PR #1, opened 2020-09-09
added jc-test hub

colliand:
Jim attempts to create a hub by editing hubs.yaml following a suggestion from Yuvi.


hubs.yaml
```diff
@@ -1,4 +1,10 @@
 hubs:
+  - name: jc-test
+   domain: jc-test.alpha.2i2c.cloud
```
yuvipanda, line 3: This isn't valid YAML since everything is offset by 1 space

PR #33, opened 2020-10-07
Only build docs on changes to docs folder

consideRatio:
I figure it makes sense to only build docs when the docs folder change. I assume there is no out of docs folder inspection though, which there may be, then we should not merge this. For example, a script that inspects content in the folder to /docs.


choldgraf:
Currently there is not any out-of-docs inspection (though there is cross-repository inspection that generates https://2i2c.org/pilot/infrastructure.html#a-list-of-current-hubs, but I don't think that's something we have control over in this repo)

PR #42, opened 2020-10-08
add quotes to fix invalid YAML?

colliand:


hubs.yaml
```diff
@@ -112,7 +112,7 @@ hubs:
               - choldgraf
               - ericvd-ucb
 - name: pfw
-    org_name: Purdue Fort Wayne
```
choldgraf, line 115: I think the issue is that `org_name` has a different indentation level from `name` (same for the following keys). THey all need to be at the same level

PR #60, opened 2020-10-08
wageningen to 2GB ram

choldgraf:
closes https://github.com/2i2c-org/pilot/issues/15


yuvipanda:
@choldgraf the current RAM *guarantee* is 256M. I think we should try keep it at about 1/4 of limit, so can you raise that to 512M?

choldgraf:
https://github.com/2i2c-org/pilot-hubs/pull/62

Issue #73, opened 2020-10-15
Remove NoVNC from the launchers


choldgraf:
I noticed that there's still a "launch desktop" launcher:

![image](https://user-images.githubusercontent.com/1839645/96163701-10ee3d80-0ecf-11eb-9f06-36fc9dbf109e.png)

but it doesn't seem to work (and I believe @yuvipanda said this is expected not to work). I looked in the environment files and couldn't figure out where it was being installed, so opening an issue that we should probably remove it

yuvipanda:
I think this is gone now?

choldgraf:
yes!

PR #79, opened 2020-10-20
Remove jupyter-desktop-server

ianabc:
This should remove jupyter-desktop-server in response to #73. If it's better to keep it and fix it please feel free to close this, I'm just trying to get familiar with the issue queue 😄 

When I build this locally I get

```
ADD failed: stat /var/lib/docker/tmp/docker-builder757307623/ipython_config.py: no such file or directory
```

Is that something I'm missing in the build process? I don't see `ipython_config.py` in images/user.

### DIFF
--- a/images/user/Dockerfile
+++ b/images/user/Dockerfile
@@ -130,5 +130,3 @@ RUN /tmp/install.R && \
 
 # Set bash as shell in terminado.
 ADD jupyter_notebook_config.py  ${CONDA_PREFIX}/etc/jupyter/
-# Disable history.
-ADD ipython_config.py ${CONDA_PREFIX}/etc/ipython/
--- a/images/user/environment.yml
+++ b/images/user/environment.yml
@@ -1,8 +1,6 @@
 channels:
   - conda-forge
 dependencies:
-  # for jupyter-desktop-server
-  - websockify
 
   # language runtimes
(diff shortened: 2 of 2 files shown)


yuvipanda:
w00p w00p, awesome @ianabc! THANK YOU!!!!! Am excited for you to poke around here :)

I think the ipython_config.py is something on my system but nowhere else, and hence the builds were succeeding before :'( Can you add a commit removing that line too? Sorry!

You'll need to build and push the image before merging this. Been trying to automate that, but have run into issues where `git log images/user` is producing the commit of the last change to that directory on my local machine, but the current commit even if it didn't touch that directory in GitHub Actions 😢

ianabc:
Cool, I removed that line from the `Dockerfile` and it builds happily now. I didn't tried pushing it anywhere, I wasn't sure if I could do that without causing problems for other people and I didn't really test the changes so I left it alone. I'll take a look through build.py and try to understand what is going on. For the moment I'm just looking for things that look easy and "probably" won't break then crossing my fingers and opening a PR! 😄 

The term is starting to settle into a pattern so I'll try to set aside some time to catch up on all the new things. Thank you!

choldgraf:
indeed @ianabc it makes me happy to see you here as well!

PR #84, opened 2020-10-23
document grafana

choldgraf:
Minimal documentation about the existence of hub grafanas. @yuvipanda does this look correct?

We'll need to eventually add some information about how to understand (and maybe create) dashboards, as well as a more principled way for gaining access. Also I couldn't find `grafana.cloudbank.2i2c.cloud` so maybe my assumption about "one grafana per cluster" is incorrect?


yuvipanda:
Ya, good to have this here - thanks @choldgraf!

PR #104, opened 2020-11-25
Add docs on kubectl & helm access to pilot hubs

yuvipanda:
Fixes https://github.com/2i2c-org/pilot-hubs/issues/75

### DIFF
--- a/docs/operate.md
+++ b/docs/operate.md
@@ -2,38 +2,75 @@
 
 Information about operating the hubs, debugging problems, and performing common actions.
 
-## Gain `kubectl` access to a hub
+## Gain `kubectl` & `helm` access to a hub
 
-Each of the hubs in the 2i2c Pilot runs on Google Cloud Platform and Kubernetes. To access the Kubernetes deployments (in order to inspect them or make changes), use the `kubectl` command line tool.
+Each of the hubs in the 2i2c Pilot runs on Google Cloud Platform and Kubernetes.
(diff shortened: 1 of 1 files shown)


choldgraf:
this looks great @yuvipanda - thanks for improving this!

Non-blocking on this PR: I was wondering what you thought about a 2i2c practice of writing *one sentence per line* in documentation. This sometimes leads to long-ish lines, but I've found that it makes patching the markdown files easier to do via UIs such as GitHub's interactive editor, and is also a fairly intuitive decision-point to use rather than "how many characters does my line have". Curious what you think about that.

yuvipanda:
Thanks @choldgraf!

She had only gone through the process of setting up for farallon hubs, nothing on Google cloud yet. We can use this as basis and modify as we go.

I mostly break lines wherever my editor does, and I think most text editors can be configured to break lines and reflow them as needed. This seems more automatic to me than breaking per sentence, since we would have to do that manually.

choldgraf:
sounds good 👍 

re: lines - the main benefit I've found is that because GitHub's interactive UI requires you to add a suggestion to only one line, rather than multiple lines, this is where having "one sentence per line" is helpful. It's nothing crucial right now (and in the future we may be able to just use an auto-formatter like https://github.com/executablebooks/mdformat) but in the future we'll need to find practices that make editing and changing the markdown as easy as possible.

PR #112, opened 2020-11-27
Allow deploying all hubs or only one hub

GeorgianaElena:
I also kept the option to deploy all hubs to be able to be fast when needed.

```bash
deploy.py deploy all-hubs
or
deploy.py deploy hub <cluster-name> <hub-name>
```

But there might be a little too many parsers and subparsers :sweat_smile: Is this ok?

Closes https://github.com/2i2c-org/pilot-hubs/issues/110

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -57,7 +57,7 @@ def build():
             cluster.build_image()
 
 
-def deploy(cluster_name=None, hub_name=None):
+def deploy(cluster_name, hub_name):
     """
     Deploy all hubs in all clusters
     """
@@ -112,29 +112,15 @@ def main():
     build_parser = subparsers.add_parser('build')
     deploy_parser = subparsers.add_parser('deploy')
 
-    deploy_subparsers = deploy_parser.add_subparsers(dest='mode')
-
-    all_parser = deploy_subparsers.add_parser(
-        'all-hubs',
-        help='Deploy all the hubs'
-    )
-    hub_parser = deploy_subparsers.add_parser(
-        'hub',
-        help='Only deploy a specific hub in a cluster'
-    )
-
-    hub_parser.add_argument('cluster_name')
-    hub_parser.add_argument('hub_name')
+    deploy_parser.add_argument('cluster_name', nargs="?")
+    deploy_parser.add_argument('hub_name', nargs="?")
 
     args = argparser.parse_args()
 
     if args.action == 'build':
         build()
     elif args.action == 'deploy':
-        if args.mode == 'all-hubs':
-            deploy()
-        else:
-            deploy(args.cluster_name, args.hub_name)
+        deploy(args.cluster_name, args.hub_name)
 
 if __name__ == '__main__':
     main()
--- a/deploy.py
+++ b/deploy.py
@@ -29,6 +29,24 @@ def parse_clusters():
     ]
 
 
+def find_cluster_by_name(cluster_name, clusters):
+    """
+    Find a cluster by name in a list of Cluster objects
+    """
+    for cluster in clusters:
+        if cluster.spec['name'] == cluster_name:
+            return cluster
+
+
+def find_hub_in_cluster(hub_name, hubs):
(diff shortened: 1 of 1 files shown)


GeorgianaElena:
We also need to update GitHub Actions to use the new deploy flags once we decide their final form.

deploy.py
```diff
@@ -68,27 +68,47 @@ def deploy():
     # proxy.secretTokens have leaked. So let's be careful with that!
     PROXY_SECRET_KEY = bytes.fromhex(os.environ['PROXY_SECRET_KEY'])
 
-    clusters = parse_clusters()
-    for cluster in clusters:
+    if cluster and hub:
         with cluster.auth():
-            for hub in cluster.hubs:
-                hub.deploy(k, PROXY_SECRET_KEY)
+            hub.deploy(k, PROXY_SECRET_KEY)
```
yuvipanda, line 73: Cluster & Hub are strings here, right? We'd have to make them cluster & hub objects somehow.

deploy.py
```diff
@@ -68,27 +68,47 @@ def deploy():
...
-        'action',
-        choices=['build', 'deploy']
+    subparsers = argparser.add_subparsers(dest='action')
+
+    build_parser = subparsers.add_parser('build')
+    deploy_parser = subparsers.add_parser('deploy')
+
+    deploy_subparsers = deploy_parser.add_subparsers(dest='mode')
+
+    all_parser = deploy_subparsers.add_parser(
```
yuvipanda, line 91: How about the following commandline forms:

`python3 deploy.py deploy` deploys all clusters and hubs

`python3 deploy.py deploy <cluster-name>` deploys all hubs in that cluster

`python3 deploy.py deploy <cluster-name> <hub-name>` deploys just that one hub.

Maybe by adding cluster name & hub name as optional arguments to the deploy parser?

yuvipanda (changes requested):
Thanks, @GeorgianaElena!

deploy.py
GeorgianaElena: I just added some search functions. Sorry for missing this :(

deploy.py
GeorgianaElena: Perfect! :+1:

choldgraf:
Just a quick note - we should document this! We should treat most 2i2c infrastructure as if it were an open-source project - API docs, intro docs, overviews, etc, that way others on the team or in the community can quickly get oriented if they need to.

At the least we should open an issue to document how / when to use this functionality 👍

yuvipanda:
@choldgraf opened https://github.com/2i2c-org/pilot-hubs/issues/113 for documentation

deploy.py
```diff
@@ -29,6 +29,24 @@ def parse_clusters():
     ]
 
 
+def find_cluster(cluster_name, clusters):
```
yuvipanda, line 32: Since these are called only in one place right now, I think we should inline them instead of splitting it out into a function. We can refactor it later if needed.

deploy.py
```diff
@@ -29,6 +29,24 @@ def parse_clusters():
...
+    """
+    Find a cluster by name in a list of Cluster objects
+    """
+    for cluster in clusters:
+        if cluster.spec['name'] == cluster_name:
+            return cluster
+
+
+def find_hub(hub_name, hubs):
```
yuvipanda, line 41: Same as for find_cluster

deploy.py
yuvipanda: If you're feeling fancy, you can also use https://stackoverflow.com/a/34894322 instead of a for loop :)

GeorgianaElena: ty <3

GeorgianaElena:
I believe this is safe to be merged now. I've used it successfully to deploy https://github.com/2i2c-org/pilot-hubs/pull/106 to the staging hub :tada:

yuvipanda:
Thanks a lot, @GeorgianaElena!

PR #118, opened 2020-12-03
Don't set OAUTH_CALLBACK_URL env var

yuvipanda:
- A typo meant we never actually set this
- Things continue to work without it, so let's clear it out

### DIFF
--- a/hub.py
+++ b/hub.py
@@ -141,11 +141,6 @@ class Hub:
                         'name': self.cluster.spec['image_repo']
                     },
                 },
-                'hub': {
-                    'extraaEnv': {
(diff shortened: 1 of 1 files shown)


GeorgianaElena (approved):
I see no reason in removing something that's not being used.

For deploying, the oauth callback is set in the `KeyProvider` class, but I'm wondering how is the redirect happening after a manual login without the callback. Is there a default callback set somewhere?

yuvipanda:
I think it's set [here](https://github.com/2i2c-org/pilot-hubs/blob/master/auth.py#L43). And auth0 just redirects to the default callback_url.

PR #119, opened 2020-12-03
Explicitly create NFS share during deploy

### LINKED ISSUE #83: FS Mounting error when logging in on a new hub
I just created `demo.cloudbank.2i2c.cloud` and tried to log in, and ran into this error when starting up the server. It seems related to mounting the filesystem:

```
2020-10-22T18:49:55Z [Warning] MountVolume.SetUp failed for volume "demo-home-nfs" : mount failed: exit status 1 Mounting command: systemd-run Mounting arguments: --description=Kubernetes transient mount for /var/lib/kubelet/pods/716b4d9d-dddd-4cf4-a791-ab83ca26a832/volumes/kubernetes.io~nfs/demo-home-nfs --scope -- /home/kubernetes/containerized_mounter/mounter mount -t nfs -o noatime,soft,vers=4.2 nfs-server-01:/export/home-01/homes/demo /var/lib/kubelet/pods/716b4d9d-dddd-4cf4-a791-ab83ca26a832/volumes/kubernetes.io~nfs/demo-home-nfs Output: Running scope as unit: run-r064b93d0547b4c2284a02ac58f794e2f.scope Mount failed: mount failed: exit status 32 Mounting command: chroot Mounting arguments: [/home/kubernetes/containerized_mounter/rootfs mount -t nfs -o noatime,soft,vers=4.2 nfs-server-01:/export/home-01/homes/demo /var/lib/kubelet/pods/716b4d9d-dddd-4cf4-a791-ab83ca26a832/volumes/kubernetes.io~nfs/demo-home-nfs] Output: mount.nfs: mounting nfs-server-01:/export/home-01/homes/demo failed, reason given by server: No such file or directory
```

# To fix this

Until we get an auto-fix, here are the steps to fix this for a new hub:

1. Log in to the NFS Server (called `nfs-server-01`).

   If you have `gcloud` installed, you can do this with `gcloud compute ssh nfs-server-01` after logging in to gcloud
2. Create a directory named `/export/home-01/homes/<hub-name>`:

   `sudo mkdir /export/home-01/homes/<hub-name>`
3. Make it be owned by the 'ubuntu' user

   `sudo chown ubuntu:ubuntu /export/home-01/homes/<hub-name>`

yuvipanda:
Fixes #83

### DIFF
--- a/hubs.yaml
+++ b/hubs.yaml
@@ -7,6 +7,9 @@ clusters:
       project: two-eye-two-see
       cluster: low-touch-hubs-cluster
       zone: us-central1
+      nfs:
+        server: nfs-server-01
+        zone: us-central1-b
     hubs:
       - name: staging
         cluster: 2i2c
@@ -279,6 +282,9 @@ clusters:
       project: cb-1003-1696
       cluster: cloudbank-2i2c-cluster
       zone: us-central1-b
+      nfs:
+        server: nfs-server-01
+        zone: us-central1-b
     hubs:
       - name: spelman
         domain: spelman.cloudbank.2i2c.cloud
--- a/hubs.yaml
+++ b/hubs.yaml
@@ -6,7 +6,7 @@ clusters:
       key: secrets/2i2c.json
       project: two-eye-two-see
       cluster: low-touch-hubs-cluster
-      zone: us-central1
+      region: us-central1
       nfs:
         server: nfs-server-01
         zone: us-central1-b
--- a/hub.py
+++ b/hub.py
@@ -81,7 +81,9 @@ class Cluster:
         config = self.spec['gcp']
         key_path = config['key']
         project = config['project']
-        zone = config['zone']
+        # If cluster is regional, it'll have a `region` key set.
+        # Else, it'll just have a `zone` key set. Let's respect either.
+        location = config.get('zone', config.get('region'))
         cluster = config['cluster']
 
         with decrypt_file(key_path) as decrypted_key_path:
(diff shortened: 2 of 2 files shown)


yuvipanda:
Should be easier to review once #118 is merged.

hubs.yaml
```diff
@@ -7,6 +7,9 @@ clusters:
       project: two-eye-two-see
       cluster: low-touch-hubs-cluster
       zone: us-central1
+      nfs:
+        server: nfs-server-01
+        zone: us-central1-b
```
GeorgianaElena, line 12: I'm not sure about this, but shouldn't the dir be created in the zone where the project is running? 

```suggestion
        zone: us-central1
```

yuvipanda: So GKE clusters can be [regional or zonal](https://cloud.google.com/kubernetes-engine/docs/concepts/types-of-clusters). Regional clusters have masters running in 3 zones in the same region for resiliency, but cost more. In some cases, it's worth it and in others it isn't. I've now updated the code with a commit that specifies that more explicitly for the cluster.

Our NFS server is just a single machine though, and they only exist in one zone. Even though our cluster is regional, all our user nodes are only in one zone (us-central1-b), so our NFS machine is there too.

You can see some of the config that creates all this [in our terraform](https://github.com/2i2c-org/pilot-hubs/blob/master/terraform/main.tf#L43)

GeorgianaElena:
TY <3

GeorgianaElena:
@yuvipanda, I just noticed that the deploy failed for this PR with:

```
ERROR: (gcloud.compute.ssh) Could not add SSH key to instance metadata:
 - Required 'compute.instances.setMetadata' permission for 'projects/cb-1003-1696/zones/us-central1-b/instances/nfs-server-01'
```

yuvipanda:
Hmm, so it worked for the 2i2c google cloud project but not for the cloudbank one.

Issue #90, opened 2020-10-28
Set limits & resource requests on all core pods


choldgraf:
### Description

*All* 'core' pods should have CPU & memory requests and limits set.

### Benefit

Setting explicit resources makes sure that critical processes / pods don't run out of resources and die, as has happened in cases like https://github.com/2i2c-org/pilot-hubs/issues/526. This helps prevent major disruptions to the service.

### Tasks to complete

- [ ] Make sure all components from support/ chart have requests and limits set
- [ ] Make sure all hub components have requests and limits set
- [ ] Make sure all dask-gateway components have requests and limits set

PR #120, opened 2020-12-03
Limit prometheus resource consumption

yuvipanda:
It's on the same nodes as the core pods, and if left to its
own devices it can take up all the resources on the nodes.

Ref #90


GeorgianaElena (approved):
The limits seem to be appropriate in my view too, so merging this <3

PR #121, opened 2020-12-03
Allow overriding more hub options from hubs.yaml

yuvipanda:
- Make hubs.yaml z2jh config top priority
- Allow disabling auth0 client creation

### DIFF
--- a/hub.py
+++ b/hub.py
@@ -172,21 +172,25 @@ class Hub:
         """
         Deploy this hub
         """
-        client = auth_provider.ensure_client(
-            self.spec['name'],
-            self.spec['domain'],
-            self.spec['auth0']['connection']
-        )
 
         proxy_secret = hmac.new(proxy_secret_key, self.spec['name'].encode(), hashlib.sha256).hexdigest()
+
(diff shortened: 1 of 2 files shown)


yuvipanda:
Should be easier to review once #119 is merged

hub.py
```diff
@@ -144,26 +149,51 @@ def get_generated_config(self):
...
             'jupyterhub':  {
                 'proxy': {
                     'secretToken': proxy_secret
                 },
-                'auth': auth_provider.get_client_creds(client, self.spec['auth0']['connection'])
             }
         }
 
+        # Allow explicilty ignoring auth0 setup
+        if self.spec['auth0'].get('enabled', True):
```
GeorgianaElena, line 187: So, with this we're ignoring the auth0 setup only for the hubs that have explicitly set `aut0.enabled` False, right?

GeorgianaElena (approved):

hub.py
yuvipanda: Yep, exactly! It defaults to 'true'

PR #122, opened 2020-12-03
Grant SA permission to login to NFS Server

yuvipanda:
These permissions are too lax, should migrate to
using
https://cloud.google.com/compute/docs/instances/managing-instance-access
instead - at least.


yuvipanda:
This should fix the permission fail in https://github.com/2i2c-org/pilot-hubs/runs/1492442960?check_suite_focus=true. I've deployed it as well, should be ok.

GeorgianaElena:
Thank you <3

Issue #114, opened 2020-12-02
Allow deploying multiple *kinds* of hubs


yuvipanda:
We deploy all the hubs specified in `hubs.yaml` from one deployment chart defined [here](https://github.com/2i2c-org/pilot-hubs/tree/master/hub). Multiple hubs types can be supported with different deployment charts.

We should steal a version of the upstream [pangeo-cloud-federatio](https://github.com/pangeo-data/pangeo-cloud-federation/tree/staging/pangeo-deploy) deployment chart, allowing us to deploy a PANGEO-style hub with the same workflow as our educational hubs.

yuvipanda:
This works now!

PR #123, opened 2020-12-03
Add an 'ephemeral' hub template

yuvipanda:
Offering multiple 'hub templates' for deployment makes life
much easier for us & our users. We can represent each type with
its own deployment chart, and have them depend on each other
to form a hierarchy of them.

Currently, it is only possible to hade a one-level hierarchy,
since `base-hub` is hardcoded in the code. This is config that
must be separated.

We add the 'ephemeral-hub' template, and make the ephemeral hub
use it.

The ephemeral hub has the following features:

- Template for a temporary, transient binder-style hub
- Same image (for now), but no authentication
- Resource limitations via memory / CPU limits, and maximum number
  of concurrent user servers
- More aggressive culling
- No persistent storage at all
- The home page template doesn't get rendered at all.

Ref #114


hub.py
```diff
@@ -114,23 +114,26 @@ def __init__(self, cluster, spec):
...
 
-        Shouldn't have anything secret here.
+        WARNING: CONTAINS SECRET VALUES!
         """
 
-        return {
+        proxy_secret = hmac.new(proxy_secret_key, self.spec['name'].encode(), hashlib.sha256).hexdigest()
+
+        generated_config = {
             'nfsPVC': {
```
GeorgianaElena, line 130: This will get overwritten by hubs.yaml for the ephemeral hub right?

hub.py
```diff
@@ -168,49 +183,38 @@ def setup_nfs_share(self):
...
-        # Allow explicilty ignoring auth0 setup
-        if self.spec['auth0'].get('enabled', True):
-            client = auth_provider.ensure_client(
-                self.spec['name'],
-                self.spec['domain'],
-                self.spec['auth0']['connection']
-            )
-            secret_values['jupyterhub']['auth'] = auth_provider.get_client_creds(client, self.spec['auth0']['connection'])
 
+        # FIXME: Don't do this for ephemeral hubs
```
GeorgianaElena, line 201: Would a template check similar with the one above, be enough?
```python
if hub_template != `ephemeral-hub`:
    self.setup_nfs_share()
```

hub.py
yuvipanda: from ephemeral-hub's values.yaml, we'll set `nfsPVC.enabled` to false, which will ignore this config. However, it would be nice to not need that - filed https://github.com/2i2c-org/pilot-hubs/issues/126 to tackle that.

hub.py
yuvipanda: Possibly, but I'd instead like to add an explicit flag to enable / disable NFS. See https://github.com/2i2c-org/pilot-hubs/issues/126

GeorgianaElena:
Thanks @yuvipanda! <3

PR #130, opened 2020-12-07
Fetch hub specific template from pilot-homepage

GeorgianaElena:
Fixes https://github.com/2i2c-org/pilot-hubs/issues/111

### DIFF
--- a/hub.py
+++ b/hub.py
@@ -168,8 +168,10 @@ class Hub:
                                 'allowPrivilegeEscalation': False,
                                 'readOnlyRootFilesystem': True,
                                 'volumeMounts': [
-                                    'name': 'custom-templates',
-                                    'mountPath': '/srv/repo'
+                                    {
+                                        'name': 'custom-templates',
+                                        'mountPath': '/srv/repo'
+                                    }
                                 ]
                             }
                         }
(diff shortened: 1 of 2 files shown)


hub.py
```diff
@@ -149,7 +149,20 @@ def get_generated_config(self, auth_provider, proxy_secret_key):
                         'name': self.cluster.spec['image_repo']
                     },
                 },
-            }
+                'hub': {
+                    'extraContainers': {
+                        'templates-sync': {
+                            'args':
+                                '-c',
+                                f'while true; do git fetch origin; \
```
yuvipanda, line 157: Consider putting this out into its own triple quoted string (possibly with a [dedent](https://docs.python.org/3/library/textwrap.html#textwrap.dedent)) instead of it being split out here.

We've had to substitute the hub name in a few places - here, nfs share name, etc. We should open an issue about being able to do that declaratively, so we could've left this in values.yaml.

hub.py
```diff
@@ -149,7 +149,20 @@ def get_generated_config(self, auth_provider, proxy_secret_key):
                         'name': self.cluster.spec['image_repo']
                     },
                 },
-            }
+                'hub': {
+                    'extraContainers': {
```
yuvipanda, line 153: extraContainers takes a list, not a dictionary, right?

yuvipanda (changes requested):
If we move this here, let's just move everything about home page / templates to python? Instead of mixing code in YAML & Python.

yuvipanda (approved):

yuvipanda:
Thanks, @GeorgianaElena!

PR #132, opened 2020-12-08
Move helm dependencies to Chart.yaml

### LINKED ISSUE #129: Remove requirements.yaml file from hub templates
A separate `Chart.yaml` and `requirements.yaml` were required in helm2.
But starting from helm3, `requirements.yaml` is redundant - it has
been merged with `Chart.yaml` - see https://helm.sh/docs/topics/charts/

We should just get rid of requirements.yaml.

yuvipanda:
Fixes #129

### DIFF
--- a/hub-templates/base-hub/Chart.yaml
+++ b/hub-templates/base-hub/Chart.yaml
@@ -2,7 +2,8 @@ apiVersion: v2
 appVersion: '1.0'
 description: Deployment Chart for JupyterHub
 name: base-hub
-version: v0.0.1
+# Let's keep this constant so other charts in this repo can depend on this easily
+version: 0.0.1
 dependencies:
  - name: jupyterhub
    # REMEMBER TO CHANGE BASE IMAGE OF images/hub/ WHEN CHANGING THIS
--- a/hub-templates/base-hub/Chart.yaml
+++ b/hub-templates/base-hub/Chart.yaml
@@ -3,3 +3,8 @@ appVersion: '1.0'
 description: Deployment Chart for JupyterHub
 name: base-hub
 version: v0.0.1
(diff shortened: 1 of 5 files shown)


hub-templates/base-hub/Chart.yaml
```diff
@@ -3,3 +3,8 @@ appVersion: '1.0'
 description: Deployment Chart for JupyterHub
 name: base-hub
 version: v0.0.1
+dependencies:
+ - name: jupyterhub
+   # REMEMBER TO CHANGE BASE IMAGE OF images/hub/ WHEN CHANGING THIS
+   version: 0.9.0-n287.hd5f1f881
+   repository: https://jupyterhub.github.io/helm-chart/
```
consideRatio, line 10: This transition will require a `apiVersion: v2` in the top of the Chart.yaml file I think. That change will also require Helm3, but Helm2 is end of life etc so go for it is my sentiment!

consideRatio:
Suggested changes
1. `apiVersion: 2` in Chart.yaml (required i think)
2. SemVer2 valid chart versions, which means to not lead with a `v` in the version field (required at least if `helm package` is run, maybe otherwise also).
3. Replace the `+` sign in the support/Chart.yaml file with something else, for example by following [the pattern Chartpress does](https://github.com/jupyterhub/chartpress#examples-chart-versions-and-image-tags), because Helm will treat what follows behind `+` as things of little importance, which may cause issues for example when using `helm package` to a Helm chart repo.

yuvipanda:
Thanks a lot for the review, @consideRatio! I've addressed them appropriately, I'd hope.

consideRatio:
Nice LGTM!

PR #134, opened 2020-12-10
Add 'daskhub' template

yuvipanda:
- Allows deploying new dask-enabled hubs with `hubs.yaml`
  workflow!
- Installs dask-gateway alongside our basic hub. We're using
  the 'dask-gateway' chart, but should move to the 'daskhub'
  chart instead at some point soon.
- Adds a 'dask-staging.pilot.2i2c.cloud' hub that has dask
  enabled!
- Defaults to a pangeo image, since we don't want to be curating
  these images ourselves.
- Factors out a function for making changes to the generated
  config. This is terrible and should be fixed.
- Put dask workers on preemptible nodes, since they
  should be resilient to lost workers
- Use e2 nodes instead of n1 nodes, since they are
  cheaper when the sustained use discount isn't applied
- Pin node versions in terraform, so it doesn't try to
  upgrade them each time we run terraform apply!
- Make catalyst coop a dask hub.
- Enable outgoing ssh for all daskhubs, since they are more
  likely to be research-oriented & need ssh push.

Ref #114

### DIFF
--- a/hub-templates/base-hub/values.yaml
+++ b/hub-templates/base-hub/values.yaml
@@ -38,6 +38,8 @@ jupyterhub:
       userPlaceholderPriority: -10
     userScheduler:
       enabled: true
+      nodeSelector:
+        hub.jupyter.org/pool-name: core-pool
       resources:
         requests:
           # FIXME: Just unset this?
@@ -54,6 +56,8 @@ jupyterhub:
     service:
       type: ClusterIP
     chp:
+      nodeSelector:
+        hub.jupyter.org/pool-name: core-pool
       resources:
         requests:
           # FIXME: We want no guarantees here!!!
@@ -63,6 +67,8 @@ jupyterhub:
         limits:
           memory: 1Gi
     traefik:
+      nodeSelector:
+        hub.jupyter.org/pool-name: core-pool
       resources:
         requests:
           memory: 64Mi
@@ -128,6 +134,8 @@ jupyterhub:
   auth:
     type: dummy
   hub:
+    nodeSelector:
+      hub.jupyter.org/pool-name: core-pool
     networkPolicy:
       enabled: true
       ingress:
--- a/hub-templates/daskhub/Chart.yaml
+++ b/hub-templates/daskhub/Chart.yaml
@@ -1,0 +1,12 @@
+apiVersion: v2
+appVersion: '1.0'
+description: Deployment Chart for an ephemeral JupyterHub
+name: ephemeral-hub
+version: 0.0.1-n2724.h1dfae31
+dependencies:
+ - name: base-hub
+   version: 0.0.1
+   repository: file://../base-hub
+ - name: dask-gateway
+   version: "0.9.0"
+   repository: 'https://dask.org/dask-gateway-helm-repo/'
--- a/hub-templates/daskhub/Chart.yaml
+++ b/hub-templates/daskhub/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 appVersion: '1.0'
-description: Deployment Chart for an ephemeral JupyterHub
-name: ephemeral-hub
+description: Deployment Chart for a dask-enabled JupyterHub
+name: daskhub
 version: 0.0.1-n2724.h1dfae31
 dependencies:
  - name: base-hub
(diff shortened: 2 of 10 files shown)


yuvipanda:
I want to move to using daskhub helm chart for this, but that requires a refactor of how values are currently generated.

I want to control:

1. The version of JupyterHub chart being used (mostly be on master)
2. The common config for JupyterHub itself (proxySecrets, auth, etc)

Currently, we've done this by making charts inherit from base-chart, making it difficult to start from an external, new chart like daskhub. This is fixable, and we should fix it. But I don't want to block this PR on that refactoring.

consideRatio:
@yuvipanda just a mention of a future solution not applicable for use now in this PR afaik, is that the z2jh can start supporting dynamic creation of proxy.secretToken etc with helm 3's `lookup` function. See https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1910.

yuvipanda:
@consideRatio YESSS I'm looking forward to that! That'll also remove need to do the secrets manually for daskhub either.

hub-templates/daskhub/Chart.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v2
+appVersion: '1.0'
+description: Deployment Chart for an ephemeral JupyterHub
+name: ephemeral-hub
```
GeorgianaElena, line 4:
```suggestion
description: Deployment Chart for a dask-enabled JupyterHub
name: daskhub
```
I think this new daskhub template is being referred to as `daskhub` in `hubs.yaml`

hub.py
```diff
@@ -261,20 +261,51 @@ def setup_nfs_share(self):
...
         # FIXME: Have a templates config somewhere? Maybe in Chart.yaml
         # FIXME: This is a hack. Fix it.
         if hub_template != 'base-hub':
-            generated_values = {
-                'base-hub': generated_values
+            generated_config = {
+                'base-hub': generated_config
+            }
+
+        # LOLSOB FIXME
```
GeorgianaElena, line 284: Does #LOLSOB mean :smile:  + :sob:? hehe
Should we open an issue about fixing this later to not block the PR?

GeorgianaElena (approved):
Thank you, @yuvipanda <3 I reviewed this to the best of my understanding and I believe it's LGTM.

I noticed the new hub has been deployed and it's working :tada: I spotted the same `nb-conda-kernels` doubled kernels behavior though.
kernels

yuvipanda:
Thanks a lot, @GeorgianaElena! I just pushed a commit just after you responded here, unfortunately.

Yeah, we should get to the bottom of the nb-conda-kernels behavior!

hub-templates/base-hub/values.yaml
```diff
@@ -128,6 +134,8 @@ jupyterhub:
   auth:
     type: dummy
   hub:
+    nodeSelector:
+      hub.jupyter.org/pool-name: core-pool
```
consideRatio, line 138: The configuration scheduling.corePods.nodeAffinity.matchNodePurpose=require will make all core pods of the jupyterhub chart have a required node affinity toward the node label `hub.jupyter.org/node-purpose=core`.

By using hub.jupyter.org/pool-name it seems like there is a predefined meaning of that label and can cause some confusion, it isn't having a predefined meaning right? If it is our meaning, we should do something like 2i2c.org/pool-name instead I think.

yuvipanda: Great catch, @consideRatio! I opened https://github.com/2i2c-org/pilot-hubs/issues/136 to track this, since changing the labels will require rotating node pools.

choldgraf:
this is very exciting! (that's all I have to say)

TomAugspurger:
>  We're using the 'dask-gateway' chart, but should move to the 'daskhub' chart instead at some point soon.

FWIW, I'd be curious to see if others have thoughts on how to better "compose" these kinds of setups. I ran into a situation where I wanted JupyterHub + Dask Gateway + [persistent binder](https://github.com/gesiscss/persistent_binderhub). Each of daskhub & persistent binder want to "wrap" the jupyterhub chart, but only one can. I'm not sure if there's actually a better solution out there, but maybe others have thoughts.

yuvipanda:
@TomAugspurger yeah, I'm not sure at all :| This might be one of the places where something like [jsonnet](https://jsonnet.org/) would've been useful for helm, but alas.

We'll probably need to write python tooling around it.

GeorgianaElena:
Thanks for this great work, @yuvipanda :tada:

yuvipanda:
Thanks for the review, @GeorgianaElena!

PR #141, opened 2020-12-16
updating funded by

choldgraf:


yuvipanda:
Should this be merged, @choldgraf?

choldgraf:
actually I think it can be closed, unless I'm missing something it seems like these have already been merged into `hubs.yaml`...confusing 😕

PR #148, opened 2020-12-21
Validate hub health after each deployment

GeorgianaElena:
Closes https://github.com/2i2c-org/pilot-hubs/issues/138

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -39,7 +39,7 @@ def build():
             cluster.build_image()
 
 
-def deploy(cluster_name, hub_name, test_notebook):
+def deploy(cluster_name, hub_name, test_notebook_path):
     """
     Deploy all hubs in all clusters
     """
@@ -76,15 +76,15 @@ def deploy(cluster_name, hub_name, test_notebook):
             hubs = cluster.hubs
             if hub_name:
                 hub = next((hub for hub in hubs if hub.spec['name'] == hub_name), None)
-                hub.deploy(k, PROXY_SECRET_KEY, test_notebook)
+                hub.deploy(k, PROXY_SECRET_KEY, test_notebook_path)
             else:
                 for hub in hubs:
-                    hub.deploy(k, PROXY_SECRET_KEY, test_notebook)
+                    hub.deploy(k, PROXY_SECRET_KEY, test_notebook_path)
     else:
         for cluster in clusters:
             with cluster.auth():
                 for hub in cluster.hubs:
-                    hub.deploy(k, PROXY_SECRET_KEY, test_notebook)
+                    hub.deploy(k, PROXY_SECRET_KEY, test_notebook_path)
 
 
 def main():
@@ -96,14 +96,14 @@ def main():
 
     deploy_parser.add_argument('cluster_name', nargs="?")
     deploy_parser.add_argument('hub_name', nargs="?")
-    deploy_parser.add_argument('--test_notebook')
+    deploy_parser.add_argument('--test-notebook-path')
 
     args = argparser.parse_args()
 
     if args.action == 'build':
         build()
     elif args.action == 'deploy':
-        deploy(args.cluster_name, args.hub_name, args.test_notebook)
+        deploy(args.cluster_name, args.hub_name, args.test_notebook_path)
 
 if __name__ == '__main__':
     main()
--- a/deploy.py
+++ b/deploy.py
@@ -9,6 +9,7 @@ import hashlib
 import hmac
 import os
 import argparse
+import asyncio
 
 from auth import KeyProvider
 from hub import Hub, Cluster
@@ -39,7 +40,7 @@ def build():
             cluster.build_image()
 
 
-def deploy(cluster_name, hub_name, test_notebook_path):
+async def deploy(cluster_name, hub_name, test_notebook_path):
     """
     Deploy all hubs in all clusters
     """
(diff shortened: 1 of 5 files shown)


deploy.py
```diff
@@ -96,13 +96,14 @@ def main():
 
     deploy_parser.add_argument('cluster_name', nargs="?")
     deploy_parser.add_argument('hub_name', nargs="?")
+    deploy_parser.add_argument('--test_notebook')
```
yuvipanda, line 99: How about `--test-notebook` or `--test-ipynb-path` or `--test-notebook-path` instead? I generally like `-` separators than `_` separators for named arguments like this. argparse automatically converts them to `_` so we can access them the same.

hub.py
```diff
@@ -260,6 +260,22 @@ def setup_nfs_share(self):
...
+        Automatically create a temporary user, start their server and run a test notebook, making
+        sure it runs to completion. Stop and delete the test server at the end. Try this steps twice
+        before declaring it a failure. If any of these steps fails, immediately halt further
+        deployments and error out.
+        """
+        hub_url = f'https://{self.spec["domain"]}'
+        # Export the hub health check service as an env var so that jupyterhub_client can read it.
+        os.environ['JUPYTERHUB_API_TOKEN']=service_api_token
+        subprocess.check_call([
+            'jhubctl', 'run', '--temporary-user', '--notebook', test_notebook, '--hub', hub_url
```
yuvipanda, line 276: do you think we can call out to the jupyterhub_client library directly instead of running a subprocess? It's ok if we wanna use the subprocess call just now, but we should move away from that soon.

hub.py
```diff
@@ -260,6 +260,22 @@ def setup_nfs_share(self):
...
+        After each hub gets deployed, validate that it 'works'.
+
+        Automatically create a temporary user, start their server and run a test notebook, making
+        sure it runs to completion. Stop and delete the test server at the end. Try this steps twice
+        before declaring it a failure. If any of these steps fails, immediately halt further
+        deployments and error out.
+        """
+        hub_url = f'https://{self.spec["domain"]}'
+        # Export the hub health check service as an env var so that jupyterhub_client can read it.
+        os.environ['JUPYTERHUB_API_TOKEN']=service_api_token
```
yuvipanda, line 274: `subprocess.check_call` can take an explicit dictionary of environment variables. To prevent accidental issues caused by setting `os.environ` in our *own* process, let's instead set the environment of the subprocess explicitly?

hub.py
```diff
@@ -260,6 +260,22 @@ def setup_nfs_share(self):
...
+        subprocess.check_call([
```
yuvipanda, line 275: We should probably also catch the exception here if it fails, and return True or False. This allows us to replace the internal implementation detail of this method later without having to modify all the callers.

hub.py
```diff
@@ -325,6 +351,27 @@ def deploy(self, auth_provider, proxy_secret_key):
...
+            if test_notebook:
+                try_idx = 1
+                if self.spec['template'] != 'base-hub':
+                    service_api_token = generated_values["base-hub"]["jupyterhub"]["hub"]["services"]["hub-health"]["apiToken"]
+                else:
+                    service_api_token = generated_values["jupyterhub"]["hub"]["services"]["hub-health"]["apiToken"]
+                # Try 2 times before declaring it a failure
+                while try_idx <= 2:
+                    try:
+                        print(f"Validate hub health try {try_idx} started...")
```
yuvipanda, line 367: This is ok for now, but let's add a FIXME to move to something like https://github.com/litl/backoff? Doing retries properly is complicated and hard, we should just re-use other people's work :D

requirements.txt
```diff
@@ -1,2 +1,3 @@
 ruamel.yaml
-auth0-python
\ No newline at end of file
+auth0-python
+git+https://github.com/GeorgianaElena/jupyterhub_client.git@user-token
```
yuvipanda, line 3: This is pinned to a branch (`user-token`). Instead, can you pin to a git commit hash? That way, we know exactly what's installed. And we can bump it when your PR gets merged, or when you add a new commit.

yuvipanda (changes requested):
This is awesome work, @GeorgianaElena! I've left some inline comments.

How do you think we can test this?

GeorgianaElena:
Thanks for the feedback @yuvipanda! I believe I have addressed all the inline comments.
I tested this manually by deploying it to the staging hub and validating the hub's health, though I didn't check it for all the hubs. I was thinking we could create a GitHub action that deploys the changes in a PR on staging and checks that hub's health. What do you think?

hub.py
yuvipanda: Is this needed by execute_notebook? Ideally this would be passed as a parameter. If it only reads from the environment variable, we should make sure the original value (if it exists) is restored once the call is done, with a `try/finally` block. See https://github.com/yuvipanda/hubploy/blob/master/hubploy/auth.py#L66 for an example.

requirements.txt
```diff
@@ -1,2 +1,5 @@
+asyncio
```
yuvipanda, line 1: asyncio is built into python now, don't need to explicitly install it.

yuvipanda (approved):
Thanks for addressing the comments. Making this `async` is very exciting!

I explicitly broke the staging cluster by doing this:

```
diff --git a/hub-templates/base-hub/values.yaml b/hub-templates/base-hub/values.yaml
index 73a8189..481da05 100644
--- a/hub-templates/base-hub/values.yaml
+++ b/hub-templates/base-hub/values.yaml
@@ -101,7 +101,7 @@ jupyterhub:
       hub.jupyter.org/pool-name: user-pool
     image:
       name: set_automatically_by_automation
-      tag: 4066a62
+      tag: 4066a6
     storage:
       type: static
       static:
```

Just removed a character from the docker image tag. Even though the deploy succeeded, staging breaks with the following error:

image

3

hub.py
GeorgianaElena: Unfortunately we can't pass the token as a parameter to this function :( (params received by the [`execute_code` function](https://github.com/GeorgianaElena/jupyterhub_client/blob/user-token/jupyterhub_client/execute.py#L45), which is called by e`xecute_notebook` [here](https://github.com/GeorgianaElena/jupyterhub_client/blob/user-token/jupyterhub_client/execute.py#L89))

I followed the example in` hubploy` (thanks <3) and just commited the changes.

GeorgianaElena:
Thanks @yuvipanda  <3
> Just removed a character from the docker image tag. Even though the deploy succeeded, staging breaks with the following error:
My intuition is that some error check needs to be fixed somewhere.
We should also print messages when the health checks are happening, so we know what it is doing.

I just tried the same thing with the current state of the PR (more log messages added and we're also catching `TimeoutError` exeptions now) and this is the output I get in the terminal:
err-staging

Do you think there's enough info here? Should I add anything else?

We should deploy before checking the hub's health, right? Otherwise the hub wouldn't be up able to be checked?
The idea is to not deploy to other prod hubs, no?

yuvipanda:
Ah, so I wasn't passing `--test-notebook-path`, so I wasn't seeing it. With that and catching the TimeoutError, this works! ❤️ I've filed https://github.com/2i2c-org/pilot-hubs/issues/149 for an improvement, but this can be merged now.

Thanks a lot, @GeorgianaElena!!!! <3

PR #154, opened 2020-12-24
Stop accidentally overriding hub config

### LINKED ISSUE #150: Fix home page templates for hubs
The current dask-staging.pilot.2i2c.cloud website looks like this:

image

the home-page setup is somehow not going through. We need to fix that.

yuvipanda:
The logic for 'fixing' the config for various hub types
was clobbering some config we had set earlier - particularly
for the front page & dask-gateway service. This fixes them
temporarily. We should really find a better way to do config.

Fixes #150


GeorgianaElena (approved):
Thanks for catching and fixing this @yuvipanda <3 I didn't notice that the addition of the hub health service had overwritten the config and  messed the nice home page we had for the hubs :(

PR #157, opened 2020-12-30
Add tests

GeorgianaElena:
First step towards https://github.com/2i2c-org/pilot-hubs/issues/155


yuvipanda:
Ah, this is great! I haven't reviewed this fully yet, but I've a suggested tweak to the workflow. We run the py.test based tests against *every* hub as it's deployed, not just on staging hubs. This way, we can make sure that all the hubs are working - with their individual customizations. It'll also cause the GitHub action to fail whenever a hub fails to check - we can use this to our advantage by specifying the staging & dask-staging hubs first in `hubs.yaml`. What do you think?

GeorgianaElena:
> It'll also cause the GitHub action to fail whenever a hub fails to check - we can use this to our advantage by specifying the staging & dask-staging hubs first in hubs.yaml. What do you think?

That is a great idea! (I believe the staging hubs are already the first in list <3). 
The reason I created a file with only the staging hubs was to be able to deploy the changes from a pull request on these staging hubs and then run the tests.

>We run the py.test based tests against every hub as it's deployed, not just on staging hubs.

This should happen on master pushes and not on pull requests, right?

What do you think about having two workflows:
- one for master pushes, where we deploy to and test **all the hubs**.
- one for pull requests, where we only deploy to and test only the **staging hubs**.

I believe it would be nice to be able to test pull request changes automatically. What do you think?

yuvipanda:
@GeorgianaElena ooh, that's interesting. I haven't thought of deploying per PR. 

If we do a deploy per PR, how would we be deal with two PRs opened around the same time? If they both try to deploy to staging at the same time, that'll probably cause issues. How do you think we can deal with that?

Excited to see where this goes.

GeorgianaElena:
I believe we could manually re-trigger the workflow if we see it failing because of this reason... Though I'm not sure if this has the potential to break the staging hubs and might not be sustainable.

### Some thoughts: 
- The main reason in my mind for this per-PR workflow would be to make sure that we don't break existing functionality, before merging the changes (by running the py.test tests on pull-requests).
- The other possible benefit would be that we could also check the new functionality added by a PR, right after it was opened or a new commit was added. But, when multiple PRs are opened, the state of the staging hubs will reflect the changes of the most recent commit of a PR, and would probably need a manual re-deploy (either locally or with the help of GHA).

Do you think this is worth it?

yuvipanda:
>     * But, when multiple PRs are opened, the state of the staging hubs will reflect the changes of the most recent commit of a PR, and would probably need a manual re-deploy (either locally or with the help of GHA).

A way around this probably is to generate an entirely new hub for each PR. However, I think it's not worth it on our case yet - we aren't making very many radical changes at this point. Also, I'd like us to eventually move away from GitHub actions into a fully self-serve UI, and so I think maybe this isn't worth it until then?

My suggestion is that we focus now only on the workflow after it's merged to master. The goal is to make sure we don't break any hubs *without us knowing* - and a failed CI deploy would definitely tell us. We can then add PR based checks for things that trip us up during deploy, as we gain experience with that. 

I'm very excited for us to get good per-PR testing! I don't think we need it right now, though.

dev-requirements.txt
```diff
@@ -0,0 +1,4 @@
+pytest
+pytest-asyncio
```
yuvipanda, line 2: Now that the code isn't async, do we still need this?

hub.py
```diff
@@ -388,11 +389,18 @@ async def deploy(self, auth_provider, proxy_secret_key, test_notebook_path=None)
...
-                print("Starting hub health validation...")
-                await self.check_hub_health(test_notebook_path, service_api_token)
+
+                exit_code = pytest.main([
+                    "-s", "-v", "tests",
+                    "--hub-name", self.spec["name"],
+                    "--cluster-name", self.cluster.spec["name"],
+                    "--api-token", service_api_token
+                ])
+                if exit_code:
```
yuvipanda, line 405:
```suggestion
                if exit_code != 0:
```

hub.py
```diff
@@ -388,11 +389,18 @@ async def deploy(self, auth_provider, proxy_secret_key, test_notebook_path=None)
...
+                    raise(RuntimeError)
```
yuvipanda, line 406: Does pytest print output to stdout? That way we can see which tests failed and debug.

tests/conftest.py
```diff
@@ -0,0 +1,39 @@
...
+        default="2i2c"
+    )
+    parser.addoption(
+        "--api-token",
+        action="store",
+        default="secret"
+    )
+
+@pytest.fixture()
+def hub(request):
```
yuvipanda, line 24: Can you add a docstring about how this fixture is coming from the commandline params? This is a very awesome way to use fixtures!

yuvipanda (commented):
Mino

hub.py
GeorgianaElena: Yes it does. It outputs everything we print inside the tests and the exception the test fails with.

Sample output for a passed test:
test-passed

Sample output for a failed test:
failed-test

tests/conftest.py
GeorgianaElena: Sure <3

PR #161, opened 2021-01-09
Document how to use a custom image

### LINKED ISSUE #160: Document how to use a custom environment / Docker image
Many organizations (and particularly the upcoming USC hub) want to be able to bring their own environments or Docker images to their user sessions. I'm not sure how to do this in the current setup, can this be documented? I think it'll be a common request. We should also document how the user image interacts with things like which UIs etc are there. A few specific questions to answer:

- Process / config for providing a user image (or some other way for letting them control the environment)
- If users provide their own image, is *nothing else* installed on top of it?
- Are there any restrictions or requirements for images that users provide?
- Where should user images be placed (e.g. Dockerhub/GCR/etc?)
- Can users use repo2docker to build images for their repositories?

yuvipanda:
Fixes #160

### DIFF
--- a/docs/custom-image.md
+++ b/docs/custom-image.md
@@ -1,0 +1,66 @@
+# Using a custom image
+
+Pilot hubs use an image we curate as the default. This can be replaced with your
+own custom image fairly easily. However, custom images should be maintained
+by the hub admins, and we won't be able to help much with things in it.
+
+## Image requirements
+
+On top of whatever you are using, your image should match the following requirements:
+
+1. The `jupyterhub` package is installed in such a way that the command `jupyterhub-singleuser`
+   works when executed with the image. 99% of the time, this just means you need to
+   install the `jupyterhub` python package.
+
+2. Everything should be able to run as a non-root user, most likely with a uid 1000. 
+
+There are a few options to help make this easier.
+
+1. [repo2docker](https://repo2docker.readthedocs.io) to build your image. 
+2. Use a [pangeo curated docker image](https://github.com/pangeo-data/pangeo-docker-images/).
+   They have an 'onbuild' variant of their images that lets you easily customize
+   them.
+3. Use a [jupyter curated docker image](https://jupyter-docker-stacks.readthedocs.io/en/latest/).
+
+Nothing will be installed on top of this by us, so you really have full control of what
+goes in your environment.
+
+## Which image registry to use?
+
+While you can push your image to [dockerhub](https://hub.docker.com), it now has
+pretty [strict usage limits](https://www.docker.com/increase-rate-limits). This
+could cause disruptions in your hub if a new image can not be pulled due to
+these rate limits. We recommend the following image repositories:
+
+1. [quay.io](https://quay.io/). Owned by RedHat / IBM. Easiest to get started in,
+   and **recommended as the default**
+2. [Google Artifact Registry](https://cloud.google.com/artifact-registry), if
+   you already have infrastructure running on Google Cloud.
+3. [AWS Public Container Registry](https://aws.amazon.com/blogs/aws/amazon-ecr-public-a-new-public-container-registry/)
+   if you already have infrastructure running on AWS.
+4. [GitHub container registry](https://docs.github.com/en/free-pro-team@latest/packages/guides/about-github-container-registry).
+   It integrates better with GitHub, but doesn't have a clear policy on rate limits
+   yet.
+
+## Configuring your hub to use the custom image
+
+`hubs.yaml` can contain arbitry `zero to jupyterhub on kubernetes` config, so we
+can use that to set the image name & tag. Here is some example config, with only
+the useful bits.
+
+```yaml
+- name: custom-image-hub
+  ...
+  config:
+    base-hub:
+      jupyterhub:
+        singleuser:
+          image:
+            name: pangeo/pangeo-notebook
+            tag: 2020.12.08
+```
+
+This can be any image name & tag. 
+
+Whenever you push a new image, you'll have to make a PR that updates the tag here.
+Only on merge will the hub get the new image.
--- a/docs/custom-image.md
+++ b/docs/custom-image.md
@@ -16,7 +16,7 @@ On top of whatever you are using, your image should match the following requirem
 
 There are a few options to help make this easier.
 
-1. [repo2docker](https://repo2docker.readthedocs.io) to build your image. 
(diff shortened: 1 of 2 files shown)


yuvipanda:
How does this look, @choldgraf?

docs/custom-image.md
```diff
@@ -0,0 +1,66 @@
+# Using a custom image
+
+Pilot hubs use an image we curate as the default. This can be replaced with your
+own custom image fairly easily. However, custom images should be maintained
+by the hub admins, and we won't be able to help much with things in it.
```
choldgraf, line 5: can we define who the "hub admin" is? Our official roles that we define are "hub engineer", "community champion", and "hub architect". We should make it clear which of those roles is "the hub admin" since I suspect that the hub engineer will also have admin rights on the hubs

docs/custom-image.md
```diff
@@ -0,0 +1,66 @@
...
+
+1. The `jupyterhub` package is installed in such a way that the command `jupyterhub-singleuser`
+   works when executed with the image. 99% of the time, this just means you need to
+   install the `jupyterhub` python package.
+
+2. Everything should be able to run as a non-root user, most likely with a uid 1000. 
+
+There are a few options to help make this easier.
+
+1. [repo2docker](https://repo2docker.readthedocs.io) to build your image.
```
choldgraf, line 19:
```suggestion
1. Use [repo2docker](https://repo2docker.readthedocs.io) to build your image. 
```

docs/custom-image.md
```diff
@@ -0,0 +1,66 @@
...
+goes in your environment.
+
+## Which image registry to use?
+
+While you can push your image to [dockerhub](https://hub.docker.com), it now has
+pretty [strict usage limits](https://www.docker.com/increase-rate-limits). This
+could cause disruptions in your hub if a new image can not be pulled due to
+these rate limits. We recommend the following image repositories:
+
+1. [quay.io](https://quay.io/). Owned by RedHat / IBM. Easiest to get started in,
```
choldgraf, line 35: that's a new one to me :-) interesting!

choldgraf (commented):
This looks great to me - just a few minor comments in there.

One extra thought - do we assume that this is the pattern that *most* people will follow when adding their own packages and such? Or in the future are we planning to make it possible for them to "start with a base 2i2c image and then add a few packages on top of it"? I foresee the latter being a much more common use-case than building the whole image from scratch.

choldgraf:
One other thought - I think we should define the *process* by which these images generally get onto the hub. E.g., do we tell the hub admins that they should open a PR to this repository that points the user image to their hub image? Do they tell the hub engineer about the location of the image and then the engineer opens the PR?

cc also @zaneselvans who I think would be interested in this workflow and might have comments himself!

choldgraf:
I'm just gonna merge this in because it's a clear improvement over what we've already got. Let's iterate from there!

zaneselvans:
Sorry I've been a bit absent. We finally got a draft of the grant sent off and I have space for other things now. So if I understand correctly, we can create our own container image based on one of the ones you've listed, like something appropriate from the stack of curated Jupyter images, add our own layer to the image with our required packages installed, and then push it to some container registry, and then specify in the `hubs.yaml` file what registry, container, and tag should be pulled, and that's the image that will show up on our hub.  And if we screw something up in the image then we need to go fix it ourselves 😄 

How would one best go about testing potential new images before deploying them to the hub for everyone to use? Is it usually enough to just try and run a notebook server locally using the image? Or is there some way to deploy it only for some users and not others? Or have a testing hub setup that runs alongside the production one?

zaneselvans:
And the repositories that control all these configurations are administered by 2i2c, with users/clients making PRs to implement changes?

choldgraf:
No worries @zaneselvans - 

> we can create our own container image based on one of the ones you've listed

You don't need to base your image based on those images, though doing so might make your life a bit easier (particularly, using repo2docker to generate the image). You just need to make sure the constraints listed in https://2i2c.org/pilot-hubs/custom-image.html#image-requirements are met

> if we screw something up in the image then we need to go fix it ourselves

We can provider pointers about what kinda stuff needs to be in the image, but yeah we probably don't have time to curate everybody's custom Docker image. We're thinking of this like an "advanced use case". In the future we probably want easier ways for people to extend the environment.

> How would one best go about testing potential new images before deploying them to the hub for everyone to use?

That is a good question - normally what we do is set up a staging hub alongside the production hub. That said, this would definitely be added complexity for the infrastructure and it's not set up to do this yet. In the short term, I think the method you describe is probably reasonable, though you wouldn't be able to test stuff like scaling with kubernetes etc. Maybe start off trying that and then see how it goes from there?

@yuvipanda what are your thoughts on recommendations for how users can "test out" their images before deploying them?

zaneselvans:
`repo2docker` definitely sounds like a good way to go for something like our (hopefully growing) collection of tutorial notebooks. It does seem like it might be a little funny to have a repo that mostly exists to specify a Docker container. Is that common? Our plan right now is to try and use exactly the same Docker container for the JupyterHub as we use for local day-to-day work, and to archive it alongside the data releases for use by future users that want to download a completely self-contained environment. But maybe setting up an `environment.yml` in our https://github.com/catalyst-cooperative/pudl-tutorials repo would be a good way to start defining the environment.

I imagine it'll also include the Intake data catalog conda packages once we create them.  I kind of want to experiment with that soon.  I've been wondering what costs would end up being associated with just having Hub users use Intake to pull data for local use -- like would there be egress fees to get it to the Hub from our cloud storage? Would every hub user necessarily end up with their own duplicated local file cache?

choldgraf:
> It does seem like it might be a little funny to have a repo that mostly exists to specify a Docker container. Is that common?

Well the reason to use repo2docker is because it is what Binder uses, so basically the repository you'd have would be ~ a Binder repository, for which yes there are many out there just to specify an environment 👍 

The main benefit of this is just that environment.yml, requirements.txt, etc tend to be much more human-readable than Dockerfiles, so easier to maintain for folks who aren't Docker experts

> like would there be egress fees to get it to the Hub from our cloud storage? Would every hub user necessarily end up with their own duplicated local file cache?

I bet it depends a lot on where hte data is stored relative to where the hub runs. Usually data transfer fees are much, much smaller if the data doesn't need to leave the cloud. If the data was (relatively) small, then it shouldn't be a big deal to effectively have copies on user drives, but definitely something we cna experiment with moving forward

PR #169, opened 2021-01-14
Use lowercase for hub name to fix helm release name

GeorgianaElena:
Can a helm release name start with an upper case?

Deploy is failing for the newly added Lassen hub with:

```
Error: release name is invalid: Lassen
```


yuvipanda:
Ah, that makes sense! I think they can't have uppercase letters - https://kubernetes.io/docs/concepts/overview/working-with-objects/names/ has the full info.

PR #177, opened 2021-01-20
Docs service

GeorgianaElena:
Ref https://github.com/2i2c-org/pilot-hubs/issues/58

### DIFF
--- a/hub-templates/base-hub/templates/docs-service-config.yaml
+++ b/hub-templates/base-hub/templates/docs-service-config.yaml
@@ -1,0 +1,14 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: nginx-docs-configmap
+  labels:
+    app: docs-service
+data:
+  nginx.conf: |
+    server {
+      listen 12222;
+      location / {
+          root /etc/nginx/html;
+      }
+    }
--- a/hub-templates/base-hub/templates/docs-service-config.yaml
+++ b/hub-templates/base-hub/templates/docs-service-config.yaml
@@ -1,3 +1,4 @@
+{{ if .Values.docsService.enabled }}
 kind: ConfigMap
 apiVersion: v1
 metadata:
@@ -12,3 +13,4 @@ data:
           root /etc/nginx/html;
       }
     }
+{{ end }}
--- a/hub-templates/base-hub/templates/docs-service-config.yaml
+++ b/hub-templates/base-hub/templates/docs-service-config.yaml
@@ -7,7 +7,7 @@ metadata:
 data:
   nginx.conf: |
     server {
-      listen 12222;
+      listen 8080;
       location / {
           root /etc/nginx/html;
       }
--- a/hub-templates/base-hub/templates/docs-service-config.yaml
+++ b/hub-templates/base-hub/templates/docs-service-config.yaml
@@ -1,4 +1,4 @@
-{{ if .Values.docsService.enabled }}
+{{ if .Values.jupyterhub.docs_service.enabled }}
 kind: ConfigMap
 apiVersion: v1
 metadata:
@@ -10,7 +10,8 @@ data:
     server {
       listen 8080;
       location / {
-          root /etc/nginx/html;
+        index index.html;
+        root /etc/nginx/html;
       }
     }
 {{ end }}
--- a/hub-templates/base-hub/templates/docs-service-deployment.yaml
+++ b/hub-templates/base-hub/templates/docs-service-deployment.yaml
@@ -1,4 +1,4 @@
-{{ if .Values.docsService.enabled }}
+{{ if .Values.jupyterhub.docs_service.enabled }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -26,10 +26,10 @@ spec:
             - clone
             - --depth=1
             - -b
-            - {{ .Values.jupyterhub.docs.branch }}
+            - {{ .Values.jupyterhub.docs_service.branch }}
             - --single-branch
             - --
-            - {{ .Values.jupyterhub.docs.repo }}
+            - {{ .Values.jupyterhub.docs_service.repo }}
             - /srv/docs
           securityContext:
               runAsUser: 1000
--- a/hub-templates/base-hub/templates/docs-service-deployment.yaml
+++ b/hub-templates/base-hub/templates/docs-service-deployment.yaml
@@ -58,7 +58,8 @@ spec:
           image: nginx
           command: ["/usr/sbin/nginx", "-g", "daemon off;"]
           ports:
-            - containerPort: 12222
+            - name: nginx-port
+              containerPort: 8080
           volumeMounts:
           - name: nginx-config
             mountPath: /etc/nginx/conf.d/default.conf
--- a/hub-templates/base-hub/templates/docs-service-deployment.yaml
+++ b/hub-templates/base-hub/templates/docs-service-deployment.yaml
@@ -21,12 +21,14 @@ spec:
       initContainers:
         - name: docs-clone
           image: alpine/git
-          args: 
+          args:
             - clone
             - --depth=1
+            - -b
+            - gh-pages
             - --single-branch
             - --
-            - https://github.com/cloudacademy/static-website-example.git
+            - {{ .Values.jupyterhub.docs.repo }}
             - /srv/docs
           securityContext:
               runAsUser: 1000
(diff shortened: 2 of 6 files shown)


choldgraf:
@GeorgianaElena do you speak Kubernetes now? This all looks like gibberish to me 😅

GeorgianaElena:
Haha :sweat_smile: I wouldn't go that far, it's more like a baby saying it's first "mama" word :baby_chick:

hub-templates/base-hub/values.yaml
```diff
@@ -193,3 +193,18 @@ jupyterhub:
             return super().start(*args, **kwargs)
 
         c.JupyterHub.spawner_class = CustomSpawner
+      06-docs-nginx-service: |
+          # Adds JupyterHub service to make the docs available at
+          # {HUB_URL}/services/docs
+          import os
+          host = os.environ["DOCS_SERVICE_SERVICE_HOST"]
```
yuvipanda, line 200: When there's a service named `docs-service`, kubernetes adds a DNS entry for `docs-service`. So you can just access `docs-service` from the hub, and that works well! No need to use environment variables.

hub-templates/base-hub/templates/docs-service-deployment.yaml
```diff
@@ -0,0 +1,67 @@
...
+              \ 5m; done"
+          securityContext:
+            runAsUser: 1000
+            allowPrivilegeEscalation: False
+            readOnlyRootFilesystem: True
+          volumeMounts:
+            - name: docs
+              mountPath: /srv/docs
+        - name: nginx-docs-service
+          image: nginx
```
yuvipanda, line 58: We should always specify a tag for all our images. This makes sure that our deployments are repeatable and reproducible. Without a tag, kubernetes tries to use the latest image. The 'latest image' can change over time without us knowing anything about it, and can break unexpectedly. For example, if the pod moves from one node to another for any reason, the new image pulled could be totally different from the old image! This could break our service when we aren't looking. So let's always find a tag and use it.

hub-templates/base-hub/values.yaml
yuvipanda: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/ has some more info

hub-templates/base-hub/templates/docs-service-config.yaml
```diff
@@ -0,0 +1,14 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: nginx-docs-configmap
+  labels:
+    app: docs-service
+data:
+  nginx.conf: |
+    server {
+      listen 12222;
```
yuvipanda, line 10: I generally like to listen on port 8080, as the most popular [alternative http port](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=http-alt).

yuvipanda (changes requested):
This is awesome! I hope you had fun with it :) 

Let's make the whole setup conditional, based on a flag in the values file - like https://github.com/2i2c-org/pilot-hubs/blob/05dc69a7ab722b417122138f155634b597ff8342/hub-templates/base-hub/templates/nfs-pvc.yaml#L1. Not all hubs will have something like this going.

hub-templates/base-hub/templates/nginx-docs-service.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: docs-service
+  labels:
+    app: docs-service
+spec:
+  type: ClusterIP
+  ports:
+    - port: 12222
```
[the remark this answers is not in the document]
consideRatio: I suggest to name the port on the pod's container and reference it by name also.

GeorgianaElena:
I did my best to address your feedback! Thank you @yuvipanda and @consideRatio for reviewing it :heart:

(duplicate of https://github.com/2i2c-org/pilot-hubs/pull/185#issuecomment-767666576 :sweat_smile: )

hub-templates/base-hub/templates/docs-service-config.yaml
```diff
@@ -0,0 +1,14 @@
+kind: ConfigMap
```
yuvipanda, line 1: Let's put this behind a conditional too, just like the deployment.

hub-templates/base-hub/values.yaml
```diff
@@ -20,6 +20,9 @@ nfsPVC:
     serverIP: nfs-server-01
     shareName: /export/home-01
 
+docsService:
+  enabled: true
```
yuvipanda, line 24: Let's disable this by default, and enable it only for the hubs that use it.

hub-templates/base-hub/templates/docs-service-deployment.yaml
```diff
@@ -0,0 +1,70 @@
...
+        - name: docs
+          emptyDir: {}
+      initContainers:
+        - name: docs-clone
+          image: alpine/git
+          args:
+            - clone
+            - --depth=1
+            - -b
+            - gh-pages
```
yuvipanda, line 29: The branch should be configurable too, since many repos just use master instead of gh-pages.

hub.py
```diff
@@ -338,6 +338,15 @@ def apply_hub_template_fixes(self, generated_config, proxy_secret_key):
 
         }
 
+        # Generate a token for the docs service
+        docs_token = hmac.new(proxy_secret_key, f'docs-{self.spec["name"]}'.encode(), hashlib.sha256).hexdigest()
+        # Register the docs service
+        generated_config['jupyterhub']['hub']['services']['docs'] = {
+            'apiToken': docs_token,
```
yuvipanda, line 345: Does the docs service need an apiToken? It isn't making any API calls to the hub, so maybe not necessary. I'm not sure if you can run a service without explicitly setting an apitoken though.

yuvipanda (changes requested):
A few more fixes, and then feel free to merge, @GeorgianaElena!

hub.py
GeorgianaElena: It turns out, it didn't need an api token :D Thanks for suggesting to try this <3!

GeorgianaElena:
This is how the [gh-pages branch of the 2i2c website](https://github.com/2i2c-org/2i2c-org.github.io/tree/gh-pages) are rendered by the docs service in the staging hub:

docs-service-gh-pages

Some things I noticed:
* When I used some random website template I found on GitHub, it got rendered properly
* When I open `index.html` in a browser, it looks the same it does in the gif above
* Unfortunately I didn't capture the mouser cursor in the gif, but the 404 happens for any link I click on the index page.
The url suggests that instead of opening for example the `about` page at `/services/docs-service/about`, it instead tries to go to `/hub/about`
(it doesn't work locally either).

I think it has something to do with the repo itself and how the resources are structured, but I don't know enough about this :(

yuvipanda:
This has probably something to do with relative vs absolute paths. @choldgraf do you know how we can make the generated docs use relative paths so they work at `/service/docs` as well, and not just at `/`?

choldgraf:
ah - so this will depend on the SSG that is being used. I guess Hugo (which we're using for the website) doesn't do this by default? @GeorgianaElena can you try the same with a Sphinx-built website? I think that Sphinx *does* do relative paths by default so this shouldn't be an issue.

consideRatio:
Out of the blue - a question: can JupyterHub be configured to register a route with the configurable proxy it relies on other than `/services/<service-name>`, for example `/docs` in this case?

choldgraf:
Wooo SSGs! The future is static!

yuvipanda:
If the SSG has to know where the docs are going to be served from to use as a prefix, that means we can't use the same site on both github pages and here :'( That makes me at least a little sad.

Ideally, all links would be *relative* - so that would allow them to work on any prefix. Perhaps https://gohugo.io/content-management/urls/#relative-urls should help with that?

GeorgianaElena:
@choldgraf , I tried with your website :smile: and it worked:

sphinx-built-website

I've opened https://github.com/2i2c-org/2i2c-org.github.io/pull/35 to try Yuvi's suggestion <3

GeorgianaElena:
I have now added the pilot-hubs docs to be served by the [staging hub](https://staging.pilot.2i2c.cloud) docs service, which I think makes more sense than having the 2i2c or Chris' website :smile: They too render and redirect ok (they are built with sphinx).

What do you all think about merging this as it is and then open another PR for documenting the feature? We can then explore more solutions and write some requirements the docs repo has to follow to work with the service.

yuvipanda:
@GeorgianaElena This is awesome! I've merged this now, awesome work on all the k8s and nginx stuff!!!!!!!

PR #181, opened 2021-01-22
Custom image for Catalyst Cooperative pilot hub

zaneselvans:
I've created an image using repo2docker based on [our tutorial repo](https://github.com/catalyst-cooperative/pudl-tutorials) and pushed it to docker hub [catalystcoop/pilot-hub:2021.01.22](https://hub.docker.com/r/catalystcoop/pilot-hub/tags?page=1&ordering=last_updated)  I think this PR would specify the use of that image for the catalyst hub's user environment. Right?

It seems to work with PUDL when I run it locally using `repo2docker` though I have not been able to run the image directly using `docker run` -- I haven't figured out how to get it to publish the right ports I don't think... it always comes up on `127.0.0.1:8888` but when I go to that address it won't load. Are there other things I can do to test it out before the hub switches over? It would be nice if I could set up local `shared` and `shared-readwrite` volumes to mount so I can test the tutorial notebooks locally as well. Not yet clear on how to mount external files into the image though.

Is it better to use a versioned tag like this (e.g. the date) and then have to make a PR every time we want to change the image? Or should I use something like `latest` so that the environment updates whenever I push a new version?

### DIFF
--- a/hubs.yaml
+++ b/hubs.yaml
@@ -327,6 +327,9 @@ clusters:
           base-hub:
             jupyterhub:
               singleuser:
+                image:
+                  name: catalystcoop/pilot-hub
+                  tag: 2021.01.22
                 memory:
                   limit: 6G
                   guarantee: 4G


yuvipanda:
This is great, @zaneselvans! I've merged this, and catalyst-cooperative.pilot.2i2c.cloud/ now uses the new image! Can you try it out?

Tags with dates / hashes and then explicit updates here is the best way to go right now. Hopefully that can change within the next few weeks.

re: repo2docker and mounts, can you open another issue? I'm off to sleep now, but will try and respond to it later :) 

Thanks!

zaneselvans:
I was able to restart my server, log out and back in, and run one of the tutorial notebooks without needing to install any additional software -- doing `conda list` in a terminal through Jupyter showed that `catalystcoop.pudl` and other specified packages were installed, so it seems to be working!

choldgraf:
yahoo!

PR #182, opened 2021-01-25
[WIP] Restrict access to a certain domain automatically

GeorgianaElena:
This should be smarter and instead of creating a new rule rule for each hub that has restriction based on domain in place, we should update an existing rule with the name of the hub.

Having lots of rules is bad also because the rules evaluate one after another in the order specified by their `order` field.

### DIFF
--- a/auth.py
+++ b/auth.py
@@ -91,21 +91,38 @@ class KeyProvider:
         
         return client
 
-    def create_google_domain_rule(self):
+    def create_google_domain_rule(self, hub_name, domain_list, admin_list):
         gt = GetToken(self.domain)
         creds = gt.client_credentials(
             self.client_id, self.client_secret,
             f'https://{self.domain}/api/v2/'
         )
         rules = Rules(self.domain, creds['access_token'])
-        request_body = {
-            "name": "testRule",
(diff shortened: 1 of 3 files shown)


yuvipanda (changes requested):
auth0 rules seem very exciting!!!

My hope is that we can do this idempotently. We have an `ensure_client` method that makes sure that a client is created with the given property if it exists, and *changes* the existing client to match given properties if it doesn't exist. We should probably also implement something like that here - if the list of admins or users change, this rule would need to be recreated. 

Encoding username and admin lists directly in the code is also problematic, since it can lead to sql injection style attacks. For example, if 'domainAllowedList' is set to `""; alert("hi");` that will be counted as valid JS and execute the arbitrary `alert`. This can be exploited to steal private data or allow unauthorized use. So in general, it's always a good idea to separate data from code. Ideally, this would be doable by adding some metadata style information to each client. If not, we should JSON encode the data, and put it in as a global variable, rather than embedding it in the code in specific places.

yuvipanda:
ah, I also just discovered that all this is moot - our auth0 pricing plan [only allows a total of 5 rules](https://auth0.com/pricing/), so I think we shouldn't use that for this :'( Can you investigate `username_pattern` instead? Sorry!

GeorgianaElena:
Since we're not allowed to more than 5 rules, I'm going to close this and use [`username_pattern`](https://jupyterhub.readthedocs.io/en/stable/api/auth.html#jupyterhub.auth.Authenticator.username_pattern) instead.

PR #185, opened 2021-01-26
Add datahub.mills.edu as an ingress domain for mills hub

GeorgianaElena:
Ref: https://github.com/2i2c-org/pilot-hubs/issues/178

### DIFF
--- a/hub.py
+++ b/hub.py
@@ -313,9 +313,9 @@ class Hub:
         """
 
         if self.spec['auth0']['domain']:
-                hub_domain = self.spec['auth0']['domain']
-            elif isinstance(self.spec['domain'], str):
-                hub_domain = self.spec['domain']
+            hub_domain = self.spec['auth0']['domain']
+        elif isinstance(self.spec['domain'], str):
+            hub_domain = self.spec['domain']
 
         hub_url = f'https://{hub_domain}'
         username='deployment-service-check'
--- a/hub.py
+++ b/hub.py
@@ -320,7 +320,7 @@ class Hub:
         hub_url = f'https://{hub_domain}'
         username='deployment-service-check'
 
-        # Export the hub health check service as an env var so that jupyterhub_client can read it.
+        # Export the hub health check service as an env var so that jhub_client can read it.
         orig_service_token = os.environ.get('JUPYTERHUB_API_TOKEN', None)
 
         try:
@@ -368,8 +368,6 @@ class Hub:
         them in this function.
         """
         hub_template = self.spec['template']
-        hub_name = self.spec['name']
-
 
         # Generate a token for the hub health service
         hub_health_token = hmac.new(proxy_secret_key, 'health-'.encode() + self.spec['name'].encode(), hashlib.sha256).hexdigest()
--- a/hub.py
+++ b/hub.py
@@ -139,11 +139,11 @@ class Hub:
             'jupyterhub': {
                 'proxy': { 'secretToken': proxy_secret },
                 'ingress': {
-                    'hosts': self.spec['domain'],
+                    'hosts': self.spec['domain'] if isinstance(self.spec['domain'], list) else [self.spec['domain']],
                     'tls': [
                         {
                             'secretName': f'https-auto-tls',
-                            'hosts': self.spec['domain']
+                            'hosts': self.spec['domain'] if isinstance(self.spec['domain'], list) else [self.spec['domain']]
                         }
                     ]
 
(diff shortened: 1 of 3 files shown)


hub.py
```diff
@@ -328,6 +328,12 @@ def apply_hub_template_fixes(self, generated_config, proxy_secret_key):
         them in this function.
         """
         hub_template = self.spec['template']
+        hub_name = self.spec['name']
+
+        if hub_name == "mills":
```
yuvipanda, line 333: In general we should try to avoid special casing any particular hub in our deployment script. I'd like for us to remove the special casing for hub-templates too, but some other time.

One suggestion here is to allow `domain` to be a list, rather than just a string as it is now. This should cause minimal issues, no hard coding, and pass through all the things we want.

yuvipanda (changes requested):

hub.py
GeorgianaElena: Making `domain` a list, is a great idea <3 Thank you!

I have one curiosity though, if we were to add the `datahub.mills.edu` ingress part into `hubs.yaml` would that have merged ok with the one in `generated_config`? (I understand that it's better to have similar config in the same place, I'm just wondering how smart the config merging process is)

yuvipanda: I think right now lists don't get merged, since there isn't unfortunately an easy way to merge them :( If we can figure that out, that'd be great! Not sure of an obvious solution though.

hub.py
```diff
@@ -234,9 +234,10 @@ def get_generated_config(self, auth_provider, proxy_secret_key):
         #
         # Allow explicilty ignoring auth0 setup
         if self.spec['auth0'].get('enabled', True):
+            auth0_org_domain = next(domain for domain in self.spec['domain'] if "2i2c" in domain)
             client = auth_provider.ensure_client(
                 self.spec['name'],
-                self.spec['domain'],
+                auth0_org_domain,
```
yuvipanda, line 240: I had forgotten that auth0 requires domain name the way we have set it up. I don't want to assume that there'll *always* be a 2i2c domain, and so this might be better explicit.

In this case, how about we:

1. If there's only one domain, it's listed as a string in `domain`.
2. If there's more than one domain, `domain` can be a list - but then you *must* specify `domain` under `auth0` as well to specify what domain should be used with auth0.

`auth0.domain` takes precedence over `domain` if it exists.

This way, we have minimal config for the general case, but *explicit* config for the more complex case.

hubs.yaml
```diff
@@ -13,7 +13,8 @@ clusters:
     hubs:
       - name: staging
         cluster: 2i2c
-        domain: staging.pilot.2i2c.cloud
+        domain:
```
yuvipanda, line 16: This should be optionally a list or string, since most hubs will not have more than one domain

tests/test_custom_templates.py
```diff
@@ -9,7 +9,14 @@ def test_templates_loaded(hub):
 
     It checks that each hub's login page has their configured institutional logo.
     """
-    url = f'https://{hub.spec["domain"]}/hub/login'
+
+    hub_org_domain = (
```
yuvipanda, line 13: We should check all hosts specified here

GeorgianaElena:
I did my best to address your feedback! Thank you @yuvipanda and @consideRatio for reviewing it :heart:

GeorgianaElena:
Ooops :blush: Posted this from the phone in the wrong PR :sweat_smile:

hub.py
```diff
@@ -299,7 +307,13 @@ async def check_hub_health(self, test_notebook_path, service_api_token):
         before declaring it a failure. If any of these steps fails, immediately halt further
         deployments and error out.
         """
-        hub_url = f'https://{self.spec["domain"]}'
+
+        hub_org_domain = (
+            self.spec["domain"]
+            if isinstance(self.spec["domain"], str)
```
yuvipanda, line 313: We should probably also use the same logic as earlier (auth0.domain overriding domain) here.

yuvipanda (commented):
This LGTM, @GeorgianaElena. Please merge after the CNAME is set up.

GeorgianaElena:
@yuvipanda, now that CNAME entry was added, should I try a manual deploy for this and see what happens?

yuvipanda:
@GeorgianaElena yep!

GeorgianaElena:
It partially worked :sweat_smile: I can see the hub login page when accessing the new domain but I can't login because of this error:
Screenshot from 2021-02-11 15-21-07

I believe something extra needs to happen on the auth0 side (I see an *Allowed Callback URLs* from which the new domain is missing there). Investigating now...

GeorgianaElena:
yep, that did the trick :heart:

yuvipanda:
\o/ awesome. when this is done, can you add some docs about how to add additional domains to a hub?

GeorgianaElena:
Sure. I've just added an issue about it and I'll probably start a documentation spree asap because I've left other things undocumented too (like the docs service).

Also, after the current deploy finishes, I'll merge this because it seems to be working properly :tada: thank you, @yuvipanda

PR #191, opened 2021-02-02
Keep the checks separated

GeorgianaElena:


consideRatio:
With modern version of helm there is a function that can help in these situations called dig: http://masterminds.github.io/sprig/dicts.html

No need to change any of this though!

GeorgianaElena:
wooooo, this is awesome :heart_eyes: thanks for letting me know about these @consideRatio

PR #192, opened 2021-02-03
Attempt to fix servers left behind after failed health check

GeorgianaElena:


hub.py
```diff
@@ -303,14 +303,20 @@ async def check_hub_health(self, test_notebook_path, service_api_token):
...
             await execute_notebook(
                 hub_url,
                 test_notebook_path,
-                temporary_user=True,
+                username='deployment-service-check',
+                server_creation_timeout=90,
+                kernel_execution_timeout=90, # This doesn't do anything yet
                 create_user=True,
-                delete_user=True
+                delete_user=True,
```
GeorgianaElena, line 317: * If we choose to delete the user, then if its server is left behind, we won't be able to find and delete it with something like:
```python
    hub = JupyterHubAPI(hub_url)
    async with hub:
        user = await hub.get_user(username)
        if(user and user['server'] and not user['pending']):
            await hub.delete_server(username)
```
     This will lead to other servers being spawned for this user :(

* If we don't delete the user, then we can delete the server left running, but we won't be able to anything else.
This is because we won't have the user's api token anymore (this is different than the service's token and we get our hands on it when we create this user the first time). We need this token to start the kernel and execute the code (ref: https://github.com/costrouc/jupyterhub_client/pull/4).

yuvipanda: Is there a JupyterHub API to create / delete user tokens?

GeorgianaElena: It is: https://jupyterhub.readthedocs.io/en/stable/_static/rest-api/index.html#path--users--name--tokens--token_id-
but we need the old token which we don't :(

I gave it a try, deleting the user too (as part of the cleanup) and it worked (I could successfully execute a notebook for a user and a server that were already present on the hub from a previous iteration).

This means:
* after a health check (successful or not) we leave the health check user behind.
* if the health check was successful, then their user server will successfully get deleted
* if the health check failed, then their user server will be left running, but it will get deleted together with the user in the next iteration.

@yuvipanda, do you think this is acceptable? :eyes:

yuvipanda: That sounds great! Let's do that for now :)

GeorgianaElena: Merge this, and see how many things I forgot? :sweat_smile:

yuvipanda:
Let's try!

PR #193, opened 2021-02-03
Update Catalyst Cooperative Docker image

zaneselvans:
Added `nbgitpuller` back into the environment (whoops!) as well as a `pip` dependency for `jupyter-offlinenotebook` since it was complaining about needing to be included in the build. Got automatic builds of the Docker image working from our examples repository using the `jupyter/repo2docker-action` GitHub action, with automatic date tagging.


choldgraf:
🚀

PR #194, opened 2021-02-04
Bump hub version to 0.11

yuvipanda:
Breaking changes to fix:

- All authenticator config has changed, and must be
  modified - https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1943
- pdb has been disabled - we should file an issue to investigate
  if we want it - https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1938
- Any networkpolicy changes we might need -
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/CHANGELOG.md#breaking-changes-1

### DIFF
--- a/hub-templates/base-hub/values.yaml
+++ b/hub-templates/base-hub/values.yaml
@@ -58,6 +58,18 @@ jupyterhub:
     chp:
       nodeSelector:
         hub.jupyter.org/pool-name: core-pool
+      networkPolicy:
+        enabled: true
+        ingress:
+        - from:
+          - namespaceSelector:
+              matchLabels:
+                name: support
+          - podSelector:
+              matchLabels:
+                hub.jupyter.org/network-access-proxy-http: "true"
+          ports:
+          - port: http
       resources:
         requests:
           # FIXME: We want no guarantees here!!!
@@ -78,18 +90,6 @@ jupyterhub:
       enabled: true
       letsencrypt:
         contactEmail: yuvipanda@gmail.com
-    networkPolicy:
-      enabled: true
-      ingress:
-      - from:
-        - namespaceSelector:
-            matchLabels:
-              name: support
-        - podSelector:
-            matchLabels:
-              hub.jupyter.org/network-access-proxy-http: "true"
-        ports:
-        - port: http
   singleuser:
     admin:
       extraVolumeMounts:
--- a/hub-templates/base-hub/values.yaml
+++ b/hub-templates/base-hub/values.yaml
@@ -75,6 +75,7 @@ jupyterhub:
         limits:
           memory: 1Gi
     https:
+      enabled: true
       letsencrypt:
         contactEmail: yuvipanda@gmail.com
     networkPolicy:
--- a/hub-templates/base-hub/values.yaml
+++ b/hub-templates/base-hub/values.yaml
@@ -131,9 +131,13 @@ jupyterhub:
         - ports:
             - port: 443
               protocol: TCP
-  auth:
-    type: dummy
   hub:
+    config:
+      JupyterHub:
+        authenticator_class: oauthenticator.generic.GenericOAuthenticator
+    image:
+      name: quay.io/2i2c/pilot-hub
+      tag: '0.0.1-n311.h51f02bb'
     nodeSelector:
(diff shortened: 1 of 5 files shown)


GeorgianaElena:
I have tested this on staging and everything appears to be working :tada:

hub-templates/base-hub/values.yaml
```diff
@@ -75,20 +87,9 @@ jupyterhub:
         limits:
           memory: 1Gi
     https:
+      enabled: true
```
yuvipanda, line 90: Would this deploy the autohttps pod? We don't need that, right? I think the letsencrypt section here can also be removed, since we use nginx-ingress + cert-manager for it.

hub-templates/base-hub/values.yaml
```diff
@@ -131,9 +132,13 @@ jupyterhub:
...
             - port: 443
               protocol: TCP
-  auth:
-    type: dummy
   hub:
+    config:
+      JupyterHub:
+        authenticator_class: oauthenticator.generic.GenericOAuthenticator
+    image:
+      name: quay.io/2i2c/pilot-hub
```
yuvipanda, line 140: I was experimenting with this hub image when I pushed it. I think we can remove this and use the default image for now.

hubs.yaml
```diff
@@ -245,6 +219,10 @@ clusters:
           connection: github
         config:
           jupyterhub:
+            singleuser:
+              image:
+                name:  quay.io/2i2c/paleohack-2021
```
yuvipanda, line 224: Let's remove this too for now - I think this was me experimenting locally :)

consideRatio:
> pdb has been disabled - we should file an issue to investigate
if we want it - jupyterhub/zero-to-jupyterhub-k8s#1938

I'm strongly opinionated after having researched and considered this in depth.

I find there is no point enabling a PDB to enforce either 1 replica availability or 0 replica unavailability on pods not supporting having two separate replicas on - this was the old z2jh default.

Note also that mybinder.org-deploy has explicitly opted out of the old z2jh default, and systematically avoids having behavior like the old z2jh default for all other deployments with 1 replica. This was discussed in https://github.com/jupyterhub/mybinder.org-deploy/pull/1730#issuecomment-751519417 also.

hubs.yaml
```diff
@@ -38,21 +38,15 @@ clusters:
...
-                  - georgiana.dolocan@gmail.com
+            hub:
+              config:
+                Authenticator:
+                  allowed_users: &staging_users
+                    - yuvipanda@gmail.com
+                    - colliand@gmail.com
+                    - choldgraf@gmail.com
+                    - georgiana.dolocan@gmail.com
+                  admin_users: *staging_users
```
consideRatio, line 49: Users added to admin_users doesn't need to also be added to allowed_users, right? No damage done doing this though.

hub-templates/base-hub/values.yaml
consideRatio: I think it would yes because letsencrypt.contactEmail is also specified and the default https type is letsencrypt.

With nginx-ingress + cert-manager, set this to false I think.

hub-templates/base-hub/values.yaml
```diff
@@ -58,6 +58,18 @@ jupyterhub:
...
+        ingress:
+        - from:
+          - namespaceSelector:
+              matchLabels:
+                name: support
+          - podSelector:
+              matchLabels:
+                hub.jupyter.org/network-access-proxy-http: "true"
+          ports:
+          - port: http
```
consideRatio, line 72: I believe this section won't be needed, because http access to the autohttps and chp pods are granted by default.

yuvipanda:
@consideRatio thanks for the pointers re: pdb. Since this PR blocks #185, I just wanted to try and keep the old behavior as much as possible :)

I've opened https://github.com/2i2c-org/pilot-hubs/issues/202 with your comment, and would love to know how you think we can accomplish the behavior we want.

GeorgianaElena:
Thank you @consideRatio and @yuvipanda for feedback <3

GeorgianaElena:
Since this seems to be working on staging, I'm going to merge it now and see what happens :crossed_fingers:

PR #204, opened 2021-02-08
Upgrade to JupyterLab 3.0

GeorgianaElena:
Closes https://github.com/2i2c-org/pilot-hubs/issues/197

- **Update2:** The `jupyter-server-proxy` version has been bumped. This only needs building the new user image and some manual testing.
    - [x] Build the new image and test this
- Update: This was tested and it works, just needs a new jupyter-server-proxy release :sun_with_face:


consideRatio:
@GeorgianaElena this is also needed I think.

```yaml
hub:
  extraEnv:
    # ref: https://github.com/jupyterhub/jupyterhub/blob/master/docs/source/reference/config-user-env.md#switching-to-jupyter-server
    JUPYTERHUB_SINGLEUSER_APP: jupyter_server.serverapp.ServerApp
```

yuvipanda:
Hmm, I'd like us to *not* switch to serverapp yet, and continue using the notebook package for now. The situation around jupter_server, nbclassic, notebook packages is very confusing to me :D And since the primary interface used in teaching is still the classic notebook package, I want us to stick to that as closely as possible for now. I'm also already running JupterLab 3.0 with the notebook package at berkeley, and it seems to be doing ok.

images/user/environment.yml
```diff
@@ -51,3 +49,7 @@ dependencies:
     - geojson
     - folium
     - geopy
+    # JupyterLab extensions
+    - jupyterlab_widgets==1.0.*
+    - jupytext==1.10.*
+    - jupyter-server-proxy==1.5.3
```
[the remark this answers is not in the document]
yuvipanda: I've merged the changes, @consideRatio! I've been able to run current version of jupyter-server-proxy with lab 3.0 without any issues though.

consideRatio:
> I'm also already running JupterLab 3.0 with the notebook package at berkeley, and it seems to be doing ok.

Oh so using ServerApp wasn't required? Ah

yuvipanda:
Nope! Check it out - https://github.com/berkeley-dsep-infra/datahub/blob/staging/deployments/datahub/images/default/infra-requirements.txt. Jupyterlab continues to function fine as a notebook extension

images/user/environment.yml
```diff
@@ -31,17 +31,15 @@ dependencies:
...
+    - jupyterlab==3.*
     - voila==0.1.23
     - nbgitpuller==0.9.*
     - jupyter-resource-usage==0.5.0
     - nbinteract==0.2.5
-    - jupytext==1.6.*
     - RISE==5.6.1
     - jupyter_contrib_nbextensions==0.5.1
     - notebook==6.1.4
     - jupyterhub==1.1.0
```
consideRatio, line 42: While it may be off topic, I think bumping jupyterhub to 1.3 should be done if we have z2jh 0.11 at least.

yuvipanda:
This LGTM, @GeorgianaElena :)

GeorgianaElena:
Thanks @yuvipanda! I'm building the image right now and want to test it on staging before merging.

GeorgianaElena:
For some reason, when trying to build the image, it gets stuck in:
```
Step 30/38 : RUN conda env update -p ${CONDA_DIR} -f /tmp/environment.yml
 ---> Running in 0b0a7ff246ae
Collecting package metadata (repodata.json): ...working... done
Solving environment: ...working... 
Warning: 2 possible package resolutions (only showing differing packages):
  - conda-forge/linux-64::pyrsistent-0.17.3-py38h497a2fe_2, conda-forge/noarch::tableschema-1.19.3-pyh9f0ad1d_0
  - conda-forge/linux-64::pyrsistent-0.16.0-py38h1e0a361_0, conda-forge/noarch::tableschema-1.20.0-pyh9f0ad1d_0done

Downloading and Extracting Packages
hdf4-4.2.13          | 951 KB    | ########## | 100% 
(...)
Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... done
```
:( @consideRatio or @yuvipanda, did you encounter this before and know what the issue may be?

yuvipanda:
hmm, how long is it stuck there? Maybe the docker image is full?

I unfortunately do not have enough conda experience to debug this :( This is why I always just use pip, though. So if we can't figure this out in a bit, let's just move the packages to pip.

GeorgianaElena:
I think yesterday I left it more than an hour and nothing happened.

The thing is that it used to work before bumping the version of some packages (I got some version conflicts) and now nothing :(
Pkgs version is probably unrelated though... I'll try again with the ones that gave conflicts, just to double check and because staring at a screen that's stuck makes me do desperate things :sweat_smile:

GeorgianaElena:
@yuvipanda, when pining pip to 20.2 (as per this comment https://github.com/conda/conda/issues/6986#issuecomment-669429070), it got pass the `conda env update` step :shrug:, but got stuck because of the nodejs version requirement of jupyterlab3.

Few questions:
* Does it make sense to have pip pinned? (I'm running `conda env update` with -v flag to see if I get any useful info about why it gets stuck with the newest pip)
* jupyterlab3 requires nodejs >=12.0.0, but we have this comment in our Dockerfile:
```
# Our pre-built R packages from rspm are built against system libs in focal
# rstan takes forever to compile from source, and needs libnodejs
# So we install older (10.x) nodejs from apt rather than newer from conda
```
I don't see it pinned to 10.x anywhere, but I guess this is the newest version available from apt, right?
I'm not sure what to do about this.

yuvipanda:
mmmm, fun.. Here's what I suggest:

1. Get rid of our environment.yaml file completely. We use python 3.8, which also comes by default with our miniforge install. We don't need to upgrade that right now, so we can let that go..
2. Move all the packages in environment.yml to a requirements.txt. This means we will install them with pip instead of conda, and that's ok.
3. JupyterLab only requires 12.x for installing extensions with `jupyter labextension` - it doesn't need nodejs at all for extensions installed via pip. So let's find the extensions that can't be installed via pip yet, and just ignore those for now. If those turn out to be super important, we can instead disable the `rstan` package for now.

How does this sound?

GeorgianaElena:
> How does this sound?

Sounds like a plan :superhero_woman: 

> JupyterLab only requires 12.x for installing extensions with jupyter labextension - it doesn't need nodejs at all for extensions installed via pip. So let's find the extensions that can't be installed via pip yet, and just ignore those for now. If those turn out to be super important, we can instead disable the rstan package for now.

So the labextentions we had before are:
```
    @jupyter-widgets/jupyterlab-manager@2 \
    @jupyterlab/server-proxy@2.1.2 \
    @jupyterlab-jupytext@1.2.1
```
I believe that only `jupyter-server-proxy` needs nodejs still, but there's an open PR for this: https://github.com/jupyterhub/jupyter-server-proxy/pull/245 :tada:

consideRatio:
In a image having quite a bit of conda packages, I got a speedup from 60 minute builds to 40 minute builds using `mamba`. The time invested in `conda` command was cut very significantly, and I wouldn't be that surprised if `conda` would stall for 60+ minutes if it was a bit starved of RAM or something like that.

For reference, this is how I've most recently installed the base python environment using `mamba` instead of `conda`, and it works great for a large image with lots of dependencies!

```Dockerfile
# Install a minimal conda based python disitribution and ensure we use a
# specific python version. Instead of using conda, we use mamba which increase
# performance. Instead of the default channel, we use conda-forge by default as
# well.
#
# Ref for latest versions and SHA: https://github.com/conda-forge/miniforge/releases
ENV MAMBAFORGE_VERSION=4.9.2-5 \
    MAMBAFORGE_SHA256SUM=7f0ad0c2f367751f7878d25a7bc1b4aa48b8dcea864daf9bc09acb595102368b \
    PYTHON_VERSION=3.7
ENV MAMBAFORGE_FILENAME=Mambaforge-${MAMBAFORGE_VERSION}-Linux-x86_64.sh
ENV MAMBAFORGE_URL=https://github.com/conda-forge/miniforge/releases/download/${MAMBAFORGE_VERSION}/${MAMBAFORGE_FILENAME}
RUN cd /tmp \
    # Download, verify checksum, install, cleanup
 && wget --quiet $MAMBAFORGE_URL \
 && echo "${MAMBAFORGE_SHA256SUM}  ${MAMBAFORGE_FILENAME}" | sha256sum --quiet -c - \
 && /bin/bash "${MAMBAFORGE_FILENAME}" -f -b -p $CONDA_DIR \
 && rm "${MAMBAFORGE_FILENAME}" \
    # Install desired Python version, cleanup
 && mamba install python=${PYTHON_VERSION} --quiet --yes \
 && mamba clean --all --yes -f \
    # Preserve behavior of miniconda - packages come from conda-forge + defaults
 && conda config --system --append channels defaults \
 && fix-permissions $CONDA_DIR $HOME
```

GeorgianaElena:
I was aware of mamba, but didn't get to use it yet. The speedup seems really impressive :tada: Thanks for sharing @consideRatio!

I won't mind giving mamba a try for this. @yuvipanda, what do you think, go mama :woman_dancing: or stick with pip? :D

GeorgianaElena:
Also, running `conda env update` with `-v` flag showed nothing, but with `-vvv`, the last debug message before hanging was this one:
```
DEBUG conda.common.signals:signal_handler(58): de-registering handler for Signals.SIGABRT
DEBUG conda.common.signals:signal_handler(58): de-registering handler for Signals.SIGINT
DEBUG conda.common.signals:signal_handler(58): de-registering handler for Signals.SIGTERM
DEBUG conda.common.signals:signal_handler(58): de-registering handler for Signals.SIGQUIT
```
It really doesn't make any sense for me, and I don't see any connection between pip and de-registering signal handlers :shrug:

consideRatio:
The approach I've evolved to is to rely on pip as much as possible, but when needed, let individual packages be installed with mamba using the conda-forge channel.

yuvipanda:
@GeorgianaElena I definitely want us to explore mamba! But I want us to finish this up as soon as we possibly can and move on to other tasks, here and in UToronto, so let's open an issue for that instead?

GeorgianaElena:
Thanks for sharing all this knowledge @consideRatio <3

yuvipanda:
My approach to packaging is also exactly @consideRatio's :D I used conda here as an experiment to see what kinda complexity *I* can deal with, and unfortunately it looks like the amount of knowledge I have isn't good enough to deal with conda packaging complexities...

GeorgianaElena:
@yuvipanda, do you want to wait for a new release of jupyter-server-proxy that doesn't require nodejs anymore or just go without it until then?

yuvipanda:
@GeorgianaElena if we can test that everything else in this PR works, we can just wait until that release is made. Not urgent :D

GeorgianaElena:
I built the image (as it is, without jupyter-server-proxy extension enabled for jupyterlab, but pkg installed), pushed it, then deployed it to staging.

### What happened:
* we have jupyterlab3
* jupytext seems to work
* ipywidgets seems to work
* health check failed with a timeout error and saw this warning  related to `jupyter-server-proxy` in the pod logs:
```bash
[W 2021-02-10 16:07:43.160 SingleUserNotebookApp notebookapp:1950] Error loading server extension jupyter_server_proxy
      File "/opt/conda/lib/python3.7/site-packages/notebook/notebookapp.py", line 1945, in init_server_extensions
        func(self)
      File "/opt/conda/lib/python3.7/site-packages/jupyter_server_proxy/__init__.py", line 27, in load_jupyter_server_extension
        server_processes += get_entrypoint_server_processes()
      File "/opt/conda/lib/python3.7/site-packages/jupyter_server_proxy/config.py", line 75, in get_entrypoint_server_processes
        make_server_process(entry_point.name, entry_point.load()())
      File "/opt/conda/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2460, in load
        self.require(*args, **kwargs)
      File "/opt/conda/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2483, in require
        items = working_set.resolve(reqs, env, installer, extras=self.extras)
      File "/opt/conda/lib/python3.7/site-packages/pkg_resources/__init__.py", line 790, in resolve
        raise VersionConflict(dist, req).with_context(dependent_req)
    pkg_resources.ContextualVersionConflict: (nbformat 4.4.0 (/opt/conda/lib/python3.7/site-packages), Requirement.parse('nbformat>=5.0'), {'nbclient'})
[W 2021-02-10 16:07:44.255 LabApp] 'terminado_settings' has moved from NotebookApp to ServerApp. This config will be passed to ServerApp. Be sure to update your config before our next release.
[W 2021-02-10 16:07:44.255 LabApp] 'terminado_settings' has moved from NotebookApp to ServerApp. This config will be passed to ServerApp. Be sure to update your config before our next release.
```

yuvipanda:
ah, fun! I think the 'InteractExporter' is from nbinteract, which we can remove - I don't think anyone is actively using that, and Jupyyterbook's thebe does similar things I think.

The other one is concerning :|

GeorgianaElena:
Removing nbinteract from my single user server seemed to have solved the 500 error issue.
Though in the logs I'm still seeing bunch of similar warnings like:
```
[W 2021-02-10 17:36:16.906 SingleUserNotebookApp configurable:168] Config option `template_path` not recognized by `ExporterCollapsibleHeadings`.  Did you mean one of: `extra_template_paths, template_name, template_paths`?
```

choldgraf:
Hey all - what is the state of this one? Are we blocked on the error messages reported above?

yuvipanda:
All the labextensions that were giving us trouble have now been upgraded to work with JupyterLab 3.0, so this can be picked up again.

choldgraf:
@GeorgianaElena do you wanna pick this one back up?

GeorgianaElena:
@choldgraf, sure! Thanks for the heads up!

GeorgianaElena:
I've rebased this, bumped jupyter-server-proxy to its latest version (JupyterLab3 compatible :tada:), manually built and pushed the new user image and deployed it to staging hub.

At a first glance, the extensions seem to still be working and the Lab version is > 3. Since we're manually testing these changes, I would love if someone else could try the changes too. See JupyterLab3  + extensions in action here https://staging.pilot.2i2c.cloud/

deploy/build.py
```diff
@@ -2,7 +2,7 @@
 import os
 import subprocess
 
-HERE = os.path.dirname(os.path.abspath(__file__))
+HERE = os.getcwd()
```
damianavila, line 5: Why did you change this line? I would keep using the existing line and just look for the proper directory if that is the problem.

damianavila:
Just a general comment about switching to pip. In the last few years, I have seen conda showing a lot of issues being the slowness the most problematic one. Recently I started to use mamba and I think it could become a super interesting alternative. I still think the agnostic nature of mamba/conda has been super useful in the past and one of the differentiators at the time to build complex environments. So, I would love to see mamba exploration in the future for these hubs (just a thought for the future, not asking mamba on this PR :wink:)

images/user/requirements.txt
```diff
@@ -0,0 +1,42 @@
...
+ipywidgets==7.5.1
+otter-grader==2.1.3
+datascience
+geojson
+folium
+geopy
+# JupyterLab extensions
+jupyterlab_widgets==1.0.*
+jupytext==1.10.*
+jupyter-server-proxy==3.0.2
```
damianavila, line 42: Why we are not defining specific versions in this requirement file? I tend to always be explicit in the versions I am deploying on a base image, otherwise, you do not "control" which versions you are actually deploying... and that can introduce subtle bugs. I understand there is some additional work in keeping the version relevant as time pass but usually, that is a small price to pay, IMHO.

images/user/requirements.txt
```diff
@@ -0,0 +1,42 @@
...
+catalystcoop.pudl
+dask==2020.12.0
+distributed==2020.12.0
+dask-gateway==0.9.0
+# Infrastructure things
+jupyterlab==3.*
+voila==0.2.6
+nbgitpuller==0.9.*
+jupyter-resource-usage==0.5.0
+RISE==5.6.1
```
damianavila, line 27: Hey, we have released 5.7.1 a long time ago :wink:

choldgraf:
I had a conversation with @damianavila yesterday where he mentioned that we might run into some weirdness with the new Jupyter Server coming bundled with newer versions of Jupyter Lab, and this might cause unpredictable behavior (e.g., nbgitpuller might not work with JupyterLab vs. with the old Jupyter Notebook server). Just flagging it for attention.

deploy/build.py
GeorgianaElena: Thanks for asking about this @damianavila!

Before changing this, I got this error when trying to build the docker image:
```
fatal: ambiguous argument 'pilot-hubs/deploy/images/user': unknown revision or path not in the working tree.
```
Which is because de ` was the right call
 build.py` has been moved to the `deploy` dir and the images are still in the old location.

My first thought was to move the images where the build script is (the `deploy` dir), but then I bumped into:
```
File "deploy/build.py", line 36, in last_modified_commit
    return substring_with_alpha(commit_hash)
  File "deploy/build.py", line 26, in substring_with_alpha
    return s[:max(min_len, first_alpha(s))]
  File "deploy/build.py", line 15, in first_alpha
    raise Exception("No alpha characters in string: {}".format(s))
Exception: No alpha characters in string: 
```
Which is from this line here: https://github.com/2i2c-org/pilot-hubs/blob/6105de7428bb9e7442c56fd61aa031ac2e8f5877/deploy/build.py#L59
That builds the new image tag using the last commit that modified the image. And since the images were moved to a new location, then no prev commit that modified them was found.
So I've decided to let the images where they were and tell the `build.py` where to find them. Not sure if this was the best decision though..

Hope I made a bit of sense in explaining my thinking process :D

images/user/requirements.txt
GeorgianaElena: I've seen it in other repos and used it myself too, but I don't know the exact answer to this question.

My intuition is that it's expected that an increase in the patch version to only fix bugs and don't cause any breaking change.
So pinning only the major and minor versions should be enough to prevent any issue caused by a pkg incompatibility.

damianavila: That's a sensible intuition. Although I should say I have been bitten by patch version incompatibilities before 😉 , this is why I am biased to be stricter on the version specification.

deploy/build.py
damianavila: Thanks for the whole explanation!

> So I've decided to let the images where they were and tell the build.py where to find them. Not sure if this was the best decision though..

I think it is a sensible decision, my question was around why you were using `getcwd` to find the images.
If you keep
```python
HERE = os.path.dirname(os.path.abspath(__file__))
```
maybe this works?
```python
tag = last_modified_commit(os.path.join(os.path.dirname(HERE), "images/user")) 
```

NOT tested, btw...

GeorgianaElena: > maybe this works?

Yep, it worked! Just added a commit with this. Thank you!

GeorgianaElena:
I've rebased this again and built a new image with the newer RISE version, but I don't know how to proceed about this, given the unpredictable behavior mentioned by @choldgraf and @damianavila. 

Do you know if there's any issue opened on `nbgitpuller`'s side about this behavior that I could  read more about?

consideRatio:
@choldgraf wrote
> I had a conversation with @damianavila yesterday where he mentioned that we might run into some weirdness with the new Jupyter Server coming bundled with newer versions of Jupyter Lab, and this might cause unpredictable behavior (e.g., nbgitpuller might not work with JupyterLab vs. with the old Jupyter Notebook server). Just flagging it for attention.

@GeorgianaElena wrote
> Do you know if there's any issue opened on nbgitpuller's side about this behavior that I could read more about?

I think I've experienced this myself. I'll open an issue in nbgitpuller and describe what I observe. Separate consideration: should a malfunctional nbgitpuller be a blocker for this PR or not?

yuvipanda:
fwiw Berkeley has been running JupyterLab 3.0 with nbgitpuller for many months now with no issues

consideRatio:
> fwiw Berkeley has been running JupyterLab 3.0 with nbgitpuller for many months now with no issues

Are you running notebook servers or jupyter_server servers?

deploy/build.py
damianavila: Super! I prefer to avoid using `getcwd` in this sort of case (when you are locating files/dirs). Thanks!

damianavila:
> Are you running notebook servers or jupyter_server servers?

That's the key... see my comment here: https://github.com/pangeo-data/pangeo-docker-images/issues/206#issuecomment-829242814

Particularly this piece (pasted here for a quick look):

> AFAIK, if you start with jupyter notebook, that is calling the old server whereas if you start with jupyterlab, that is calling the new one. So you may end up in a situation where you started with lab (using the underlying new server) but then redirected to /tree to use the legacy notebook view.
In that scenario, if you have a server extension that is not compatible with the new server, it will fail... I think...

The fix for this sort of scenarios is to make your extension also compatible with the new server: https://jupyter-server.readthedocs.io/en/latest/developers/extensions.html#migrating-an-extension-to-use-jupyter-server

And support both, the legacy and the new server, for at least some time, until the transition completes...

consideRatio:
@damianavila do you think what I observed (https://github.com/jupyterhub/nbgitpuller/issues/176) as a web-frontend issue is a sympton of this, or do you think there are two separate issues to report in nbgitpuller?

damianavila:
I think it is a symptom... if you control that environment, try launching the legacy notebook app vs the jupyterlab app and check if things behave in a different way.

consideRatio:
@damianavila thanks, I opened https://github.com/jupyterhub/nbgitpuller/issues/177!

damianavila:
>  @damianavila is that your understanding as well?

IIRC JupyterLab 3 actually has the new jupyter_server as a requirement but, I think, we are defaulting to `/tree` in our hubs: https://github.com/2i2c-org/pilot-hubs/blob/master/hub-templates/base-hub/values.yaml#L144

So we should be fine, maybe? We should definitely test nbgitpuller on top of the current PR and see how it behaves...

choldgraf:
Hey all - what are the blockers to getting this one merged?

damianavila:
> Hey all - what are the blockers to getting this one merged?

⬇️ 

>We should definitely test nbgitpuller on top of the current PR and see how it behaves...

I guess it is just a matter of testing if the nbgitpuller works as expected. And solve the conflicts.
I left some other general comments about package version specification but those should not be a blocker, IMHO.

GeorgianaElena:
I've just fixed the conflicts and re-built the user image. I also tried to test out `nbgitpuller` and it seemed to be working.

What I did:
-  created nbgitpuller links that were either for lab or for the classical interface. The repo got cloned ok in both cases.
- defaulted the staging hub to `/lab` and then tried the same thing from above again. Repos got cloned ok.

A strage thing I noticed was that running ` jupyter server extension list` (as per https://github.com/pangeo-data/pangeo-docker-images/issues/206#issuecomment-829242814), didn't show `nbgitpuller` at all when the hub was defaulted to `/lab`.
`nbgitpuller` only showed up when opening up a terminal from lab, when the hub was defaulting to `/tree`.

I would highly appreciate if this nbgitpuller situation got another pair of :eyes: If anyone wants to give this a try, the staging hub (https://staging.pilot.2i2c.cloud) is using lab v3 and it's defaulting to this interface.

GeorgianaElena:
I haven't used `nbgitpuller` much either. So I cannot be sure that everything works with this new lab3 setup. 

I'm thinking of two approaches, in order to unblock this PR:
- Merge this tomorrow morning if nobody has any objection until then. Then watch for user bug reports and try to solve them, or revert this entirely
- Implement [@consideRatio's suggestion](https://github.com/jupyterhub/nbgitpuller/issues/177), then merge this PR and solve any issues that may come up. 

What do you think?

damianavila:
>Merge this tomorrow morning if nobody has any objection until then. Then watch for user bug reports and try to solve them, or revert this entirely

I am not sure to what extent our users are actually using nbgitpuller, but I think this ⬆️  is an OK approach since we have some previous reports that things are going OK with this in other places and we have already identified the possible issue, so if anything arises, we should be able to quickly fix it/workaround it/revert it.

choldgraf:
wahooo!!! congrats @GeorgianaElena and thanks everybody else for helping out :-)

PR #208, opened 2021-02-09
Username pattern for mills

GeorgianaElena:
Didn't know how to make this smarter and don't copy the admins again in the regex, since the admin users have to comply with the `username_pattern` too :(

I have tried:
* Adding the admins in `hub.py` but `hubs.yaml` takes precedence over `generated_config`
* Some functions from here https://helm.sh/docs/chart_template_guide/function_list/ but didn't work and have no idea how to use them or they should work with this at all

### DIFF
--- a/auth.py
+++ b/auth.py
@@ -109,7 +109,7 @@ class KeyProvider:
             'username_key': username_key,
             'client_id': client['client_id'],
             'client_secret': client['client_secret'],
-            'scopes': ['openid', 'name', 'profile', 'email']
+            'scope': ['openid', 'name', 'profile', 'email']
         }
 
         return auth
--- a/hubs.yaml
+++ b/hubs.yaml
@@ -709,4 +709,4 @@ clusters:
                     - georgiana.dolocan@gmail.com
                     - aculich@berkeley.edu
                     - jpercy@berkeley.edu
-                  username_pattern: '^.+@2i2c.org$|yuvipanda@gmail.com|choldgraf@gmail.com|georgiana.dolocan@gmail.com|aculich@berkeley.edu|jpercy@berkeley.edu|deployment-service-check'
+                  username_pattern: '^.+@mills.edu$|yuvipanda@gmail.com|choldgraf@gmail.com|georgiana.dolocan@gmail.com|aculich@berkeley.edu|jpercy@berkeley.edu|deployment-service-check'
--- a/hubs.yaml
+++ b/hubs.yaml
@@ -709,4 +709,4 @@ clusters:
                     - georgiana.dolocan@gmail.com
                     - aculich@berkeley.edu
                     - jpercy@berkeley.edu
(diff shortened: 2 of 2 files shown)


auth.py
```diff
@@ -109,7 +109,7 @@ def get_client_creds(self, client, connection_name):
             'username_key': username_key,
             'client_id': client['client_id'],
             'client_secret': client['client_secret'],
-            'scopes': ['openid', 'name', 'profile', 'email']
+            'scope': ['openid', 'name', 'profile', 'email']
```
yuvipanda, line 112: Can you leave a note in the commit message about what this change is for?

hubs.yaml
```diff
@@ -703,11 +703,10 @@ clusters:
...
-                    - yuvipanda@2i2c.org
+                  admin_users: &mills_admins
                     - yuvipanda@gmail.com
-                    - choldgraf@2i2c.org
-                    - georgianaelena@2i2c.org
+                    - choldgraf@gmail.com
                     - georgiana.dolocan@gmail.com
                     - aculich@berkeley.edu
                     - jpercy@berkeley.edu
+                  username_pattern: '^.+@mills.edu$|yuvipanda@gmail.com|choldgraf@gmail.com|georgiana.dolocan@gmail.com|aculich@berkeley.edu|jpercy@berkeley.edu|deployment-service-check'
```
yuvipanda, line 712:
```suggestion
                  username_pattern: '^(.+@mills\.edu|yuvipanda@gmail\.com|choldgraf@gmail\.com|georgiana\.dolocan@gmail\.com|aculich@berkeley\.edu|jpercy@berkeley\.edu|deployment-service-check)$'
```

Regular expressions are an easy way to shoot yourself in the foot - particularly around '^' and '$'. Without those, the email 'notactuallyyuvipanda@gmail.com' would match the regex, since it'll match yuvipanda@gmail.com! Similar with '.', since it matches 'any character' regex.

Longer term we should do something else but this should do for now :D

auth.py
GeorgianaElena: Sorry about this, it should have probably been better to open an issue or another PR about this.
But when I checked the logs to see if the regex works, I saw this error:
```
[W 2021-02-09 10:37:19.991 JupyterHub configurable:190] Config option `scopes` not recognized by `GenericOAuthenticator`.  Did you mean `scope`?
```
I checked with `oauthenticator` too:
https://github.com/jupyterhub/oauthenticator/blob/3d249bba92db34519c016e81c8281ace089ea01d/oauthenticator/oauth2.py#L266

consideRatio: This is from the old z2jh `jupyterhub_config.py`

```python
set_config_if_not_none(c.OAuthenticator, "scope", "auth.scopes")
```

And this is from the deprecation logic which seem to be correct also, pweh!

```yaml
{{- define "jupyterhub.authDep.remapOldToNew.map" }}
scopes: OAuthenticator.scope
# ...
```

GeorgianaElena:
I have deployed this already to the mills hub and seems to be working as desired. Merging :D

PR #224, opened 2021-02-12
Document how to add multiple domains to a hub

GeorgianaElena:
Closes https://github.com/2i2c-org/pilot-hubs/issues/214

### DIFF
--- a/docs/configure.md
+++ b/docs/configure.md
@@ -200,16 +200,6 @@ Assuming there is a hub named `foo` in the `2i2c` cluster and we want to add the
           domain: foo.pilot.2i2c.cloud
         (...)
       ```
-    * In Auth0's **Application Settings page**:
-
-      ```{figure} images/auth-applications.png
-      ```
-
-      Add the callback URL of the new domain to the list of **Allowed Callback URLs**:
-
-      ```{figure} images/auth0-allowed-callback-urls.png
-      ```
-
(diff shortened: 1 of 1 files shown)


docs/configure.md
```diff
@@ -148,6 +148,69 @@ So we want to manage authentication by:
...
+
+      ```yaml
+      - name: foo
+        (...)
+        auth0:
+          connection: google-oauth2
+          domain: foo.pilot.2i2c.cloud
+        (...)
+      ```
+    * In Auth0's **Application Settings page**:
```
yuvipanda, line 203: Aaah, can you automate this [here](https://github.com/2i2c-org/pilot-hubs/blob/master/auth.py#L50) instead?

yuvipanda (changes requested):
A point of automation that should make this easier :) I want to try remove manual steps as much as possible - not everyone will have access to auth0.

docs/configure.md
GeorgianaElena: Sure! I just followed the instructions in the error page and missed that function :sweat_smile:

GeorgianaElena: I just opened https://github.com/2i2c-org/pilot-hubs/pull/225 to address this :)

yuvipanda:
\o/ love it!

PR #225, opened 2021-02-12
Ensure we are adding all the callback urls if hub has multiple domains

GeorgianaElena:
If hub has multiple domains linked to it, we need to add a callback url of the form `https://{domains}/hub/oauth_callback` in auth0 for each of them.

This should automate this task.

### DIFF
--- a/auth.py
+++ b/auth.py
@@ -36,36 +36,46 @@ class KeyProvider:
             for connection in self.auth0.connections.all()
         }
 
-    def create_client(self, name, domain):
+    def create_client(self, name, domains):
+        callbacks = self._get_callback_url_list(domains)
+
         client = {
             'name': name,
             'app_type': 'regular_web',
-            'callbacks': [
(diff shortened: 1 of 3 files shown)


yuvipanda:
Awesome! Does this mean we no longer have to explicitly set auth0 domain in hubs.yaml?

GeorgianaElena:
Unfortunately it's still being used in health checks:

https://github.com/2i2c-org/pilot-hubs/blob/16fcdf557cd26429caffd099150378e30eb4c3f4/hub.py#L316

Does it make any sense to do a health check per domain?

yuvipanda:
Ah, no I think we don't need to do a health check per-domain. But then we might need to move the location of that config item, since it's not related to auth0 anymore.

How about two things:

1. Run the home page health check on all domains
2. For the spawn-server health check, run it on a random domain? Since theoretically they should all work.

yuvipanda:
@GeorgianaElena feel free to merge this after testing it out :)

PR #227, opened 2021-02-12
Update the version of our Docker image

zaneselvans:
Updated some dependencies and integrated changes from sprint30.


yuvipanda:
Thanks, @zaneselvans!

PR #230, opened 2021-02-17
Dump generated values as JSON

yuvipanda:
I was getting some syntax errors when dumping as YAML,
possibly as discrepancies between ruamel.yaml vs go YAML.
I didn't wanna investigate too deply, so I switched the
generated values to JSON.

### DIFF
--- a/hub.py
+++ b/hub.py
@@ -416,9 +416,9 @@ class Hub:
         # Temporary disable this until https://github.com/2i2c-org/pilot-hubs/issues/206 is fixed
         # self.setup_nfs_share()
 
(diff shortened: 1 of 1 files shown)


GeorgianaElena (approved):
Go json :rocket: 

The help page of `helm upgrade` says `-f` is to `specify values in a YAML file or a URL (can specify multiple)`. But since both `yaml.dump` and `json.dump` produce strings, shouldn't be an issue, right?

yuvipanda:
@GeorgianaElena yeah shouldn't be an issue. Plus, *all JSON* is valid YAML by design! So it works out.

PR #250, opened 2021-02-18
Set image explictly for grenoble hub

yuvipanda:


yuvipanda:
Was missing this, and the manual workaround in https://github.com/2i2c-org/pilot-hubs/issues/206#issuecomment-780803655

PR #265, opened 2021-02-23
Rename and move new hub and goal issue templates

GeorgianaElena:
The new hub and goal issue templates were in a `ISSUE_TEMPLATES` directory, while the debrief one was in `ISSUE_TEMPLATE` and only the latter was shown.

This puts all the templates in the same `ISSUE_TEMPLATE` directory and changes the template file type from `yml` to `md`. The `yml` probably works too, but haven't tried it and all the GitHub examples I've seen so far were markdown files.

We do use a `yml` file for the issue templates called `config.yaml` file in the `jupyterhub` org, but that's to create issue templates that redirect to other sites (Discourse in jupyterhub's case).

Sorry for not spotting this in time.


choldgraf:
Oops that was a silly mistake if mine, thanks for noticing!

PR #266, opened 2021-02-23
Make sure volumes don't get duplicated

GeorgianaElena:


yuvipanda:
\o/ Thanks :) I forgot that dictionary equivalents ok in Python for the most part.

PR #270, opened 2021-02-24
Bump pangeo notebook image

yuvipanda:


GeorgianaElena:
@yuvipanda, I believe the latest image tag is `2021.02.19` instead of (2020.02.19)
2020 doesn't deserve to be repeated  :sweat_smile:

yuvipanda:
Haha, good catch @GeorgianaElena! https://github.com/2i2c-org/pilot-hubs/pull/271 has it

PR #278, opened 2021-03-01
Replace nodeSelector with nodeAffinity for singleuser notebook servers.

GeorgianaElena:
This will allow the user servers to run on core nodes from the core pool too and potentially make server spawn faster.
Should fix the timeout errors seen in the deployments tests that we run.

This PR tries to replicate the z2jh behavior of  `scheduling.userPods.nodeAffinity`.

### Things that I'm unsure of:
- I'm not sure if setting the node affinity in both `KubeSpawner` and `jupyterhub.userAffinity` was necessary, as setting it only in KubeSpawner seems to produce the desired behaviour:
single-user-server-affinity
- Is there an issue that we have 2 node affinities set for the single user servers and they have the same weight?
- If so, is there a way to remove the one coming from z2jh?
 
cc @yuvipanda and @consideRatio

### DIFF
--- a/hub-templates/base-hub/values.yaml
+++ b/hub-templates/base-hub/values.yaml
@@ -47,16 +47,6 @@ jupyterhub:
           memory: 64Mi
         limits:
           memory: 1G
-    userPods:
-      nodeAffinity:
-        preferredDuringSchedulingIgnoredDuringExecution:
-        - preference:
-            matchExpressions:
-            - key: hub.jupyter.org/pool-name
-              operator: In
-              values:
-              - user-pool
-          weight: 100
   prePuller:
     continuous:
       enabled: false
@@ -96,6 +86,14 @@ jupyterhub:
         subPath: _shared
     startTimeout: 600 # 10 mins, because sometimes we have too many new nodes coming up together
     defaultUrl: /tree
+    extraNodeAffinity:
+      preferred:
+        - weight: 100
+          preference:
+            matchExpressions:
+            - key: hub.jupyter.org/pool-name
+              operator: In
+              values: [user-pool]
     image:
       name: set_automatically_by_automation
       tag: 4066a62
--- a/hub-templates/base-hub/values.yaml
+++ b/hub-templates/base-hub/values.yaml
@@ -78,6 +78,16 @@ jupyterhub:
       enabled: false
       letsencrypt:
         contactEmail: yuvipanda@gmail.com
+  userAffinity:
+    nodeAffinity:
+        preferredDuringSchedulingIgnoredDuringExecution:
+        - preference:
+            matchExpressions:
+            - key: hub.jupyter.org/pool-name
+              operator: In
+              values:
+              - user-pool
+          weight: 100
   singleuser:
     admin:
       extraVolumeMounts:
@@ -86,8 +96,6 @@ jupyterhub:
         subPath: _shared
     startTimeout: 600 # 10 mins, because sometimes we have too many new nodes coming up together
     defaultUrl: /tree
-    nodeSelector:
-      hub.jupyter.org/pool-name: user-pool
     image:
       name: set_automatically_by_automation
       tag: 4066a62
@@ -186,3 +194,19 @@ jupyterhub:
             return super().start(*args, **kwargs)
 
         c.JupyterHub.spawner_class = CustomSpawner
+      06-custom-affinity: |
+        node_selector = dict(
+            matchExpressions=[
+                dict(
+                    key="hub.jupyter.org/pool-name",
+                    operator="In",
+                    values=["user-pool"],
+                )
+            ],
+        )
+        c.KubeSpawner.node_affinity_preferred.append(
+            dict(
+                weight=100,
+                preference=node_selector,
+            ),
+        )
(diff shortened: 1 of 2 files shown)


consideRatio (commented):
> I'm not sure if setting the node affinity in both KubeSpawner and jupyterhub.userAffinity was necessary, as setting it only in KubeSpawner seems to produce the desired behaviour:

`scheduling.userPods.nodeAffinity` is a way to configure KubeSpawner via the Helm chart. If you configure KubeSpawner directly that would also be fine but use either or.

> Is there an issue that we have 2 node affinities set for the single user servers and they have the same weight?

No! You can end up with score of 0, 100, or 200 for a node, which will influence the preference of nodes. Since only one of these labels may show up, in practice you will never be influenced by the other label.

> If so, is there a way to remove the one coming from z2jh?

Yes `scheduling.userPods.nodeAffinity.matchNodePurpose=ignore` disables the automatic injection of the affinity for hub.jupyterorg/node-purpose=user label.

---

### Suggested changes
- [x] Remove the hub.extraConfig entry directly configuring KubeSpawner.
- [x] Fix the Helm chart config currently under `userAffinity.nodeAffinity` to be under `scheduling.userPods.nodeAffinity`.

consideRatio (approved):
This LGTM @GeorgianaElena! :)

GeorgianaElena:
Thank you @consideRatio :)

I tried testing this in the following scenario, but I'm a bit unsure if the right thing it's happening:
1. I started a user server when only one core node and no user node was available.
    - the user server got scheduled on the core node (:heavy_check_mark:  this is what should have happened, right?)
2. I started a user server when only one core node and one user node were available.
    -  the user server got scheduled on the core node again (:x: It should have preferred the user node instead, right?)
3. I'm only seeing the z2jh default injected node affinity now :confused: 
current-node-affinity

yuvipanda:
@consideRatio can we not just use https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html#singleuser-extranodeaffinity instead? I thought that was the way to set node affinity for user pods.

I think the two scenarios you tested are right, @GeorgianaElena! The third one is that core node is *full*, new node needs to come up. It should spawn a user node, not a core node.

consideRatio:
> I'm only seeing the z2jh default injected node affinity now confused

> @consideRatio can we not just use https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html#singleuser-extranodeaffinity instead? I thought that was the way to set node affinity for user pods.

Ah, yes @yuvipanda! My suggestion was wrong @GeorgianaElena!

I got confused by the documentation. I thought `scheduling.userPods.nodeAffinity` was a configuration option, while in practice it was just a category of configuration wherein we had the option `matchNodePurpose`.

GeorgianaElena:
Thank you @yuvipanda and @consideRatio for reviewing this :tada: 

> The third one is that core node is full, new node needs to come up

I was thinking to test this by doing a manual deploy to all of the hubs while checking the number of nodes being created in the cluster. Alternatively we could merge this and let the CI do the deployment. @yuvipanda, what do you think?

Also, should we also increase the 5 min spawn timeout of the hub health check? I'm thinking that at some point (when the core node is full, a user node will need to be created) it will need more?

yuvipanda:
@GeorgianaElena manually doing a deploy to test this seems like a good idea.

We might have to increase the node timeout unfortunately. Can you get a measure of how long it takes to launch a new node + pod on it and we can set the timeout to that?

GeorgianaElena:
> Can you get a measure of how long it takes to launch a new node + pod on it and we can set the timeout to that?

I tried this by creating a server on the dask-staging hub when no user node was available and with the user-pool selector still in place. This is what happened:
server-pod-events

Server appeared *ready* in the hub logs:
```
[I 2021-03-02 10:19:15.348 JupyterHub users:664] Server georgiana.dolocan@gmail.com is ready
```

So, it looks like it took the server 3m2s to start. This includes:
- launch a new node and pod
- image pull
- notebook server start

yuvipanda:
That is actually quite fast! Although, just before that I actually logged in to https://swot-ocean-pangeo.pilot.2i2c.cloud and maybe that spawned a node? If it spawned a new node, there's usually an entry about 'triggered scale up' in the event history. Sorry :( I'll log off now...

GeorgianaElena:
This time I made sure there was only the core node available and the scale up event was present:
Screenshot from 2021-03-02 13-17-18
```
[I 2021-03-02 11:15:41.449 JupyterHub users:664] Server georgiana.dolocan@gmail.com is ready
```

Everything still happened in about 3m...

I'll try the manual deploy to all the hubs once the user pool scales down to 0 again (is there a way to force this somehow?) and check the pod events if/when it gives a timeout.

yuvipanda:
woah, that's awesome! Yeah, please try a full deploy and see what happens

GeorgianaElena:
Everything passed :sweat_smile: 

I'll however increase the timeout with 60s (around how much it takes the user node pool to scale up) because I saw some tests taking about 4min even though there was a user node available.
tests

Hope this will make the tests fail less often at least :confused:

GeorgianaElena:
@yuvipanda, merge this and :crossed_fingers: the tests become more reliable?

PR #282, opened 2021-03-03
Restructure hub template doc, update diagram and add more info

GeorgianaElena:
Ref: https://github.com/2i2c-org/pilot-hubs/issues/274
Closes https://github.com/2i2c-org/pilot-hubs/issues/162


docs/configure.md
```diff
@@ -49,25 +49,35 @@ The hubs are configured and deployed using *hub templates*. Because each hub
...
-- No persistent storage
-- No home page template
+- `ephemeral-hub`
 
+  The **ephemeral-hub template** helps deploying temporary, transient binder-style hubs.
+    - Doesn't require any user authentication.
+    - Permits limiting of resources:
+      * memory / CPU limits
+      * maximum number of concurrent user servers
+    - Has a more aggressive culling:
```
yuvipanda, line 66:
```suggestion
    - Has more aggressive culling:
```

yuvipanda (changes requested):
One comment, but text looks good! The image is empty?

yuvipanda:
In addition, I think we need to document https://github.com/2i2c-org/pilot-hubs/blob/master/hub.py#L356. If you're using daskhub or ephemeral hub, config in `hubs.yaml` needs to be nested under a `base-hub` key. This is because `base-hub` is now a sub chart of `daskhub`, and `jupyterhub` (z2jh) is a sub chart of `base-hub` (https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#overriding-values-from-a-parent-chart). Does this make sense?

GeorgianaElena:
> The image is empty?

It shouldn't be: https://github.com/GeorgianaElena/pilot-hubs/blob/docs/hub-templates/docs/images/config-flow.png

>  Does this make sense?

Yes! Ty!

yuvipanda:
Ah, GitHub shows the images like this:

image

But they're clearly there! 

And they're pretty damn awesome! I'll merge this now :)

docs/configure.md
```diff
@@ -49,29 +49,60 @@ The hubs are configured and deployed using *hub templates*. Because each hub
...
 
-Currently there are two hub templates available:
-- the `base-hub` template
-- the `ephemeral-hub` template
+The graphic below, shows the relationship between the hub templates and the other
+config files and how they are merged together when deploying a pilot hub.
 
-For example, the **ephemeral hub** is a special kind of hub that is built using the *ephemeral-hub template*
-and has the following features:
+```{figure} images/config-flow.png
```
choldgraf, line 55: @GeorgianaElena can you add a comment in these docs (with `% `) that says where these images were created, and how a different team member could update them if they wanted to do so?

PR #283, opened 2021-03-03
Add a SCRATCH_BUCKET per dask hub

yuvipanda:
A temporary bucket that's cleared every 7 days,
and provides full access to all the users on that
hub. See
https://github.com/pangeo-data/pangeo-cloud-federation/issues/610
for reasons why this is very useful.

### DIFF
--- a/hub-templates/daskhub/templates/env-vars.yaml
+++ b/hub-templates/daskhub/templates/env-vars.yaml
@@ -1,0 +1,7 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: cloud-env-vars
+data:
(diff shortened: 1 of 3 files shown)


choldgraf:
Is this documented somewhere? We really need to be in the habit of documenting these things! Or opening issues to document them after adding them.

yuvipanda:
@choldgraf I opened https://github.com/2i2c-org/pilot-hubs/issues/284

choldgraf:
Great thanks - sorry to be annoying about this, it's just that if we don't document, nobody will know this functionality exists! And team members also won't know that it exists!

PR #295, opened 2021-03-09
Document health check

GeorgianaElena:
Closes https://github.com/2i2c-org/pilot-hubs/issues/234


GeorgianaElena:
@yuvipanda, did I included enough info in this? It doesn't follow yet the new docs structure, but I can open another PR to move things around. What do you think?

yuvipanda:
Yeah, we can re-arrange this later! Thanks, @GeorgianaElena

Issue #277, opened 2021-02-26
Allow admins to rename user home directories


yuvipanda:
When a hub switches authentication methods, the users' get a new name. For example, with GitHub my username is `yuvipanda` while with Google it might be `yuvipanda@gmail.com`. Most of our infrastructure doesn't care about usernames, *except* home directories. So while my home directory with GitHub might be at `/export/homes/yuvipanda`, with Google it might be `/export/homes/yuvipanda@gmail.com`. So we need to support renaming users when authentication methods change.

This is a somewhat rare occasion, so we don't need to over-do it. This should also be as self-serve as possible.

My suggestion is that we mount *all* users' home directories at something like `~/shared/admin/homes` for admins, and then admins can do the renaming themselves. This shouldn't be on by default, since an accidental `rm -rf .` on an admin's home directory will wipe out everyone's work

GeorgianaElena:
> My suggestion is that we mount all users' home directories at something like ~/shared/admin/homes for admins, and then admins can do the renaming themselves. This shouldn't be on by default, since an accidental rm -rf . on an admin's home directory will wipe out everyone's work

So the workflow would be something like:
- hub admin decides to switch auth methods
- we switch auth methods for that hub
- we "move" user's home dirs to the new location own by the admins
- admins rename the home dirs

Some questions:
* After the renaming process finishes, do we leave this hub with the users home dirs at `~/shared/admin/homes` or do we move them back to `/export/homes/`? Do we still want/need to give admin access to the user's home dirs?
* If we do decide to mount them back, wouldn't be easier for us to have something like a `username` -> `new-username` map and do the renaming ourselves? (I have no idea if this is something that can be done automatically)
* What if a user logins with using the new auth method before the admin had the chance to  rename their home dir?

yuvipanda:
Hey @GeorgianaElena! This looks good. One thing to think of is that admins might need to change user names incrementally, rather than in one go. If you are switching from Google to GitHub, admins might not know all the GitHub usernames immediately.

So I think we should just mount all user homes to something like `~/shared-readwrite/user-homes`, so they can rename as they need whenever they want. So then the sequence becomes:

1. Mount user home directory under `~/shared-readwrite/user-homes` for *all* admins
2. When auth methods are switched, do the switch + do renames for admins
3. After that, admins have to do renames for their users.

This also enables a few other use cases:

1. Admins can poke around users' home directories to help with other things as well. 
[...]

PR #300, opened 2021-03-12
[WIP] Allow admins to rename user home directories

GeorgianaElena:
I'm not sure at all about this and I would be super grateful if got some help understanding this.

What *I think* the pilot-hubs persistent storage setup looks like:
- there is a nfs server for persistent storage
    - it's running at: `serverIP: nfs-server-01`
- nfs baseShareName is the location on the nfs server where all the home directories are
    - `baseShareName: /export/home-01/homes`
    - each "release" (staging, dask-staging, etc) gets their own volume (`staging-home-nfs`, `dask-staging-home-nfs` etc)
- `baseShareName` is also used to mount this nfs share inside the cluster (is it ???). We're mirroring the file hierarchy of the nfs server inside our cluster.

The way I understood the issue is that we want another "pointer" in our cluster to the location in the nfs server where the home dirs actually are. This pointer has to be somewhere where all the admins have rw access.

What this PR tries to do is to mount the nfs share (where the home dirs are) to a new location inside the cluster (`clusterMountPath`).

P.S. I read the https://docs.datahub.berkeley.edu/en/latest/admins/storage.html which seemed similar with the setup we have here (it's just implemented differently)

Closes https://github.com/2i2c-org/pilot-hubs/issues/277

### DIFF
--- a/hub-templates/base-hub/templates/nfs-share-creater.yaml
+++ b/hub-templates/base-hub/templates/nfs-share-creater.yaml
@@ -27,14 +27,14 @@ spec:
           image: busybox
           env:
             - name: NFS_SHARE_NAME
-              value: "{{ .Values.nfsPVC.nfs.baseShareName }}/{{ .Release.Name }}"
+              value: "{{ .Values.nfsPVC.nfs.clusterMountPath }}/{{ .Release.Name }}"
           command:
             - /bin/sh
             - -c
             - "mkdir -p ${NFS_SHARE_NAME} && chown 1000:1000 ${NFS_SHARE_NAME}"
           volumeMounts:
             - name: home-base
-              mountPath: {{ .Values.nfsPVC.nfs.baseShareName | quote }}
+              mountPath: {{ .Values.nfsPVC.nfs.clusterMountPath | quote }}
       volumes:
         - name: home-base
           nfs:
--- a/hub-templates/base-hub/templates/nfs-share-creater.yaml
+++ b/hub-templates/base-hub/templates/nfs-share-creater.yaml
@@ -27,14 +27,14 @@ spec:
           image: busybox
           env:
             - name: NFS_SHARE_NAME
-              value: "{{ .Values.nfsPVC.nfs.clusterMountPath }}/{{ .Release.Name }}"
+              value: "{{ .Values.nfsPVC.nfs.baseShareName }}/{{ .Release.Name }}"
           command:
             - /bin/sh
             - -c
             - "mkdir -p ${NFS_SHARE_NAME} && chown 1000:1000 ${NFS_SHARE_NAME}"
           volumeMounts:
             - name: home-base
-              mountPath: {{ .Values.nfsPVC.nfs.clusterMountPath | quote }}
+              mountPath: {{ .Values.nfsPVC.nfs.baseShareName | quote }}
       volumes:
         - name: home-base
           nfs:
--- a/hub-templates/base-hub/values.yaml
+++ b/hub-templates/base-hub/values.yaml
@@ -19,6 +19,7 @@ nfsPVC:
   nfs:
     serverIP: nfs-server-01
     baseShareName: /export/home-01/homes
+    clusterMountPath: /home/jovyan/shared-readwrite/user-homes
 
 jupyterhub:
   ingress:
@@ -81,8 +82,10 @@ jupyterhub:
   singleuser:
     admin:
       extraVolumeMounts:
-      - name: home
-        mountPath: /home/jovyan/shared-readwrite
+      - name: homes
+        mountPath: /home/jovyan/shared-readwrite/user-homes
+      - name: public
+        mountPath: /home/jovyan/public
         subPath: _shared
     startTimeout: 600 # 10 mins, because sometimes we have too many new nodes coming up together
     defaultUrl: /tree
@@ -103,8 +106,8 @@ jupyterhub:
         pvcName: home-nfs
         subPath: '{username}'
       extraVolumeMounts:
-      - name: home
-        mountPath: /home/jovyan/shared
+      - name: public
+        mountPath: /home/jovyan/public
         subPath: _shared
         readOnly: true
     memory:
--- a/hub-templates/base-hub/values.yaml
+++ b/hub-templates/base-hub/values.yaml
@@ -82,7 +82,7 @@ jupyterhub:
     admin:
       extraVolumeMounts:
       - name: home
-        mountPath: /home/jovyan/admin-readwrite/user-homes
+        mountPath: /home/jovyan/admin/user-homes
       - name: home
         mountPath: /home/jovyan/shared-readwrite
         subPath: _shared
--- a/hub-templates/base-hub/values.yaml
+++ b/hub-templates/base-hub/values.yaml
@@ -19,7 +19,6 @@ nfsPVC:
   nfs:
     serverIP: nfs-server-01
     baseShareName: /export/home-01/homes
-    clusterMountPath: /home/jovyan/shared-readwrite/user-homes
 
 jupyterhub:
   ingress:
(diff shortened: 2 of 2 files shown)


yuvipanda:
Thanks for opening this, @GeorgianaElena! We don't have many docs on our storage layer, so hopefully this will help.

## NFS Server setup

1. We have 1 NFS Server per cluster, often located at nfs-server-01. This is currently hand configured, but should change.
2. The NFS Server has a [persistent disk](https://cloud.google.com/persistent-disk) mounted specifically for the home directories of users. This makes it independent from rest of the VM, and can be grown / snapshotted differently.
3. This disk is mounted inside the NFS server at `/export/home-01`. You can check this out by sshing into the nfs server (` gcloud compute ssh nfs-server-01 --zone=us-central1-b`) and running `mount` or opening `/etc/fstab`.
4. This disk is then made available via NFS to be mounted by everything in the cluster, via this in `/etc/exports`: `/export/home-01 10.0.0.0/8(all_squash,anonuid=1000,anongid=1000,no_subtree_check,rw,sync)`. `/etc/exports` is an NFS thing, there sohuld be documentation around it in places.

## NFS Client setup

For reach hub, we want to:

1. **Hub directory creation**: Make sure there's a directory created under `/export/home-01` for that hub. So `/export/home-01` is `nfsPVC.nfs.baseShareName` - the base directory under which each hub has a directory.
2. **Hub user mounting**: Mount `/export/home-01/<hub-name>/<username>` for each user pod. This name is dynamically determined as a combination of `nfsPVC.nfs.baseShareName` and the current release name.

### Hub Directory Creation

For (1), we previously automated it with `gcloud` (#119) but now have a better solution with helm hooks (#255).

1. For each deployment, we create [this job](https://github.com/2i2c-org/pilot-hubs/blob/master/hub-templates/base-hub/templates/nfs-share-creater.yaml) via [helm hooks](https://helm.sh/docs/topics/charts_hooks/). It'll be created pre-deploy, run, and cleaned up before deployment proceeds.
2. This job will mount `nfsPVC.nfs.baseShareName`, and make sure the directory for the hub is present there with appropriate permissions. Then it exits.

Ideally, this would only happen once per hub setup - but we don't have a clear way to do that yet.

### Hub user mounting

Now comes the interesting part - actually giving a home directory to each user.

1. For each hub, we create a [PersistentVolumeClaim and PersistentVolume](https://github.com/2i2c-org/pilot-hubs/blob/c4986d52c1f83605683956492c2e3353122478ff/hub-templates/base-hub/templates/nfs-pvc.yaml#L1) that points to the hub directory created for it. This is the kubernetes *Volume* that refers to our actual storage. z2jh then mounts this PersistentVolumeClaim on the pod as a [volume named "home"](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/c86268b174e70e37fd68168450ea8737db3c8f61/jupyterhub/files/hub/jupyterhub_config.py#L277). So this `volume` points to `/exports/home-01/<hub-name>`
2. We can then *mount* (make visible) different parts of this volume in different places for our user. :
   1. [user home directory](https://github.com/2i2c-org/pilot-hubs/blob/master/hub-templates/base-hub/values.yaml#L100). Note the `subPath` - that refers to a *subdirectory* in the volume to mount at that given location. This is special syntax for z2jh - it will now mount into `/home/jovyan` (the mount path) the contents of the path `/exports/home-01/<hub-name>/<username>`, which is exactly what we want. This subpath is created if it doesn't already exist.
   2. The shared directory is mounted at `/home/jovyan/shared` for all users, and it is the contents of `/exports/home-01/<hub-name>/_shared`. The `_shared` is just convention for a username that won't be allowed in most cases (because of the leading underscore) - WHICH IS MOST LIKELY AN INVALID ASSUMPTION AND SHOULD BE CHANGED! not right now though. We use the `readOnly` option to make this particular mount readonly - users can't write to it.
   3. The shared directory is *again* mounted at `/home/jovyan/shared-readwrite` but [just for admins](https://github.com/2i2c-org/pilot-hubs/blob/master/hub-templates/base-hub/values.yaml#L105). This is a feature [we have](https://github.com/2i2c-org/pilot-hubs/blob/master/hub-templates/base-hub/values.yaml#L182) that lets us provide extra config for admin users only. This volumeMount is *not* readonly, so admins can write to it. 

To recap, we have:

1. 1 NFS Server (nfs-server-01)
2. With one disk mounted at `/exports/home-01`.
3. With one directory (share in NFS terminology for this use case) created at `/exports/home-01/<hub-name>` (via [helm hook](https://github.com/2i2c-org/pilot-hubs/blob/master/hub-templates/base-hub/templates/nfs-share-creater.yaml))
4. With one PersistentVolumeClaim (https://github.com/2i2c-org/pilot-hubs/blob/c4986d52c1f83605683956492c2e3353122478ff/hub-templates/base-hub/templates/nfs-pvc.yaml#L1) created for each hub, pointing to `/exports/home-01/<hub-name>`
5. Made available by z2jh as a [volume named "home"](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/c86268b174e70e37fd68168450ea8737db3c8f61/jupyterhub/files/hub/jupyterhub_config.py#L277) for each pod
6. With a `volumeMount` at `/home/jovyan` of the `subPath` of `<username>` in the  volume `home` - so the subdirectory `/exports/home-01/<hub-name>/<user-name>`. This is created if needed.
7. A *readOnly* volumeMount of the same volume, but with a subpath of `_shared` (so directory `/exports/home-01/<hub-name>/_shared`) for all users (https://github.com/2i2c-org/pilot-hubs/blob/c4986d52c1f83605683956492c2e3353122478ff/hub-templates/base-hub/values.yaml#L105)
8. A *readWrite* volumeMount of the same volume, with subpath of `_shared` (so directory `/exports/home-01/<hub-name/_shared`) [just for admins](https://github.com/2i2c-org/pilot-hubs/blob/c4986d52c1f83605683956492c2e3353122478ff/hub-templates/base-hub/values.yaml#L82). So we can *mount the same subdirectory in different places with different options*



For this task, what we want to do is mount `/exports/home-01/<hub-name>` (referenced by the volume `home` in the pod) to `/home/jovyan/shared/user-homes` (or similar paths) just for admins. We can do this similar to how we setup shared-readwrite, just without a subPath.

This is a lot, so I hope it makes sense! This is a lot of stuff! To have a good handle on how this works a little deeper, I'd suggest reading about: NFS, NFS Shares, Mounts and Bind Mounts.

GeorgianaElena:
Thanks A LOT @yuvipanda for making this clear :heart: 

I have mounted the home dirs to `/home/jovyan/admin-readwrite/user-homes` to make it clear that is only visible and modifiable by admins and it's not shared anywhere else. Is this ok?

yuvipanda:
Thanks a lot, @GeorgianaElena! This looks good, but I want us to have a consistent naming scheme for shared stuff before we merge this. See https://github.com/2i2c-org/pilot-hubs/issues/262/

So maybe we can make 'admin' be something that's visible to just admins, and is always readwrite. So we could make this into `admin/user-homes`? And later, we can move `shared-readwrite` to `admin/user-shared` or similar. 

How does that sound?

GeorgianaElena:
> So maybe we can make 'admin' be something that's visible to just admins, and is always readwrite. So we could make this into admin/user-homes? And later (if we so choose), we can move shared-readwrite to admin/user-shared or similar.

I really like this!

choldgraf:
Should we consider this PR blocked until https://github.com/2i2c-org/pilot-hubs/issues/262 is resolved?

yuvipanda:
@choldgraf yep! These conventions are going to stick for a while, so let's get it right. There's no current pressing need.

I'll mark this as a WIP

choldgraf:
Gotcha - I've removed this from our active board, added some tasks to #277, and added that to our projects board 👍

GeorgianaElena:
I will close this PR since it's been opened for a long time and if we decide to pick this up again I believe we should open a new one.
The discussion here (that has turned into docs) is the truly valuable part!

PR #304, opened 2021-03-15
Restructure docs to match the new structure

GeorgianaElena:
This doesn't add any new content or modifies existing one, it just moves things around to match the [new docs structure](https://github.com/2i2c-org/pilot-hubs/pull/293).


choldgraf:
IMO just go for it and we can spot-check improvements to the structure in future PRs. This PR will get out of date immediately if we leave it open for very long :-)

yuvipanda:
YAY! Let's try this out and modify as we go along.

PR #305, opened 2021-03-16
Document NFS storage layer

GeorgianaElena:
Inspired from https://github.com/2i2c-org/pilot-hubs/pull/300#issuecomment-797450560 <3


choldgraf:
This looks great and is pretty understandable even for somebody who understands none of this (me!) :-)

That being said I should probably defer to somebody else's 👀 in giving a review haha

yuvipanda:
THIS IS PRETTY COOL

PR #310, opened 2021-03-17
Document manual deploy

GeorgianaElena:
Closes https://github.com/2i2c-org/pilot-hubs/issues/113
Related issue https://github.com/2i2c-org/pilot-hubs/issues/116

### DIFF
--- a/docs/howto/operate/manual-deploy.md
+++ b/docs/howto/operate/manual-deploy.md
@@ -1,4 +1,4 @@
-# Manual deploy a hub
+# Manually deploy a hub
 
 The [`deploy.py`](https://github.com/2i2c-org/pilot-hubs/blob/master/deploy.py) script is a cmd utility built to help deploy one or all the hubs to their Kubernetes clusters, but  also to build the user image. Checkout the [](../configure/update-env.md) section on how to use the deploy script to update the user environment.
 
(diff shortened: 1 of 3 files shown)


docs/howto/operate/manual-deploy.md
```diff
@@ -0,0 +1,73 @@
...
+    ```bash
+    python3 deploy.py deploy --skip-hub-health-test
+    ```
+
+### Pre-requisites
+
+Before using any of these `deploy.py` commands, you first need to make sure you are authenticated in all the right places and have the right credentials exported as environmental variables. This is because the deploy script need to be able to authenticate into the given cluster and have access to the unique `proxy.secretToken` needed to deploy the proxy pod. It also needs to know how to connect to the Auth0 account in order to determine and configure each hub's auth provider.
+
+```{note}
+To enable the deploy script to authenticate into the cluster, checkout the [Fetch Cluster credentials](cmd-access.md) section.
```
yuvipanda, line 56: You only need to be authenticated to the two-eye-two-see project, since we have direct credentials for everything else under `secrets/`. They're encrypted with [sops](https://github.com/mozilla/sops), and the key is in the two-eye-two-see project.

yuvipanda:
Minor change, but otherwise great!

yuvipanda:
Thanks, @GeorgianaElena!

PR #311, opened 2021-03-17
Add 2i2c staff as admins to all hubs

GeorgianaElena:
Closes https://github.com/2i2c-org/pilot-hubs/issues/49

P.S I haven't modified all the hubs and I'm probably missing some staff members, because I first wanted to know what you think about having  `extra_allowed_users` and `extra_admins` under `auth0` config :eyes:


hub.py
```diff
@@ -240,8 +240,37 @@ def get_generated_config(self, auth_provider, proxy_secret_key):
...
+            # Add 2i2c staff to the list of admins and allowed users
+            admin_users, allowed_users = self.get_hub_allowed_users()
+            generated_config['jupyterhub']['hub']['config']['Authenticator'] = {
+                'admin_users': admin_users,
+                'allowed_users': allowed_users
+            }
+
         return self.apply_hub_template_fixes(generated_config, proxy_secret_key)
 
+    def get_hub_allowed_users(self):
```
yuvipanda, line 252: Can you put this in `hubs.yaml`? `clusters` is a top level variable, this could be another top level.

hub.py
```diff
@@ -240,8 +240,37 @@ def get_generated_config(self, auth_provider, proxy_secret_key):
...
+            'choldgraf@gmail.com',
+            'georgiana.dolocan@gmail.com'
+        ]
+        staff_github_ids = [
+            'yuvipanda',
+            'choldgraf',
+            'GeorgianaElena'
+        ]
+
+        extra_admins = self.spec['auth0']['extra_admins'] if self.spec['auth0'].get('extra_admins', False) else []
```
yuvipanda, line 266: Hmm, so we can't 'merge' this list with what's already in config, right? Is there any other way to manage this without inventing our own schema?

hub.py
GeorgianaElena: The `staff_emails` and `staff_github_ids`, rigth?

hub.py
GeorgianaElena: Unfortunately I don't think so :( Or I just couldn't find a different working approach

GeorgianaElena:
@yuvipanda, I've just rebased this. Do you think this is good to go? Is there anything I can do to make it better?

hubs.yaml
```diff
@@ -15,6 +26,12 @@ clusters:
         template: base-hub
         auth0:
           connection: github
+          extra_allowed_users: &paleohack_users
```
choldgraf, line 29: I am not sure exactly what `@paleohack_users` means, but this makes me wonder if we could just keep calling the key `allowed_users`, and then just define an explicit group for 2i2c staff that we manually add to each. Could we do something like:

```yaml
auth0:
    allowed_users: &paleohack_users &2i2c_staff
        - user1
        - user2
```

and so on?

(apologies if this question makes no sense)

choldgraf:
Is there anything else to finish up here? Or can we merge? Maybe @yuvipanda can give a quick approval?

choldgraf:
Another ping on this one @GeorgianaElena / @yuvipanda - or is this no longer relevant to merge?

yuvipanda:
Sorry for the late response, @GeorgianaElena. I'll mark this as WIP and rebase once #352 lands, since I think we can put the list of admins in the centralized config we're using for secrets there. Does that sound ok?

yuvipanda:
Here's a different pattern I suggest we employ:

1. Put list of staff admins in `https://github.com/2i2c-org/pilot-hubs/blob/36e83a5c3c289750a75fb240455c3e8eb1bf9e98/hub-templates/base-hub/values.yaml`, say under `staff.names.google`, `staff.names.github`, etc
2. Add a config item [here](https://github.com/2i2c-org/pilot-hubs/blob/36e83a5c3c289750a75fb240455c3e8eb1bf9e98/hub-templates/base-hub/values.yaml#L256) that will get the list of staff names for the currently selected authenticator, and add them to `c.Authenticator.admin_users`

This lets us keep the code in hub.py simple, and uses z2jh specific (rather than 2i2c specific) mechanisms for doing this.

GeorgianaElena:
Sure! I will give this a try @yuvipanda! Thanks <3

hub-templates/basehub/values.yaml
```diff
@@ -318,6 +329,8 @@ jupyterhub:
             return resp
 
         c.JupyterHub.authenticator_class = CustomOAuthenticator
+        # TODO: find selected authenticator
+        c.CustomOAuthenticator.admin_users = get_config('staff.names.google')
```
damianavila, line 333: This is looking nice, but wondering how this will play with the `allowed_users` stuff...

choldgraf:
@GeorgianaElena do you think that it would work if we just loaded the list of admin users from the YAML config in the `__main__.py` file, and then appended them to the list of admin users around here:

https://github.com/2i2c-org/pilot-hubs/blob/36e83a5c3c289750a75fb240455c3e8eb1bf9e98/deploy/__main__.py#L76-L83

Then I think we wouldn't need to write the custom authenticator stuff? But maybe I am not understanding this properly.

GeorgianaElena:
@choldgraf, it will probably work (this PR did something similar some time ago: https://github.com/2i2c-org/pilot-hubs/commit/9fa027c7237fbdc247e54640fb64ceff598e06ed).

This PR is blocked because I don't know how to achieve this: 
> This lets us keep the code in hub.py simple, and uses z2jh specific (rather than 2i2c specific) mechanisms for doing this. What do you think, @GeorgianaElena?

I don't know how to achieve the TODO comment I left here https://github.com/2i2c-org/pilot-hubs/blob/411df13b887cf0c14abde455d5a48ba1b62d16d6/hub-templates/basehub/values.yaml#L332

Givien, [this configuration merging structure](https://pilot-hubs.2i2c.org/en/latest/topic/hub-templates.html) + dealing with allowed_user and username_pattern https://github.com/2i2c-org/pilot-hubs/blob/380a8bf3cfc8fbf1d3f5791cfebe0bc3a745227f/config/hubs/cloudbank.cluster.yaml#L387

:(

choldgraf:
hmmm - OK thanks for this context. I think this is something that we should put on our next sprint, and make some time for intentional discussion so that we can figure out how to move past it.

My feeling is that we if we can't figure out how to do this the "right way" (using JupyterHub vs. 2i2c-specific config) within a couple of days, then we should just add this code into our deploy code as you had done before. We can always improve things with refactoring in the future!

GeorgianaElena:
This should be ready for review again. But the implementation might need to change potentially entirely depending on how we're going to solve the OAuthenticator issue.

hub-templates/basehub/values.yaml
```diff
@@ -298,8 +298,24 @@ jupyterhub:
         c.JupyterHub.spawner_class = CustomSpawner
 
       06-custom-authenticator: |
+        from z2jh import get_config
         from oauthenticator.auth0 import Auth0OAuthenticator
 
+        staff_names_google = [
+          "yuvipanda@gmail.com",
+          "choldgraf@gmail.com",
+          "georgiana.dolocan@gmail.com",
```
sgibson91, line 307: Is there a reason we are using personal emails here and not our 2i2c.org addresses? They should work because they're powered by Google, yes? Or am I missing something?

choldgraf (approved):
This seems pretty good to me! Just a couple of comments in there.

edit: oops that was supposed to be "comment" rather than "approve" 🙃

GeorgianaElena:
Thanks everyone for their reviews! I will mark this PR as draft again, until we remove support for ORCID as suggested [here](https://github.com/2i2c-org/pilot-hubs/pull/732#pullrequestreview-771330665).

I will tackle the orcid removal first, then come back to this and potentially move the staff list logic back in `hub.py`

GeorgianaElena:
This should be ready for review again, but FYI, the implementation changed entirely again :sweat_smile:

 Now that `CustomAuthenticator` is gone and the other solution I had identified implied listing the extra admin users under a different key than [`Authententicator.allowed_users`](https://github.com/2i2c-org/pilot-hubs/pull/311#discussion_r598330247). I decided to go with @choldgraf's idea [here](https://github.com/2i2c-org/pilot-hubs/pull/667) which seemed more intuitive. LMK what you think.

GeorgianaElena:
Hmm, not sure if aliases will still work...

choldgraf:
> since our names do not make any sense if someone wants to replicate the stuff, right?

That's a good point - I like the idea that the configuration of a given hub is roughly copy/pasteable with minimal changes into a users's own jupyterhub config.

config/hubs/staff.yaml
```diff
@@ -0,0 +1,15 @@
+staff:
+  githubIds:
+    - choldgraf
+    - consideRatio
+    - damianavila
+    - GeorgianaElena
+    - sgibson91
+    - yuvipanda
+  googleIds:
```
choldgraf, line 9:
```suggestion
  staff_github_ids:
    - choldgraf
    - consideRatio
    - damianavila
    - GeorgianaElena
    - sgibson91
    - yuvipanda
  staff_google_ids:
```

deployer/utils.py
```diff
@@ -48,29 +48,35 @@ def decrypt_file(encrypted_path):
...
+
+    staff_list = []
+
+    if "staff_google_ids" in user_list:
+        user_list = user_list.remove("staff_google_ids")
+        staff_list =  staff["googleIds"]
+    elif "staff_github_ids" in user_list:
+        user_list = user_list.remove("staff_github_ids")
+        staff_list =  staff["githubIds"]
+
```
choldgraf, line 66:
```suggestion
    for staff_list_type, staff_ids in staff.items():
        if staff_list_type in user_list:
            user_list = user_list.remove(staff_list_type)
            staff_list = staff_ids
            break
```

choldgraf: What do you think about re-using the keys that are in our `staff.yaml` file, so that if we add more there it will automatically be checked here?

config/hubs/2i2c.cluster.yaml
```diff
@@ -48,10 +48,10 @@ hubs:
           config:
             Authenticator:
               allowed_users: &staging_users
-                - yuvipanda@gmail.com
+                # this will be replaced with the actual list of
+                # staff google ids by the deployer script
+                - staff_google_ids
```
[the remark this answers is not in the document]
choldgraf: This seems reasonable to me - I like the idea that it's explicit. Could we use some special character that is still valid YAML that would make it clear that this will be replaced? Like could we do `{staff_google_ids}` or something? (or would that break YAML somehow?)

config/hubs/schema.yaml
```diff
@@ -119,6 +119,24 @@ properties:
...
+      List of different login providers staff ids
+    properties:
+      githubIds:
+        type: array
+        description: |
+            List of staff GitHub handles
+        items:
+          type: string
+      googleOauthIds:
+        type: array
```
choldgraf, line 135:
```suggestion
      github_ids:
        type: array
        description: |
            List of staff GitHub handles
        items:
          type: string
      google_ids:
        type: array
```

to match the new `staff.yaml` keys right?

config/hubs/2i2c.cluster.yaml
```diff
@@ -48,10 +48,10 @@ hubs:
           config:
             Authenticator:
               allowed_users: &staging_users
-                - yuvipanda@gmail.com
+                # this will be replaced with the actual list of
+                # staff google ids by the deployer script
+                - <staff_google_ids>
```
GeorgianaElena, line 53: @choldgraf, what do you think about this? (I kept the `staff` prefix here, otherwise `google_ids` would've lost its explicitly )

deployer/utils.py
```diff
@@ -46,4 +46,33 @@ def decrypt_file(encrypted_path):
...
+def add_staff(staff, user_list):
+    """
+    Add the staff ids to the existing user list
+    """
+    if isinstance(user_list, str):
+        user_list = [user_list]
+
+    custom_users = user_list[:]
+    for staff_list_type, staff_ids in staff.items():
+        staff_placeholder = "<staff_" + staff_list_type + ">"
```
GeorgianaElena, line 60: @choldgraf, I'm adding the `staff` prefix here

GeorgianaElena:
@choldgraf, my GitHub handle is wrong in that sheet :grimacing: Also, the order is different than what we thought, right? :laughing: 

> And then wget it and add the appropriate staff users as a part of the hub deployment. One benefit of this is that there would be a single list for the 2i2c engineering team and their relevant github/gmail/etc handles.

Hmm, there's should be another list in the repo that powers our website, right? :sweat_smile: 

> I can see pros/cons though, so feel free to tell me that this is making it way too complicated :-)

This isn't complicated, we would just need to parse the csv to get relevant info. However, I'm not very convinced that this is better than the current approach. One thing that I fear can happen, is someone changing a column name or order in that file that would cause things to fail :shrug: Also, I kind of like things being on GitHub as I feel they are more visible.

I don't have a strong opinion though. Curious what others think.

choldgraf:
In the meeting today we agreed that we should just go with @GeorgianaElena's suggestion of embedding this list of usernames etc in the `config.yml` file instead of using a separate google sheet. I've deleted that google sheet (and sorry @GeorgianaElena for getting your username wrong 😬)

@GeorgianaElena is this one now ready to go?

deployer/utils.py
```diff
@@ -46,4 +46,33 @@ def decrypt_file(encrypted_path):
...
+    """Prepare a hub's configuration file for deployment."""
+    # Load the staff config file
+    with open('config/hubs/staff.yaml') as f:
+        staff = yaml.load(f)
+
+    authenticator = config.get("jupyterhub", {}).get("hub", {}).get("config", {}).get("Authenticator", {})
+
+    # `Allowed_users` list doesn't exist for hubs where everyone is allowed to login
+    if authenticator.get("allowed_users", None):
+        authenticator["allowed_users"] = add_staff(staff["staff"], authenticator["allowed_users"])
```
damianavila, line 76: If "allowed_users" does not exist, then `authenticator["allowed_users"]` would be `None` as per line 75, right?
How the `add_staff` function will handle that `None`?

choldgraf: I think that if `allowed_users` is `None`, then the if-statement won't be triggered and it will just be skipped

damianavila: > Feel free to say that my thoughts do not make any sense 😜 .

It is 2 am now so that would make sense 😉 💤

choldgraf (approved):
This looks good to me! We can improve things iteratively in follow-up PRs, so my vote is that we merge this unless somebody has a strong objection, and then see if this works to provide all of us admin access

PR #329, opened 2021-03-29
Update Catalyst hub image

zaneselvans:
Changed the name of the image (new repo on DockerHub) because we're using the same image for other PUDL Jupyter things now too.  Updated the version tag.

### DIFF
--- a/hubs.yaml
+++ b/hubs.yaml
@@ -355,8 +355,8 @@ clusters:
             jupyterhub:
               singleuser:
                 image:
-                  name: catalystcoop/pilot-hub
-                  tag: 2021.02.23
+                  name: catalystcoop/pudl-jupyter
+                  tag: 2021.03.27
                 memory:
                   limit: 6G
                   guarantee: 4G


yuvipanda:
Hey @zaneselvans! We've *just* built a GUI that lets you configure the image yourself, without needing to wait on us. The docs aren't merged yet, but are at https://github.com/2i2c-org/pilot/pull/68/files#diff-aaa6369658b00d12f7edcc005ad3146066f5f7ddc588eeaf71df6a9709c2007fR1. Can you try it out and let us know if that works?

choldgraf:
Specifically [here's a demo of the instructions](https://2i2c-pilot-documentation--68.org.readthedocs.build/en/68/admin/howto/configurator.html) - indeed feedback would be most welcome :-)

zaneselvans:
Hokay here we go...

* I logged in and went to the Admin > Service > configurator menu as directed.
* In the configurator, I see a "User docker image" text field and a "Default user interface" radio button.
* But I don't want to update the "user" docker image, I want to update the docker image for my entire JupyterHub, so for now I'm leaving this field blank, and wondering how to set the "default" image so that everyone on the Hub gets it.
* Also, there's no indication as to what format the entry in the "User docker image" field should take. An example or two would be good (should this be `catalystcoop/pudl-jupyter` or what would it look like on another container registry?)
* However I *do* want to set my personal default interface to JupyterLab, so I select that radio button.
* Then I click "Submit" and nothing happens.

yuvipanda:
That is very valuble feedback, @zaneselvans! I'll look through it today. I've merged this for now to not block you :)

zaneselvans:
Don't know if it matters, and maybe you have it in the logs anyway, but I was using Firefox 87.0 on Ubuntu 20.10. and connecting via the MozillaVPN from a server in Denver, CO, USA.

PR #334, opened 2021-03-31
Fix actions

GeorgianaElena:
Unfortunately we cannot use other actions inside our composite action (ref: https://github.com/actions/runner/issues/646)
So I've moved all of them back to the workflows. We're creating some duplicate code with this, but it's still better than not having the composite action at all.

### Note:
I couldn't expand the values in `secrets` as much as I tried :( (ref: https://github.com/GeorgianaElena/pilot-hubs/runs/2237985671?check_suite_focus=true) so instead I'm passing them over as inputs from the workflow and it doesn't seem to expose them in the logs (https://github.com/GeorgianaElena/pilot-hubs/runs/2238407062?check_suite_focus=true). The alternative would be to export them from the workflows.

### DIFF
--- a/.github/actions/deploy/action.yml
+++ b/.github/actions/deploy/action.yml
@@ -14,16 +14,17 @@ runs:
         helm dep up base-hub
         helm dep up daskhub
         helm dep up ephemeral-hub
+      shell: bash
     - name: Setup gcloud auth for docker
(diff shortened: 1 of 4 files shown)


GeorgianaElena:
Yaay! They worked :D

Issue #207, opened 2021-02-09
Hub for Univ. Grenoble satellite team


choldgraf:
We need to deploy a Pangeo-style hub for the University of Grenoble. Here's a rough list of "extra" packages that they listed in their questionnaire:

```
cartopy, dask, distributed, gsw, intake, mamba, matplotlib, numpy, scipy, xarray, xesmf, xgcm, xhistogram,
xmitgcm, xrft, xscale, zarr
```

They are also going to have like 10-20TB of data stored in Wasabi hot storage somewhere (@rabernat maybe you have context on this?)

# Hub Info

* **Hub Type**: Pangeo
* **Authentication Type**: GitHub
* **Administrators**:
  * roxyboy
  * lesommer
  * auraoupa
* **Timeline**: ASAP
* **Link to config**: TODO
* **Link to URL**: https://grenoble-swot.pilot.2i2c.cloud

choldgraf:
I've added some metadata about this hub after hearing back from the Grenoble team. I wonder what @yuvipanda or @GeorgianaElena thinks about using that `hub info` section as a GitHub Issue template for when we add new hubs, so that we can keep track of what we are running (either in this repo, or in a `hubs/` repo or something?).

Note that the "link to config" section also reminds me of a question @yuvipanda had a few weeks back, about whether hub configs should be in a sub-directory rather than in a single YAML file. One benefit of this would be that we can more easily create permalinks to the config...

choldgraf:
Is the process for creating this hub as simple as just adding a new entry to the hubs.yaml file? If so I can try this tomorrow and send to them

yuvipanda:
Adding more context - @roxyboy and others have gotten a GCP project + billing account where we're moving the hub to (with #237). This will let them scale up more than they can now.

roxyboy:
Currently the Jupyterlab option (which was there during the pilot version) seems to be gone with only the notebook format available but could we add the lab option?

yuvipanda:
@roxyboy we default to opening in JupyterLab now - can you give it a shot?

roxyboy:
> @roxyboy we default to opening in JupyterLab now - can you give it a shot?

Yes, it is now JupyterLab :)

yuvipanda:
@roxyboy yay, ok. You have a 1GB memory limit now, and dask-gateway limits aren't set by default. What would you like these to be?

roxyboy:
[...]

PR #340, opened 2021-04-07
Add docs on setting up a new project

yuvipanda:
Ref https://github.com/2i2c-org/pilot-hubs/issues/335

### DIFF
--- a/docs/howto/operate/setup-new-project.md
+++ b/docs/howto/operate/setup-new-project.md
@@ -23,7 +23,7 @@ account and give us access.
 
 Once the client organization provides us with a billing account,
 we provision a project for them inside the 2i2c.org GCP organization.
-ALl this work happens in the [org-ops](https://github.com/2i2c-org/org-ops)
+All this work happens in the [org-ops](https://github.com/2i2c-org/org-ops)
 repository.
 
 1. Edit `projects.tfvars`, adding an entry to `fully_managed_projects`.
@@ -33,7 +33,7 @@ repository.
    issue.
 
 2. Run terraform (with instructions in the org-ops repo README). This will give you a
-   detaile plan on what exactly terraform will do. **Scrutinize this carefully** -
+   detailed plan on what exactly terraform will do. **Scrutinize this carefully** -
    only a new project should be created, nothing else should change. Particularly,
    there should be no **delete** actions.
 
--- a/docs/howto/operate/setup-new-project.md
+++ b/docs/howto/operate/setup-new-project.md
@@ -33,15 +33,16 @@ repository.
    issue.
 
 2. Run terraform (with instructions in the org-ops repo README). This will give you a
-   detailed plan on what exactly terraform will do. **Scrutinize this carefully** -
-   only a new project should be created, nothing else should change. Particularly,
-   there should be no **delete** actions.
+   detailed plan on what exactly terraform will do. [prevent_destroy](https://www.terraform.io/docs/language/meta-arguments/lifecycle.html#prevent_destroy)
+   is set to most sensitive resources to prevent accidental destruction - however,
+   you should still scrutinize the plan produced by `terraform apply` carefully before applying it.
+   Particularly, there shouldn't be any *delete* actions with this.
 
 3. Validate that the project has been created, and you have access to it with
    your user account. 
 
 4. Commit the change, make a PR to the repo with it, and merge it. This is
-   preferable to pushing to the repo directly. You can self-merge your PR.
+   preferable to pushing to the repo directly. You can self-merge it.
 
 ## Without billing account access
 
--- a/docs/howto/operate/setup-new-project.md
+++ b/docs/howto/operate/setup-new-project.md
@@ -32,17 +32,19 @@ repository.
    information about the client organization, ideally pointing to a GitHub
    issue.
 
-2. Run terraform (with instructions in the org-ops repo README). This will give you a
-   detailed plan on what exactly terraform will do. [prevent_destroy](https://www.terraform.io/docs/language/meta-arguments/lifecycle.html#prevent_destroy)
-   is set to most sensitive resources to prevent accidental destruction - however,
-   you should still scrutinize the plan produced by `terraform apply` carefully before applying it.
-   Particularly, there shouldn't be any *delete* actions with this.
+2. Make a commit, and create a pull request in the org-ops repository.
 
-3. Validate that the project has been created, and you have access to it with
+3. Run `terraform plan -var-file project.tfvars` to give you a detailed plan on
+   what exactly terraform will do. Paste this in the PR description, and ask
+   someone else for review.
+
+4. Someone else should review the plan, and merge the PR. There should be co-ordination
+   on when this is merged, so immediately after either the PR creator or merger
+   can run `terraform apply -var-file projects.tfvars` after merging. They should
+   make sure the diff isn't that different, and comment on the PR once it's done.
+
+5. Validate that the project has been created, and you have access to it with
    your user account. 
-
-4. Commit the change, make a PR to the repo with it, and merge it. This is
-   preferable to pushing to the repo directly. You can self-merge it.
 
 ## Without billing account access
 
--- a/docs/howto/operate/setup-new-project.md
+++ b/docs/howto/operate/setup-new-project.md
@@ -1,0 +1,72 @@
+# Setup a new project for a client organization
+
+Each client organization of 2i2c gets their own [GCP Project](https://cloud.google.com/resource-manager/docs/creating-managing-projects).
+This allows us to:
+
+1. Maintain billing easily, since projects are the easiest unit of
+   billing. 
+2. Give members of client organizations (such as IT teams) full access to
+   the cloud project, without granting access to other client organizations'
+   resources
+3. Allow client organizations to take over running of the infrastructure
+   themselves, with minimal disruption.
+
+Client organizations can give us access to a billing account, or
+just to a project. This document describes what 2i2c engineers should
+do once client organizations give us access.
+
+## Client organization provides billing account
+
+[Our pilot docs](https://pilot.2i2c.org/en/latest/admin/howto/create-billing-account.html#full-billing-account-access)
+has information on how client organizations can provision a billing
+account and give us access.
+
+Once the client organization provides us with a billing account,
+we provision a project for them inside the 2i2c.org GCP organization.
+ALl this work happens in the [org-ops](https://github.com/2i2c-org/org-ops)
+repository.
+
+1. Edit `projects.tfvars`, adding an entry to `fully_managed_projects`.
+   The key should be the name of the project we want to create, and the
+   value should be the billing account ID. Add a comment referencing more
+   information about the client organization, ideally pointing to a GitHub
+   issue.
+
+2. Run terraform (with instructions in the org-ops repo README). This will give you a
+   detaile plan on what exactly terraform will do. **Scrutinize this carefully** -
+   only a new project should be created, nothing else should change. Particularly,
+   there should be no **delete** actions.
+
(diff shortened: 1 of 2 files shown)


yuvipanda:
We can test this out on #207

docs/howto/operate/setup-new-project.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Client organization provides billing account
+
+[Our pilot docs](https://pilot.2i2c.org/en/latest/admin/howto/create-billing-account.html#full-billing-account-access)
+has information on how client organizations can provision a billing
+account and give us access.
+
+Once the client organization provides us with a billing account,
+we provision a project for them inside the 2i2c.org GCP organization.
+ALl this work happens in the [org-ops](https://github.com/2i2c-org/org-ops)
```
damianavila, line 26: All typo

docs/howto/operate/setup-new-project.md
```diff
@@ -0,0 +1,72 @@
...
+repository.
+
+1. Edit `projects.tfvars`, adding an entry to `fully_managed_projects`.
+   The key should be the name of the project we want to create, and the
+   value should be the billing account ID. Add a comment referencing more
+   information about the client organization, ideally pointing to a GitHub
+   issue.
+
+2. Run terraform (with instructions in the org-ops repo README). This will give you a
+   detaile plan on what exactly terraform will do. **Scrutinize this carefully** -
```
damianavila, line 36: detailed

yuvipanda:
Thanks @damianavila! Fixed :)

docs/howto/operate/setup-new-project.md
```diff
@@ -0,0 +1,72 @@
...
+All this work happens in the [org-ops](https://github.com/2i2c-org/org-ops)
+repository.
+
+1. Edit `projects.tfvars`, adding an entry to `fully_managed_projects`.
+   The key should be the name of the project we want to create, and the
+   value should be the billing account ID. Add a comment referencing more
+   information about the client organization, ideally pointing to a GitHub
+   issue.
+
+2. Run terraform (with instructions in the org-ops repo README). This will give you a
```
GeorgianaElena, line 35:
```suggestion
2. Run terraform (with instructions in the org-ops repo [README](https://github.com/2i2c-org/org-ops/blob/main/README.md)). This will give you a
```

GeorgianaElena (approved):
:rocket: Thank you for writing these instructions @yuvipanda!

docs/howto/operate/setup-new-project.md
```diff
@@ -0,0 +1,72 @@
...
+2. Run terraform (with instructions in the org-ops repo [README](https://github.com/2i2c-org/org-ops/blob/main/README.md)). This will give you a
+   detailed plan on what exactly terraform will do. **Scrutinize this carefully** -
+   only a new project should be created, nothing else should change. Particularly,
+   there should be no **delete** actions.
+
+3. Validate that the project has been created, and you have access to it with
+   your user account. 
+
+4. Commit the change, make a PR to the repo with it, and merge it. This is
+   preferable to pushing to the repo directly. You can self-merge your PR.
```
damianavila, line 44: Are you sure you want to allow self-merge when the PR has the potential to destroy things? (the second point you wrote above)

yuvipanda: The danger is in the `terraform apply` step, which is why it doesn't run automatically. Merging doesn't actually trigger anything.

damianavila: OK, thanks for the clarification, so how do we make sure others can see/review that "state" before being applied?

yuvipanda: Great question! What kinda workflow do you think would work here? I don't want to run `terraform apply` in CI, so not sure how to approach review.

damianavila: I have not worked with the terraform stuff yet (although I have some experience with AWS Cloudformation), but looking quickly at the terraform docs, it seems you have a `terraform plan` command. That should surface any problem (or deletion you are trying to prevent), right? Then, you could run that command by the CI and detect any "bad" message. And others can review that plan.
Additionally, it seems you can run that command with a `-out` parameter and get some file that you can eventually run manually later with `terraform apply` command (modulo you still do not want to run `apply` within the CI which could be an event happening on the merge event after a successful CI run and approval from others, but I totally get it is **frightening** to give the CI such destructive power): https://www.terraform.io/docs/cli/commands/plan.html
Finally, I may be saying stupids things (as I did in the other PR) because I do not know yet the deep stuff, so be patient with me :wink:

yuvipanda: That sounds great! I think @sgibson91 also mentioned something like this elsewhere.

What do you think about the permission side of things? Right now, `terraform apply` (or even `terraform plan`, I think) requires a lot of access - to create and edit projects! What do you think are ways we can minimize risk there?

If I'm understanding your suggestion correctly, the process could be:

1. Whoever submits PR also runs `terraform plan` and posts output in PR
2. Someone else reviews it, and actually runs `terraform apply`, passing in the plan from (1).

Does that seem right?

sgibson91: You would want to be careful about posting the output of terraform apply anywhere public. Secrets won't be encrypted. Terraform developed Terraform Cloud specifically to share state files safely across teams.

yuvipanda: Ah, fun :| What do you think would be a good way to do review for this, @sgibson91?

yuvipanda: Another factor to consider here is that I don't think we'll be doing this a lot - only happens once per customer. There are also lifecycle guards that prevent deletions, so it isn't *that* scary. So maybe it's ok?

damianavila: > Whoever submits PR also runs terraform plan and posts output in PR
> Someone else reviews it, and actually runs terraform apply, passing in the plan from (1).
> Does that seem right?

Yes, except that the first person can also run the apply command after a positive review from the second one.
The important piece, I think, is having a second pair of eyes looking at the output before making the change.

> You would want to be careful about posting the output of terraform apply anywhere public.

Looking quickly into the docs, it seems there is a way to mask the sensitive info coming from the outputs: https://www.terraform.io/docs/language/values/variables.html#suppressing-values-in-cli-output
Again, just reading docs, so I might be missing some stuff you know...

> Another factor to consider here is that I don't think we'll be doing this a lot 

Yeah, maybe I am overcomplicating this stuff if this is just an "infrequent" event.
But I guess this will scale as more clients arise, right?

yuvipanda: > Yeah, maybe I am overcomplicating this stuff if this is just an "infrequent" event.
> But I guess this will scale as more clients arise, right?

I'm not sure how exactly the process would be, though. I think Terraform is a poor fit for 'many simple operations, repeated and automated' - a script or a web app maybe, once we hit scale. 

I've added:

1. More prevent_destroy attributes to our terraform resources. It shouldn't destroy anything now
2. Toned down the warnings here.

yuvipanda: Another option here is to switch to a python script that can parse output of terraform plan, and validate that it won't destroy anything. However, I think that's a bit overkill for now - prevent_destroy should do most of what we want.

sgibson91: > Looking quickly into the docs, it seems there is a way to mask the sensitive info coming from the outputs: https://www.terraform.io/docs/language/values/variables.html#suppressing-values-in-cli-output
> Again, just reading docs, so I might be missing some stuff you know...

I admit, I may have gotten confused with the state file :) https://www.terraform.io/docs/language/state/sensitive-data.html

yuvipanda: @sgibson91 just pointed to https://github.com/2i2c-org/pilot-hubs/, which might be useful here.

However, reading https://learn.hashicorp.com/tutorials/terraform/automate-terraform?in=terraform/automation, it tells me that it is 'advanced', and seems fairly complicated to do right.

sgibson91: Yeah, looks like. Do you have a staging deployment that could be used as a manual test? So an example review workflow might look like:

1. Checkout PR
2. Run script X that will mimic deployment Y but without dependencies Z (where Z could be users, anything that is "live")
3. Run `terraform plan` of the checked out PR for inspection
4. Run `terraform apply` of the checked out PR (it's then ok if things go wrong here because it's not a live deployment)
5. Then either: approve and merge PR, or report issues in PR body.
6. `terraform destroy` the staging deployment

This would also mean that states wouldn't be problematic. It's very manual though and possibly expensive in time and cost so I'm not sure it's the greatest idea 😕

yuvipanda: At least for this workflow, we're creating GCP projects (https://github.com/2i2c-org/org-ops/blob/072054aded99652a717b171fbb59ee502a6a9a16/main.tf#L12), so I don't know how we can have a staging deployment. Agree re: expensive in time and cost, though.

My inclination is to do the following for project creation (which is what this documentation is specifically for):

1. Make changes
2. Make a PR
3. Someone else approves PR
4. You or other person does `terraform apply`.

This seems like a reasonable balance without adding too much cost in terms of time to the workflow. What do you think?

docs/howto/operate/setup-new-project.md
```diff
@@ -0,0 +1,73 @@
...
+1. Edit `projects.tfvars`, adding an entry to `fully_managed_projects`.
+   The key should be the name of the project we want to create, and the
+   value should be the billing account ID. Add a comment referencing more
+   information about the client organization, ideally pointing to a GitHub
+   issue.
+
+2. Run terraform (with instructions in the org-ops repo README). This will give you a
+   detailed plan on what exactly terraform will do. [prevent_destroy](https://www.terraform.io/docs/language/meta-arguments/lifecycle.html#prevent_destroy)
+   is set to most sensitive resources to prevent accidental destruction - however,
+   you should still scrutinize the plan produced by `terraform apply` carefully before applying it.
```
sgibson91, line 38:
```suggestion
   you should still scrutinize the plan produced by `terraform plan` carefully before applying it.
```

Is this supposed to be `terraform plan` or do you mean before accepting the `Continue with apply?` prompt?

docs/howto/operate/setup-new-project.md
sgibson91: I'm a little bit confused now as step 3 is "Check the project has been created and you have access to it" and step 4 is "open the PR" which implies `terraform apply` was run before opening the PR?

sgibson91: Ah, my brain just caught up with me! Yes, I think we should approve/merge the PR before running `terraform apply` - always good to have a second pair of eyes on code/configs before actioning anything I think.

damianavila: > 3. Someone else approves PR

This is where `terrraform plan` (with `sensitive` on to prevent leaking info/secret) plays a role, IMHO.
Otherwise, the person reviewing the PR does not have any info about how the plan looks like.

yuvipanda: @damianavila @sgibson91 I updated the instructions based on the conversations here. LMK what you think? I wanna balance code review with appropriate load on our small team.

sgibson91: I think this looks good!

damianavila (approved):
LGTM, the only remaining thing would be to mark the secrets as `sensitive`... but looking at https://github.com/2i2c-org/org-ops/blob/main/projects.tfvars, you already have "visible" data there, so I think this is ready as-is.

yuvipanda:
Thanks, @damianavila! Wanna hit merge? :D

damianavila:
Done 🎉 😉

PR #341, opened 2021-04-07
Add deps

GeorgianaElena:
Ref: https://github.com/2i2c-org/pilot/issues/66

Todos left from the initial issue:
- nodejs with the ability to manage node versions
- gatsby

### DIFF
--- a/hub-templates/base-hub/values.yaml
+++ b/hub-templates/base-hub/values.yaml
@@ -96,7 +96,7 @@ jupyterhub:
                 values: [user-pool]
     image:
       name: set_automatically_by_automation
-      tag: 638350c
+      tag: 1b83c4f
     storage:
       type: static
       static:
--- a/images/user/Dockerfile
+++ b/images/user/Dockerfile
@@ -109,8 +109,6 @@ RUN apt-get update && \
 RUN apt-get update && \
 	apt-get install -y --no-install-recommends \
                    ffmpeg > /dev/null
-# gatsby: https://github.com/2i2c-org/pilot/issues/66
-RUN npm install -g gatsby-cli
 
 # R_LIBS_USER is set by default in /etc/R/Renviron, which RStudio loads.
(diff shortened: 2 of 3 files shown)


damianavila:
Just curious about this one, I noticed you are installing r packages from the r side...
And I guess most of the packages are already available from conda-forge.
Actually, in the Dockerfile, you are installing r-base and friends with conda.
So... are these install.R packages specific enough that they are not being packaged in conda-forge yet?
Additionally, did not you find any incompatibilities mixing conda R and native R packages?

yuvipanda:
@damianavila so r-base is installed from apt - specifically from https://cran.r-project.org/bin/linux/ubuntu/. The people who maintain that also maintain it in debian / ubuntu, so that's nice.

Then, we get R packages from packagemanager.rstudio.com/. This produces *binary packages*, so install times are much, much better than when installing from source. It supports everything in CRAN, and uses R native installation methods familiar to R users (`install.packages`), so that wins over conda-forge's packages.

So currently, we get:

1. Python from conda-forge
2. scientific python packages from conda-forge / pypi
3. other python packages from pypi
4. R packages from packagemanager.rstudio.com
5. R from https://cran.r-project.org/bin/linux/ubuntu/

I hope this helps?

GeorgianaElena:
All the R libraries installed here seem to be loading ok on staging, so merging this!

damianavila:
> @damianavila so r-base is installed from apt - specifically from https://cran.r-project.org/bin/linux/ubuntu/. The people who maintain that also maintain it in debian / ubuntu, so that's nice.

Ja... I read conda in the first lines and I supposed you were installing r-base with it, my bad.
I should read the whole thing before commenting :wink:.
Now that I looked twice it is clear you are getting this from apt, thanks for the clarification!

> R packages from packagemanager.rstudio.com

Interesting!

> I hope this helps?

Yep, thank you!

PR #343, opened 2021-04-14
Revert replace nodeSelector with nodeAffinity

GeorgianaElena:
Reverts #278

### DIFF
--- a/hub-templates/base-hub/values.yaml
+++ b/hub-templates/base-hub/values.yaml
@@ -86,14 +86,8 @@ jupyterhub:
           subPath: _shared
     startTimeout: 600 # 10 mins, because sometimes we have too many new nodes coming up together
     defaultUrl: /tree
-    extraNodeAffinity:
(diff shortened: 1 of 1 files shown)


damianavila:
@GeorgianaElena, I noticed this is not an exact revert of #278, there are still some changes in hub.py. Is that expected?

damianavila:
From the last comment in that PR, it seems to be the case, just double-checking...

GeorgianaElena:
> @GeorgianaElena, I noticed this is not an exact revert of #278, there are still some changes in hub.py. Is that expected?

@damianavila, yes! The changes in `hub.py` are related to the [hub health test](https://pilot-hubs.2i2c.org/en/latest/howto/operate/test-hub.html?highlight=health#the-hub-health-test). It's just an increase of the timeout we wait for the test user server to get up, before calling it a failure (the timeout error you probably noticed in the GitHub actions tests :( )

Issue #139, opened 2020-12-14
Turn our deployment scripts into a package


yuvipanda:
We have a smattering of `.py` files in this repo for deployment. They should be organized into a python package

PR #352, opened 2021-04-20
Refactor deployment script

yuvipanda:
- Move to its own package under deploy/ to reduce
  clutter in repo root
- Move tests inside deploy/ too
- Tests earlier had a recursive dependency - hub.py will call them,
  but they will import hub.py to know more about the hub they are
  testing! This didn't break because the tests were called via
  pytest.main. This did prevent the tests from being used elsewhere.
  The tests now are independent, and get just the info they need
  from commandline arguments
- Temporarily disable the custom home page template test, since I
  haven't quite figured out how to pass the information it
  needs
- Use encrypted YAML file for secret keys
    
    These secret keys were set in GitHub secrets and then
    provisioned manually to users who needed to deploy from
    their machine:
    
    - Auth0 Client ID & Client Secret, to talk to the Auth0
      API
    - PROXY_SECRET_TOKEN, which was used to derive not just the
      hub's proxy secret token but many other secrets
    
    This PR puts them in a deployment.config.yaml file,
    encrypted with sops and checked-in to the repo.
    Document per-hub config file schema
    
 -  Adds a JSON Schema (written in YAML) to document what the
    fields in per-hub config file do. They are validated strictly
    (no additional properties allowed) on each deploy, to prevent
    addition of fields that aren't documented.
    
    Some of the documentation made redundant by the schema have
    been removed.

Fixes https://github.com/2i2c-org/pilot-hubs/issues/139


deploy/__main__.py
```diff
@@ -109,6 +110,10 @@ def main():
         build()
     elif args.action == "deploy":
         deploy(args.cluster_name, args.hub_name, args.skip_hub_health_test)
+    else:
+        # FIXME: Not sure how to achieve this with just argparse?
+        print(argparser.format_help(), file=sys.stderr)
```
GeorgianaElena, line 115: Is this to print a help message when no arguments are passed to the script?

I noticed that when you pass it anything other than `build/deploy`, it prints a `usage` help message. But when you run it with no arguments, it  just exits and doesn't print anything.

deploy/hub.py
```diff
@@ -385,7 +317,8 @@ def deploy(self, auth_provider, proxy_secret_key, skip_hub_health_test=False):
             generated_values_file.flush()
 
             cmd = [
-                'helm', 'upgrade', '--install', '--create-namespace', '--wait',
+                'helm', 'upgrade', '--install', '--create-namespace',
+                # '--wait',
```
GeorgianaElena, line 321: I think this was to help with testing and can be un-commented, right?

deploy/tests/test_hub_health.py
```diff
@@ -0,0 +1,90 @@
...
+
+def failure_handler(details):
+    print(f"Hub check health validation failed, hub not healthy.")
+
+def success_handler(details):
+    print(f"Notebook execution finished successfully.")
+
+
+
+# Try 2 times before declaring it a failure
```
GeorgianaElena, line 20: We're only trying on time now. Sorry for failing to update/remove the comment when introducing this change :(

deploy/tests/test_hub_health.py
```diff
@@ -0,0 +1,90 @@
...
+    max_tries=1
+)
+async def check_hub_health(hub_url, test_notebook_path, service_api_token):
+    """
+    After each hub gets deployed, validate that it 'works'.
+
+    Automatically create a temporary user, start their server and run a test notebook, making
+    sure it runs to completion. Stop and delete the test server at the end. Try this steps twice
+    before declaring it a failure. If any of these steps fails, immediately halt further
+    deployments and error out.
```
GeorgianaElena, line 36: Same here:
```suggestion
    Automatically create a temporary user, start their server and run a test notebook, making
    sure it runs to completion. Stop and delete the test server at the end. If any of these steps fails, immediately halt further
    deployments and error out.
```

tests/test_custom_templates.py
```diff
@@ -1,31 +0,0 @@
-import pytest
-import requests
-from bs4 import BeautifulSoup
-
-def test_templates_loaded(hub):
```
GeorgianaElena, line 5: I think that for this to be refactored too and run under the current testing structure, it needs the list of urls where a hub runs (the only one with multiple domains should be just mills at this moment), the url where the logo can be found and the hub template (hub-type that we're already passing).

GeorgianaElena (commented):
Moving the tests to their own module it's a great idea and reduces the confusion a bit.
I left a few comments with a couple of comments suggestions and a question :)

yuvipanda:
@GeorgianaElena thanks! I've addressed most of your comments :) But also did some more refactoring...

GeorgianaElena (approved):
Encrypting the secret keys is just "yaaaay!" \o/

I think there are some docs pieces that need to be changed to match the current deployment structure, but it can definitely go into another PR and don't block this one.

GeorgianaElena:
Thanks @yuvipanda :tada:

yuvipanda:
There's now several references to `hubs.yaml` that will need to be rewritten in the docs.

choldgraf:
@yuvipanda - I've added a "tasks" section to #139 and added an item to document the changes, also updated the top comment of this so that it no longer resolves the issue on its own (unless you want to add documentation changes to this PR as well?)

damianavila:
I have one general comment for this PR... you are thinking here into splitting the big hubs.yaml file into 3 yaml files representing the 3 clusters? What do you think about having more granularity with **each** hub being represented by an independent yaml file?
So you would **keep** these 3 now shorter cluster files and add a dozen of hub yaml files in some "ordered" directory layout... 

clusters/
  foo.yaml
  foo/
    hub1.yaml
    hub2.yaml
  bar.yaml
  bar/
    hub3.yaml
  baz/

I think that will make it easy to isolate each hub caractheristics to deal with them on a daily basis and/or show them in docs, for instance.

Thoughts?
PS: If you already discussed this, feel free to redirect me to that previous conversation).

yuvipanda:
ok, I've cleaned up the docs, removed references to `hubs.yaml` - this is ready now.

@damianavila one file per hub is absolutely the way to go! Right now, we rely on the fact that 'staging' is the first hub in the list, and that any failing deploy will stop subsequent deploys, to make sure that we 'test' on staging first. Need to figure out some mechanism for doing that when there's no natural ordering - as will be the case with files. I don't want to end up with 00-staging.yaml, etc though. Some prior discussion in https://github.com/2i2c-org/pilot-hubs/issues/124 but not a lot.

yuvipanda:
@choldgraf I've added docs too in this PR!

damianavila:
> one file per hub is absolutely the way to go! 

Glad we agree :wink: and thanks for linking the preexisting issue!

> Need to figure out some mechanism for doing that when there's no natural ordering - as will be the case with files. I don't want to end up with 00-staging.yaml, etc though. 

Well, maybe adding some `priority` key (or any other name) in each of those hubs yaml files is the way to go? Then you can put a 100 prio on the staging one and let the others with minor values accordingly to some deployment prioritization... just an idea...

yuvipanda:
@damianavila or, we can explicitly list 'staging' hubs in a cluster's yaml, making sure those are run first before anything else - the ordering on the rest mostly doesn't matter. What do you think?

yuvipanda:
Do you want me to explore that in this PR, or in a separate one?

damianavila:
> What do you think?

That is another alternative, I like the fact we are not introducing a new key, I do not like the fact there is a privileged hub living in cluster yaml file... but, in fact, the staging one is somehow a special case, so I would not be against that alternative.

 > Do you want me to explore that in this PR, or in a separate one?

Separate one, tied to the previous issue you shared with me.

choldgraf:
+1 on having a separate YAML file per hub. From a hub administrator's perspective, I think this is also much friendlier because we'll have a canonical place to point them to "your hub's configuration is *here*". (right now, if we add config above a hub, it'll push the line number down so old links won't work anymore)

Also agreed that it doesn't have to be in this PR! This is much welcome improvement as-is :-)

docs/howto/customize/configure-login-page.md
```diff
@@ -6,4 +6,8 @@ For an example, see [the log-in page of the staging hub](https://staging.pilot.2
 
 The log-in pages are built with [the base template at this repository](https://github.com/2i2c-org/pilot-homepage). Values are inserted into each template based on each hub configuration.
 
-You may customize the configuration for a hub's homepage at `config.jupyterhub.homepage.templateVars`. Changing these values for a hub will make that hub's landing page update automatically. We recommend [using the `hubs.yaml` file as a reference](https://github.com/2i2c-org/pilot-hubs/blob/master/hubs.yaml). Copy the configuration from another hub, and then modify the contents to fit the new hub that is being configured.
+You may customize the configuration for a hub's homepage at
+`config.jupyterhub.homepage.templateVars`. Changing these values for a hub
+will make that hub's landing page update automatically. 
+Copy the configuration from another hub, and then modify the contents to fit
+the new hub that is being configured.
```
choldgraf, line 13:
```suggestion
You may customize the configuration for a hub's homepage at `config.jupyterhub.homepage.templateVars`. Changing these values for a hub will make that hub's landing page update automatically.
Copy the configuration from another hub, and then modify the contents to fit the new hub that is being configured.
```

Not blocking on this PR, but what do folks think about adopting a "one line per sentence" documentation policy? I find that this strikes a nice balance between "gigantic lines because a whole paragraph is in it" and "shorter lines but with broken up sentences"

docs/howto/customize/custom-image.md
```diff
@@ -44,8 +44,8 @@ these rate limits. We recommend the following image repositories:
 
 ## Configuring your hub to use the custom image
 
-`hubs.yaml` can contain arbitry `zero to jupyterhub on kubernetes` config, so we
-can use that to set the image name & tag. Here is some example config, with only
+Hub config under `config/hubs` can contain arbitry `zero to jupyterhub on kubernetes`
```
choldgraf, line 47:
```suggestion
The hub config under `config/hubs` can contain arbitry [`zero to jupyterhub on kubernetes`](https://zero-to-jupyterhub.readthedocs.io/en/stable/resources/reference.html) 
```

choldgraf (commented):
This looks great to me - a couple pretty minor comments. I think that we can 🚢 and continue iterating on the implementation if you and @damianavila are happy w/ it

docs/howto/customize/configure-login-page.md
yuvipanda: ah, hmm - my text editor can automatically either reflow lines to match a max width, or automatically wrap them so they don't horizontally scroll. However, with this reflow will have to be manual - how do you deal with that in your writing workflow?

I'm also curious how to make this determination without it becoming a timesuck bikeshed :D

choldgraf: haha yeah - for me it has just become natural when I'm writing documentation - I finish a sentence then hit enter. I basically follow the first few steps of https://sembr.org/ but don't spend the time worrying about the whole "split concepts into separate lines" parts of it.

choldgraf: Other than the "make the source code easier to semantically follow" part, I also like the fact that it increases the likelihood that I can suggest an edit with a one-liner in the github PR interface (because edits often happen within-sentence, and the whole sentence would be on the same line)

choldgraf: either way, please do not block this PR on these thoughts!

deploy/hub.py
```diff
@@ -388,28 +282,37 @@ def deploy(self, auth_provider, proxy_secret_key, skip_hub_health_test=False):
...
                     service_api_token = generated_values["base-hub"]["jupyterhub"]["hub"]["services"]["hub-health"]["apiToken"]
                 else:
                     service_api_token = generated_values["jupyterhub"]["hub"]["services"]["hub-health"]["apiToken"]
 
+                domain = self.spec["domain"]
+                if type(domain) == str:
+                    hub_url = f'https://{domain}'
+
+                else:
+                    hub_url = f'https://{random.choice(domain)}'
```
damianavila, line 309: Should not be using always one (the first? the last?) instead of a random one? Just curious about why are you adding that layer of variability...

damianavila (approved):
Left a minor question but it should not be a blocker for this one.

deploy/hub.py
yuvipanda: Nothing particular, I was just porting the old logic. We should probably instead strive to just have one domain per hub though.

choldgraf (approved):

choldgraf:
@yuvipanda wanna merge whenever you think it's ready to go?

yuvipanda:
Yay, thanks @choldgraf, @GeorgianaElena and @damianavila!

PR #357, opened 2021-04-21
Fix missed rename

yuvipanda:
hackathon-alpha was the project name when we thought we
might re-use it for other purposes - but now that the
hub is running constantly, it isn't being used like that.
It's just used for the paleohack hackathon.


yuvipanda:
Follow-up to #352

PR #358, opened 2021-04-26
Fix manual build of cluster images

GeorgianaElena:
This allows building the user image for a given cluster. Before the deploy script refactor, we could build the images for all the cluster.

Because the deploy can only happen now on a per-cluster basis, I though that the build should follow the same pattern.
@yuvipanda, what do you think? Build the image per cluster, or re-implement the old `parse_clusters` function and build for all clusters at once?

### DIFF
--- a/deploy/__main__.py
+++ b/deploy/__main__.py
@@ -20,16 +20,17 @@ from utils import decrypt_file
 # Without `pure=True`, I get an exception about str / byte issues
 yaml = YAML(typ="safe", pure=True)
 
-def build():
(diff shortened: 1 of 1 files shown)


yuvipanda:
Yeah, this makes sense. Can you make `cluster_name` a required commandlien argument as well, and pass it to build?

GeorgianaElena:
> Can you make cluster_name a required commandlien argument as well, and pass it to build?

Ups :facepalm: This is done now. Thanks @yuvipanda!

PR #361, opened 2021-04-27
Remove 'ephemeral hub' type

yuvipanda:
This was supposed to be a 'binder-like' hub, with
the following config:

1. No authentication - uses tmpauthenticator by default to let
   anyone in
2. No persistent storage!
3. More aggressive culling of users

The goal was to provide 'demonstration' hubs for workshops,
etc. However, there has been no current *demand* for this kinda
hub, and the workshops we've supported definitely wanted authentication.
The amount of customization needed is also fairly small, and can
be done on a per-hub basies without much effort.

So, let's just remove this hub type and simplfiy our code.

This removes the two ephemeral hubs we were running. 
@colliand confirmed the callysto one isn't currently in use.


damianavila:
> So, let's just remove this hub type and simplfiy our code.

+ 1 on simplifying the code base!

> The amount of customization needed is also fairly small, and can
be done on a per-hub basies without much effort.

Where that customization needed (that is currently collected in the ephemeral config file) could be "saved"?
It is maybe a subject for a how-to guide specific article in the docs?

yuvipanda:
>  It is maybe a subject for a how-to guide specific article in the docs?

I think part of the problem is that these were arbitrary, and not well defined by use cases. IMO, it's ok to let them die and build up hubs based on use cases when we do get use cases.

damianavila (approved):
LGTM, I always love simplification!

damianavila:
> IMO, it's ok to let them die and build up hubs based on use cases when we do get use cases.

Sounds reasonable to me!

yuvipanda:
@damianavila wanna hit merge? :)

PR #364, opened 2021-04-27
Cleaning up documentation structure

choldgraf:
This is a quick cleanup of the pilot-hubs documentation, so that navbar items do not appear twice. It also expands the left navbar so that second-level items are always visible.


damianavila:
First, thanks for fixing those nav bars!!

Now, looking at the PR build, I still see some kind of mismatch...

Screen Shot 2021-04-27 at 20 00 37

Is it possible to sync both somehow? (I hope I am clear enough, let me know if that is not the case).

choldgraf:
Ahh I bet that is taken from the H2 headers. I quite like the 3-category structure there. Though it is unclear to me the difference between configuration and customization. Is there an obvious difference between them?

choldgraf:
Here's another slight re-org. This time I used top-level `caption` blocks to separate out our sections, which means we don't have to click-through levels as much. Think this is a sensible structure?

image

damianavila:
> Is there an obvious difference between them?

Yep, maybe it makes sense to merge those sections?

> Think this is a sensible structure?

This is far better, IMHO. Thanks!

> Also for what it's worth, I feel like the "Configuration structure" page is more of a "Reference" page than a "Topic Guide" page (since it is basically just a factual table of configuration locations). I didn't change it in this PR though to keep the scope narrower

I agree, but let's keep the scope, as you said.

damianavila (approved):
LGTM now!

choldgraf:
ok cool - let's merge it in and spot-check as needed to keep improving! thanks for the feedback :-)

PR #367, opened 2021-04-28
Document how to delete a hub

### LINKED ISSUE #354: Delete the Peddie Hub
# Summary

I'm pretty sure that the Peddie Academy hub isn't being used anymore - it hasn't had activity that I've seen, and I emailed the instructor that was going to be testing it out, and he didn't respond. I think that we can delete the hub.

This is the hub: https://github.com/2i2c-org/pilot-hubs/blob/master/hubs.yaml#L311

For reference: [Here are the instructions to delete a hub](https://pilot-hubs.2i2c.org/en/latest/howto/operate/delete-hub.html)

yuvipanda:
Doesn't actually delete any hub, as we discovered
that peddie is still in use

Ref #354

### DIFF
--- a/docs/howto/operate/delete-hub.md
+++ b/docs/howto/operate/delete-hub.md
@@ -3,6 +3,21 @@
 If you'd like to delete a hub, there are a few steps that we need to take:
 
 1. **Backup the hub database**. Backup the `jupyterhub.sqlite` db off the hub.
-2. **Backup the home directory contents**. Especially if we think that users will want this information in the future (or if we plan to re-deploy a hub elsewhere).
-3. **Delete the Helm namespace**. Run `helm -n <hub-name> delete <hub-name>`.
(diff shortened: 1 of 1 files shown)


choldgraf:
whoah hold up a sec! I just checked the Admin panel to be sure that this hub wasn't being used, and this is what I saw:

image

choldgraf:
that actually suggests to me that they have been using the hub quite a lot! Did I totally misinterpret this?

yuvipanda:
@choldgraf actually no, I think that's the correct interpretation. I'll hold off...

choldgraf (changes requested):
We shouldn't merge this! After a closer inspection it turns out that Peddie is using this hub after all (yay! but also phew!). Let's change this PR to just be a docs improvement and not remove the Peddie config

yuvipanda:
@choldgraf done!

PR #374, opened 2021-05-03
Put GCP-only 'scratch bucket' behind a flag

yuvipanda:
Pangeo hubs have a `PANGEO_SCRATCH` env variable that
points to a GCS bucket, used to share data between users.
We implement that here too, but with a more generic `SCRATCH_BUCKET`
env var (`PANGEO_SCRATCH` is also set for backwards compat).
https://github.com/pangeo-data/pangeo-cloud-federation/issues/610
has some more info on the use cases for `PANGEO_SCRATCH`

Right now, we use Google Config Connector
(https://cloud.google.com/config-connector/docs/overview)
to set this up. We create Kubernetes CRDs, and the connector
creates appropriate cloud resources to match them. We use this
to provision a GCP Serivce account and a Storage bucket for each
hub.

Since these are GCP specific, running them on AWS fails. This
PR puts them behind a switch, so we can work on getting things to
AWS.

Eventually, it should also support AWS resources via the
AWS Service broker (https://aws.amazon.com/partners/servicebroker/)

Ref https://github.com/2i2c-org/pilot-hubs/issues/366

### DIFF
--- a/hub-templates/daskhub/values.yaml
+++ b/hub-templates/daskhub/values.yaml
@@ -1,3 +1,10 @@
+scratchBucket:
+  # Enable a 'scratch' bucket per-hub, with read-write permissions for all
+  # users. This will set a `SCRATCH_BUCKET` env variable (and a PANGEO_SCRATCH variable
+  # too, for backwards compatibility). Users can share data with each other using
+  # this bucket.
+  enabled: true
+
 base-hub:
   # Copied from https://github.com/dask/helm-chart/blob/master/daskhub/values.yaml
   # FIXME: Properly use the upstream chart.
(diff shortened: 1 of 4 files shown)


choldgraf:
Nice - do we have another deliverable to track implementing Scratch Buckets for AWS/Azure? If not maybe we should make one to track adding this functionality for those providers?

hub-templates/daskhub/values.yaml
```diff
@@ -45,6 +52,7 @@ base-hub:
         # The default worker image matches the singleuser image.
         DASK_GATEWAY__CLUSTER__OPTIONS__IMAGE: '{JUPYTER_IMAGE_SPEC}'
 
+        # FIXME: Only set these if scratchBucket.enabled is true
```
damianavila, line 55: Any plans about how to make this happen? Doing stuff with some `extraConfig` or something else?

damianavila (approved):
Left a minor question, LGTM as an iterative step.

hub-templates/daskhub/values.yaml
yuvipanda: Yeah - that, or something with whatever we decide to do in https://github.com/2i2c-org/pilot-hubs/issues/362.

PR #375, opened 2021-05-03
Fix imports in pytest

yuvipanda:
Missed during earlier rename of deploy to deployer

### DIFF
--- a/deployer/tests/context.py
+++ b/deployer/tests/context.py
@@ -3,4 +3,4 @@ import sys
 sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
 
 import auth
-import deploy
+import deployer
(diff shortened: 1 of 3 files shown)


choldgraf (approved):

damianavila (approved):
LGTM

deployer/tests/context.py
```diff
@@ -3,4 +3,4 @@
 sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
 
 import auth
-import deploy
+import deployer
```
GeorgianaElena, line 6: Since we don't have a hub fixture anymore and  the deployer is in it's own module now, this entire file can be deleted.

GeorgianaElena (approved):
Thanks @yuvipanda! I left a comment about deleting the `context.py` file, but other than this, LGTM!

deployer/tests/context.py
yuvipanda: Ahh, thanks @GeorgianaElena. done

PR #376, opened 2021-05-03
Adding script to pull list of running hubs and display it

choldgraf:
This ports over the code to generate a list of running hubs in this repository. It currently generates the table [in the team compass](https://team-compass.2i2c.org/en/latest/reference/hubs.html#running-jupyterhubs), but I think it'd be better to keep the code here. This way we know that the table is up-to-date, and also know if we change the config in a way that breaks the table. We also don't need authentication to grab the data from GitHub directly since the YAML files are already in this repository.

Here's the list of running hubs:

https://2i2c-pilot-hubs--376.org.readthedocs.build/en/376/reference/hubs.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -131,3 +131,6 @@ dmypy.json
 # helm charts
 hub/charts
 support/charts
+
+# Documentation build assets
+docs/tmp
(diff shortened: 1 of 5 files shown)


damianavila:
So, are you planning to remove the relevant code from the team-compass repo?

choldgraf:
@damianavila yep, if this is merged I will remove that section and cross link it to here

damianavila (approved):
LGTM

choldgraf:
sounds good!

PR #377, opened 2021-05-03
Remove 'nfs' spec from clusters yaml

yuvipanda:
We aren't actually using this anywhere! If we want
to override NFS servers, we should do that in
jupyterhub.nfsPVC

### DIFF
--- a/config/hubs/2i2c.cluster.yaml
+++ b/config/hubs/2i2c.cluster.yaml
@@ -6,8 +6,6 @@ gcp:
   project: two-eye-two-see
   cluster: low-touch-hubs-cluster
   zone: us-central1-b
-  nfs:
-    server: nfs-server-01
 hubs:
   - name: staging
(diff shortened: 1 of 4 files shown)


damianavila:
> We aren't actually using this anywhere! If we want
> to override NFS servers, we should do that in
> jupyterhub.nfsPVC

Defined here, right? https://github.com/2i2c-org/pilot-hubs/blob/master/hub-templates/base-hub/values.yaml#L20

But... should not be a per cluster config instead of a per hub config?

yuvipanda:
@damianavila yeah, I agree that would be nice - per-cluster common hub config. However, this is a remnant from https://github.com/2i2c-org/pilot-hubs/issues/206. If we implement per-cluster common hub config, it should be more generic than just for NFS though.

damianavila (approved):
OK, let's push this one forward, LGTM.

PR #378, opened 2021-05-03
Require trailing slash in baseSharePath

yuvipanda:
In places like EFS, the base path is `/` - and right now
there's no way to actually specify that!


choldgraf:
Huh - there's no way that we can do like an `append` filter like you'd do in Jinja?

damianavila:
I think the trailing slash is a good compromise to make this agnostic from the cloud provider without complex stuff (looking quickly at the helm stuff I do not see an easy way to make it happen).

damianavila (approved):
LGTM if there is not an **easy** way to conditionally add the trailing slash...

PR #382, opened 2021-05-06
Rename base-hub to basehub

yuvipanda:
Referencing base-hub from helm charts templates is basically
impossible - go templates can not use '-' in their name!
This makes referencing them easier


hub-templates/daskhub/Chart.lock
```diff
@@ -1,9 +1,9 @@
 dependencies:
-- name: base-hub
-  repository: file://../base-hub
-  version: 0.0.1
+- name: basehub
+  repository: file://../basehub
+  version: 0.0.1-n569.hb296398
```
GeorgianaElena, line 4: I was a bit confused about the basehub version here.
The `.lock` file is generated based on the `.yaml` file right? Because in `daskhub/Chart.yaml` the basehub dependency is set to just `0.0.1`.

Also, when reading about this `Chart.lock` file, I learned that these are the helm3 equivalents of `requirements.lock` and `.yaml`. But we have both for our helm charts. Is there any reason for this?

GeorgianaElena (approved):
This is LGTM :rocket:

I left a comment about the hub charts, but it's mostly for my education.

damianavila (approved):
LGTM

hub-templates/daskhub/Chart.lock
yuvipanda: I think just removing them is probably the way to go actually.

PR #383, opened 2021-05-06
Set default tolerations for share creator

yuvipanda:
Without this, regular deploys fail

Depends on #382


hub-templates/basehub/values.yaml
```diff
@@ -16,6 +16,8 @@ etcJupyter:
 
 nfsPVC:
   enabled: true
+  shareCreator:
+    tolerations: []
```
damianavila, line 20: Can you explain a little bit more about this change? Why did you need it?

yuvipanda:
Ah, https://github.com/2i2c-org/pilot-hubs/pull/379/files#diff-4c3982c1f7e655df9d9dd69e0d8c75f85d4874a1e7a6f7967191d7c1c6df67f9R25 is what needs this - but I had gotten my PRs confused :)

damianavila:
Ja, Ja... I was thinking I was missing a huge thing because I did not see any `tolerations` key in nfs-share-creator.yaml on this PR and I could not find anything informative online... Thanks for the link!

damianavila (approved):
With the link provided, this one LGTM now.

Issue #368, opened 2021-04-28
Migrate Farallon hub to this repo


yuvipanda:
# Background

The Farallon institute JupyterHub is run from github.com/2i2c-org/pangeo-hubs/, and is [not well automated](https://github.com/2i2c-org/pilot-hubs/discussions/365). We should move that to this repository!

https://github.com/2i2c-org/pilot-hubs/issues/366 tackles fully setting up on AWS, so this goal will focus only on migrating the *helm* config + deployment to this repository.

# Expected timeline

???

# Steps to complete this goal

- [x] Add support for a cluster authentication method that can be used on the farallon hub to `deployer/`
- [x] Move the staging farallon hub 
- [ ] https://github.com/2i2c-org/pilot-hubs/issues/381
- [x] Figure out how to do authentication - right now, Farallon hub uses GitHub orgs for authorization, which auth0 doesn't support yet https://github.com/2i2c-org/pilot-hubs/pull/487
- [x] Move the production farallon hub
- [x] ~Consider removing the farallon staging hub~ (gonna keep it)
- [x] Archive https://github.com/2i2c-org/pangeo-hubs

choldgraf:
Just a quick note on this - I think we should prioritize this below #363 - so I'm attaching a label to mark this as such :-)

yuvipanda:
Once https://github.com/2i2c-org/pilot-hubs/pull/391 is merged, @damianavila is going to go through moving the farallon hub over to the new kops based setup. That will help complete this.

damianavila:
A little reminder to re-revert https://github.com/2i2c-org/pilot-hubs/pull/400 once the above is complete.

yuvipanda:
We should archive https://github.com/2i2c-org/pangeo-hubs once this is complete

yuvipanda:
I've carefully deleted the old EFS instance. I'll now carefully delete the old cluster

yuvipanda:
I've deleted the staging and prod hubs by hand:

```
helm -n farallon-staging delete farallon-staging
helm -n farallon-prod delete farallon-prod
```

This ensures that any external cloud resources created by kubernetes (disks for the hub db dir, load balancer for the `proxy-public` service) are deleted

yuvipanda:
[...]

PR #386, opened 2021-05-06
Cleanup how GCP storage bucket is created

yuvipanda:
- Move everything into the base chart, rather than the daskhub
  chart. No reason this has to be coupled with dask. And if we
  want to use `z2jh.get_config` in `hub.extraConfig` to dynamically
  set environment values, the config needs to be under jupyterhub.
- Scope the resources more clearly to be gcp specific.
- Move some repetitive name constructions to a _helpers.tpl file

Ref #368


damianavila (approved):
This is great @yuvipanda!
It makes sense to move this into some specific GCP place.

PR #390, opened 2021-05-10
Set proxy.https.hosts by default too

yuvipanda:
In some hubs, we currently just use a loadbalancer + autohttps
to get traffic into the cluster. This is simpler than getting
nginx-ingress + certmanager setup, although not entirely
sure if that's the right thing to do long term.

By putting the domains in proxy.https.hosts as well, we can
decide to use or not use autohttps on a per-hub basis, without
having to repeat the domains in multiple places. Staging and
prod hubs that have the exact same config but differ in
domains can be easily constructed thus.

### DIFF
--- a/config/hubs/farallon.cluster.yaml
+++ b/config/hubs/farallon.cluster.yaml
@@ -94,8 +94,6 @@ hubs:
               type: LoadBalancer
             https:
               enabled: true
-              hosts:
-              - staging.farallon.2i2c.cloud 
             chp:
               nodeSelector: {}
               tolerations:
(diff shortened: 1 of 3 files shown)


damianavila:
> In some hubs, we currently just use a loadbalancer + autohttps
to get traffic into the cluster. This is simpler than getting
nginx-ingress + certmanager setup, although not entirely
sure if that's the right thing to do long term.

What about terminating SSL at the load balancer as another option, maybe?

> By putting the domains in proxy.https.hosts as well, we can
decide to use or not use autohttps on a per-hub basis, without
having to repeat the domains in multiple places. Staging and
prod hubs that have the exact same config but differ in
domains can be easily constructed thus.

I like that! This is I am approving this one in a few seconds!

damianavila (approved):

PR #389, opened 2021-05-10
Generate carbonplan kops file with jsonnet

yuvipanda:
Each kops cluster we want to deploy will have some things in
common, but many differences. 2i2c engineers should be able to
create and manage these without a lot of cognitive overhead.

1. All clusters should have some basic opinionated defaults
2. But everything should be customizable as necessary
3. kops configuration is already declerative, and we should
   *not* add another layer on top.

jsonnet helps with this. It's a powerful declarative language
generating JSON, with conditionals, prototype inheritance,
deep merging, maps, etc.

In this PR, we create two base objects that can be customized -
a `cluster` object that generates a kops Cluster object,
and a `instancegroup` object that generates a kops InstanceGroup
object. We use these to generate our kops config for
carbonplan.jsonnet. jsonnet is only used for some abstraction
and deep merging - there isn't an extra level of abstraction
here.

Issue #136, opened 2020-12-11
Standardize on nodepool labeling conventions


yuvipanda:
From @consideRatio in https://github.com/2i2c-org/pilot-hubs/pull/134#discussion_r541023273:

> The configuration scheduling.corePods.nodeAffinity.matchNodePurpose=require will make all core pods of the jupyterhub chart have a required node affinity toward the node label hub.jupyter.org/node-purpose=core.
>
> By using hub.jupyter.org/pool-name it seems like there is a predefined meaning of that label and can cause some confusion, it isn't having a predefined meaning right? If it is a label with meaning for the 2i2c.org pilot hubs, I think we should name it like 2i2c.org/pool-name instead.

We should investigate either node-purpose or switch to a different namespace.

yuvipanda:
We'll3

consideRatio:
> By using hub.jupyter.org/pool-name it seems like there is a predefined meaning of that label and can cause some confusion, it isn't having a predefined meaning right? If it is a label with meaning for the 2i2c.org pilot hubs, I think we should name it like 2i2c.org/pool-name instead.

The k8s label `hub.jupyter.org/pool-name` is no longer in use, we are instead referencing the k8s label `node.kubernetes.io/instance-type` for such purposes I think. We have also stopped using the `hub.jupyter.org` label prefix for labels we define (#1256).

I'll go for a close here, lets open a new issue if there is some concrete aspect to focus on remaining that I've failed to capture.

PR #391, opened 2021-05-10
Add carbonplan cluster + hubs

yuvipanda:
- staging and prod hubs that are exactly the same,
  with just domain differences
- Uses traditional autohttps + LoadBalancer to get traffic
  into the cluster. Could be nginx-ingress later on if necessary.
- Manual DNS entries for staging.carbonplan.2i2c.cloud and
  carbonplan.2i2c.cloud. Initial manual deploy with
  `proxy.https.enabled` set to false to complete deployment,
  fetch externalIP of `proxy-public` service, setup DNS,
  then re-deploy with `proxy.https.enabled` set to true.
- Simple hacky script to setup EFS properly for a cluster
- Use image automatically built at https://github.com/carbonplan/trace/tree/main/envs/python-notebook
- Add ssh keys used to build the cluster
- Standardize labels used for our nodes - see #136 

Depends on #389 and #390 

Ref https://github.com/2i2c-org/pilot-hubs/issues/291


config/hubs/carbonplan.cluster.yaml
```diff
@@ -0,0 +1,182 @@
...
+              c.Backend.cluster_options = cluster_options
+            idle: |
+              # timeout after 30 minutes of inactivity
+              c.KubeClusterConfig.idle_timeout = 1800
+  - name: prod
+    domain: carbonplan.2i2c.cloud
+    template: daskhub
+    auth0:
+      connection: github
+    config: *carbonPlanHubConfig
```
damianavila, line 182: OK, this is nice, I have missed that before...

kops/libsonnet/instancegroup.jsonnet
```diff
@@ -0,0 +1,42 @@
...
+        image: "099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20210119.1",
+        cloudLabels: {
+            // Tell autoscaler to scale up this instancegroup when something asks for a node with the label
+            // node.kubernetes.io/instance-type: <machineType>
+            "k8s.io/cluster-autoscaler/node-template/label/node.kubernetes.io/instance-type": $.spec.machineType
+        } + makeCloudLabels(self.nodeLabels) + makeCloudTaints(self.taints),
+        taints: [],
+        nodeLabels: {},
+        machineType: "",
+        maxSize: 20,
```
damianavila, line 37: Maybe this one should be empty? Since you are populating it above...

kops/libsonnet/instancegroup.jsonnet
```diff
@@ -0,0 +1,42 @@
...
+        minSize: 0,
+        role: "Master",
```
damianavila, line 39: Same with the role, maybe an empty one?

kops/setup-efs.py
```diff
@@ -0,0 +1,63 @@
...
+        {'Name':'tag:KubernetesCluster', 'Values': [cluster_name]}
+    ])['Subnets']
+
+def find_security_groups(cluster_name, region):
+    """
+    Find security groups of master and nodes
+    """
+    ec2 = boto3.client('ec2', region_name=region)
+    return ec2.describe_security_groups(Filters=[
+        {'Name':'tag:Name', 'Values': [f'{t}.{cluster_name}' for t in ['masters', 'nodes']]}
```
damianavila, line 30: Do you really need the SG associated with `masters`?

kops/setup-efs.py
```diff
@@ -0,0 +1,63 @@
...
+    ])['SecurityGroups']
+
+def create_filesystem(token, name, region):
+    efs = boto3.client('efs', region_name=region)
+    subnets = find_subnets(name, region)
+    security_groups = find_security_groups(name, region)
+    fs = efs.create_file_system(
+        CreationToken=token,
+        Encrypted=True,
+        Backup=True,
```
damianavila, line 40: I like that!!

kops/carbonplan.jsonnet
```diff
@@ -0,0 +1,100 @@
...
+];
+
+local data = {
+    cluster: cluster {
+        metadata+: {
+            name: "carbonplanhub.k8s.local"
+        },
+        spec+: {
+            // FIXME: Not sure if this is necessary?
+            configBase: "s3://2i2c-carbonplan-kops-state/%s" % data.cluster.metadata.name
```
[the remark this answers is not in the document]
yuvipanda: kops seems to automatically set this up, I removed it from the openscapes hub and it seems ok.

hub-templates/daskhub/values.yaml
```diff
@@ -217,6 +216,6 @@ dask-gateway:
       type: jupyterhub  # Use JupyterHub to authenticate with Dask Gateway
   traefik:
     nodeSelector:
-      hub.jupyter.org/pool-name: core-pool
+      k8s.dask.org/node-purpose: core
```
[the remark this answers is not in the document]
yuvipanda: Yeah, they're orthogonal to each other, and are much more self-documenting.

kops/libsonnet/instancegroup.jsonnet
yuvipanda: @damianavila so the jsonnet convention seems to be to say `error "Must set role"`, so the output will fail unless it is overriden. Probably the way to go

kops/setup-efs.py
yuvipanda: I checked - the 'masters' is for all master nodes. Since the share creator runs there, we should probably keep that.

The alternative is to do one EFS per hub, but I don't know if I entirely like that.

yuvipanda: I've added some comments to the script.

kops/carbonplan.jsonnet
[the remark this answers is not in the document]
damianavila: OK, so you can remove it from here,  I guess...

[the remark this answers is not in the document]
yuvipanda: @damianavila i'm not sure what it'll do now that we've deployed the carbonplan hub, so I'll leave it in here. That's another advantage of the jsonnet approach - we can make changes in the base like we have without having to redeploy everywhere

kops/setup-efs.py
damianavila: > The alternative is to do one EFS per hub, but I don't know if I entirely like that.

Having an EFS per hub could be something needed from a security point of view, maybe?
Btw, I do think it is a conversation for later (not something we need to figure it out for this PR).

yuvipanda:
I'll need to deploy and merge https://github.com/2i2c-org/pilot-hubs/pull/397 first before merging this, so as to not break existing clusters.

kops/libsonnet/cluster.jsonnet
```diff
@@ -1,4 +1,33 @@
...
+//    with the default settings. Explicitly decided against AWS-VPC cluster networking
+//    due to pod density issues - see https://github.com/2i2c-org/pangeo-hubs/issues/28.
+// 3. Nodes in only one subnet in one AZ. Ideally, the master would be multi-AZ but
+//    the nodes would be single AZ. Multi AZ workers run into problems attaching PVs
+//    from other AZs (for hub db PVC, for example), and incurs networking cost for no
+//    clear benefit in our use case. An opinionated set of IP ranges is picked here,
+//    and the subnet is created in _config.zone.
+// 4. A /16 network for the entire cluster, with a /19 allocated to the one subnet
+//    currently in use. This allows for ~8000 currently active pods.
+//    FIXME: Consider a /18 instead?
```
damianavila, line 18: Can you elaborate a little bit more about these decisions? 
Why a /16 network for the entire cluster?
Btw, /19 (~ 8000 pods seems) seems a sensible number for now, IMHO.

kops/libsonnet/cluster.jsonnet
```diff
@@ -1,4 +1,33 @@
...
+// 5. Kubernetes API and SSH access allowed from everywhere.
+// 6. IAM Permissions to pull from ECR.
+// 7. Enables feature gates to allow hub services to run on master node as well.
+// 8. Docker as the container runtime.
+//
```
damianavila, line 23: This is a super useful summary!
Thanks for taking the time to document it!!

kops/libsonnet/cluster.jsonnet
yuvipanda: @damianavila ah, these are just the kops defaults! I'll document that too.

yuvipanda: Updated!

damianavila: Since we are on this discussion :wink: what is the network story in GKE?
Looking at this one right now: https://cloud.google.com/kubernetes-engine/docs/how-to/flexible-pod-cidr
Are we using the default values?

yuvipanda: Yeah, we're just using the default values. This is from terraform in this repo at `terraform/`

damianavila: Yep, I just wanted to double-check because I looked into the terraform stuff in this repo and I did not find any customization (and sometimes things are deployed on the fly and we - at least I - forgot about them 😉 ).

yuvipanda:
With #397 all GKE clusters have new labels. I've removed Farallon from CI/CD in https://github.com/2i2c-org/pilot-hubs/pull/400

yuvipanda:
Ok this one is ready to merge :)

damianavila (approved):
I have been reviewing this one as it was evolving.
Last final check LGTM
This is great @yuvipanda!
Feel free to push the green button!

yuvipanda:
@damianavila done!

PR #401, opened 2021-05-14
pilot-hubs repo no longer has a gh-master branch

yuvipanda:
Builds are now served from RTD rather than GitHub pages,
so there's no gh-pages branch to pull. nbgitpuller repo
does, so let's use that for now.

This causes current deployments to fail.

### DIFF
--- a/config/hubs/2i2c.cluster.yaml
+++ b/config/hubs/2i2c.cluster.yaml
@@ -15,8 +15,8 @@ hubs:
     config:
       docs_service:
         enabled: true
-        repo: https://github.com/2i2c-org/pilot-hubs
-        branch: gh-pages
+        repo: https://github.com/jupyterhub/nbgitpuller
+        branch: master
       jupyterhub:
         homepage:
           templateVars:


choldgraf:
Just realized you had requested reviews... Sorry! I merged because it seemed important and the fix seemed reasonable to me as a short term solution

yuvipanda:
@choldgraf I always request reviews as 'people who can maybe hit merge' :D Not as something someone is required to do

config/hubs/2i2c.cluster.yaml
```diff
@@ -15,8 +15,8 @@ hubs:
     config:
       docs_service:
         enabled: true
-        repo: https://github.com/2i2c-org/pilot-hubs
-        branch: gh-pages
+        repo: https://github.com/jupyterhub/nbgitpuller
+        branch: master
```
damianavila, line 19: Should not be the `gh-pages` branch as well? Otherwise, why we are not pointing to the master pilot-hubs?

damianavila: @yuvipanda @choldgraf ⬆️

yuvipanda: Haha great catch, @damianavila. https://github.com/2i2c-org/pilot-hubs/pull/408 fixes it

PR #407, opened 2021-05-14
Mark CI service account key as explicit

yuvipanda:
Required for compatibility with newer terraform
versions. Doesn't affect anything.


damianavila (approved):
It makes sense to mark the key as sensitive to prevent leakage.

PR #408, opened 2021-05-14
Use gh-pages branch of nbgitpuller

yuvipanda:
Docs are rendered and pushed there.
Follow-up to https://github.com/2i2c-org/pilot-hubs/pull/401


damianavila (approved):
LGTM, merging now.

PR #411, opened 2021-05-14
Add flake8 checker action + fix flakes

yuvipanda:
- Copied .flake8 file from upstream JupyterHub, but removed
  the ignore for unused module imports and local variables
- Fix flake8 warnings
- Setup a github action for the lint

- Turn on flake8 warning for unused local variable
  and fix bug found due to this!

  Don't overwrite user's KUBECONFIG

  Without this, your local authentication with kubectl
  is reset to that of the serviceaccount used to deploy the
  cluster. If you try and switch cluster, it'll fail with
  strange errors.

  Now we properly set the env vars while authenticated so
  it does not leak out.

  Found because I turned on flake8 warning for unused
  local variables!

### DIFF
--- a/.github/workflows/python-lint.yaml
+++ b/.github/workflows/python-lint.yaml
@@ -1,0 +1,21 @@
+name: Python lint
+
+on:
+  push:
+    paths:
+      - deployer/
+
+jobs:
+  flake8_py3:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Setup Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.8
+      - name: Install flake8
+        run: pip install flake8
+      - name: Run flake8
+        run: |
+          flake8 deployer/
--- a/.github/workflows/python-lint.yaml
+++ b/.github/workflows/python-lint.yaml
@@ -11,7 +11,7 @@ jobs:
     steps:
       - uses: actions/checkout@master
       - name: Setup Python
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v2
         with:
           python-version: 3.8
       - name: Install flake8
--- a/.github/workflows/python-lint.yaml
+++ b/.github/workflows/python-lint.yaml
@@ -18,4 +18,4 @@ jobs:
         run: pip install flake8
       - name: Run flake8
         run: |
-          flake8 deployer/
+          flake8 deployer/ docs/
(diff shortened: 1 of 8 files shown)


choldgraf:
Could we just use a precommit check? We might wanna add things like black one day too and this could make it easier to extend with just a single job

yuvipanda:
I've never managed to get pre-commit to work any time in any of my projects... For JupyterHub I manually run black :D Happy to take that if someone else can set it up tho.

This commit (https://github.com/2i2c-org/pilot-hubs/pull/411/commits/ea2c213865597dbed2bd19839517847637a82cd9) does fix a bug tho.

choldgraf:
Yeah fair enough - maybe something we can just keep on our radar, don't think it should block this PR though since some lint is better than none 😅

.github/workflows/python-lint.yaml
```diff
@@ -0,0 +1,21 @@
...
+    paths:
+      - deployer/
+
+jobs:
+  flake8_py3:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Setup Python
+        uses: actions/setup-python@v1
```
GeorgianaElena, line 14: Is there any reason for not using the v2 version of the action? (I don't know how they differ in terms of features tho...)

yuvipanda: Ah good catch! I've updated :)

deployer/hub.py
```diff
@@ -71,22 +68,28 @@ def auth_gcp(self):
         location = config.get('zone', config.get('region'))
         cluster = config['cluster']
         with tempfile.NamedTemporaryFile() as kubeconfig:
-            with decrypt_file(key_path) as decrypted_key_path:
+            orig_kubeconfig = os.environ.get('KUBECONFIG')
```
GeorgianaElena, line 71: Oh wow... You caught this because KUBECONFIG wasn't used, right?
Hooray to having a linter now :tada:

GeorgianaElena (approved):
Thanks for setting this up, @yuvipanda!
I believe this is ready to go and hooray for having it prove its utility already! <3

deployer/hub.py
yuvipanda: EXACTLY! It was messing with my local credentials too

yuvipanda:
Thanks, @GeorgianaElena!

damianavila:
> Don't overwrite user's KUBECONFIG

That was a nice catch from the linter... and you!
Belated LGTM 😉

PR #413, opened 2021-05-15
Add UToronto to list of running hubs

yuvipanda:
Ref #175

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -103,5 +103,4 @@ df = pd.DataFrame(hub_list)
 path_tmp = Path("tmp")
 path_tmp.mkdir(exist_ok=True)
 path_table = path_tmp / "hub-table.csv"
-if not path_table.exists():
-    df.to_csv(path_table, index=None)
+df.to_csv(path_table, index=None)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -72,7 +72,7 @@ hub_list = [{
     'name': 'University of Toronto',
     'domain': 'jupyter.utoronto.ca',
     'id': 'utoronto',
-    'template': 'base-hub (external repo)'
+    'template': 'base-hub ([deployment repo](https://github.com/utoronto-2i2c/jupyterhub-deploy/))'
 }]
 for cluster_info in clusters:
     if "schema" in cluster_info.name:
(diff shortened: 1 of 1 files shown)


choldgraf:
We do keep track of this here: https://team-compass.2i2c.org/en/latest/reference/hubs.html

I imagined that this list was *just* for the pilot hubs repository, and the list in the team compass would be for all hub infrastructure across repositories. Do you think there's a better way to structure that?

yuvipanda:
Ah, I didn't know of that one. I'd like them all to be in one place though - mostly as a way to show people 'here is all the hubs we run', regardless of location of repository.

choldgraf:
Well yes that was the goal of https://team-compass.2i2c.org/en/latest/reference/hubs.html :-)

It used to have the table from this repository as well, but we moved it here because it was brittle having a table in one repo that used data from another repo.

Another option would be to move everything in that page to the docs for this repo, I just want sure if that would confuse things

yuvipanda:
> Another option would be to move everything in that page to the docs for this repo, I just want sure if that would confuse things

I like that :)

Once #368 is done we'll archive the pangeo-hubs repo as well, so that entry would be gone too.

choldgraf:
Ok cool so then I think the order of operations is:

1. Add the extra hub repos to the list here (we should make clear that those are hosted at a different repo somehow)
2. Remove from team compass and link to these docs from the TC page

yuvipanda:
Ok, done

image

I don't want to add pangeo-hubs repo here - Farallon is already listed here.

choldgraf:
Nice! Do you wanna just leave the Pangeo hubs in the team compass table? I was imagining we'd just remove that table entirely and point here...

yuvipanda:
Yeah, I was thinking we'll remove that table entirely too. I'll do a PR there later

choldgraf:
But then won't we not have any record of the pangeo hubs repo existing? Or do you plan to delete that super quickly?

yuvipanda:
@choldgraf yeah, no new work can happen there now anyway. I'm tempted to archive it already...

yuvipanda:
@choldgraf I opened https://github.com/2i2c-org/team-compass/pull/103, but also left a comment there about utoronto hub

choldgraf (approved):
Sounds good!

damianavila (approved):
LGTM, merging now.

Issue #362, opened 2021-04-27
Simplify enabling / disabling dask-gateway for a hub


yuvipanda:
While most hubs we offer do not need dask-gateway, some do. We want to support this in the easiest way possible. Currently, this is done via the `daskhub` chart in `hub-templates/daskhub`, which has a dependency on dask-gateway and `hub-templates/base-hub`. This lets us put dask specific config in `hub-templates/daskhub/values.yaml`.

This is extreme overkill, and causes problems in our deploy scripts - look at the amount of special casing for `daskhub` in `deploy/`. The only things the `hub-templates` mechanism offers is:

1. Conditional installation of a helm chart
2. Conditional setting of large swaths of config (currently in `hub-templates/daskhub/values.yaml`.

There should be simpler ways to accomplish both of these without resorting to a full extra chart.

yuvipanda:
I've removed the `ephemeral-hub` template in https://github.com/2i2c-org/pilot-hubs/pull/361 as a precursor to this.

yuvipanda:
My current idea is to turn our YAML config into [jsonnet](https://jsonnet.org/), a non-turing complete, expressive configuration language that is JSON native. I quite like it, and it was built primarily to generate k8s objects and similar deeply nested JSON structures - and works well for that. There are python bindings as well.

damianavila:
In this case, I think the cost of adopting a new language is less than the complexity around the hierarchical structure and all the specific code to handle that special case.

yuvipanda:
Thinking of dask as an additional service we want to deploy with JupyterHub, BinderHub is too! And right now it's difficult to mix and match since we use *inheritance* of charts, rather than *composing* them. For example, https://github.com/jupyterhub/binderhub/blob/0b4462c0b1fb15e355b32b38b99c60f440666e94/helm-chart/binderhub/values.yaml#L56 is the jupyterhub block in binderhub, some that conflicts with the jupyterhub block in the dask hub chart. But there's no reason they need to conflict, and if we had a better way of composing them, it will be all good. We'll end up with linear complexity rather than the combinatorial one we have now.

I think @jacobtomlinson talked about this too, and so did @rabernat

choldgraf:
> if we had a better way of composing them, it will be all good.

This feels like a good summary of a good chunk of the reason for https://github.com/jupyterhub/team-compass/issues/382

Agreed that making this more straightforward would be very impactful, "composability" and "modularity" seems like things we should be striving for!

[...]

PR #415, opened 2021-05-16
Move cloudResources under jupyterhub.custom

yuvipanda:
With hhttps://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2200,
helm will default to not allowing arbitrary extra properties anywher
except under jupyterhub.custom. This helps us move towards 1.0

Ref https://github.com/2i2c-org/pilot-hubs/issues/414

### DIFF
--- a/hub-templates/basehub/values.yaml
+++ b/hub-templates/basehub/values.yaml
@@ -28,12 +28,13 @@ nfsPVC:
     baseShareName: /export/home-01/homes/
 
 jupyterhub:
-  cloudResources:
-    provider: null
-    gcp:
-      projectId: null
-    scratchBucket:
-      enabled: false
+  custom:
+    cloudResources:
+      provider: null
+      gcp:
+        projectId: null
+      scratchBucket:
+        enabled: false
   ingress:
     enabled: true
     annotations:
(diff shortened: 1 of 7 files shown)


consideRatio (approved):
LGTM, i'm on a mobile phone atm - perhaps also do a final check searching for not-renamed references and then go for merge.

yuvipanda:
Thanks, @consideRatio! Can you tell me more about what 'not-named references' are?

consideRatio:
Ah sorry, i meant to do a final search for references to the old name of the passed value: .Values.jupyterhub.cloudResources

Just to check not missing out renaming a reference to that value.

yuvipanda:
Ah yeah - I already did that. I used search and replace, and now I've verified it again

consideRatio:
Wieee! I pressed merge 🥳😱

yuvipanda:
yay! Thank you, @consideRatio

hub-templates/basehub/values.yaml
```diff
@@ -287,3 +288,28 @@ jupyterhub:
...
+          env = {
+            'SCRATCH_BUCKET_PROTOCOL': bucket_protocol,
+            # Matches "daskhub.scratchBUcket.name" helm template
+            'SCRATCH_BUCKET_NAME': bucket_name,
+            # Use k8s syntax of $(ENV_VAR) to substitute env vars dynamically in other env vars
+            'SCRATCH_BUCKET': f'{bucket_protocol}://{bucket_name}/$(JUPYTERHUB_USER)',
+            'PANGEO_SCRATCH': f'{bucket_protocol}://{bucket_name}/$(JUPYTERHUB_USER)',
+          }
+
+          c.KubeSpawner.environment.update(env)
```
damianavila, line 315: @yuvipanda, I had the impression this piece was daskhub-specific? Why did you moved it to the basehub?

yuvipanda: cloudResoures was already under `jupyterhub.custom`, rather `daskhub.custom` - `get_config` can only read things under `jupyterhub`. None of this is specific to dask-gateway either - you can just as well want an s3 bucket without dask-gateway. Finally, for https://github.com/2i2c-org/pilot-hubs/issues/362 I've been trying to slowly remove everything under daskhub/ that doesn't strictly belong there...

consideRatio: > None of this is specific to dask-gateway either - you can just as well want an s3 bucket without dask-gateway. Finally, for #362 I've been trying to slowly remove everything under daskhub/ that doesn't strictly belong there...

@yuvipanda it could be reasonable to let the JupyterHub helm chart stash away `global` values into the k8s Secret that is exposed to the hub pod, but I'm not sure. Open an issue in z2jh if you want to suggest such feature is implemented.

yuvipanda: I want to remove the extra chart anyway, for #362. So I think it's ok to keep it like this for now

damianavila: > None of this is specific to dask-gateway either - you can just as well want an s3 bucket without dask-gateway.

Gotcha! I had the wrong impression because I heard about the scratch bucket in a Dask context, but you are actually right about the scratch bucket could be also useful in other environments.
Thanks for the clarification!

PR #416, opened 2021-05-17
Fix image build script + upgrade otter

yuvipanda:
- Fix deployer build setup so builds & pushes actually work
- Authenticate to docker by setting up appropriate credential
   helper. Currently supports google artifact registry only, since
   that's the only thing we use this piece of code for.
    
   https://docs.docker.com/engine/reference/commandline/login/#credential-helpers
   has more information
- Upgrade version of otter

Fixes https://github.com/2i2c-org/pilot/issues/71

### DIFF
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -30,6 +30,7 @@ class Cluster:
         ]
 
     def build_image(self):
+        self.ensure_docker_credhelpers()
         build_image(self.spec['image_repo'])
 
     @contextmanager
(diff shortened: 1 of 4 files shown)


yuvipanda:
I want to split out the image from this repo and do something a bit more long term sustainable, but for now this fixes the immediate issue of older otter grader version

deployer/hub.py
```diff
@@ -41,6 +42,39 @@ def auth(self):
...
+        if helper is not None:
+            dockercfg_path = os.path.expanduser('~/.docker/config.json')
+            try:
+                with open(dockercfg_path) as f:
+                    config = json.load(f)
+            except FileNotFoundError:
+                config = {}
+
+            helpers = config.get('credHelpers', {})
+            helpers[registry] = helper
```
GeorgianaElena, line 73: I think we should add a check here to see if there is already the expected pair of `registry-helper`.
Otherwise, I think we will end up modifying the config file every time we build the image.

But since building the image already takes a loooooooot of time, the time to update the config file doesn't even matter, so my suggestion shouldn't block this PR  that already looks LGTM :rocket:

yuvipanda: @GeorgianaElena good catch! Fixed :)

GeorgianaElena:
Thanks @yuvipanda! :)

deployer/hub.py
```diff
@@ -41,6 +42,40 @@ def auth(self):
...
+        image_name = self.spec['image_repo']
+        registry = image_name.split('/')[0]
+
+        helper = None
+        # pkg.dev is used by Google Cloud Artifact registry
+        if registry.endswith('pkg.dev'):
+            helper = 'gcloud'
+
+        if helper is not None:
+            dockercfg_path = os.path.expanduser('~/.docker/config.json')
```
damianavila, line 65: I usually do not like to "mess" with user's config files but given your other comment about splitting this into their own repo, I think it is OK for now :wink:

PR #417, opened 2021-05-17
Move homepage under custom.homepage

yuvipanda:
Ref https://github.com/2i2c-org/pilot-hubs/issues/414

In preparation for strict schema validation of helm config
from https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2200

### DIFF
--- a/config/hubs/2i2c.cluster.yaml
+++ b/config/hubs/2i2c.cluster.yaml
@@ -18,21 +18,22 @@ hubs:
         repo: https://github.com/jupyterhub/nbgitpuller
         branch: gh-pages
       jupyterhub:
-        homepage:
-          templateVars:
-            org:
-              name: 2i2c Staging
-              url: https://2i2c.org
-              logo_url: https://2i2c.org/media/logo.png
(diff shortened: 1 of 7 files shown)


config/hubs/2i2c.cluster.yaml
```diff
@@ -57,21 +58,21 @@ hubs:
...
-              designed_by:
-                name: 2i2c
-                url: https://2i2c.org
-              operated_by:
-                name: 2i2c
-                url: https://2i2c.org
-              funded_by:
-                name: 2i2c
-                url: https://2i2c.org
+            homepage:
```
consideRatio, line 61: Custom: missing here i think

consideRatio (approved):
This LGTM! If i press merge, it deploys? Im a bit afriad of causing a disruption so wont do this right now on a mobile phone. 👍👍👍 for merge though

yuvipanda:
Thanks for the review, @consideRatio :) And yes, it does deploy if you press merge :)

PR #418, opened 2021-05-17
Support username / password auth provider

yuvipanda:
Uses a 'database' connector[1] with auth0 to provide
username / password authentication.

demo.cloudbank.2i2c.cloud is now moved over to this,
making it easier for people to give demos. This also
removes any current restrictions on who can log in,
and opens it up to everyone. We add some resource
restrictions to match this.

Fixes https://github.com/2i2c-org/pilot-hubs/issues/403

[1]: https://auth0.com/docs/connections/database

### DIFF
--- a/config/hubs/cloudbank.cluster.yaml
+++ b/config/hubs/cloudbank.cluster.yaml
@@ -186,7 +186,7 @@ hubs:
     domain: demo.cloudbank.2i2c.cloud
     template: basehub
     auth0:
-      connection: google-oauth2
+      connection: password
     config:
       jupyterhub:
         homepage:
@@ -206,15 +206,15 @@ hubs:
               url: http://cloudbank.org/
         hub:
           config:
-            Authenticator:
-              allowed_users: &demo_users
-                - yuvipanda@gmail.com
-                - choldgraf@gmail.com
-                - georgiana.dolocan@gmail.com
-                - ericvd@gmail.com
-                - sean.smorris@berkeley.edu
-                - colliand@gmail.com
-              admin_users: *demo_users
+            JupyterHub:
+              # No more than 100 users at a time
+              active_server_limit: 100
+        cull:
+          # Cull after 30min of inactivity
+          every: 300
+          timeout: 1800
+          # No pods over 12h long
+          maxAge: 43200
   - name: lassen
     domain: lassen.cloudbank.2i2c.cloud
     template: basehub
--- a/deployer/auth.py
+++ b/deployer/auth.py
@@ -6,7 +6,8 @@ from auth0.v3.management import Auth0
 USERNAME_KEYS = {
     'github': 'nickname',
     'google-oauth2': 'email',
-    'ORCID': 'sub'
+    'ORCID': 'sub',
+    'password': 'email'
 }
 
 
@@ -79,6 +80,7 @@ class KeyProvider:
                 }
             )
 
+
     def ensure_client(self, name, domains, connection_name):
         current_clients = self.get_clients()
         if name not in current_clients:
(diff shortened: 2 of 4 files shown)


yuvipanda:
image

is how it looks.

choldgraf:
Nice! So what is the expected UX here? Is it:

- user asks for access to demo hub
- we tell them to make an account with auth0
- we manually add that to the hub?

Or can they sign up automatically somehow?

yuvipanda:
@choldgraf there's a 'sign up' button there, so they can just sign up. There's a 100 user active limit there, but otherwise no limits.

choldgraf:
ah very cool - so "sign up" means "sign up for the hub", basically. Is that some kind of check to prevent abuse? Or do we just need to trust that whoever signs up will not be able to mess things up too badly?

yuvipanda:
> Or do we just need to trust that whoever signs up will not be able to mess things up too badly?

Basically ^. We can bring the max number of currently active users down (maybe to 40?) if we are worried.



> Oh and are they signing up with some kind of centralized Auth0 account? Or is this unique to our JupyterHub? (e.g., are we asking users to share their data with auth0 as a part of this?)

This is unique to each JupyterHub.

choldgraf:
Awesome, I am not too worried about abuse (particularly since these are not highly-advertised hubs)...I suppose we'd have options at our disposal if this ever became a major issue though.

deployer/auth.py
```diff
@@ -89,12 +91,31 @@ def ensure_client(self, name, domains, connection_name):
             self._ensure_client_callback(client, domains)
 
         current_connections = self.get_connections()
+
+        if connection_name == 'password':
+            # All hubs attached to a single database 'connection'
+            # will share username / password. So we create one
+            # connection per hub.
```
GeorgianaElena, line 98: When I first read this comment, my english failed me and actually understood it the other way around (that the hubs share the same database).
@yuvipanda, do you want to rephrase this a bit to make it "english-bad-day" proof? :sweat_smile:

yuvipanda: @GeorgianaElena done!

GeorgianaElena (approved):
LGTM :rocket: Thanks @yuvipanda 

Reiterating Chris' question about what user data we're sharing with auth0. The databases are hosted by auth0, right? and they're only storing emails and encrypted passwords?

deployer/auth.py
GeorgianaElena: Ty! :D

config/hubs/cloudbank.cluster.yaml
```diff
@@ -206,15 +206,15 @@ hubs:
...
           config:
-            Authenticator:
-              allowed_users: &demo_users
-                - yuvipanda@gmail.com
-                - choldgraf@gmail.com
-                - georgiana.dolocan@gmail.com
-                - ericvd@gmail.com
-                - sean.smorris@berkeley.edu
-                - colliand@gmail.com
-              admin_users: *demo_users
```
damianavila, line 217: Who are the admin users now?

deployer/hub.py
```diff
@@ -100,7 +101,7 @@ def __init__(self, cluster, spec):
         self.cluster = cluster
         self.spec = spec
 
-    def get_generated_config(self, auth_provider, secret_key):
+    def get_generated_config(self, auth_provider: KeyProvider, secret_key):
```
damianavila, line 104: Thinking about typing?

deployer/auth.py
```diff
@@ -89,12 +91,32 @@ def ensure_client(self, name, domains, connection_name):
             self._ensure_client_callback(client, domains)
 
         current_connections = self.get_connections()
+
+        if connection_name == 'password':
+            # Users should not be shared between hubs - each hub
+            # should have its own username / password database.
+            # So we create a new 'database connection' per hub,
+            # instead of sharing one across hubs.
+            db_connection_name = f'database-{name}'
```
damianavila, line 100: Wondering if there is a way to support the same user auth against several (but not all) of our hubs.
The use case I am thinking around (and avoid) is having several demo hubs and asking the user to sign up for each of them...

deployer/hub.py
yuvipanda: haha yes, now that I'm on an editor  (vscode) that uses a proprietary (pylance) plugin to provide better autocomplete

deployer/auth.py
yuvipanda: Yeah, if we get to that we should be able to add an extra field here to set connection name. However, their user data will not transfer, making the idea of 'same user' tricky.

damianavila: > However, their user data will not transfer, making the idea of 'same user' tricky.

Yep, that's a good point.

PR #420, opened 2021-05-18
update carbonplan image

jhamman:
We'll be iterating on the contents of this image for a bit so its a good time to just point to latest.


yuvipanda:
Alright!

PR #424, opened 2021-05-18
Added SJCC to hubs

sean-morris:
This is the configuration for the sjcc hub.


yuvipanda:
The readthedocs failure is related to https://github.com/2i2c-org/pilot-hubs/pull/417, should be investigated and fixed.

Merging this PR before that since @sean-morris has a 3PM Pacific deadline :)

PR #425, opened 2021-05-18
Move homepage under jupyterhub.custom

yuvipanda:
Follow-up https://github.com/2i2c-org/pilot-hubs/pull/424


yuvipanda:
Turns out the failure in #424 did matter! With https://github.com/2i2c-org/pilot-hubs/pull/417 homepage now needs to be under jupyterhub.custom, not just jupyterhub.

yuvipanda:
Going to self-merge this to unbreak the build, and make sure the SJCC hub is working.

Issue #421, opened 2021-05-18
Allow setting admin users for hubs with username / password auth


yuvipanda:
# Summary

We now support username / password to authenticate with JupyterHub (#418). Currently, there seems to be no easy way to specify that some users are admins, but everyone else can just log in. If I set anyone to admin, other users get a permission denied.

# User Stories

As someone using the demohub (demo.cloudbank.2i2c.cloud) to demo things, I want to be able to access the admin interface and stop / start user servers if needed.

As someone running a workshop with username / password access, I want to use the admin interface so I can access users' home directories to help them debug problems.

# Acceptance criteria

- The demo cloudbank hub has 2i2c admins on it

damianavila:
>If I set anyone to admin, other users get a permission denied.

So, if you set `admin_users` to somebody and avoid `allowed_users`, are you getting the permission denied?

Accordingly to the docs, an empty `allowed_user`should not restrict anyone: https://jupyterhub.readthedocs.io/en/latest/api/auth.html#jupyterhub.auth.Authenticator.allowed_users

yuvipanda:
> So, if you set `admin_users` to somebody and avoid `allowed_users`, are you getting the permission denied?

Yeah, that's what is happening. And I'm not sure why.

damianavila:
@consideRatio, do you have any thoughts about this? Wondering if something is happening at the z2jh layer instead of the underlying JupyterHub...

consideRatio:
@damianavila hmmm no the authentication layer in z2jh is just passing through information directly to JupyterHub.

I think this is the logic that decides it. If users are denied, do you see an error log message about "user name <...> is not in Authenticator.allowed_users" ?

yuvipanda:
Derp, this was my mistake. I had set allowed_users as well, so it was behaving as expected.

https://github.com/2i2c-org/pilot-hubs/pull/426 fixes this.

PR #426, opened 2021-05-20
Set admin users for demo hub

yuvipanda:
Since we want everyone to be able to sign up, we don't set
allowed_users.

Fixes https://github.com/2i2c-org/pilot-hubs/issues/421

### DIFF
--- a/config/hubs/cloudbank.cluster.yaml
+++ b/config/hubs/cloudbank.cluster.yaml
@@ -212,6 +212,15 @@ hubs:
                 url: http://cloudbank.org/
         hub:
           config:
+            Authenticator:
+              # Everyone should be able to sign up, so we don't set allowed_users
+              # These folks should still have admin tho
+              admin_users:
+              - yuvipanda@gmail.com
+              - choldgraf@gmail.com
+              - georgiana.dolocan@gmail.com
+              - ericvd@gmail.com
+              - sean.smorris@berkeley.edu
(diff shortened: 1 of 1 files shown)


consideRatio:
Hmmm the changes suggested only adds a admins section, not anything about removing allowed_users. How then could it fix #421?

yuvipanda:
@consideRatio ah, because when I filed #421 I was setting both allowed_users and admin_users. For spelman and other hubs, we want the admins to explicitly add other users who should be allowed to login (see [docs](https://pilot.2i2c.org/en/latest/admin/howto/manage-users.html)). So by config we'll only allow admin users to log in - everyone else is manually entered in the hub admin interface. The demo hub is different.

consideRatio:
Btw, shouldn't admin users always have access? I wonder if something like this could work?

```
              allowed_users: []
              admin_users:
                - yuvipanda@gmail.com
                - colliand@gmail.com
                - choldgraf@gmail.com
                - georgiana.dolocan@gmail.com
                - ericvd@gmail.com
                - sean.smorris@berkeley.edu
                - jerry.volcy@gmail.com
```

damianavila:
> Btw, shouldn't admin users always have access?

I was saying the same in a previous conversation with @GeorgianaElena: https://github.com/2i2c-org/pilot-hubs/pull/395#discussion_r630516365

PR #428, opened 2021-05-21
Custom logout handler

GeorgianaElena:
Fixes https://github.com/2i2c-org/pilot-hubs/issues/422

### DIFF
--- a/deployer/auth.py
+++ b/deployer/auth.py
@@ -165,10 +165,8 @@ class KeyProvider:
             'username_key': USERNAME_KEYS[connection_name],
             'client_id': client['client_id'],
             'client_secret': client['client_secret'],
-            'scope': ['openid', 'name', 'profile', 'email']
+            'scope': ['openid', 'name', 'profile', 'email'],
+            'logout_redirect_url': f'https://{self.domain}/v2/logout?client_id={client["client_id"]}'
         }
-
-        if connection_name == 'password':
-            auth['logout_redirect_url'] = f'https://{self.domain}/v2/logout?client_id={client["client_id"]}'
 
         return auth
--- a/deployer/auth.py
+++ b/deployer/auth.py
@@ -57,22 +57,11 @@ class KeyProvider:
         return created_client
 
     def _get_callback_url_list(self, domains):
-        if isinstance(domains, list):
-            callbacks = []
-            for domain in domains:
-                callbacks.append(f'https://{domain}/hub/oauth_callback')
-            return callbacks
-
-        return [f'https://{domains}/hub/oauth_callback']
+        return [f'https://{domain}/hub/oauth_callback' for domain in domains] if isinstance(domains, list) else [f'https://{domains}/hub/oauth_callback']
 
     def _get_allowed_logout_url_list(self, domains):
-        if isinstance(domains, list):
-            logout_urls = []
-            for domain in domains:
-                logout_urls.append(f'https://{domain}/hub/')
-            return logout_urls
+        return [f'https://{domain}/hub/' for domain in domains] if isinstance(domains, list) else [f'https://{domains}/hub/']
 
-        return [f'https://{domains}/hub/']
 
     def _ensure_client_callback(self, client, domains):
         callback_urls = self._get_callback_url_list(domains)
--- a/deployer/auth.py
+++ b/deployer/auth.py
@@ -43,11 +43,9 @@ class KeyProvider:
             for connection in self.auth0.connections.all()
         }
 
-    def create_client(self, name, domains, connection_name):
+    def create_client(self, name, domains):
         callbacks = self._get_callback_url_list(domains)
-        logout_urls = []
-        if connection_name == 'password':
-            logout_urls = self._get_allowed_logout_url_list(domains)
+        logout_urls = self._get_allowed_logout_url_list(domains)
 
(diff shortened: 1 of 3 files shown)


deployer/auth.py
```diff
@@ -43,13 +43,17 @@ def get_connections(self):
             for connection in self.auth0.connections.all()
         }
 
-    def create_client(self, name, domains):
+    def create_client(self, name, domains, connection_name):
         callbacks = self._get_callback_url_list(domains)
+        logout_urls = []
+        if connection_name == 'password':
```
GeorgianaElena, line 49: I noticed that if we define this `allowed_logout_url` list even if we're not using a password auth, it doesn't cause any harm. So, these checks aren't necessarily needed. I was thinking to remove them. What do you think?

yuvipanda: Yeah, doing it unconditionally is probably not a bad idea - we want everyone to log out regardless of connection type. Right now I think if you log out of our hub and log out of github as well, if you try to log in again you'll just immediately log in - no need to go via github. Making this unconditional should fix this.

yuvipanda:
Thanks for working on this, @GeorgianaElena!

Once this is deployed, I think we should explore:

1. Switching to the [auth0 authenticator](https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/auth0.py) instead of the Generic OAuthenticator
2. Adding the logout functionality to the upstream Auth0 authenticator.

What do you think?

GeorgianaElena:
> What do you think?

I think this would be best. We should open a separate issue for it to keep track of the progress.

Also, was there any other reasons for not going with the Auth0 OAuthenticator in the first place? I'm wondering if there's anything else we should have in mind when exploring switching from GenericOAuthenticator.

yuvipanda:
I think I didn't check to see if there was an auth0 specific one before using GenericOAuthenticator, so we should definitely switch :)

> We should open a separate issue for it to keep track of the progress.

Yeah, can you open one in the OAuthenticator repo?

GeorgianaElena:
> Yeah, can you open one in the OAuthenticator repo?

Sure! <3

Also, I deployed this to the staging hub, if you want to give it a try :D

yuvipanda:
@GeorgianaElena awesome :) Can you remove the conditional checks and deploy it on demo too?

GeorgianaElena:
Done!

Just to make sure: I should do the switch to Auth0 OAuthenticator as part of this PR?

deployer/auth.py
```diff
@@ -63,6 +65,15 @@ def _get_callback_url_list(self, domains):
 
         return [f'https://{domains}/hub/oauth_callback']
 
+    def _get_allowed_logout_url_list(self, domains):
+        if isinstance(domains, list):
+            logout_urls = []
```
yuvipanda, line 70: Since this is a farily simple for loop, can we make it a list comprehension instead?

deployer/auth.py
```diff
@@ -63,6 +65,15 @@ def _get_callback_url_list(self, domains):
...
+            for domain in domains:
+                logout_urls.append(f'https://{domain}/hub/')
+            return logout_urls
+
+        return [f'https://{domains}/hub/']
```
yuvipanda, line 75: Why the `/hub` prefix? Can you add a comment about what is being authorized here?

deployer/auth.py
```diff
@@ -150,4 +179,7 @@ def get_client_creds(self, client, connection_name):
             'scope': ['openid', 'name', 'profile', 'email']
         }
 
+        if connection_name == 'password':
```
yuvipanda, line 182: Is this different than for other type of connections?

hub-templates/basehub/values.yaml
```diff
@@ -271,9 +271,43 @@ jupyterhub:
...
+          Handle custom logout URLs. If a custom logout url
+          is specified, the 'logout' button will log the user out of that identity
+          provider in addition to clearing the session with Jupyterhub, otherwise
+          only the Jupyterhub session is cleared.
+          """
+
+          async def get(self):
+            await self.default_handle_logout()
+            await self.handle_logout()
+            if self.authenticator.logout_redirect_url:
```
yuvipanda, line 290: Let's do this conditional check inside render_logout_page instead? If all `super()` calls are only accesing the same function they are calling from, that makes life simpler I think.

yuvipanda (changes requested):
Left some stylistic comments but otherwise we can merge I think.

Moving to the auth0 authenticator should definitely be a different PR, doesn't need to block this one.

Thanks for working on this, @GeorgianaElena!

deployer/auth.py
GeorgianaElena: Right, we don't actually need the `/hub` part. Fixed and commented :D

deployer/auth.py
GeorgianaElena: If we want to logout everyone at the Auth0 session layer regardless of the connection type, then it shouldn't be different.
I just removed the condition. Ty! :D

GeorgianaElena:
Thanks, @yuvipanda :D I think I addressed all the comments!

yuvipanda:
Thanks, @GeorgianaElena! Let's move this upstream :)

choldgraf:
🎉

PR #430, opened 2021-05-25
Reduce memory request for carbonplan r5.8x instance

yuvipanda:
They apparently have less than 250G total allocatable space,
despite having 256G total RAM

Ref https://github.com/2i2c-org/pilot-hubs/issues/291


damianavila (approved):
This one LGTM.
This is already deployed, IIRC... am I right @yuvipanda?

yuvipanda:
Thanks, @damianavila. And yep, already deployed...

PR #434, opened 2021-05-27
Switch to Auth0OAuthenticator

GeorgianaElena:
This should work once we'll be using an `oauthenticator` version that has these changes: https://github.com/jupyterhub/oauthenticator/pull/439 and https://github.com/jupyterhub/oauthenticator/pull/437

Steps to have this PR ready for review:
- [x] Have the two `oauthenticator` PRs mentioned above, merged
- [x] Use OAuthenticator main branch in `pilot-hubs` hub image (https://github.com/2i2c-org/pilot-hubs/tree/master/hub-templates/images/hub)

### DIFF
--- a/deployer/auth.py
+++ b/deployer/auth.py
@@ -1,6 +1,8 @@
 from auth0.v3.authentication import GetToken
 from auth0.v3.management import Auth0
 
+import re
+
 # What key in the authenticated user's profile to use as hub username
 # This shouldn't be changeable by the user!
 USERNAME_KEYS = {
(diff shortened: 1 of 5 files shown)


yuvipanda:
@GeorgianaElena we already use a custom hub image - https://github.com/2i2c-org/pilot-hubs/tree/master/hub-templates/images/hub. You can just install master OAuthenticator there and move this forward, and make sure we move back to released version of OAuthenticator when it's released. IMO keeping a local fork and pushing things upstream is an important part of running infra like ours, since we shouldn't rust releases to match our schedule

GeorgianaElena:
Thanks @yuvipanda <3 ! I updated the top comment and switched to using the main branch of `OAuthenticator`. I also deployed the changes to staging.

yuvipanda:
Awesome, thanks @GeorgianaElena!

damianavila:
@GeorgianaElena this is great!

Just some tiny comment below...

> You can just install master OAuthenticator there and move this forward, and make sure we move back to released version of OAuthenticator when it's released. IMO keeping a local fork and pushing things upstream is an important part of running infra like ours, since we shouldn't rust releases to match our schedule

I agree we should not rush upstream releases and I agree a local fork is OK, but we should not be building the fork from master, IMHO, because you are introducing uncertainty from other changes that live in master and were not validated by the full test + release process. My suggestion, in these cases, is to fork from the last released version and apply your patch on top of it to reduce the uncertainty. It is a little bit more conservative approach that saved some a lot of time in the past 😜 .

yuvipanda:
@damianavila yeah, i generally prefer that for most cases! But I also consider myself (and us, in general) somewhat *as* upstream as well, so finding bugs in master is often a feature not a (hah) bug. But we should try lay out criteria on when to do what, I think.

damianavila:
> But we should try lay out criteria on when to do what, I think.

👍

PR #436, opened 2021-05-31
For reference: work towards adding JMTE hub (AWS, eksctl, cloudformation)

consideRatio:
For reference and transparency, this PR represents the state of my work to deploy hub.jupytearth.org using 2i2c configuration etc.

This is what remains to be done for this PR to be mergeable, as updated 2021-08-27.
- [x] #539
  I've configured the deployment to rely on automatic seeding of some secrets so they don't have to be managed by SOPS etc.
- [ ] https://github.com/2i2c-org/pilot-hubs/issues/502
  I've installed the `jupyterhub-ssh` Helm chart, but doing so required me modifying the daskhub Helm chart to install it for me. We need to make charts composable rather than have the current situation that instead have dependencies.
  - [ ] Allow the helm chart to optionally depend on jupyterhub-ssh as well
- [ ] #636
  https://github.com/2i2c-org/pilot-hubs/pull/436#issuecomment-852560099
  Using `eksctl`, we can't store a kubeconfig with credentials to access the cluster because they time out in ~1 day. Instead, we must make the deployment script be able to use some service account to acquire access.
- [ ] Either upstream the changes of how I've configured daskhub or make them specific to the JMTE deployment.
  https://github.com/2i2c-org/pilot-hubs/pull/436/files#diff-27ac079997ab718271acef5b6fef5940182968946a57f78458fef5822e26b9ec


eksctl/eksctl-cluster-config.yaml
```diff
@@ -0,0 +1,252 @@
...
+#
+# Choosing instance type?
+# - Maximum pods: https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt
+# - Node specs:   https://aws.amazon.com/ec2/instance-types/
+# - Cost:         https://ec2pricing.net/
+#
+# Management advice:
+# - Always use a suffix for node group names that you can replace with something
+#   to create a new node group and delete the old. You will run into issues if
+#   you name it "core" and "core-a" instead of "core-a" and "core-b", such as
```
yuvipanda, line 86: +1, good idea. This has been another advantage with kops though, you can update node-pools in place.

hub-templates/basehub/Chart.yaml
```diff
@@ -2,10 +2,9 @@ apiVersion: v2
...
 name: basehub
-# Let's keep this constant so other charts in this repo can depend on this easily
-version: 0.0.1-n569.hb296398
+version: "0.1.0"
 dependencies:
   - name: jupyterhub
-   # REMEMBER TO CHANGE BASE IMAGE OF images/hub/ WHEN CHANGING THIS
-    version: 0.11.1-n277.h61b0e003
+    # REMEMBER TO CHANGE BASE IMAGE OF images/hub/ WHEN CHANGING THIS
+    version: 1.0.0-beta.1
```
yuvipanda, line 9: Aha, amazing! I'm hoping we can extract this out to its own PR and merge it independently

hub-templates/daskhub/values.yaml
```diff
@@ -51,6 +50,9 @@ basehub:
       extraEnv:
         # The default worker image matches the singleuser image.
         DASK_GATEWAY__CLUSTER__OPTIONS__IMAGE: '{JUPYTER_IMAGE_SPEC}'
+        DASK_DISTRIBUTED__DASHBOARD_LINK: '/user/{JUPYTERHUB_USER}/proxy/{port}/status'
```
yuvipanda, line 53: Awesome - another thing I'm hoping we can extract and merge independently :)

yuvipanda (commented):
AMAZE!

yuvipanda:
@consideRatio if you'd like, we (or I?) can extract the parts of this PR that aren't related to JMTE and merge them separately.

consideRatio:
❤️ THANKS For your energy injection as usual @yuvipanda :) :tada:

Abolutely feel free to go for it! I didnt feel i had the time to prioritize any upstream PRs so i just just tried to make various PRs be commit like chunks.

yuvipanda:
I've https://github.com/2i2c-org/pilot-hubs/pull/439 which pulls out commits from here for v1.0-bea bump

deployer/hub.py
```diff
@@ -40,6 +40,8 @@ def auth(self):
             yield from self.auth_gcp()
         elif self.spec['provider'] == 'kubeconfig':
             yield from self.auth_kubeconfig()
+        elif self.spec['provider'] == 'none':
```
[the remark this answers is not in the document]
consideRatio: Yepp!

config/hubs/jmte.cluster.yaml
```diff
@@ -0,0 +1,137 @@
...
+              name: pangeo/pangeo-notebook
+              tag: "2021.05.15" # https://hub.docker.com/r/pangeo/pangeo-notebook/tags
+            profileList:
+              # The mem-guarantees are here so k8s doesn't schedule other pods
+              # on these nodes.
+              - display_name: "Small: 4 CPU and 16GB RAM (m5.xlarge)"
+                description: "This machine will be shared with up to four other users."
+                kubespawner_override:
+                  # Unset to avoid default values from basehub/values.yaml
+                  cpu_guarantee: 0.8
```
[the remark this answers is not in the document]
consideRatio: > ALso, if you don't specify memory limits one user can very easily OOM everyone else causing others' pods to die.

Oh, so it won't kill the main villain hogging all memory?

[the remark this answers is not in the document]
yuvipanda: hmm, I'm not exactly sure how the OOMKiller will work in this case, but my intuition is that if all the processes are running at the same OOM score, it's not clear which will be killed. However, it's also possible that one process immediately takes up all the RAM, and other processes get killed when they ask for tiny bits. Limits and guarantees make this easier to reason about as well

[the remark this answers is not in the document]
consideRatio: Ah hmmm, the score is influenced by requests/limits somehow i know. I don't remember where I read that though. I think its fair to assume it is more complicated than the one taking up most memory though

consideRatio:
I rebased on master that now includes some of the previous commits. I have no way of creating a kubeconfig that would be functional for more than ~1 day though so this is not something we can start deploying automatically yet.

This is related to https://github.com/2i2c-org/pilot-hubs/issues/381 but a more extreme situation where the ~30 day manual update of kubeconfig isn't possible.

eksctl/cloudformation-extras.yaml
```diff
@@ -0,0 +1,293 @@
...
+  #
+  S3Bucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      AccessControl: Private
+      BucketName: !Sub ${EksClusterName}-scratch
+
+  # ref: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecr-repository.html
+  #
+  EcrRepository:
```
damianavila, line 200: I have not seen any ECR reference in the config file you pushed above.
Are you planning to use the ECR repos later down the road?

damianavila: OK, I re-read the top comment you posted above just now... I should have read better 😉 .

eksctl/eksctl-cluster-config.yaml
```diff
@@ -0,0 +1,274 @@
...
+apiVersion: eksctl.io/v1alpha5
+kind: ClusterConfig
+metadata:
+  name: jmte
+  # region:
+  #   The region was chosen to to us-west-2 (Oregon) to be close to a CMIP-6
+  #   dataset.
+  #
+  region: us-west-2
+  version: "1.19"
```
damianavila, line 39: I think 1.20 is available from AWS, are you selecting 1.19 to be in sync with the rest of the pilots-hubs stuff, I guess, just curious :wink: ?

damianavila:
This is awesome work, Erik! Thanks for sharing it!

yuvipanda:
@consideRatio I extracted another individual commit from this - https://github.com/2i2c-org/pilot-hubs/pull/455

eksctl/eksctl-cluster-config.yaml
consideRatio: Sorry for super late response @damianavila! I'm not sure what I thought any more at that time, but I guess I'd cared to avoid piloting usage of the k8s manifests on a newer k8s version than they have previously been used on.

consideRatio:
### Rebased

I rebased this on current master before doing a redeploy, everything still works. I'm still running in a fork due to AWS permissions etc and to feel freedom to quickly adjust things based on needs.

consideRatio:
## Extending JMTE to use jupyterhub-ssh

If this was a custom deployment I'd manage myself, I'd have a meta helm chart and add another Helm chart to its dependency. I can do this here as well, but it would influence other Helm charts etc. An idea of what can be done is to use a dependency that is disabled by default via a condition set via values.

@2i2c-org/tech-team what do you think about me adding a opt-in Helm chart dependency to jupyterhub-ssh in the daskhub helm chart? Btw, it would also make sense to make the basehub helm chart have a opt-in dependency on the dask-gateway helm chart and just have a single meta chart instead of multiple.

### Reference
- A practical example of using [opt in dependencies](https://github.com/neurohackademy/nh2021-jupyterhub/blob/c7fab6574e9986ec9486f180e4c5eb725c157522/chart/Chart.yaml#L17-L37) that are enabled [via passed values](https://github.com/neurohackademy/nh2021-jupyterhub/blob/c7fab6574e9986ec9486f180e4c5eb725c157522/chart/values.yaml#L12-L18).
- [Helm documentation](https://helm.sh/docs/chart_best_practices/dependencies/#conditions-and-tags) about `tags` (enable/disable multiple charts dependencies with one value) and `condition` (enable/disable single chart dependency with one value).

choldgraf:
@consideRatio I'm not quite sure what are the implications of your suggestion...it also feels related to #502 right? Should we discuss and converge there? Or perhaps bring this up in a team meeting to brainstorm?

consideRatio:
> Should we discuss and converge there? Or perhaps bring this up in a team meeting to brainstorm?

Yes to discuss and converge there and yes to bringing this up in a team meeting to brainstorm! :)

choldgraf:
For sure - this feels very relevant to ones is the points that we discussed in the Pangeo collaboration meeting as well around allowing binder functionality to be a feature flag of jupyterhub https://discourse.pangeo.io/t/notes-from-the-pangeo-2i2c-kick-off-meeting/1587

As well as that team compass conversation in jupyterhub about multi application deployments

hub-templates/daskhub/values.yaml
yuvipanda: Even without this, I see dask-dashboard is functional, and does not use jupyter-server-proxy 
image

Is this necessary for dask-gateway? The scheduler (and hence the dashboard process) lives in another pod anyway.

consideRatio: Im generally unsure about this config, it is a migrated pattern so probably no longer relevant like this.

choldgraf:
As this is an ongoing PR that we aren't actively trying to merge soon, could we convert this to `draft mode`, so that it's clear it's something we'll use as reference over time?

yuvipanda:
This has been subsumed into our infrastructure now (https://github.com/2i2c-org/infrastructure/issues/2201), except for the SFTP part (#2208). With gratitude to @consideRatio, I close this PR!

PR #437, opened 2021-05-31
Use yarl instead of tornado

GeorgianaElena:
Should fix the currently [failing GitHub actions](https://github.com/2i2c-org/pilot-hubs/runs/2709666249).


yuvipanda:
Can we use https://pypi.org/project/yarl/ instead? I think it's already present as a dependency of aiohttp

GeorgianaElena:
Done. Thanks, @yuvipanda!

PR #438, opened 2021-05-31
Move paleohack back to 2i2c cluster

yuvipanda:
- Moved NFS home directories - https://github.com/2i2c-org/pilot-hubs/issues/409#issuecomment-851415247
- DNS needs to be switched to the new cluster
- Kept CPU & Memory limits, but dropped guarantees. Helps increase
  utilization on the cluster, since we expect the hub to be rarely
  used.

Fixes https://github.com/2i2c-org/pilot-hubs/issues/409


yuvipanda:
Need to move the sqlite database as well

GeorgianaElena (approved):
This looks good to me :rocket:

I'm just wondering, do we have any documentation around moving hubs out an existing cluster?

Same for moving them back and what are the pieces that need to be permanently deleted and what needs to be kept. @yuvipanda, maybe you can give more details about this comment specifically `*I move the old contents of this directory to _old - this is ok, since the users will now see the same contents.*` from https://github.com/2i2c-org/pilot-hubs/issues/409#issuecomment-851415247

I guess what I'm saying is that when doing operations like moving a hub from a cluster to another, there are a few "invisible" steps from a PR's point of view, that are potentially only explained in comments (like [this one](https://github.com/2i2c-org/pilot-hubs/issues/409#issuecomment-851415247)) or commit messages. I would love if we get some of these steps into the docs, even if it's in the form of some generic points (I imagine there's no silver bullet that fits every scenario)

damianavila (approved):

damianavila:
> I guess what I'm saying is that when doing operations like moving a hub from a cluster to another, there are a few "invisible" steps from a PR's point of view, that are potentially only explained in comments (like this one) or commit messages. I would love if we get some of these steps into the docs, even if it's in the form of some generic points (I imagine there's no silver bullet that fits every scenario)

This is a very good point, @GeorgianaElena. If not in this PR, we should have a follow-up task to document the process.

yuvipanda:
@damianavila @GeorgianaElena I added docs to this PR :) LMK if that looks ok, and I'll complete the move?

GeorgianaElena:
It's perfect @yuvipanda, thank you! Adding the docs shouldn't have been a blocker for this PR though, sorry for not following up on this!

damianavila:
Thanks for adding the docs @yuvipanda!

PR #439, opened 2021-05-31
Bump z2jh to v1.0 beta

yuvipanda:
- Move `singleuser.admin` to `custom.singleuserAdmin` to satisfy the helm
  JSON schema, which doesn't allow extra properties under `singleuser`
- Bump our hub base image to inherit from the v1.0-beta1 image
- Merge `requirements.yaml` file into `Chart.yaml` files - with helm3 these 
   do not need to be together
- Add couple more things to .gitignore

Splitting up code from https://github.com/2i2c-org/pilot-hubs/pull/429 that
works on just upgrading code to z2jh 1.0-beta1

Fixes https://github.com/2i2c-org/pilot-hubs/issues/414


consideRatio (approved):
Thanks @yuvipanda this LGTM!

I saw that you have pushed the k8s-hub replacement image as well. Merge?

hub-templates/basehub/values.yaml
```diff
@@ -67,8 +72,6 @@ jupyterhub:
     hook:
       enabled: false
   proxy:
-    https:
-      enabled: false
```
damianavila, line 71: I guess this is now `false` by default?

hub-templates/daskhub/values.yaml
```diff
@@ -41,7 +41,6 @@ basehub:
               - port: 22
                 protocol: TCP
       cloudMetadata:
-        enabled: true
```
damianavila, line 44: And this is now `true` by default, correct?

hub-templates/basehub/values.yaml
consideRatio: Yepp!

hub-templates/daskhub/values.yaml
consideRatio: No it renamed to the other field below called blockWithIpTables or something, and inverted its value. Having this set leads to an error as the helm charts forces a migration. There is no actual change in outcome by this change now.

hub-templates/basehub/values.yaml
```diff
@@ -29,10 +29,15 @@ nfsPVC:
...
 jupyterhub:
   custom:
+    singleuserAdmin:
+      extraVolumeMounts:
+        - name: home
+          mountPath: /home/jovyan/shared-readwrite
+          subPath: _shared
     cloudResources:
-      provider: null
+      provider:
```
GeorgianaElena, line 38: Is there any difference between leaving this black and not setting them at all? Or is this a way to unset default values?

hub-templates/basehub/values.yaml
yuvipanda: Yeah but it was also repeated down in the same file

hub-templates/basehub/values.yaml
yuvipanda: ruamel.yaml formatted it this way, I assume it's equivalent?

hub-templates/basehub/values.yaml
damianavila: Gotcha!

hub-templates/daskhub/values.yaml
damianavila: OK, thanks for the additional info!

damianavila (approved):
My questions were answered, this LGTM now!

yuvipanda:
Thanks, @damianavila!

PR #445, opened 2021-06-02
Add docs service api token

GeorgianaElena:
I couldn't identify exactly what changed with the new z2jh to v1.0 beta, but it seems that we need to specify an api token for the docs service now. When we first added the docs service, we concluded that because the service didn't make any API calls, then we didn't need a token ([ref](https://github.com/2i2c-org/pilot-hubs/pull/177#discussion_r564686610))

### DIFF
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -294,9 +294,11 @@ class Hub:
 
         }
 
+        docs_token = hmac.new(secret_key, f'docs-{self.spec["name"]}'.encode(), hashlib.sha256).hexdigest()
         if 'docs_service' in self.spec['config'].keys() and self.spec['config']['docs_service']['enabled']:
             generated_config['jupyterhub']['hub']['services']['docs'] = {
-                'url': f'http://docs-service.{self.spec["name"]}'
+                'url': f'http://docs-service.{self.spec["name"]}',
+                'apiToken': docs_token
             }
 
 


yuvipanda:
@consideRatio do you know what might be causing this?

yuvipanda:
Merging to unbreak CI.

consideRatio:
@yuvipanda @GeorgianaElena hmmm I don't know... Hmmm...

I have touched this logic in z2jh recently I recall so there may be some change there, but it may also be a change in JupyterHub itself.

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/30b3647591af121dc46ddc08a4c517a530f11eee/jupyterhub/files/hub/jupyterhub_config.py#L360-L372

Hmmm... Could it be that it is required to be set explicitly to None, and with my change, it isn't set at all?

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2042/files#diff-22223d4c6126ae9811f49f6477a770d372019099e5fc613aeb7dcf2b5b5a5323R368-R380

consideRatio:
> I couldn't identify exactly what changed with the new z2jh to v1.0 beta, but it seems that we need to specify an api token for the docs service now. 

I'd like to get a hold of the error shown again, I found this failure, but I understand we don't emit any logs after this automatically and we rely on manual inspection at the moment.

https://github.com/2i2c-org/pilot-hubs/runs/2720750369?check_suite_focus=true

consideRatio:
Hmmm hmmm no I'm lost. I don't understand what if anything changed of relevance. Reproduction or clear logs about the error is needed.

Note that if the deployer script fails, we could given access credentials for k8s setup, consider using https://github.com/jupyterhub/action-k8s-namespace-report to get information on failures.

PR #446, opened 2021-06-02
Emit array from openscapes.jsonnet

yuvipanda:
- `jsonnet -y` expects an array to emit as multiple documents.
  It unfortunately ends with a `...` at the end (since it is a YAML
  stream) which has to be manually removed before we can pass it to
  kops.
- -m <dirname> can emit different files in a directory, and we can
  emit one file each for cluster, master, etc. The code in this
  PR didn't actually do that either, it was just invalid YAML.

Right now, I prefer the 1 file format, so sticking to the format
that works with `-y`.

### DIFF
--- a/kops/openscapes.jsonnet
+++ b/kops/openscapes.jsonnet
@@ -90,7 +90,7 @@ local data = {
     ]
 };
 
-{ 'kops.yaml': [
+[
     data.cluster,
     data.master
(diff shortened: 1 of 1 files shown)


damianavila (approved):
> Right now, I prefer the 1 file format, so sticking to the format
that works with -y.

Makes sense to me, this is going in the same direction I was thinking when I found this one.

PR #448, opened 2021-06-02
Fix logout handling - transition to a new naming scheme on registered OAuth2 application in Auth0

### LINKED ISSUE #447: Logout on JMTE redirected to Grenoble
# Summary

Logging out from JupyterHub on hub.jupytearth.org redirects users to the wrong location in the end.

This is a redirection chain observed.

1. https://hub.jupytearth.org/hub/logout
2. https://2i2c.us.auth0.com/v2/logout?client_id=JptUUZT5RgWmjsLGYvcVS4CWvLsVptYt&returnTo=https://grenoble.2i2c.cloud
3. https://grenoble.2i2c.cloud

I observe that the client_id relates to the `prod` application rather than the `jmte` application that I registered, and that `grenoble` was the first logout url listed among the allowed logout urls. This is probably related. Is the deployer script not using the `jmte` application for logout stuff, only during login stuff?

It seems like the following code from the `class Hub` definition is using the `name` entry on the listed Hub. I guess I expected it to use the cluster name with regards to authentication rather then the hub name within the cluster.

```yaml
# This name is used during /hub/login
name: jmte
provider: none
hubs:
  # This name is used during /hub/logout
  - name: prod
    # ...
```

https://github.com/2i2c-org/pilot-hubs/blob/f68da3c2e65da1c8ae6df95aa32e4d9d7d897fd6/deployer/hub.py#L250-L251

consideRatio:
This closes #447.

Previously we were using a `<clustername>` format for the name of the Auth0 registered OAuth2 application, but the logout action assumed a `<hubname>` format. It seems we have consensus on opting for a `<clustername>-<hubname>` format at this point so this PR is about transitioning to that.

## Action points pre-merge

- [x] Decide on the new format `clutsername-hubname` (DECIDED ON THIS)
- [x] Alternatively, decide it is preferable with more explicit configuration.
    I imagine an option to assuming a `clusternane-hubname` pattern, that a more explicit config option mentioning the Auth0 registered OAuth2 application name could be less magic and make it more clear what's going on.
    ```yaml
    hubs:
      - name: staging
        domain: staging.carbonplan.2i2c.cloud
        template: daskhub
        auth0:
          connection: github
          application_name: carbonplan-staging
    ```
- [x] Ensure we reference the new OAuth2 application also on /hub/logout
- [x] Ensure we reference the new OAuth2 application format on /hub/login
- [x] Ensure we don't run into issues based on the logic discussed here: https://github.com/2i2c-org/pilot-hubs/pull/448#discussion_r644910435
- [x] In Auth0: register new OAuth2 applications following the new format for all hubs we have
  Update: this is not needed, they are automatically created.

## Action points post-merge

- [ ] Verify successful deployment, what needs testing is the login and logout action
- [ ] In Auth0: delete the old OAuth2 applications following the old format for all hubs we have

### DIFF
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -246,9 +246,8 @@ class Hub:
         #
         # Allow explicilty ignoring auth0 setup
         if self.spec['auth0'].get('enabled', True):
-
             client = auth_provider.ensure_client(
-                self.spec['name'],
+                self.cluster.spec['name'],
                 self.spec['domain'],
                 self.spec['auth0']['connection']
             )
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -247,7 +247,7 @@ class Hub:
         # Allow explicilty ignoring auth0 setup
         if self.spec['auth0'].get('enabled', True):
             client = auth_provider.ensure_client(
-                self.cluster.spec['name'],
+                f"{self.cluster.spec['name']}-{self.spec['name']}",
                 self.spec['domain'],
                 self.spec['auth0']['connection']
             )
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -247,9 +247,10 @@ class Hub:
         # Allow explicilty ignoring auth0 setup
         if self.spec['auth0'].get('enabled', True):
             client = auth_provider.ensure_client(
-                f"{self.cluster.spec['name']}-{self.spec['name']}",
-                self.spec['domain'],
-                self.spec['auth0']['connection']
+                name=               self.spec['auth0'].get('application_name', f"{self.cluster.spec['name']}-{self.spec['name']}"),
+                domains=            self.spec['domain'],
+                connection_name=    self.spec['auth0']['connection'],
+                connection_config=  self.spec['auth0'].get(self.spec['auth0']['connection'], {}),
             )
             # FIXME: We're hardcoding GenericOAuthenticator here
             # We should *not*. We need dictionary merging in code, so
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -247,10 +247,10 @@ class Hub:
         # Allow explicilty ignoring auth0 setup
         if self.spec['auth0'].get('enabled', True):
             client = auth_provider.ensure_client(
-                name=               self.spec['auth0'].get('application_name', f"{self.cluster.spec['name']}-{self.spec['name']}"),
-                domains=            self.spec['domain'],
-                connection_name=    self.spec['auth0']['connection'],
-                connection_config=  self.spec['auth0'].get(self.spec['auth0']['connection'], {}),
+                name=self.spec['auth0'].get('application_name', f"{self.cluster.spec['name']}-{self.spec['name']}"),
+                domains=self.spec['domain'],
+                connection_name=self.spec['auth0']['connection'],
+                connection_config=self.spec['auth0'].get(self.spec['auth0']['connection'], {}),
             )
             # FIXME: We're hardcoding GenericOAuthenticator here
             # We should *not*. We need dictionary merging in code, so
--- a/config/hubs/cloudbank.cluster.yaml
+++ b/config/hubs/cloudbank.cluster.yaml
@@ -192,6 +192,8 @@ hubs:
     template: basehub
     auth0:
       connection: password
+      password:
+        database_name: database-demo
     config:
       jupyterhub:
         custom:
--- a/config/hubs/schema.yaml
+++ b/config/hubs/schema.yaml
@@ -1,4 +1,4 @@
-$schema: 'http://json-schema.org/draft-04/schema#'
+$schema: 'http://json-schema.org/draft-07/schema#'
 type: object
 additionalProperties: false
 properties:
--- a/config/hubs/schema.yaml
+++ b/config/hubs/schema.yaml
@@ -126,6 +126,28 @@ properties:
                   Authentication method users of the hub can use to log in to the hub.
                   We support a subset of the [connectors](https://auth0.com/docs/identityproviders)
                   that auth0 supports
+              application_name:
+                type: string
+                description: |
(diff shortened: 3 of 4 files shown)


deployer/hub.py
```diff
@@ -246,9 +246,8 @@ def get_generated_config(self, auth_provider: KeyProvider, secret_key):
         #
         # Allow explicilty ignoring auth0 setup
         if self.spec['auth0'].get('enabled', True):
-
             client = auth_provider.ensure_client(
-                self.spec['name'],
+                self.cluster.spec['name'],
```
yuvipanda, line 250:
```suggestion
                f"{self.cluster.spec['name']}-{self.spec['name']}",
```

yuvipanda: Are there other places this change needs to happen?

consideRatio: I don't understand your suggested change. When I look in the auth0 client list, I see no indication that the clients are named like this and therefore only created a single client called `jmte` rather than `jmte-staging` or similarly.

Do you suggest we start using `<clustername>-<hubname>` instead, and update our auth0 OAuth2 applications?

consideRatio: > Are there other places this change needs to happen?

I think there are if we accept your commit suggestion as /hub/login has correctly used the correct client_id even though the /hub/logout didn't - which this was fixing.

GeorgianaElena: Just a side note: I think that if we go with `<clustername>` and have a connection per-cluster, then `redirectTo` should pick the right, hub-specific logout url, rather than the first one in the list:
https://github.com/2i2c-org/pilot-hubs/blob/27be2de984009da7733832ed188bc6e0e945328d/deployer/auth.py#L166

consideRatio:
I updated the PR description. A decision needs to be made, and then the implementation can follow.

yuvipanda:
https://auth0.com/docs/api/management/v2#!/Clients/post_clients has a bunch of other options we can set for each client. Maybe we can allow setting name and description under `auth0`, with `name` defaulting to `<clustername>-<hubname>`?

consideRatio:
> Maybe we can allow setting name and description under auth0, with name defaulting to <clustername>-<hubname>?

Would this be a benefit? I figure being explicit can help readability and reduce implementation logic's complexity, allowing for both would increase the complexity from either option.

yuvipanda:
> Would this be a benefit?

Right, I think it's ok to not allow name to be overrideable. Description being overrideable might be useful (it's shown in the login screen) though. Not necessary for this PR!

consideRatio:
> Right, I think it's ok to not allow name to be overrideable. Description being overrideable might be useful (it's shown in the login screen) though. Not necessary for this PR!

Oh I think i misunderstood you entirely, okay yes sure I get what you are saying now.

What do you suggest? Default to `clustername-hubname`, not let that be configurable explicitly, but in the future let description etc be configurable?

What do you think about instead of having a magic default that must be documented somewhere, we have config like this now that can be extended in the future?

```
hubs:
  - name: staging
    domain: staging.carbonplan.2i2c.cloud
    template: daskhub
    auth0:
      connection: github
      application_name: carbonplan-staging
      # application_description: bla bla
```

Or wait? Is the auth0 application created automatically? I created one manually. If it is created automatically, I'm more inclined to appreciate a naming scheme being assumed.

yuvipanda:
> What do you suggest? Default to `clustername-hubname`, not let that be configurable explicitly, but in the future let description etc be configurable?

Way to go in this PR, I think.

yuvipanda:
> Or wait? Is the auth0 application created automatically? I created one manually. If it is created automatically, I'm more inclined to appreciate a naming scheme being assumed.

It is created automatically! And maintained automatically - see https://github.com/2i2c-org/pilot-hubs/blob/master/deployer/auth.py. If you delete the on you created manually and run the deployer, it should auto-create it. Auto-creation is why  I'm suggesting the magic name :)

damianavila:
> Auto-creation is why I'm suggesting the magic name :)

I agree with the magic name as well ;-)

yuvipanda (approved):
Ok, so this LGTM. Would you be able to deploy and test this, @consideRatio?

consideRatio:
> Ensure we reference the new OAuth2 application format on /hub/login

I still guess we will have a /hub/login failure so I don't want to merge before checking for that. I need to do some errands now but I'll come back to this.

deployer/hub.py
consideRatio: @yuvipanda it is perhaps out of scope for this PR to also address this, but I figure this is a bug if many hubs use the same oauth2 application at least. Hmmm... But by going for a hub specific oauth2 application, this won't be an issue...

Okay! All good!

yuvipanda:
> I guess we will have a /hub/login vs /hub/logout discrepancy with regards to what oauth2 applications we use, so I don't want to merge before checking for that.

Can you tell me what you mean by this? I'm not sure I understand

consideRatio:
> > I guess we will have a /hub/login vs /hub/logout discrepancy with regards to what oauth2 applications we use, so I don't want to merge before checking for that.
> 
> Can you tell me what you mean by this? I'm not sure I understand

Hmmm... Well I tried following up in the code, but failed to see what I expected.

I believed that there was evidence that my hub had consumed two different Oauth2 applications, something i recall based on inspecting `client_id` in requests made if i recall correctly. To me, that was an indication that updating the logic only at one place would only change one oauth2 client being used or similarly.

I think I got the evidence to support that understanding wrong somehow.

consideRatio:
I still don't feel comfortable merging this myself as I don't understand the deployment sequence. What needs to be done if a deployment fails in the sequence of updates made etc? Perhaps we can do this together tomorrow?

yuvipanda (approved):
This LGTM! The test failure is because the whole internet is melting (pypi is down), so let's do it

PR #450, opened 2021-06-04
Fix typo on openscapes jsonnet

yuvipanda:
This didn't have any effect on the deployed cluster. I
verified this with:

1. jsonnet openscapes.jsonnet -y > openscapes.kops.yaml
2. Remove `...` at end of the file
3. kops replace -f openscapes.kops.yaml
4. kops rolling-update cluster. This has no effects, so it was already
   the correct setup.

### DIFF
--- a/kops/openscapes.jsonnet
+++ b/kops/openscapes.jsonnet
@@ -27,7 +27,7 @@ local data = {
             name: "master"
         },
         spec+: {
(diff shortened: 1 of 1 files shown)


damianavila:
> This has no effects, so it was already the correct setup.

Great, approving and merging now.

damianavila (approved):

PR #456, opened 2021-06-08
Provision per-cluster support components

yuvipanda:
- Set up cert-manager explicitly, since it can not be set up as
  a dependent chart of our 'support' chart.
- 'support' chart deploys prometheus, grafana and nginx-ingress
  in a standard configuration.
- Allow per-cluster overrides with config under support/clusters.
- Automates auth to cluster for automated deployment
- Puts config overrides in same place as cluster config, to
  avoid duplication
- Sets up HTTPS for the 2i2c pilot hubs grafana

Ref https://github.com/2i2c-org/pilot-hubs/issues/388

TODO
- [x] Setup auth for grafana
- [x] Deploy with CI / CD

### DIFF
--- a/.github/actions/deploy/action.yml
+++ b/.github/actions/deploy/action.yml
@@ -27,6 +27,10 @@ runs:
         python3 -m pip install -r requirements.txt
         python3 -m pip install -r dev-requirements.txt
       shell: bash
+    - name: Deploy support components
+      run: |
+        python3 deployer deploy-support ${{ inputs.cluster }}
+      shell: bash
     - name: Deploy and test hubs
       run: |
         python3 deployer deploy ${{ inputs.cluster }}
--- a/.sops.yaml
+++ b/.sops.yaml
@@ -5,3 +5,5 @@ creation_rules:
     gcp_kms: projects/two-eye-two-see/locations/global/keyRings/sops-keys/cryptoKeys/similar-hubs
   - path_regex: config/secrets.yaml$
     gcp_kms: projects/two-eye-two-see/locations/global/keyRings/sops-keys/cryptoKeys/similar-hubs
+  - path_regex: support/secrets.yaml$
+    gcp_kms: projects/two-eye-two-see/locations/global/keyRings/sops-keys/cryptoKeys/similar-hubs
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -8,6 +8,7 @@ import subprocess
 import tempfile
 from contextlib import contextmanager, redirect_stderr, redirect_stdout
 from textwrap import dedent
+from pathlib import Path
 
 import pytest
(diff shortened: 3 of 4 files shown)


support/values.yaml
```diff
@@ -15,18 +15,20 @@ prometheus:
...
     enabled: false
-  rbac:
-    create: true
   server:
     resources:
       # Without this, prometheus can easily starve users
       requests:
         cpu: 0.2
-        memory: 768Mi
+        memory: 512Mi
```
damianavila, line 31: Can you comment on this adjustment?

deployer/__main__.py
```diff
@@ -97,6 +114,7 @@ def main():
     build_parser = subparsers.add_parser("build")
     deploy_parser = subparsers.add_parser("deploy")
     validate_parser = subparsers.add_parser("validate")
+    deploy_support_parser = subparsers.add_parser("deploy-support")
```
damianavila, line 117: Are you envisioning some cases where we do not deploy-support by default?

support/values.yaml
yuvipanda: Right now at least, these are all just estimates. The hope is that we can measure resource use of these and adjust accordingly.

deployer/__main__.py
damianavila: Ping 😉

yuvipanda: @damianavila I think for small clusters like farallon, maybe. Although experience with carbonplan suggests otherwise.

I made it a separate command primarily because `deploy` can be used to deploy individual hubs as well, not just an entire cluster. I've added this as a CD step now, and we should be able to deploy this to all hubs that are continuously deployed.

sgibson91 (approved):
This looks great to me!

yuvipanda:
Copying testing instructions from https://github.com/2i2c-org/pilot-hubs/pull/533#issuecomment-883290848:

1. Add the 'support' key in your `cluster.yaml`, and change grafana DNS there accordingly
2. Run `python3 deployer deploy-support <cluster-name>`
3. Once it's deployed, look at the external ip of `k -n support get svc support-ingress-nginx-controller`, and make a DNS entry for the grafana DNS you picked in (1)
4. The grafana username is admin, password is present in `support/secrets.yaml` (sops encrypted)
5. Once done, try to deploy github.com/jupyterhub/jupyterhub-grafana to the grafana so we have useful dashboards
6. Once this all works, let's document this

damianavila:
For future readers, this is not the complete PR as it is described in the first message.
The complement lives actually as part of another @yuvipanda's PR: https://github.com/2i2c-org/pilot-hubs/pull/429.

Btw, @sgibson91, I think you will find that historical context useful as well 😉 .

PR #429, opened 2021-05-25
Refactor GCP terraform code + add  MOEM-IGE hub

yuvipanda:
- Setup the cluster with the [terraform google provider][1], instead
  of the higher level [gke module][2]. The code gets simpler, and
  makes more terraform features (like for_each) accessible more easily.
- Allow multiple notebook and dask nodepools to be set up. Most
  research hubs want 2-3 options of notebook sizes to optimize for
  spend. I attempted to use [gke node autoprovisioning][3] instead of
  requiring manual nodepool provisioning, but it consistently
  provisioned nodes bigger than required. We should re-evaluate it
  later.
- Expose the GCP SA used by the k8s nodes to the user pods. A highly
  restricted SA is used for this, to prevent damage as much as possible.
  Users can then make requests to GCS buckets in other
  projects on behalf of this project.
- Dask Nodepools will default to matching sizes of the notebook
  nodepools. Can be overriden if necessary.
- Split terraform code into multiple files for easier maintenance
- Move tfvars files into a subdirectory and split terraform code
  into multiple files for easier maintenance
- Remove unused terraform variables
- Add some inline terraform docs
- Setup MOEM-IGE cluster + hub with new terraform code
- traefik tag bump was required to get LE working. It's
  already bumped in newer z2jh versions
- NFS server was again set up manually, and needed the `insecure`
  flag - even though other hubs are setup the same way and didn't
  need this. NFS situation needs to be sorted.

Ref https://github.com/2i2c-org/pilot-hubs/issues/207

[1]: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster
[2]: https://registry.terraform.io/modules/terraform-google-modules/kubernetes-engine/google/latest
[3]: https://cloud.google.com/kubernetes-engine/docs/how-to/node-auto-provisioning

TODO:

- [x] Move Paleohack hub back into this repo, remove the paleohack terraform setup https://github.com/2i2c-org/pilot-hubs/issues/409 https://github.com/2i2c-org/pilot-hubs/pull/438
- [x] Migrate pilot hubs to new cluster structure
- [x] Migrate cloudbank hubs to new cluster structure
- [x] Document decisions made better
- <s>[ ] Add a template for new clusters</s>  (another PR)

PR #457, opened 2021-06-08
Move pilot-hubs to new cluster

yuvipanda:
- Setup core, user and dask worker pods to match current
  configuration
- User and dask worker pods will switch to e2- instances rather
  than n1- instances - our usage is not constant enough to get
  the esustained use discounts
- The cluster is renamed from 'low-touch-hubs' to 'pilot-hubs'
- Point to new docker registry - an image has been built and
  pushed there.


yuvipanda:
Builds on deployment of #429. New pilot-hubs cluster has been deployed as well

Issue #454, opened 2021-06-08
AVC - avc.cloudbank.2i2c.cloud


ericvd-ucb:
# Background

Community College in Lancaster CA

# Setup Information

- **Hub auth type**: 
What is our plan for Microsoft campuses - just get them to make google accounts?

- **Hub administrators**: 
rbiritwum@avc.edu
James Brownlow <jbrownlow@avc.edu>,
William Kitto <wkitto1@avc.edu>,

- **Hub url**: 
rbiritwum@avc.edu
- **Hub logo**: 
- **Hub logo URL**: 
- **Hub type**: 
education - community college - data 8 
- **Hub cluster**: 
- **Hub image**: 

# Important Information

- **Link to leads issue**: 
- **Hub config name**: 
- **Community champion**: 
@sean-morris 
- **Hub start date**: 
- **Hub end date**: 
- **Hub important dates**: 

# Deploy To Do

- [ ] Initial Hub deployment
- [ ] Administrators able to log on
- [ ] Community Champion satisfied with hub environment
- [ ] Hub now in steady-state

yuvipanda:
@ericvd-ucb @sean-morris we could have them make username / passwords instead of using google auth?

PR #458, opened 2021-06-08
Added AVC(Antelope Valley) to hubs

sean-morris:
I added the avc.cloudbank.212c.cloud hub to the cloudbank.clusters file
using our standard format as well as google-oauth2

Resolves: #454

### DIFF
--- a/config/hubs/cloudbank.cluster.yaml
+++ b/config/hubs/cloudbank.cluster.yaml
@@ -443,3 +443,41 @@ hubs:
                 - christiaan.desmond@sjcc.edu
                 - sanjay.dorairaj@sjcc.edu
               admin_users: *sjcc_users
+  - name: avc
+    domain:
+      - avc.cloudbank.2i2c.cloud
+    template: basehub
+    auth0:
+      connection: google-oauth2
+    config:
+      jupyterhub:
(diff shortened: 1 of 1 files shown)


config/hubs/cloudbank.cluster.yaml
```diff
@@ -443,3 +443,41 @@ hubs:
                 - christiaan.desmond@sjcc.edu
                 - sanjay.dorairaj@sjcc.edu
               admin_users: *sjcc_users
+  - name: avc
+    domain:
+      - avc.cloudbank.2i2c.cloud
+    template: basehub
+    auth0:
+      connection: google-oauth2
```
damianavila, line 451: There was some comment in Slack from @yuvipanda about maybe using Auth password here instead of Google, IIRC.

yuvipanda:
https://github.com/2i2c-org/pilot-hubs/pull/462 moves cloudbank over to new cluster, needs to be merged before this can be.

config/hubs/cloudbank.cluster.yaml
yuvipanda: @ericvd-ucb let me know that avc is on google so this should be ok.

PR #460, opened 2021-06-09
Remove support for more than 1 domain per hub

yuvipanda:
Multiple domains were supported primarily to make sure hubs
are reachable by both datahub.mills.edu and mills.cloudbank.2i2c.cloud.
In practice, nobody uses the second URL, and this just confuses
the deployer code. datahub.mills.edu is a CNAME to
mills.cloudbank.2i2c.cloud - but that doesn't require the latter
to actually serve any content (over HTTP). CNAME is a DNS feature,
so we don't actually have to do anything wrt HTTP serving.

Includes slight refactoring to make auth.KeyProvider a little more
agnostic to just using with JupyterHubs - could perhaps be used
with https://github.com/2i2c-org/pilot-hubs/pull/456 in providing
OAuth credentials for grafana too

### DIFF
--- a/deployer/auth.py
+++ b/deployer/auth.py
@@ -46,70 +46,56 @@ class KeyProvider:
             for connection in self.auth0.connections.all()
         }
 
-    def create_client(self, name, domains):
-        callbacks = self._get_callback_url_list(domains)
-        logout_urls = self._get_allowed_logout_url_list(domains)
-
+    def create_client(self, name, callback_url, logout_url):
         client = {
             'name': name,
             'app_type': 'regular_web',
-            'callbacks': callbacks,
-            'allowed_logout_urls': logout_urls
+            'callbacks': [callback_url],
+            'allowed_logout_urls': [logout_url]
         }
         created_client = self.auth0.clients.create(client)
         return created_client
 
-    def _get_callback_url_list(self, domains):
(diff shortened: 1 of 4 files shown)


yuvipanda:
Ping to @GeorgianaElena since she worked on this first. Ping to @consideRatio since we talked about this during a video call yesterday

deployer/auth.py
```diff
@@ -46,70 +46,56 @@ def get_connections(self):
...
-        if missing_logout_urls:
+    def _ensure_client_logout_url(self, client, logout_url):
+        if client['allowed_logout_urls'] != [logout_url]:
             self.auth0.clients.update(
                 client['client_id'],
                 {
-                    # Don't remove other logout URLs
-                    'allowed_logout_urls': client.get('allowed_logout_urls', []) + missing_logout_urls
+                    # Overwrite any other logout URL - users should only land on
+                    # the hug home page after logging out.
```
damianavila, line 85: `hug`? I like that home page :wink:

damianavila (approved):
+ 1 on the simplification!

deployer/auth.py
yuvipanda, line 85:
```suggestion
                    # the hub home page after logging out.
```

yuvipanda:
Thanks, @damianavila!

Issue #453, opened 2021-06-07
Document our current AWS infrastrcuture


damianavila:
# Summary

We have been working on our new AWS infrastructure setup through several issues and PRs.
A summary of the work performed is outlined in #366.

The last step to close that ticket is documenting our current setup.
Replication and validation of the current process are needed before documenting the setup.

damianavila:
#461 (containing the real changes) was merged and we have a draft PR with docs at #467.

damianavila:
The draft PR was moved into a ready PR, reviewed, and finally merged.
We can close this one.

PR #461, opened 2021-06-09
Update the Farallon cluster to be compatible with latest AWS deployment story

damianavila:
These are the resulting files after replicating and validating the process to deploy a new AWS cluster with our latest AWS deployment story. This was a prerequisite for #453.

The ride was bumpy because I originally had a jsonnet specification referring to an inexistent instance (m5.medium).
After fixing that problem, I faced a new one because I was referring to an old label scheme we recently changed which was preventing the hub and proxy pod to be started.
Finally, that same problem hit me again when I tried to launch a new user server because now the cluster config file was still having old references to the old pools.

You can get more details from the commit messages, hopefully 😉  .

I will follow up this one with a new PR with the documentation I am writing from the notes I have taken so others do not need to drive bumpy again :wink: .


yuvipanda (approved):
I LIKE THIS!

PR #463, opened 2021-06-10
Fix schema definition for cluster config

yuvipanda:
- Follow-up to https://github.com/2i2c-org/pilot-hubs/pull/458 -
  domain can not be list anymore with https://github.com/2i2c-org/pilot-hubs/pull/460
- Support components config for each cluster is also now present
  in the cluster.yaml config files, with
  https://github.com/2i2c-org/pilot-hubs/pull/456. Even though
  that has not been merged yet, it is already deployed. This PR
  updates the schema to allow the support config.


consideRatio (approved):
LGTM! I love that we have schema.yaml stuff going on in this project!

PR #464, opened 2021-06-11
Cleanup pytest output in GitHub actions

yuvipanda:
- On failure, we've been printing per-hub API token in GitHub
  action. This hides it.
- Rotating the secret base key helps change all previously
  exposed api keys


consideRatio (approved):
LGTM nice Python tricks!

PR #468, opened 2021-06-14
Tune-up cloudbank demo hub for UCB Workshop

yuvipanda:
UC Berkeley is running their national data science teaching
workshop this week, and will be using our demo hub for
most of their classes. To ensure smooth use, we:

1. Setup pre-puller. We'll also increase minimum size of the cluster
   separately
2. Increase resource guarantees to prevent possibility of users running
   out of RAM


yuvipanda:
Already deployed, self-merging before conference starts: https://data.berkeley.edu/academics/resources/data-science-education-resources/2021-national-workshop-data-science-education

PR #470, opened 2021-06-14
Tweak daskhub config defaults

yuvipanda:
- Currently, users have to set these manually, so limiting it
   to 32 prevents provisioning instances larger than that.
- Set daskhub default to /lab - orgs using dask mostly want to
  use lab as the default

Ref https://github.com/2i2c-org/pilot-hubs/issues/291#issuecomment-858075940


damianavila (approved):
I think /lab is a sensible default for daskhub.
LGTM.

PR #471, opened 2021-06-14
Provixion x1 instances for carbonplan

yuvipanda:
- Requested by Joe in https://github.com/2i2c-org/pilot-hubs/issues/291#issuecomment-858075940
- Refresh auth credentials, they had expired. Fixed in
   https://github.com/2i2c-org/pilot-hubs/issues/381

Depends on https://github.com/2i2c-org/pilot-hubs/pull/470


yuvipanda:
Already deployed!

damianavila (approved):
Huge instances!
LGTM, merging now since it is already deployed.

PR #472, opened 2021-06-16
Run deploy actions only when necessary

yuvipanda:


sgibson91 (approved):
LGTM!

damianavila:
What about changes in the images/user?

yuvipanda:
> What about changes in the images/user?

Those need to be manually built and deployed - there's no autobuilding yet.

damianavila:
> Those need to be manually built and deployed - there's no autobuilding yet.

Yep, you are right... I was thinking about the tag change but that actually happens in the hub template, IIRC.

damianavila (approved):
LGTM

yuvipanda:
Thanks, @damianavila

PR #477, opened 2021-06-21
Make hub domains strings, not lists

yuvipanda:
Missed in https://github.com/2i2c-org/pilot-hubs/pull/460


yuvipanda:
Deployed, + unbreaks palomar

PR #480, opened 2021-06-23
Handle clients where callback / logout URLs were never set

yuvipanda:
Our code now assumes these keys are present, but in cases
where the client was created before we added logout_url support,
the key does not exist. This commit handles these situations
gracefully.


GeorgianaElena (approved):
LGTM. Thanks for fixing this @yuvipanda!

PR #483, opened 2021-06-23
[WIP] Leads repo mentioned in New Hub issue template no longer exists

sgibson91:
The leads repo that is mentioned in the New Hub issue template no longer exists in the 2i2c org. Where should the issue point to instead, or should the point be removed from the template?

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-hub.md
+++ b/.github/ISSUE_TEMPLATE/new-hub.md
@@ -22,7 +22,7 @@ title: "[Hub] - [Hub name]"
 
 # Important Information
 
-- **Link to leads issue**: <!--  link to the leads issue that generated this hub (see https://github.com/2i2c-org/leads/issues) -->
+- **Link to < REPO > issue**: <!--  link to the leads issue that generated this hub (see < LINK >) -->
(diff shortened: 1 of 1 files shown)


GeorgianaElena (commented):
@sgibson91, I've just checked and the leads repo still exists, but it's shown as `Private` from my account.

I will check to see if I have the right permissions to make it visible for you too and report back :D

GeorgianaElena:
@sgibson91, I think I managed to change the org role. Can you please check if you can now see the repo?

sgibson91:
Yes, I can see it! Thanks @GeorgianaElena!

damianavila:
mmm... I can not see that repo either... @GeorgianaElena can you help me with that as well?
Maybe we should adjust that repo permissions to be visible by the team group?

choldgraf:
hey all - the `leads` repo is an old experiment in keeping track of groups we were working with. It isn't up to date, because we were transitioning to using Monday.com which is the platform that ICSI was using. Since we are switching to a new fiscal sponsor, we may need to move away from Monday.com anyway. 

I set up this issue to track figuring out a new leads pipeline:

https://github.com/2i2c-org/team-compass/issues/132

PR #484, opened 2021-06-24
Add Altair to the default user image

yuvipanda:
Re-arrange Dockerfile to reduce rebuilding of our R packages

Fixes https://github.com/2i2c-org/pilot-hubs/issues/478

### DIFF
--- a/hub-templates/basehub/values.yaml
+++ b/hub-templates/basehub/values.yaml
@@ -106,7 +106,7 @@ jupyterhub:
       hub.jupyter.org/node-purpose: user
     image:
       name: set_automatically_by_automation
-      tag: e3bbe25
+      tag: "135823b"
     storage:
       type: static
       static:
--- a/images/user/Dockerfile
+++ b/images/user/Dockerfile
@@ -120,23 +120,24 @@ RUN sed -i -e '/^R_LIBS_USER=/s/^/#/' /etc/R/Renviron && \
 
 USER ${NB_USER}
 
-COPY requirements.txt /tmp/
-
-RUN python3 -m pip install -r /tmp/requirements.txt
-
 # Set CRAN mirror to rspm before we install anything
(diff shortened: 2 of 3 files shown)


yuvipanda:
I built and pushed the images.

images/user/Dockerfile
```diff
@@ -120,23 +120,24 @@ RUN sed -i -e '/^R_LIBS_USER=/s/^/#/' /etc/R/Renviron && \
...
 COPY Rprofile.site /usr/lib/R/etc/Rprofile.site
 # RStudio needs its own config
 COPY rsession.conf /etc/rstudio/rsession.conf
 
-# Install IRKernel
-RUN r -e "install.packages('IRkernel', version='1.1.1')" && \
-    r -e "IRkernel::installspec(prefix='${CONDA_DIR}')"
-
 # Install R packages, cleanup temp package download location
+# Install R packages before python packages, since these take longer to install
```
GeorgianaElena, line 129: Great idea!

hub-templates/basehub/values.yaml
```diff
@@ -106,7 +106,7 @@ jupyterhub:
       hub.jupyter.org/node-purpose: user
     image:
       name: set_automatically_by_automation
-      tag: e3bbe25
+      tag: "135823b"
```
GeorgianaElena, line 109: Probably a stupid question, but curious if the `"`around the image tag make any difference here?

yuvipanda: https://github.com/helm/helm/issues/1694 happens whenever a hash is either just a number, or a string of the form `123e44` (it is taken as a number in scientific notation). So I started putting quotes around tags as a precaution...

GeorgianaElena: Interesting. Thanks for explaining this!

GeorgianaElena:
Great! Thanks @yuvipanda!

PR #496, opened 2021-07-06
docs: Don't mention multiple domains

yuvipanda:
This is no longer supported - one hub should just stay in
one domain.

Ref https://github.com/2i2c-org/pilot-hubs/pull/460


sgibson91:
So just to make sure I've followed the discussion in #460 - So long as the CNAME is applied in the DNS settings, we don't have to do anything in the hub config for redirection? But users can still use a domain name that doesn't explicitly mention 2i2c to reach a 2i2c managed hub?

I'm thinking about the Pangeo case, if that helps add context to my thought process :)

yuvipanda:
> So just to make sure I've followed the discussion in #460 - So long as the CNAME is applied in the DNS settings, we don't have to do anything in the hub config for redirection?

Yep, that is correct!

sgibson91 (approved):
Whoop! 🎉

yuvipanda:
Thanks, @sgibson91!

PR #497, opened 2021-07-06
docs: Add section on helpful tools

yuvipanda:
This document can collect tools that are required to work
with the infrastructure deployed from this repo, and any
additional tips / resources that are specifically useful
to us. The hope is that this list grows over time as we gain
more experience.

Ref https://github.com/2i2c-org/pilot-hubs/issues/476

### DIFF
--- a/docs/reference/tools.md
+++ b/docs/reference/tools.md
@@ -1,0 +1,67 @@
+# Tools used in this repo
+
+## Base tools
+
+These are helpful for everyone, and not cloud-specific.
+
+### [`kubectl`](https://kubernetes.io/docs/tasks/tools/)
+
+The canonical commandline tool for talking to kubernetes clusters.
+Debugging and understanding runtime behavior of our hubs is greatly
+enhanced by knowledge of how to use `kubectl`. Understanding how to
+use `kubectl` really helps understand how kubernetes itself works.
+
+#### Tips
+
+- Consider aliasing it to `k`, as you might be typing it a lot!
+- It is easy to accidentally operate on the wrong kubernetes cluster!
+  Putting the current kubernetes context in your terminal prompt
+  (via something like [starship](https://starship.rs/)) can help a
+  lot!
+
+#### Additional resources and tools
+
+- The [official cheatsheet](https://kubernetes.io/docs/reference/kubectl/cheatsheet/),
+  with extremely helpful tips.
+- [Lens](https://k8slens.dev/) is an amazing GUI for exploring various objects
+  in a kubernetes cluster. It can do almost everything `kubectl` does, in
+  a fairly intuitive way. Highly recommended as a supplement (or even alternative)
+  to `kubectl`.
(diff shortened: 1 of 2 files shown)


docs/reference/tools.md
```diff
@@ -0,0 +1,67 @@
...
+### [`kubectl`](https://kubernetes.io/docs/tasks/tools/)
+
+The canonical commandline tool for talking to kubernetes clusters.
+Debugging and understanding runtime behavior of our hubs is greatly
+enhanced by knowledge of how to use `kubectl`. Understanding how to
+use `kubectl` really helps understand how kubernetes itself works.
+
+#### Tips
+
+- Consider aliasing it to `k`, as you might be typing it a lot!
```
damianavila, line 16: That is a good idea although I would keep `kubectl` in our docs when we showcase commands...

damianavila (approved):
This is a great start!!
And I agree this is something to iterate on... so 👍  on merging early than later!

docs/reference/tools.md
yuvipanda: Yeah, I agree!

sgibson91 (approved):
Love it! 💖 Also agree with @damianavila that this can be iterated upon and I think it's fine to merge as is and open new PRs as folk discover new things :)

yuvipanda:
Pinging @consideRatio who probably knows a bunch of useful tools and tricks

consideRatio:
Some tricks using the `helm` cli are...

- the `--repo` flag is nice, helps you do a single command to do something with a chart without using `helm repo add` etc. first.
  ```shell
  helm template --repo https://jupyterhub.github.io/helm-chart/ jupyterhub
  ```
- the `--show-only` flag is nice, helps you review a specific helm chart template instead of the entire chart.
  ```shell
  # example on how to access a specific template
  helm template --repo https://jupyterhub.github.io/helm-chart/ jupyterhub \
      --show-only templates/hub/deployment.yaml
  
  # example on how to access a specific template from a subchart
  # note we need to specify version, as there is no non-pre-release
  # version available
  helm template --repo https://jupyterhub.github.io/helm-chart/ binderhub \
      --version=0.2.0-n611.he777436 \
      --set jupyterhub.hub.services.binder.apiToken=asd \
      --show-only charts/jupyterhub/templates/hub/deployment.yaml
  ```
- the `--validate` flag with `helm template` will not only render the templates, but also send --dry-run requests to the configured k8s api-server to see if the k8s api-server would accept those rendered manifests. This is useful as `helm template` would render almost anything, even if its not valid k8s resources.

damianavila:
> Some tricks using the helm cli are...

A PR adding these tricks is more than welcome 😜  !

consideRatio:
Thanks for the nudge @damianavila.

I'm relieved to see markdown based documentation instead of rST :relieved: - I'm totally sold on the use of MyST for project documentation atm.

https://github.com/2i2c-org/pilot-hubs/pull/499 opened

PR #498, opened 2021-07-06
docs: Revamp info on how to add a new hub

yuvipanda:
Is also a jumping off point for other things that need
to be better documented.

### DIFF
--- a/docs/howto/operate/add-hub.md
+++ b/docs/howto/operate/add-hub.md
@@ -13,7 +13,8 @@ decide which cluster it should live in.
 
 1. For standalone hubs where we are being paid or the client has cloud
    credits, we will need to create a cluster specifically for this hub
-   before deploying it.
+   before deploying it. As a general rule, we will create one cluster per
+   billing account.
 2. For hubs we are running for free, they will go into the `pilot-hubs`
    cluster.
 3. Hubs run with the support of [cloudbank](https://www.cloudbank.org/),
--- a/docs/howto/operate/add-hub.md
+++ b/docs/howto/operate/add-hub.md
@@ -18,7 +18,8 @@ Each cluster has configuration + list of hubs it supports under
 
 ## Step 2: Decide the template used for the hub
 
-Based on the kinda usage expected of the hub select a [hub template](../../topic/hub-templates.md)
+Hub templates map on to different major use-cases that we aim to enable.
+Ask the hub user(s) about their needs and expected usage, and use their answer to select a [hub template](../../topic/hub-templates.md).
 to be used.
 
 ## Step 4: Decide the authentication provider to be used
@@ -65,3 +66,7 @@ Make a PR with your changes, referencing the issue for creation of the hub. Seek
 review from someone else, and get this merged!
 
 Getting this merged will mark the hub as being deployed.
+
+Finally, notify the users of the hub that it is now deployed, and ask that they test it out to make sure it works as expected.
+If they are happy with the result, then close the issue for deploying the new hub.
+Congratulations, you are done! 🎉 
--- a/docs/howto/operate/add-hub.md
+++ b/docs/howto/operate/add-hub.md
@@ -59,12 +59,13 @@ tries. Specifically test that at least the following work ok:
 
 1. Authentication provider
 2. Admin user access
-3. User image in use
+3. User image in use (check the `JUPYTER_IMAGE` environment variable in a spawned server)
 4. Default user interface when the user logs in (lab, notebook, rstudio, retrolab, etc)
 5. Home page display configuration
 6. (Optionally) `dask-gateway` functionality
 7. (Optionally) Access to any cloud resources (like storage buckets, etc)
    granted to the hub users
+
 ## Step 7: Make a PR
 
 Make a PR with your changes, referencing the issue for creation of the hub. Seek
--- a/docs/howto/operate/add-hub.md
+++ b/docs/howto/operate/add-hub.md
@@ -1,8 +1,67 @@
 # Add a new hub
 
-Adding a new hub requires that we do three things:
+## Step 1: Decide where the hub will live
 
-1. [Create a new GitHub issue for the hub](https://github.com/2i2c-org/pilot-hubs/issues/new).
-2. Create the hub via appending a new hub entry in the appropriate cluster file under
-   `config/hubs` file (see [](../../topic/config.md) for more information).
-3. Start following [team process](team-process) around operating the hubs.
+Hubs need to exist on a kubernetes cluster, so we will need to
+decide which cluster it should live in.
+
+1. For standalone hubs where we are being paid or the client has cloud
+   credits, we will need to create a cluster specifically for this hub
+   before deploying it.
+2. For hubs we are running for free, they will go into the `pilot-hubs`
+   cluster.
+3. Hubs run with the support of [cloudbank](https://www.cloudbank.org/),
+   will go in the `cloudbank` cluster.
+
+Each cluster has configuration + list of hubs it supports under
+`config/hubs/<cluster-name>.cluster.yaml`.
+
+## Step 2: Decide the template used for the hub
+
+Based on the kinda usage expected of the hub select a [hub template](../../topic/hub-templates.md)
+to be used.
+
+## Step 4: Decide the authentication provider to be used
+
+In consultation with the users, decide
+[which authentication provider](https://pilot.2i2c.org/en/latest/admin/configuration/login.html#authentication)
+the hub should use. While this can be changed later, it's a messy
+process - so we should try to get this right the first time.
+
+## Step 5: Decide the user image for the hub
+
+The default user image is present in this repository (`images/user`),
+and is geared towards simple datascience classes - with both R and
+Python. Determine if this is acceptable - if not, you need to
+make a [custom image](../configure/update-env.md) for them.
+
+## Step 6: Add hub config
+
+Add an entry for the hub in `config/hubs/<cluster-name>.cluster.yaml` for this hub.
+The following docs might be helpful:
+
+1. [Configuration Structure](../../topic/config.md)
+2. [JSON Schema for `.cluster.yaml` file](https://github.com/2i2c-org/pilot-hubs/blob/master/config/hubs/schema.yaml)
+3. [Zero to JupyterHub on k8s Docs](https://zero-to-jupyterhub.readthedocs.io/en/latest/), since ultimately
+   that is what we are configuring.
+
+## Step 7: Deploy manually and test
+
+[Deploy the hub manually](./manual-deploy.md) and test to make sure it works
+the way you would like it to. This might take some iteration and multiple
+tries. Specifically test that at least the following work ok:
+
+1. Authentication provider
+2. Admin user access
+3. User image in use
+4. Default user interface when the user logs in (lab, notebook, rstudio, retrolab, etc)
+5. Home page display configuration
+6. (Optionally) `dask-gateway` functionality
+7. (Optionally) Access to any cloud resources (like storage buckets, etc)
+   granted to the hub users
+## Step 8: Make a PR
+
+Make a PR with your changes, referencing the issue for creation of the hub. Seek
+review from someone else, and get this merged!
+
+Getting this merged will mark the hub as being deployed.
--- a/docs/howto/operate/add-hub.md
+++ b/docs/howto/operate/add-hub.md
@@ -1,5 +1,11 @@
 # Add a new hub
 
+## Step 0: Open a GitHub issue to track hub creation
+
+In order to track progress on deploying the new hub, [open a GitHub issue in the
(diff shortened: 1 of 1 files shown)


GeorgianaElena (approved):
Just love it! :tada:

I noticed that opening up a new GitHub issue is not in the steps anymore. Not sure if this was the intention, but I think we should keep it, since it helps us organize :D Maybe add a note abut it at the end to instruct ticking the right boxes after the hub was added, or open a new issue if one wasn't opened already?

docs/howto/operate/add-hub.md
```diff
@@ -1,8 +1,67 @@
...
+decide which cluster it should live in.
+
+1. For standalone hubs where we are being paid or the client has cloud
+   credits, we will need to create a cluster specifically for this hub
+   before deploying it.
+2. For hubs we are running for free, they will go into the `pilot-hubs`
+   cluster.
+3. Hubs run with the support of [cloudbank](https://www.cloudbank.org/),
+   will go in the `cloudbank` cluster.
+
```
damianavila, line 15: I like this categorization!

docs/howto/operate/add-hub.md
```diff
@@ -1,8 +1,67 @@
...
+Each cluster has configuration + list of hubs it supports under
+`config/hubs/<cluster-name>.cluster.yaml`.
+
+## Step 2: Decide the template used for the hub
+
+Based on the kinda usage expected of the hub select a [hub template](../../topic/hub-templates.md)
+to be used.
+
```
damianavila, line 23: Step 3 is missing here 😛

docs/howto/operate/add-hub.md
```diff
@@ -1,8 +1,67 @@
...
+## Step 4: Decide the authentication provider to be used
+
+In consultation with the users, decide
+[which authentication provider](https://pilot.2i2c.org/en/latest/admin/configuration/login.html#authentication)
+the hub should use. While this can be changed later, it's a messy
```
damianavila, line 28: Maybe convert the last sentence into a note?

docs/howto/operate/add-hub.md
```diff
@@ -1,8 +1,67 @@
...
+process - so we should try to get this right the first time.
+
+## Step 5: Decide the user image for the hub
+
+The default user image is present in this repository (`images/user`),
+and is geared towards simple datascience classes - with both R and
+Python. Determine if this is acceptable - if not, you need to
+make a [custom image](../configure/update-env.md) for them.
+
```
damianavila, line 37: Maybe a note about the upcoming process to self-serve the custom image?

docs/howto/operate/add-hub.md
```diff
@@ -1,8 +1,67 @@
...
+The following docs might be helpful:
+
+1. [Configuration Structure](../../topic/config.md)
+2. [JSON Schema for `.cluster.yaml` file](https://github.com/2i2c-org/pilot-hubs/blob/master/config/hubs/schema.yaml)
+3. [Zero to JupyterHub on k8s Docs](https://zero-to-jupyterhub.readthedocs.io/en/latest/), since ultimately
+   that is what we are configuring.
+
+## Step 7: Deploy manually and test
+
+[Deploy the hub manually](./manual-deploy.md) and test to make sure it works
```
damianavila, line 50: We have automated deployment for the pilots and cloudbank hubs... are we encouraging here to do the manual deployment instead? Or we should include also the automatic deployment and then explain the testing steps indicated below?

docs/howto/operate/add-hub.md
```diff
@@ -1,8 +1,67 @@
...
+the way you would like it to. This might take some iteration and multiple
+tries. Specifically test that at least the following work ok:
+
+1. Authentication provider
+2. Admin user access
+3. User image in use
+4. Default user interface when the user logs in (lab, notebook, rstudio, retrolab, etc)
```
damianavila, line 57: Do we have retrolab? I think it is not actually on our images, but I might be wrong about it...

docs/howto/operate/add-hub.md
```diff
@@ -1,8 +1,67 @@
...
+5. Home page display configuration
+6. (Optionally) `dask-gateway` functionality
+7. (Optionally) Access to any cloud resources (like storage buckets, etc)
+   granted to the hub users
```
damianavila, line 61: For the optional ones, we should reference something more specific about the test to perform, I think...

docs/howto/operate/add-hub.md
```diff
@@ -1,8 +1,67 @@
...
+
+## Step 7: Deploy manually and test
+
+[Deploy the hub manually](./manual-deploy.md) and test to make sure it works
+the way you would like it to. This might take some iteration and multiple
+tries. Specifically test that at least the following work ok:
+
+1. Authentication provider
+2. Admin user access
+3. User image in use
```
damianavila, line 56: How we can quickly identify the image we are using? In the past, I have "injected" env variables in the Dockerfile to have a quick reference...

docs/howto/operate/add-hub.md
```diff
@@ -1,8 +1,67 @@
...
+4. Default user interface when the user logs in (lab, notebook, rstudio, retrolab, etc)
+5. Home page display configuration
+6. (Optionally) `dask-gateway` functionality
+7. (Optionally) Access to any cloud resources (like storage buckets, etc)
+   granted to the hub users
+## Step 8: Make a PR
```
damianavila, line 62: Missing blank line before line 62.

docs/howto/operate/add-hub.md
sgibson91: +1!

docs/howto/operate/add-hub.md
sgibson91: Agreed. I always prefer good documentation for stuff I do less frequently as I won't build up the muscle memory like with other tasks. (And it goes without saying, we should have good documentation for everything anyway 😛 )

sgibson91 (approved):
Looks great to me! Agree with everyone else's comments so far

docs/howto/operate/add-hub.md
```diff
@@ -1,8 +1,67 @@
...
 
-Adding a new hub requires that we do three things:
+## Step 1: Decide where the hub will live
 
-1. [Create a new GitHub issue for the hub](https://github.com/2i2c-org/pilot-hubs/issues/new).
-2. Create the hub via appending a new hub entry in the appropriate cluster file under
-   `config/hubs` file (see [](../../topic/config.md) for more information).
-3. Start following [team process](team-process) around operating the hubs.
+Hubs need to exist on a kubernetes cluster, so we will need to
+decide which cluster it should live in.
```
choldgraf, line 6:
```suggestion
decide which cluster it should live in.
As a general rule, there is **one cluster per billing account**.
```

docs/howto/operate/add-hub.md
choldgraf: I wonder if, rather than hard-coding "cloudbank" here, we should keep a list of "special" clusters like the CloudBank one (basically any cluster that has more than a 1-1 mapping onto customers) and then reference that table?

I seem to remember us having an issue about having such a "source of truth" for clusters, but maybe this was just for hubs?

docs/howto/operate/add-hub.md
```diff
@@ -1,8 +1,67 @@
...
+   cluster.
+3. Hubs run with the support of [cloudbank](https://www.cloudbank.org/),
+   will go in the `cloudbank` cluster.
+
+Each cluster has configuration + list of hubs it supports under
+`config/hubs/<cluster-name>.cluster.yaml`.
+
+## Step 2: Decide the template used for the hub
+
+Based on the kinda usage expected of the hub select a [hub template](../../topic/hub-templates.md)
```
choldgraf, line 21:
```suggestion
Hub templates map on to different major use-cases that we aim to enable.
Ask the hub user(s) about their needs and expected usage, and use their answer to select a [hub template](../../topic/hub-templates.md).
```

docs/howto/operate/add-hub.md
```diff
@@ -1,8 +1,67 @@
...
+to be used.
+
+## Step 4: Decide the authentication provider to be used
+
+In consultation with the users, decide
```
choldgraf, line 26: On this page, perhaps we should define a **Community Champion** role for a hub (or maybe we should re-name this "user representative"?) and then suggest that we ask this person questions throughout the process, rather than to use a generic "user" identity like here? I'd like us to move away from the possibility that multiple users on a hub might be needed to answer questions.

docs/howto/operate/add-hub.md
choldgraf: Additionally, perhaps we can document a basic list of the current user images and when each should be used? That feels like it could be another PR though...

docs/howto/operate/add-hub.md
choldgraf: yeah a TODO: here sounds good too (you can add comments in MyST markdown with %, so we can use that to add TODO similar to what we'd do in Python)

docs/howto/operate/add-hub.md
choldgraf: Or more generally, it'd be useful to have some guide for when we should try a manual deploy vs. just auto-deploying vs. config.yaml. The manual deploy step I suspect will be a big amount of extra work per hub (and will mean that the only people that can add new hubs will also need to have the skills to manually deploy a hub), so do we _really_ need to do it each time?

choldgraf: Another approach could be to tell the reader to auto-deploy the hub via editing `config.yaml`, and to then confirm the following things work as expected before telling the user representative about it?

docs/howto/operate/add-hub.md
choldgraf: perhaps we could create a new issue to track these? they might be more involved and I think will require separate sections. In particular for dask-gateway and non-standard cloud resources, that will require more expertise and knowledge so we should give those topics extra attention (and this can probably be another PR)

docs/howto/operate/add-hub.md
```diff
@@ -1,8 +1,67 @@
...
+5. Home page display configuration
+6. (Optionally) `dask-gateway` functionality
+7. (Optionally) Access to any cloud resources (like storage buckets, etc)
+   granted to the hub users
+## Step 8: Make a PR
+
+Make a PR with your changes, referencing the issue for creation of the hub. Seek
+review from someone else, and get this merged!
+
+Getting this merged will mark the hub as being deployed.
```
choldgraf, line 67:
```suggestion
Getting this merged will mark the hub as being deployed.

Finally, notify the users of the hub that it is now deployed, and ask that they test it out to make sure it works as expected.
If they are happy with the result, then close the issue for deploying the new hub.
Congratulations, you are done! 🎉 
```

choldgraf (commented):
This is really great and very helpful for people like me who have not deployed a ton of hubs :-). I added in a few comments about extra information / context we might want to add. Happy to iterate on this a bit more!

Note that this may also close https://github.com/2i2c-org/team-compass/issues/128

docs/howto/operate/add-hub.md
damianavila: It was for hubs, indeed (IIRC).
I am +1 about having something more dynamic here for "special" clusters...

damianavila: Although maybe we can add that later... when we have more than one "special" cluster?

docs/howto/operate/add-hub.md
damianavila: > I'd like us to move away from the possibility that multiple users on a hub might be needed to answer questions.

💯

docs/howto/operate/add-hub.md
yuvipanda: Incorporated!

docs/howto/operate/add-hub.md
```diff
@@ -1,8 +1,67 @@
 # Add a new hub
 
-Adding a new hub requires that we do three things:
+## Step 1: Decide where the hub will live
```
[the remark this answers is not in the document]
yuvipanda: I incorporated this, but pointing to pilot-hubs still. https://github.com/2i2c-org/team-compass/issues/151 talks about the support/ repository.

yuvipanda (commented):
I think I've addressed everything, by either amending the docs or opening new issues :)

docs/howto/operate/add-hub.md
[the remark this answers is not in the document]
damianavila: I think that is OK for now, @choldgraf?

docs/howto/operate/add-hub.md
damianavila: @choldgraf, until we have the Community champion role well defined, I think this is OK for now. Let us know if you disagree.

damianavila (approved):
I think the remaining things could be changed in a follow-up PR. Approving...

choldgraf (approved):
I'm playing a game kind of like charades with my french family. But it's all in french and i have *no idea* what anybody is saying, so I'm approving this PR 😅

PR #503, opened 2021-07-12
Add some docs describing exporting and encrypting the CD SA

sgibson91:
This PR adds some docs on how to export and encrypt the Continuous Deployment Service Account of a new cluster, ready to begin deploying and managing hubs installed on the cluster.

### DIFF
--- a/docs/howto/operate/new-tf-cluster.md
+++ b/docs/howto/operate/new-tf-cluster.md
@@ -97,34 +97,36 @@ Congratulations, you've just deployed a new cluster!
 
 To begin deploying and operating hubs on your new cluster, we need to export the Continuous Deployment Service Account created by terraform, encrypt it using `sops`, and store it in the `secrets` directory of the `pilot-hubs` repo.
 
-First, create a JSON key for the service account.
-Output it into a JSON file that we will then encrypt.
-In this example, we'll use `~/out.json`.
+Check you are still in the correct terraform workspace
 
 ```bash
-gcloud iam service-accounts keys create \
-    ~/out.json \
-    --iam-account=PREFIX-cd-sa@PROJECT_ID.iam.gserviceaccount.com
+terraform workspace show
 ```
 
-where `PREFIX` and `PROJECT_ID` are the values we set in our `.tfvars` file.
+If you need to change, you can do so as follows
 
-Encrypt the key using `sops` and move it into the `secrets` directory
+```bash
+terraform workspace list  # List all available workspaces
+terraform workspace select WORKSPACE_NAME
+```
+
+Then, output the JSON key for the service account created by terraform to a file under the `secrets` directory.
+
+```bash
+terraform output -raw ci_deployer_key > ../secrets/CLUSTER_NAME.json
+```
+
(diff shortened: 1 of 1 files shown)


docs/howto/operate/new-tf-cluster.md
```diff
@@ -92,3 +92,39 @@ terraform apply -var-file=projects/CLUSTER.tfvars
...
+
+## Exporting and Encrypting the Continuous Deployment Service Account
+
+To begin deploying and operating hubs on your new cluster, we need to export the Continuous Deployment Service Account created by terraform, encrypt it using `sops`, and store it in the `secrets` directory of the `pilot-hubs` repo.
+
+First, create a JSON key for the service account.
+Output it into a JSON file that we will then encrypt.
+In this example, we'll use `~/out.json`.
+
+```bash
```
yuvipanda, line 104: I think the key is also generated as part of the terraform run, as a terraform `output` variable. Try `terraform output -raw ci_deployer_key  `?

yuvipanda (changes requested):

docs/howto/operate/new-tf-cluster.md
sgibson91: You're right. Using gcloud would be robust against being in the wrong terraform workspace though. How do you feel about documenting both approaches?

yuvipanda: @sgibson91 if we create another key using `gcloud`, it is unfortunately invisble to terraform. So taking down a setup with `terraform destroy`, or changing the permissions of the service account itself might not translate. If we get the wrong terraform workspace (as I have done, haha) actions can cause major destruction anyway, so I think we should try document making sure you're in the right terraform workspace rather than using individual `gcloud` commands. 

I [display the current terraform workspace](https://starship.rs/config/#terraform) (and gcloud project, aws profile, k8s context) with the wonderful [starship](https://starship.rs) project. I think similar things might exist for other terminal setups as well - I think that is one way to validate this. The JSON generated for the key will also have the name of the project it was issued for - we can use that too!

How does this approach sound, rather than documenting a `gcloud` command to create a new key?

sgibson91: Yes, my brain didn't connect with the `create` part of the command - what I really wanted was a `show` 😂 Yep sure, working effectively with terraform workspaces sounds like a good idea.

yuvipanda: > Yes, my brain didn't connect with the create part of the command - what I really wanted was a show 

Yeah, once you create a key you can't retrieve it again in GCP. But since terraform stores 'state', it actually saves the key! So we can get the same one back multiple times :)

sgibson91: (P.S. Turns out that my shell prompt, PowerLevel10k, also displays the terraform workspace name - I just hadn't connected this with the little tree symbol. It's Monday ok? This is why I don't do destructive things on Mondays 😂 )

sgibson91: Ok, got my head wrapped around everything again and updated this

yuvipanda:
Thanks, @sgibson91!

PR #511, opened 2021-07-13
Document authentication mechanisms for gcloud

yuvipanda:
Ref https://github.com/2i2c-org/pilot-hubs/issues/476


sgibson91 (approved):
Thanks for this, didn't actually know this! (I always do both anyway :D )

GeorgianaElena (approved):
I only used the second one. Thanks @yuvipanda!

damianavila (approved):
LGTM

PR #513, opened 2021-07-13
Add Terraform setup for Azure cluster

yuvipanda:
- Prefixes in GCP keep resources separate from each other.
  Azure has the native concept of a ResourceGroup for the same,
  so we do not need prefixes.
- Overall structure tries to be as close to GCP as possible,
  but Azure specific naming is used when necessary. For example,
  Azure uses the word `vm_size` where GCP uses `machine_type`,
  and we use `vm_size` here
- Setup notebook and dask nodes with same labels and tains as
  we do on GCP. We can ruse the same logic in our deployers
  this way.
- Remove the NFS server + ansible setup. Similar to EFS, we
  should try use https://azure.microsoft.com/en-in/services/storage/files/
  on Azure
- Brings in the terraform setup from https://github.com/utoronto-2i2c/jupyterhub-deploy/tree/staging/terraform,
  preserving history with some git tricks, thanks to
  https://bneijt.nl/blog/merge-a-subdirectory-of-another-repository-with-git/
    
closes https://github.com/2i2c-org/pilot-hubs/issues/512


sgibson91 (approved):
Looks good! Neat trick to preserve history :D

yuvipanda:
Thanks, @sgibson91! I've now modified it quite a bit, to try bring it closer to our GCP terraform. Would <3 a review

terraform/azure/main.tf
```diff
@@ -0,0 +1,164 @@
...
+  auto_scaler_profile {
+    skip_nodes_with_local_storage = true
+  }
+
+  identity {
+    type = "SystemAssigned"
+  }
+
+  network_profile {
+    # I don't trust Azure CNI
```
sgibson91, line 74: Good instinct - we found a bug on the Turing mybinder deployment that meant network policies were not being applied as expected!

terraform/azure/variables.tf
```diff
@@ -0,0 +1,84 @@
+variable "tenant_id" {
+  type        = string
+  description = <<-EOT
+  Tenant ID containing the subscription to use.
+
+  A UUID referencing the
+  This
```
sgibson91, line 7: Some unfinished docstrings?

sgibson91 (approved):
LGTM! Thanks for the revamp

yuvipanda:
@choldgraf yeah, that's been done in this PR and deployed. 

Justice Innovation League hub is using this infrastructure now. Per our code review guideline (https://github.com/2i2c-org/pilot-hubs/blob/master/docs/contributing/code-review.md#changes-to-set-up-new-infrastructure), I think I can merge this now but after that I should destroy and recreate the JusticeInnovationLab cluster. However, they've been directly using it for a while, so I'm going to *not* do that step. This PR predates that guideline - I think with that guideline in mind, this PR would've been a lot smaller and merged quicker...

I hope this is an ok decision to make this way!

choldgraf:
Amazing! And sounds good, your rationale makes sense to me. I think the only thing remaining here is to update the docs here: https://pilot-hubs.2i2c.org/en/latest/howto/operate/new-cluster/azure.html

PR #514, opened 2021-07-13
Move current terraform into a GCP specific module

yuvipanda:
We will use completely different terraform code for each cloud provider,
under `terraform/<cloud-provider>`. This is much simpler than trying to abstract
them away into a 'lowest common denominator' set of modules.

Ref https://github.com/2i2c-org/pilot-hubs/issues/512


docs/topic/terraform.md
```diff
@@ -13,4 +13,10 @@ file with variable definitions for that cluster.
 
 Workspaces are stored centrally in the `two-eye-two-see-org` GCP project, even
 when we use Terraform for projects running on AWS / Azure. You must have
-access to this project before you can use terraform for our infrastructure.
\ No newline at end of file
+access to this project before you can use terraform for our infrastructure.
+
+## Cloud provider abstractions
+
+We will use completely different terraform code for each cloud provider,
```
yuvipanda, line 20: Would love to know what you think of this, @consideRatio @damianavila @sgibson91 @GeorgianaElena

sgibson91: This is similar to what I was thinking about over lunch. Each cloud provider we deploy to via terraform has it's own submodule, and we have a top-level variable like `cloud_provider` that we use in some logic to import the correct module when doing a terraform plan/apply. It means more code to maintain, but as of yet I don't know of a more elegant solution and terraform isn't a _truly_ cloud-agnostic tool (it's dependent on the provider.

yuvipanda: > Each cloud provider we deploy to via terraform has it's own submodule, and we have a top-level variable like cloud_provider that we use in some logic to import the correct module when doing a terraform plan/apply.

I tried this, and unfortunately you can't use a variable to specify which module to load in terraform :(

sgibson91: Boo! :( In which case, I think this solution is good enough.

consideRatio: I wonder if @andersy005 or @cmd-ntrf with significant experience with terraform could comment about this for us? Anderson and Félix, do you think it is a reasonable approach to have separate terraform code for deploying infrastructure to separate cloud providers?

docs/howto/operate/new-tf-cluster.md
```diff
@@ -56,7 +56,7 @@ gcloud auth application-default login
 gcloud auth application-default print-access-token
 ```
 
-Add the access token to the [terraform backend block](https://github.com/2i2c-org/pilot-hubs/blob/2ef8a4bf35bb5ee9bf04ab3db1218b8c183c5da2/terraform/main.tf#L2-L5) in `main.tf`.
+Add the access token to the [terraform backend block](https://github.com/2i2c-org/pilot-hubs/blob/master/terraform/gcp/main.tf#L2-L5) in `main.tf`.
```
consideRatio, line 59:
```suggestion
Add the access token to the [terraform backend block](https://github.com/2i2c-org/pilot-hubs/blob/HEAD/terraform/gcp/main.tf#L2-L5) in `main.tf`.
```

docs/howto/operate/new-tf-cluster.md
```diff
@@ -18,7 +18,7 @@ The _minimum_ inputs this file requires are:
 - `project_id`: GCP Project ID to create resources in.
   Should be the id, rather than display name of the project.
 
-See the [variables file](https://github.com/2i2c-org/pilot-hubs/blob/master/terraform/variables.tf) for other inputs this file can take and their descriptions.
+See the [variables file](https://github.com/2i2c-org/pilot-hubs/blob/master/terraform/gcp/variables.tf) for other inputs this file can take and their descriptions.
```
consideRatio, line 21:
```suggestion
See the [variables file](https://github.com/2i2c-org/pilot-hubs/blob/HEAD/terraform/gcp/variables.tf) for other inputs this file can take and their descriptions.
```

docs/howto/operate/new-tf-cluster.md
```diff
@@ -1,14 +1,14 @@
...
+This giude will walk through the process of adding a new cluster to our [terraform configuration](https://github.com/2i2c-org/pilot-hubs/tree/master/terraform/gcp).
 
 ## Cluster Design
 
 This guide will assume you have already followed the guidance in {ref}`/topic/cluster-design` to select the appropriate infrastructure.
 
 ## Creating a Terraform variables file for the cluster
 
-The first step is to create a `.tfvars` file in the [`terraform/projects` directory](https://github.com/2i2c-org/pilot-hubs/tree/master/terraform/projects).
+The first step is to create a `.tfvars` file in the [`terraform/gcp/projects` directory](https://github.com/2i2c-org/pilot-hubs/tree/master/terraform/gcp/projects).
```
consideRatio, line 11:
```suggestion
The first step is to create a `.tfvars` file in the [`terraform/gcp/projects` directory](https://github.com/2i2c-org/pilot-hubs/tree/HEAD/terraform/gcp/projects).
```

docs/howto/operate/new-tf-cluster.md
```diff
@@ -1,14 +1,14 @@
 # Add a new GKE cluster managed by our Terraform configuration
 
-This guide will walk through the process of adding a new cluster to our [terraform configuration](https://github.com/2i2c-org/pilot-hubs/tree/master/terraform).
+This giude will walk through the process of adding a new cluster to our [terraform configuration](https://github.com/2i2c-org/pilot-hubs/tree/master/terraform/gcp).
```
consideRatio, line 3:
```suggestion
This giude will walk through the process of adding a new cluster to our [terraform configuration](https://github.com/2i2c-org/pilot-hubs/tree/HEAD/terraform/gcp).
```

consideRatio (commented):
This LGTM from what I can tell and feel comfortable relying on @yuvipanda's judgement given that it has been explicit deliberation on the decision regarding making room for separate terraform code for different cloud providers.

I went ahead and made some off topic adjustments to the URLs :hear_no_evil:

docs/howto/operate/new-tf-cluster.md
yuvipanda: Oh very cool! TIL!

docs/topic/terraform.md
damianavila: I think this is a sensible proposal. Since terraform is not agnostic from the cloud provider I think any attempt to make that happen will look-like weird and, most likely, difficult if not impossible to do it in a sane way. But, hey, I do not have a lot of experience with terraform and I might be missing something...

damianavila (approved):
As I said in the specific comment, I think this is a sensible idea.

yuvipanda:
@sgibson91 this might require you to rebase some of the terraform code you are working on wrt private clusters

PR #525, opened 2021-07-17
Make carbonplan cluster more resilient

yuvipanda:
- Increase VM size and count of carbonplan master nodes
- Increase instancegroup limit for carbonplan


yuvipanda:
These have already been deployed

damianavila:
Belated LGTM.

PR #532, opened 2021-07-20
Make carbonplan hub more resilient - part 2

yuvipanda:
- We had another outage caused by resource exhaustion
  on the k8s master nodes. This commit separates the k8s
  master nodes from the hub control plane, giving them
  more room to breathe
- We had a single-node etcd cluster without a lot of
  resource commitments. It's now bumped up to 3 nodes,
  with more resources.
- Actually use coredns for our clusters
  A typo in the schema meant we were using legacy kube-dns,
  which was causing problems - DNS saturation seems to be
  a problem with the earlier kube-dns in carbonplan
- Increase the node size of our k8s master nodes for now -
      we can tone this down later if necessary

### DIFF
--- a/config/hubs/carbonplan.cluster.yaml
+++ b/config/hubs/carbonplan.cluster.yaml
@@ -22,11 +22,6 @@ hubs:
             - noresvport
             serverIP: fs-8a4e4f8d.efs.us-west-2.amazonaws.com
             baseShareName: /
-          shareCreator:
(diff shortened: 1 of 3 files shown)


yuvipanda:
Going to self merge this, as I've already deployed everything mentioned here.

damianavila:
I like the general idea of separating the master nodes from the hub core stuff!

PR #536, opened 2021-07-20
Specify dep version, otherwise improper constraint error

GeorgianaElena:
Somehow, without this, helm dependency's upgrade [fails](https://github.com/2i2c-org/pilot-hubs/runs/3111724001) with:

```
Error: dependency "basehub" has an invalid version/constraint format: improper constraint: 
...Successfully got an update from the "https://dask.org/dask-gateway-helm-repo/" chart repository
Error: Process completed with exit code 1.
``` 

It was failing even before the logo modification deployment attempt and I tracked it down to this commit: https://github.com/2i2c-org/pilot-hubs/commit/d9d79370abe2f2ef046c510e502f1172903b7edd

It says that the version is obsolete, but I found this piece of info here (https://kb.novaordis.com/index.php/Helm_Dependencies):
```
Note that even if it seems unnecessary, as an unpacked chart can have just one and only one version, the "version" information in the dependent's chart is actually required, and it is matched against ../b/Chart.yaml.

If the dependency declaration does not contain version, helm dependency update fails with:

helm dependency update ./a
[...]
Saving 1 charts
Save error occurred:  dependency b has an invalid version/constraint format: improper constraint:
Deleting newly downloaded charts, restoring pre-update state
Error: dependency b has an invalid version/constraint format: improper constraint:
```

cc @consideRatio, @yuvipanda :sunflower:

### DIFF
--- a/hub-templates/daskhub/Chart.yaml
+++ b/hub-templates/daskhub/Chart.yaml
@@ -5,6 +5,7 @@ name: daskhub
 version: "0.1.0"
 dependencies:
  - name: basehub
+   version: "0.1.0"
    repository: file://../basehub
  - name: dask-gateway
    version: "0.9.0"


GeorgianaElena:
Unfortunately this still doesn't fix the deploy... Only the dep upgrade.
When I tried to deploy the `ccsf` hub manually I got this:
```
Error: UPGRADE FAILED: parse error at (basehub/charts/jupyterhub/templates/hub/_helpers-passwords.tpl:35): function "dig" not defined
```

But maybe it's just because I have something outdated in my deployment setup?

consideRatio:
> function "dig" not defined

This is because of an outdated version of `helm` - install 3.5+ and this error goes away. I wonder if that is also why you required to have the version specified in Chart.yaml?

GeorgianaElena:
Thanks a lot for the info and guidance @consideRatio!
I managed to deploy the ccsf hub, yaaay :tada:

> This is because of an outdated version of helm - install 3.5+ and this error goes away. I wonder if that is also why you required to have the version specified in Chart.yaml?

* The `dig` error did indeed went away with the new helm version (`v3.6.3`) I installed :heart: 
* I re-tried to run `helm dep up` for the `daskub` chart without a version specified and it still threw the `improper constraint` error :(

choldgraf:
nice catch @GeorgianaElena :-)

damianavila:
> The dig error did indeed went away with the new helm version (v3.6.3) I installed

Should we add a note in https://pilot-hubs.2i2c.org/en/latest/reference/tools.html about the minimum version we suggest using?

consideRatio:
@damianavila I think it will be quickly outdated advice. If we add docs about it, I suggest it mention that one can run into issues having outdated versions of the `helm` or `kubectl` CLI and it's good to update them from time to time quite. The `helm` version required for z2jh specifically is 3.5.0 currently as is documented in z2jh.jupyter.org's landing page, but that is just one of several charts in the end and just one of a few tools we use.

PR #537, opened 2021-07-20
Add openscapes credentials to repo

yuvipanda:
These were never added to the repo!


GeorgianaElena (approved):
Yay! thanks @yuvipanda!

choldgraf:
sounds great - I'm not quite sure what this means (e.g., why we were able to deploy the hub at all given that we didn't have credentials for it), but this sounds like a good thing to do :-)

damianavila:
@choldgraf, the AWS clusters we deploy use kops, and kops saves the state of the cluster in an S3 bucket, so if you have access to the AWS account (and access to that bucket), you can actually get admin credentials using the kops CLI (exporting the kubeconfig file) and do stuff on the cluster: https://kops.sigs.k8s.io/cli/kops_export_kubeconfig/

damianavila:
More details about generating this file lives here: https://pilot-hubs.2i2c.org/en/latest/howto/operate/add-aws-hub.html#deploy-the-new-hub

PR #540, opened 2021-07-21
Label nodes with their instancegroup name

yuvipanda:
https://github.com/jupyterhub/grafana-dashboards
needs some label to distinguish different node pools,
and kops doesn't seem to add one by default.


yuvipanda:
I've already deployed this

sgibson91 (approved):
LGTM!

PR #541, opened 2021-07-21
Bump resource requests for prometheus

yuvipanda:
Earlier limits were too little, causing the server
to crash.


damianavila (approved):
LGTM since matches observed resource usage.

PR #544, opened 2021-07-22
Template updates after feedback

choldgraf:
Incorporate feedback from the OceanHackWeek group


choldgraf:
(this is a relatively minor update per some feedback so I'll plan to merge in a bit if nobody suggests otherwise!)

PR #546, opened 2021-07-22
Expose grafana as a service on COESSING hub

sgibson91:
Make grafana available at HUB_URL/services/grafana
instead of at grafana.HUB_URL
Minimises the need for extra A records in DNS.

### DIFF
--- a/config/hubs/pangeo-181919.cluster.yaml
+++ b/config/hubs/pangeo-181919.cluster.yaml
@@ -5,16 +5,6 @@ gcp:
   project: pangeo-181919
   cluster: pangeo-hubs-cluster
   zone: us-central1-b
-support:
-  config:
-    grafana:
-      ingress:
-        hosts:
-          - grafana.pangeo.2i2c.cloud
-        tls:
-          - secretName: grafana-tls
-            hosts:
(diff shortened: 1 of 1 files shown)


config/hubs/pangeo-181919.cluster.yaml
```diff
@@ -75,3 +68,11 @@ hubs:
               memory:
                 request: 1G
                 limit: 2G
+      grafana:
```
yuvipanda, line 71: This should be under `support.config`, since the grafana chart is a subchart of the support chart, not a hub template

yuvipanda:
This will expose the URL to both staging and prod clusters, which is fine I guess. The auth will still use the separate password - if we want it to use hub oauth, we need to explicitly set it up. But that can happen as a different PR!

If this works well, we should use this for all single-hub clusters. For stuff like pilot hubs and cloudbank we should still have a separate A record

sgibson91:
I don't think I'm going to merge this one as the COESSING workshop is now underway and the last thing I want to do is either stress Paige out or bring the hub down during a critical moment. But I still learned lots from making it and getting feedback on it!

Issue #535, opened 2021-07-20
Consider exposing grafana via `https://<hub domain>/services/grafana`


consideRatio:
Via this configuration, grafana can be exposed via a subpath under /services/grafana, where traffic is routed via JupyterHub's configurable proxy, which in turn exposes that path to the grafana service.

I've not looked into detail if this is already done or not or how we have it setup. I'm just dumping information into this issue at this point.

```yaml
jupyterhub:
  hub:
    # ref: https://jupyterhub.readthedocs.io/en/stable/reference/services.html#properties-of-a-service
    services:
      grafana:
        # This will make the CHP proxy let /services/grafana route to the
        # grafana service in the k8s namespace, which lets us make use of
        # JupyterHub's HTTPS setup without needing something like nginx-ingress
        # + cert-manager and additional ingress k8s resources.
        url: http://grafana
# Reference on the Grafana Helm chart's configuration options:
# https://github.com/helm/charts/blob/master/stable/grafana/values.yaml
grafana:
  # Reference on Grafana's configuration options:
  # http://docs.grafana.org/installation/configuration
  grafana.ini:
    server:
      domain: hub.neurohackademy.org
      # NOTE: Don't use %(protocol)s in root_url, but hardcode https. If not, it
      #       will when redirecting the user to external authentication set with
      #       a redirect back query parameter to use http instead of https,
      #       which will be wrong. This is because the TLS termination is done
      #       without Grafanas knowledge by the ingress controller. If we would
      #       specify protocol to be https, then it would want to do the TLS
      #       termination itself so that also would fail.
      root_url: 'https://%(domain)s/services/grafana'
      serve_from_sub_path: true
      enforce_domain: true
      enable_gzip: true
      router_logging: true
```

### Action points to take

1. Decide if there is action points to take.

choldgraf:
I think it's a cool idea! What would the alternative be?

1. No grafana at all
2. grafana.<somehub>.2i2c.cloud

Are there pros and cons for number 2 above vs. this issue's suggestion?

sgibson91:
> I think it's a cool idea! What would the alternative be?
> 
> 1. No grafana at all
> 2. grafana..2i2c.cloud
> 
> Are there pros and cons for number 2 above vs. this issue's suggestion?

We currently do 2 and @consideRatio mentioned in Slack that this approach avoids setting up multiple A records.

I like it, but how does it fit in with the way we configure grafana in the support chart? Or is this no longer required?

Example of current config:

[...]

PR #550, opened 2021-07-23
Document how to deploy the support chart and setup Grafana dashboards

sgibson91:
This PR documents the process of deploying the support chart and setting up grafana dashboards using jupyterhub/grafana-dashboards. Also automates installation of the jetstack helm repo in deploy-support instead of CI

Note: if #535 is accepted, the example config should be updated


docs/howto/operate/setup-grafana.md
```diff
@@ -0,0 +1,83 @@
+# Setting up Grafana Dashboards for a cluster
+
+This guide will walk through the steps required to setup a suite of Grafana dashboards for a cluster.
+
+## Before you start
+
+Make sure you are authenticated against the correct cloud platform and have selected the correct project to deploy to.
```
yuvipanda, line 7: `deploy-support` should automatically do this for you - https://github.com/2i2c-org/pilot-hubs/blob/a6f2e354399cc08275c16f49b0d92f75e11e6030/deployer/__main__.py#L46. It uses the credentials in `secrets/` and sets project info accordingly.

docs/howto/operate/setup-grafana.md
```diff
@@ -0,0 +1,83 @@
...
+            hosts:
+              - GRAFANA_URL
+```
+
+### Deploy the `support` chart via the `deployer`
+
+If you don't have the `cert-manager` chart added to your helm setup already, you can add it and update it like so:
+
+```bash
+helm repo add jetstack https://charts.jetstack.io
```
yuvipanda, line 44: Let's just automate this with `deploy-support` instead? We don't nee to add the other repos (for nginx-ingress, prometheus, etc) because helm does it for us from the `Chart.yaml`. We'll just have to recreate a tiny bit of that.

docs/howto/operate/setup-grafana.md
```diff
@@ -0,0 +1,83 @@
...
+You can do this by selecting the gear icon from the left-hand menu, and then selecting API keys.
+The key you create needs admin permissions.
+
+**Keep this key safe as you won't be able to retrieve it!**
+
+Some default grafana dashboards for JupyterHub can then be deployed using [`jupyterhub/grafana-dashboards`](https://github.com/jupyterhub/grafana-dashboards).
+
+1. Create a local clone of the repository
+2. Create an environment using `venv` or `conda`
+3. `pip` install the `jsonnet` package into your env
```
yuvipanda, line 82: We shell out to the jsonnet binary (https://github.com/jupyterhub/grafana-dashboards/blob/775a09f6a24689e42676b1709ec92631c4df4d63/deploy.py#L46), so you should install the go jsonnet binary instead https://github.com/google/jsonnet.

yuvipanda: I considered using the jsonnet python wrapper, but we are using the `-J` option for vendoring in the grafana definitions. I didn't want to re-implement that logic in python

sgibson91: Ah, so this worked for me because I had previously brew installed jsonnet

docs/howto/operate/setup-grafana.md
sgibson91: I had originally done this [here](https://github.com/2i2c-org/pilot-hubs/pull/548/commits/a2ddda1243d58cdc9d25f4af4205f108359a88c2), but wasn't certain that was the right choice and so migrated it into the action.yml file instead

docs/howto/operate/setup-grafana.md
sgibson91: Ah ok. I tend to stick with double-checking my auth before I do anything anyway, just in case! 😆

.github/actions/deploy/action.yml
```diff
@@ -13,8 +13,6 @@ runs:
       shell: bash
     - name: Fetch helm charts
       run: |
-        helm repo add jetstack https://charts.jetstack.io
-        helm repo update
```
damianavila, line 17: These lines lasted just a breath 😜

docs/howto/operate/setup-grafana.md
```diff
@@ -0,0 +1,66 @@
...
+Eventually, visiting `GRAFANA_URL` will present you with a login page.
+the username is `admin` and the password is located in `support/secrets.yaml` (`sops` encrypted).
+
+## Setting up Grafana Dashboards
+
+Once you have logged into grafana as the admin user, create a new API key.
+You can do this by selecting the gear icon from the left-hand menu, and then selecting API keys.
+The key you create needs admin permissions.
+
+**Keep this key safe as you won't be able to retrieve it!**
```
damianavila, line 58: Where we should save this key so others can access it?

docs/howto/operate/setup-grafana.md
```diff
@@ -0,0 +1,66 @@
...
+
+Some default grafana dashboards for JupyterHub can then be deployed using [`jupyterhub/grafana-dashboards`](https://github.com/jupyterhub/grafana-dashboards).
+
+1. Create a local clone of the repository
+2. Install the [`jsonnet` binary](https://github.com/google/jsonnet#packages).
+   Homebrew is the best option if you're on MacOS.
```
damianavila, line 64: Maybe these two lines could be a note to highlight them?

.github/actions/deploy/action.yml
sgibson91: I know 😆 I was so concerned with fixing what I'd broke on merge that I kinda panicked over "best placement"

docs/howto/operate/setup-grafana.md
sgibson91: Good Q. What scenario do we envision others needing to use it once the grafana dashboards have been deployed? We either need to figure that out and define that process, or the comment should really read "Keep this key safe for the next step"

docs/howto/operate/setup-grafana.md
```diff
@@ -0,0 +1,66 @@
...
+   The Python package will not suffice here as we directly call the `jsonnet` library.
```
sgibson91, line 65:
````suggestion

```{note}
Homebrew is the best option if you're on MacOS.
The Python package will not suffice here as we directly call the `jsonnet` library.
```

````

@damianavila ?

docs/howto/operate/setup-grafana.md
damianavila: > What scenario do we envision others needing to use it once the grafana dashboards have been deployed? We either need to figure that out and define that process

My question was going in that direction, indeed.

docs/howto/operate/setup-grafana.md
damianavila: Super! Thanks!

choldgraf (approved):
This looks good to me! With one question: didn't we discuss deploying `grafana` as a service on top of the JupyterHubs? (e.g. at `myhub.org/grafana`). How does this PR fit in with those plans?

sgibson91:
@choldgraf Yeah we did, though as Yuvi pointed out in [this comment](https://github.com/2i2c-org/pilot-hubs/pull/546#issuecomment-885548152) the service method exposes grafana for all hubs on a cluster. So that method should only really be implemented for single-hub, single-community clusters, and clusters like 2i2c/Cloudbank so maintain a separate A record.

I think we should merge this now as our current best practice, and if we can solve (i) JupyterHub based authentication for accessing grafana, and (ii) separation of grafana dashboards for the hubs, then we can update as needed.

docs/howto/operate/setup-grafana.md
sgibson91: For now, I have added a note to the docs and will open an issue in team-compass

sgibson91:
I will merge this for now as there are issues opened to track the questions that have been raised here.

PR #553, opened 2021-07-26
Add actions for incident reports

choldgraf:
In https://github.com/2i2c-org/pilot-hubs/issues/524 we identified the need for some specific "actions" to take in writing incident reports so that we know when to close the reports. This adds such a list to our incident report template.


choldgraf:
merging this one as I think it's a pretty minor update

PR #557, opened 2021-07-27
Update new hub template to suggest two community representatives

choldgraf:
Updates our New Hub template to make it easier to list 2 community representatives, and also asks for a bit more information about each of them so that it's easier to track.


yuvipanda:
Thanks, @choldgraf

PR #558, opened 2021-07-27
Move carbonplan hub to eksctl

yuvipanda:
- Use jsonnet to generate the eksctl config YAML file.
  This reduces duplication a lot, and matches what we
  do with kops. We can easily set defaults, generate
  nodegroup names, and set cloud tags equivalent to our
  node labels / taints automatically.
- Install the cluster autoscaler as part of the support
  charts, since eksctl doesn't build that by default.
  It is turned off by default.
- No changes were needed for the hub configuration, which
  is great!

Inspired by https://github.com/2i2c-org/pilot-hubs/pull/436

~~TODO:~~
~~- [ ] eksctl nodegroups are immutable, so modifying them requires manually creating a new one and removing the old one. Figure out a plan for this~~
~~- [ ] Document how to create an eksctl cluster~~
~~- [ ] Add guidelines on when to use eksctl vs kops~~


support/Chart.yaml
```diff
@@ -21,4 +21,9 @@ dependencies:
   # that references this controller.
   - name: ingress-nginx
     version: 3.33.0
-    repository: https://kubernetes.github.io/ingress-nginx
\ No newline at end of file
+    repository: https://kubernetes.github.io/ingress-nginx
+
+  - name: cluster-autoscaler
```
consideRatio, line 26:
```suggestion
  # cluster-autoscaler for clusters where it isn't available.
  # https://github.com/kubernetes/autoscaler/tree/master/charts/cluster-autoscaler
  - name: cluster-autoscaler
```

eksctl/libsonnet/nodegroup.jsonnet
```diff
@@ -0,0 +1,32 @@
...
+  labels+: {
+	  // Add instance type as label to nodegroups, so they
+	  // can be picked up by the autoscaler
+    'node.kubernetes.io/instance-type': $.instanceType,
+  },
+  taints+: {},
+  tags: makeCloudLabels(self.labels) + makeCloudTaints(self.taints),
+  iam: {
+    withAddonPolicies: {
+      autoScaler: true,
```
consideRatio, line 29: Do we want `efs: true` here as well?

```yaml
      withAddonPolicies:
        autoScaler: true
        efs: true
```

consideRatio (approved):
From what I can tell, this looks good! I left a comment or two

eksctl/libsonnet/nodegroup.jsonnet
yuvipanda: That's only needed if we use the [EFS CSI Driver](https://docs.aws.amazon.com/eks/latest/userguide/efs-csi.html). We instead mount via NFS, so it is not needed.

yuvipanda:
I struck out the TODO items. Let's get this merged first, and then we can work on the subtasks.

choldgraf:
Is this waiting on anything? I see an approve from @consideRatio but I don't feel comfortable merging a PR when I don't understand what it does

consideRatio:
Same here, i just approved based on my understanding - +1 for self-merge!

yuvipanda:
Yeah, we should figure out policies in https://github.com/2i2c-org/team-compass/issues/175. I always feel uncomfortable self-merging, but with infrastructure as code that's often what needs to happen. I'll do that here for now.

damianavila:
Belated review (because of holidays 😉 )

This looks nice to me except for the fact you put the autoscaler under the support chart.
Can you elaborate on that decision? AFAIR, the support stuff is actually optional (right?) and it rings bells on my head that you could potentially deploy a broken? eksctl cluster if you missed adding the support chart? Or am I missing something here?

PR #559, opened 2021-07-27
Run helm dep up for support chart

yuvipanda:
GH Actions fails otherwise:
https://github.com/2i2c-org/pilot-hubs/runs/3171244245


sgibson91 (approved):
Good catch! 🎉

PR #561, opened 2021-07-28
Add backend configs for terraform remote state storage

### LINKED ISSUE #560: Terraform remote state backends: Dynamic configuration implementation and storage
# Background

I have been trying to work with terraform across two authentications for a while now. The current workflow looks like the below.

- I need to be logged into 2i2c to be able to access our remote state backend on `two-eye-two-see`
- My 2i2c account doesn't have access to the project where the infrastructure will be deployed, so...
- Print out an access token while logged into my 2i2c account and add it to the backend block in `main.tf`. Always be vigilant not to check this into version control.
- Log into another account and rerun `terraform init`, `terraform plan`, etc...

This was working fine until my token expired mid-way through a terraform action and then state lock got me into a _whole heap of trouble_! Thankfully I was the only one using that workspace and the infrastructure is definitely not ready for production deployment yet, so not much is lost.

It would be a lot easier in this case if the remote backend was stored in the same project I was trying to deploy to so I didn't have to worry about multi-authentication. So...

# How do we dynamically configure the remote backend?

First off, I tried to parameterise this problem. Have a variable such as `backend` that would point towards `two-eye-two-see` when set to `default`, and then set it to something else to point to another bucket somewhere else. I tried [every trick I could think of](https://blog.gruntwork.io/terraform-tips-tricks-loops-if-statements-and-gotchas-f739bbae55f9), but it was all quite hacky and didn't really work.

But, `terraform init` has an optional flag `-backend-config` that accepts a path to a `backend.hcl` file. This file is formatted similarly to a `tfvars` file with key-value pairs. This seems like the solution to the problem!

First off, I need to test to see if providing a backend config file on the command line will automatically overwrite what is provided in [`main.tf`](https://github.com/2i2c-org/pilot-hubs/blob/824d1dd53a6b5170a32a0458bcf2724f8a412a95/terraform/main.tf#L2-L5). If yes, then that's great! Behaviour won't change for the majority of the team.

# Actions

- [x] Create a storage bucket in Columbia project for remote state storage
- [x] Create a `backend.hcl` file pointing to Columbia bucket
- [x] Test this correctly overwrites the config in `main.tf` as expected

sgibson91:
This PR adds a backend config files for terraform that points to various buckets stored in Google projects. This is particularly useful when working on projects we don't have access to via our 2i2c accounts (e.g., Columbia/Pangeo) as we avoid having to work with multiple authentications.

To use a backend file, run:

```
terraform init -backend-config=backends/NAME_OF_FILE.hcl
```

Our default remote storage can be configured by

```
terraform init -backend-config=backends/default-backend.hcl
```

Remote storage for Pangeo can be configured by

```
terraform init -backend-config=backends/pangeo-backend.hcl
```

The `-reconfigure` option may be required.

fixes #560

### DIFF
--- a/docs/howto/operate/new-tf-cluster.md
+++ b/docs/howto/operate/new-tf-cluster.md
@@ -41,7 +41,7 @@ Then you can change into the terraform directory and initialise
 
 ```bash
 cd terraform
-terraform init -backend-config=backends/default-backend.hcl
+terraform init -backend-config=backends/default-backend.hcl -reconfigure
 ```
 
 ````{note}
@@ -51,7 +51,7 @@ This saves us the pain of having to handle multiple authentications as these sto
 For example, to work with Pangeo you would initialise terraform like so:
 
 ```bash
-terraform init -backend-config=pangeo-backend.hcl
+terraform init -backend-config=pangeo-backend.hcl -reconfigure
 ```
 ````
 
--- a/docs/howto/operate/new-tf-cluster.md
+++ b/docs/howto/operate/new-tf-cluster.md
@@ -31,7 +31,7 @@ Once you have created this file, open a Pull Request to the `pilot-hubs` repo fo
 
 ## Initialising Terraform
 
-The terraform state is located centrally in our `two-eye-two-see-org` GCP project, therefore you must authenticate `gcloud` to your `@2i2c.org` account before initialising terraform.
+Our default terraform state is located centrally in our `two-eye-two-see-org` GCP project, therefore you must authenticate `gcloud` to your `@2i2c.org` account before initialising terraform.
 
 ```bash
 gcloud auth application-default login
@@ -41,26 +41,18 @@ Then you can change into the terraform directory and initialise
 
 ```bash
 cd terraform
-terraform init
+terraform init -backend-config=backends/default-backend.hcl
 ```
 
 ````{note}
-If you are deploying a cluster to a project you have access to via a different account, such as a university-affliated account, there are a few extra steps to take.
-Hopefully, this will be a rare scenario.
+If you are working on a project which you cannot access with your 2i2c account, there are other backend config files stored in `terraform/backends` that will configure a different storage bucket to read/write the remote terraform state.
+This saves us the pain of having to handle multiple authentications as these storage buckets are within the project we are trying to deploy to.
 
-First, you will need to provide terraform with an [access token](https://www.terraform.io/docs/language/settings/backends/gcs.html#configuration-variables) to use the state files.
-You can generate one using the below commands and logging in with your 2i2c.org account.
+For example, to work with Pangeo you would initialise terraform like so:
 
 ```bash
-gcloud auth application-default login
-gcloud auth application-default print-access-token
+terraform init -backend-config=pangeo-backend.hcl
 ```
-
-Add the access token to the [terraform backend block](https://github.com/2i2c-org/pilot-hubs/blob/2ef8a4bf35bb5ee9bf04ab3db1218b8c183c5da2/terraform/main.tf#L2-L5) in `main.tf`.
-**DO NOT COMMIT THIS CHANGE.**
-Then run `terraform init` or `terraform init -reconfigure`.
-
-You can now login to your other gcloud account and proceed with the guide.
 ````
 
 ## Creating a new terraform workspace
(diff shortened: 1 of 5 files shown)


yuvipanda:
This is awesome!

Can you add a blurb to the docs too?

sgibson91:
@2i2c-org/tech-team can we all check this PR please? It changes how we initialise terraform so we can dynamically choose backends for terraform state storage

docs/howto/operate/new-tf-cluster.md
```diff
@@ -41,26 +41,18 @@ Then you can change into the terraform directory and initialise
...
 ```
-
-Add the access token to the [terraform backend block](https://github.com/2i2c-org/pilot-hubs/blob/2ef8a4bf35bb5ee9bf04ab3db1218b8c183c5da2/terraform/main.tf#L2-L5) in `main.tf`.
-**DO NOT COMMIT THIS CHANGE.**
-Then run `terraform init` or `terraform init -reconfigure`.
-
-You can now login to your other gcloud account and proceed with the guide.
 ````
 
 ## Creating a new terraform workspace
```
yuvipanda, line 58: Let's add a note that workspaces are per backend? So if you don't see a particular workspace, it might be because you aren't in the right backend.

Is there a way to dynamically switch between backends on the same checkout? Or do we need different git clones?

sgibson91: > Is there a way to dynamically switch between backends on the same checkout? Or do we need different git clones?

Backends doesn't have anything to do with git, you choose it when you run `terraform init -backend-config=path/to/config`

sgibson91: > Let's add a note that workspaces are per backend? So if you don't see a particular workspace, it might be because you aren't in the right backend.

Done!

yuvipanda: @sgibson91 if I was working on the pilot hubs infra (in workspace `pilot-hubs` repo), and want to work on justiceinnovation lab instead (in workspace `justiceinnovationlab`), I can run `terraform workspace switch justiceinnovationlab`. But that only works if they're in the same backend. If they're in a different backend, how would this switch happen?

sgibson91:
```
terraform init -backend-config=/path/to/backend/config/that/contains/that/workspace
terraform workspace select desiredworkspace
```

At the minute we only have two backends: our default one that contains all our currently known workspaces, and the Pangeo one on the Columbia project, that only contains one workspace for the Pangeo hubs.

sgibson91: I was having to re-run `terraform init` _a lot_, because every time I needed to update my `access_token` terraform would recognise this as a change in the backend and ask to be reinitialised. [This comment](https://github.com/hashicorp/terraform/issues/17288#issuecomment-363851018) might explain it a bit better.

sgibson91: This is why I originally didn't abstract out our default backend, because Pangeo is a special case.

sgibson91: We may also want to agree on some team process around whether we prefer `-reconfigure` or `-migrate-state` when rerunning `terraform init`. https://www.terraform.io/docs/cli/commands/init.html#backend-initialization

sgibson91: terraform is also smart enough to know which workspaces exist where

```
tf init -backend-config=backends/pangeo-backend.hcl -reconfigure

Initializing the backend...

Successfully configured the backend "gcs"! Terraform will automatically
use this backend unless the backend configuration changes.

The currently selected workspace (sandbox) does not exist.
  This is expected behavior when the selected workspace did not have an
  existing non-empty state. Please enter a number to select a workspace:
  
  1. default
  2. pangeo-hubs

  Enter a value:
```

yuvipanda: Reading it, I think we should suggest doing `terraform init -reconfigure`, since we'll just be switching workspaces. Let's document that as 'how to switch workspaces', and then we can merge this?

Thank you so much for working on this! It'll save a lot of pain going forward :D

sgibson91: > Let's document that as 'how to switch workspaces', and then we can merge this?

Done! Will merge tomorrow if I receive no further comments. I've linked to this in the team-sync issue and I want to make sure the team know this is happening since it changes our workflow.

yuvipanda:
Thanks a lot, @sgibson91! This is sweeeet!

damianavila:
This is great @sgibson91, thanks!!

PR #563, opened 2021-07-28
Use extraFiles feature to cull idle kernels

GeorgianaElena:
Fixes https://github.com/2i2c-org/pilot-hubs/issues/518

### DIFF
--- a/hub-templates/basehub/values.yaml
+++ b/hub-templates/basehub/values.yaml
@@ -1,19 +1,3 @@
-etcJupyter:
-  jupyter_notebook_config.json:
-    # if a user leaves a notebook with a running kernel,
-    # the effective idle timeout will typically be CULL_TIMEOUT + CULL_KERNEL_TIMEOUT
-    # as culling the kernel will register activity,
-    # resetting the no_activity timer for the server as a whole
-    MappingKernelManager:
-      # shutdown kernels after no activity
-      cull_idle_timeout: 3600
-      # check for idle kernels this often
-      cull_interval: 300
-      # a kernel with open connections but no activity still counts as idle
-      # this is what allows us to shutdown servers
-      # when people leave a notebook open and wander off
-      cull_connected: true
-
 nfsPVC:
   enabled: true
   shareCreator:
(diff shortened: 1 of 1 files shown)


yuvipanda:
This is great! Can you add a little docs too?

- [ ] (In pilot repo, for users) when culling happens
- [ ] (In pilot-hubs repo, for us) where it is configured, and how to raise it

docs/howto/configure/culling.md
```diff
@@ -0,0 +1,51 @@
...
+  jupyterhub:
+    cull:
+      # Cull after 30min of inactivity
+      every: 300
+      timeout: 1800
+      # No pods over 12h long
+      maxAge: 43200
+```
+
+More culling options and info about them, can be found in the [idle-culler documentation](https://github.com/jupyterhub/jupyterhub-idle-culler#readme).
```
damianavila, line 24:
```suggestion
More culling options and information about them can be found in the [idle-culler documentation](https://github.com/jupyterhub/jupyterhub-idle-culler#readme).
```

docs/howto/configure/culling.md
```diff
@@ -0,0 +1,51 @@
...
+          # shutdown kernels after no activity
+          cull_idle_timeout: 3600
+          # check for idle kernels this often
+          cull_interval: 300
+          # a kernel with open connections but no activity still counts as idle
+          cull_connected: true
+```
+
+### Note
+If a user leaves a notebook with a running kernel, the idle timeout will typically be `CULL_TIMEOUT` + `CULL_KERNEL_TIMEOUT`, as culling the kernel will register activity, resetting the `no_activity` timer for the server as a whole.
```
damianavila, line 50: `CULL_TIMEOUT` and `CULL_KERNEL_TIMEOUT`, both are not mentioned before... maybe you can reference these with the names you used in the examples yaml you provided above?

GeorgianaElena: Thanks @damianavila! I just added a commit that addresses this.

damianavila (approved):
This one looks OK to me.
Wondering how this change will impact our users... I mean, do we need to ANN this somehow to be sure we are not breaking established workflows?

choldgraf:
Good question - I guess the question is "what is the default behavior" for jupyterhub when this isn't configured? Because I assume that this what our behavior was, since this config wasn't taking effect right?

PR #567, opened 2021-07-30
Add admins to carbonplan hub

### LINKED ISSUE #566: Add JupyterHub admin users to CarbonPlan hub
# Summary

I'd like to add 3 folks to the list of JupyterHub admins on the CarbonPlan hub. Those are:

- @jhamman (me!)
- @orianac
- @tcchiao 

# User Stories

- We need the ability for a few folks on our core team to administer JupyterHub sessions.

# Acceptance criteria

- All users listed above should be able to access the admin dashboard on the hub

# Important information

# Tasks to complete

- [ ] add users to list of hub admins

choldgraf:
I tried to mimic how this is done on the OpenScapes hub

closes #566

### DIFF
--- a/config/hubs/carbonplan.cluster.yaml
+++ b/config/hubs/carbonplan.cluster.yaml
@@ -151,6 +151,16 @@ hubs:
             readinessProbe:
               enabled: false
             nodeSelector: {}
+            config:
+              Authenticator:
+                admin_users:
+                  - yuvipanda
+                  - choldgraf
+                  - GeorgianaElena
+                  - damianavila
+                  - sgibson91
+                  - jhamman
+
       dask-gateway:
         traefik:
           resources:
--- a/config/hubs/carbonplan.cluster.yaml
+++ b/config/hubs/carbonplan.cluster.yaml
@@ -153,14 +153,14 @@ hubs:
             nodeSelector: {}
             config:
               Authenticator:
-                allowed_users: &staging_users
+                allowed_users: &users
                   - yuvipanda
                   - choldgraf
                   - GeorgianaElena
                   - damianavila
                   - sgibson91
                   - jhamman
-                admin_users: *staging_users
+                admin_users: *users
 
       dask-gateway:
         traefik:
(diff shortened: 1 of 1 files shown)


yuvipanda:
Let's do it like https://github.com/2i2c-org/pilot-hubs/blob/151faa89e92d7ca596a31bbe9e58e16de758a2f7/config/hubs/2i2c.cluster.yaml#L50 instead? That'll mean only the listed users can log in (right now anyone with GitHub can login lol), and they can add more users in the admin panel when they need to log in.

yuvipanda:
Also https://github.com/2i2c-org/pilot-hubs/pull/558 is needed to deploy to the new carbonplan hub

choldgraf:
ok I think this is what you mean? Do we also need to do this in the openscapes hub? https://github.com/2i2c-org/pilot-hubs/blob/151faa89e92d7ca596a31bbe9e58e16de758a2f7/config/hubs/openscapes.cluster.yaml

config/hubs/carbonplan.cluster.yaml
```diff
@@ -151,6 +151,17 @@ hubs:
             readinessProbe:
               enabled: false
             nodeSelector: {}
+            config:
+              Authenticator:
+                allowed_users: &staging_users
```
yuvipanda, line 156:
```suggestion
                allowed_users: &users
```

config/hubs/carbonplan.cluster.yaml
```diff
@@ -151,6 +151,17 @@ hubs:
...
+                  - yuvipanda
+                  - choldgraf
+                  - GeorgianaElena
+                  - damianavila
+                  - sgibson91
+                  - jhamman
+                admin_users: *staging_users
```
yuvipanda, line 163:
```suggestion
                admin_users: *users
```

yuvipanda (changes requested):
Yep! I think we should do this for openscapes (and all our hubs really), since otherwise anyone with a github login can get in.

The * and & are YAML anchors that need to be unique within the document, so they were `<hubname>_users` in 2i2c.clusters.yaml. Not the case here though.

There's docs on adding new users in https://pilot.2i2c.org/en/latest/admin/howto/manage-users.html, but we should have equivalent in the '2i2c engineer handbook'

Issue #381, opened 2021-05-04
Use an AWS IAM Role / User to fetch kubernetes credentials in our deployer


yuvipanda:
# Background

For AWS, we've checked in a static version of kubeconfig [exported by kops](https://github.com/2i2c-org/pilot-hubs/pull/379/commits/c81b36921307033fc6c491d3913c249993314dc3), with a 30d lifetime. Instead, we should create an AWS IAM Role with just enough permissions to run `KUBECONFIG=secrets/farallon.yaml kops export kubecfg --admin=30m farallon-2i2c.k8s.local`, and use that instead.

# Expected timeline

The credentials for farallon in https://github.com/2i2c-org/pilot-hubs/pull/379 last 30days, so this is a ticking time bomb :D

# Steps to complete this goal

- [x] Make an AWS IAM Role (or User) that can fetch kubeconfig via kops
- [x] Modify our deployer to accept the role (or user) credentials instead of checking in a static kubeconfig file

damianavila:
Definitely! Checking in the static kubeconfig file, even when it is a quick way to do it and I think is OK for now for fast iteration, is somehow ringing 🔔  in my head 😉 .

damianavila:
> In farallon, we would need to create a new deployer user under the 2i2c-engineers group

Created a new user:
aws iam create-user --user-name deployer

Added the user to the 2i2c-engineers group
aws iam add-user-to-group --group-name 2i2c-engineers --user-name deployer

Checked if the user was added to the group
aws iam get-group --group-name 2i2c-engineers

Got the credentials:
aws iam create-access-key --user-name deployer | tee /tmp/farallon.json

damianavila:
So, we now have a PR for this one 😉 , #673.

PR #569, opened 2021-07-30
Deploy some more hub automatically

yuvipanda:
Ref https://github.com/2i2c-org/pilot-hubs/issues/507

I'm worried that incomplete tests might leave notebook
nodes running, but the culler should take care of those? We
must verify this.

Added automatic deployment for everything except farallon
and openscapes hubs - those are using `kops` and I think
their kubernetes credentials will have expired. #381 
tracks a fix.

 Once this is merged, PRs like https://github.com/2i2c-org/pilot-hubs/pull/580
can be deployed without needing a local setup

### DIFF
--- a/.github/workflows/deploy-carbonplan.yaml
+++ b/.github/workflows/deploy-carbonplan.yaml
@@ -1,0 +1,30 @@
+name: Deploy and Test carbonplan hub
+
+on:
+  push:
+    branches: [ master ]
+    paths:
+      - deployer/**
+      - hub-templates/**
+      - requirements.txt
+      - dev-requirements.txt
+      - config/secrets.yaml
+      - config/hubs/carbonplan.cluster.yaml
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: google-github-actions/setup-gcloud@master
+        with:
+          version: '290.0.1'
+          # This is used for KMS only
+          project_id: two-eye-two-see
+          service_account_key: ${{ secrets.GCP_KMS_DECRYPTOR_KEY }}
+          export_default_credentials: true
+      - uses: azure/setup-helm@v1
+      - uses: mdgreenwald/mozilla-sops-action@v1
+      - uses: ./.github/actions/deploy
+        with:
+          cluster: 'carbonplan'
--- a/.github/workflows/deploy-pangeo-181919.yaml
+++ b/.github/workflows/deploy-pangeo-181919.yaml
@@ -1,0 +1,30 @@
+name: Deploy and Test pangeo hub
+
+on:
+  push:
+    branches: [ master ]
+    paths:
+      - deployer/**
+      - hub-templates/**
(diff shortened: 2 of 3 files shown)


yuvipanda:
I thought I had named this branch cp-cd but apparently I was too sleepy to get it right

choldgraf:
I'm a bit confused - I though all hubs were auto deployed but maybe that is just for the pilot hubs?

It sounds like I need to do another deep dive in the pilot docs since it has been a while and they've changed a lot

yuvipanda:
@choldgraf Yeah, this repo basically deploys all our hubs now. I opened https://github.com/2i2c-org/pilot-hubs/issues/570 to rename this repo to clarify.

.github/workflows/deploy-carbonplan.yaml
```diff
@@ -0,0 +1,30 @@
...
+on:
+  push:
+    branches: [ master ]
+    paths:
+      - deployer/**
+      - hub-templates/**
+      - requirements.txt
+      - dev-requirements.txt
+      - config/secrets.yaml
+      - config/hubs/carbonplan.cluster.yaml
```
sgibson91, line 12: Is handling this "run on changes to a specific cluster.yaml file" the crux of the reason we are using workflow files per cluster rather than a matrix job?

.github/workflows/deploy-pangeo-181919.yaml
```diff
@@ -0,0 +1,30 @@
+name: Deploy and Test pangeo hub
```
sgibson91, line 1:
```suggestion
name: Deploy and Test COESSING hub
```

Just so we don't get confused with naming!

sgibson91 (approved):
LGTM!

.github/workflows/deploy-carbonplan.yaml
yuvipanda: @sgibson91 It wasn't intentional - i was just following the pattern @GeorgianaElena set up. It is nice to not run them when there's no changes though.

sgibson91: For sure! I was going to suggest using something like this action so set a conditional https://github.com/dorny/paths-filter
So we maintain the same behaviour of not doing a deploy when nothing's changed, but we can consolidate into a single workflow file and new clusters can just be added to the matrix element list.

sgibson91: (This shouldn't block the PR btw - merge away and we can open an issue to track this discussion if you like)

choldgraf (approved):
This looks good to me! I had the same thought that @sgibson91 did and I see she opened up #581 so I'm +1 on merging this and tracking that improvement there

sgibson91:
I've just opened #582 which would replace this PR if merged because it already contains the clusters @yuvipanda was adding in here

choldgraf:
Now that #582 has been merged, I _think_ this one can be closed, so closing this but please reopen if I'm wrong!

PR #571, opened 2021-08-02
ohw: Increase memory limit for the workshop

yuvipanda:
Ref https://github.com/2i2c-org/pilot-hubs/issues/549


yuvipanda:
I've deployed this for the workshop. Am merging so that other merges don't accidentally un-deploy this change

PR #572, opened 2021-08-02
Bump up resource requests for OHW again

yuvipanda:
See https://github.com/2i2c-org/pilot-hubs/issues/549#issuecomment-891307408


yuvipanda:
Already deployed!

PR #574, opened 2021-08-03
Add docs on manual setup of NFS server

sgibson91:
### Summary

This PR adds a step-by-step guide to manually deploying a NFS server, including any extra steps I had to take to configure this on the Pangeo project where external IPs are restricted.

### Reviewing

- @yuvipanda Do these steps look right, have I missed anything?
- @choldgraf With fresh eyes, do these make sense?

### DIFF
--- a/docs/howto/operate/manual-nfs-setup.md
+++ b/docs/howto/operate/manual-nfs-setup.md
@@ -6,17 +6,7 @@ More information about the NFS Server can be found in {ref}`/topic/storage-layer
 ## Deploy the host Virtual Machine
 
 We need to first deploy a small virtual machine with a persistent disk that will host the NFS server.
-Using `gcloud`, the command is:
-
-```bash
-gcloud compute instances create nfs-server-01 \
-  --image=ubuntu-2004-focal-v20210720 \
-  --image-project=ubuntu-os-cloud \
-  --machine-type=g1-small \
-  --boot-disk-device-name=nfs-server-01 \
-  --boot-disk-size=100GB \
-  --boot-disk-type=pd-standard
-```
+You can use `gcloud` commands to achieve this.
 
 ````{note}
 To find the values of `--image` and `--image-project`, run the following:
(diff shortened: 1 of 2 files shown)


yuvipanda (approved):
Some minor changes (based off https://github.com/utoronto-2i2c/jupyterhub-deploy/blob/staging/terraform/nfs-playbook.yaml) but overall looks good to me. This is far from an ideal setup - particularly,

1. We should use a separate data disk that is mounted at `/exports`
2. We should setup backups for it
3. We should use XFS there so we can setup project quota

However, the manual setup is a temporary workaround, *just* for Google Cloud until we figure out something better. For that, I think this is ok.

docs/howto/operate/manual-nfs-setup.md
```diff
@@ -0,0 +1,89 @@
...
+```{note}
+If the cluster you are setting up the NFS for is **private**, you will need to add the `--tunnel-through-iap` flag to the above command.
+This is because the VM will not have an external IP address and will therefore need to be routed differently.
+```
+
+1. Install the dependencies
+
+   ```bash
+   sudo apt update
+   sudo apt install nfs-kernel-server nfs-common xfsprogs
```
yuvipanda, line 59:
```suggestion
   sudo apt install nfs-kernel-server nfs-common
```

We don't actually setup XFS here, so this is not necessary

docs/howto/operate/manual-nfs-setup.md
```diff
@@ -0,0 +1,89 @@
...
+We need to first deploy a small virtual machine with a persistent disk that will host the NFS server.
+Using `gcloud`, the command is:
+
+```bash
+gcloud compute instances create nfs-server-01 \
+  --image=ubuntu-2004-focal-v20210720 \
+  --image-project=ubuntu-os-cloud \
+  --machine-type=g1-small \
+  --boot-disk-device-name=nfs-server-01 \
+  --boot-disk-size=100GB \
```
yuvipanda, line 17: Let's note that the data is also in the boot disk here, and suggest people make it bigger if they need to?

yuvipanda: @sgibson91 feel free to self-merge after adding this

PR #576, opened 2021-08-04
Open port 8080 and 22 outbound for OHW

yuvipanda:
Ref https://github.com/2i2c-org/pilot-hubs/issues/549#issuecomment-892276020


yuvipanda:
Merging since this is already deployed

PR #580, opened 2021-08-05
Update SWOT image to latest pangeo image

choldgraf:
cc'ing @yuvipanda so he can confirm that this is how the Pangeo hub image updates should work.

Also - should we suggest that the SWOT team try the configurator?


yuvipanda:
We should definitely ask them to try out the configurator! That frees us from having to do this kinda work. The configurator needs some work though.

https://github.com/2i2c-org/pilot-hubs/pull/569 will enable automatic deployment for the SWOT (aka meom-ige) hub. After that is merged, this can be merged. Regardless of what happens with the configurator, we should get this merged

sgibson91:
We merged #582 instead, so merging this PR should **only** trigger a deploy for the meom cluster

yuvipanda:
oh, you caught that @sgibson91. The meom-ige folks used to have a hub under 2i2c.cluster.yaml but we moved it to their own project. Guess we didn't decomission the old cluster!

Should definitely be in the other file.

sgibson91:
> Guess we didn't decomission the old cluster!

I did see the low-touch-cluster still going this morning in the GCP web GUI!

choldgraf:
@sgibson91 mmm, so does that mean that this PR is not correct?

sgibson91:
> @sgibson91 mmm, so does that mean that this PR is not correct? It would be a different cluster yaml file?

I think so, I think it should be `meom-ige.cluster.yaml` but it also sounds like there's some historical confusion and switching about where this config lives.

sgibson91:
If I understand correctly, the changes in your PR should be applied here

https://github.com/2i2c-org/pilot-hubs/blob/d6b109e425885232312e462d15dbcc81bf56ca23/config/hubs/meom-ige.cluster.yaml#L89-L91

And the block you have actually modified should be removed?

choldgraf:
@sgibson91 OK I think the latest commit does this, does that look right?

I assume we'll have to manually delete the old cluster? I am not sure how to do this these days though

sgibson91:
> I think the latest commit does this, does that look right?

Looks right to me, but I'll let @yuvipanda officially sign off on it. We don't need to delete the 2i2c cluster, but we will need to delete the `grenoble` deployment.

Something like:

```
helm delete grenoble --namespace grenoble
helm delete namespace grenoble
```

But we should do that after this has been successfully deployed and the SWOT folk confirm.

damianavila:
> but we will need to delete the grenoble deployment.

There might be some other old stuff (other namespaces) in the cluster (in fact I have an internal note to open a ticket about that), it is probably worth checking if there is anything else old (and not used) and deletes it as well...

yuvipanda (approved):
Let's open another issue to tackle decomissioning the existing hub fully? Otherwise this is good to go.

sgibson91:
I opened #612

yuvipanda:
Great! @choldgraf do you feel comfortable merging (and hence deploying) this and checking out if the new image works?

choldgraf:
OK let's see if this works 🙃

PR #583, opened 2021-08-05
Update deploy action to also be triggered by changes to workflow files

sgibson91:
We may want to update this to specify only changes the deploy-hubs.yaml. I'm not sure what is gained if we trigger on changes to, say, python-lint.yaml?

### DIFF
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -12,6 +12,7 @@ on:
       - config/secrets.yaml
       - config/hubs/**
       - .github/workflows/deploy-hubs.yaml
+      - .github/actions/deploy/**
 
 jobs:
   deploy:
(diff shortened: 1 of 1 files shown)


yuvipanda:
Makes sense to restrict it to just that file. Do we still use a separate action file? If so we should include that too.

sgibson91:
> Makes sense to restrict it to just that file. Do we still use a separate action file? If so we should include that too.

Yes we do, will add that

yuvipanda (approved):

yuvipanda:
Thanks, @sgibson91! I've hit approve, so merge whenever you think you can look at the deploys?

sgibson91:
Thanks! Will probably be tomorrow morning for me now :)

yuvipanda:
@sgibson91 yay! I am very happy with how the review process in this worked for me :)

yuvipanda:
I often find it slightly awkward to merge PRs because what if they broke deployment? Now that i"m just reviewing and not deploying, it's much easier to commit :D

PR #584, opened 2021-08-06
Add deployer key for meom-ige cluster

sgibson91:
This adds the deployer key (sops encrypted) for the meom-ige cluster to the repo for CI/CD workflows.

I extracted this from the meom-ige terraform workspace (there is also a meom workspace but that didn't have any outputs)


yuvipanda (approved):
Thank you!

PR #597, opened 2021-08-10
Add config and deploy Pangeo staging hub

### LINKED ISSUE #599: Staging Hub deployment for Pangeo
### Description

We should deploy a **staging hub** for Pangeo that has the same infrastructure setup on less-costly infrastructure. This may also generate some other tasks that we need to accomplish in order to get the base infrastructure running.

### Benefit

This will help us iterate more quickly and get feedback from the Pangeo team. It will also be a place where we can stage changes in the future without affecting `prod`, since Pangeo is a more complex and dynamic setup than most of our community hubs.

### Tasks to complete

- [x] Create initial staging deployment
  - [x] Figure out NFS configuration: https://github.com/2i2c-org/pilot-hubs/issues/599#issuecomment-903860481 
- [x] Merge in #597 
- [ ] https://github.com/2i2c-org/pilot-hubs/pull/651
- [ ] Iterate with feedback from @rabernat 
  - [x] Choices for session types: https://github.com/2i2c-org/pilot-hubs/issues/599#issuecomment-911597920
- [ ] Confirm that major functionality works as expected

  

---

### Updates

- 2021-08-24: We've got a hub deployed but the user servers aren't being created properly due to some NFS errors. We've agreed to **try fixing this for two weeks** (@yuvipanda will give this a shot). If we cannot fix it after that time, we will:
  - Re-deploy the Pangeo hub using Google File Storage
  - Track the deployment of in-cluster NFS as a separate enhancement here: #50  
- 2021-08-31: in the sprint planning meeting today, we discussed that, now that NFS is ready to go (#50) we should be ready to review this PR and merge it in, and then ask Pangeo folks to take a look at the hub and make sure it looks good. In a future step, we will finish up #598 and deploy it, but that's not necessary for the initial deployment

sgibson91:
This PR adds a new `.cluster.yaml` file that points to the private cluster in the Pangeo/Columbia project. This file contains the config to deploy a staging hub.

fixes #599 
needs #613 to be merged


config/hubs/pangeo-hubs.cluster.yaml
```diff
@@ -0,0 +1,73 @@
...
+                  - jhamman
+                  - rabernat
+                  - TomAugspurger
+                  - scottyhq
+                  - sgibson91
+                  - yuvipanda
+                  - damianavila
+              GitHubOAuthenticator:
+                allowed_organizations:
+                  # Allow access to any member of the us-central1-b-gcp team
```
yuvipanda, line 49: Let's make a note that this doesn't have any effect right now?

yuvipanda (approved):
Thanks, @sgibson91!

config/hubs/pangeo-hubs.cluster.yaml
sgibson91: Yeah, I'm having trouble with GitHub and my local copy because I've removed this and pushed, but the commit isn't showing up here. Might close this and start afresh (silly GitHub outage!)

sgibson91:
I am having trouble with the NFS server again. I either missed a step when writing the docs or this is related to the private network.

```
Events:
  Type     Reason       Age                 From                                   Message
  ----     ------       ----                ----                                   -------
  Normal   Scheduled    116s                gke.io/optimize-utilization-scheduler  Successfully assigned staging/nfs-share-creator-9gdfb to gke-pangeo-hubs-cluster-core-pool-6f942896-gtqk
  Warning  FailedMount  51s (x8 over 115s)  kubelet                                MountVolume.SetUp failed for volume "home-base" : mount failed: exit status 1
Mounting command: /home/kubernetes/containerized_mounter/mounter
Mounting arguments: mount -t nfs nfs-server-01:/export/home-01/homes/ /var/lib/kubelet/pods/8e9ae982-6e80-48f3-b61e-59dc23db2549/volumes/kubernetes.io~nfs/home-base
Output: Mount failed: mount failed: exit status 32
Mounting command: chroot
Mounting arguments: [/home/kubernetes/containerized_mounter/rootfs mount -t nfs nfs-server-01:/export/home-01/homes/ /var/lib/kubelet/pods/8e9ae982-6e80-48f3-b61e-59dc23db2549/volumes/kubernetes.io~nfs/home-base]
Output: mount.nfs: access denied by server while mounting nfs-server-01:/export/home-01/homes/
```

sgibson91:
> what are the user permissions on `:/export/home-01/homes/`?

`chmod 0700` and `chown 100`:1000`

> It is also possible that the `10.0.0.0/8` IP range we have 'allowed' in our NFS setup's '/etc/exports' is not including the k8s nodes.

Oooh, this might need to be `10.128.0.0/8` instead

sgibson91:
So I changed the IP address in `/etc/exports` to `10.128.0.0/16` (because the current node and nfs server are at `.13` and `.10` respectively), re-ran `sudo exportfs -ra` on the nfs server and did another deploy, but still no luck

yuvipanda:
@sgibson91 I think `10.0.0.0/8` includes `10.8.0.0/24` and `10.128.0.0/8` - `10.128.0.0/8` and `10.0.0.0/8` might actually be equivalent. CIDR notation is often very confusing, I have found https://cidr.xyz/ to be a good tool to help me understand 'is this IP part of this range?'.

> chmod 0700 and chown 100:1000`

Is it `chown 1000:1000` or `chown 100:1000`?

sgibson91:
### Debugging with Yuvi

- We tried manually running the mount command while logged into the server and received the same permission denied error which indicated it was not a networking issue
- We tried various log outputs and tracing commands to try and get more information, but nothing was very helpful
- After more poking around and not getting anywhere, we decided that it would be best to try to automate this to save future selves time
- Where to automate: terraform & ansible or within Kubernetes?
  - We could deploy the VM with terraform and then setup the NFS server using ansible
    - Ansible is yet another tool to add to our stack though
  - Quansight have a Terraform config to deploy an NFS server in the k8s cluster
    - On closer inspection, this was "just" a PVC, Service and Deployment definition. Rather than add this to terraform, we can convert to YAML and deploy the NFS server as part of our support chart (where we're keeping other one-off-per-cluster deployment things)
- Wrote some draft YAML to deploy the NFS server into the k8s cluster and tested with `kubectl apply -f`
- Tried doing another deploy and this time we got a different error! The process was basically timing out
- This now _could_ be network-related so we disabled the network policy in the cluster
  - We expected all the calico pods to disappear from the kube-system namespace and they didn't, so this may need more poking
  - Also unexpectedly triggered a node upgrade that lasted longer than the terraform process, so we ended up in a weird state for a while
- Last thing we tried before agreeing to take a break was to ssh into a node and try the mount command from there. We saw the same hanging/probably gonna time out behaviour

sgibson91:
# Sarah doing some more testing

I'm going to try a few testing scenarios on `two-eye-two-see-sandbox` to see if I can debug the NFS issue further

### Test scenario 1

| Private cluster? | Network Policies enabled? |
| --- | --- |
| No | No |

NFS mounting failed due to server resolution timeout. Output from describing the share-creator pod

```
Events:
  Type     Reason       Age              From                                   Message
  ----     ------       ----             ----                                   -------
  Normal   Scheduled    7s               gke.io/optimize-utilization-scheduler  Successfully assigned testing/nfs-share-creator-qckhn to gke-testhub-cluster-core-pool-01e7dc9d-c1wb
  Warning  FailedMount  3s (x4 over 7s)  kubelet                                MountVolume.SetUp failed for volume "home-base" : mount failed: exit status 1
Mounting command: /home/kubernetes/containerized_mounter/mounter
Mounting arguments: mount -t nfs nfs-server-01:/export/home-01/homes/ /var/lib/kubelet/pods/150e4c95-7d25-465f-aa0d-3fddb9a02b2c/volumes/kubernetes.io~nfs/home-base
Output: Mount failed: mount failed: exit status 32
Mounting command: chroot
Mounting arguments: [/home/kubernetes/containerized_mounter/rootfs mount -t nfs nfs-server-01:/export/home-01/homes/ /var/lib/kubelet/pods/150e4c95-7d25-465f-aa0d-3fddb9a02b2c/volumes/kubernetes.io~nfs/home-base]
Output: mount.nfs: Failed to resolve server nfs-server-01: Name or service not known
```

sgibson91:
The most recent commits ([`c330c12`](https://github.com/2i2c-org/pilot-hubs/pull/597/commits/c330c1275b7a37f1bb5367b1fee70988d68e6f1b) and [`999a178`](https://github.com/2i2c-org/pilot-hubs/pull/597/commits/999a1787060b82a0065ce409df648e524cf9e005)), disable the nfs-share-creator and configures the singleuser to provision an NFS-based PVC.

This change results in each user provisioning a PVC and we should discuss if that is a sensible decision.

cc/ @consideRatio

config/hubs/pangeo-hubs.cluster.yaml
```diff
@@ -0,0 +1,74 @@
...
+                  logo_url: "https://coessing.files.wordpress.com/2016/08/ghana-logo-21.png?w=262&h=376&zoom=2"
+                designed_by:
+                  name: 2i2c
+                  url: https://2i2c.org
+                operated_by:
+                  name: 2i2c
+                  url: https://2i2c.org
+                funded_by:
+                  name: Pangeo
+                  url: https://pangeo.io
```
sgibson91, line 40: @rabernat Is this correct reference for funding source or should it be Columbia and/or something else?

rabernat: Pangeo is not really a funding agency. We _receive_ funding for various Pangeo-related projects from sponsors.

The current funding source is The Gordon and Betty Moore Foundation (https://www.moore.org/), so that's probably what should go here for now.

Going forward, we will soon switch to NSF Earthcube funding.

config/hubs/pangeo-hubs.cluster.yaml
```diff
@@ -0,0 +1,74 @@
+name: pangeo-hubs
+provider: gcp
+gcp:
+  key: secrets/pangeo-hubs.json
+  project: pangeo-integration-te-3eea
+  cluster: pangeo-hubs-cluster
+  zone: us-central1-b
+support:
+  nfs-server-provisioner:
+    enabled: true
```
sgibson91, line 10:
```suggestion
support:
  config:
    nfs-server-provisioner:
      enabled: true
```

choldgraf:
@sgibson91 do you think you could provide some guidance for what kind of review you're looking for here? E.g., general thoughts, "go try out the hub here and give feedback", "I have this specific concern" etc?

sgibson91:
> This change results in each user provisioning a PVC and we should discuss if that is a sensible decision.

This is what needs discussing right now. There isn't a hub any more because I tried to do a deploy by pairing this PR with #613 and ran up against problems in https://github.com/2i2c-org/pilot-hubs/pull/613#issuecomment-901956729

yuvipanda:
> This change results in each user provisioning a PVC and we should discuss if that is a sensible decision.

If we have one PVC per user, that means we can't map the users back to their data if the PVC objects are deleted! This adds another thing we must back up (PVC objects)

Can we instead have one PVC per hub that is bound to NFS, and continue to use subpath? This would be taking the approach we currently have (https://github.com/2i2c-org/pilot-hubs/blob/master/hub-templates/basehub/templates/nfs-pvc.yaml) and just getting rid of the explicit PV, and setting storageClass on the PVC.

sgibson91:
> Can we instead have one PVC per hub that is bound to NFS, and continue to use subpath?

@consideRatio Could you help me configure this please?

consideRatio:
> @consideRatio Could you help me configure this please?

Sure, but I want to challenge the idea of doing that a bit. Technically, one should then keep creating a PVC dedicated to the Hub as `nfsPVC.enabled=true` has done. This PVC should have a `storageClass=nfs`, and it should no longer create a `PV` that it references etc.

### A hidden assumption?

> If we have one PVC per user, that means we can't map the users back to their data if the PVC objects are deleted!

I think there are some assumptions being made here that I'd like to surface. A deleted PVC can but don't have to imply deleted files. A [storageClass can be defined with `reclaimPolicy: Retain`](https://kubernetes.io/docs/concepts/storage/storage-classes/), which I think is just a default of how the created PVs `reclaimPolicy` will be configured which is what actually matters. There is [documentation about this here](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reclaiming).

### Additional input of relevance

- If you have a per hub PVC, each home folder would be located under `/nfs-exports/pvc-................/user1`. One could have quotas per hub somewhat simple then, and you could configure something like `jupyterhub-sftp` to work well with that as you could locate the users folder without knowing the user's associated PVC name.
- If you have a per user PVC, each home folder would be located under `/nfs-exports/pvc-................` directly, and one could have quotas per user but not per hub.
- There are [Volume Snapshots](https://kubernetes.io/docs/concepts/storage/volume-snapshots/). I have not used them or read much about them. Planning for potential use of them may influence what one deems to be preferred in this decision as well.

### Action point

- Make a decision on having a single PVC for home folders per hub or per user in the hub
- Make a decision on a preferred reclaimPolicy for the `nfs` storageClass and its dynamically created PVs
- While we are at it, make a preferred reclaimPolicy for the `standard` storageClass and its dynamically created PVs, the default is `Delete` on GCP at least and I personally think that is to be preferred to reduce maintenance burden.

yuvipanda:
> If you have a per user PVC, each home folder would be located under `/nfs-exports/pvc-................` directly, and one could have quotas per user but not per hub.

This would be ideal, but the problem is that the pvc-XXXXX is an unpredictable random string. The failure case here is not the PVC being manually deleted, but the entire *cluster* being deleted for whatever reason (has happened many times!). If that happens, we can not reassociate the users with their home directories, which becomes a big problem. If we can introduce deterministic naming to the provisioner, we can do one PVC per user and all these concerns disappear.

I opened an issue about it in the former home of the external NFS provisioner laying out the issues, and the possible solutions (https://github.com/kubernetes-retired/external-storage/issues/1281)

choldgraf:
@yuvipanda and @consideRatio do you think that you two could converge on a specific recommendation for this PR and https://github.com/2i2c-org/pilot-hubs/pull/613, and then we can decide on next steps from there so that @sgibson91 can continue implementing this?

sgibson91:
Initial deployment of the staging hub: https://staging.pangeo.2i2c.cloud/

Note the spawn currently fails so the NFS still needs some tweaking. Not sure if that needs to happen here or in #613

sgibson91:
```
Spawn failed: (422) Reason: error HTTP response headers: HTTPHeaderDict({'Audit-Id': '2a805ba7-085a-475b-8d71-b2bc69696ceb', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'X-Kubernetes-Pf-Flowschema-Uid': 'fa135df3-09aa-4e18-b9c7-367f9d15e2e8', 'X-Kubernetes-Pf-Prioritylevel-Uid': 'b756986c-8baa-49d5-aec9-0c879d4ede21', 'Date': 'Mon, 23 Aug 2021 15:14:35 GMT', 'Content-Length': '565'}) HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Pod \"jupyter-sgibson91\" is invalid: [spec.containers[0].volumeMounts[1].name: Not found: \"home\", spec.containers[0].volumeMounts[2].name: Not found: \"home\"]","reason":"Invalid","details":{"name":"jupyter-sgibson91","kind":"Pod","causes":[{"reason":"FieldValueNotFound","message":"Not found: \"home\"","field":"spec.containers[0].volumeMounts[1].name"},{"reason":"FieldValueNotFound","message":"Not found: \"home\"","field":"spec.containers[0].volumeMounts[2].name"}]},"code":422}
```

hub-templates/basehub/values.yaml
```diff
@@ -27,6 +27,9 @@ nfsPVC:
     # MUST HAVE TRAILING SLASH
     baseShareName: /export/home-01/homes/
 
+# In-cluster NFS storage
+nfsGanesha:
+  enabled: false
```
consideRatio, line 32: I was thinking about a more explicit name than this. I miss two aspects captured in this naming:
- this influence 2i2c specific templates, not a subchart like `nfs-server-provisioner` running NFS ganesha
- this influence the creation of a PV resource specifically.

In practice I can't come up with good names without breaking misc patterns in how config has been named, but I think what I suggest is...


```suggestion
# Configuration related to integrating with the nfs-server-provisioner Helm chart that run
# nfs-ganesha and provides the custom "nfs" StorageClass. Together they provide in cluster
# NFS storage for PVs configured with a storageClassName of "nfs".
nfsGanesha:
  enabled: false
```

hub-templates/basehub/templates/nfs-ganesha.yaml
```diff
@@ -0,0 +1,13 @@
+{{ if .Values.nfsGanesha.enabled }}
```
consideRatio, line 1:
```suggestion
{{- if .Values.nfsGanesha.enabled -}}
```

hub-templates/basehub/templates/nfs-ganesha.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: home-nfs
+spec:
+  accessModes:
+    - ReadWriteMany
+  storageClassName: "nfs"
+  resources:
+    requests:
+      storage: {{ .Values.nfsGanesha.size }}
+{{ end }}
```
consideRatio, line 13:
```suggestion
{{- end }}
```

hub-templates/basehub/templates/nfs-ganesha.yaml
consideRatio: My rules of thumb:
- always chomp left
- always chomp right before the first content to be rendered in a YAML file

yuvipanda:
@consideRatio I renamed it to `inclusterNFS`, as that is more descriptive than `nfsGanesha`.

PR #603, opened 2021-08-12
Updated otter-grader to 2.2.5

sean-morris:
The otter-grader version is updated to 2.2.5, cloudbank and 2i2c cluster images a built and deployed, and the image tag
version is updated.

Fixes this issue [2i2c.org/pilot/86](https://github.com/2i2c-org/pilot/issues/86)

### DIFF
--- a/hub-templates/basehub/values.yaml
+++ b/hub-templates/basehub/values.yaml
@@ -106,7 +106,7 @@ jupyterhub:
       hub.jupyter.org/node-purpose: user
     image:
       name: set_automatically_by_automation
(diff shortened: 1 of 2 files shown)


yuvipanda:
Thanks, @sean-morris! Based on the guidelines for review we're developing in https://github.com/2i2c-org/team-compass/issues/175, I'd normally have 'approved' this pull request and you can deploy it at your convenience so you can make sure that the deployment in GitHub actions works fine. However, I just merged this to unblock use of new otter in the ccsf  hub, as it might be urgent.

PR #604, opened 2021-08-12
Updated otter-grader and bumped image tag

sean-morris:
After the rpy2 dependency [otter-grader issue](https://github.com/ucbds-infra/otter-grader/issues/293) was resolved, I update the version here to 2.2.6 as well as bumped the image tag.


damianavila (approved):
Similar to #603, changes LGTM.

PR #605, opened 2021-08-13
Use spot instances for dask workers

yuvipanda:
- Don't change the instance type of the dask workers - so no
  mixed instances. This is a straight port of what we have now
  to just use spot instances.
- x1 spot instances can't be created due to quota issues,
  so we remove them from the pool.

Fixes https://github.com/2i2c-org/pilot-hubs/issues/490

### DIFF
--- a/eksctl/carbonplan.jsonnet
+++ b/eksctl/carbonplan.jsonnet
@@ -7,16 +7,24 @@ local clusterRegion = "us-west-2";
 local masterAzs = ["us-west-2a", "us-west-2b", "us-west-2c"];
 local nodeAz = "us-west-2a";
 
-// Node definitions for use with dask and notebook nodes
-// These are merged in with the defaults for either node type,
-// and so can contain any overrides.
-local nodes = [
+// Node definitions for notebook nodes. Only `instanceType` is
+// supported as a property.
+local notebookNodes = [
     { instanceType: "r5.large" },
     { instanceType: "r5.xlarge" },
     { instanceType: "r5.2xlarge" },
     { instanceType: "r5.8xlarge" },
     { instanceType: "x1.16xlarge" },
     { instanceType: "x1.32xlarge" }
+];
+
+// Node definitions for notebook nodes. Only `instanceType` is
+// supported as a property.
+local daskNodes = [
+    { instanceType: "r5.large" },
+    { instanceType: "r5.xlarge" },
+    { instanceType: "r5.2xlarge" },
+    { instanceType: "r5.8xlarge" },
 ];
 
 cluster {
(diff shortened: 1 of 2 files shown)


yuvipanda:
I've deployed and tested this.

yuvipanda:
Issues to create from here:

- Fix x1 instances, for both notebooks and dask
- Consider using mixed-instance types, to ensure better pricing & availability of nodes

choldgraf:
@yuvipanda I've created a tracking issue for this one, think we could use that to track the extra items you wanted to follow-up on? And merge this PR since it is deployed? https://github.com/2i2c-org/pilot-hubs/issues/607

eksctl/carbonplan.jsonnet
```diff
@@ -19,6 +18,15 @@ local nodes = [
     { instanceType: "x1.32xlarge" }
 ];
 
+// Node definitions for notebook nodes. Only `instanceType` is
+// supported as a property.
```
consideRatio, line 22:
```suggestion
// Node definitions for dask worker nodes. Only `instanceType` is
// supported as a property.
```

eksctl/carbonplan.jsonnet
```diff
@@ -60,12 +69,12 @@ cluster {
                 "hub.jupyter.org/dedicated": "user:NoSchedule"
             },
 
-        } + n for n in nodes
+        } + n for n in notebookNodes
```
consideRatio, line 72: I don't know jsonnet, but there is an inconsistency here.

Below it sais `} for n in daskNodes`, but here it sais `} + n for n in notebookNodes`. I'll go ahead and guess this is wrong, but perhaps its the other that is wrong? No clue.

```suggestion
        } n for n in notebookNodes
```

eksctl/libsonnet/nodegroup.jsonnet
```diff
@@ -15,12 +15,12 @@ local makeCloudTaints(taints) = {
   availabilityZones: [],
   minSize: 0,
   desiredCapacity: self.minSize,
-  instanceType: '',
   volumeSize: 80,
   labels+: {
 	  // Add instance type as label to nodegroups, so they
-	  // can be picked up by the autoscaler
-    'node.kubernetes.io/instance-type': $.instanceType,
+	  // can be picked up by the autoscaler. If using spot instances,
```
consideRatio, line 21:
```suggestion
    // can be picked up by the autoscaler. If using spot instances,
```

Indentation nitpick

consideRatio (commented):
With resolved comments, this LGTM

eksctl/carbonplan.jsonnet
yuvipanda: @consideRatio yeah, I was trying to be 'clever' and just be able to set `instanceType` for both non-spot and spot node pools. However, this becomes unmaintainable soon, so I've reverted that. Now the config is the same as what you'll pass to eksctl.

yuvipanda:
Thanks a lot, @consideRatio! what do you think of it now?

consideRatio (approved):
LGTM!

I don't feel comfortable in this situation: either I press merge or I don't.

If I merge, I'm not feeling comfortable as I don't know what happens in the CI system and feel that I have a responsibility if I merge something and the CI system messes things up in a deployment or something happen making me think so.

If I don't press merge, I feel that it would slowed down a process that isn't worth slowing down after one approved review in this case.

I'll opt to not press merge this time and count on self-merge is to be considered an acceptable policy given this PR review approval.

choldgraf:
@consideRatio we're discussing that topic over in here too: https://github.com/2i2c-org/team-compass/issues/175

PR #606, opened 2021-08-13
Change default value for enable_network_policy to false

sgibson91:
### Summary

This PR changes the default value of the `enable_network_policy` variable in our terraform config to false and hence also replicates the default behaviour of the underlying terraform resource. I think this also follows logically that if we have a Boolean flag to enable something, then we should explicitly enable it, rather than explicitly disable it.

### What's changed?

- default value of `enable_network_policy` in `terraform/gcp/variables.tf` changed to false

### What should a reviewer concentrate their feedback on?

- [ ] Does this change make sense?
- [ ] Everything looks ok?

### DIFF
--- a/terraform/gcp/projects/hackathon-2i2c-project-alpha.tfvars
+++ b/terraform/gcp/projects/hackathon-2i2c-project-alpha.tfvars
@@ -1,6 +1,8 @@
 prefix     = "alpha"
 project_id = "hackathon-2i2c-project-alpha"
 
+enable_network_policy = true
+
 # Inane CPU requests mean we need at least 3 CPUs for a base node?!?!
 # But, we can't have custom machine sizes with odd number of CPUs -
 # only even numbers. So we go with 4. 3840 is the smallest amount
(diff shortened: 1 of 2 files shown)


yuvipanda (changes requested):
It's disabled when you create a GKE cluster via the console or terminal, so having the same here makes sense to me.

What is this set to for *current* clusters? We should make sure that whatever values they have now (true or false) should continue to be the same after this is deployed.

sgibson91:
> What is this set to for _current_ clusters?

If the variable is not explicitly listed as `enable_network_policy = false` in the tfvars file, then it will be set to `true`. So we should set `enable_network_policy = true` in all the tfvars files unless otherwise stated.

yuvipanda:
@sgibson91 yeah, that makes sense. Happy to approve this once that's changed!

sgibson91:
I could only find one cluster file that didn't explicitly set enable_network_policy - have pushed that change :)

yuvipanda (approved):

PR #613, opened 2021-08-17
Use nfs-server-provisioner subchart in support

sgibson91:
This PR adds the `nfs-server-provisioner` chart as a dependency of `support` and allows PVCs to be provisioned as an nfs type. We would no longer need to manually provision an NFS server.

### TODOs

- [x] Figure out a cleaner install procedure and document it

_New issues:_

- Gather knowledge on what provisioning NFS storage in this manner means https://github.com/2i2c-org/pilot-hubs/issues/629
- Create reference documentation for the above https://github.com/2i2c-org/pilot-hubs/issues/628

Fixes https://github.com/2i2c-org/pilot-hubs/issues/50


support/Chart.yaml
```diff
@@ -29,3 +29,8 @@ dependencies:
     version: 9.10.3
     repository: https://kubernetes.github.io/autoscaler
     condition: cluster-autoscaler.enabled
+
+  # In-cluster NFS server provisioner
+  - name: nfs-server-provisioner
+    version: 1.3.1
+    repository: https://kvaps.github.io/charts
```
consideRatio, line 36: Probably make sense to have this be something one can opt-in/opt-out from when installing the support chart and all its dependencies.

```suggestion
    repository: https://kvaps.github.io/charts
    condition: nfs-server-provisioner.enabled
```

support/values.yaml
```diff
@@ -71,3 +71,21 @@ grafana:
...
           editable: false
+
+nfs-server-provisioner:
+  storageClass:
+    # Users will have to opt-in to using NFS-type storage under
+    # singleuser.storage.dynamic.storageClass
+    name: nfs
+    create: true
+    defaultClass: false
+    # Mount Options: default v3, should we use v4?
```
yuvipanda, line 82: We should definitely use v4! Lots of improvements in the protocol, leads to fewer 'stuck' clients

support/Chart.yaml
```diff
@@ -29,3 +29,9 @@ dependencies:
...
+    condition: nfs-server-provisioner.enabled
```
yuvipanda, line 37: Let's turn this off by default?

yuvipanda (changes requested):
This is great! Thank you, @sgibson91 and @consideRatio!

I've a few questions about how the data is stored longer term.

1. What are the common ways data loss can occur? Does deleting the PV delete the backing disk? 
2. Can we resize them later if required? If so, how? 
3. Will we have one disk per hub or one per cluster?
4. How can we do backups?

Losing user home dirs is always the worst case scenario, so I just want to make sure we understand what is happening here.

None of these need to block this PR, though.

support/Chart.yaml
sgibson91: Done in [`8242a88`](https://github.com/2i2c-org/pilot-hubs/pull/613/commits/8242a88778bdc968b8fa0b934ff3dccb32814c94))

support/values.yaml
sgibson91: Done in [`b714fdc`](https://github.com/2i2c-org/pilot-hubs/pull/613/commits/b714fdc74179a44eff9aec0a377764129d8477c7)

support/values.yaml
```diff
@@ -71,3 +71,22 @@ grafana:
...
+nfs-server-provisioner:
+  enabled: false
+  storageClass:
+    # Users will have to opt-in to using NFS-type storage under
+    # singleuser.storage.dynamic.storageClass
+    name: nfs
+    create: true
+    defaultClass: false
+    mountOptions:
+      - vers=4
```
yuvipanda, line 84:
```suggestion
      - vers=4.2
      - soft
      - noatime
```

From https://github.com/2i2c-org/pilot-hubs/blob/35a830bf5894f8a96244e265c9c9c0968eda490c/hub-templates/basehub/values.yaml#L22, which is our current set of options

yuvipanda (changes requested):
One suggestion, but please merge after that! Thank you :)

support/values.yaml
```diff
@@ -71,3 +71,22 @@ grafana:
           access: proxy
           isDefault: true
           editable: false
+
```
sgibson91, line 74:
```suggestion

# Values reference for nfs-server-provisioner
# https://github.com/kubernetes-sigs/nfs-ganesha-server-and-external-provisioner/blob/master/deploy/helm/values.yaml
```

sgibson91:
I'm noticing some circular dependencies between running `python deployer deploy` and `python deployer deploy-support`. With NFS enabled, the support chart needs to be deployed first so the hub-dir-db can be granted an NFS pvc, but that leaves the grafana and prometheus pods pending and hence the install fails. Similarly, deploying the *hub chart first leaves the hub pod in a pending state. So I currrently don't have a way of deploying either the support chart or hub-template chart first without a failure.

Possible solutions:

* Give all support chart dependencies an `condition: <chart name>.enabled` flag so we can control the order of deployment
* Something else?

yuvipanda:
> but that leaves the grafana and prometheus pods pending and hence the install fails.

What are they waiting on, do you know? At first glance I don't know what there would be dependent on a hub being deployed.

Separately, *only* the user home directories should be on NFS. Putting our hub db on NFS will lead to poor performance, since sqlite (which is what we use with our hub) works poorly on NFS.

choldgraf:
I have a meta question - this PR is making this NFS behavior true for **all of our hubs**, not just for Pangeo, right? If so, is there a place in `pilot-hubs` where this new functionality could be documented? I worry that, 3 months from now, it will be hard to reverse-understand how our NFS stuff is being provisioned without some digging.

Does it make sense to have a `reference` section that simply describes the major components of our hub template? (e.g. "where does XXX come from?" "where is YYY configured?") or is it enough to just use lots of comments in our YAML?

sgibson91:
> this PR is making this NFS behavior true for all of our hubs, not just for Pangeo, right?

Not quite true. It has to be enabled and is disabled by default (see line 37 in `support/Chart.yaml` - doing links to lines on the mobile app is not easy!). So other hubs shouldn't be impacted if/when this is merged.

choldgraf:
Ah that is good to know! But I think that is actually another good example of something that we'd ideally have reference documentation on. E.g. how would somebody unfamiliar with this infrastructure quickly learn how NFS is provisioned, what the default behavior is, etc? Since it is disabled by default, how does somebody know when to enable it?

sgibson91:
All good questions that I'm trying to figure out myself :)

choldgraf:
@consideRatio you mentioned that you had been doing a dive into NFS stuff lately. Would you mind taking a look at @yuvipanda's questions and my questions above? Would you be willing to help advise on what kind of documentation would best-capture this? (or, whether there should be upstream documentation improvements for this?)

sgibson91:
Ok, I think I have some solutions to my issues in https://github.com/2i2c-org/pilot-hubs/pull/613#issuecomment-901956729

(I deleted all previously created namespaces `cert-manager`, `support` and `staging` and tried again)

There wasn't a default `storageClass` set, so I set `standard` as the default and then deployed the `support` chart. This time all the pods came up Running, but the install failed with the following:

```
Error: failed to create resource: Internal error occurred: failed calling webhook "validate.nginx.ingress.kubernetes.io": Post "https://support-ingress-nginx-controller-admission.support.svc:443/networking/v1beta1/ingresses?timeout=10s": no endpoints available for service "support-ingress-nginx-controller-admission"
```

Then I deployed the hub chart into `staging`, all pods came up but again received:

```
Error: Internal error occurred: failed calling webhook "validate.nginx.ingress.kubernetes.io": Post "https://support-ingress-nginx-controller-admission.support.svc:443/networking/v1beta1/ingresses?timeout=10s": context deadline exceeded
```

consideRatio:
>  Would you be willing to help advise on what kind of documentation would best-capture this? (or, whether there should be upstream documentation improvements for this?)

While there may be multiple points of documentation of relevance, I think one thing I would advocate for pilot hubs would be to let its Helm chart config follow a pattern similar to z2jh. In z2jh we have a schema.yaml file that defines configuration used both to generate documentation and to generate a `values.schema.json` file that `helm install|upgrade|template` use to validate configuration.

For NFS specifically, I think the practice of deploying the nfs-server-provisioner Helm chart could be a generally relevant discourse post and at some point be integrated into z2jh docs under a Tutorial or How to section perhaps.

support/values.yaml
```diff
@@ -71,3 +71,26 @@ grafana:
...
+  storageClass:
+    # Users will have to opt-in to using NFS-type storage under
+    # singleuser.storage.dynamic.storageClass
+    name: nfs
+    create: true
+    defaultClass: false
+    mountOptions:
+      - vers=4.2
+      - soft
+      - noatime
```
consideRatio, line 88: @yuvipanda the default options in the nfs-server-provisioner helm chart is `- vers=3`.

Would it make sense to create a PR suggesting a new default in the helm chart perhaps? I know too little to say that it does indeed make sense or not.

This is the upstream repo to update if so: https://github.com/kubernetes-sigs/nfs-ganesha-server-and-external-provisioner

yuvipanda: Yeah, I think the default should be 4.2 since nfs-ganesha supports it, unless they have a strong reason to not do that. Would be nice to open a PR upstream!

sgibson91:
So a lot of what I was experiencing in https://github.com/2i2c-org/pilot-hubs/pull/613#issuecomment-902597951 was to do with deploying charts that I wasn't ready to use yet. In [`ab9e77d`](https://github.com/2i2c-org/pilot-hubs/pull/613/commits/ab9e77d068433c9c1ebc55d2cb76af9b5c48dfad), I have added similar `condition: <chart name>.enabled` flag but set them to be _enabled by default_ - so behaviour doesn't change for any of the current hubs. I could then disable (mostly) everything I didn't want in order to deploy the `nfs-server-provisioner`, then deploy the hub, and then go back and turn on `ingress-nginx`, etc.

The exception to this is `cert-manager`. I tried adding a `--skip-cert-manager-deploy` flag to `deploer deploy-support`, but the Python logic didn't seem quite right. (It complained about number of arguments even though I followed what had been done for the `--skip-hub-health-check` flag to `deployer deploy`.) I think it's further complicated by the `ClusterIssuer` under `templates` too.

support/values.yaml
consideRatio: Opened https://github.com/kubernetes-sigs/nfs-ganesha-server-and-external-provisioner/issues/64

sgibson91:
So after some discussion in Slack, @yuvipanda and I decided to remove the following lines from `support/values.yaml` since they are never passed to `cert-manager` directly.

https://github.com/2i2c-org/pilot-hubs/blob/386d9a0b351878d693d7fd8a500222e726a90792/support/values.yaml#L3-L4

We are explicitly passing this argument in the deployer here:

https://github.com/2i2c-org/pilot-hubs/blob/386d9a0b351878d693d7fd8a500222e726a90792/deployer/hub.py#L97-L103

---

After making the above change, I deleted all helm deployments and k8s namespaces and tried another fresh install.

---

1. I disabled `prometheus`, `grafana` and `ingress-nginx` in the `support` chart and enabled `nfs-server-provisioner`
2. I then ran `python deployer deploy-support pangeo-hubs` which would deploy `nfs-server-provisioner` **and** `cert-manager`
3. I received the following error RE: the `cert-manager` webhook

```
Error: failed to create resource: Internal error occurred: failed calling webhook "webhook.cert-manager.io": Post "https://cert-manager-webhook.cert-manager.svc:443/mutate?timeout=10s": x509: certificate signed by unknown authority
```

Upon some searching, I found the following documentation around using the webhook on private GKE clusters - which this is! https://docs.cert-manager.io/en/release-0.9/getting-started/webhook.html#running-on-private-gke-clusters

sgibson91:
I'm actually not sure that the above issue is regarding firewall rules since we already have IP aliasing that allows outbound access to the internet? Inspecting the webhook pod, it is already using one of the [two ports that private GKE clusters automatically grant access to the control plane](https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters#:~:text=by%20default%2C%20firewall%20rules%20restrict%20your%20cluster%20control%20plane%20to%20only%20initiate%20tcp%20connections%20to%20your%20nodes%20and%20pods%20on%20ports%20443%20(https)%20and%2010250%20(kubelet).) anyway: `10250`.

yuvipanda:
@sgibson91 how do you feel about just turning off the webhook validation for now, tracking it in an issue, and continuing with getting everything else going?

sgibson91:
> @sgibson91 how do you feel about just turning off the webhook validation for now, tracking it in an issue, and continuing with getting everything else going?

That's exactly what I did! (minus opening the issue, but happy to)

I'm now frustrated that I still seem to get an error about failing to connect to the `ingress-nginx` controller - of course you can't connect, I didn't install `ingress-nginx`! I have `basehub.jupyterhub.proxy.https.enabled = false`, but I'm wondering if there's anything in `daskhub` that's trying to talk to `ingress-nginx` that needs disabling?

sgibson91:
What do you think about tracking the "clean install" story in a new issue before this PR gets too unwieldy? It might also help distinguish between "technically blocked because something doesn't work" and "something needs more discussion/process/documenting"

yuvipanda:
@sgibson91 +1 on opening a new clean issue. 



>  have `basehub.jupyterhub.proxy.https.enabled = false`, but I'm wondering if there's anything in `daskhub` that's trying to talk to `ingress-nginx` that needs disabling?

Anything from `kubectl get ingress -A`? Might show us if ingress objects are being added.

sgibson91:
> Anything from `kubectl get ingress -A`? Might show us if ingress objects are being added.

```
$ kubectl get ingress -A
No resources found
```

consideRatio:
Oh I had to allow 6443 before, but it seems now that they [have changed the port](https://github.com/jetstack/cert-manager/blob/38ab6f4bddabaac6ba46451f52c8720ed02e6831/deploy/charts/cert-manager/values.yaml#L300-L305).

```
  # The port that the webhook should listen on for requests.
  # In GKE private clusters, by default kubernetes apiservers are allowed to
  # talk to the cluster nodes only on 443 and 10250. so configuring
  # securePort: 10250, will work out of the box without needing to add firewall
  # rules or requiring NET_BIND_SERVICE capabilities to bind port numbers <1000
  securePort: 10250
```

@sgibson91 I believe that cert-manager may give you trouble if you try to disable the webhook in modern versions. What version are you installing? You pointed documentation about version 0.9 which is very old at this point (on Jul 23, 2019).

sgibson91:
Found the same info in v1.5: https://cert-manager.io/docs/concepts/webhook/#webhook-connection-problems-on-gke-private-cluster

Version being installed: https://github.com/2i2c-org/pilot-hubs/blob/386d9a0b351878d693d7fd8a500222e726a90792/deployer/hub.py#L84

I would reenable it after I'd deployed the hub and setup a DNS record. The whole sticking point of this PR is that other stuff in the `support` chart is trying to do it's thing before I'm ready for it because I don't have a hub up yet.

What I want to do is the following:

1. Deploy `nfs-server-provisioner`
2. Deploy a hub
3. Deploy other support components, i.e. `prometheus`, `grafana`, `ingress-nginx`, setup the DNS and then let `cert-manager` do it's thing.

But at the minute, helm seems to be complaining about not finding things that haven't been installed yet and I'm not sure what part of `basehub`/`daskhub`/whatever is making those calls.

sgibson91:
Deleting cluster, starting again from scratch. Only condition should be to run `deploy-support` and then `deploy`.

sgibson91:
So on a fresh cluster:

The **whole** of the support chart installed fine (no disabling of any sub-charts, just enabled `nfs-server-provisioner`)

Then tried to deploy the hub and got:

```
Error: Internal error occurred: failed calling webhook "validate.nginx.ingress.kubernetes.io": Post "https://support-ingress-nginx-controller-admission.support.svc:443/networking/v1beta1/ingresses?timeout=10s": context deadline exceeded
```

But all pods came up successfully

consideRatio:
Hmmmm? What is the `support-ingress-nginx-controller-admission` service in the support namespace? It seems like a k8s resource validation webhook is registered by the ingress-nginx controller as well, is that correct? I didn't know they did that.

sgibson91:
@consideRatio yes it's in the docs here https://kubernetes.github.io/ingress-nginx/how-it-works/#avoiding-outage-from-wrong-configuration

consideRatio:
@sgibson91 perhaps you can update the config so this template renders to a already allowed port?

```
          {{- if .Values.controller.admissionWebhooks.enabled }}
            - name: webhook
              containerPort: {{ .Values.controller.admissionWebhooks.port }}
              protocol: TCP
          {{- end }}
```

https://github.com/kubernetes/ingress-nginx/blob/605c243d7ae49e11202ea106bebc205b45b26333/charts/ingress-nginx/templates/controller-deployment.yaml#L169-L173

sgibson91:
I tried adding another firewall rule to terraform as well, but that didn't seem to help

```
resource "google_compute_firewall" "ingress_nginx_webhook_validation_ingress" {
  count = var.enable_private_cluster ? 1 : 0

  name    = "ingress-ngix-webhook-validation"
  project = var.project_id
  network = data.google_compute_network.default_network.name

  allow {
    protocol = "tcp"
    ports    = ["8443"]
  }

  // This range contains all IP addresses that IAP uses for TCP forwarding.
  // https://cloud.google.com/iap/docs/using-tcp-forwarding
  source_ranges = ["35.235.240.0/20"]
}
```

yuvipanda:
let's turn off the nginx-ingress validation webhook too and try? https://github.com/kubernetes/ingress-nginx/blob/605c243d7ae49e11202ea106bebc205b45b26333/charts/ingress-nginx/values.yaml#L525

sgibson91:
Disabling the admissions webhook seemed to work. Got a hub in CrashLoopBackOff though

```
  Warning  Unhealthy               61s (x12 over 83s)  kubelet                  Readiness probe failed: Get "http://10.8.0.32:8081/hub/health": dial tcp 10.8.0.32:8081: connect: connection refused
```

Probably need to allow traffic to/from port 8081 as well

sgibson91:
I disabled the webhook [over here](https://github.com/2i2c-org/pilot-hubs/pull/597/commits/a8bad3407242756cec88580ceefe9f03b4c48429) in #597

sgibson91:
\*collapses\* https://staging.pangeo.2i2c.cloud/

support/values.yaml
```diff
@@ -71,3 +69,26 @@ grafana:
...
+  persistence:
+    enabled: true
+    accessMode: ReadWriteOnce
+    size: 100Gi
+    # Future option is to reference an xfs storage class. This will allow us to enable quotas.
+    # https://github.com/pangeo-data/pangeo-cloud-federation/issues/654#issuecomment-861771398
```
yuvipanda, line 92: On most cloud providers, the storageClass is a reference to the underlying *physical* storage only - spinning disks or SSDs. I think the nfs external provisioner already uses XFS, and so the storageClass wouldn't come up. I think the reference in the provided link is for bare metal instances only.

yuvipanda:
I made some minor changes, this is good to go!

yuvipanda:
Great work, @consideRatio and @sgibson91!

choldgraf:
Amazing! thanks everybody for collaborating on this and getting it through!

I've updated https://github.com/2i2c-org/pilot-hubs/issues/629 to track documenting this configuration, so that we don't lose it. I think we should document it relatively soon while the information is still fresh.

PR #617, opened 2021-08-18
Update catalystcoop hub image

zaneselvans:
Updating our docker image to use our v0.4.0 software release.

Also omg we are finally working on the Sloan projects.

### DIFF
--- a/config/hubs/2i2c.cluster.yaml
+++ b/config/hubs/2i2c.cluster.yaml
@@ -327,7 +327,7 @@ hubs:
           singleuser:
             image:
               name: catalystcoop/pudl-jupyter
-              tag: 2021.03.27
+              tag: 2021.08.17
             memory:
               limit: 6G
               guarantee: 4G


choldgraf:
Any chance you wanna try doing this with the configurator? I'm pretty sure it is installed on your hub now...some docs here: https://pilot.2i2c.org/en/latest/admin/howto/configurator.html

(if it doesn't work we can merge this though)

zaneselvans:
Oh sure, I'm happy to. I did this because I searched the docs and [found that making a PR](https://pilot-hubs.2i2c.org/en/latest/howto/customize/custom-image.html#configuring-your-hub-to-use-the-custom-image) was still listed as the recommended way to update the config.

zaneselvans:
Well, the configurator is there, but I'm not sure if it worked.  There wasn't any confirmation or acknowledgement after I hit "Submit"

It might be good to provide an example or three as to what one is supposed to put in the "User docker image" field too.

choldgraf:
@zaneselvans thanks for the feedback - as you can probably tell, this is still super early-phase tooling :-) just thought I'd share in case you thought it was interesting or useful. I think you should just be able to start up a new user session and see if the environment is different.

zaneselvans:
Okay it seems like it updated successfully! So I'll close this PR.

Issue #625, opened 2021-08-23
Support OAuthenticator for our hubs


yuvipanda:
### Description

Right now, we only support using Auth0 for our hub authentication. Individual hubs should be able to opt-in to using OAuthenticator for specific features, rather than being forced to *just* use Auth0.

### Benefit

While our Auth0 setup allows us to create arbitrary new hubs without needing to provision external auth credentials (like GitHub OAuth secrets, Google OAuth secrets, etc), it doesn't allow us to use provider specific features present in [the JupyterHub OAuthenticator](https://github.com/jupyterhub/oauthenticator) - such as allowing users with a GitHub org or team to login (https://github.com/2i2c-org/pilot-hubs/issues/598).

Many communities may want to use these features, particularly the more advanced ones. Most importantly, this would allow us to mirror Pangeo's authentication setup.

### Implementation

- Note that if we set up OAuthentication for our hubs, this will be a manual step
- We should explore whether we can do the same thing but using OAuth instead of a manual OAuthentication

### Tasks to complete

- [x] Make a release of Z2JH (since changes to OAuthenticator are on master right now)
- [x] Bump our z2jh image in basehub chart and /images/hub: #635 
- [x] Implementation proposal for how we'd support both Auth0 + OAuthenticator
- [x] Support specifying secret config on a per-hub / per-cluster basis, to store auth credentials
- [ ] Determine policy for *when* we should use OAuthenticator directly vs Auth0

### Updates

- [x] **2021-09-14** - For this sprint, @sgibson91 is going to put together a design proposal for how we could balance using both Auth0 and the native authenticator in our hub deployments.
- [x] **2021-09-21** - Give a shot at implementing native OAuthenticator on our hubs + using that to authenticate with GitHub Teams. Will open another issue for long-term options using `Auth0`.
- **2021-09-29** - Merging #706 caused some JSON validation errors so we rolled back the merge and will try to debug
  - Deploy manually, might be being bitten by obscured error output in CI
  - Use online JSON validation checker to make absolutely sure

sgibson91:
Changelog for v1.1.3 of z2jh merged and tag release made https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2361

sgibson91:
Opened #635 to bump the version of z2jh we're using

choldgraf:
Next steps

- Come up with a design proposal for how we'd implement this in our hub deployments.

sgibson91:
[...]

PR #635, opened 2021-08-25
Bump jupyterhub chart and hub image to version 1.1.3

### LINKED ISSUE #539: Upgrade to z2jh 1.1.0+
# Summary

Task: to upgrade to z2jh 1.1.0 ([CHANGELOG](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/CHANGELOG.md))

- hub.services.<service>.apiToken is now automatically generated, and a bug in jupyterhub is fixed so it can be set to a blank even if it wouldn't be autogenerated by z2jh. I recall @GeorgianaElena ran into this bug before and perhaps there is some cleanup that can be done.
- I believe we install a specific commit of kubespwner that previously wasn't released in the pilot-hubs hub image, but that won't be needed any more
- KubeSpawner now intelligently sort environment variables to help them resolve. This can make our logic to set SCRATCH_BUCKET environment variables less complicated if i recall correctly.

sgibson91:
This brings in authenticating with GitHub Teams feature from
OAuthenticator

first step towards #625

EDIT: closes #539

### DIFF
--- a/hub-templates/basehub/Chart.yaml
+++ b/hub-templates/basehub/Chart.yaml
@@ -2,7 +2,9 @@ apiVersion: v2
 appVersion: '1.0'
 description: Deployment Chart for JupyterHub
 name: basehub
-version: "0.0.1-n1159.h5b045cd"
+# Updates to this version should be kept in sync
+# with the dependency reference in the daskhub chart.
+version: "0.1.0"
 dependencies:
   - name: jupyterhub
     # Updates to this version should follow go hand in hand with updates to
--- a/hub-templates/basehub/Chart.yaml
+++ b/hub-templates/basehub/Chart.yaml
@@ -6,5 +6,5 @@ version: "0.1.0"
 dependencies:
   - name: jupyterhub
     # REMEMBER TO CHANGE BASE IMAGE OF images/hub/ WHEN CHANGING THIS
-    version: 1.0.0-beta.1
+    version: 1.1.3
     repository: https://jupyterhub.github.io/helm-chart/
--- a/hub-templates/basehub/Chart.yaml
+++ b/hub-templates/basehub/Chart.yaml
@@ -5,6 +5,9 @@ name: basehub
 version: "0.0.1-n1159.h5b045cd"
 dependencies:
   - name: jupyterhub
-    # REMEMBER TO CHANGE BASE IMAGE OF images/hub/ WHEN CHANGING THIS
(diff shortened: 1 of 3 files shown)


yuvipanda (changes requested):
I think you need to run `chartpress --push` to build and push the new hub image with the new base.

sgibson91:
Happy to let someone else do that if it can't wait until I'm back? (About to go pick up my cat-sitter!) I don't think this is super urgent as we're not working on the OAuthenticator stuff yet :)

yuvipanda:
@sgibson91 absolutely! PLEASE GO VACATION! ROME NEEDS YOU!!!!!! SEND PICS :)

choldgraf:
**Update**: We plan to merge this in ~12 hours (@GeorgianaElena will shepherd and others should be ready to assist if something unexpected happens!)

hub-templates/basehub/Chart.yaml
```diff
@@ -2,9 +2,12 @@ apiVersion: v2
 appVersion: '1.0'
 description: Deployment Chart for JupyterHub
 name: basehub
-version: "0.1.0"
+version: "0.0.1-n1159.h5b045cd"
```
GeorgianaElena, line 5:
```suggestion
version: "0.1.0"
```

GeorgianaElena:
Before merging this, I tried to deploy it first manually only on the staging hub.
But, I bumped into this chart version mismatch when trying to do so. We declare a new version for the `basehub` chart, but the `daskhub` chart depends still on the old one.

My guess is that the `basehub` chart version was changed by mistake. @consideRatio, is this right or am I missing something?
Should we instead (of my code suggestion) have the `daskhub` chart depend on the new basehub chart version (`version: "0.0.1-n1159.h5b045cd"`)?

consideRatio:
Oh doh, great catch @GeorgianaElena! Either make daskhub reference basehubs new version, or make basehubs version be 0.1.0 again or whatever it was before.

Also update the inline comment about using chartpress about the need to keep basehubs chart version in sync with the dependency reference in daskhub.

GeorgianaElena:
Thanks @consideRatio! The staging deploy went ok and I was able to login and start a server on that hub afterwards.
So, I'll go ahead and push the merge button now :crossed_fingers:

Thanks everyone! :sun_with_face:

sgibson91:
Thank you for stewarding @GeorgianaElena and @consideRatio for assisting! 🎉

Issue #632, opened 2021-08-23
Automatically deploy carbonplan hub


yuvipanda:
### Description

Currently when you change things in pilot-hubs, it doesn't actually update CarbonPlan's hub. We should build auto-deployment infrastructure instead.

Was tried in https://github.com/yuvipanda/pilot-hubs/commit/f687264dadad5d126f3e281fa22c87d39dba8b17 but reverted, as the auth provided by the checked-in kubeconfig is not enough. We need https://github.com/2i2c-org/pilot-hubs/issues/381 with eks support to be able to deploy automatically from CI

### Tasks

- [x] Get @damianavila access to work on this hub
- [x] Create an AWS user with credentials setup with access to the EKS cluster
- [x] Add support to our deployer to fetch k8s credentials from eks (porting [this code](https://github.com/yuvipanda/hubploy/blob/dfc00254abb30d8887230472e83b405ff5872933/hubploy/auth.py#L298))
- [ ] Deploy changes whenever a change is merged here

yuvipanda:
@damianavila what's your carbonplan AWS username? I can give you access to the EKS cluster by following https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html

damianavila:
> @damianavila what's your carbonplan AWS username? 

I have pinged you in Slack with that info :wink: .

yuvipanda:
@damianavila I added you to list of users who can access the EKS cluster, with

`eksctl create iamidentitymapping -f carbonplan.yaml --group system:masters --username admin --arn arn:aws:iam::631969445205:user/damian`

damianavila:
> Add support to our deployer to fetch k8s credentials from eks (porting this code)

Update: Draft PR #636 tries to deal with this item.

damianavila:
Update:

* Created a `deployer` user in the AWS carbonplan space under the 2i2c group.
* Added the identity mapping for that new `deployer` user so it can access the EKS cluster
* Got access keys for that user and encrypted them

Tested manually (after some adjusting in the logic) and it seems to be working as expected.

damianavila:
Some commands I have used for the above update:

* Create a new user:
`aws iam create-user --user-name deployer`

* Add the user to the 2i2c group
`aws iam add-user-to-group --group-name 2i2c --user-name deployer`

* Check if the user was added to the group
`aws iam get-group --group-name 2i2c`

* Get the credentials:
[...]

PR #636, opened 2021-08-28
Auth AWS EKS cluster with encrypted AWS credentials

damianavila:
~~This is a draft PR for early feedback.~~

I have added a new auth_aws function following the existing pattern for
the general auth context manager. The main idea was to get a temporary
kubeconfig file with the update-kubeconfig AWS CLI command [1] so we can
successfully connect with the EKS cluster. The command looks for the IAM
entity in your default AWS CLI to properly authenticate. In this case,
we are going to provide some 2i2c-deployer AWS credentials in a sops
encrypted json file and set up the Access Key ID and Secret Access Key
as environment variables.

I have also added some new values to the schema with the aim to replace
the current auth_kubeconfig with a future (more general) auth_aws
function working for kops-based clusters as well.

~~Btw, I have not provided the encrypted credentials file because currently contains my AWS credentials and I would like to test it with credentials associated with a 2i2c-deployer user I still need to create and give access to the cluster. More stuff in the next few days.~~

This is actually one step forward in #632 and half of #381 (which is intended for kops-based clusters instead of EKS ones).

### DIFF
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -26,7 +26,7 @@ jobs:
           # add its name to the list
           - 2i2c
           - cloudbank
-          # - carbonplan  # We need to edit deployer to connect to aws before we can autodeploy carbonplan
+          - carbonplan
           - meom-ige
           - pangeo-181919
 
--- a/config/hubs/schema.yaml
+++ b/config/hubs/schema.yaml
@@ -79,25 +79,25 @@ properties:
     type: object
     additionalProperties: false
     description: |
-      Configuration to connect to the cluster on AWS. Is used when
-      provider is set to `aws`
+      Configuration to connect to the cluster on AWS.
+      Is used when provider is set to `aws`.
     properties:
       key:
         type: string
         description: |
-          Path to a `sops` encrypted JSON key that contains AWS credentials
-          with enough rights to get full access to the kubernetes cluster.
+          Path to a `sops` encrypted JSON file that contains the AWS credentials for the
+          "deployer" user (with enough rights to get full access to the cluster).
       clusterType:
         type: string
         description: |
-          We currently support kops and eks based kubernetes cluster on AWS.
+          We currently support `kops` and EKS-based cluster on AWS.
         enum:
           - kops
           - eks
       clusterName:
         type: string
         description: |
-          Name of the kubernetes cluster.
+          Name of the cluster.
       region:
         type: string
         description: |
--- a/config/hubs/schema.yaml
+++ b/config/hubs/schema.yaml
@@ -30,6 +30,7 @@ properties:
     enum:
       - gcp
       - kubeconfig
+      - aws
   kubeconfig:
     type: object
     description: |
@@ -74,6 +75,35 @@ properties:
       - project
       - cluster
       - zone
+  aws:
+    type: object
+    additionalProperties: false
+    description: |
+      Configuration to connect to the cluster on AWS. Is used when
+      provider is set to `aws`
+    properties:
+      key:
+        type: string
+        description: |
+          Path to a `sops` encrypted JSON key that contains AWS credentials
+          with enough rights to get full access to the kubernetes cluster.
+      clusterType:
+        type: string
+        description: |
+          We currently support kops and eks based kubernetes cluster on AWS.
+      clusterName:
+        type: string
+        description: |
+          Name of the kubernetes cluster.
+      region:
+        type: string
+        description: |
+          The AWS region the cluster is in.
+    required:
+      - key
+      - clusterType
+      - clusterName
+      - region
   hubs:
     type: array
     description: |
--- a/config/hubs/schema.yaml
+++ b/config/hubs/schema.yaml
@@ -91,6 +91,9 @@ properties:
         type: string
         description: |
           We currently support kops and eks based kubernetes cluster on AWS.
+        enum:
+          - kops
+          - eks
       clusterName:
         type: string
         description: |
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -164,8 +164,8 @@ class Cluster:
                     with open(decrypted_key_abspath) as f:
                         creds = json.load(f)
 
-                    os.environ['AWS_ACCESS_KEY_ID'] = creds['AccessKeyId']
-                    os.environ['AWS_SECRET_ACCESS_KEY'] = creds['SecretAccessKey']
+                    os.environ['AWS_ACCESS_KEY_ID'] = creds["AccessKey"]['AccessKeyId']
+                    os.environ['AWS_SECRET_ACCESS_KEY'] = creds["AccessKey"]['SecretAccessKey']
 
                 subprocess.check_call([
                     'aws', 'eks', 'update-kubeconfig',
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -145,7 +145,11 @@ class Cluster:
     def auth_aws(self):
         """
         Reads `aws` nested config and temporarily sets environment variables
-        like `KUBECONFIG`, `AWS_ACCESS_KEY_ID`, and `AWS_SECRET_ACCESS_KEY`.
+        like `KUBECONFIG`, `AWS_ACCESS_KEY_ID`, and `AWS_SECRET_ACCESS_KEY`
+        before trying to authenticate with the aws eks update-kubeconfig command.
+
+        Finally get those environment variables to the original values to prevent
+        side-effects on existing local configuration.
         """
         config = self.spec['aws']
         key_path = config['key']
@@ -158,7 +162,6 @@ class Cluster:
             orig_access_key_id = os.environ.get('AWS_ACCESS_KEY_ID', None)
             orig_secret_access_key = os.environ.get('AWS_SECRET_ACCESS_KEY', None)
             try:
-                os.environ['KUBECONFIG'] = kubeconfig.name
                 with decrypt_file(key_path) as decrypted_key_path:
 
                     decrypted_key_abspath = os.path.abspath(decrypted_key_path)
@@ -170,6 +173,8 @@ class Cluster:
 
                     os.environ['AWS_ACCESS_KEY_ID'] = creds["AccessKey"]['AccessKeyId']
                     os.environ['AWS_SECRET_ACCESS_KEY'] = creds["AccessKey"]['SecretAccessKey']
+
+                os.environ['KUBECONFIG'] = kubeconfig.name
 
                 subprocess.check_call([
                     'aws', 'eks', 'update-kubeconfig',
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -40,6 +40,8 @@ class Cluster:
     def auth(self):
         if self.spec['provider'] == 'gcp':
             yield from self.auth_gcp()
+        elif self.spec['provider'] == 'aws':
+            yield from self.auth_aws()
         elif self.spec['provider'] == 'kubeconfig':
             yield from self.auth_kubeconfig()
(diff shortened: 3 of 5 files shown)


config/hubs/schema.yaml
```diff
@@ -74,6 +75,35 @@ properties:
...
+    properties:
+      key:
+        type: string
+        description: |
+          Path to a `sops` encrypted JSON key that contains AWS credentials
+          with enough rights to get full access to the kubernetes cluster.
+      clusterType:
+        type: string
+        description: |
+          We currently support kops and eks based kubernetes cluster on AWS.
```
damianavila, line 93: This should be an enum...

deployer/hub.py
```diff
@@ -140,6 +142,46 @@ def auth_kubeconfig(self):
...
+
+                    os.environ['AWS_ACCESS_KEY_ID'] = creds['AccessKeyId']
+                    os.environ['AWS_SECRET_ACCESS_KEY'] = creds['SecretAccessKey']
+
+                subprocess.check_call([
+                    'aws', 'eks', 'update-kubeconfig',
+                    f'--name={cluster_name}',
+                    f'--region={region}'
+                ])
+
```
damianavila, line 175: We could add support for kops-based clusters here with subprocessing `kops export kubecfg --admin` if the `cluster_type` is kops. We would need an additional parameter for the state store.

choldgraf:
@damianavila could you either request a review from somebody specifically, or put this PR up on the sprint board? It's not linked to the original issue (#632 so it's not going to show up on the board UI unless people go digging through the issue)

damianavila:
> It's not linked to the original issue (#632 so it's not going to show up on the board UI unless people go digging through the issue)

I hope the linking was done automatically, but that does not seem to be the case unless you use some specific keyword.
I have now linked it manually.

damianavila:
Btw, I tested this one locally with my AWS carbonplan credentials and it seems to run successfully.

```
(pilot-hubs) Damians-MacBook-Pro:pilot-hubs damian$ python3 deployer deploy carbonplan staging --skip-hub-health-test
Added new context arn:aws:eks:us-west-2:631969445205:cluster/carbonplanhub to /var/folders/mn/8h0hm_395l31nrtn235h29900000gn/T/tmppa__63d4
Running helm upgrade --install --create-namespace --wait --namespace staging staging hub-templates/daskhub -f /var/folders/mn/8h0hm_395l31nrtn235h29900000gn/T/tmpa4311uug -f /var/folders/mn/8h0hm_395l31nrtn235h29900000gn/T/tmpnxy0id1w
NAME: staging
LAST DEPLOYED: Mon Aug 30 15:19:40 2021
NAMESPACE: staging
STATUS: deployed
REVISION: 9
TEST SUITE: None
```

damianavila:
OK, added the encrypted file after creating/enabling the "deployer" user (more details here: https://github.com/2i2c-org/pilot-hubs/issues/632#issuecomment-911649179). I think this one is now ready for review.

config/hubs/schema.yaml
```diff
@@ -74,6 +75,38 @@ properties:
...
+    additionalProperties: false
+    description: |
+      Configuration to connect to the cluster on AWS. Is used when
+      provider is set to `aws`
+    properties:
+      key:
+        type: string
+        description: |
+          Path to a `sops` encrypted JSON key that contains AWS credentials
+          with enough rights to get full access to the kubernetes cluster.
```
consideRatio, line 89: I'm not so used to AWS credentials and I have not yet worked with a .json file to contain them.

Are these to contain the credentials to an account with enough rights to get full access, or is it more indirect like an account that contains credentials to take on a AWS Role, and then the role has credentials etc?

Could you be a bit more specific about this?

config/hubs/schema.yaml
```diff
@@ -74,6 +75,38 @@ properties:
...
+      clusterType:
+        type: string
```
consideRatio, line 91: Is this acceptable in the schema of JSONschema file when we also have an `enum` as part of a JSON schema file restricting it to two specific string values already?

config/hubs/schema.yaml
```diff
@@ -74,6 +75,38 @@ properties:
...
+        description: |
+          We currently support kops and eks based kubernetes cluster on AWS.
+        enum:
+          - kops
+          - eks
+      clusterName:
+        type: string
+        description: |
+          Name of the kubernetes cluster.
```
consideRatio, line 100: I guess there is an EKS concept of a name, but is kops installing k8s on instances without AWS being aware of the k8s cluster? Then does it mean that this name have meaning to kops separate to the meaning it would have if used by eksctl?

deployer/hub.py
```diff
@@ -140,6 +142,46 @@ def auth_kubeconfig(self):
             os.environ['KUBECONFIG'] = decrypted_key_path
             yield
 
+    def auth_aws(self):
```
consideRatio, line 145:
```suggestion
    def auth_aws(self):
        """
        Reads `aws` nested config and temporarily sets environment variables
        like `KUBECONFIG`, `AWS_ACCESS_KEY_ID`, and `AWS_SECRET_ACCESS_KEY`.
        """
```

deployer/hub.py
```diff
@@ -140,6 +142,46 @@ def auth_kubeconfig(self):
...
+                    if not os.path.isfile(decrypted_key_abspath):
+                        raise FileNotFoundError(
+                            'The decrypted key file does not exist')
+                    with open(decrypted_key_abspath) as f:
+                        creds = json.load(f)
+
+                    os.environ['AWS_ACCESS_KEY_ID'] = creds["AccessKey"]['AccessKeyId']
+                    os.environ['AWS_SECRET_ACCESS_KEY'] = creds["AccessKey"]['SecretAccessKey']
+
+                subprocess.check_call([
```
consideRatio, line 170: Do we want the KUBECONFIG to be set as late as possible?

I believe the sops encryption operation can take a while. Will there be race conditions in the logic using this function? I'm a bit vague on the use of generators. That is a separate question in a way that also may relate to the question on where to set KUBECONFIG.

It vaguely feels better that its set just before calling aws eks update-kubeconfig to minimize the time its influencing things and such.

deployer/hub.py
consideRatio: What does it do? Does it consume AWS_ credentials to create a kubeconfig file?

I guess then that makes sense to me. Can we make it be written to the temp file allocated by python so we can benefit of having it automatically be deleted later then also?

deployer/hub.py
```diff
@@ -140,6 +142,46 @@ def auth_kubeconfig(self):
...
+                ])
+
+                yield
+            finally:
+                if orig_kubeconfig is not None:
+                    os.environ['KUBECONFIG'] = orig_kubeconfig
+                if orig_access_key_id is not None:
+                    os.environ['AWS_ACCESS_KEY_ID'] = orig_access_key_id
+                if orig_kubeconfig is not None:
+                    os.environ['AWS_SECRET_ACCESS_KEY'] = orig_secret_access_key
```
consideRatio, line 183: In AWS, sometimes you have an account with permissions to act as a certain role. We don't have support for that currently, so if you manage a cluster you need to make every user individually a k8s admin, rather than define what it means to be a k8s admin once and let everyone be able to act as such role.

The JMTE cluster is setup to have a k8s role and a ecr role (where we push an image).

consideRatio (commented):
This looks like a good first iteration to me and could be merged. I left non-blocking comments:
- On adding of a docstring
- On adding details to a schema description for `key`
- A question about a potential race condition stemming from me not understanding generators so well.

consideRatio (approved):
I'd be fine if this is merged as it is after my comment suggestion adding a docstring.

config/hubs/schema.yaml
damianavila: I will be more specific in the description, of course.
In this case, we have created a `deployer` user in the carbonplan AWS account with enough credentials to interact and work with the EKS cluster (the deployer user belongs to a group with full admin access).
In the future, we may want to revisit this approach and work with AWS roles so we have better granularity and easiness at the time to add users (such as the deployer one) to the EKS cluster (without the mapping we currently need to do). I tried this approach first because it is the most straightforward and because none of our current docs or practices in AWS land includes AWS roles... yet.

config/hubs/schema.yaml
damianavila: I followed the pattern I saw in the rest of the schema file and it seems to be valid and acceptable, AFAIK.

config/hubs/schema.yaml
damianavila: I would say they have a separate meaning as long as with kops, you are deploying the whole k8s thing whereas, with EKS, the control plane is under Amazon guard. I should probably be more descriptive here as well...

deployer/hub.py
damianavila: Thanks!

deployer/hub.py
damianavila: File operations here should be blocking, am I right? So, the subprocess call should proceed after we get the kubeconfig file properly written (and closed) and the AWS env variables properly set up.
Btw, I have not seen that race condition so far (it does not mean it could not happen if I am wrong about the above :wink:).

damianavila: > It vaguely feels better that its set just before calling aws eks update-kubeconfig to minimize the time its influencing things and such.

I was writing a reply to this one and get posted before it was ready :wink:
In any case, I now see where you are aiming at... I feel better as well if we set the KUBECONFIG env variable just before the `update-config` command, I will make the change.

deployer/hub.py
damianavila: > What does it do? Does it consume AWS_ credentials to create a kubeconfig file?

Yep, `update-kubeconfig` in EKS (and `kops export kubecfg` on kops world) will create a new kubeconfig file (or append and existing one) modulo you have provided the required credentials (and have access to the state store in kops world).

> Can we make it be written to the temp file allocated by python so we can benefit of having it automatically be deleted later then also?

We are creating a temporary kubeconfig file in line 152, are you asking for something else?

deployer/hub.py
damianavila: Yep, as I said in another reply:

> I tried this approach first because it is the most straightforward and because none of our current docs or practices in AWS land includes AWS roles... yet.

I would be happy to explore using roles in the near future.

config/hubs/schema.yaml
damianavila: Or get rid of the kubernetes word :wink:

damianavila:
> I'd be fine if this is merged as it is after my comment suggestion adding a docstring.

Thanks for your feedback @consideRatio.
I have pushed some commits hopefully addressing some of your comments.

Btw, since you approved it, I will be merging this one soon.

.github/workflows/deploy-hubs.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
           # add its name to the list
           - 2i2c
           - cloudbank
-          # - carbonplan  # We need to edit deployer to connect to aws before we can autodeploy carbonplan
+          - carbonplan
```
sgibson91, line 29: Do we also need to add an AWS login step into CD? Such as https://github.com/marketplace/actions/configure-aws-credentials-action-for-github-actions

sgibson91 (commented):
Looking good @damianavila! I have a small Q about Continuous Deployment :)

.github/workflows/deploy-hubs.yaml
damianavila: From my understanding, we would not need that step... this PR is retrieving the credentials from the encrypted file and setting up the needed AWS environment variables by itself, so any AWS CLI configuration should not be needed.

PR #649, opened 2021-08-31
Fix broken links and add linkcheck jobs

choldgraf:
This adds a linkcheck to our pilot hubs documentation. Decide to tackle this quickly before writing some extra docs to make sure that we are tracking broken links.

part of https://github.com/2i2c-org/pilot/issues/76


choldgraf:
assuming that the link check build passes, I will merge this one in, since it doesn't change content, only fixes links!

PR #662, opened 2021-09-07
requestor-pays-access

sgibson91:


docs/howto/configure/pangeo-requester-pays-access.md
```diff
@@ -0,0 +1,103 @@
+# Pangeo Data Access via Requester Pays
```
choldgraf, line 1: Perhaps we should make this generic instead of Pangeo-specific, since I could imagine other hubs would want the same setup?

choldgraf, line 1: I think we could use a top-level section like `Data access` and then this could be the first content in that section. I think that could bring this in-line with our Hub Administrator's guide "topic guides" here: https://pilot.2i2c.org/en/latest/

docs/howto/configure/pangeo-requester-pays-access.md
```diff
@@ -0,0 +1,103 @@
...
+
+For some hubs, such as our Pangeo deployments, the communities they serve require access to data stored in other projects.
+Accessing data normally comes with a charge that the folks _hosting_ the data have to take care of.
+However, there is a method by which those making the request are responsible for the charges instead: [Requester Pays](https://cloud.google.com/storage/docs/requester-pays).
```
choldgraf, line 5: Is "Requestor pays" access something unique to GCP? Or are we just prototyping this first on GCP?

docs/howto/configure/pangeo-requester-pays-access.md
```diff
@@ -0,0 +1,103 @@
...
+This document shows the steps required to set this method up.
+
+## Setting up Requester Pays Access on GCP
+
+```{note}
+We may automate these steps in the future.
```
choldgraf, line 11: This definitely feels like something that is semi-common for research groups, and that is not scalable in its current implementation of manual steps :-)

docs/howto/configure/pangeo-requester-pays-access.md
```diff
@@ -0,0 +1,103 @@
...
+  --member "serviceAccount:requester-pays-sa@PROJECT_ID.iam.gserviceaccount.com" \
+  PROJECT_ID
+
+gcloud projects add-iam-policy-binding \
+  --role roles/storage.objectViewer \
+  --member "serviceAccount:requester-pays-sa@PROJECT_ID.iam.gserviceaccount.com" \
+  PROJECT_ID
+```
+
+where `PROJECT_ID` is the ID of the Google Cloud project, **not** the display name!
```
choldgraf, line 47: maybe add moustache brackets so it's clear this is variable to be replaced? (e.g. `{{PROJECT_ID}}`)

docs/howto/configure/pangeo-requester-pays-access.md
```diff
@@ -0,0 +1,103 @@
...
+  Note: this is the **ID**, not the display name!
+
+5. RESTART THE HUB
+
+This is a very important step.
+If you don't do this you won't see the changes applied.
+
+You can restart the hub by heading to `https://<hub_url>/hub/admin` (you need to be logged in as admin), clicking the "Shutdown Hub" button, and waiting for it to come back up.
+
+You can now test the requester pys access by starting a server on the hub and running the below code in a script or Notebook.
```
choldgraf, line 96:
```suggestion
You can now test the requester pays access by starting a server on the hub and running the below code in a script or Notebook.
```

choldgraf (commented):
Just a few quick thoughts and suggestions from me. In general this looks quite nice, thanks for explaining all of this as you figured it out!

docs/howto/configure/pangeo-requester-pays-access.md
sgibson91: AFAIK, it is unique to GCP. I've not come across it before being introduced to the Pangeo deployment.

docs/howto/configure/pangeo-requester-pays-access.md
sgibson91: Absolutely. We should open an issue to figure out how to automate this as I suspect that it can't be confined to just the terraform scripts or just the deployer module, as for the last step you need a hub up and running. In practice though, this is a 5 min manual job.

sgibson91:
@choldgraf I think I've addressed your comment, let me know what you think!

docs/howto/configure/data-access.md
```diff
@@ -0,0 +1,113 @@
...
+Here we will document various ways to grant hubs access to external data.
+
+## Data Access via Requester Pays
+
+For some hubs, such as our Pangeo deployments, the communities they serve require access to data stored in other projects.
+Accessing data normally comes with a charge that the folks _hosting_ the data have to take care of.
+However, there is a method by which those making the request are responsible for the charges instead: [Requester Pays](https://cloud.google.com/storage/docs/requester-pays).
+This section demonstrates the steps required to setup this method.
+
+## Setting up Requester Pays Access on GCP
```
sgibson91, line 12:
```suggestion
### Setting up Requester Pays Access on GCP
```

docs/howto/configure/data-access.md
```diff
@@ -0,0 +1,113 @@
...
+Steps 1 and 2 only need to be run **when you create a new cluster**!
+
+From step 3 onwards only need to be repeated when **adding a new hub to an already existing cluster**!
+```
+
+1. Create a new Service Account
+
+```bash
+gcloud iam service-accounts create requester-pays-sa \
+  --description="Service Account to allow access to Pangeo data stored in the cloud" \
```
sgibson91, line 30:
```suggestion
  --description="Service Account to allow access to external data stored elsewhere in the cloud" \
```

yuvipanda:
Thanks for fixing this one, @sgibson91. I had already automated this with [config connector](https://cloud.google.com/config-connector/docs/overview), creating the appropriate bindings in https://github.com/2i2c-org/pilot-hubs/blob/4d84506587142a36205b56dd3592e45aca649e21/hub-templates/basehub/templates/cloud-resources/gcp/service-account.yaml#L25. So this should have maybe already worked - the question for me is why did it not? Was there a feature we needed to turn on? Are these config connector resources being created appropriately?

sgibson91:
@yuvipanda Some random thoughts:

- I wonder of the annotation (step 4 in the docs) wasn't happening?
- Role `storage.objectViewer` also seems to be missing from that link
- Object Viewer and Service Usage consumer should be granted _on the project_

Let's compare `requestor-pays-sa` and `pangeo-hubs-cluster-sa` [here](https://console.cloud.google.com/iam-admin/iam?authuser=2&project=pangeo-integration-te-3eea) to debug

sgibson91:
Something is definitely broken in that config though as I had to manually set this up for COESSING too https://github.com/2i2c-org/pilot-hubs/issues/473#issuecomment-883445851 which also had the config connector enabled https://github.com/2i2c-org/pilot-hubs/blob/4d84506587142a36205b56dd3592e45aca649e21/terraform/gcp/projects/pangeo-181919.tfvars#L10

yuvipanda:
@sgibson91 I think it needs to be explicitly enabled with helm config like https://github.com/2i2c-org/pilot-hubs/blob/4d84506587142a36205b56dd3592e45aca649e21/config/hubs/2i2c.cluster.yaml#L65.

sgibson91:
Hmmm, then we need to make that more discoverable/apparent that that's what it's doing. Even from that snippet, I don't think it's clear that that is also enabling requester pays access. Any recommendations?

yuvipanda:
@sgibson91 let's try and see if it works, and if it does, we'll have to find a way to document that. I haven't touched it since we set up the first 2i2c clusters. I didn't add any docs when I first implemented that - I'm really sorry about that :( 

Let's try turning it on and see if it works?

sgibson91:
Sure - no worries!

sgibson91:
@yuvipanda the config connector stuff in the helm chart doesn't work as it creates an annotation that looks like this:

```
Annotations: iam.gke.io/gcp-service-account: staging-user-sa@pangeo-integration-te-3eea.iam.gserviceaccount.com
```

But that service account doesn't exist anywhere in the Google project - I don't think it's created by terraform (we only create `*-cluster-sa` and `*-cd-sa`).

I'll open a new issue to track fixing this up (I think it's doable if we set a fixed display name, like `requester-pays-sa` as I've suggested in the docs) and I'll reenable the manual setup for now.

yuvipanda:
@sgibson91 that service account should've been also created by configconnector with https://github.com/2i2c-org/pilot-hubs/blob/f6db9a19b4058e1080312a75e9d5798cb3989a60/hub-templates/basehub/templates/cloud-resources/gcp/service-account.yaml#L3. We can't create it in terraform since it's one per hub rather than one per deployment.

sgibson91:
I'll double check but I checked the Google project after running the helm chart with that config enabled and noticed the name of the service account and it wasn't there.

sgibson91:
I opened https://github.com/2i2c-org/pilot-hubs/issues/669 to track this

sgibson91:
I've updated this PR so the manual steps reflect what we _expect_ the automation to be doing. Can we get this merged and then concentrate on figuring out why the automation doesn't work in #669?

damianavila (approved):
Let's get this one merged and figure it out why the automation is not working as part of another ticket, as @sgibson91 proposed.

choldgraf:
thanks @sgibson91 for digging into this + the follow up issue!

PR #665, opened 2021-09-08
Use n1- instances instead of e2 instances

yuvipanda:
n1 instances get a sustained use discount while e2 instances
do not. I had previously believed that n1 instances did *not*
get this discount if they were not running all the time. However,
this is not true, based on a re-reading of
https://cloud.google.com/compute/docs/sustained-use-discounts.

Switching to n1 instances will give us pretty good cost savings.

# Tasks

Deploy this change to:

- [x] CloudBank
- [x] Pilot Hubs
- [ ] MEOM-IGE

closes https://github.com/2i2c-org/pilot-hubs/issues/715

### DIFF
--- a/config/hubs/meom-ige.cluster.yaml
+++ b/config/hubs/meom-ige.cluster.yaml
@@ -56,35 +56,35 @@ hubs:
                   mem_limit: 8G
                   mem_guarantee: 5.5G
                   node_selector:
-                    node.kubernetes.io/instance-type: e2-standard-2
+                    node.kubernetes.io/instance-type: n1-standard-2
               - display_name: "Medium"
                 description: "~8 CPU, ~32G RAM"
                 kubespawner_override:
                   mem_limit: 32G
                   mem_guarantee: 25G
                   node_selector:
-                    node.kubernetes.io/instance-type: e2-standard-8
+                    node.kubernetes.io/instance-type: n1-standard-8
               - display_name: "Large"
                 description: "~16 CPU, ~64G RAM"
                 kubespawner_override:
                   mem_limit: 64G
                   mem_guarantee: 55G
                   node_selector:
-                    node.kubernetes.io/instance-type: e2-standard-16
+                    node.kubernetes.io/instance-type: n1-standard-16
               - display_name: "Very Large"
                 description: "~32 CPU, ~128G RAM"
                 kubespawner_override:
                   mem_limit: 128G
                   mem_guarantee: 115G
                   node_selector:
-                    node.kubernetes.io/instance-type: e2-standard-32
+                    node.kubernetes.io/instance-type: n1-standard-32
               - display_name: "Huge"
                 description: "~64 CPU, ~256G RAM"
                 kubespawner_override:
                   mem_limit: 256G
                   mem_guarantee: 230G
                   node_selector:
-                    node.kubernetes.io/instance-type: n2-standard-64
+                    node.kubernetes.io/instance-type: n1-standard-64
             defaultUrl: /lab
             image:
               name: pangeo/pangeo-notebook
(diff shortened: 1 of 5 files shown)


damianavila (approved):
If it saves us money 😉 
LGTM.

yuvipanda:
I've applied this for cloudbank and pilot-hubs workspaces. meom-ige is currently actively being used, so haven't applied there yet. will try to do later.

choldgraf:
What is the state of this one? Are we just waiting to apply it to MEOM-IGE before merging and closing this?

yuvipanda:
Yep! I'll try do meom-ige today. They're always actively using it tho :D

choldgraf:
Hey all - I updated the task list at the top with the clusters I think we need to deploy to. I've also created an issue for this, and am adding it to our sprint board so that we don't lose track of it. (in the future, we should create an issue first, lay out the problem there, and make sure that it's on either the backlog or the sprint board before we start work on it, so that we don't lose track of it)

yuvipanda:
When applying this for the grenoble cluster, terraform wanted to replace the entire cluster! Because it thought somehow the default nodepool (which we don't use?) is changing from preemptible to non-preemptible.

```
      ~ node_config {
          ~ disk_size_gb      = 100 -> (known after apply)
          ~ disk_type         = "pd-ssd" -> (known after apply)
          ~ guest_accelerator = [] -> (known after apply)
          ~ image_type        = "COS_CONTAINERD" -> (known after apply)
          ~ labels            = {
              - "k8s.dask.org/node-purpose" = "worker"
            } -> (known after apply)
          ~ local_ssd_count   = 0 -> (known after apply)
          ~ machine_type      = "e2-standard-2" -> (known after apply)
          ~ metadata          = {
              - "disable-legacy-endpoints" = "true"
            } -> (known after apply)
          ~ oauth_scopes      = [
              - "https://www.googleapis.com/auth/cloud-platform",
            ] -> (known after apply)
          ~ preemptible       = true -> false # forces replacement
          - tags              = [] -> null
          ~ taint             = [
              - {
                  - effect = "NO_SCHEDULE"
                  - key    = "k8s.dask.org_dedicated"
                  - value  = "worker"
                },
            ] -> (known after apply)
            # (1 unchanged attribute hidden)

          ~ shielded_instance_config {
              ~ enable_integrity_monitoring = true -> (known after apply)
              ~ enable_secure_boot          = false -> (known after apply)
            }

          ~ workload_metadata_config {
              ~ node_metadata = "SECURE" -> (known after apply)
            }
        }
```

I have now asked terraform to ignore any changes it thinks have been made to the `node_config` block. We don't use it, terraform [tells us to not use it](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#node_config), and any changes it thinks are happening there are either accidental or unintentional. Changes there are also no-op, since we use `  remove_default_node_pool = true` to get rid of it anyway.

yuvipanda:
And n1- instances have slightly less memory than e2- instances, so I had to adjust memory guarantees as well. Otherwise new nodes were not being spawned.

yuvipanda:
I was able to easily debug the reason for nodes not scaling up by:

1. Viewing the [GKE cluster autoscaler logs](https://cloud.google.com/compute/docs/autoscaler/viewing-autoscaler-logs)
2. Searching for the name of the pod that was not triggering a scale up
3. In the big JSON payload that shows up in the result, look for the name of the nodepool you think should've scaled up. There should be a  `reason` field that gives you information. In this case, it was:

```
              "reason": {
                "parameters": [
                  "NodeResourcesFit",
                  "Insufficient memory"
                ],
                "messageId": "no.scale.up.mig.failing.predicate"
```

This told me that I needed to adjust the memory request on the pod.

yuvipanda:
This is done now.

damianavila:
> I was able to easily debug the reason for nodes not scaling up by:

We should have a section in our docs for this sort of troubleshooting...

PR #667, opened 2021-09-09
Prototype single list of admin users

choldgraf:
In creating the AUP hub, I ran into the issue of needing to choose admin users from the other lists on the page. I wanted to see if we could define a quick way to streamline this process so that know that all 2i2c Engineers have access to all hubs.

I am curious what people think about the current PR. It uses YAML references to define admin users once and insert them throughout the hubs.

There is one problem with this: when you insert YAML lists like this, they aren't flattened, so something like:

```yaml
keyone: &mykey
  - a
  - b
keytwo:
- *mykey
- c
```

effectively becomes

```yaml
```yaml
keyone: &mykey
  - a
  - b
keytwo:
- - a
  - b
- c
```

To get around this, I added a little utility function that just searches through the `Authenticator` config and flattens its contents. See diff for details!

# My question

Is this a feasible stop-gap solution to only require maintaining a single list of users? Is there any way for us to process that list with a `flatten` function before passing it to JupyterHub?

# Tasks

- [ ] Decide if this is a reasonable approach
- [ ] If so, add docs to describe how this affects new hub creation
- [ ] Update our pre-existing hubs to follow this pattern

### DIFF
--- a/config/hubs/2i2c.cluster.yaml
+++ b/config/hubs/2i2c.cluster.yaml
@@ -16,6 +16,23 @@ support:
           - secretName: grafana-tls
             hosts:
               - grafana.pilot.2i2c.cloud
+staff:
+  # Alphabetical for github handles, and OAuth usernames follow same order as github
+  names:
+    github: &staff-github
+      - choldgraf
+      - consideRatio
+      - damianavila
+      - GeorgianaElena
+      - sgibson91
+      - yuvipanda
+    google_oauth: &staff-google-oauth
+    - choldgraf@2i2c.org
+    - erik@2i2c.org
+    - damianavila@2i2c.org
+    - georgiana.dolocan@gmail.com
+    - sgibson@2i2c.org
+    - yuvipanda@2i2c.org
 hubs:
   - name: staging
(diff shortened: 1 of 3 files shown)


config/hubs/2i2c.cluster.yaml
```diff
@@ -48,10 +65,8 @@ hubs:
           config:
             Authenticator:
               allowed_users: &staging_users
-                - yuvipanda@gmail.com
-                - colliand@gmail.com
-                - choldgraf@gmail.com
-                - georgiana.dolocan@gmail.com
+                - *staff-google-oauth
+                - anextrausername
```
choldgraf, line 69: this is just to show off what I'm talking about here - obviously we don't want to actually add `anextrausername` to our hub :-)

damianavila:
It is actually pretty frustrating that YAML does not support this somehow...
It seems to be a pretty common use case accordingly to a quick search.

choldgraf:
I'm gonna close this one so that @GeorgianaElena can focus on figuring this out in https://github.com/2i2c-org/pilot-hubs/pull/311

PR #668, opened 2021-09-09
Add AUP hub

choldgraf:
This adds a hub for the American University of Paris. It starts off with a small environment (2G RAM per user) but hopefully this is enough for most light educational cases. @shaolintl and @swalker can bump up the memory via the configurator if need be (https://pilot.2i2c.org/en/latest/admin/howto/configurator.html).

I'd love if somebody from the @2i2c-org/tech-team could take a look and approve to make sure I got the config right.

Note also that I opened up #667 as a quick idea while I was implementing this. Would love thoughts there :-)

closes https://github.com/2i2c-org/pilot-hubs/issues/659

### DIFF
--- a/config/hubs/2i2c.cluster.yaml
+++ b/config/hubs/2i2c.cluster.yaml
@@ -436,3 +436,43 @@ hubs:
                 - fzhu2e
                 - alexkjames
               admin_users: *paleohack_users
+  - name: aup
+    domain: aup.pilot.2i2c.cloud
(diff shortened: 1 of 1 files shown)


sgibson91 (approved):
LGTM!

damianavila (approved):
LGTM as well.

choldgraf:
ok let's see how it goes :-)

PR #672, opened 2021-09-10
Update AUP logo

choldgraf:
Updates the logo of AUP so it shows up better


choldgraf:
merging this since it's just a logo URL change

Issue #359, opened 2021-04-26
Tighten permissions granted to secrets.GITHUB_TOKEN used in GitHub Workflows across the 2i2c-org


consideRatio:
### Description

The 2i2c-org GitHub org provides read & write permissions to the automatically provided `GITHUB_TOKEN` that can be used to read/write the content of a repo.

As discussed in https://github.com/jupyterhub/team-compass/issues/404, I think we should also tighten the permissions given to this to be read only by default across the org. We should then explicitly set each of our GitHub Actions to have this permission where we want it so.

### Benefit

Since it can be used in GitHub workflows to write content to branches, update comments, make PRs etc, the suggestion is to provide specific permissions to the token on a GitHub Job level (workflows has many jobs, jobs has many steps).

### Tasks to complete

- [x] Agree to reduce default permissions to read only. Decision prompt here: https://github.com/2i2c-org/pilot-hubs/issues/359#issuecomment-919500130
- [x] Explicitly add the required permissions where GITHUB_TOKEN is referenced in the 2i2c-org's GitHub workflows across various repos. See https://github.com/jupyterhub/team-compass/issues/404 for instructions on how to do this.
- [x] Make the 2i2c-org's default permissions read only at: https://github.com/organizations/2i2c-org/settings/actions

consideRatio:
Meta questions: 
- is this too big to be a task, or is it more of a deliverable?
- would it made more sense for this to be posted in the meta git repository?

choldgraf:
Let's add this to the deliverables board since it's important and requires a bit of thinking. I imagine that a clear "do X then Y" task will emerge from that discussion, which we can use to close out the deliverable. Does that makes sense?

choldgraf:
@consideRatio to your other meta question, perhaps this issue can tackle both the "what do we do in this repository" and the "what is our team policy moving forward" question at the same time. So another task associated with this deliverable would be a change to the team compass to define the policy

consideRatio:
> Let's add this to the deliverables board [...]

Yes!

> [...], perhaps this issue can tackle both the "what do we do in this repository" and the "what is our team policy moving forward" question at the same time.

Do you mean that we could define a set of "what we do in repository X" policies in the team-compass, and track that as tasks part of this deliverable?

choldgraf:
@consideRatio I didn't imagine it needed to be repository-specific, just general team guidelines for this. But ya, in the team compass somewhere

choldgraf:
[...]

PR #673, opened 2021-09-10
Use an AWS deployer user to fetch kubernetes credentials in our deployer

### LINKED ISSUE #381: Use an AWS IAM Role / User to fetch kubernetes credentials in our deployer

damianavila:
Summary:

The auth_aws function I recently added to the deployer is now authenticating kops-based clusters.
Farallon would be the first one.
You can see more details in the commit messages.
You may find interesting the usage of if and else keyword in the schema to properly validate the config file.

Btw, this one still needs to be more tested but I believe is complete so putting it out there to get early feedback.

Closes #381

### DIFF
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -28,6 +28,7 @@ jobs:
           - cloudbank
           - carbonplan
           - farallon
+          # Uncomment openscapes once a deployer user is created in openscapes AWS land
           # - openscapes
           - meom-ige
           - pangeo-181919
@@ -88,9 +89,11 @@ jobs:
           (steps.base_files.outputs.files == 'true') ||
           (steps.config_files.outputs.hub_config == 'true')
         run: |
-          curl -Lo kops https://github.com/kubernetes/kops/releases/download/$(curl -s https://api.github.com/repos/kubernetes/kops/releases/latest | grep tag_name | cut -d '"' -f 4)/kops-linux-amd64
+          curl -Lo kops https://github.com/kubernetes/kops/releases/download/$KOPS_VERSION/kops-linux-amd64
           chmod +x kops
           sudo mv kops /usr/local/bin/kops
+        env:
+          KOPS_VERSION: "v1.21.1"
 
       - name: Deploy ${{ matrix.cluster_name }}
         if: |
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -87,7 +87,10 @@ jobs:
         if: |
           (steps.base_files.outputs.files == 'true') ||
           (steps.config_files.outputs.hub_config == 'true')
-        uses: hiberbee/github-action-kops@1.0.0
+        run: |
+          curl -Lo kops https://github.com/kubernetes/kops/releases/download/$(curl -s https://api.github.com/repos/kubernetes/kops/releases/latest | grep tag_name | cut -d '"' -f 4)/kops-linux-amd64
+          chmod +x kops
+          sudo mv kops /usr/local/bin/kops
 
       - name: Deploy ${{ matrix.cluster_name }}
         if: |
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -27,6 +27,8 @@ jobs:
           - 2i2c
           - cloudbank
           - carbonplan
+          - farallon
+          # - openscapes
           - meom-ige
           - pangeo-181919
           - pangeo-hubs
@@ -80,6 +82,12 @@ jobs:
           (steps.base_files.outputs.files == 'true') ||
           (steps.config_files.outputs.hub_config == 'true')
         uses: mdgreenwald/mozilla-sops-action@v1
+
+      - name: Setup kops
+        if: |
+          (steps.base_files.outputs.files == 'true') ||
+          (steps.config_files.outputs.hub_config == 'true')
+        uses: hiberbee/github-action-kops@1.0.0
 
       - name: Deploy ${{ matrix.cluster_name }}
         if: |
--- a/config/hubs/schema.yaml
+++ b/config/hubs/schema.yaml
@@ -102,11 +102,23 @@ properties:
         type: string
         description: |
           The AWS region the cluster is in.
+      stateStore:
+        type: string
+        description: |
+          A dedicated S3 bucket for kops to use in order to store the state
+          (and the representation) of your of your kops cluster.
     required:
       - key
       - clusterType
       - clusterName
       - region
+    if:
+      properties:
+        clusterType:
+          const: kops
+    then:
+      required:
+        - stateStore
   hubs:
     type: array
     description: |
--- a/config/hubs/farallon.cluster.yaml
+++ b/config/hubs/farallon.cluster.yaml
@@ -1,7 +1,11 @@
 name: farallon
-provider: kubeconfig
-kubeconfig:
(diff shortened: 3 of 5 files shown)


.github/workflows/deploy-hubs.yaml
```diff
@@ -27,6 +27,8 @@ jobs:
           - 2i2c
           - cloudbank
           - carbonplan
+          - farallon
+          # - openscapes
```
GeorgianaElena, line 31: Maybe add a comment here about when this should be un-commented (if the plan is not to do it as part of this PR)?

.github/workflows/deploy-hubs.yaml
```diff
@@ -81,6 +83,12 @@ jobs:
           (steps.config_files.outputs.hub_config == 'true')
         uses: mdgreenwald/mozilla-sops-action@v1
 
+      - name: Setup kops
+        if: |
+          (steps.base_files.outputs.files == 'true') ||
+          (steps.config_files.outputs.hub_config == 'true')
+        uses: hiberbee/github-action-kops@1.0.0
```
GeorgianaElena, line 90: This doesn't look like it's maintained anymore and its CI seems to be failing :(
But I agree that it's worth trying it and see if it works. In case it's working, maybe also leave a comment about the possibly of it failing in the future?

config/hubs/schema.yaml
```diff
@@ -102,11 +102,23 @@ properties:
...
       - clusterType
       - clusterName
       - region
+    if:
+      properties:
+        clusterType:
+          const: kops
+    then:
+      required:
+        - stateStore
```
GeorgianaElena, line 121: Really interesting pattern :star:

GeorgianaElena (commented):
This looks generally awesome! Thanks @damianavila

.github/workflows/deploy-hubs.yaml
damianavila: I think the plan is actually to enable it on this PR as well :wink:

.github/workflows/deploy-hubs.yaml
damianavila: The master version seems to be failing whereas 1.0.0 seems to be green... still, as you said, this is just a try before diving into our own kops install (if we need to do it).

damianavila:
Thanks for the review @GeorgianaElena!

Btw, local testing on top of staging seems to be working as expected:

```
$ python3 deployer deploy farallon staging --skip-hub-health-test
kops has set your kubectl context to farallonhub.k8s.local
Running helm upgrade --install --create-namespace --wait --namespace staging staging hub-templates/daskhub -f /var/folders/mn/8h0hm_395l31nrtn235h29900000gn/T/tmpt92t1imr -f /var/folders/mn/8h0hm_395l31nrtn235h29900000gn/T/tmp903tmgl3
NAME: staging
LAST DEPLOYED: Mon Sep 13 16:16:16 2021
NAMESPACE: staging
STATUS: deployed
REVISION: 13
TEST SUITE: None
(pilot-hubs) Damians-MacBook-Pro:pilot-hubs damian$ kubectl -n staging get pods
NAME                                               READY   STATUS    RESTARTS   AGE
api-staging-dask-gateway-f47fb549-p9vf4            1/1     Running   0          82d
autohttps-6b8c745849-vcq7r                         2/2     Running   0          82s
controller-staging-dask-gateway-54fdd4bfdc-drjrw   1/1     Running   0          99d
hub-8567fc7d8f-dkcn5                               2/2     Running   0          50s
proxy-56cdf9b78c-kk86b                             1/1     Running   0          72s
traefik-staging-dask-gateway-7df6cffd45-8jdl2      1/1     Running   2          99d
```

Also, run successfully with the test enabled 😉 

```
Running hub health check...
.                                                                                                                                                                                                                                      [100%]
1 passed in 332.02s (0:05:32)
```

sgibson91 (approved):
LGTM! Thank you for this work @damianavila!

damianavila:
@GeorgianaElena and @sgibson91, I pushed another commit that I think you would be interested in.
I am not a GH actions user, so let me know if you have any further feedback about it.
Thanks!

sgibson91:
@damianavila Looks good to me! That's a totally valid workflow step :)

sgibson91:
The only thing we _might_ want to do is, instead of always pulling the latest release of `kops` is use an environment variable to pin the version we want to use for now, and then we're in control of bumping to the latest version. See what we do with the `HELM_VERSION` variable over in mybinder.org-deploy https://github.com/jupyterhub/mybinder.org-deploy/blob/89cfc75dff4f711bb9427da7cb9af0a24a28dfa9/.github/workflows/cd.yml#L29

damianavila:
@sgibson91, I think the last commit addressed what you ask for, let me know if that is not the case.

damianavila:
OK, I will proceed with the merge since I have at least one approval and nobody is working in the prod hub now.
Let's see how this one goes...

damianavila:
And it obviously failed 😛 
Looking at it soon...

damianavila:
@sgibson91, I am surprised this is failing because, accordingly to the [docs](https://docs.github.com/en/actions/reference/environment-variables#about-environment-variables) that should work

> If you use the workflow file's run key to read environment variables from within the runner operating system (as shown in the example above), the variable is substituted in the runner operating system after the job is sent to the runner.

unless I am missing something, which is obviously the case :wink:
Thoughts? Maybe this is related to the fact we are using a matrix somehow?

damianavila:
Well... testing the same on my test repo works as expected... mmm...

https://github.com/damianavila/test/blob/main/.github/workflows/test_kops_gha.yml

damianavila:
So, it seems there is some permission error preventing curl to write the file on this repo... wondering why the behavior is different from my repo.

choldgraf:
Hmmm could it be that we have run into https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/ ?

I believe @consideRatio had set the default permissions of all repos to be read only

sgibson91:
@choldgraf i would agree except GITHUB_TOKEN isn't being invoked 🤔

sgibson91:
This could be a closed pipe issue and I suspect that if we reran CI, it would be a flaky test and pass sometimes and not others

https://github.com/helm/helm/issues/2802

consideRatio:
@choldgraf no change made by me regarding that, but I suggest that permissions for GITHUB_TOKEN are explicitly specified - especially in any templates passed to others, so our org and other orgs can narrow down permissions as a security measure and still have the workflows functional.

This is #359

damianavila:
I think this is curl-specific, I will switch to wget and try again :wink:

damianavila:
For posterity, this was not a curl not a wget issue.
The problem was we already have a kops directory in the repo and I was trying to overwrite it at the time to install kops 🤦 .

PR #682, opened 2021-09-14
Tweak CI so kops setup step only runs for AWS hubs

sgibson91:
Merging #673 introduced a bug to do with curl when installing kops manually. This is breaking CI for all hubs, but should only affect hubs deployed on AWS not GCP. This PR adds in some provider information into the matrix job and updates the condition on the kops install step to only run when the provider is AWS.

### DIFF
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -21,18 +21,26 @@ jobs:
       # Don't stop other deployments if one fails
       fail-fast: false
       matrix:
-        cluster_name:
+        include:
           # To enable auto-deployments for other clusters,
           # add its name to the list
-          - 2i2c
-          - cloudbank
-          - carbonplan
-          - farallon
+          - cluster_name: 2i2c
+            provider: gcp
+          - cluster_name: cloudbank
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-hubs.yaml
```diff
@@ -21,18 +21,26 @@ jobs:
       # Don't stop other deployments if one fails
       fail-fast: false
       matrix:
-        cluster_name:
+        include:
           # To enable auto-deployments for other clusters,
           # add its name to the list
```
sgibson91, line 26:
```suggestion
          # add its name and provider to the list
```

yuvipanda (approved):

damianavila:
This is lovely @sgibson91, thanks!
Btw, we should "technically" install kops with kops-based clusters such as Farallon (and eventually Openscapes).
Carbonplan one, even when it is a AWS cluster, it is supported by EKS, so we do not need kops there.
Still, I think this PR is "conceptually" easier for others to understand, so I would be +1 on keeping it as is even when we are installing kops in the Carbonplan job when that is not actually needed :wink:.

PR #686, opened 2021-09-14
Replace curl with wget to download kops

damianavila:
Curl is weirdly failing at the time to download kops, let's use wget instead to see if that solves the problem.


damianavila:
This is one-liner to replace curl with wget, I will self-merge without review to, hopefully, unblock the failing aws builds on master.

PR #687, opened 2021-09-14
Do not try to override the existing kops directory

damianavila:
We already have a kops directory in the repo, I think that is the cause of the failure, let's try downloading it into /tmp.


damianavila:
Another attempt after seeing #686 failing... going to self-merge now (same reasons as before).

PR #690, opened 2021-09-14
Revert to obscuring output from pytest

sgibson91:
Pytest is continuing to leak tokens in URLs when tests fail. This PR reverts to the state where we redirect pytest's output to /dev/null. We should iterate on this and move away from pytest completely.

P.S. there's a weird double commit here because I accidentally committed to master on my fork, then did a cherry-pick but I had untracked files so git complained and made me do `git commit --allow-empty`. Why does everything break when you're trying to be efficient?

### DIFF
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -473,15 +473,20 @@ class Hub:
 
                 hub_url = f'https://{self.spec["domain"]}'
 
+                # On failure, pytest prints out params to the test that failed.
+                # This can contain sensitive info - so we hide stderr
+                # FIXME: Don't use pytest - just call a function instead
                 print("Running hub health check...")
-                exit_code = pytest.main([
-                    "-q",
-                    "deployer/tests",
-                    "--hub-url", hub_url,
-                    "--api-token", service_api_token,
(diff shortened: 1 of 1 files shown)


damianavila:
> Why does everything break when you're trying to be efficient?

That's a great question! :wink:

damianavila (approved):
LGTM

damianavila:
I guess after this one lands, we can try to put Farallon in auto-deploy mode again (revert #689), correct?

sgibson91:
Yes. It may be best to rotate the keys again before we do that too, though I'm not sure how @yuvipanda did that. It would be nice to have some SRE-style for handling these kinds of scenarios

damianavila:
> Yes. It may be best to rotate the keys again before we do that too, though I'm not sure how @yuvipanda did that. It would be nice to have some SRE-style for handling these kinds of scenarios

@yuvipanda, can you comment on that?

damianavila:
>  It would be nice to have some SRE-style for handling these kinds of scenarios

I agree, @sgibson91.

PR #693, opened 2021-09-20
Update incident report template

choldgraf:
This updates our incident report template so that it is a bit more actionable _during_ the incident. It splits up the report into two sections: one for during the incident that should be updated as we learn more, and one for after the incident that is more about reflection.


damianavila (approved):
I like the idea, LGTM.

choldgraf:
I'll leave it open a day or two to see if other team members have objections

choldgraf:
merging this one in as I think it's a nice improvement and not too big :-)

PR #704, opened 2021-09-22
Adding tips section and updating viz

choldgraf:
This does a couple of quick things:

- Minor visual update so that it is more in-line with our org-wide docs (at docs.2i2c.org)
- Adds a `troubleshooting` topic and a section that has the commands that @sgibson91 gave @GeorgianaElena in a recent debugging session
- Minor title rename so that it is less-long

Would love feedback from @sgibson91 and @GeorgianaElena that I have the commands here correct (and let me know if I have missed an important troubleshooting command or something)


choldgraf:
@sgibson91 you mentioned that there was a command needed to rotate a secret...should we add that here as well? I am happy to do so as part of this PR if you point me in the right direction

sgibson91:
> @sgibson91 you mentioned that there was a command needed to rotate a secret...should we add that here as well? I am happy to do so as part of this PR if you point me in the right direction

@damianavila and I have been trying to get @yuvipanda to share what he did as he was the one who rotated the secret.

docs/topic/troubleshooting.md
```diff
@@ -0,0 +1,28 @@
...
+
+:::{note}
+Ideally, this would have happened automatically via CI/CD, but sometimes a manual deploy is still necessary!
+:::
+
+If you'd simply like to revert back to the state of the Kubernetes infrastructure from before you ran `helm install`, try the following commands:
+
+- **Get the deployment name and revision number for the latest deploy**. To do so, run this command:
+  ```bash
+  helm list -n {{NAMESPACE}}
```
sgibson91, line 18:
```suggestion
  helm list --namespace {{NAMESPACE}}
```

My preference is to use the long flags in documentation so it's clear what is being run and it's easier to find out more about that flag through a web search

docs/topic/troubleshooting.md
```diff
@@ -0,0 +1,28 @@
...
+  ```
+- Roll back the deployment to the previous revision, using the information output from the above command:
+  
+  ```bash
+  helm rollback -n {{NAMESPACE}} {{DEPLOYMENT_NAME}} {{REV_NUM}}-1
```
sgibson91, line 23:
```suggestion
  helm rollback --namespace {{NAMESPACE}} {{DEPLOYMENT_NAME}} {{REV_NUM - 1}}
```

- Using long flag form
- `REV_NUM` statement updated to match the statement immediately below

docs/topic/troubleshooting.md
```diff
@@ -0,0 +1,28 @@
+# Troubleshooting and debugging
+
+These sections describe a few common and useful tips for troubleshooting our infrastructure.
+
+## Roll back (revert) a helm deploy
+
+Sometimes it is useful to simply **revert** a Kubernetes deployment with Helm.
+For example, if you've manually deployed something via `helm install`, and notice that something is wrong with our deployment.
```
sgibson91, line 8:
```suggestion
For example, if you've manually deployed something via `helm upgrade`, and notice that something is wrong with our deployment.
```

`helm upgrade` is the much more likely command we will be using a rollback to recover from. `helm install` implies that the chart wasn't deployed before the command was run, therefore there's nothing to rollback to. A "fix" in that scenario is to just delete the deployment.

docs/topic/troubleshooting.md
```diff
@@ -0,0 +1,28 @@
...
+  ```bash
+  helm list -n {{NAMESPACE}}
+  ```
+- Roll back the deployment to the previous revision, using the information output from the above command:
+  
+  ```bash
+  helm rollback -n {{NAMESPACE}} {{DEPLOYMENT_NAME}} {{REV_NUM}}-1
+  ```
+
+  The {{REV_NUM - 1}} simply means "deploy the previous revision number"
```
sgibson91, line 26:
```suggestion
  The {{REV_NUM - 1}} simply means "deploy the previous revision number".
  Usually, `NAMESPACE` and `DEPLOYMENT_NAME` are identical, but always best to double check.
```

docs/topic/troubleshooting.md
```diff
@@ -0,0 +1,28 @@
...
+
+This should revert the Helm deployment to the previous revision (the one just before you ran `helm install`).
```
sgibson91, line 28:
```suggestion
This should revert the Helm deployment to the previous revision (the one just before you ran `helm upgrade`).
```

sgibson91 (approved):
Left a few clarifying comments, but all good!

docs/topic/troubleshooting.md
damianavila: > Usually, `NAMESPACE` and `DEPLOYMENT_NAME` are identical, but always best to double check.

Maybe adding a few lines about how to double-check?

sgibson91: It should be obvious from the output of `helm list -n NAMESPACE` above, no? e.g.

```
$ helm list -n staging
NAME   	NAMESPACE	REVISION	UPDATED                             	STATUS  	CHART        	APP VERSION
staging	staging  	57      	2021-09-23 12:39:49.844059 +0100 BST	deployed	daskhub-0.1.0	1.0
```

damianavila:
> @damianavila and I have been trying to get @yuvipanda to share what he did as he was the one who rotated the secret.

Indeed, I suspect the process is:

1. Unencrypt https://github.com/2i2c-org/pilot-hubs/blob/master/config/secrets.yaml
2. Generate a key with openssl: `openssl rand -hex 32`
3. Replace the key
4. Encrypt the file again

but I could not get confirmation from @yuvipanda yet.

yuvipanda:
That is exactly correct, @damianavila! I did `sops config/secrets.yaml`, which opened an editor for me. I then pasted the output of `openssl rand -hex 32` there, and saved my editor - which encrypted the file automatically.

Additional tip: I didn't actually copy paste, i just deleted the current secret and typed `<ESC>:r!openssl rand -hex 32`, which ran the command and substituted its output where the cursor was. This works in vim as well as in most editors where vim emulation is turned on. I find this helpful as `openssl rand -hex 32` seems to be something we end up typing often :) Not sure if it belongs in the troubleshooting guide though.

choldgraf:
ok I'm going to merge this one and open another to add a section on rotating keys :-)

docs/topic/troubleshooting.md
damianavila: Yep, I was reading this comment isolated from the above lines, my bad 😉 .

PR #705, opened 2021-09-23
Automatically run `helm dep up` for hub templates in deployment scripts

sgibson91:
Working towards https://github.com/2i2c-org/pilot-hubs/issues/556

### DIFF
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -440,12 +440,11 @@ class Hub:
         Deploy this hub
         """
         # Ensure helm charts are up to date
-        os.chdir("hub-templates/basehub")
-        subprocess.check_call(["helm", "dep", "up"])
+        os.chdir("hub-templates")
+        subprocess.check_call(["helm", "dep", "up", "basehub"])
         if self.spec["template"] == "daskhub":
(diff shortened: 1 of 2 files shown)


deployer/hub.py
```diff
@@ -439,6 +439,13 @@ def deploy(self, auth_provider, secret_key, skip_hub_health_test=False):
         """
         Deploy this hub
         """
+        # Ensure helm charts are up to date
+        os.chdir("hub-templates/basehub")
+        subprocess.check_call(["helm", "dep", "up"])
+        if self.spec["template"] == "daskhub":
+            os.chdir("../daskhub")
+            subprocess.check_call(["helm", "dep", "up"])
+        os.chdir("../..")
```
damianavila, line 448: What about going to hub-templates, `helm dep up basehub` and `helm dep up daskhub` if the template is daskhub?
That should reduce the need to line 446 and mimic better what you are removing from the CI (and what we have with `support` above.
Just a suggestion :wink:

sgibson91: Done!

damianavila (approved):
LGTM now!

PR #706, opened 2021-09-23
Support native JupyterHub OAuthenticator in 2i2c-managed hubs

sgibson91:
### Summary

This PR is implementing support for the native JupyterHub OAuthenticator to authenticate against GitHub orgs/teams in our hubs. A config matching the thought process laid out here is currently deployed on the Pangeo staging hub and works! :tada: 

fixes https://github.com/2i2c-org/pilot-hubs/issues/625

### What has changed?

- Schema config for `config/hubs/` now allows `auth0` block to have additional properties (such as `enabled: false`) and makes the `auth0.connection` field not required so we don't have to provide it when using JupyterHub OAuthenticator
- Updates in `deployer/hub.py`:
  - Explicitly set the `JupyterHub.authenticator_class` as part of the generated config
  - Check for a partner `secrets/config/hubs/*.cluster.yaml` file. If one exists, read it in and parse the config to the `helm upgrade` command. Merging of config happens at the helm level, instead of merging dicts in Python.
- Updated `auth-management.md` docs to add process for enabling JupyterHub OAuthenticator for GitHub orgs and teams.
- Changes in `basehub/values.yaml`:
  - Where we were explicitly setting the `JupyterHub.authenticator_class` here before, it is now being set in the deployer as part of the generated config (see above)
- Added a schema for `secrets/config/hubs/*.cluster.yaml` files
  - Only the `name` and `config` parameters are allowed under the `hubs` key. Any other config should live in the main config file under `config/hubs`.


config/hubs/schema.yaml
```diff
@@ -158,7 +158,7 @@ properties:
               - basehub
               - daskhub
           auth0:
-            additionalProperties: false
+            additionalProperties: true
```
sgibson91, line 161: I needed to add this so that I could add `enabled: false` in a `*.cluster.yaml` file without experiencing a failure. It would also be nice to modify this schema in such a way that if `auth0.enabled: false` then the `auth0.connection` key is no longer required.

docs/howto/configure/auth-management.md
```diff
@@ -53,4 +55,107 @@ So we want to manage authentication by:
...
+   You should make sure the matching hub config takes one of the following forms.
+
+   To authenticate against a GitHub organisation:
+
+    ```yaml
+    hubs:
+    - name: HUB_NAME
+      auth0:
+        enabled: false
+        connection: ""  # This key is still required by the schema
```
sgibson91, line 118: Would be nice is `connection` wasn't required if `enabled: false`

docs/howto/configure/auth-management.md
```diff
@@ -53,4 +55,107 @@ So we want to manage authentication by:
...
+        jupyterhub:
+          hub:
+            config:
+              JupyterHub:
+                authenticator_class: github
+              GitHubOAuthenticator:
+                oauth_callback_url: https://{{ HUB_DOMAIN }}/hub/oauth_callback
+                allowed_organizations:
+                  - 2i2c-org:tech-team
+                  - ORG_NAME:TEAM_NAME
```
sgibson91, line 156: I included the `2i2c-org:tech-team` here since, when I tried this on the Pangeo staging deployment, I was denied access until I added myself to that team. I wasn't sure which way round we'd need like to manage this going forward, either always have the `2i2c-org:tech-team` listed in `allowed_organizations` or always add ourselves to `ORG_NAME:TEAM_NAME`.

Same discussion applies to the "just org" example above this.

docs/howto/configure/auth-management.md
```diff
@@ -53,4 +55,107 @@ So we want to manage authentication by:
...
+                scope:
+                  - read:org
+            extraConfig:
+              06-custom-authenticator: ""  # Required to override our overrides...
```
sgibson91, line 160: I needed to override this in our values, otherwise the JupyterHub returned a 500 error when you clicked on the "Login" button and the hub logs showed:

```
[E 2021-09-23 10:41:12.961 JupyterHub web:1789] Uncaught exception GET /hub/oauth_login?next=%2Fhub%2F (10.128.0.21)
    HTTPServerRequest(protocol='https', host='staging.pangeo.2i2c.cloud', method='GET', uri='/hub/oauth_login?next=%2Fhub%2F', version='HTTP/1.1', remote_ip='10.128.0.21')
      File "/usr/local/lib/python3.8/dist-packages/tornado/web.py", line 1702, in _execute
        result = method(*self.path_args, **self.path_kwargs)
      File "/usr/local/lib/python3.8/dist-packages/oauthenticator/oauth2.py", line 115, in get
        self.authorize_redirect(
      File "/usr/local/lib/python3.8/dist-packages/tornado/auth.py", line 585, in authorize_redirect
        url = self._OAUTH_AUTHORIZE_URL  # type: ignore
      File "/usr/local/lib/python3.8/dist-packages/oauthenticator/oauth2.py", line 72, in _OAUTH_AUTHORIZE_URL
        return self.authenticator.authorize_url
      File "/usr/local/lib/python3.8/dist-packages/traitlets/traitlets.py", line 575, in __get__
        return self.get(obj, cls)
      File "/usr/local/lib/python3.8/dist-packages/traitlets/traitlets.py", line 538, in get
        default = obj.trait_defaults(self.name)
      File "/usr/local/lib/python3.8/dist-packages/traitlets/traitlets.py", line 1578, in trait_defaults
        return self._get_trait_default_generator(names[0])(self)
      File "/usr/local/lib/python3.8/dist-packages/traitlets/traitlets.py", line 975, in __call__
        return self.func(*args, **kwargs)
      File "/usr/local/lib/python3.8/dist-packages/oauthenticator/auth0.py", line 69, in _authorize_url_default
        return "https://%s.auth0.com/authorize" % self.auth0_subdomain
      File "/usr/local/lib/python3.8/dist-packages/traitlets/traitlets.py", line 575, in __get__
        return self.get(obj, cls)
      File "/usr/local/lib/python3.8/dist-packages/traitlets/traitlets.py", line 538, in get
        default = obj.trait_defaults(self.name)
      File "/usr/local/lib/python3.8/dist-packages/traitlets/traitlets.py", line 1578, in trait_defaults
        return self._get_trait_default_generator(names[0])(self)
      File "/usr/local/lib/python3.8/dist-packages/traitlets/traitlets.py", line 975, in __call__
        return self.func(*args, **kwargs)
      File "/usr/local/lib/python3.8/dist-packages/oauthenticator/auth0.py", line 52, in _auth0_subdomain_default
        raise ValueError(
    ValueError: Please specify $AUTH0_SUBDOMAIN env or CustomOAuthenticator.auth0_subdomain config
```

If there's a more elegant way to do this, I'd be happy to hear suggestions.

deployer/hub.py
```diff
@@ -439,14 +439,28 @@ def deploy(self, auth_provider, secret_key, skip_hub_health_test=False):
         """
         Deploy this hub
         """
+        if self.spec["auth0"]['enabled'] == False:
```
yuvipanda, line 442: I think we should read secret config file unconditionally if it exists, regardless of what auth0 is set to. This lets us put in other secret values (like a GCP service account JSON in environment, for example) without hassle.

docs/howto/configure/auth-management.md
```diff
@@ -53,4 +55,107 @@ So we want to manage authentication by:
...
+Presently, this involves a few more manual steps than the `auth0` setup described above.
+
+1. **Create a GitHub OAuth App.**
+   This can be achieved by following [GitHub's documentation](https://docs.github.com/en/developers/apps/building-oauth-apps/creating-an-oauth-app).
+   - When naming the application, please follow the convention `<CLUSTER_NAME>-<HUB_NAME>` for consistency, e.g. `2i2c-staging` is the OAuth app for the staging hub running on the 2i2c cluster.
+   - The Homepage URL should match that in the `domain` field of the appropriate `*.cluster.yaml` file in the `pilot-hubs` repo.
+   - The authorisation callback URL is the homepage url appended with `/hub/oauth_callback`
+   - Once you have created the OAuth app, make a new of the client ID, generate a client secret and then hold on to these values for a future step
+
+2. **Transfer the OAuth App to the `2i2c-org` GitHub account.**
```
yuvipanda, line 78: I'm slightly confused about wether we can't just create it ourselves in the 2i2c-org org.

docs/howto/configure/auth-management.md
yuvipanda: We can just make `connection` have a `required: false` field in the schema.

secrets/config/hubs/schema.yaml
```diff
@@ -0,0 +1,24 @@
+$schema: 'http://json-schema.org/draft-07/schema#'
```
yuvipanda, line 1: So the only things that can go here are name and config - everything else should be in the regular config file. This is because we only merge things under `config` via helm, and not the whole file. Is my understanding here correct?

docs/howto/configure/auth-management.md
yuvipanda: We currently have auth0 set as the authenticator unconditionally here: https://github.com/2i2c-org/pilot-hubs/blob/8ab995af71cd3c21481176061c51cbc0647f8b60/hub-templates/basehub/values.yaml#L216.

We then set it unconditionally *again* in https://github.com/2i2c-org/pilot-hubs/blob/8ab995af71cd3c21481176061c51cbc0647f8b60/hub-templates/basehub/values.yaml#L320.

Let's unset both of those, and set it conditionally here instead: https://github.com/2i2c-org/pilot-hubs/blob/8ab995af71cd3c21481176061c51cbc0647f8b60/deployer/hub.py#L355

yuvipanda (commented):
\o/ AWESOME. I love the general approach here, +1 to the simplicity of it :)

docs/howto/configure/auth-management.md
sgibson91: Yes, I think there is a "Switch account" button at the top of the page that slipped my notice when I went through this.

secrets/config/hubs/schema.yaml
sgibson91: Yes, correct

docs/howto/configure/auth-management.md
```diff
@@ -53,4 +55,107 @@ So we want to manage authentication by:
...
+    ```
+
+    ```{note}
+    Add the `basehub` key between `config` and `jupyterhub` for `daskhub` deployments.
+    ```
+
+    ```{admonition}
+    Make sure this is encrypted with `sops` before committing it to the repository!
+
+    `sops -i -e secrets/config/hub/*.cluster.yaml`
```
sgibson91, line 105:
```suggestion
    `sops -i -e secrets/config/hubs/*.cluster.yaml`
```

Fix typo

docs/howto/configure/auth-management.md
sgibson91: Done in commit 99e7f82

deployer/hub.py
sgibson91: Done in 4f53a4c

docs/howto/configure/auth-management.md
sgibson91: I gave this a shot in commit ffb4b95 but please let me know if I got it right. I was a bit confused about the mixture of `oauthenticator.auth0.Auth0OAuthenticator` and `CustomOAuthenticator`.

docs/howto/configure/auth-management.md
```diff
@@ -38,19 +40,113 @@ So we want to manage authentication by:
...
 ```{admonition} Switching auth
-Switching authentication for a pre-existing hub will simply create new usernames. Any pre-existing users will no longer be able to access their accounts (although administrators will be able to do so). If you have pre-existing users and want to switch the hub authentication, rename the users to the new auth pattern (e.g. convert github handles to emails).
-```
\ No newline at end of file
+Switching authentication providers (e.g. from GitHub to Google) for a pre-existing hub will simply create new usernames. Any pre-existing users will no longer be able to access their accounts (although administrators will be able to do so). If you have pre-existing users and want to switch the hub authentication, rename the users to the new auth pattern (e.g. convert github handles to emails).
+```
+
+## Native JupyterHub OAuthenticator for GitHub Orgs and Teams
+
+```{admonition}
```
choldgraf, line 60:
````suggestion
```{note}
````

If you use `admonition` then you need to give a title, but `note` just lets you add the comment w/o a title :-)

File: docs/howto/configure/auth-management.md [hunk 10]
```diff
@@ -38,19 +40,113 @@ So we want to manage authentication by:
...
+This setup is currently only supported for communities that **require** authentication via a GitHub organisation or team.
+
+We may update this policy in the future.
+```
+
+For communities that require authenticating users against a GitHub organisation or team, we instead use the native JupyterHub OAuthenticator.
```
choldgraf, line 66:
```suggestion
For communities that require authenticating users against [a GitHub organisation or team](https://docs.github.com/en/organizations), we instead use the [native JupyterHub OAuthenticator](https://github.com/jupyterhub/oauthenticator).
```

docs/howto/configure/auth-management.md
```diff
@@ -38,19 +40,113 @@ So we want to manage authentication by:
...
+Presently, this involves a few more manual steps than the `auth0` setup described above.
+
+1. **Create a GitHub OAuth App.**
+   This can be achieved by following [GitHub's documentation](https://docs.github.com/en/developers/apps/building-oauth-apps/creating-an-oauth-app).
+   Use the "Switch account" button at the top of your settings page to make sure you have `2i2c-org` selected.
+   That way, the app will be owned by the `2i2c-org` GitHub org, rather than your personal GitHub account.
```
choldgraf, line 72:
```suggestion
   - Use the "Switch account" button at the top of your settings page to make sure you have `2i2c-org` selected.
     That way, the app will be owned by the `2i2c-org` GitHub org, rather than your personal GitHub account.
```

docs/howto/configure/auth-management.md
```diff
@@ -38,19 +40,113 @@ So we want to manage authentication by:
...
+   - When naming the application, please follow the convention `<CLUSTER_NAME>-<HUB_NAME>` for consistency, e.g. `2i2c-staging` is the OAuth app for the staging hub running on the 2i2c cluster.
+   - The Homepage URL should match that in the `domain` field of the appropriate `*.cluster.yaml` file in the `pilot-hubs` repo.
+   - The authorisation callback URL is the homepage url appended with `/hub/oauth_callback`
```
choldgraf, line 75:
```suggestion
   - The authorisation callback URL is the homepage url appended with `/hub/oauth_callback`. For example, `staging.pilot.2i2c.cloud/hub/oauth_callback`.
```

docs/howto/configure/auth-management.md
choldgraf: it sounds like this one got resolved? is that right?

choldgraf (commented):
Some quick thoughts from me - in general this looks good! I like the simplicity of it, it is surprisingly not as much effort as I thought it would be :-)

I won't "approve" as I don't understand the implementation details as well

docs/howto/configure/auth-management.md
sgibson91: I don't think we've had the discussion yet? I'd like the team to agree on a process to either always including `2i2c-org:tech-team` in the config, or being added to the community's org/team

docs/howto/configure/auth-management.md
sgibson91: Also if I understand correctly, each hub in every `*.cluster.yaml` file will need to be updated to have `JupyterHub.authenticator_class` explicitly specified now

deployer/hub.py
```diff
@@ -439,14 +440,29 @@ def deploy(self, auth_provider, secret_key, skip_hub_health_test=False):
         """
         Deploy this hub
         """
+        # Check if this hub has any secret config. If yes, read it in.
```
sgibson91, line 443:
```suggestion
        # Check if this cluster has any secret config. If yes, read it in.
```

deployer/hub.py
```diff
@@ -444,6 +461,7 @@ def deploy(self, auth_provider, secret_key, skip_hub_health_test=False):
                 # we should put the config from config/hubs last.
                 '-f', generated_values_file.name,
                 '-f', values_file.name,
+                '-f', secret_values_file.name,
```
damianavila, line 464: I see value in having this one separated from the main thing and applied at the helm step.
I think it is in agreement with the overall idea and the encrypted config pieces under secret.

deployer/hub.py
```diff
@@ -426,14 +427,30 @@ def deploy(self, auth_provider, secret_key, skip_hub_health_test=False):
...
+        secret_config_path = Path(os.getcwd()) / "secrets/config/hubs" / f'{self.cluster.spec["name"]}.cluster.yaml'
+
+        if os.path.exists(secret_config_path):
+            with decrypt_file(secret_config_path) as decrypted_file_path:
+                with open(decrypted_file_path) as f:
+                    secret_config = yaml.load(f)
+
+            hubs = secret_config["hubs"]
+            secret_hub_config = next((hub for i, hub in enumerate(hubs) if hubs[i]["name"] == self.spec["name"]), None)
+            secret_hub_config = secret_hub_config["config"]
```
damianavila, line 441: If `secret_hub_config` is `None` (ie. putting the wrong name?), 441 will trigger an exception, right?

docs/howto/configure/auth-management.md
```diff
@@ -38,19 +40,112 @@ So we want to manage authentication by:
...
-Switching authentication for a pre-existing hub will simply create new usernames. Any pre-existing users will no longer be able to access their accounts (although administrators will be able to do so). If you have pre-existing users and want to switch the hub authentication, rename the users to the new auth pattern (e.g. convert github handles to emails).
-```
\ No newline at end of file
+Switching authentication providers (e.g. from GitHub to Google) for a pre-existing hub will simply create new usernames. Any pre-existing users will no longer be able to access their accounts (although administrators will be able to do so). If you have pre-existing users and want to switch the hub authentication, rename the users to the new auth pattern (e.g. convert github handles to emails).
+```
+
+## Native JupyterHub OAuthenticator for GitHub Orgs and Teams
+
+```{note}
+This setup is currently only supported for communities that **require** authentication via a GitHub organisation or team.
```
damianavila, line 60:
```suggestion
This setup is currently only supported for communities that **require** authentication via a GitHub organization or team.
```

I learnt `organisation` is accepted in British, but since we are using organization in other places, let's be consistent about it?

docs/howto/configure/auth-management.md
```diff
@@ -38,19 +40,112 @@ So we want to manage authentication by:
...
+
+We may update this policy in the future.
+```
+
+For communities that require authenticating users against [a GitHub organisation or team](https://docs.github.com/en/organizations), we instead use the [native JupyterHub OAuthenticator](https://github.com/jupyterhub/oauthenticator).
```
damianavila, line 65:
```suggestion
For communities that require authenticating users against [a GitHub organization or team](https://docs.github.com/en/organizations), we instead use the [native JupyterHub OAuthenticator](https://github.com/jupyterhub/oauthenticator).
```

docs/howto/configure/auth-management.md
```diff
@@ -38,19 +40,112 @@ So we want to manage authentication by:
...
+        jupyterhub:
+          hub:
+            config:
+              GitHubOAuthenticator:
+                client_id: CLIENT_ID
+                client_secret: CLIENT_SECRET
+    ```
+
+    ```{note}
+    Add the `basehub` key between `config` and `jupyterhub` for `daskhub` deployments.
```
damianavila, line 93: A short snippet showing where to put it usually helps in addition to words, IMHO.

docs/howto/configure/auth-management.md
```diff
@@ -38,19 +40,112 @@ So we want to manage authentication by:
...
+    ```{note}
+    Make sure this is encrypted with `sops` before committing it to the repository!
+
+    `sops -i -e secrets/config/hubs/*.cluster.yaml`
+    ```
+
+3. **Edit the non-secret config under `config/hubs`.**
+   You should make sure the matching hub config takes one of the following forms.
+
+   To authenticate against a GitHub organisation:
```
damianavila, line 105: OK, maybe forget about my previous comment about `organization` :wink:

docs/howto/configure/auth-management.md
damianavila: I would go with the first option, I think...

docs/howto/configure/auth-management.md
```diff
@@ -38,19 +40,112 @@ So we want to manage authentication by:
...
+                client_id: CLIENT_ID
+                client_secret: CLIENT_SECRET
+    ```
+
+    ```{note}
+    Add the `basehub` key between `config` and `jupyterhub` for `daskhub` deployments.
+    ```
+
+    ```{note}
+    Make sure this is encrypted with `sops` before committing it to the repository!
```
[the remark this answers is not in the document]
sgibson91: I wonder if we can use a `.gitattributes` file the same way the mybinder.org-deploy repo encrypts it's secrets with git-crypt? https://github.com/jupyterhub/mybinder.org-deploy/blob/master/.gitattributes

hub-templates/basehub/values.yaml
```diff
@@ -317,8 +314,6 @@ jupyterhub:
               resp['name'] = resp['name'].split('|')[-1]
             return resp
 
-        c.JupyterHub.authenticator_class = CustomOAuthenticator
-
```
damianavila, line 321: I do not get this removal... don't you need to keep this in auth0 enabled clusters?

docs/howto/configure/auth-management.md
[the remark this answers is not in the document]
sgibson91: There's a blog implementing this with git-hooks here https://levelup.gitconnected.com/using-sops-and-git-hook-to-share-secrets-part-2-38518be67a14

deployer/hub.py
sgibson91: Good question, I don't know! I copied this line from here:

https://github.com/2i2c-org/pilot-hubs/blob/e96e7bcded187870dc2e07d6626de8a12586ed32/deployer/__main__.py#L89-L96

And since that part handles sorting the hub names and parsing them to `deploy`, I don't think the `None` would ever actually be triggered here.

However, syntactically, I now think `None` should be something like `{"config": {}}` in order to match what is being output by the `else` clause.

damianavila: > However, syntactically, I now think None should be something like {"config": {}}

Yep, I was expecting something like that.

docs/howto/configure/auth-management.md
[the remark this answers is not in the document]
damianavila: Interesting links, thanks for sharing them.
I guess we should capture this conversation in another issue, WDYT?

hub-templates/basehub/values.yaml
damianavila: I am not sure the alternative you wrote in the previous comment will work in your pangeo (non-auth0) case.
What about doing all the stuff under 06-custom-authenticator conditional if `auth0` is configured? As it is done [here](https://github.com/2i2c-org/pilot-hubs/blob/master/hub-templates/basehub/values.yaml#L328) with the `scratch_bucket`?

docs/howto/configure/auth-management.md
choldgraf: @sgibson91 ah sorry I had misunderstood - my feeling is that we should go with option 1 because that's an action we have more control over, and I think it will be more reliable to rely on ourselves rather than other communities to take action. Does that make sense?

hub-templates/basehub/values.yaml
GeorgianaElena: > So it _might_ be that this line

> https://github.com/sgibson91/pilot-hubs/blob/fe4ab0101b6900997b9c1c0fc90e62b1b57e1720/deployer/hub.py#L351

> should be the following instead

> ```
> generated_config['jupyterhub']['hub']['config']['JupyterHub']['authenticator_class'] = 'CustomOAuthenticator'
> ```

@sgibson91, I believe so too!

At some point, I did [this chart](https://pilot-hubs.2i2c.org/en/latest/topic/hub-templates.html#hub-templates) about how different configs are put together. I'm not sure if it's super accurate now, but for me, at least it provides a visual starting point, though I'm still a bit confused about them (hence, my forever-taking PR with the staff admin users :roll_eyes: )

One thing that's not accurate in that chart (*I think*), is that although `values.yaml` template is the first one that gets loaded, , the `extra_config` in `values.yaml` is actually the last one that get loaded by the hub. At least that's what I understand from the hub logs first few lines:
```
Loading extra config: 01-working-dir
Loading extra config: 02-prometheus
Loading extra config: 03-no-setuid
Loading extra config: 04-custom-theme
Loading extra config: 05-custom-admin
Loading extra config: 06-custom-authenticator
Loading extra config: 07-cloud-storage-bucket
```

> I am not sure the alternative you wrote in the previous comment will work in your pangeo (non-auth0) case.

@damianavila, it [shouldn't be set](https://github.com/sgibson91/pilot-hubs/blob/fe4ab0101b6900997b9c1c0fc90e62b1b57e1720/deployer/hub.py#L336) for non-auth0 cases.

sgibson91: Thank you @GeorgianaElena! I will update that line and see what people think. Also very nice chart! 😉 

---

@damianavila for clarity's sake, we now need to _dynamically_ set the value of `JupyterHub.authenticator_class` based on which authentication method we are using, hence why these values are being removed from the chart. So when using auth0, `authenticator_class` will be set in the deployer in this block:

https://github.com/sgibson91/pilot-hubs/blob/fe4ab0101b6900997b9c1c0fc90e62b1b57e1720/deployer/hub.py#L336-L352

And when using the native GitHubOAuthenticator, we set the class in the `*.cluster.yaml` hub config, as demonstrated in the accompanying docs, e.g. Step 3 here: https://github.com/sgibson91/pilot-hubs/blob/jupyterhub-oauth-github/docs/howto/configure/auth-management.md#native-jupyterhub-oauthenticator-for-github-orgs-and-teams

---

> What about doing all the stuff under 06-custom-authenticator conditional if auth0 is configured? As it is done here with the scratch_bucket?

I am not convinced that method will work. If I understand the code correctly, the highest level config key that `get_config` reads in is everything under `jupyterhub` and the `auth0.enabled` key is much higher in the hierarchy of keys, adjacent to hubs, so I'm not sure the `z2jh` module can even see it?

Image to clarify what I mean by the hierarchy:

config

docs/howto/configure/auth-management.md
sgibson91: Ok great, in which case there's no action as the documentation is already reflecting option 1 :)

deployer/hub.py
sgibson91: Done in [`c937694`](https://github.com/2i2c-org/pilot-hubs/pull/706/commits/c93769442483127f4fd7a5cbe5ba98860ae6217d)

docs/howto/configure/auth-management.md
sgibson91: Done in [`5ca78c9`](https://github.com/2i2c-org/pilot-hubs/pull/706/commits/5ca78c96d690e7ce8ff6b30228e18631d1c23b01)

hub-templates/basehub/values.yaml
sgibson91: I updated the `authenticator_class` for auth0 hubs in [`b1d6aac`](https://github.com/2i2c-org/pilot-hubs/pull/706/commits/b1d6aac999bc8d51890da7a99f0f6aab2ea18af6)

docs/howto/configure/auth-management.md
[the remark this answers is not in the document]
sgibson91: I opened https://github.com/2i2c-org/pilot-hubs/issues/719

hub-templates/basehub/values.yaml
damianavila: Hey, thanks @GeorgianaElena and @sgibson91 for the discussion and the clarifications!
I now understand why I did not catch it... I totally missed [this line](https://github.com/sgibson91/pilot-hubs/blob/fe4ab0101b6900997b9c1c0fc90e62b1b57e1720/deployer/hub.py#L336). This is why I proposed the alternative "conditional approach" (without any testing, of course 😉 ). 
In fact, I was expecting that auth0 "enabled" property to have some representation in the schema (I guess that is another discussion to have...), and since I did not find it there, and because I did not check some lines above 🤦 , I assumed there was not a conditional approach.
I concur with the last commit you pushed @sgibson91.

docs/howto/configure/auth-management.md
sgibson91: From discussion in [this thread](https://github.com/2i2c-org/pilot-hubs/pull/706/#discussion_r717712945) we agreed that `authenticaotr_class` should be `CustomOAuthenticator` instead. Change was pushed [here](https://github.com/2i2c-org/pilot-hubs/pull/706/commits/b1d6aac999bc8d51890da7a99f0f6aab2ea18af6)

hub-templates/basehub/values.yaml
damianavila: >so I'm not sure the z2jh module can even see it?

Yep, you would need to have somehow an `auth` under custom as well for being able to catch it... but no need to test that now 😜 .

damianavila (approved):
I think this one is ready!
Thanks, @sgibson91 for the work you have done here and to bear with my comments/questions.

PR #707, opened 2021-09-23
Enabling GitHub Teams Authentication for the Pangeo Staging Hub

sgibson91:
This PR enables authentication against the `@pangeo-data/us-central1-b-gcp` GitHub Team on the Pangeo staging hub. Presently, I have explicitly defined the config for the prod hub so that we can control when we update prod to match staging, since prod has active users.

**2021-10-05 10:22 BST:** I will merge this in 24 hours unless I receive an approving review before then


sgibson91:
I have deployed this onto staging and it works for me. I have asked @jhamman and @scottyhq in the `#pangeo-pilot-hubs` slack channel to double check they can login too (since neither of them have a user account on the staging hub yet :) )

config/hubs/pangeo-hubs.cluster.yaml
```diff
@@ -24,14 +24,12 @@ support:
       controller:
         admissionWebhooks:
           enabled: false
-    nfs-server-provisioner:
-      enabled: false
```
damianavila, line 28: I guess this is not related at all, just removing old stuff??

config/hubs/pangeo-hubs.cluster.yaml
```diff
@@ -74,6 +72,15 @@ hubs:
                   - choldgraf
                   - rabernat
                 admin_users: *staging_users
+              JupyterHub:
+                authenticator_class: github
+              GitHubOAuthenticator:
+                oauth_callback_url: https://staging.pangeo.2i2c.cloud/hub/oauth_callback
+                allowed_organizations:
+                  - pangeo-data:us-central1-b-gcp
+                  - 2i2c-org:tech-team
```
damianavila, line 81: Since the tech team is here now, maybe we can remove the allowed_users from above and just keep them as admin_users? Am I correct in my interpretation?

config/hubs/pangeo-hubs.cluster.yaml
sgibson91: Correct, it's disabled by default here: https://github.com/2i2c-org/pilot-hubs/blob/85a9adb707aef3bbcb8a95947960114a860a8761/support/values.yaml#L75-L76

config/hubs/pangeo-hubs.cluster.yaml
sgibson91: I don't think so no. I don't think `allowed_organizations` grants admin privileges. It would be a nice to have though!

sgibson91: Sorry, just re-read this and parsed it properly. I _thought_ you meant removing the admin block! Yes, we can probably remove the `allowed_users` block.

sgibson91: You are in fact correct and this block needs to go, otherwise it prevents folk from logging in _even if_ they are a member of the appropriate org/team. Fixed in https://github.com/2i2c-org/pilot-hubs/pull/707/commits/ee39b165fa5a91c61517ff2ad9d441e0d446ddef

sgibson91:
We fixed this so ready to be reviewed/merged!

config/hubs/pangeo-hubs.cluster.yaml
```diff
@@ -143,4 +149,109 @@ hubs:
...
+                  name: 2i2c
+                  url: https://2i2c.org
+                funded_by:
+                  name: The Gordon and Betty Moore Foundation
+                  url: https://www.moore.org/
+          hub:
+            config:
+              Authenticator:
+                allowed_users: &prod_users
+                  - sgibson91
```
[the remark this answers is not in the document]
yuvipanda: Or rather, they should just be listed as admin users, not in both.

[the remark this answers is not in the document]
sgibson91: Not yet. This PR is only updating the **staging** hub, and you've wandered into the config for the **prod** hub. I intended to coordinate with Ryan when we move the authentication method for prod over since it will affect his class and they'll need to be members of `pangeo-data:us-central1-b-gcp`, no? Or does that fact that Ryan has already added them as `allowed_users` not matter?

(Either, I was intending to move prod across in a second PR)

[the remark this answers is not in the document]
yuvipanda: aaah, sorry - forgot this was partially already deployed!

sgibson91:
I will merge this in 24 hours unless I receive an approving review before then

config/hubs/pangeo-hubs.cluster.yaml
damianavila: Nice to know my intuition is still accurate 😜  .

damianavila (approved):
LGTM, great work @sgibson91!

PR #711, opened 2021-09-24
Add secrets section

choldgraf:
Follow up to #704 - this adds a `secrets` section and a short guide on rotating the secrets.

### DIFF
--- a/docs/topic/secrets.md
+++ b/docs/topic/secrets.md
@@ -1,0 +1,40 @@
+# Secrets and private keys
+
+## Where are secrets stored
+
+Most secrets are stored in this location:
+
+https://github.com/2i2c-org/pilot-hubs/blob/master/config/secrets.yaml
+
+This contains a configuration file encrypted with [`sops`](https://github.com/mozilla/sops).
+It is used for things like authentication with `Auth0`, image registries, etc.
+
+:::{seealso}
+For information about how to set up `sops`, see [the team compass documentation](tc:secrets:sops)
+:::
+
+## How to rotate / change secrets
+
+Sometimes we need to rotate the secret keys used in our repository.
+For example, if a service we use has become compromised, and we need to generate new keys in order to protect the infrastructure.
+
+To rotate our secrets, take these steps:
+
+1. Unencrypt [our configuration file](https://github.com/2i2c-org/pilot-hubs/blob/master/config/secrets.yaml) (see [the team compass documentation](tc:secrets:sops) for instructions).
+2. Generate a new key with `openssl`:
+   
+   ```
+   openssl rand -hex 32
+   ```
+   
+   This will return a random hash that looks something like this:
+
+   ```
+   4a87d32d435f5471b5852f30f1adcc29d11b39035d68b81720130701e65fa585
+   ```
+
+3. Find the key you'd like to replace, and replace its value with the hash that you've generated above.
+4. Re-encrypt the file with `sops`.
+5. Commit the file to the repository and push.
+
+You have now rotated the secret for this key!
--- a/docs/index.md
+++ b/docs/index.md
@@ -44,6 +44,7 @@ topic/hub-templates.md
 topic/storage-layer.md
 topic/terraform.md
 topic/cluster-design.md
+topic/secrets.md
 topic/troubleshooting.md
 ```
 


docs/topic/secrets.md
```diff
@@ -0,0 +1,40 @@
+# Secrets and private keys
+
+## Where are secrets stored
+
+Most secrets are stored in this location:
+
+https://github.com/2i2c-org/pilot-hubs/blob/master/config/secrets.yaml
```
damianavila, line 7: Well, actually, most secrets live here instead :wink: : https://github.com/2i2c-org/pilot-hubs/tree/master/secrets

docs/topic/secrets.md
```diff
@@ -0,0 +1,40 @@
...
+   openssl rand -hex 32
+   ```
+   
+   This will return a random hash that looks something like this:
+
+   ```
+   4a87d32d435f5471b5852f30f1adcc29d11b39035d68b81720130701e65fa585
+   ```
+
+3. Find the key you'd like to replace, and replace its value with the hash that you've generated above.
```
damianavila, line 36: Maybe we should be explicit about replacing the secret_key? https://github.com/2i2c-org/pilot-hubs/blob/master/config/secrets.yaml#L5

damianavila: Although re-reading this one, it seems you wrote it in a general way...

docs/topic/secrets.md
choldgraf: Ha good point - do we have it documented when to use one folder and not another? If not we can add that here

damianavila: > do we have it documented when to use one folder and not another?

Not sure if we have an explicit reference...

choldgraf: If we don't have an explicit reference, could you eyeball an informal guideline to use? Like are there certain kinds of secrets that we store in one folder vs. another?

damianavila: Sure, 

> https://github.com/2i2c-org/pilot-hubs/blob/master/config/secrets.yaml

The above one deals auth0 credentials and a general proxy secret for our hubs. More info here: https://github.com/2i2c-org/pilot-hubs/blob/master/deployer/__main__.py#L57-L83

Whereas 

> https://github.com/2i2c-org/pilot-hubs/tree/master/secrets

deals with cluster authentication for all the cloud providers we support. More info here: https://github.com/2i2c-org/pilot-hubs/blob/master/deployer/hub.py#L39-L46 and https://github.com/2i2c-org/pilot-hubs/blob/master/deployer/hub.py#L126-L216.

choldgraf: OK I'll add those into this PR.

Though, I was looking for more like a _reason_ that we use two different places. Like "We store authentication credentials in XXX, because of YYY, and we store cluster authentication credentials in ZZZ, because of WWW"

damianavila: Oh... sorry, I should have been more clear...
* General secrets (used by all of our clusters): https://github.com/2i2c-org/pilot-hubs/blob/master/config/secrets.yaml
* Cluster specific secrets: https://github.com/2i2c-org/pilot-hubs/tree/master/secrets

docs/topic/secrets.md
```diff
@@ -0,0 +1,50 @@
...
+(secrets:locations)=
+## Where are secrets stored
+
+Most secrets are stored in one of two locations:
+
+https://github.com/2i2c-org/infrastructure/blob/master/config/secrets.yaml
+: Secrets that are shared across all of our hub and cluster deployments.
+
+https://github.com/2i2c-org/pilot-hubs/tree/master/secrets
+: Secrets that are specific to each cluster / hub that we run (one JSON file for each cluster). For example, `Auth0` and proxy secrets.
```
damianavila, line 12: > For example, `Auth0` and proxy secrets.

This is incorrect, Auth0 and the secret key (not proxy secret) live in the above config file (these are general stuff, not cluster-specific stuff): https://github.com/2i2c-org/infrastructure/blob/master/config/secrets.yaml, you can see the keys without decrypting the file.

docs/topic/secrets.md
```diff
@@ -0,0 +1,50 @@
+# Secrets and private keys
+
+(secrets:locations)=
+## Where are secrets stored
+
+Most secrets are stored in one of two locations:
+
+https://github.com/2i2c-org/infrastructure/blob/master/config/secrets.yaml
+: Secrets that are shared across all of our hub and cluster deployments.
```
choldgraf, line 9:
```suggestion
: Secrets that are shared across all of our hub and cluster deployments, such as Auth0 secrets.
```

docs/topic/secrets.md
choldgraf: I just tried to update this - does that look better?

PR #712, opened 2021-09-24
Redirect test output when we are running on the CI

### LINKED ISSUE #694: Redirecting the test output (to prevent the CI from leaking secrets) is preventing us to get the actual error locally
### Description

I faced this issue investigating failures at the time to deploy openscapes.
More details here: https://github.com/2i2c-org/pilot-hubs/pull/691#issuecomment-922320344

### Value / benefit

We should be able to easily get the test output when we are working locally whereas we redirect the output in the CI environment.

### Implementation details

A proposal lives in this comment: https://github.com/2i2c-org/pilot-hubs/pull/691#issuecomment-922320344 (check diff) and people showed interest to have this piece as a new PR.

### Tasks to complete

- [x] PR diff with the proposed implementation
- [ ] Discuss details about the env variable that we may use to detect the GitHub env

### Updates

_No response_

damianavila:
The idea of this PR is just redirecting to dev/null only if we are running the deployer on the CI and get the full
output when running locally.

Accordingly, to [GitHub docs about environment variables](https://docs.github.com/en/actions/learn-github-actions/environment-variables), the `CI` is always set to `true`. I think that should be enough to "detect" the CI environment.

Closes #694


damianavila:
Picked some reviewers but everyone is invited to review :wink:

deployer/hub.py
```diff
@@ -458,7 +458,20 @@ def deploy(self, auth_provider, secret_key, skip_hub_health_test=False):
                 # This can contain sensitive info - so we hide stderr
                 # FIXME: Don't use pytest - just call a function instead
                 print("Running hub health check...")
-                with open(os.devnull, 'w') as dn, redirect_stderr(dn), redirect_stdout(dn):
+                # Show errors locally but redirect on CI
+                gh_ci = os.environ.get('CI', "false")
+                if gh_ci == "true":
+                    print("Testing on CI, redirected output")
```
choldgraf, line 464:
```suggestion
                    print("Testing on CI, not printing output")
```
just to be a bit more explicit

deployer/hub.py
```diff
@@ -458,7 +458,20 @@ def deploy(self, auth_provider, secret_key, skip_hub_health_test=False):
...
+                    with open(os.devnull, 'w') as dn, redirect_stderr(dn), redirect_stdout(dn):
+                        exit_code = pytest.main([
+                            "-q",
+                            "deployer/tests",
+                            "--hub-url", hub_url,
+                            "--api-token", service_api_token,
+                            "--hub-type", self.spec['template']
+                        ])
```
choldgraf, line 472: Do you think that we could define these arguments in a single variable like:

```python
pytest_args = [
    "-q",
    "deployer/tests",
    "--hub-url", hub_url,
    "--api-token", service_api_token,
    "--hub-type", self.spec['template']
]
```

and then reuse them in each block below, like:

```python
exit_code = pytest.main(pytest_args)
```

so that it's easier to tell that both parts of the `if` statement are basically doing the same thing?

choldgraf (commented):
a few quick thoughts but it LGTM in general!

deployer/hub.py
damianavila: Makes total sense to me!

damianavila: Pushed change in the last commit.

GeorgianaElena (approved):
LGTM :rocket:

sgibson91 (approved):

PR #714, opened 2021-09-26
Add special cases section

choldgraf:
There are some hubs that are "different" from our main `pilot-hubs/` deployments. These are harder to maintain for a variety of reasons, and we should discourage this practice in general and migrate these special-case hubs to our pilot-hubs/ infrastrucutre.

In the meantime, we should document these hubs more explicitly so it is clear what makes these hubs special, and how we can maintain them nonetheless.

This PR adds a short section on "special cases" to our list of running JupyterHubs. It also adds the U of T hub to this list, and notes a few special considerations.

### DIFF
--- a/docs/reference/hubs.md
+++ b/docs/reference/hubs.md
@@ -1,4 +1,9 @@
-# List of running hubs
+# JupyterHubs we manage
+
+Most of our JupyterHubs are centrally configured and deployed in [the `pilot-hubs/` repository](https://github.com/2i2c-org/pilot-hubs).
+However, there are a few hubs that are special-cased as well, and we list both on this page.
+
+## A complete list of our JupyterHubs
 
 Here's a list of running hubs.
 
@@ -27,3 +32,28 @@ $(document).ready( function () {
     });
 } );
 </script>
+
+## Special-cased hubs
+
+There are some hubs that we treat differently from others.
+This is heavily discouraged, but may be necessary to meet the needs of certain communities before our deployment infrastructure can accommodate them.
+
+### University of Toronto
+
+We run a JupyterHub for the University of Toronto.
+This is configured and deployed at [the `utoronto-deploy` repository](https://github.com/utoronto-2i2c/jupyterhub-deploy).
+
+**Special cases**
+
+There are a few special cases about this deployment:
+
+- **Authentication**: Uses [Azure Active Directory](https://azure.microsoft.com/en-us/services/active-directory/).
+  
+  This assigns each username a unique ID, like `adb7ebad-9fb8-481a-be4c-6c0a8b4se670`.
+- **U. Toronto provides us credentials for authentication**: Our hub uses a credential that allows us to authenticate with the U of T AzureAD service.
+  This credential must be **renewed annually** by the University of Toronto.
+  They must send us the new credential each year, or the hub will no longer be able to authenticate users.
+  See [this incident issue](https://github.com/2i2c-org/pilot-hubs/issues/637) for an example of what happens if these credentials aren't renewed.
+- **User image must be manually built and pushed**: The [user image for this hub](https://github.com/utoronto-2i2c/jupyterhub-deploy/tree/staging/deployments/utoronto/image) is quite large, as it accommodates a number of different use-cases. GitHub actions may not be able to build it, and so we need to build and pushed to the image registry.
+- **Runs on U Toronto Azure project**. This hub runs on the U of T Azure project, and their administrators must provide us access to their cloud in order to directly work on the hub.
+  Currently, only @yuvipanda and @GeorgianaElena have access to this hub.


docs/reference/hubs.md
```diff
@@ -26,4 +31,29 @@ $(document).ready( function () {
...
+  
+  This assigns each username a unique ID, like `adb7ebad-9fb8-481a-be4c-6c0a8b4se670`.
+- **U. Toronto provides us credentials for authentication**: Our hub uses a credential that allows us to authenticate with the U of T AzureAD service.
+  This credential must be **renewed annually** by the University of Toronto.
+  They must send us the new credential each year, or the hub will no longer be able to authenticate users.
+  See [this incident issue](https://github.com/2i2c-org/pilot-hubs/issues/637) for an example of what happens if these credentials aren't renewed.
+- **User image must be manually built and pushed**: The [user image for this hub](https://github.com/utoronto-2i2c/jupyterhub-deploy/tree/staging/deployments/utoronto/image) is quite large, as it accommodates a number of different use-cases. GitHub actions may not be able to build it, and so we need to build and pushed to the image registry.
+- **Runs on U Toronto Azure project**. This hub runs on the U of T Azure project, and their administrators must provide us access to their cloud in order to directly work on the hub.
+  Currently, only @yuvipanda and @GeorgianaElena have access to this hub.
```
sgibson91, line 59: We could make a similar statement about the Pangeo project

docs/reference/hubs.md
```diff
@@ -26,4 +31,29 @@ $(document).ready( function () {
...
+
+We run a JupyterHub for the University of Toronto.
+This is configured and deployed at [the `utoronto-deploy` repository](https://github.com/utoronto-2i2c/jupyterhub-deploy).
+
+**Special cases**
+
+There are a few special cases about this deployment:
+
+- **Authentication**: Uses [Azure Active Directory](https://azure.microsoft.com/en-us/services/active-directory/).
+
```
damianavila, line 51: Whiteline to be removed?

damianavila:
As a general comment, it is not clear to me when we classify a hub as a special case... can we be more specific about that definition?

choldgraf:
Thanks both for your comments!

For @damianavila's general point above, that's a good question. Let me give a stab at first defining our "standard stack" and then maybe that'll make it easier to decide what is a special case.

Standard stack:

- Configuration and CI/CD: In `pilot-hubs/`
- Cloud project: Owned by 2i2c
- Environment: uses a base environment, or uses an image deployed with the repo2docker action
- Authentication: GitHub or Google OAuth
- Cloud data: something 2i2c doesn't need to know about, handle authorization credentials for, etc

Is that a reasonable start? Am I missing anything?

As we develop the infrastructure further, things that are currently "bespoke" can become "standardized". For example, right now the Pangeo hub + the Toronto hub both run on cloud projects that we don't control. Right now I'd consider that "non-standard" because we haven't yet figured out a pattern to sustainably / repeatably follow there.

For the hubs where we do non-standard things, we should:

1. Decide if the non-standard thing *should* be standardized, and use the hub as inspiration to drive development that makes this hapen.
2. If the thing *should not* be standardized, we should either move towards migrating them off of that pattern, or we should decide how to charge them extra $$$ to cover the extra labor we incur in doing this non-standard thing.

Does that make sense? 

So then for the Toronto hub, it is non-standard because:

- Vendor: U Toronto project, not ours
- Authentication: AzureAD, not Github/Google OAuth
- Environment: Custom Dockerfile embedded in their repo
- CI/CD: Dedicated repository using hubploy

For Pangeo, it is non-standard because:

- Vendor: Columbia project, not ours
- Authentication: Soon to be GitHub Teams, but @sgibson91 is doing awesome work to standardize the deployment of that in our hubs :-)


# Does this sound reasonable?

If so, maybe I can write this up in the same page above? What do you think?

sgibson91:
This makes broad sense to me and I can't think of any obvious points you've missed :)

damianavila:
Thanks for that summary/proposal @choldgraf!

So, given your proposal, how do you classify the AWS clusters supporting Carbonplan, Farallon, and Openscapes hubs?
Technically, all of them live in their own AWS accounts and we are been given permissions to deploy on them.
Would not be that something close to the Pangeo (except for the Github Teams, of course)?

I think, according to your guidelines above, the only special case is the Toronto one, do you agree?

choldgraf:
For sure - just wanted to get a gut check before updating the PR :-)

For your question, I think that I'd direct it back to the @2i2c-org/tech-team :-)  do we think that our process and infrastructure is in a state where we'd consider hubs like carbon plan / pangeo / etc to no-longer be special cases? E.g., if a new community came with similar needs, would we know what to do, how to do it, and in a sustainable fashion? To some degree, "special case" also means "there are likely many unknown unknowns which is why we're doing this to learn". Do we think we are there yet? (maybe something to discuss tomorrow?)

sgibson91:
In the case of Pangeo, it's only getting access to the project that requires a different process. All the other work I did for that cluster (e.g. private nodes) could be applied to a new hub from the pilot-hubs repo just by using a flag somewhere in the config.

damianavila:
In the Farallon and Openscapes, which are kops-based clusters, I think we have the pieces in a good shape.
For the Carbonplan cluster, which is EKS-based, we still need documentation about setting up the cluster.
I think I would probably consider it a special case because this is our first EKS deployment (originated from the fact kops was not behaving enough well for the carbonplan usage case), but that is debatable as well :wink: .

choldgraf:
Hey all - I think this PR is stale now, and I don't have the time to follow up to figure out what needs to be updated. We've changed a bunch of the cluster deployments since it was opened, and have also opened some issues specific to each hub (e.g. Toronto access) so it may not be as urgent as before. I'll close this but please somebody ping here if there's something we should follow up on

PR #720, opened 2021-09-29
Pin prometheus chart version to an older version

GeorgianaElena:
Fixes https://github.com/2i2c-org/pilot-hubs/issues/616

### DIFF
--- a/support/Chart.yaml
+++ b/support/Chart.yaml
@@ -7,7 +7,7 @@ dependencies:
   # Prometheus for collection of metrics.
   # https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus
   - name: prometheus
-    version: 14.1.1
+    version: 13.8.0
     repository: https://prometheus-community.github.io/helm-charts
 
   # Grafana for dashboarding of metrics.


GeorgianaElena:
I'm currently trying to deploy it locally with:

```
python3 deployer deploy-support 2i2c
```

to see if it works

GeorgianaElena:
Running `kubectl describe` on that pod showed this error:
```
Multi-Attach error for volume "pvc-dae5a17d-763f-4dd5-b73a-3006d436a7ed" Volume is already used by pod(s) support-p
rometheus-server-59b9b6f98c-9x462
```

So, I guess I need to manually delete the old pod?

yuvipanda:
@GeorgianaElena yep, you should manually delete it. Prometheus should have config set [like this](https://github.com/berkeley-dsep-infra/datahub/blob/5718a15c9340c4d10e2fece33df70dbddc307f0e/support/values.yaml#L87) to prevent this from happening. When the strategy is set to `Recreate`, the first pod is fully deleted (thus releasing the PVC) before the new pod comes up. The default is `RollingUpdate` which tries to bring up the new pod before old pod is deleted - causing this issue wrt PVCs. This only happens *sometimes* I think - I'm not so sure why, perhaps only if the new pod is in a different node?

GeorgianaElena:
Thanks @yuvipanda! The deployment went ok after deleting the pod :tada: 

However, it doesn't look like anything's changed in https://grafana.pilot.2i2c.cloud. Is there anything else I should do, or the dashboards should just have magically appeared if this `prometheus` pin would had fixed the issue?

yuvipanda:
@GeorgianaElena I think github.com/jupyterhub/jupyterhub-grafana need to be deployed there to see if it works. Can you give that a shot too?

GeorgianaElena:
> @GeorgianaElena I think github.com/jupyterhub/jupyterhub-grafana need to be deployed there to see if it works. Can you give that a shot too?

I just did and I think [they work](https://grafana.pilot.2i2c.cloud/d/8Dl-i3N7z/cluster-information?orgId=1&from=now-15m&to=now) :tada:
grafana

yuvipanda (approved):

yuvipanda:
Yay awesome!

We need to also automate deployment of rhe gradana dashboard too

damianavila (approved):

choldgraf:
yesssss dashboards!!! thanks @GeorgianaElena 

@yuvipanda I assume you mean that automation should be tackled in the future, not as part of this PR?

yuvipanda:
@choldgraf yep!

damianavila:
> I assume you mean that automation should be tackled in the future, not as part of this PR?

I quickly searched about grafana and prometheus open issues and there are several of them.
Maybe it is time to group those together under the topic "infra to get info about our infra" 😉 .
I know this could be view as something belonging to the "Managed JupyterHubs Infrastructure" column in the deliverables board, but maybe we need more granularity there?

choldgraf:
@damianavila I don't think that I follow. Do you mean have a high-level issue about grafana/prometheus specifically?

@yuvipanda what are the things that we need to do in order to automatically deploy Grafana for our hubs? Is that something we should track in a new issue?

yuvipanda:
@choldgraf i opened https://github.com/2i2c-org/pilot-hubs/issues/739 to track that

damianavila:
`reporting`? How that would work? It is not clear to me...

damianavila:
Thanks for the clarification... `reporting` is too wide and prone to be confusing, IMHO, but I guess it is OK since I can come up with a better word :wink: .

damianavila:
`infrastructure reporting`, I guess?

damianavila:
Yep, :100: to `monitoring`.

PR #723, opened 2021-09-29
Use latest pangeo docker image for staging

rabernat:
Cartopy is broken in the tag we are pointing to. Should be fixed in this one.

I hope this is the correct way to propose this update. 🙏

### DIFF
--- a/config/hubs/pangeo-hubs.cluster.yaml
+++ b/config/hubs/pangeo-hubs.cluster.yaml
@@ -77,7 +77,7 @@ hubs:
           singleuser:
             image:
               name: pangeo/pangeo-notebook
-              tag: bcfacc5
+              tag: e6128a7
             profileList:
               # The mem-guarantees are here so k8s doesn't schedule other pods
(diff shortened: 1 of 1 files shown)


sgibson91:
Thanks @rabernat! I'm happy to merge this, but for future reference it is possible to change the image using the configurator https://pilot.2i2c.org/en/latest/admin/howto/configurator.html The image set in the configurator will always override what is in the config, so we don't cause a bottle neck for you :)

sgibson91 (approved):

rabernat:
In other words, RTFM! 🙃  Thanks Sarah.

sgibson91:
Hub health check failed when I merged this but the helm upgrade succeeded so I expect that was just test flakiness. Do holler in `#hub-support` on Slack though if something doesn't seem right

rabernat:
So I screwed up and pointed to a tag that exists in GitHub but not docker! 🤦 

Let's just revert this and I will use the admin console.

PR #724, opened 2021-09-29
Add nox builds for our docs

choldgraf:
This adds a quick way to auto-generate the documentation via one command:

```
nox -s docs
```

or

```
nox -s docs_live
```

after a request from @damianavila :-)

the implementation is a little bit hacky because nox doesn't handle native conda environments, so I split our `docs/environment.yml` file into a dedicated `docs/requirements.txt` file that we now use on its own.


noxfile.py
```diff
@@ -0,0 +1,26 @@
+import nox
+
+nox.options.reuse_existing_virtualenvs = True
+
+BUILD_COMMAND = ["-b", "html", "docs", "docs/_build/html"]
+
+def install_deps(session):
+    # Manually installing this because conda is a bit wonky w/ nox
+    session.conda_install("--channel=conda-forge", "go-terraform-docs")
```
damianavila, line 9: I see `python=3.8` in the environment file, don't you need the same here?

choldgraf, line 9:
```suggestion
    session.conda_install("--channel=conda-forge", "go-terraform-docs", "python=3.8")
```

damianavila (approved):
LGTM now!

damianavila (approved):
LGTM 2 😉

choldgraf:
Just updated from master to fix a version conflict, will merge now, thanks!

PR #726, opened 2021-10-01
Recovering PR #706: Supporting native JupyterHub OAuthenticator alongside Auth0 for our hubs

sgibson91:
### Summary

This PR is a reconstruction of PR #706 which failed on merge due to JSON schema validation errors. In addition to reconstructing that, this PR also aims to address the JSON schema validation errors and extends the `validate()` function in `deployer` to also validate the secret config, if it exists.

fixes https://github.com/2i2c-org/pilot-hubs/issues/625

### Changes to `config/hubs/schema.yaml`

I have [borrowed a pattern from @damianavila](https://github.com/2i2c-org/pilot-hubs/blob/master/config/hubs/schema.yaml#L115-L121) to make the `auth0.connection` property conditional on `auth0.enabled` which is now a fully fledged property of `auth0`.

I have also set the default value of `auth0.enabled` to be `true` so that we don't have to go through every `*.cluster.yaml` file and add the `enabled` key for `auth0`.

### DIFF
--- a/config/hubs/schema.yaml
+++ b/config/hubs/schema.yaml
@@ -170,7 +170,7 @@ properties:
                 default: true
                 description: |
                   Whether or not to enable Auth0 authentication for this hub.
-                  This should almost always true. It should only be false if
+                  This should almost always be true. It should only be false if
                   authenticating against GitHub Orgs or Teams using the
                   JupyterHub OAuthenticator. Defaults to true.
               connection:
--- a/config/hubs/schema.yaml
+++ b/config/hubs/schema.yaml
@@ -161,7 +161,7 @@ properties:
             additionalProperties: true
             type: object
             description: |
-              All hubs use Auth0 for authentication, and we dynamically fetch the credentials
+              Most hubs use Auth0 for authentication, and we dynamically fetch the credentials
               needed for each hub - client_id, client_secret, callback_url - on deploy. This
               block contains configuration on how auth0 should be configured for this hub.
             properties:
@@ -176,7 +176,6 @@ properties:
                   Authentication method users of the hub can use to log in to the hub.
                   We support a subset of the [connectors](https://auth0.com/docs/identityproviders)
                   that auth0 supports
-                required: false
               application_name:
                 type: string
                 description: |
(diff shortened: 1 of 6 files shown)


sgibson91:
I have deployed this to Pangeo staging (partnered with #707) and it works! So marking this as ready for review :)

config/hubs/schema.yaml
```diff
@@ -161,10 +161,18 @@ properties:
...
+              Most hubs use Auth0 for authentication, and we dynamically fetch the credentials
               needed for each hub - client_id, client_secret, callback_url - on deploy. This
               block contains configuration on how auth0 should be configured for this hub.
             properties:
+              enabled:
+                type: boolean
+                default: true
+                description: |
+                  Whether or not to enable Auth0 authentication for this hub.
+                  This should almost always true. It should only be false if
```
sgibson91, line 173:
```suggestion
                  This should almost always be true. It should only be false if
```

choldgraf:
Just a quick question - is this basically just the same PR as before, but now the original bug that we tripped has been fixed? If so, and if you've already tested it out, I'd be +1 on merging unless you want fresh eyes on any new stuff in particular, since the last PR already had some approves

config/hubs/schema.yaml
```diff
@@ -198,8 +206,13 @@ properties:
                       "<clustername>-<hubname>".
 
                       See https://manage.auth0.com/dashboard/us/2i2c/connections/database.
-            required:
-              - connection
+            if:
```
yuvipanda, line 209: this is super cool, TIL! Thanks @damianavila!

deployer/hub.py
```diff
@@ -348,6 +348,7 @@ def get_generated_config(self, auth_provider: KeyProvider, secret_key):
             # FIXME: We're hardcoding Auth0OAuthenticator here
```
yuvipanda, line 348: Update the comment to match the code change?

yuvipanda (changes requested):
A code comment doesn't match what the code is doing now, so we should just update that? Otherwise this is good to me.

sgibson91:
I've updated the comment @yuvipanda pointed out, so I will merge once tests pass :)

config/hubs/schema.yaml
damianavila: Yeah... I really liked it when I discovered it 😉

PR #728, opened 2021-10-04
[BUGFIX] Create JupyterHub key in generated config

sgibson91:
Merging #726 failed in non-Auth0 cases due to a Python error where the `JupyterHub` key in the generated config didn't exist. This PR fixes that issue by tweaking the Python code so the key is created.

Also, setting `authenticator_class` to `CustomOAuthenticator` was the wrong move as the `extra_config` is loaded into the hub last and this caused a bad config error (see below comment). Reverting back to the Auth0 class resulted in a successful deploy to 2i2c-staging hub.

### DIFF
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -348,7 +348,7 @@ class Hub:
             # NOTE: Some dictionary merging might make these lines prettier/more readable.
             # Since Auth0 is enabled, we set the authenticator_class to the CustomOAuthenticator we define in the basehub chart
             # https://github.com/2i2c-org/pilot-hubs/blob/master/hub-templates/basehub/values.yaml#L303-L318
-            generated_config['jupyterhub']['hub']['config']['JupyterHub']['authenticator_class'] = 'CustomOAuthenticator'
+            generated_config['jupyterhub']['hub']['config']['JupyterHub'] = {'authenticator_class': 'CustomOAuthenticator'}
             generated_config['jupyterhub']['hub']['config']['Auth0OAuthenticator'] = auth_provider.get_client_creds(client, self.spec['auth0']['connection'])
 
         return self.apply_hub_template_fixes(generated_config, secret_key)
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -348,7 +348,7 @@ class Hub:
             # NOTE: Some dictionary merging might make these lines prettier/more readable.
             # Since Auth0 is enabled, we set the authenticator_class to the CustomOAuthenticator we define in the basehub chart
             # https://github.com/2i2c-org/pilot-hubs/blob/master/hub-templates/basehub/values.yaml#L303-L318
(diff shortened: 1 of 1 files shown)


sgibson91:
I tried deploying this fix on 2i2c-staging but the hub pod gets stuck with the following error

```
[C 2021-10-04 10:52:38.190 JupyterHub application:89] Bad config encountered during initialization: The 'authenticator_class' trait of <jupyterhub.app.JupyterHub object at 0x7f5ec96e0340> instance must be a type, but 'CustomOAuthenticator' could not be imported
```

I think this is related to what @GeorgianaElena said [here](https://github.com/2i2c-org/pilot-hubs/pull/706#discussion_r718279289) about extra config being the last thing to be loaded by the hub. Can anyone help?

sgibson91:
OK, setting `authenticator_class` back to `oauthenticator.auth0.Auth0OAuthenticator` worked on 2i2c-staging! I think this is good to go!

sgibson91:
I'm going to merge this without review since I got it working on 2i2c-staging and `master` is currently broken without this fix (merging this now saves reverting #726 and starting from scratch again)

damianavila:
OK, two things here:

1. the comment above the line changed should be updated otherwise it does not reflect the code change: https://github.com/2i2c-org/pilot-hubs/pull/728/files#diff-eedaf02b81cd907a3feb5e4389e9825226bf7dc82a0fb582f9ad367c00ba6651R349-R350
2. since we are no longer configuring the `CustomOAuthenticator` here, don't we need to get back these lines: https://github.com/2i2c-org/pilot-hubs/pull/726/files#diff-30738e7aecdcaa7cac0f7072ff874406795aca126a1ec89051237cb52829588fL320-L321

sgibson91:
> 2\. since we are no longer configuring the `CustomOAuthenticator` here, don't we need to get back these lines? [#726 (files)](https://github.com/2i2c-org/pilot-hubs/pull/726/files#diff-30738e7aecdcaa7cac0f7072ff874406795aca126a1ec89051237cb52829588fL320-L321)

All I know for certain is that including this line here will break for JupyterHub OAuthenticator cases. Since this is read **after** the main config, it will overwrite this line: https://github.com/2i2c-org/pilot-hubs/blob/41a797146de147bf43bcf647460545a7e98ca9ae/config/hubs/pangeo-hubs.cluster.yaml#L75-L76

I think you're right that we need to go back to `CustomOAuthenticator` but it's not a straight-forward reversion since we need to _dynamically_ set this. We somehow need an if-statement in that block. I wonder if something like this would work?

```
if c.JupyterHub.authenticator_class != 'github':
    c.JupyterHub.authenticator_class = CustomOAuthenticator
```

sgibson91:
Making a note of this config for myself in case it's needed: https://github.com/2i2c-org/pilot-hubs/pull/311/files#diff-30738e7aecdcaa7cac0f7072ff874406795aca126a1ec89051237cb52829588fR333-R335

PR #732, opened 2021-10-05
Dynamically set authenticator_class

sgibson91:
This resets the authenticator_class to `CustomOAuthenticator` if the `Auth0`
class has originally been instantiated. This allows us to continue
supporting ORCiD login.

This is a short-term fix that addresses concerns raised in https://github.com/2i2c-org/pilot-hubs/discussions/731

### DIFF
--- a/hub-templates/basehub/values.yaml
+++ b/hub-templates/basehub/values.yaml
@@ -298,6 +298,7 @@ jupyterhub:
         c.JupyterHub.spawner_class = CustomSpawner
 
       06-custom-authenticator: |
+        from z2jh import get_config
         from oauthenticator.auth0 import Auth0OAuthenticator
 
         class CustomOAuthenticator(Auth0OAuthenticator):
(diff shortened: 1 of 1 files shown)


GeorgianaElena (approved):
LGTM! :+1:

yuvipanda (changes requested):
My preference is to just get rid of the entire `06-custom-authenticator` config completely, and remove mention of ORCID support from our docs (https://github.com/2i2c-org/pilot-hubs/discussions/731#discussioncomment-1428432). We don't even have a clear way to test this properly as we don't have any hubs with ORCID running. This will simplify our auth config, and we can just deal with this if / when someone asks us for ORCID support.

damianavila:
I would agree with @yuvipanda here. If we are not using this actively and we can not test it, then let's simplify the stuff and re-introduce it (we now know how to deal with this :wink: ) when it is actually needed.

PR #734, opened 2021-10-05
Update auth docs

sgibson91:
This PR adds a small note to our auth docs not to mix `allowed_users` with `allowed_organizations` when authenticatings with the JupyterHub OAuthenticator.

This is because `allowed_users` takes precedence even if the user authenticating is a valid member of the specified org/team. (This caught me out in #707)


yuvipanda (approved):
Thanks, @sgibson91!

PR #747, opened 2021-10-08
Add workflow to automatically deploy grafana dashboards to our hubs

GeorgianaElena:
Fixes https://github.com/2i2c-org/pilot-hubs/issues/739

** Note **
1. The following clusters don't have the grafana support chart deployed:
- `farallon`
- `openscapes`
- `meom-ige`
2. The tokens created in grafana are for 1y

### DIFF
--- a/.github/workflows/deploy-grafana-dashboards.yaml
+++ b/.github/workflows/deploy-grafana-dashboards.yaml
@@ -23,6 +23,7 @@ jobs:
         run: |
           python3 -m pip install -r requirements.txt
           python3 -m pip install -r dev-requirements.txt
+          sudo apt install jsonnet
 
       - name: Deploy grafana dashboards for ${{ matrix.cluster_name }}
         run: |
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,4 +5,3 @@ ruamel.yaml
 auth0-python
 jhub-client==0.1.2
 jsonschema
-jsonnet


.github/workflows/deploy-grafana-dashboards.yaml
```diff
@@ -0,0 +1,39 @@
...
+    strategy:
+      # Don't stop other deployments if one fails
+      fail-fast: false
+      matrix:
+        include:
+          # To enable auto-deployments of grafana dashboards
+          # for other clusters add its name and, a GitHub
+          # secret for it and save its token to `grafana_token`
+          - cluster_grafana_name: grafana.pilot
+            grafana_token: secrets.GRAFANA_TOKEN
```
damianavila, line 17: I do not think we are using GitHub secrets on other workflows... maybe we want to keep that state and have all these secrets in some of the existing (or maybe a new) encrypted file?

yuvipanda: @sgibson91's work in https://github.com/2i2c-org/pilot-hubs/pull/726 means we already have a space for per-cluster sops encrypted secrets (secrets/config). Let's use that? But that means we'll need a python script or something that can decrypt the secrets and use them. But that's cleaner I think.

yuvipanda (changes requested):

yuvipanda:
I've removed all github secrets except the one we need to decrypt `sops`, and `GRAFANA_TOKEN` which was added as part of this PR I think. We should definitely use sops here!

damianavila:
> Let's use that? But that means we'll need a python script or something that can decrypt the secrets and use them. But that's cleaner I think.

I thought about that as well... in that case, you have per cluster secrets but now your grafana stuff is scattered...
I would be OK with either approach @yuvipanda... since you can not reconcile both things at the same time.

deployer/__main__.py
```diff
@@ -46,6 +46,43 @@ def deploy_support(cluster_name):
...
+    if cluster.support:
+        grafana_url = cluster.support.get("config", {}).get("grafana", {}).get("ingress", {}).get("hosts", {})
+        uses_tls = cluster.support.get("config", {}).get("grafana", {}).get("ingress", {}).get("tls", {})
+
+        if grafana_url:
+            grafana_url = "https://" + grafana_url[0] if uses_tls else "http://" + grafana_url[0]
+
+        with tempfile.TemporaryDirectory() as grafana_dashboards:
+            print("Cloning jupyterhub/grafana-dashboards.git...")
+            Repo.clone_from("https://github.com/jupyterhub/grafana-dashboards.git", grafana_dashboards)
```
yuvipanda, line 78: Can we just shell out to `git` instead? Keeps things simpler I think...

deployer/__main__.py
```diff
@@ -46,6 +46,47 @@ def deploy_support(cluster_name):
...
+        if grafana_url:
+            grafana_url = "https://" + grafana_url[0] if uses_tls else "http://" + grafana_url[0]
+
+            # Use the jupyterhub/grafana-dashboards deployer to deploy the dashboards to this cluster's grafana
+            with tempfile.TemporaryDirectory() as grafana_dashboards:
+                print("Cloning jupyterhub/grafana-dashboards...")
+                subprocess.check_call(["git", "clone", "https://github.com/jupyterhub/grafana-dashboards", grafana_dashboards])
+
+                print(f"Deploying grafana dashboards to {cluster_name}...")
+                subprocess.check_call([grafana_dashboards + "/deploy.py", grafana_url])
```
yuvipanda, line 85: let's pass the environment variable GRAFANA_TOKEN here specifically, instead of modifying our own environment variables?

deployer/__main__.py
```diff
@@ -46,6 +47,48 @@ def deploy_support(cluster_name):
...
+        uses_tls = cluster.support.get("config", {}).get("grafana", {}).get("ingress", {}).get("tls", {})
+        if grafana_url:
+            grafana_url = "https://" + grafana_url[0] if uses_tls else "http://" + grafana_url[0]
+
+            # Use the jupyterhub/grafana-dashboards deployer to deploy the dashboards to this cluster's grafana
+            with tempfile.TemporaryDirectory() as grafana_dashboards:
+                print("Cloning jupyterhub/grafana-dashboards...")
+                subprocess.check_call(
+                    ["git", "clone", "https://github.com/jupyterhub/grafana-dashboards", grafana_dashboards],
+                    env={'GRAFANA_TOKEN': config["grafana_token"]}
```
yuvipanda, line 83: The env var should be passed to the deploy.py call, rather than the git clone call

deployer/__main__.py
```diff
@@ -46,6 +47,50 @@ def deploy_support(cluster_name):
...
+
+        # Read and set GRAFANA_TOKEN from the cluster specific secret config file
+        with decrypt_file(secret_config_file) as decrypted_file_path:
+            with open(decrypted_file_path) as f:
+                config = yaml.load(f)
+
+        # Get the url where grafana is running from the cluster config
+        grafana_url = cluster.support.get("config", {}).get("grafana", {}).get("ingress", {}).get("hosts", {})
+        uses_tls = cluster.support.get("config", {}).get("grafana", {}).get("ingress", {}).get("tls", {})
+        if grafana_url:
```
damianavila, line 75: if support is deployed, grafana should be there, right?
Or, are we envisioning some cluster with support enabled but without grafana specified?

PR #748, opened 2021-10-13
Enable scratch buckets on Pangeo staging and prod hubs

sgibson91:
This adds missing config required to enable scratch buckets on the Pangeo hubs


damianavila (approved):
LGTM

PR #753, opened 2021-10-15
Update docs on migrating NFS home directories

sgibson91:
This PR clarifies the steps required to migrate user home directories for hubs. It also documents an `rsync` command as an alternative to `scp`, and explains the process for migrating GCP filestores.

### DIFF
--- a/docs/howto/operate/move-hub.md
+++ b/docs/howto/operate/move-hub.md
@@ -57,6 +57,34 @@ The trailing slashes are important to copy the contents of the directory, withou
 
 See the [`rsync` man page](https://ss64.com/bash/rsync.html) to understand these options.
 
+### GCP Filestores
+
+We also use GCP Filestores as in-cluster NFS storage and can transfer the home directories between them in a similar fashion to the NFS servers described above.
+
+The filestores must be mounted in order to be accessed.
+
+1. Create VMs in the projects of the source and target filestores.
+2. For **both** filestores, get the server address from the [GCP console](https://cloud.google.com/filestore/docs/mounting-fileshares).
+3. On each VM for the source and target filestores:
+   1. Install `nfs-common`:
+      ```bash
+      sudo apt-get -y update && sudo apt-get -y install nfs-common
+      ```
+   2. Create a mount point:
+      ```bash
+      sudo mkdir -p /mnt/filestore
+      ```
+   3. Mount the Filestore:
+      ```bash
+      sudo mount SERVER_ADDRESS /mnt/filestore
+      ```
+
+The user directories can then be transferred in the same manner as [NFS Servers](#nfs-servers) with the locations updated to be the following:
+
+```bash
+ubuntu@nfs-source-server-public-IP:/mnt/filestore/<hub-name> /mnt/filestore/<hub-name>
+```
+
 ### EFS
 
 [AWS DataSync](https://aws.amazon.com/datasync/)
--- a/docs/howto/operate/move-hub.md
+++ b/docs/howto/operate/move-hub.md
@@ -48,7 +48,11 @@ Primarily used with GKE right now.
 As an alternative to `scp` you can use `rsync` as follows:
 
 ```bash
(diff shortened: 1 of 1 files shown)


docs/howto/operate/move-hub.md
```diff
@@ -23,29 +23,67 @@ home directory storage!
 
 Primarily used with GKE right now.
 
-1. In the target NFS server, create a new ssh key-pair, with
+1. SSH into **both** the target server and source server and then enable the `ubuntu` user on **both** servers with `sudo su ubuntu`.
+   This will ensure that the keys we are about to create will be assigned to the correct user.
+2. In the target NFS server, create a new ssh key-pair, with
```
damianavila, line 28: Don't you need to set up some specific permissions after creating the key?

docs/howto/operate/move-hub.md
```diff
@@ -23,29 +23,67 @@ home directory storage!
...
 
    This makes sure the target is owned by the `ubuntu` user, which has
    uid `1000`. Our user pods run as uid `1000`, so this makes sure they
    can mount their home directories.
 
+As an alternative to `scp` you can use `rsync` as follows:
+
+```bash
+rsync -e 'ssh -i nfs-transfer-key' -rougvhP ubuntu@nfs-source-server-public-IP:/export/home-01/homes/<hub-name>/ /export/home-01/homes/<hub-name>/
+```
```
damianavila, line 52: Could it be the case the `-e 'ssh -i nfs-transfer-key'` is not actually needed if you ssh-add the key?

docs/howto/operate/move-hub.md
sgibson91: No, by making sure you activate the `ubuntu` user first and using the `ssh-keygen` command, all the permissions are taken care of.

docs/howto/operate/move-hub.md
sgibson91: I don't know, I'm not familiar enough with `ssh add`. What I found is that because I prefaced the command with `sudo`, `rsync` was looking at `root`'s ssh keys, so I was just specific in saying which key I wanted to use.

docs/howto/operate/move-hub.md
```diff
@@ -23,29 +23,67 @@ home directory storage!
 
 Primarily used with GKE right now.
 
-1. In the target NFS server, create a new ssh key-pair, with
+1. SSH into **both** the target server and source server and then enable the `ubuntu` user on **both** servers with `sudo su ubuntu`.
```
GeorgianaElena, line 26: Might be a stupid question, but do our clusters all have an `ubuntu` user on them?

docs/howto/operate/move-hub.md
```diff
@@ -23,29 +23,67 @@ home directory storage!
...
    `ssh-keygen -f nfs-transfer-key`
-
-2. Copy the public key `nfs-transfer-key.pub` to the source NFS
-   server's `/home/ubuntu/.ssh/authorized_keys`, so the target
-   NFS server can `scp` from the source NFS server.
-
-3. Copy the NFS home directories from the source NFS server to
+3. Append the public key `nfs-transfer-key.pub` to the source NFS
+   server's `/home/ubuntu/.ssh/authorized_keys` file, so the target
+   NFS server can open SSH connections to the source NFS server.
```
GeorgianaElena, line 32: What do you think about splitting the sentence? (I read it 2 times until I made a visual model and was sure I understood :sweat_smile: )
```suggestion
3. Append the public key `nfs-transfer-key.pub` to the source NFS
   server's `/home/ubuntu/.ssh/authorized_keys` file. This way, the target
   NFS server will be able to open SSH connections to the source NFS server.
```

GeorgianaElena (approved):
LGTM! :rocket:  Thanks @sgibson91!

docs/howto/operate/move-hub.md
sgibson91: It's not our clusters but a VM that is either acting as an NFS server or a VM that has some kind of storage mounted (GCP Filestore, Azure Filestore, etc).

I _think_ most linux distros ship with an `ubuntu` user by default - I didn't have to `sudo useradd ubuntu` - but I'm not 100% sure.

PR #756, opened 2021-10-18
Remove outdated references to PROXY_SECRET_TOKEN

yuvipanda:
Unused since d6fc272

Ref https://github.com/2i2c-org/pilot-hubs/issues/755


yuvipanda:
@consideRatio we don't actually specify proxy.secretToken anymore, it's automatically generated. This is primarily now used to deterministically generate the service token for the hub service check (and other services). Hence it was renamed.

yuvipanda:
The code has fundamentally changed enough now that this PR is no longer helpful. Going to close this :)

PR #758, opened 2021-10-18
Rename pilot hubs to infrastructure

choldgraf:
This renames references of `pilot hubs` to `infrastructure` or `community hubs` when we refer to specific hubs.

### DIFF
--- a/docs/structure.md
+++ b/docs/structure.md
@@ -79,7 +79,7 @@ repository, available at [2i2c.org/pilot](https://2i2c.org/pilot/)
 
 These are folks tasked with building, maintaining, debugging and fixing
 2i2c infrastructure. Documentation targetting them should be in
-[2i2c-org/infrastructure](https://github.com/2i2c-org/infrastructure).
+[2i2c-org/infrastructure](https://github.com/2i2c-org/infrastructure)
 repository (2i2c-org/infrastructure), regardless of the kind of hub it 
 relates to. Since we run our hubs openly, with best practices that can
 be adopted by anyone, we should try write these to be as accessible to
--- a/docs/topic/terraform.md
+++ b/docs/topic/terraform.md
@@ -47,7 +47,12 @@ If you want to switch to a different terraform workspace that is stored in the s
 terraform workspace switch WORKSPACE_NAME
 ```
 
-For example, if you were working in the `infrastructure` workspace but want to switch to `justiceinnovationlab`, these are both stored in the same backend and so the command looks like:
+
+:::{note}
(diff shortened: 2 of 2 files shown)


damianavila:
General comment, you are making some changes in our templates to fix comment links...
I would be hesitant to do that because we are going to trigger unneeded deployments just to fix a comment link.
I would prefer to change those at the time we implement change in the terraform files, IMHO.

choldgraf:
Resolved conflicts! I wonder if @damianavila would be willing to take one last look to make sure I'm not overwriting something important?

docs/structure.md
```diff
@@ -79,8 +79,8 @@ repository, available at [2i2c.org/pilot](https://2i2c.org/pilot/)
 
 These are folks tasked with building, maintaining, debugging and fixing
 2i2c infrastructure. Documentation targetting them should be in
-[2i2c-org/pilot-hubs](https://github.com/2i2c-org/pilot-hubs).
-repository (2i2c-org/pilot-hubs), regardless of the kind of hub it 
+[2i2c-org/infrastructure](https://github.com/2i2c-org/infrastructure).
+repository (2i2c-org/infrastructure), regardless of the kind of hub it
```
sgibson91, line 83:
```suggestion
[2i2c-org/infrastructure](https://github.com/2i2c-org/infrastructure)
repository (2i2c-org/infrastructure), regardless of the kind of hub it 
```

sgibson91:
For future reference, the path list the would cause new deployments can be found here:

https://github.com/2i2c-org/infrastructure/blob/be6c30f5ce31f49081c6426fdcbac712137110f5/.github/workflows/deploy-hubs.yaml#L7-L15

I left a couple of small comments, otherwise it looks good.

docs/topic/terraform.md
```diff
@@ -71,7 +71,7 @@ terraform init -backend-config=backends/<REQUIRED_CONFIG>.hcl -reconfigure
 terraform workspace select WORKSPACE_NAME
 ```
 
-For example, if you were working on our `pilot-hubs`, with our default backend initialised, but wanted to switch to working on our Pangeo deployments, the commands would look as follows:
+For example, if you were working on our `infrastructure`, with our default backend initialised, but wanted to switch to working on our Pangeo deployments, the commands would look as follows:
```
choldgraf, line 74:
```suggestion
For example, if you were working on our `pilot-hubs`, with our default backend initialised, but wanted to switch to working on our Pangeo deployments, the commands would look as follows:
```

damianavila:
> Resolved conflicts! I wonder if @damianavila would be willing to take one last look to make sure I'm not overwriting something important?

I was willing but people have beaten me at it 😜  .

PR #763, opened 2021-10-19
Fix mills deploy. Another #311 follow-up

GeorgianaElena:
Add the 2i2c addrs to the username pattern. Should find a better way to do this though


damianavila:
> The admin users must respect the username_pattern too, this is why they were added to the regex (this was also mentioned in jupyterhub/jupyterhub#3525)

This is indeed something I would not expect because, IMHO, the admin state should have a higher heararchy.
We should try to fix this upstream as indicated/desribed in that issue.

Issue #790, opened 2021-10-28
[Incident] PaleoHack User sessions won't start


choldgraf:
# Summary

The PaleoHackWeek hub is not putting users on the new nodes that were pre-cooked for the event. Here are the hub logs:

```
2021-10-28T17:38:46Z [Warning] 0/23 nodes are available: 20 Insufficient cpu, 3 node(s) didn't match node selector.
2021-10-28T17:38:48Z [Normal] pod didn't trigger scale-up: 3 node(s) didn't match node selector, 1 max node group size reached
```

In addition, I'm not sure if this is relevant but it seems like all of the CPU commit %s are pinned at 62.6% and I'm not sure why. I wonder if that's related to the `20 Insufficient cpu` error above. Here's an image of the grafana plot:

![image](https://user-images.githubusercontent.com/1839645/139307444-2831399b-8409-450b-8dc2-52cc3a8d59a6.png)

## Timeline (if relevant)

See the comments below for the timeline details, all happened within a few hours. Here is a summary:

- [Initial report](https://github.com/2i2c-org/infrastructure/issues/790#issuecomment-954063415)
- [Identified that the issue was one user per node](https://github.com/2i2c-org/infrastructure/issues/790#issuecomment-954068393)
- [Realized that we did not add the proper helm config value to fix this](https://github.com/2i2c-org/infrastructure/issues/790#issuecomment-954081154)
- [We realized that the paleohack image didn't have nbgitpuller](https://github.com/2i2c-org/infrastructure/issues/790#issuecomment-954094014) and updated the environment there
- [We realized that the configurator had its own Docker image configured that was clashing with our updates](https://github.com/2i2c-org/infrastructure/issues/790#issuecomment-954110457)
- [Confirmed the incident was resolved](https://github.com/2i2c-org/infrastructure/issues/790#issuecomment-954120784)

---

# After-action report

## What went wrong

Two major things:

- We had a sub-optimal helm config that was causing Kubernetes to request *all* of the resources available to a pod.
- The PaleoHack image didn't have nbgitpuller on it, so their nbgitpuller links didn't work

## Action items

These are only sample subheadings. Every action item should have a GitHub issue
(even a small skeleton of one) attached to it, so these do not get forgotten. These issues don't have to be in `infrastructure/`, they can be in other repositories.

### Process improvements

1. Add steps to send to the community representative to make sure the hub is working properly before the event (#  ) 

### Documentation improvements

[...]

PR #801, opened 2021-11-01
Set 0.05 CPU guarantee for all user pods

yuvipanda:
If no CPU limit is set, it is possible for a single user or group of users to
starve everyone else of CPU time on a node, even causing new user pods to completely
fail as the notebook server process gets no CPU to complete auth handshake with
the server, and even trivial cells like `print("hello world")` may not run.
Unlike memory guarantees, CPU guarantees are actually enforced by the Linux Kernel
(see https://medium.com/@betz.mark/understanding-resource-limits-in-kubernetes-cpu-time-9eff74d3161b)
By giving each user a 5% CPU guarantee (represented by 0.05), we ensure that:
1. Simple cells will always execute
2. Notebook server processes will always start - so users won't have server spawn failure
3. We don't accidentally set just a high limit for a particular hub and not set a
   guarantee, at which point kubernetes treats the limit as the guarantee! This causes
   far more nodes to be scaled up than needed, making everything super slow (like in
   https://github.com/2i2c-org/infrastructure/issues/790)
4. Most of our workloads are still memory bound, and we want scaling to happen only
   when a node is full on its memory guarantees. But a 0.05 guarantee means a n1-highmem-8
   node can fit 160 user pods, and since kubernetes already caps us at 100 pods a node,
   this guarantee doesn't actually change our scheduling.

Ref https://github.com/2i2c-org/infrastructure/issues/790

### DIFF
--- a/hub-templates/basehub/values.yaml
+++ b/hub-templates/basehub/values.yaml
@@ -150,6 +150,14 @@ jupyterhub:
     memory:
       guarantee: 256M
       limit: 1G
+    cpu:
+      # When limit is set but not guarantee, guarantee is set to match limit! For most
+      # use cases, when we set limit but not guarantee, we want to offer no guarantee and
+      # a limit. This doesn't seem to be possible at all (need to investigate why). In the
+      # meantime, setting a super low guarantee here means we can guard against issues like
+      # https://github.com/2i2c-org/infrastructure/issues/790, where setting the limit
+      # but not guarantee just gave users a huge guarantee, causing node spin ups to fail
+      guarantee: 0.01
     networkPolicy:
       # Allow unrestricted access to the internet but not local cluster network
       enabled: true
--- a/hub-templates/basehub/values.yaml
+++ b/hub-templates/basehub/values.yaml
@@ -151,13 +151,24 @@ jupyterhub:
       guarantee: 256M
       limit: 1G
     cpu:
-      # When limit is set but not guarantee, guarantee is set to match limit! For most
-      # use cases, when we set limit but not guarantee, we want to offer no guarantee and
-      # a limit. This doesn't seem to be possible at all (need to investigate why). In the
-      # meantime, setting a super low guarantee here means we can guard against issues like
-      # https://github.com/2i2c-org/infrastructure/issues/790, where setting the limit
-      # but not guarantee just gave users a huge guarantee, causing node spin ups to fail
-      guarantee: 0.01
+      # If no CPU limit is set, it is possible for a single user or group of users to
+      # starve everyone else of CPU time on a node, even causing new user pods to completely
+      # fail as the notebook server process gets no CPU to complete auth handshake with
+      # the server, and even trivial cells like `print("hello world")` may not run.
+      # Unlike memory guarantees, CPU guarantees are actually enforced by the Linux Kernel
+      # (see https://medium.com/@betz.mark/understanding-resource-limits-in-kubernetes-cpu-time-9eff74d3161b)
+      # By giving each user a 5% CPU guarantee (represented by 0.05), we ensure that:
+      # 1. Simple cells will always execute
+      # 2. Notebook server processes will always start - so users won't have server spawn failure
+      # 3. We don't accidentally set just a high limit for a particular hub and not set a
+      #    guarantee, at which point kubernetes treats the limit as the guarantee! This causes
+      #    far more nodes to be scaled up than needed, making everything super slow (like in
+      #    https://github.com/2i2c-org/infrastructure/issues/790)
+      # 4. Most of our workloads are still memory bound, and we want scaling to happen only
+      #    when a node is full on its memory guarantees. But a 0.05 guarantee means a n1-highmem-8
+      #    node can fit 160 user pods, and since kubernetes already caps us at 100 pods a node,
+      #    this guarantee doesn't actually change our scheduling.
+      guarantee: 0.05
     networkPolicy:
       # Allow unrestricted access to the internet but not local cluster network
       enabled: true


yuvipanda:
@consideRatio how would kubespawner currently represent 'no guarantee but a limit'? Is that even possible?

consideRatio:
From the k8s docs at https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits

> Note: If a Container specifies its own memory limit, but does not specify a memory request, Kubernetes automatically assigns a memory request that matches the limit. Similarly, if a Container specifies its own CPU limit, but does not specify a CPU request, Kubernetes automatically assigns a CPU request that matches the limit.

I assume kubespawner would represent no cpu/memory guarantee as an unset resource requests, which then makes the k8s api-server automatically set the resource requests to the explicit limit as documented.

hub-templates/basehub/values.yaml
```diff
@@ -150,6 +150,14 @@ jupyterhub:
...
       guarantee: 256M
       limit: 1G
+    cpu:
+      # When limit is set but not guarantee, guarantee is set to match limit! For most
+      # use cases, when we set limit but not guarantee, we want to offer no guarantee and
+      # a limit. This doesn't seem to be possible at all (need to investigate why). In the
+      # meantime, setting a super low guarantee here means we can guard against issues like
+      # https://github.com/2i2c-org/infrastructure/issues/790, where setting the limit
+      # but not guarantee just gave users a huge guarantee, causing node spin ups to fail
+      guarantee: 0.01
```
consideRatio, line 160: This can make sense in a way, but it also forces everyone to learn about 2i2c specific configuration.

Its a trade off between that users can assume k8s native behavior vs that they can't and potentially avoids mistakes due to it.

I'm conflicted about setting various defaults in the base chart overall due to this, but since we already have defaults about resource requests/limits I don't see it as an escalation of such complexity.

yuvipanda:
@consideRatio right - so this means that 'limit but no guarantee' is unrepresentable in kubernetes?

hub-templates/basehub/values.yaml
yuvipanda: > I'm conflicted about setting various defaults in the base chart overall due to this, but since we already have defaults about resource requests/limits I don't see it as an escalation of such complexity.

+1. Specifically, we already set default memory requests and limits here.

consideRatio:
> so this means that 'limit but no guarantee' is unrepresentable in kubernetes?

You can set resource requests to zero explicitly to represent 'limit but no guarantee', but having them unset makes it imply something else than 'no guarantee'.

damianavila:
Since we are here... do we want to also have a `limit` on the cpu?

choldgraf:
@damianavila I believe that this PR *is* setting a guarantee on the CPU - it already exists for RAM I believe.

also it feels like this is something that any community would run into themselves if they ran their own Z2JH setup, so to me this feels non 2i2c-specific

consideRatio:
@damianavila for very efficient CPU utilization and avoiding starvation of CPU resources, I think a good strategy is to set CPU guarantee + but use unbounded CPU limits.

By using a cpu limit, you make pods constrainted to use CPU without accomplishing anything I'd say - _unless_ - there was a container that got _CPU starved_. A pod/container can be starved from cpu resources if it is guaranteed/requested something very very small (minimum of 0.002 I think - you always get this minimum at least), and there was another container that went full throttle and had a request/guarantee of a lot more, like 2 - then they would share the CPU capacity based on the difference in request. So, in the 2 vs 0.002 example, on a 4 core machine would mean the pod with 2 in CPU requests/guarantee would make use of 3.992 of the CPU, and the other would be using 0.008 of the CPU which could have been to little to function.

damianavila:
So, given that all the user pods have the same request/guarantee CPU value, you are not expecting a starvation scenario like the one you described above, and then, there are no additional benefits for using a CPU limit, am I correct?

sgibson91:
I see there's some discussion in this PR but I'm not quite sure how/if it's blocking an approval?

consideRatio:
I'm conflicted about this kind of change.

It removes some friction, but adds some friction.
- It makes users need to understand how basehub chart works (what the default values are mainly) to understand how their configuration works
- It avoids situations where you set a limit and expected not to have to set the guarantee

The gist is that its complicates how things work, but at the same time it can reduce the mistakes made _in this chart_ specifically. It can at the same time enforce a conception that it is acceptable to only set limit in other charts, and then one may feel the need to add this kind of default to other charts resource requests as well.

Thinking about it, I suggest this change isn't accepted, but that instead whenever the resources are specified for cpu or memory respectively, that they are specified in one place fully - instead of having them specified in parts in multiple places that merge together into some final setting. I find that to add complexity to avoid having an issue that we will end up having no matter what unless we understand this k8s quirk:

> From the k8s docs at https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits
> 
> > Note: If a Container specifies its own memory limit, but does not specify a memory request, Kubernetes automatically assigns a memory request that matches the limit. Similarly, if a Container specifies its own CPU limit, but does not specify a CPU request, Kubernetes automatically assigns a CPU request that matches the limit.

damianavila:
> I think the best bet we have is to embrace the complexity without trying to mitigate it. Practically, I suggest we don't set either guarantee or limits without setting both whenever we set it for either cpu or memory in a chart - unless it's the final user config for some chart.

If we are going with this approach, we need to have a unified direction... I guess it should be all or nothing... closing this PR without further modifications leaves us in the middle of something where we have limits/guarantees for RAM but not CPU.

In addition, if we are going with the above approach, we need to make sure we are setting up limits and guarantee on all the final/end hubs config and, maybe, establish some "checks" so we are sure we do not miss adding those values, right?

consideRatio:
I think the practice is worth thinking about, but I don't feel confident to champion any one specific choice because there are trade offs.

I don't want to hold back on taking any kind of action, or invest my own time to think much more about it since I think everything is acceptable even though I have a preference towards either closing options.
- merging this
- closing this and doing nothing
- closing this and systematically making basehub not request anything and require the end configuration set it all

yuvipanda:
I think our alternatives are:

1. Ensure somehow that each hub in our config has explicit CPU and memory set, and validate this on CI
2. Merge this PR as is, and open an issue for (1)

My current preference is for (2), mostly because I think the advantage we get out of it is worth that. (1) is more perfect, but I don't want that to block this good enough.

sgibson91:
@yuvipanda I agree with your assessment. Let's merge this and open a new issue to iterate on implementation improvements!

sgibson91 (approved):

damianavila (approved):
Yep, let's merge this one as is and open a follow-up issue for option 1.

yuvipanda:
So I ended up encountering a related problem on the Berkeley DataHub a few days ago (https://github.com/berkeley-dsep-infra/datahub/issues/2966), where a few users running expensive simulations meant basically the entire hub was 'down' for most people - even `print("hello world")` won't work. In investigating and fixing that, I've actually changed my approach to this PR and have increased the CPU guarantee as well. I've laid out my new rationale in the PR description now, although I believe this is something that should actually make it into the z2jh docs...

damianavila (approved):
The new rationale is sensible to me so re-approving this one.

PR #804, opened 2021-11-04
Reduce guarantee memory to avoid going beyond the allocatable value

damianavila:
Each node has some allocatable resources [1]
If you try to schedule a pod with guarantee requirements above the
allocatable values, it will fail to be spawned.
With the "very large" option, we are requesting 110G at a minimum and that
is dangerously close to the theoretical allocatable memory for the
n1-standard-32 node. Then, let's give that guarantee value some
breath ;-)

[1] https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-architecture#memory_cpu

More details about the whole debugging lives on the corresponding freshdesk-support thread

### DIFF
--- a/config/hubs/meom-ige.cluster.yaml
+++ b/config/hubs/meom-ige.cluster.yaml
@@ -75,7 +75,7 @@ hubs:
                 description: "~32 CPU, ~128G RAM"
                 kubespawner_override:
                   mem_limit: 128G
-                  mem_guarantee: 110G
+                  mem_guarantee: 100G
                   node_selector:
                     node.kubernetes.io/instance-type: n1-standard-32
               - display_name: "Huge"


damianavila:
I have manually deployed this one to meom staging and it seems to work as intended!

config/hubs/meom-ige.cluster.yaml
```diff
@@ -75,7 +75,7 @@ hubs:
                 description: "~32 CPU, ~128G RAM"
```
choldgraf, line 75:
```suggestion
                description: "~32 CPU, ~100G RAM"
```

config/hubs/meom-ige.cluster.yaml
```diff
@@ -75,7 +75,7 @@ hubs:
...
                 kubespawner_override:
                   mem_limit: 128G
-                  mem_guarantee: 110G
+                  mem_guarantee: 100G
                   node_selector:
                     node.kubernetes.io/instance-type: n1-standard-32
               - display_name: "Huge"
```
choldgraf, line 81: is there something similar we should do for "huge"?

choldgraf (commented):
hmmm - I gave this a shot and ran into the `pod didn't trigger scale-up` issue :-/  

also a couple quick comments!

damianavila:
> hmmm - I gave this a shot and ran into the pod didn't trigger scale-up issue :-/

Did you try with staging, right? https://staging.meom-ige.2i2c.cloud/

config/hubs/meom-ige.cluster.yaml
damianavila: Fetching info from the nodes:

```
small >> 5758056Ki >> 5.75 GB allocatable (mem_guarantee: 5GB)
medium  >> 27155328Ki >> 27.15 GB allocatable (mem_guarantee: 25GB)
large >> 56183024Ki >> 56.18 GB allocatable (mem_guarantee: 50GB)
very large >> 114336712Ki >> 114.33G allocatable (mem_guarantee: 100GB with this PR)
huge >> 235367296Ki >> 235.37GB allocatable (mem_guarantee: 220G)
```

I think we are OK with the huge one but happy to discuss all the other values 😉  .

config/hubs/meom-ige.cluster.yaml
damianavila: Well, if we modify the description, it should be ~110G.

choldgraf:
Confirmed that this does work with staging!

config/hubs/meom-ige.cluster.yaml
consideRatio: I think the description should be what the node has capacity for, not what we guarantee.

The point of the guarantee is set to make sure we don't end up with multiple user pods on that node as I think of it. So if its ~128 or ~110 I don't know, but i recall 3.5G per CPU, so 112, but with a bit of memory less for system things, so ~110 seems about right for a n1 node.

consideRatio:
```suggestion
                description: "~32 CPU, ~110G RAM"
```

consideRatio (approved):
+1 For updating the description of in spawn options to say ~110 GB of memory, but this LGTM and can be merged.

damianavila:
I will merge it as is because this is what I validated on `staging` and open a follow-up issue to correct all the values in the profile descriptions (that actually needs better calculations first).

damianavila:
Follow-up: https://github.com/2i2c-org/infrastructure/issues/809

Issue #799, opened 2021-10-29
Migrate Pangeo prod's URL


sgibson91:
### Summary

To complete the migration of the hub, Pangeo prod's URL needs to be redirected to the `pangeo.io` address from `2i2c.cloud`.

### Tasks to Complete

- [x] Create a CNAME record in `pangeo.io` that points to `pangeo.2i2c.cloud`
- [x] Update GitHub OAuth app to use `pangeo.io` URL
- [ ] https://github.com/2i2c-org/infrastructure/pull/812

PR #812, opened 2021-11-08
Update pangeo prod hub's URL to pangeo.io domain

sgibson91:
fixes https://github.com/2i2c-org/infrastructure/issues/799

Updates Pangeo prod hub's domain to pangeo.io instead of pangeo.2i2c.cloud


rabernat (approved):
Awesome! I think we are ready!

PR #813, opened 2021-11-08
update Pangeo cluster funding source

rabernat:
We are now on NSF funding.


choldgraf:
thanks for updating this 👍

Issue #815, opened 2021-11-08
[Solved] Pangeo Hub has several critical issues


choldgraf:
# Summary

There are a variety of critical issues that have been reported on the Pangeo JupyterHub.

### Certificate errors

Some users reported a certificate error when connecting to the hub. Here's an example of the error message:

![image](https://user-images.githubusercontent.com/1839645/140835807-995a8b51-5641-47c0-a7cb-2de18bc9a165.png)

FreshDesk tickets:

- https://2i2c.freshdesk.com/a/tickets/41
- https://2i2c.freshdesk.com/a/tickets/42

### JupyterLab and kernal usability errors

Many users on the Pangeo hub are reporting slugging behavior. In particular, the following two actions take upwards of 30 seconds to complete, or do not complete at all:

- Starting Kernels
- Opening Terminals

FreshDesk tickets:

- https://2i2c.freshdesk.com/a/tickets/43

### Grafana is not reachable

I tried to go to grafana.us-central1-b.gcp.pangeo.io but received a "your connection is not private" error. So I haven't been able to look at any dashboards to understand what might be going on.

---

# After-action report

## What went wrong

We setup a redirect between two URLs where one was a CNAME for the other. This turned out to be a Very Bad Idea ™️ . In PR  https://github.com/2i2c-org/infrastructure/pull/812, we **_replaced_** the `pangeo.2i2c.cloud` address with `us-central1-b.gcp.pangeo.io` meaning that cert-manager was no longer issuing certificates for `pangeo.2i2c.cloud` and our load balancer would no longer accept traffic from `pangeo.2i2c.cloud`. All issues were resolved by undoing the redirect and visiting the `us-central1-b.gcp.pangeo.io` address instead.

Pangeo has been special-cased in that it has had active users **before** setup development was complete and I think the switch in URL is what confused people. Normally we would only invite users **after** the DNS has been set and so I don't see the issue arising again.

### Documentation improvements

1. Pull grafana links from hub config files and add them to documentation sites so it's clear which grafana URL goes with which hub: https://github.com/2i2c-org/infrastructure/pull/817

# Actions

- [x] Incident has been dealt with or is over
- [x] Sections above are filled out
- [x] Incident title and after-action report is cleaned up
- [x] All actionable items above have linked GitHub Issues

sgibson91:
> ### Grafana is not reachable
[...]

PR #817, opened 2021-11-09
Add grafana links to hubs list

choldgraf:
The grafana dashboards for our clusters are configured in the cluster-specific files in this repository. Since those exist in a structured space, this PR grabs the grafana configuration for each cluster and generates a grafana link per hub. That way we can more easily reference the grafana URL for a given hub, as this may be a common action and this can help reduce confusion.

you can preview how this looks here: https://2i2c-pilot-hubs--817.org.readthedocs.build/en/817/reference/hubs.html

This is a follow-up to #815.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -91,6 +91,13 @@ def render_hubs():
         yaml = cluster_info.read_text()
         cluster = safe_load(yaml)
 
+        # Pull support chart information to populate fields (if it exists)
+        support = cluster.get("support", {}).get("config", {})
+        grafana_url = support.get("grafana", {}).get("ingress", {}).get("hosts", "")
+        if isinstance(grafana_url, list):
+            grafana_url = grafana_url[0]
+            grafana_url = f"[{grafana_url}](http://{grafana_url})"
+
         # For each hub in cluster, grab its metadata and add it to the list
         for hub in cluster['hubs']:
             config = hub['config']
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -91,6 +91,13 @@ def render_hubs():
         yaml = cluster_info.read_text()
         cluster = safe_load(yaml)
 
+        # Pull support chart information to populate fields (if it exists)
+        support = cluster.get("support", {}).get("config", {})
+        grafana_url = support.get("grafana", {}).get("ingress", {}).get("hosts", "")
+        if isinstance(grafana_url, list):
```
damianavila, line 97: Are we really expecting a list of hosts urls? And if that is the case, why do we choose just the first one?

sgibson91: Yeah, the `grafana` chart expects `ingress.hosts` to be a list:

https://github.com/grafana/helm-charts/blob/a0b7d4d2222976bd11af38b2fec3f654a764030a/charts/grafana/values.yaml#L180-L196

I think if we are following the same policy for grafana that we do for hubs, that we don't support multiple domains, then there *should* only be one entry here. Maybe that is something we can enforce in the schema?

choldgraf: yeah I was just taking the first item because it always seemed to be a list in the config. Is this something that should block this PR? I think building the docs will fail if this condition is ever not met so we'd catch it thne

GeorgianaElena (approved):
:rocket:

choldgraf:
I am going to merge this one in because I would personally find it very useful to be able to quickly reference the grafana URLs for hubs (and make it clear when we *don't* have a grafana set up for a hub...but happy to iterate on questions about the right configuration structure etc in follow-up issues!

docs/conf.py
damianavila: Should we capture @sgibson91's suggestion in a new issue?

choldgraf: +1 from me

Issue #315, opened 2021-03-22
Enable authentication via CILogon


yuvipanda:
### Problem statement

Right now we only offer two ways to authenticate for our hubs: GitHub and Google OAuth. However, this falls short in a few ways:

- Many organization-wide hubs will want to authenticate by a given domain (e.g., `@berkeley.edu`) rather than with an individual list of users
- For organizations that don't with to use Google OAuth or GitHub, we do not have a means of authenticating their users.

### Solution

[CILogon](https://www.cilogon.org/home) is an authentication system focused on north american & western european academia. It is the authentication mechanism behind the "eduroam" network and a few others ([some more info is here](https://www.cilogon.org/faq)).

It is free for 'academic users' (https://www.cilogon.org/subscribe), and I would like to think we qualify.

If we support CILogon as a login provider, it lets us provide direct institutional logins to a large swath of potential users. This is probably particularly attractive to communities that are specific to an organization / institution, because it allows them to use the same login credentials that they are already using in their institution.

### Appetite

1 sprint

If we are integrating this with Auth0, it hopefully shouldn't be too much work to get running. It might take a bit of time to get the authentication approval from the CILogon folks, however.

### Guidance / implementation

CILogin has [specific instructions](https://www.cilogon.org/auth0) on integrating with Auth0, and that is what we should do! 

#### 2022-01-26

We'll aim to get the following done by the next sprint:

- [x] Get a valid approved client registered with CILogon (https://cilogon.org/oauth2/register)
- [x] Integrate that with [our Auth0 setup](https://www.cilogon.org/auth0)
- [x] Make it possible to set up CILogon for our hubs: #823
- [x] Choose a hub to use as a "demo" or "sandbox" hub for use by 2i2c team
- [x] Test out CILogon with this hub (maybe a demo hub). Two use-cases to consider:
  - Allow all users that match `@place.edu` pattern
  - Allow a list of users explicitly
  - Allow for a combination of `@place.edu` users *AND* 2i2c admin users
- [x] Use this experience to write up documentation about how to do this.
- [x] Investigate a deeper integration with CILogon to use for other authentication patterns as well :arrow_right: https://github.com/2i2c-org/infrastructure/issues/967
  - Look into their [service offerings](https://www.cilogon.org/subscribe)
  - Look into the [CILogon Authenticator for JupyterHub](https://github.com/jupyterhub/oauthenticator/blob/main/oauthenticator/cilogon.py)
  - Questions to answer:
[...]

PR #823, opened 2021-11-10
Add cilogon credentials and support

GeorgianaElena:
Reference: https://github.com/2i2c-org/infrastructure/issues/315

### DIFF
--- a/config/hubs/schema.yaml
+++ b/config/hubs/schema.yaml
@@ -179,6 +179,7 @@ properties:
                   - google-oauth2
                   - github
                   - password
+                  - CILogon
                 description: |
                   Authentication method users of the hub can use to log in to the hub.
                   We support a subset of the [connectors](https://auth0.com/docs/identityproviders)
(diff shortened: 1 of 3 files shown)


GeorgianaElena:
Friendly ping to @2i2c-org/tech-team :arrow_right:  Do you see any reason for not merging this (given the context in https://github.com/2i2c-org/infrastructure/issues/315#issuecomment-975887945)? (Other than there are no docs around this yet,  but promise to add them once this gets approved)

choldgraf:
Here's the comment where we decided which one to implement: https://github.com/2i2c-org/infrastructure/issues/315#issuecomment-982960242

It looks like the main thing that this does is adds our CILogon credentials to the deployment configuration etc. So is this correct:

1. We have set up an application on CILogon, and this gave us some kind of credential which we're adding in this PR.
2. If a community wants to use CILogon, in their hub configuration we'll need to configure the regex to support it. Here's an example of the configuration we would need:
  - https://github.com/2i2c-org/infrastructure/blob/d3da0f3e722bec6b3a5bc72a5614d0c7f6bbfd08/config/hubs/cloudbank.cluster.yaml#L358-L367
3. So once this PR is merged, we'll be able to add the hub authentication configuration to make Auth0 provide a "CILogon" button for authentication.

Is that right?

choldgraf (approved):
I think we should merge this one in, and then go through the process of setting up CILogon for a test hub to see if / how it works. I'm +1 to merge as long as we track these next steps in the other issue. Will update it so we don't forget

GeorgianaElena:
I follow @choldgraf's suggestion and merge this, then track follow-up doc PRs in the original issue. Thanks @choldgraf!

config/hubs/schema.yaml
```diff
@@ -179,6 +179,7 @@ properties:
                   - google-oauth2
                   - github
                   - password
+                  - CILogon
```
damianavila, line 182: I was expecting all lowercase here, but it is not a big deal at all 😉.

GeorgianaElena: Ah wow! Yes, lowercase would have been better. Will keep this in mind and when I'll document this, I will also change the casing :sweat_smile: TY!

PR #822, opened 2021-11-10
Add rule to allow traffic from singleuser pods to proxy pod

sgibson91:
This PR fixes the issue raised here https://discourse.pangeo.io/t/trying-to-open-gateway-cluster/1912 that was temporarily mitigated in PR https://github.com/2i2c-org/infrastructure/pull/819

It adds an egress rule to the singleuser pods allowing traffic to the proxy pod, this is particularly important for daskhubs that use the proxy to create clusters (schedulers and workers)

PR #819, opened 2021-11-09
Temporarily disable singleuser networkpolicy for Pangeo

sgibson91:
Temporarily mitigates issue reported in https://discourse.pangeo.io/t/trying-to-open-gateway-cluster/1912 but we should come up with a better fix

PR #824, opened 2021-11-10
Reenable singleuser networkpolicy for Pangeo hubs

sgibson91:
Reverts 2i2c-org/infrastructure#819 Bug was fixed in #822


sgibson91:
Merging this without a review as it's a reversion of a temporary bug fix.

Issue #594, opened 2021-08-09
Use nginx-ingress as entrypoint for all hubs


yuvipanda:
For some hubs that don't have the support chart deployed (#388), we use individual LoadBalancer IPs to expose the hubs. This prevents easily exposing something like grafana, and also adds complexity - some hubs use LoadBalancer, and some use nginx-ingress.

We should move everything to nginx-ingress, and new hubs should all use the nginx-ingress entrypoint.

We need to move the following hubs:

- [ ] Farallon
- [ ] Openscapes
- [x] Meom-IGE
- [ ] Pangeo-181919 (COESSING hub I think)

We can do so by:

1. Changing TTL of current DNS entry for a hub to 1minute, so changes can propagate faster. This will take some time to take effect - as long as the current TTL (often 5 minutes) of the record.
2. Change the record, pointing it to the IP / CNAME of the nginx-ingress load balancer service in the `support/` namespace
3. Set `proxy.https.enabled` to `false` for the hub, and unset `proxy.service.type`. Do a deploy immediately. The `autohttps` deployment should go away, and `proxy-public` should change type. You might have to delete the `proxy-public` service before doing a deploy, or the deploy might fail, since it'll try to change the service from a LoadBalancer to a ClusterIP (and this is often not fully supported)
4. Make this change in a PR and merge it

This might lead to a minute or two of downtime, so has to be done at a time that'll hopefully affect nobody.

damianavila:
> We should move everything to nginx-ingress, and new hubs should all use the nginx-ingress entrypoint.

So this means we are deploying support in all our clusters, right? I would agree with that.

choldgraf:
I'm setting this as medium priority and adding to our backlog now...these kinds of "standardize and simplify our infrastructure" issues seem particularly important to make sure that we can iterate quickly and without confusing ourselves in the future

yuvipanda:
This is done now, thanks to @GeorgianaElena!

damianavila:
For future readers, Farallon is being decommissioned here: https://github.com/2i2c-org/infrastructure/issues/1454.

PR #533, opened 2021-07-20
Setup prometheus + grafana for carbonplan

yuvipanda:
Without this we were flying blind.

Based off https://github.com/2i2c-org/pilot-hubs/pull/456

I manually updated the DNS records, and had to delete
`proxy-public` svc to get this to work.

Issue #431, opened 2021-05-25
Guidelines for using kops vs EKS


yuvipanda:
We currently use [kops](https://kops.sigs.k8s.io/) to manage AWS clusters. This is primarily driven by clunkiness of EKS and lack of features (see https://github.com/aws/containers-roadmap/issues/724 for example), and has worked out well for us. However, this does mean that we're responsible for the k8s master - and that's a pretty big responsibility! I also made the decision single-handedly at that time, and it's useful to properly evaluate it with some set criteria I think.

We initially started with EKS, and [this issue](https://github.com/2i2c-org/pangeo-hubs/issues/28) documents some of the process behind switching to `kops`. However, it's not structured enough to give me confidence that it is the right thing to do, and to help re-evaluate the decision as EKS is fast moving.

### Resolution

We've decided to use `EKS` instead of `kops` for AWS. See this comment for a rationale: https://github.com/2i2c-org/infrastructure/issues/431#issuecomment-968495599

Here's a start for some criteria we could use. 

| Criteria | `kops` | EKS |
| - | - | - |
| Base cost | | |
| Automation support | | |
| Master debuggability | | |
| Node debuggability | | |
| Operational UX | | |
| Node upgrade experience | | |
| Node config change experience | | |
| Master upgrade experience | | |
| Master config change experience | | |
| Security | | |
| Access to k8s features | | |
| AWS community support | | |
| z2jh community experience | | |
| Migrateability (to other option) | | |

damianavila:
I think making this comparison is a great idea!
I will start filling some of those empty buckets soon (and maybe add some other rows).

consideRatio:
> is is primarily driven by clunkiness of EKS and lack of features (see aws/containers-roadmap#724 for example), and has worked out well for us.

I agree on some clunkiness, but to me it mostly relate to the idea of managed nodes which is something I figured that if avoided is quite unproblematic. I have not used `kops` though. I'll try list all the observed clunkiness of the various options in this issue together going onwards.

yuvipanda:
[...]

PR #827, opened 2021-11-11
Deploy support chart for the openscapes hub

GeorgianaElena:
After this PR is merged and the support chart is deployed, we should manually trigger the [grafana deployer action](https://github.com/2i2c-org/infrastructure/actions/workflows/deploy-grafana-dashboards.yaml) in order to have the grafana dashboards available for this hub.

Ref: https://github.com/2i2c-org/infrastructure/issues/810

### Note:
(from  [jupyterhub/grafana-dashboards](https://github.com/jupyterhub/grafana-dashboards#deployment))

*NOTE: ANY CHANGES YOU MAKE VIA THE GRAFANA UI WILL BE OVERWRITTEN NEXT TIME YOU RUN deploy.bash. TO MAKE CHANGES, EDIT THE JSONNET FILE AND DEPLOY AGAIN*

So just a heads up that if we deploy the dashboards through the action, any changes made to the other's hub grafanas will be ovewriten.

### DIFF
--- a/config/hubs/openscapes.cluster.yaml
+++ b/config/hubs/openscapes.cluster.yaml
@@ -6,6 +6,16 @@ aws:
   clusterName: openscapeshub.k8s.local
   region: us-west-2
   stateStore: s3://2i2c-openscapes-kops-state
+support:
+  config:
+    grafana:
+      ingress:
+        hosts:
+          - grafana.openscapes.2i2c.cloud
+        tls:
+          - secretName: grafana-tls
+            hosts:
+              - grafana.openscapes.2i2c.cloud
 hubs:
   - name: staging
     domain: staging.openscapes.2i2c.cloud


damianavila (approved):
LGTM, although I think there are two follow-ups:
1. DNS entry for the grafana DNS you picked: https://github.com/2i2c-org/infrastructure/pull/533#issuecomment-883290848
2. The grafana dashboards you mentioned

GeorgianaElena:
> LGTM, although I think there are two follow-ups:

Thanks @damianavila! :D

> DNS entry for the grafana DNS you picked: #533 (comment)

Waa, didn't know we had steps for this written somewhere! This is great <3! TY!
(We have them in the docs too https://infrastructure.2i2c.org/en/latest/howto/operate/grafana.html?highlight=support#deploy-the-support-chart btw :tada: )

GeorgianaElena:
@damianavila, I deployed the support chart manually, but finished with a timeout error.
The support pods is in a pending state and I see this warning during scheduling:

```
0/3 nodes are available: 1 node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn't tolerate, 2 node(s) had taint {hub.jupyter.org_dedicated: user}, that the pod didn't tolerate.
```

Any ideas on how to spin up a node for the support pods?

damianavila:
Mmm... I was expecting most of the support-related pods to be scheduled on the master node that also serves as the core node in our kops-based deployments.
Wondering if we might need to work around some of the support-related pods to run on master nodes as we do with CoreDNS 🤔

@yuvipanda might have more details/thoughts/ideas...

Also, wondering if we would face this one as well: https://github.com/2i2c-org/infrastructure/issues/594
Notice that, in the old carbonplan PR (https://github.com/2i2c-org/infrastructure/pull/533), a few more things were needed to set up support (check the LoadBalancer and traefik removals). And I think that PR was on top of some previos updates where the master node was actually split into master + core independent nodes: https://github.com/2i2c-org/infrastructure/pull/532. So, adding support to kops-based deployments could be more complex than we originally thought... 😢

damianavila:
> So, adding support to kops-based deployments could be more complex than we originally thought... 😢

Btw, if things get really complicated here, my inclination would be to default to the current stability and keep the _status quo_ even when that means no support (nor grafana) stuff in the openscapes cluster. We are super close to the event start date and changes should be as minimal and simple as possible if we want to avoid surprises during the event...

yuvipanda:
Yeah, you definitely need the same extra tolerations the core dns pod gets on all the pods we schedule in the master nodes. 


Might help to timebox the support node deployment setup. You don't *have* to use the nginx ingress for the hubs, so the load balancer can just stay for the hubs - we don't have to touch that. However, we must make sure that the extra pods being present on the master node - particularly prometheus - doesn't put undue extra stress on the master node components. So let's make sure they have cpu and memory limits set, and maybe are on a node by themselves?

yuvipanda:
But let's think of this as 'adding Prometheus and grafana' rather than 'migrating a kops based deployment to use the support chart'. Let's migrate off kops soon instead!

choldgraf:
From these conversations, how about this proposal:

### Proposal

- Try deploying the support resources for, say, 1 hour. If that doesn't work, then go with @damianavila's suggestion to focus more on doing what we think will make the infrastructure stable ahead of Monday's event. If somebody thinks "this will definitely take more than an hour", then we should just skip this bit and know that we'll need to pay extra attention via `kubectl` during the event.
- Regarding EKS, I [made this proposal here to remove `kops` from our infrastructure](https://github.com/2i2c-org/infrastructure/issues/431#issuecomment-967286679) and would love to hear thoughts once we have bandwidth to revisit.

damianavila:
This will definitely take more than 1 hour, IMHO.

And, more importantly, implies modifications bigger enough I am not conformable with pushing forward on a Friday evening before the Monday start of the event.

I would suggest closing this PR now and focusing on the transition to EKS after the event.

choldgraf:
> I would suggest closing this PR now and focusing on the transition to EKS after the event.

This sounds like a good plan to me. I agree this feels like we are doing too many things "for the first time and with uncertainty about whether it'll work" just before an event...

damianavila:
Closing this one now to avoid confusion about what things are really active for the upcoming event.

Thanks for opening the PR, @GeorgianaElena!!

Issue #440, opened 2021-06-01
Correctly apply `chown` to `/shared` folders across hubs


consideRatio:
# Summary

I needed to add a volume mount hack for the ~/shared folder. I'm not fully confident about this in general though, but this is what I did.

This change can now be seen in https://github.com/2i2c-org/pilot-hubs/pull/436/commits/9c5e6ee373545a2a4565e36b1342d1e16d35dd7c but below is a summary.

### Before

```yaml
          singleuser:
          # ...
            initContainers:
              # Need to explicitly fix ownership here, since EFS doesn't do anonuid
              - name: volume-mount-ownership-fix
                image: busybox
                command: ["sh", "-c", "id && chown 1000:1000 /home/jovyan && ls -lhd /home/jovyan"]
                securityContext:
                  runAsUser: 0
                volumeMounts:
                  - name: home
                    mountPath: /home/jovyan
                    subPath: "{username}"
```

### Current

```yaml
          singleuser:
          # ...
            initContainers:
              # Need to explicitly fix ownership here, since EFS doesn't do anonuid
              - name: volume-mount-ownership-fix
                image: busybox
                command: ["sh", "-c", "id && chown 1000:1000 /home/jovyan /home/jovyan/shared && ls -lhd /home/jovyan"]
                securityContext:
                  runAsUser: 0
                volumeMounts:
                  - name: home
                    mountPath: /home/jovyan
                    subPath: "{username}"

                  # Added along with the /home/jovyan/shared entry in the `chown` command
                  - name: home
                    mountPath: /home/jovyan/shared
                    subPath: _shared
```

### Action points
- [ ] Conclude if this is needed and wanted
- [ ] Do this for the other hubs with this kind of logic in place.

  Note that I copied carbonplan.cluster.yaml for jmte.cluster.yaml that I've currently not pushed to the repo.

yuvipanda:
Very interesting that I didn't do this for carbonplan or openscapes and it seems to have just worked fine?!

_shared only needs to be created once per hub, so maybe it should be part of nfs-share-creator?

consideRatio:
Yes i think that may have solved these matters, i'll explore that

damianavila:
This was also reported here: https://github.com/2i2c-org/infrastructure/issues/810#issuecomment-967407615.

> Very interesting that I didn't do this for carbonplan or openscapes and it seems to have just worked fine?!

[...]

PR #832, opened 2021-11-14
Fix shared directory ownership and enable w+r on it

damianavila:
The first piece of this commit is executing on the workaround described
in [1] for the missing ownership on the _shared directory [2].
The second pieces essentially disable the ReadOnly attribute on the
mounted _shared directory so it is writable by everyone (admin and
users)[3].

[1] https://github.com/2i2c-org/infrastructure/issues/440
[2] https://github.com/2i2c-org/infrastructure/issues/810#issuecomment-968307309
[3] https://github.com/2i2c-org/infrastructure/issues/810#issuecomment-968308530

### DIFF
--- a/config/hubs/openscapes.cluster.yaml
+++ b/config/hubs/openscapes.cluster.yaml
@@ -60,7 +60,7 @@ hubs:
                   [
                     "sh",
                     "-c",
-                    "id && chown 1000:1000 /home/jovyan && ls -lhd /home/jovyan",
+                    "id && chown 1000:1000 /home/jovyan /home/jovyan/shared && ls -lhd /home/jovyan",
                   ]
                 securityContext:
                   runAsUser: 0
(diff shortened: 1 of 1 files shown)


damianavila:
As you can see from the linked comment, I have successfully tested this on openscapes `staging`.
Wondering if they are not going to write themselves to top of each other... but I made that clear as a risk in my refereced comment. Do you foresee any other risk?

yuvipanda (approved):
If they want this, this looks good! Curious to see how people prevent overwriting each other's work

config/hubs/openscapes.cluster.yaml
```diff
@@ -60,18 +60,27 @@ hubs:
                   [
                     "sh",
                     "-c",
-                    "id && chown 1000:1000 /home/jovyan && ls -lhd /home/jovyan",
+                    "id && chown 1000:1000 /home/jovyan /home/jovyan/shared && ls -lhd /home/jovyan",
```
consideRatio, line 63: Is we chown on a dir that doesn't exist, this fails right?

Q: can/should we assume this to exist?

consideRatio (approved):
LGTM from what I can tell!

choldgraf (approved):
+1 since they asked for it, we should ask them how they're using this functionality, as this is a new request that we haven't tried before!

Thanks @damianavila for tackling this on a weekend 🙏

damianavila:
The openscapes folks re-confirmed they want the writable shared space so going to merge this one now.
Thanks all for the review on Sunday time!

Issue #800, opened 2021-10-31
New Hub: CarbonPlan / Azure


jhamman:
### Hub Description

CarbonPlan is a non-profit organization that builds open data and tools to accelerate the deployment of quality and transparent climate solutions. We are working on a new project that is utilizing the Microsoft Azure cloud platform. The hub we'd like to see deployed is a "Pangeo-style" hub, including spot instances for Dask workloads and GPU instances for some single-user sessions.

### Community Representative(s)

- @jhamman 
- @norlandrhagen

### Important dates

No specific dates. We'll be using this hub for exploratory data analysis and prototyping in November, data production in December, and continued research through at least April 2022.

### Target start date

11/5/2021

### Preferred Cloud Provider

Microsoft Azure

### Do you have your own billing account?

- [X] Yes, I have my own billing account.

### Hub Authentication Type

GitHub Authentication (e.g., @mygithubhandle)

### Hub logo URL

https://carbonplan.org/

### Hub image service

Dockerhub / quay.io

### Hub image

tbd, for now: pangeo/pangeo-notebook:latest

### Extra features you'd like to enable

- [X] Dedicated Scalable Dask Cluster with [Dask Gateway](https://gateway.dask.org/)

### Hub ID

`carbonplan-azure`

### Hub Cluster

(New!) `carbonplan-azure`

### Hub URL

`azure.carbonplan.2i2c.cloud`

### Hub Template

daskhub

### Tasks to deploy the hub

- [x] Engineer who will deploy the hub is assigned
- [x] Deploy information filled in above
- [x] Initial Cluster deployment PR: https://github.com/2i2c-org/infrastructure/pull/833
- [x] Initial Hub deployment PR: https://github.com/2i2c-org/infrastructure/issues/838
- [x] Administrators able to log on
- [x] Community Representative satisfied with hub environment
- [x] Hub now in steady-state

sgibson91:
Confirming that @jhamman has given me access to the Azure subscription to be used 🎉

sgibson91:
I have opened a PR to deploy a new cluster on Azure here: https://github.com/2i2c-org/infrastructure/pull/833

sgibson91:
PR for new hubs is here: #838

sgibson91:
I opened https://github.com/2i2c-org/infrastructure/issues/840 to track the object storage connection

sgibson91:
Hey @jhamman the hubs are up!

[...]

PR #844, opened 2021-11-23
Change CarbonPlan's user image on Azure hubs

sgibson91:
As requested in https://github.com/2i2c-org/infrastructure/issues/800#issuecomment-975799504


sgibson91:
I have deployed this to the staging and prod hubs so will merge

Issue #853, opened 2021-11-24
Deploy a new cluster for the U. Toronto community


choldgraf:
### Description

As a part of https://github.com/2i2c-org/infrastructure/issues/638 we need to deploy a new cluster/hub for U. Toronto that runs off of our CI/CD infrastructure.

### Value / benefit

This is a pre-requisite for us being able to transition the U. Toronto hub onto the new setup!
We now have Azure terraform support in this repo (#800) and can set up a new hub infra with that, to run in parallel with current setup. The primary infra difference will be the use of Azure File for home directories rather than the hand-spun NFS VM we have there right now - I think this is a big positive change.

### Tasks to complete

- [x] **Deploy a new cluster**
  - [x] Create a new Azure based infra using the documentation instructions in https://infrastructure.2i2c.org/en/latest/howto/operate/new-cluster/new-cluster.html

choldgraf:
closed by https://github.com/2i2c-org/infrastructure/pull/856

PR #859, opened 2021-11-29
Update Azure new cluster instructions

GeorgianaElena:
Ref: https://github.com/2i2c-org/infrastructure/issues/853

### DIFF
--- a/docs/howto/operate/new-cluster/new-cluster.md
+++ b/docs/howto/operate/new-cluster/new-cluster.md
@@ -84,7 +84,7 @@ We recommend the following conventions using `lowercase`:
 - `{CLUSTER_NAME}hubstorage` for storage accounts
 
 ````{ note}
-Changes in Azure's own requirements might break our recommended convension. If any such failure occurs, please signal it.
+Changes in Azure's own requirements might break our recommended convention. If any such failure occurs, please signal it.
 ````
 
 This increases the probability that we won't take up a namespace that may be required by the Hub Community, for example, in cases where we are deploying to Azure subscriptions not owned/managed by 2i2c.
--- a/docs/howto/operate/new-cluster/new-cluster.md
+++ b/docs/howto/operate/new-cluster/new-cluster.md
@@ -216,7 +216,7 @@ You must be logged into Google with your `@2i2c.org` account at this point so `s
 
 ```bash
 cd ../..
-sops --encrypt --in-place secrets/CLUSTER_NAME.yaml
+sops --encrypt --in-place secrets/CLUSTER_NAME.{{ json | yaml }}
 ```
 
 This key can now be committed to the `infrastructure` repo and used to deploy and manage hubs hosted on that cluster.
--- a/docs/howto/operate/new-cluster/new-cluster.md
+++ b/docs/howto/operate/new-cluster/new-cluster.md
@@ -78,7 +78,7 @@ Names for Azure container registries and storage accounts **must** conform to th
 A failure will occur if you try to create a storage account whose name is not entirely lowercase.
 ````
 
-We explictly recommend the following conventions using `lowercamelcase`:
+We recommend the following conventions using `lowercase`:
 
 - `{CLUSTER_NAME}hubregistry` for container registries
 - `{CLUSTER_NAME}hubstorage` for storage accounts
--- a/docs/howto/operate/new-cluster/new-cluster.md
+++ b/docs/howto/operate/new-cluster/new-cluster.md
@@ -74,10 +74,18 @@ Names for Azure container registries and storage accounts **must** conform to th
 - alphanumeric strings between 5 and 50 characters for [container registries](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftcontainerregistry), e.g., `myContainerRegistry007`
 - lowercase letters and numbers strings between 2 and 24 characters for [storage accounts](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftstorage), e.g., `mystorageaccount314`
 
-We explictly recommend the following conventions using `lowerCamelCase`:
+````{ note}
+A failure will occur if you try to create a storage account whose name is not entirely lowercase.
+````
+
+We explictly recommend the following conventions using `lowercamelcase`:
 
 - `{CLUSTER_NAME}hubregistry` for container registries
 - `{CLUSTER_NAME}hubstorage` for storage accounts
+
+````{ note}
+Changes in Azure's own requirements might break our recommended convension. If any such failure occurs, please signal it.
+````
 
 This increases the probability that we won't take up a namespace that may be required by the Hub Community, for example, in cases where we are deploying to Azure subscriptions not owned/managed by 2i2c.
 ```
--- a/docs/howto/operate/new-cluster/new-cluster.md
+++ b/docs/howto/operate/new-cluster/new-cluster.md
@@ -46,8 +46,8 @@ The _minimum_ inputs this file requires are:
 
 - `subscription_id`: Azure subscription ID to create resources in.
   Should be the id, rather than display name of the project.
-- `resourcegroup_name`: The name of the Resource Group that the cluster and other resources will be deployed into.
-- `global_container_registry_name`: The name of an Azure Container Registry to use for our image.
+- `resourcegroup_name`: The name of the Resource Group to be created by terraform, where the cluster and other resources will be deployed into.
+- `global_container_registry_name`: The name of an Azure Container Registry to be created by terraform to use for our image.
   This must be unique across all of Azure.
   You can use the following [Azure CLI](https://docs.microsoft.com/en-us/cli/azure/) command to check your desired name is available:
 
@@ -55,7 +55,7 @@ The _minimum_ inputs this file requires are:
   az acr check-name --name ACR_NAME --output table
   ```
 
-- `global_storage_account_name`: The name of a storage account to use for Azure File Storage.
+- `global_storage_account_name`: The name of a storage account to be created by terraform to use for Azure File Storage.
   This must be unique across all of Azure.
   You can use the following [Azure CLI](https://docs.microsoft.com/en-us/cli/azure/) command to check your desired name is available:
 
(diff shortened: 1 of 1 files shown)


docs/howto/operate/new-cluster/new-cluster.md
```diff
@@ -46,16 +46,16 @@ The _minimum_ inputs this file requires are:
 
 - `subscription_id`: Azure subscription ID to create resources in.
   Should be the id, rather than display name of the project.
-- `resourcegroup_name`: The name of the Resource Group that the cluster and other resources will be deployed into.
-- `global_container_registry_name`: The name of an Azure Container Registry to use for our image.
+- `resourcegroup_name`: The name of the Resource Group to be created by terraform, where the cluster and other resources will be deployed into.
```
GeorgianaElena, line 49: I wasn't sure of this. Will the resource group be created by terraform if it doesn't exist?

docs/howto/operate/new-cluster/new-cluster.md
```diff
@@ -72,12 +72,12 @@ See the [variables file](https://github.com/2i2c-org/infrastructure/blob/HEAD/te
...
 
 - alphanumeric strings between 5 and 50 characters for [container registries](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftcontainerregistry), e.g., `myContainerRegistry007`
-- alphanumeric strings between 2 and 24 characters for [storage accounts](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftstorage), e.g., `myStorageAccount314`
+- lowercase letters and numbers strings between 2 and 24 characters for [storage accounts](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftstorage), e.g., `mystorageaccount314`
 
 We explictly recommend the following conventions using `lowerCamelCase`:
 
-- `{CLUSTER_NAME}HubRegistry` for container registries
-- `{CLUSTER_NAME}HubStorage` for storage accounts
+- `{CLUSTER_NAME}hubregistry` for container registries
```
sgibson91, line 79:
```suggestion
- `{CLUSTER_NAME}HubRegistry` for container registries
```

So we _can_ use upper case in container registry names, the lower case restriction only applies to storage accounts. So the question is, what standard do we want to go with?

docs/howto/operate/new-cluster/new-cluster.md
```diff
@@ -113,14 +113,16 @@ terraform init -backend-config=backends/default-backend.hcl -reconfigure
 ```
 
 ````{note}
-If you are working on a project which you cannot access with your 2i2c account, there are other backend config files stored in `terraform/backends` that will configure a different storage bucket to read/write the remote terraform state.
+There are other backend config files stored in `terraform/backends` that will configure a different storage bucket to read/write the remote terraform state for projects which we cannot access with our 2i2c account,.
```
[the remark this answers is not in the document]
sgibson91:
```suggestion
There are other backend config files stored in `terraform/backends` that will configure a different storage bucket to read/write the remote terraform state for projects which we cannot access with our 2i2c account.
```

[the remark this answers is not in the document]
sgibson91: "our 2i2c account" is "logging into GCP with your `@2i2c.org` email"

docs/howto/operate/new-cluster/new-cluster.md
choldgraf: Let's just go with CamelCase since that's what we've had documented so far, and add a note that in some cases technical limitations might prevent this from happening, in which case it's OK to just use all-lowercase.

sgibson91: I have no strong preference either way. I think the important thing to document is that _a failure will occur if you try to create a storage account whose name is not 100% lowercase_.

GeorgianaElena: > Let's just go with CamelCase since that's what we've had documented so far, and add a note that in some cases technical limitations might prevent this from happening, in which case it's OK to just use all-lowercase.

We have documented CamelCase indeed, but all existing Azure clusters use lowercase. My preference is use lower case for both, for the sake of consistency and keeping it simple. Otherwise, we would end up with all sorts of naming schemas: "-", CamelCase and lower case:

```
subscription_id                = "my-awesome-subscription-id"
resourcegroup_name             = "my-awesome-resource-group"
global_container_registry_name = "myAwesomeHubRegistry"
global_storage_account_name    = "myawesomestorageaccount"
ssh_pub_key                    = "ssh-rsa my-public-ssh-key"
```

> I think the important thing to document is that a failure will occur if you try to create a storage account whose name is not 100% lowercase.

:+1:

docs/howto/operate/new-cluster/new-cluster.md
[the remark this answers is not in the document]
GeorgianaElena: I edited your suggestion with this info @sgibson91! Hope that's ok.

docs/howto/operate/new-cluster/new-cluster.md
choldgraf: ah - in that case I think our decision criteria should be:

- A. If we expect to be *required* to use `lower-case` only occasionally, then we should use `CamelCase` and on rare occasions use `lower-case`
- B. If we instead expect to be required to use `lower-case` fairly often, and as long as we can *always* use `lower-case` for naming, then I think we should just standardize on `lower-case`

So which seems more likely?

sgibson91: I think A is most likely. For example, I know `CamelCase` is allowed for the Container Registry.

> We have documented CamelCase indeed, but all existing Azure clusters use lowercase. My preference is use lower case for both, for the sake of consistency and keeping it simple. Otherwise, we would end up with all sorts of naming schemas: "-", CamelCase and lower case:

@GeorgianaElena I also the Azure terraform config could use a lot of cleaning up to bring it in-line with the GCP config, we can standardise during that cleanup process.

damianavila: Even if we are in scenario A, I do concur with @GeorgianaElena about the fact that mixing "schemas" could be potentially confusing...

sgibson91: Sorry, I meant the _requirement to use lowercase only occasionally_ part of the statement is most likely.

damianavila: > requirement to use lowercase only occasionally

Still, I think it might introduce a waste of time and confusion but...

> I also the Azure terraform config could use a lot of cleaning up to bring it in-line with the GCP config, we can standardise during that cleanup process.

if we are really going to standardize later I guess I would be OK with a temporary scenario with "mixed schemas".

sgibson91: > Still, I think it might introduce a waste of time and confusion but...

Yikes, I'm sorry, I'm still not doing well at communicating this.

The occasional requirement for lowercase in some resource names comes from Azure. So I was saying that if we wanted to standard on CamelCase, we would only occasionally run into this error. I still have no preference of what standard we adopt. I think the important thing to document is that whatever standard we adopt, Azure's own requirements will probably break it at some point. Because Azure is annoying.

damianavila: > I think the important thing to document is that whatever standard we adopt, Azure's own requirements will probably break it at some point. Because Azure is annoying.

💯

GeorgianaElena:
Thanks everyone for their reviews :rocket: ! I added a couple of notes about Azure's naming requirements and hope I captured everyone's ideas around this.

Please LTMK if there's anything else I should address.

docs/howto/operate/new-cluster/new-cluster.md
```diff
@@ -72,12 +72,20 @@ See the [variables file](https://github.com/2i2c-org/infrastructure/blob/HEAD/te
...
 - alphanumeric strings between 5 and 50 characters for [container registries](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftcontainerregistry), e.g., `myContainerRegistry007`
-- alphanumeric strings between 2 and 24 characters for [storage accounts](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftstorage), e.g., `myStorageAccount314`
+- lowercase letters and numbers strings between 2 and 24 characters for [storage accounts](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftstorage), e.g., `mystorageaccount314`
 
-We explictly recommend the following conventions using `lowerCamelCase`:
+````{ note}
+A failure will occur if you try to create a storage account whose name is not entirely lowercase.
+````
+
+We explictly recommend the following conventions using `lowercamelcase`:
```
choldgraf, line 81:
```suggestion
We recommend the following conventions using `lowercase`:
```

I believe that "lower camel case" means something like `lowerCamelCase` (as opposed to "camel case" which is `CamelCase`). So we are just doing plain old `lowercase`

choldgraf (commented):
just a quick note! I'm +1 if others think this looks good

docs/howto/operate/new-cluster/new-cluster.md
```diff
@@ -72,12 +72,20 @@ See the [variables file](https://github.com/2i2c-org/infrastructure/blob/HEAD/te
...
+
+We recommend the following conventions using `lowercase`:
 
-- `{CLUSTER_NAME}HubRegistry` for container registries
-- `{CLUSTER_NAME}HubStorage` for storage accounts
+- `{CLUSTER_NAME}hubregistry` for container registries
+- `{CLUSTER_NAME}hubstorage` for storage accounts
+
+````{ note}
+Changes in Azure's own requirements might break our recommended convension. If any such failure occurs, please signal it.
```
sgibson91, line 87:
```suggestion
Changes in Azure's own requirements might break our recommended convention. If any such failure occurs, please signal it.
```

docs/howto/operate/new-cluster/new-cluster.md
```diff
@@ -200,8 +215,8 @@ You must be logged into Google with your `@2i2c.org` account at this point so `s
 ```
 
 ```bash
-cd ..
-sops --encrypt --in-place secrets/CLUSTER_NAME.json
+cd ../..
+sops --encrypt --in-place secrets/CLUSTER_NAME.yaml
```
sgibson91, line 219:
```suggestion
sops --encrypt --in-place secrets/CLUSTER_NAME.{{ json | yaml }}
```

This depends on what sort of file we're outputting. GCP's Service Accounts will be JSON, as will Azure's Service Principals when I figure that out. But raw kubeconfig files are YAML. I'm not sure of the best way to write this?

sgibson91 (approved):
I have left a couple of minor comments on this, but I'm +1 too!

docs/howto/operate/new-cluster/new-cluster.md
GeorgianaElena: Oooh, right! Makes sense now. Thank you!

GeorgianaElena:
Thanks all! :rocket:

Issue #841, opened 2021-11-19
Automating deployments to Azure hubs


sgibson91:
### Description

Add deployer and CI support to automatically deploy upgrades to Azure-hosted hubs. This will complete the trifecta of automated deployments to the three major cloud providers we support.

### Value / benefit

- All major cloud deployments can be deployed automatically

### Tasks to complete

- [x] Update Azure Terraform config to create Service Principals for authentication beyond a raw kubeconfig files: https://github.com/2i2c-org/infrastructure/pull/864
- [x] Update the deployer to support Azure as an auth provider
- [x] Update the hub config's general schema to support Azure as a provider
- [ ] Update CI workflow to include Azure tools and deploy Azure hubs
  - Azure CLI is already available in GitHub-hosted runners, so this didn't need to be addressed: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#cli-tools

choldgraf:
yesss 💯 

this is one that you'd like to prioritize soon, is that right @sgibson91 ? 

(oh I see it is already on the sprint board so I guess the answer is yes haha)

sgibson91:
PR opened #865

Issue #840, opened 2021-11-19
Setting up read/write access to blob storage for Azure


sgibson91:
### Description

In our GCP and AWS deployments, we enable read/write access to object storage with _cloud_ service accounts that are assigned to _kubernetes_ service accounts. We should implement a similar connection on our Azure deployments.

Relates to: https://github.com/2i2c-org/infrastructure/issues/800
Initial discussion: https://github.com/2i2c-org/infrastructure/pull/838#issuecomment-972967683

### Value / benefit

We should have as similar as possible services/capabilities across the major cloud providers where possible

### Tasks to complete

**In terraform:**

- [ ] Create a service principal: https://docs.microsoft.com/en-us/azure/active-directory/develop/app-objects-and-service-principals
  - https://github.com/2i2c-org/infrastructure/pull/864
- [ ] Assign blob storage read/write role to the service principal on behalf of the kubernetes application: https://docs.microsoft.com/en-us/azure/storage/blobs/assign-azure-role-data-access
- [ ] Create the blob storage

**Then:**

- [ ] Do further research around linking Azure roles and kubernetes service accounts
  - Some details here: https://docs.microsoft.com/en-us/azure/aks/manage-azure-rbac#create-role-assignments-for-users-to-access-cluster
- [ ] Add more tasks as they arise here

sgibson91:
I think Azure's version of GCP's Config Connector is called Service Operator https://cloudblogs.microsoft.com/opensource/2020/06/25/announcing-azure-service-operator-kubernetes/

PR #864, opened 2021-12-01
Create a Service Principal for Azure clusters

sgibson91:
This PR updates our Azure Terraform config to create a service principal and secret and output these so we can use the info to login in a similar way to how GCP uses Service Accounts, and hence replacing the raw kubeconfig auth method.

This is the first piece for both #840 and #841


yuvipanda (changes requested):
\o/ This is super awesome.

Our UToronto access doesn't let us create service principals. While we should eventually get that fixed, I don't want the migration to block on that. So we should try make these temporarily conditional.

sgibson91:
Good to know, cheers!

sgibson91:
Ok @yuvipanda I have no made it optional to create service principals

yuvipanda (approved):
LGTM!

sgibson91:
Hmmmm, seems like I still don't have enough privileges - @jhamman?

> Error: Could not list existing service principals
> 
> with azuread_service_principal.service_principal[0],
>   on service-principal.tf line 1, in resource "azuread_service_principal" "service_principal":
>    1: resource "azuread_service_principal" "service_principal" {
> 
> ServicePrincipalsClient.BaseClient.Get(): unexpected status 403 with OData error: Authorization_RequestDenied: Insufficient privileges to complete the
> operation.

sgibson91:
Ok, I'm going to rescope this PR to not apply the change to carbonplan and then merge it. We can figure out the permissions to create a service principal for carbonplan later.

jhamman:
Thanks @sgibson91 for working on this. Happy to adjust permissions as needed on our side. Confirming that right now, you do have the `Owner` role in the subscription being used for our deployment.

sgibson91:
Thanks @jhamman. Knowing Azure, I am 100% confident that it's something far more complicated and annoying :) I will update the main issue tomorrow

PR #865, opened 2021-12-02
Support and automate Azure deployments

sgibson91:
This PR brings in support for automatically deploying hubs on Azure clusters.

Changes in this PR:

- New function added to `deployer/hub.py` to authenticate against an Azure cluster
- Update to `config/hubs/schema.yaml` to support Azure config
- Update `.github/workflows/deploy-hubs.yaml` to install the Azure CLI for Azure-based hubs. <- I reverted this change since the Azure CLI is already available in GitHub-hosted runners: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#cli-tools

Note that we currently do not have any Azure-based hubs with a Service Principal to run this workflow on yet, so it may need some tweaking in the future.

fixes https://github.com/2i2c-org/infrastructure/issues/841

### DIFF
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -222,6 +222,13 @@ class Cluster:
                 os.environ['KUBECONFIG'] = kubeconfig.name
 
                 with decrypt_file(key_path) as decrypted_key_path:
+
+                    decrypted_key_abspath = os.path.abspath(decrypted_key_path)
+                    if not os.path.isfile(decrypted_key_abspath):
+                        raise FileNotFoundError(
+                            'The decrypted key file does not exist'
+                        )
+
                     with open(decrypted_key_path) as f:
                         service_principal = json.load(f)
 
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -42,6 +42,8 @@ class Cluster:
             yield from self.auth_gcp()
         elif self.spec['provider'] == 'aws':
             yield from self.auth_aws()
+        elif self.spec['provider'] == 'azure':
+            yield from self.auth_azure()
         elif self.spec['provider'] == 'kubeconfig':
             yield from self.auth_kubeconfig()
         else:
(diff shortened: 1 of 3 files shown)


sgibson91:
Azure CLI is already available on GitHub-hosted runners: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#cli-tools

deployer/hub.py
```diff
@@ -202,6 +204,53 @@ def auth_aws(self):
...
+        resource_group = config['resource_group']
+
+        with tempfile.NamedTemporaryFile() as kubeconfig:
+            orig_kubeconfig = os.environ.get('KUBECONFIG', None)
+
+            try:
+                os.environ['KUBECONFIG'] = kubeconfig.name
+
+                with decrypt_file(key_path) as decrypted_key_path:
+                    with open(decrypted_key_path) as f:
```
damianavila, line 225: On the AWS one, I added this piece:
```python
                    if not os.path.isfile(decrypted_key_abspath):
                        raise FileNotFoundError(
                            'The decrypted key file does not exist')
```
If you feel it makes sense, maybe you can add it here as well...

damianavila (approved):
Minor comment, otherwise, LGTM (modulo the Azure command are the correct ones, I have not interacted with AKS before 😉 ).

deployer/hub.py
sgibson91: Done!

sgibson91:
Thanks @damianavila!

choldgraf:
amazing! do you think this warrants updates to the docs? I don't think we discuss Azure much in the hub deployment docs, so maybe this just brings azure hubs in-line with what is already there?

sgibson91:
@choldgraf possibly? I'd need to have a deeper dive into the docs to know what to change, but I want to focus on https://github.com/2i2c-org/meta/issues/262 next because that's a much tighter deadline

Issue #638, opened 2021-08-30
Migrate UToronto hub to our pilot-hubs repository


yuvipanda:
## Problem statement

We currently deploy the University of Toronto hub via a dedicated GitHub organization and repository (https://github.com/utoronto-2i2c/jupyterhub-deploy/), using [hubploy](https://github.com/yuvipanda/hubploy) for deployment. This pre-dates our `pilot-hubs/` infrastructure which is why it is special-cased.

The problem is that, because it is on a dedicated repository / deployment infrastructure, it:

- doesn't run the same infrastructure setup as our main CI/CD, which makes it harder to understand
- is a duplication of some machinery which means we have more code to maintain
- is an information and access silo (e.g., only @yuvipanda and @GeorgianaElena could resolve issues #637 currently

## Solution

We should move this infrastructure to be deployed from `pilot-hubs/`, the same as any other Azure deployment (see the #288 for one example). This will involve a few steps, described below

### Step 1: Setup new hub Infra and prepare

We now have Azure terraform support in this repo (https://github.com/2i2c-org/infrastructure/issues/800) and can set up a new hub infra with that, to run in parallel with current setup. The primary infra difference will be the use of Azure File for home directories rather than the hand-spun NFS VM we have there right now - I think this is a big positive change.

Step 1 is all the things we can do that don't involve any downtime, and can be done early on.

- [x] https://github.com/2i2c-org/infrastructure/issues/854
- [x] https://github.com/2i2c-org/infrastructure/issues/853
- [x] #867
- [x] https://github.com/2i2c-org/infrastructure/issues/851
- [x] https://github.com/2i2c-org/infrastructure/issues/876

When these tasks are completed, I think we should have a *staging hub* that can be used by UToronto, but runs out of this cluster.

### Step 2: Migrate production hub

This should be done between Dec 13-15.

- [x] #868

TADA! New hub!

### Step 3: Clean up any loose ends

- [x] https://github.com/2i2c-org/infrastructure/issues/873
- [x] https://github.com/2i2c-org/infrastructure/issues/877
- [x] Copy the old `jupyterhub.log` :arrow_right: `gzip` it, and leave it in the new hub db dir. We can do usage analysis on this later, just gotta save it.
[...]

PR #866, opened 2021-12-02
Add utoronto staging hub

GeorgianaElena:
Closes https://github.com/2i2c-org/infrastructure/issues/867

### DIFF
--- a/config/hubs/utoronto.cluster.yaml
+++ b/config/hubs/utoronto.cluster.yaml
@@ -30,10 +30,10 @@ hubs:
     config: &utorontoHubConfig
       azureFile:
           enabled: true
-        nfs:
+      nfs:
+        enabled: false
+        shareCreator:
           enabled: false
-          shareCreator:
-            enabled: false
       jupyterhub:
         custom:
           homepage:
--- a/config/hubs/utoronto.cluster.yaml
+++ b/config/hubs/utoronto.cluster.yaml
@@ -92,7 +92,7 @@ hubs:
             AzureAdOAuthenticator:
               username_claim: oid
               login_service: "University of Toronto ID"
-              oauth_callback_url: https://staging.us-central1-b.gcp.pangeo.io/hub/oauth_callback
+              oauth_callback_url: https://staging.utoronto.2i2c.cloud/hub/oauth_callback
               tenant_id: 78aac226-2f03-4b4d-9037-b46d56c55210
           extraConfig:
             10-dynamic-subpath: |
--- a/config/hubs/utoronto.cluster.yaml
+++ b/config/hubs/utoronto.cluster.yaml
@@ -106,4 +106,3 @@ hubs:
                   "subPath": f"{pod_namespace}/{{username}}"
                 },
               ]
-
--- a/config/hubs/utoronto.cluster.yaml
+++ b/config/hubs/utoronto.cluster.yaml
@@ -29,7 +29,7 @@ hubs:
       enabled: false
     config: &utorontoHubConfig
       azureFile:
-          enabled: true
+        enabled: true
       nfs:
         enabled: false
         shareCreator:
@@ -56,11 +56,11 @@ hubs:
             name: quay.io/2i2c/utoronto-image
             tag: 83a724f5b829
           storage:
-              type: none
-              extraVolumes:
-                - name: home
-                  persistentVolumeClaim:
-                    claimName: home-azurefile
+            type: none
+            extraVolumes:
+              - name: home
+                persistentVolumeClaim:
+                  claimName: home-azurefile
         scheduling:
           userPlaceholder:
             enabled: false
--- a/config/hubs/utoronto.cluster.yaml
+++ b/config/hubs/utoronto.cluster.yaml
@@ -17,6 +17,10 @@ support:
       ingress:
         hosts:
           - grafana.utoronto.2i2c.cloud
+        tls:
+          - secretName: grafana-tls
+            hosts:
+              - grafana.utoronto.2i2c.cloud
 hubs:
   - name: staging
     domain: staging.utoronto.2i2c.cloud
@@ -24,6 +28,12 @@ hubs:
     auth0:
       enabled: false
     config: &utorontoHubConfig
+      azureFile:
+          enabled: true
+        nfs:
+          enabled: false
+          shareCreator:
+            enabled: false
       jupyterhub:
         custom:
           homepage:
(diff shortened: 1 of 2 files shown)


config/hubs/utoronto.cluster.yaml
```diff
@@ -0,0 +1,108 @@
...
+                memory: 128Mi
+              limits:
+                memory: 512Mi
+          traefik:
+            resources:
+              requests:
+                cpu: 0.1
+                memory: 256Mi
+              limits:
+                memory: 512Mi
```
GeorgianaElena, line 73: These are some resource limits from the old utoronto config. I wasn't sure whether to still use those or go with the default values in `basehub` template.

choldgraf: IMO we should stick with the current UToronto configuration values

choldgraf (commented):
This looks awesome! A couple questions to clarify:

- Now that #841 is done will this auto-deploy when we merge?
- This probably doesn't quite close #853 , because we'll need to deploy a `prod` hub as well, right?

config/hubs/utoronto.cluster.yaml
```diff
@@ -0,0 +1,108 @@
...
+          requests:
+            cpu: 0.5
+            memory: 4Gi
+          limits:
+            cpu: 2
+            memory: 16Gi
+    grafana:
+      ingress:
+        hosts:
+          - grafana.utoronto.2i2c.cloud
```
yuvipanda, line 19: DNS change needs to be done for this

yuvipanda (approved):
This lgtm, @GeorgianaElena. You'll have to make DNS changes for staging.utoronto.2i2c.cloud to point to the IP of the nginx-ingress for this to work. But once that's done, hopefully everything (including auth!) should work.

Do deal with DNS issues, my suggested workflow is:

1. Run `deploy-support` locally to deploy nginx-ingress
2. Change DNS to point domains to the nginx-ingress service's external IP
3. Then deploy the hub after 5-10 minutes, to allow time for the DNS change to propogate

This should avoid problems with let's encrypt trying to get HTTPS working properly.

config/hubs/utoronto.cluster.yaml
damianavila: Mmm... can I suggest otherwise, maybe? 😉 
I mean, do we know if these values are the proper ones for the current incantation of our hubs from this repo?
If we set the values without validation (and only because we inherit those values), we may be setting conditions that are not actually needed (nor optimal) and makes the whole thing more complex than it should be...
Maybe we can trace back why these values were needed and infer if we need to tweak them here as well before actually doing it?

damianavila:
> Now that #841 is done will this auto-deploy when we merge?

AFAIK, nop, this one is using a raw kubeconfig and auto-deploy needs setting up some Service Principal. AFAIK, the terraform stuff for that was merged quite recently, so I am not sure the cluster supporting this hub actually was built before or after that PR was merged. In addition, I think there are still some permission issues to be figured out yet accordingly to this comment: https://github.com/2i2c-org/infrastructure/pull/864#issuecomment-984463972

GeorgianaElena:
> Now that #841 is done will this auto-deploy when we merge?

@damianavila  is right, #841 won't work for the utoronto hub because of not enough permissions to create a service principal (https://github.com/2i2c-org/infrastructure/pull/864#pullrequestreview-820146574) :( Also, the cluster was built before the PR
 
> This probably doesn't quite close #853 , because we'll need to deploy a prod hub as well, right? Since nobody will be using this hub yet anyway, why don't we deploy both staging and prod in this PR?

I believe we should deploy the prod hub after we make sure that everything works on this staging hub. It should be a fairly straightforward task, that can happen as part of Step 2 of https://github.com/2i2c-org/infrastructure/issues/638. Also, we might need to take over the prod hub URL too (like we are doing here with the staging one).

config/hubs/utoronto.cluster.yaml
GeorgianaElena: That was my feeling too @damianavila, esp since I believe the utoronto hub was the first version of our infra.

> Maybe we can trace back why these values were needed and infer if we need to tweak them here as well before actually doing it?

I will dig into it to try find out if these were tweaked afterwards or were part of the initial deployment.

GeorgianaElena:
Ah, oops! I just noticed that in the original issue https://github.com/2i2c-org/infrastructure/issues/638, there this piece of context from @yuvipanda:

*We now have Azure terraform support in this repo (#800) and can set up a new hub infra with that, to run in parallel with current setup. The primary infra difference will be the use of Azure File for home directories rather than the hand-spun NFS VM we have there right now - I think this is a big positive change.*

I believe that NFS is enabled by default rather than Azure File, so that needs to change before merging this.

config/hubs/utoronto.cluster.yaml
choldgraf: Ah - I was making an assumption here, let me re-phrase my suggestion:

- **If they have explicitly requested those limits** then we should keep them.
- **If they have not and this was just set by us without explicit rationale for special-casing it** then we should just go with whatever our other configurations use

GeorgianaElena: Looking at the commit history, it seems that the limits were set by us.
But I'm not 100% sure this means that there wasn't a request from utoronto or a limitation observed at some point. 

l would love if @yuvipanda could confirm this :eyes: 

This is a side-by-side resource differences. Hope this makes it more easier to follow
resource-diffs

GeorgianaElena:
I would like to move this forward since we have a migration due date.
I was thinking about the following plan:

1. Deploy the support chart for the utoronto cluster per [@yuvipanda's comment](https://github.com/2i2c-org/infrastructure/pull/866#pullrequestreview-822467441).
2. If everything goes ok in step 1, merge this PR and check if the deploy goes well.
3. Open an issue about the differences in resource allocation that we have for the utoronto hub and discuss there if we should stick with those or try use the defaults we have in the infra repo

GeorgianaElena:
Since I have Yuvi's approval and we can tweak resources later, I will try a merge and check how the deploy goes :crossed_fingers:

damianavila:
> Open an issue about the differences in resource allocation that we have for the utoronto hub and discuss there if we should stick with those or try use the defaults we have in the infra repo

Totally OK with me (but let's not forget about that!! 😉  )

damianavila:
Btw, this is great progress @GeorgianaElena, thanks for pushing forward on it!!

GeorgianaElena:
> Totally OK with me (but let's not forget about that discussion!! )

Absolutely, @damianavila!

PR #869, opened 2021-12-03
Update events template for new event.

choldgraf:
This updates our issue template for events with some extra information that has been useful in recent iterations.


choldgraf:
I'm going to merge this in quickly, so that we can test it out for new event issues that we'll need to open for AGU 2021.

PR #872, opened 2021-12-06
Consider the case when there is no staff id list

GeorgianaElena:
Needed this for the UToronto hub since that is using azureadauthenticator.

Ref https://github.com/2i2c-org/infrastructure/pull/866


GeorgianaElena:
Merging this since without it, the UToronto manual deploy doesn't work :sweat_smile:

damianavila:
Good catch! Now wondering how we can make sure "staff" is actually present on this hub config somehow so we do not have a single point of failure 🤔

PR #881, opened 2021-12-08
Add a new field asking about resource location

sgibson91:


.github/ISSUE_TEMPLATE/2_new-hub.yml
```diff
@@ -81,6 +81,15 @@ body:
     validations:
       required: true
 
+  - type: input
+    id: resource-location
+    attributes:
+      label: Preferred Location of the Cloud Resources
+      description: |
+        If you have restrictions about where in the world your resources should be located, please indicate here. **Only choose a cloud location if you have a strong opinion.** You may wish to specify a specific location if you have a dataset that exists in a particular region/zone already. If you have no preference, we'll use US-based locations.
```
GeorgianaElena, line 89:
```suggestion
        If you have restrictions about where in the world your resources should be located, please indicate here. **Only choose a cloud location if you have a strong reason to do so.**  For example, you may wish to specify a specific location if you have a dataset that exists in a particular region/zone already. If you have no preference, we'll use US-based locations.
```

GeorgianaElena (approved):
:rocket: LGTM!

PR #884, opened 2021-12-08
Offer support for enabling NFS protocol for azure storage

GeorgianaElena:


sgibson91 (approved):
Looks good to me! Thank you for pulling this together so quickly!

yuvipanda:
To use this, you've to enable the hub to use NFS as well - with config like https://github.com/2i2c-org/infrastructure/blob/095a4e80e8255eba9abd8600b38cc6029a7379ce/config/hubs/carbonplan.cluster.yaml#L40. Need to turn AzureFile mounting off here: https://github.com/2i2c-org/infrastructure/blob/095a4e80e8255eba9abd8600b38cc6029a7379ce/config/hubs/utoronto.cluster.yaml#L32

GeorgianaElena:
@yuvipanda, is it ok if we go with basehub template defaults?
i.e., AzureFile disabled and nfs with this config:
https://github.com/2i2c-org/infrastructure/blob/095a4e80e8255eba9abd8600b38cc6029a7379ce/hub-templates/basehub/values.yaml#L14-L26

Thank you!!

yuvipanda:
@GeorgianaElena you'll need to specify the URL still, otherwise the base setup might be ok.

GeorgianaElena:
@yuvipanda, I think I'm a bit confused and would really appreciate some guidance.

1. "the URL" means `serverIP`, right? But how do I know the address of the nfs server if there isn't any right now?
2. what happens with the `home-azurefile` volume when I deploy these changes? Do I need to manually clean it up?

yuvipanda:
@GeorgianaElena ah, so I think you've to terraform apply to get the server IP, and then add that into the config. Given there are no users on this, I think you should just apply it - the config looks ok to me. 

For (2), helm should clean that up.

GeorgianaElena:
Perfect! Thanks a lot @yuvipanda!

sgibson91:
@GeorgianaElena You can ask Terraform to output the URL of the Fileshare: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_share#attributes-reference

sgibson91 (approved):
LGTM! Thank you bearing with me in Slack! 🌺 Feel free to run tf apply and then merge this

GeorgianaElena:
Applied :rocket: 

Thanks for bearing with me too :sparkling_heart:

PR #886, opened 2021-12-09
Add config for a prod UToronto hub

GeorgianaElena:
Part of https://github.com/2i2c-org/infrastructure/issues/638

### DIFF
--- a/config/hubs/utoronto.cluster.yaml
+++ b/config/hubs/utoronto.cluster.yaml
@@ -128,3 +128,10 @@ hubs:
               login_service: "University of Toronto ID"
               oauth_callback_url: https://staging.utoronto.2i2c.cloud/hub/oauth_callback
               tenant_id: 78aac226-2f03-4b4d-9037-b46d56c55210
+  - name: prod
+    domain: utoronto.2i2c.cloud
+    template: basehub
+    auth0:
+      enabled: false
+    config: *utorontoHubConfig
+
--- a/config/hubs/utoronto.cluster.yaml
+++ b/config/hubs/utoronto.cluster.yaml
@@ -111,7 +111,7 @@ hubs:
               memory: 512Mi
             limits:
               memory: 2Gi
-          allowNamedServers: &staging_jhub_allowNamedServers
+          allowNamedServers: true
           readinessProbe: &staging_jhub_redinessProbe
             enabled: false
           config:
@@ -143,7 +143,7 @@ hubs:
         proxy: *staging_jhub_proxy
         hub:
           resources: *staging_jhub_resources
-          allowNamedServers: *staging_jhub_allowNamedServers
+          allowNamedServers: true
           readinessProbe: *staging_jhub_redinessProbe
           config:
             Authenticator: *staging_jhub_authenticator
@@ -153,4 +153,3 @@ hubs:
               login_service: "University of Toronto ID"
               oauth_callback_url: https://utoronto.2i2c.cloud/hub/oauth_callback
               tenant_id: 78aac226-2f03-4b4d-9037-b46d56c55210
-          extraConfig: *staging_jhub_extraConfig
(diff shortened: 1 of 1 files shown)


GeorgianaElena:
@sgibson91, wonder if you remember why aliasing the entire hub config wasn't enough for the pangeo prod hub. I see lots of aliases and anchors in that config and wonder if we need the same here:
https://github.com/2i2c-org/infrastructure/blob/e26c5cddd92528415792a1dbd29a9359fb237431/config/hubs/pangeo-hubs.cluster.yaml#L161-L181

sgibson91:
> wonder if you remember why aliasing the entire hub config wasn't enough for the pangeo prod hub

It's because I wanted to keep the same config except change one value. When I used a YAML formatter with an anchor for the whole config and my single value change, it didn't render correctly. While a messier approach, I was more confident that using multiple anchors would recreate the correct config. This will become easier to manage if we separate each hub helm config into it's own file as we could then use the prod config to override _parts_ of the staging config.

sgibson91 (approved):

sgibson91 (changes requested):
You will need to provide the oauth_callback_url for the AzureADOAuthenticator

GeorgianaElena:
> You will need to provide the oauth_callback_url for the AzureADOAuthenticator in the prod config

Oooh, right! This is the differing bit! Thanks a loot! :sparkles:

GeorgianaElena:
@sgibson91, I added the anchor for `oauth_callback_url` and rebased this PR. 

Wondering if deploying this prod hub and me poking around with it could interfere in any way with the rsync that's still happening :grimacing: (I might as well just be paranoid at this point)

sgibson91 (approved):
The config looks good to me, but while we're beingparanoid, you can run it through [this YAML parser](https://yaml-online-parser.appspot.com/) to double check 😄 

> Wondering if deploying this prod hub and me poking around with it could interfere in any way with the rsync that's still happening 😬 (I might as well just be paranoid at this point)

I don't think it should as the deployer only mounts the NFS, it doesn't write to it. And it can apparently cope with being mounted to 2 places at once as it's mounted to the VM running the `rsync` command. However, if you are really nervous. You could probably deploy this to prod on Monday morning before running the back-up `rsync` and switching the DNS records. Have you verified that this works on staging?

GeorgianaElena:
Didn't deploy this on staging yet either, but will do so before deploying to prod (probably on Monday), but not before checking the config on the yaml parser :face_in_clouds: :sweat_smile:

Thank youuuu! :sparkles:

yuvipanda:
I opened https://github.com/GeorgianaElena/pilot-hubs/pull/1 with some changes to this.

GeorgianaElena:
I ran the config through the [YAML parser](https://yaml-online-parser.appspot.com/) again and it looks ok to me.
With @yuvipanda's changes merged, I will try a deploy to staging and see how it goes, then deploy the prod hub if everything looks alright

GeorgianaElena:
Pff, got an error:
```
Error: UPGRADE FAILED: cannot patch "hub-db-dir" with kind PersistentVolumeClaim: PersistentVolumeClaim "hub-db-dir" is invalid: spec: Forbidden: spec is immutable after creation except resources.requests for bound claims
  core.PersistentVolumeClaimSpec{
  	... // 2 identical fields
  	Resources:        {Requests: {s"storage": {i: {...}, s: "1Gi", Format: "BinarySI"}}},
  	VolumeName:       "pvc-522e0230-555b-4695-be67-f5f770e80f9e",
- 	StorageClassName: &"managed-premium",
+ 	StorageClassName: &"default",
  	VolumeMode:       &"Filesystem",
  	DataSource:       nil,
  }

```

GeorgianaElena:
Update:

- Deleting the pv, pvc and the hub pod (where the pv was mounted), then redeploying solved the issue!
- The staging and prod hub have been successfully deployed with the changes in this PR, so I'll go ahead and merge it now.

GeorgianaElena:
Thanks @sgibson91 and @yuvipanda!

Issue #876, opened 2021-12-07
Copy UToronto home dirs into the new cluster


GeorgianaElena:
Once we have the staging hub set up and confirmed it works as-expected, start copying the home directories from the NFS setup they currently live in.

- [x] We should mount the Azure File system in the NFS VM, and use rsync to copy

GeorgianaElena:
> We should mount the Azure File system in the NFS VM, and use rsync to copy

@yuvipanda, wondering if you know if there are any Azure docs about how to do this. I know we have our own docs about migrating data from one hub to another here https://infrastructure.2i2c.org/en/latest/howto/operate/move-hub.html and probably this migration to AzureFiles would be a combination of those steps

sgibson91:
I'd see if these help

- https://docs.microsoft.com/en-us/azure/storage/files/storage-how-to-use-files-linux?tabs=smb311
- https://docs.microsoft.com/en-us/azure/storage/files/storage-files-how-to-mount-nfs-shares

yuvipanda:
I've done the following:

1. Grant access to our new AzureFile from the existing VNet the NFS VM is in using the [azure portal](https://portal.azure.com/#blade/Microsoft_Azure_FileStorage/FileShareMenuBlade/overview/storageAccountId/%2Fsubscriptions%2Fead3521a-d994-4a44-a68d-b16e35642d5b%2FresourceGroups%2F2i2c-utoronto-cluster%2Fproviders%2FMicrosoft.Storage%2FstorageAccounts%2F2i2cutorontohubstorage/path/homes/protocol/NFS) ot facilitate mounting the AzureFile in the NFS VM, so we can copy them over. Documentation on how to do that is in https://docs.microsoft.com/en-us/azure/storage/files/storage-files-networking-endpoints?tabs=azure-portal#restrict-public-endpoint-access.
2. I've mounted the AzureFile onto the NFS VM with the following commands, run inside the NFS VM (copy pasted [from the portal](https://portal.azure.com/#blade/Microsoft_Azure_FileStorage/FileShareMenuBlade/overview/storageAccountId/%2Fsubscriptions%2Fead3521a-d994-4a44-a68d-b16e35642d5b%2FresourceGroups%2F2i2c-utoronto-cluster%2Fproviders%2FMicrosoft.Storage%2FstorageAccounts%2F2i2cutorontohubstorage/path/homes/protocol/NFS)):
    ```bash
     sudo mkdir -p /mnt/new-nfs
[...]

PR #889, opened 2021-12-10
Add docs on how to move data from an NFS server to AzureFile NFS storage

GeorgianaElena:
Docs for https://github.com/2i2c-org/infrastructure/issues/876

### DIFF
--- a/docs/howto/operate/move-hub.md
+++ b/docs/howto/operate/move-hub.md
@@ -57,6 +57,10 @@ The trailing slashes are important to copy the contents of the directory, withou
 
 See the [`rsync` man page](https://ss64.com/bash/rsync.html) to understand these options.
 
+```{note}
+For long-running tasks, especially those ran on a remote machine, we recommend using [`screen`](https://www.gnu.org/software/screen/manual/screen.html). Screen is an utility that allows starting `screen` sessions that can then be put to run in the background with the possiblity of re-attaching to them. Processes running in `screen` will continue to run in the background even if you get disconnected.
+```
+
 ### GCP Filestores
 
 We also use GCP Filestores as in-cluster NFS storage and can transfer the home directories between them in a similar fashion to the NFS servers described above.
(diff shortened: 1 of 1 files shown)


docs/howto/operate/move-hub.md
```diff
@@ -106,6 +110,53 @@ do your modifications. This prevents needing to create another EC2
 instance just for this.
 ```
 
+### Azure Files
+
+```{note}
+The exception: even if the source and target cluster are in the same Azure subscription, re-using the same home directory storage might prove more difficult than copying files from one storage to the other.
```
damianavila, line 116: This note just at the beginning of the Azure Files section feels a little bit weird... I mean, it seems the result of some previous paragraph discussing something and that paragraph does not exist above, AFAIK.
I am missing something, maybe?

docs/howto/operate/move-hub.md
```diff
@@ -57,6 +57,10 @@ The trailing slashes are important to copy the contents of the directory, withou
 
 See the [`rsync` man page](https://ss64.com/bash/rsync.html) to understand these options.
 
+```{note}
+For long-running tasks, especially those ran on a remote machine, we recommend using [`screen`](https://www.gnu.org/software/screen/manual/screen.html). Screen is an utility that allows starting `screen` sessions that can then be put to run in the background with the possiblity of re-attaching to them. Processes running in `screen` will continue to run in the background even if you get disconnected.
```
sgibson91, line 61:
```suggestion
For long-running tasks, especially those running on a remote machine, we recommend using [`screen`](https://www.gnu.org/software/screen/manual/screen.html). Screen is a utility that allows starting `screen` sessions that can then be put to run in the background with the possibility of re-attaching to them. Processes running in `screen` will continue to run in the background even if you get disconnected.
```

docs/howto/operate/move-hub.md
sgibson91: Yeah, I'd maybe rephrase it as "You may be tempted to attach an existing NFS server across two clusters. However, this is not possible since Azure VMs cannot simultaneously exist in two virtual networks and each cluster will have it's own network."

docs/howto/operate/move-hub.md
```diff
@@ -106,6 +110,53 @@ do your modifications. This prevents needing to create another EC2
 instance just for this.
 ```
 
+### Azure Files
+
+```{note}
+The exception: even if the source and target cluster are in the same Azure subscription, re-using the same home directory storage might prove more difficult than copying files from one storage to the other.
+```
+
+We also use AzureFiles as in-cluster NFS storage. These are the steps to transfer the home directories between a NFS server located in another cluster, but in the same Azure subscription with an AzureFile NFS storage.
```
sgibson91, line 119:
```suggestion
We also use AzureFiles as in-cluster NFS storage. These are the steps to transfer the home directories from a NFS server located in another cluster, but in the same Azure subscription, into AzureFile NFS storage.
```

docs/howto/operate/move-hub.md
```diff
@@ -106,6 +110,53 @@ do your modifications. This prevents needing to create another EC2
...
+
+AzureFile needs to be mounted in the source NFS VM in order to copy the data.
+
+1. Make sure AzureFile has access to the VNet in which the source NFS VM is in. Documentation on how to do that is can be found [here](https://docs.microsoft.com/en-us/azure/storage/files/storage-files-networking-endpoints?tabs=azure-portal#restrict-public-endpoint-access).
+
+2. On the source NFS VM:
+   1. Locate the private ssh key counterpart of the public one set for this cluster. The private key is usually encrypted with sops and stored into the repository where the infrastructure config of the cluster resides.
```
sgibson91, line 126: General note that shouldn't block this PR.

We should update our Azure cluster creation docs to include a step that creates a new SSH key pair and uploads the sops-encrypted, private part to the `secrets` directory. Atm, I think Georgiana and I are using our own ssh keys which is unsustainable.

docs/howto/operate/move-hub.md
```diff
@@ -106,6 +110,53 @@ do your modifications. This prevents needing to create another EC2
...
+      ```bash
+      sops -d secrets/ssh-key > secrets/ssh-key.unsafe
+      ```
+   2. Make sure `kubectl` is set to talk to the Azure cluster, that host the source VM, as we'll be 'hopping' through it to access the NFS VM.
```
sgibson91, line 130:
```suggestion
   2. Make sure `kubectl` is authenticated with the Azure cluster that hosts the source VM, as we'll be 'hopping' through it to access the NFS VM.
```

docs/howto/operate/move-hub.md
```diff
@@ -106,6 +110,53 @@ do your modifications. This prevents needing to create another EC2
...
+      ```bash
+      kubectl config use-context <the-desired-context>
```
sgibson91, line 132: You may wish to add a note here that suggests using `kubectl config get-contexts` to list which contexts are available. Also, if the desired context isn't available, you can authenticate using `az aks get-credentials --name CLUSTER_NAME --resource-group RESOURCE_GROUP_NAME`

docs/howto/operate/move-hub.md
```diff
@@ -106,6 +110,53 @@ do your modifications. This prevents needing to create another EC2
...
+      ```
+   3. Ssh into the source NFS VM using the [`./terraform/proxycommand.py`](https://github.com/2i2c-org/infrastructure/blob/master/terraform/azure/proxycommand.py) script, passing it the private key from step 1, the authorized user to connect to the VM using this key pair (this is usually `hubadmin` or `hub-admin`, and can be found in the terraform config `azurerm_kubernetes_cluster.linux_profile.admin_username`), and the address of the NFS VM.
```
sgibson91, line 134:
```suggestion
   3. Ssh into the source NFS VM using the [`./terraform/proxycommand.py`](https://github.com/2i2c-org/infrastructure/blob/master/terraform/azure/proxycommand.py) script, passing it the private key from step 1, the authorized user to connect to the VM using this key pair (this is usually `hubadmin` or `hub-admin`, and can be found in the terraform config `azurerm_kubernetes_cluster.linux_profile.admin_username`), and the address of the NFS VM.
```

Can we turn the `azurerm_***` part into a link please?

sgibson91 (approved):
Thanks @GeorgianaElena, this is really great! I just have a few comments :)

docs/howto/operate/move-hub.md
GeorgianaElena: Great rephrasing @sgibson91, thanks! I will use it, if that's ok with you :D 

> This note just at the beginning of the Azure Files section feels a little bit weird... I mean, it seems the result of some previous paragraph discussing something and that paragraph does not exist above, AFAIK.
I am missing something, maybe?

@damianavila, I was refering to the paragraph [here](https://github.com/2i2c-org/infrastructure/blob/master/docs/howto/operate/move-hub.md#copy-home-directories):
*"This might not entirely be necessary - if the source and target cluster
are in the same GCP Project / AWS Account, we can just re-use the same
home directory storage!"*

But now I see I didn't do a good job capturing that idea :sweat_smile: I will go with Sarah's rephrasing, since that's much better.

Thank you! :sparkles:

docs/howto/operate/move-hub.md
GeorgianaElena: > Atm, I think Georgiana and I are using our own ssh keys which is unsustainable.

Yessss! I guess we could still upload them to the `secrets` dir and generate new personal ones to fix this mhm... (But that kind of has consequences :grimacing: )

*I think* that if we change them now in terraform with a it will force a resource recreation (as per terraform's docs) of the cluster.

Absolutely  + 1 on documenting this!!!

sgibson91: Tbh, right now there's more than just the SSH keys blocking other team members from working on the UToronto and CarbonPlan clusters 😅 so I wouldn't worry too much about those specific clusters now and instead focus on not generating future toil

GeorgianaElena:
Thanks @damianavila and @sgibson91 for reviewing this! I hope I've addressed everything!
Since I have one approval, I will merge this now :rocket:

docs/howto/operate/move-hub.md
GeorgianaElena: Done! Thank you! :sparkles:

PR #894, opened 2021-12-13
Add grafana token and prod hub AzureAd client creds

GeorgianaElena:
This adds the grafana token needed to deploy JupyterHub dashboards and copies the AzureAd client credentials from staging


GeorgianaElena:
Dashboards deployed :tada:

PR #895, opened 2021-12-13
Use correct domain for UToronto prod hub

yuvipanda:
This is a CNAME to utoronto.2i2c.cloud. So we just set the
DNS for utoronto.2i2c.cloud, but *HTTPS* will work with the
new domain


GeorgianaElena:
Thanks @yuvipanda! Merging this, then I will deploy!

GeorgianaElena (approved):

PR #897, opened 2021-12-13
Support pulling from private git repos

yuvipanda:
You can test this workflow with the procedure in
https://github.com/utoronto-2i2c/jupyterhub-deploy/#pulling-from-private-github-repos-with-nbgitpuller

You can test it with this link:
https://staging.utoronto.2i2c.cloud/hub/user-redirect/git-pull?repo=https%3A%2F%2Fgithub.com%2Futoronto-2i2c%2Fprivate-test-1&branch=main&urlpath=tree%2Fprivate-test-1%2FREADME.md

Fixes https://github.com/2i2c-org/infrastructure/issues/851

PR #898, opened 2021-12-13
Add a comment on where app-id comes from

yuvipanda:
Just the commit I thought I had pushed (and gotten approved) in
https://github.com/2i2c-org/infrastructure/pull/897#issuecomment-992726535


GeorgianaElena (approved):
:rocket:

PR #904, opened 2021-12-16
Add more info about hub migration between clusters

GeorgianaElena:
It also ~links~ copies @yuvipanda's script for parallel `rsync` from https://github.com/yuvipanda/datahub/blob/cache/scripts/rsync-active-users.py to `extra-scripts`, but it definetly doesn't do it justice and we need more exposure and info about it. 

Reference: https://github.com/2i2c-org/infrastructure/issues/638


docs/howto/operate/move-hub.md
```diff
@@ -179,7 +179,28 @@ AzureFile needs to be mounted in the source NFS VM in order to copy the data.
...
-## Transfer the JupyterHub Database
+```{note}
+If the total size of the home directories is considerable, then copying the files from one cluster to another might take a long time. So make sure you have enough time to perform this operation and check the trasnfer rates once the data transfer starts.
+
+Tip: You can use [this script](https://github.com/yuvipanda/datahub/blob/cache/scripts/rsync-active-users.py) that performs a parallel `rsync` of home directories for active users only.
+```
+
+
+## 3. Set up Grafana Dashboards for the new cluster
+Make sure the new cluster has Grafana Dashboards deployed. If not, follow the steps in [](./grafana.md:new-grafana). Also, verify if the old cluster had Prometheus deployed and whether you also need to migrate that.
```
GeorgianaElena, line 190: I don't think I'm using :arrow_right: `[](./grafana.md:new-grafana)` correctly

sgibson91, line 190:
```suggestion
Make sure the new cluster has Grafana Dashboards deployed. If not, follow the steps in {ref}`grafana:new-grafana`. Also, verify if the old cluster had Prometheus deployed and whether you also need to migrate that.
```

I think this syntax will work

docs/howto/operate/move-hub.md
```diff
@@ -179,7 +179,28 @@ AzureFile needs to be mounted in the source NFS VM in order to copy the data.
    <your_scp_or_rsync_command> <homes-directory-location-on-the-source-nfs-server> /mnt/new-nfs/prod/
    ```
 
-## Transfer the JupyterHub Database
+```{note}
+If the total size of the home directories is considerable, then copying the files from one cluster to another might take a long time. So make sure you have enough time to perform this operation and check the trasnfer rates once the data transfer starts.
```
sgibson91, line 183:
```suggestion
If the total size of the home directories is considerable, then copying the files from one cluster to another might take a long time. So make sure you have enough time to perform this operation and check the transfer rates once the data transfer starts.
```

docs/howto/operate/move-hub.md
```diff
@@ -179,7 +179,28 @@ AzureFile needs to be mounted in the source NFS VM in order to copy the data.
...
+
+Tip: You can use [this script](https://github.com/yuvipanda/datahub/blob/cache/scripts/rsync-active-users.py) that performs a parallel `rsync` of home directories for active users only.
```
sgibson91, line 185: I wonder if we should move this script in out [`extra_scripts`](https://github.com/2i2c-org/infrastructure/tree/master/extra_scripts) directory?

sgibson91 (approved):
LGTM!

docs/howto/operate/move-hub.md
```diff
@@ -200,6 +221,27 @@ This step preserves user information, since they might be added via the admin UI
...
+```bash
+kubectl --namespace support get svc support-ingress-nginx-controller
+```
+
+Edit the existing DNS entry in NameCheap that matches the old hub domain and type in this external IP address.
+
+This will move DNS from old cluster to new cluster, thus completing the move.
 
-Move DNS from old cluster to new cluster, thus completing the move.
+## 9. Cleanup the old cluster and NFS VM, preserving anything we want from there
```
damianavila, line 247: >preserving anything we want from there

such as?? (I guess some examples could be useful here, maybe some specific logs?

GeorgianaElena:
Thanks everyone for their feedback! I believe I've address all the comments and since there's one approval I will merge this now :rocket:

docs/howto/operate/move-hub.md
damianavila: @GeorgianaElena sorry for the belated comment, but that link should be eventually updated to point to the script living now in the `extra_script` directory, right?

GeorgianaElena: You're totally right @damianavila! I forgot to update the link too.

> @GeorgianaElena sorry for the belated comment

Better late than forget about it! Thank you @damianavila :pray: Will open a PR to update it now

PR #906, opened 2021-12-24
Block access to howard cloudbank hub

yuvipanda:
The instructor wants the students to not be able to log in,
while preserving the ability to use the admin panel to access
students' work. By moving the hub to a randomly generated
domain name that is kept secret, the instructor can log in but
students can not.

Longer term, we need better user profile support in JupyterHub
for this. Perhaps the work in JupyterHub 2.0 with groups
will help.

Ref https://2i2c.freshdesk.com/a/tickets/61

### DIFF
--- a/config/hubs/cloudbank.cluster.yaml
+++ b/config/hubs/cloudbank.cluster.yaml
@@ -148,7 +148,6 @@ hubs:
           config:
             Authenticator:
               allowed_users: &howard_users
-                - <staff_google_ids>
                 - ericvd@berkeley.edu
                 - gwashington@scs.howard.edu
                 - anthony.fgordon64@gmail.com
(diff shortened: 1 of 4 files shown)


yuvipanda:
I had privately deployed this already to make sure we don't give away this impending removal to any students who might be watching this repo. I'm also going to merge this now, because deploying a *different* PR will restore the current howard domain unintentionally.

config/hubs/cloudbank.cluster.yaml
```diff
@@ -148,7 +148,6 @@ hubs:
           config:
             Authenticator:
               allowed_users: &howard_users
-                - <staff_google_ids>
```
damianavila, line 151: @yuvipanda, any reason to remove the staff from this hub?

PR #885, opened 2021-12-09
Automate and document setting up eksctl clusters

yuvipanda:
# Part 1: Move openscapes cluster from kops to eksctl

- Deploy support chart to this cluster, so we now have a working
  grafana for this cluster!
- Upgraded k8s to 1.21 from 1.19 that was used in kops
- Reduce single user memory guarantees to match what seems to be
  available now in Kubernetes 1.21
- Create a new IAM User and use it for infrastructure deployer.
  Access to the k8s user needs to be explicitly granted with
  `eksctl create iamidentitymapping --cluster  openscapeshub --arn arn:aws:iam::783616723547:user/infrastructure-auto-deployer --group system:masters --username infrastructure-auto-deployer --region=us-west-2`
  Otherwise, just EKS:Read permission is needed.
- EFS has been moved to the new VPC eksctl created, and appropriate
  security group has been added to it.

# Part 2: Automate and document setting up eksctl clusters

- Rewrite cluster documentation for AWS to use eksctl, remove older piece of AWS specific ci/cd
- Use Terraform to setup IAM user for CI/CD
- Use Terraform to setup EFS, along with proper network access rules there
- Import manually created EFS for openscapes into terraform with 
   `terraform import`, so it is fully managed by terraform now! I had to recreate
   the mount target, but that happened without disruption. Yay. So this means 
   the entire openscapes infra (other than EFS) is now managed by terraform!
- New AWS IAM creds checked into this repo, generated by terraform
- CarbonPlan also moved into being managed by terraform

TODO:
- [x] Document how to create a new cluster with eksctl
- [x] Document how to setup EFS for the new cluster
- [x] Document how to setup CI/CD user account for new cluster
- [x] Import Carbonplan into Terraform also

PR #912, opened 2022-01-03
Revert "Stop auto-deploying openscapes"

yuvipanda:
Reverts 2i2c-org/infrastructure#911

Now that https://github.com/2i2c-org/infrastructure/pull/885 is merged.

### DIFF
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -37,6 +37,8 @@ jobs:
             provider: aws
           - cluster_name: farallon
             provider: aws
+          - cluster_name: openscapes
+            provider: aws
           - cluster_name: meom-ige
             provider: gcp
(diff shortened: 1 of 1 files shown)


GeorgianaElena (approved):
:rocket:  Thanks for fixing this @yuvipanda!

choldgraf:
Is there a reason that we haven't merged this one?

yuvipanda:
@choldgraf I got back to work only today, hence :) I also don't think this PR will actually do anything, as no files specific to the deployment have been touched. So another PR will need to be done to test if this actually deploys.

yuvipanda:
This is now tested in https://github.com/2i2c-org/infrastructure/pull/915

yuvipanda:
Aha, I was wrong - the hubs are redeployed whenever github workflow changes! So this is being tested without needing #915.

PR #914, opened 2022-01-03
DOCS: Ignore howard link in docs

choldgraf:
Temporarily skips the link check for Howard since we've changed the hub name

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -62,7 +62,7 @@ html_static_path = ["_static"]
 # Disable linkcheck for anchors because it throws false errors for any JS anchors
 linkcheck_anchors = False
 linkcheck_ignore = [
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -61,6 +61,9 @@
 html_static_path = ["_static"]
 # Disable linkcheck for anchors because it throws false errors for any JS anchors
 linkcheck_anchors = False
+linkcheck_ignore = [
+    "howard.cloudbank.2i2c.cloud*",  # Temporarily ignore because we've changed the hub name
```
choldgraf, line 65:
```suggestion
    "https://howard.cloudbank.2i2c.cloud*",  # Temporarily ignore because we've changed the hub name
```

PR #920, opened 2022-01-05
Setup bioconductor for UToronto

yuvipanda:
Ref https://2i2c.freshdesk.com/a/tickets/60


yuvipanda:
Thanks for making the connection, @GeorgianaElena!

PR #924, opened 2022-01-11
Update link to parallel rsync script

GeorgianaElena:
Thanks @damianavila for spotting this :rocket:


GeorgianaElena:
I think it makes sense to merge this without an approval, since it's just a small update. So merging this now :rocket:

PR #922, opened 2022-01-06
Move pilot-hubs user image to its own repo

GeorgianaElena:
### This PR:
- removes the user image files
- removes the build script and references to it
- updates docs on how to update default user environment

### Not in this PR:
- New pilot-hubs-image repository that uses repo2docker action is :arrow_right: ~https://github.com/2i2c-org/pilot-hubs-image~ https://github.com/2i2c-org/2i2c-hubs-image
- Images are pushed to this quay.io registry :arrow_right: ~https://quay.io/repository/2i2c/pilot-hubs-image?tab=info~ https://quay.io/repository/2i2c/2i2c-hubs-image?tab=info

Closes https://github.com/2i2c-org/infrastructure/issues/710

PR #925, opened 2022-01-11
Use the right registry repo for the user image

GeorgianaElena:
When testing out https://github.com/2i2c-org/infrastructure/pull/922 on staging, this was the log when I tried to start a server:

```
2022-01-11T16:09:55Z [Normal] Pulling image "us-central1-docker.pkg.dev/two-eye-two-see/pilot-hubs-registry/base-user:532c5eab47a1"
2022-01-11T16:09:55Z [Warning] Failed to pull image "us-central1-docker.pkg.dev/two-eye-two-see/pilot-hubs-registry/base-user:532c5eab47a1": rpc error: code = NotFound desc = failed to pull and unpack image "us-central1-docker.pkg.dev/two-eye-two-see/pilot-hubs-registry/base-user:532c5eab47a1": failed to resolve reference "us-central1-docker.pkg.dev/two-eye-two-see/pilot-hubs-registry/base-user:532c5eab47a1": us-central1-docker.pkg.dev/two-eye-two-see/pilot-hubs-registry/base-user:532c5eab47a1: not found
```

I believe `image_repo` is overwriting what we have in the basehub template as the image repo.

### DIFF
--- a/config/hubs/2i2c.cluster.yaml
+++ b/config/hubs/2i2c.cluster.yaml
@@ -1,5 +1,5 @@
 name: 2i2c
-image_repo: "us-central1-docker.pkg.dev/two-eye-two-see/pilot-hubs-registry/base-user"
+image_repo: "quay.io/2i2c/2i2c-hubs-image"
 provider: gcp
 gcp:
   key: secrets/2i2c.json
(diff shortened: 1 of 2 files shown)


GeorgianaElena:
Actually, maybe we should not use `image_repo` at all and leave it be set only in the template https://github.com/2i2c-org/infrastructure/blob/master/hub-templates/basehub/values.yaml#L136

GeorgianaElena:
I'll merge this in without approval I thing, just to fix the issue. But feel free to provide feedback about where would be best to set the image repo.

choldgraf:
@GeorgianaElena thanks for merging this one in - I think if we're putting out an active fire, and the fix doesn't seem too nasty and complex, then you should feel empowered to put out the fire :-)

PR #932, opened 2022-01-17
Support Chart: Add daemonset to install GPU drivers to nodes

sgibson91:
This PR adds a template file to the support chart that creates a daemonset for installing the Nvidia device plugin for GPU nodes and places it behind a flag. The config is taken from these Azure docs: https://docs.microsoft.com/en-us/azure/aks/gpu-cluster#manually-install-the-nvidia-device-plugin

I have also updated CarbonPlan's Azure config to enable this in their support namespace, but haven't tested it yet.

### Update

- **2022-01-17 16:50 GMT:** I have deployed the current state of this PR to the support namespace successfully. Now I intend on providing the GPU option in JupyterHub's interface and test/debug from there.


support/templates/nvidia-device-plugin.yaml
```diff
@@ -0,0 +1,50 @@
+{{ if .Values.config.nvidiaDevicePlugin.enabled }}
```
yuvipanda, line 1: maybe nvidiaDevicePlugin.azure.enabled? This is using what looks like an Azure specific image.

sgibson91: Done!

sgibson91: I renamed the template file too, to be extra clear

support/templates/nvidia-device-plugin.yaml
```diff
@@ -0,0 +1,50 @@
+{{ if .Values.config.nvidiaDevicePlugin.azure.enabled }}
```
consideRatio:
```suggestion
{{- if .Values.config.nvidiaDevicePlugin.azure.enabled -}}
```

This is not at all important change, in practice it will mean this file won't start out with a blank line which is non-problematic, but I find it useful to be consistent with a practice so one doesn't end up needing to know when it will be or wont be important.

My rule of thumb is to:

1. Always chomp the whitespace to the left
2. When its the first content of a yaml file, or a named template defined in a _helpers.tpl file, also chomp right

config/hubs/azure.carbonplan.cluster.yaml
```diff
@@ -4,6 +4,9 @@ kubeconfig:
   file: secrets/azure.carbonplan.yaml
 support:
   config:
+    nvidiaDevicePlugin:
+      azure:
+        enabled: true
```
consideRatio, line 9: If you add a configuration value for a specific cluster, you should probably have a default value as well so that `{{ if .Values.config.nvidiaDevicePlugin.azure.enabled }}` won't error by having the object `.Values.config.nvidiaDevicePlugin` be null and then saying "can't read azure of null object" or similar.

support/templates/nvidia-device-plugin.yaml
sgibson91: Is "chomp" the hyphen? I don't know what this word means in this context! (I agree with the suggestion though)

config/hubs/azure.carbonplan.cluster.yaml
sgibson91: Think I have done this now

consideRatio:
### k8s-device-plugin vs driver-installer

I don't think this is a plugin to install a driver, but you need another DaemonSet for that. You typically need two daemonsets. One makes sure that a node starting up with a GPU is exposing the GPU and couples with k8s to have a GPU or similar, perhaps also ensure the node isn't considered ready until things are setup. Another daemonset typically installs drivers on the node, and you can choose what driver version etc as well.

For GPU setup to work, I think you will need both daemonsets.

While I think its quite clear that the driver installer installs the graphics card driver, it would be great to have clarity on what specifically the device plugin does and not just vaguely understand it.

I have used https://github.com/NVIDIA/k8s-device-plugin in GKE and AWS clusters, but I have no GPU experience on Azure. I wonder if you really need the Azure specific thing or can work with the nvidia-general thing.

Googling some I found these nvidia-driver-installer links:
- https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/nvidia-driver-installer/ubuntu/daemonset.yaml

### Related - eksctl installation

See https://github.com/2i2c-org/infrastructure/pull/436/files#diff-2913cb40ae1ad80f097b0f052b51771f17cde8d998b99b6bb924e8e0b18bf36fR232-R302

### About tolerating dedicated user nodes

You want the device plugin to run also on the dedicated user nodes if the cluster has such. To do that, you need to have a toleration for it. I'll make a code suggestion about this.

support/templates/azure-nvidia-device-plugin.yaml
```diff
@@ -0,0 +1,50 @@
...
+      # This, along with the annotation above marks this pod as a critical add-on.
+      - key: CriticalAddonsOnly
+        operator: Exists
+      - key: nvidia.com/gpu
+        operator: Exists
+        effect: NoSchedule
+      - key: "sku"
+        operator: "Equal"
+        value: "gpu"
+        effect: "NoSchedule"
```
consideRatio, line 35:
```suggestion
        effect: "NoSchedule"
      - effect: NoSchedule
        key: hub.jupyter.org/dedicated
        operator: Equal
        value: user
      - effect: NoSchedule
        key: hub.jupyter.org_dedicated
        operator: Equal
        value: user
```

support/templates/nvidia-device-plugin.yaml
consideRatio: Yepp, it means that whitespace left or right within the helm template (but not whitespace generated by the helm template) should be trimmed on the left or right side.

Chomping/trimming left doesn't do anything here since there were no whitespace before `{{- if`, but trimming right will make sure that the whitespace (a new line) following `...enabled -}}` is removed, so that the rendered k8s resource manifest starts with `# ...` rather than `(new line)# ...` for example.

For some helm documentation about this, see https://helm.sh/docs/chart_template_guide/control_structures/

sgibson91: Thanks! I applied your suggestion in [`49082fb`](https://github.com/2i2c-org/infrastructure/pull/932/commits/49082fbd0beaaf0d097e84f53b027b129c322790)

config/hubs/azure.carbonplan.cluster.yaml
consideRatio: Yepp! Is there also some schema you need/should update as part of this? I don't remember if there was one schema for the values you can have for the the deployer scripts config.

sgibson91: No the schema only goes as deep as the `config` key https://github.com/2i2c-org/infrastructure/blob/c7bbde9b03288b99835225c4da6929fe98b94b4f/config/hubs/schema.yaml#L13-L23

consideRatio:
Nice @sgibson91! I think the only concrete action point I have is to update a deployer script configuration schema if there were such in place for what configuration is allowed to pass to it.

Oh... wait... You added something to the support chart configuration, not the deployer script configuration... I assume there isn't a schema for that helm chart? Then this LGTM.

config/hubs/azure.carbonplan.cluster.yaml
consideRatio: Nevermind, you are updating the support charts input values, but that chart doesn't have a schema (even though helm charts can have schemas to be enforced by `helm` btw). I was thinking you added configuration for the deployer scripts configuration itself, which I recall had a schema defined.

consideRatio (approved):

sgibson91:
After fixing a small typo, this deployed successfully to the support namespace.

Now what I think I'll do is the following:

- Hold off on merging this for now
- Start a new PR to add the GPU option to JupyterHub
- Deploy that to staging
- Test and debug from there

Issue #930, opened 2022-01-15
Enable GPUs on the Carbon Plan hub


choldgraf:
### Description of problem and opportunity to address it

Joe with the Carbon Plan hub asked if we could enable GPUs on their infrastructure. Here are some guidelines that they provided:

- Ideal timeline: Operational in the next 2 weeks
- Constraints: 1 T4 gpu per user pod. The Standard_NC16as_T4_v3 looks like it would work well for us.

### Implementation guide and constraints

Their profile list is here:

https://github.com/2i2c-org/infrastructure/blob/943cae3286aad954b90d45438530d6023bd3d03c/config/hubs/carbonplan.cluster.yaml#L90-L101

Though this might require some change on the cluster itself to enable these machines first? I believe that @sgibson91 set this one up so maybe she could advise on what needs to happen.

sgibson91:
I have updated the top comment with next steps. I'm the only one with direct access to the cluster so I'll have to do the terraform part.

sgibson91:
After we changed some network setting to allow AzureFile to be used as a NFS server in https://github.com/2i2c-org/infrastructure/pull/887, I was no longer able to edit the cluster due to the below error (also documented in https://github.com/2i2c-org/infrastructure/issues/890)

> $ tf plan -var-file=projects/carbonplan.tfvars -out=carbonplan -refresh-only
> 
> [snip]
> 
> Error: shares.Client#GetProperties: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code="AuthorizationFailure" Message="This request is not authorized to perform this operation.\nRequestId:a7c9006d-d01a-004f-5e8d-0be45f000000\nTime:2022-01-17T10:34:16.3962556Z"
> 
>   with azurerm_storage_share.homes,
>   on storage.tf line 21, in resource "azurerm_storage_share" "homes":
>   21: resource "azurerm_storage_share" "homes" {

@yuvipanda @GeorgianaElena did you ever fix this on the UToronto cluster?

I'm blocked moving forward on this issue until I regain access to the infrastructure via terraform

sgibson91:
Temporary fix and link to upstream work that will provide the _actual_ here: https://github.com/2i2c-org/infrastructure/issues/890#issuecomment-1014404932

sgibson91:
Added an extra TODO item here around installing GPU drivers. See https://github.com/2i2c-org/infrastructure/pull/931#issuecomment-1014459426

[...]

PR #933, opened 2022-01-17
Add a profile for GPUs to CarbonPlan's Azure hub

sgibson91:
This PR will complete https://github.com/2i2c-org/infrastructure/issues/930


config/hubs/azure.carbonplan.cluster.yaml
```diff
@@ -126,6 +126,13 @@ hubs:
                   mem_guarantee: 1950G
                   node_selector:
                     hub.jupyter.org/node-size: Standard_M128s_v2
+              - display_name: "GPU: NC16as T4 v3"
+                description: "~16 CPU, ~90G RAM"
+                kubespawner_override:
+                  mem_limit: null
+                  mem_guarantee: 90G
+                  node_selector:
+                    hub.jupyter.org/node-size: Standard_NC16as_T4_v3
```
jhamman, line 135: this will also need a different docker image. For now, you can set this as `docker pull pangeo/ml-notebook:master`. In parallel, we'll work on a second carbonplan image for our application.

sgibson91: Just a note on this. The configurator doesn't (yet) support updating images listed in these profiles, so to update image tags for this profile would involve opening a PR to this repo.

sgibson91:
@jhamman I'm struggling to create a GPU node from the profile list, I wonder if we need to increase limits in the location again like we did for the M-series machines?

jhamman:
> @jhamman I'm struggling to create a GPU node from the profile list, I wonder if we need to increase limits in the location again like we did for the M-series machines?

I've submitted a support ticket to sort this out. @sgibson91, I think I've cc'd you on the ticket so hopefully you see the resolution when it comes.

sgibson91:
> I think I've cc'd you on the ticket so hopefully you see the resolution when it comes.

I got it, thanks!

sgibson91:
@jhamman I got the email notification that the quota bump was accepted and applied, but I still cannot spin up a GPU node. I get the following error:

```
2022-01-21T10:54:58Z [Normal] pod didn't trigger scale-up: 7 node(s) didn't match Pod's node affinity, 7 node(s) had taint {k8s.dask.org_dedicated: worker}, that the pod didn't tolerate, 1 in backoff after failed scale-up
```

This may be a silly question, but just to help me narrow things down... I noticed the mention of "vCPU" in the support emails from Microsoft, and I just wanted to confirm that this does indeed include GPUs and we don't have to do something separate/different to increase that quota? If not, then I'll dig in further into what's happening on the hub side of things.

jhamman:
@sgibson91 - Thanks for taking a look. I've submitted a second ticket. My original request was for a quota increase on `Standard NCSv3 Family vCPUs`. It seems I also needed to make a request for `Standard NCASv3_T4 Family vCPUs` which I have now done. 

To your question about vCPUs vs GPUs, the quota system does not have any GPU specific categories. Instead, you request VM types that come packaged with GPUs (https://docs.microsoft.com/en-us/azure/machine-learning/how-to-manage-quotas#azure-machine-learning-compute).

> To learn more about which VM family to request a quota increase for, check out virtual machine sizes in Azure. For instance GPU VM families start with an "N" in their family name (eg. NCv3 series)

yuvipanda:
With my latest commit to this PR, and a deploy of #932, I think GPU will work. See output of `nvidia-smi`:

image

yuvipanda:
I've merged #932 so just this PR is left to merge.

Issue #938, opened 2022-01-20
Standardise our code formatting using automated tools in pre-commit, e.g., pyupgrade, black, flake8


sgibson91:
### Description of problem and opportunity to address it

**Context to understand the problem**

When working on code, I find it very beneficial to have a standardised format as I find it easier to quickly parse and understand the content. I think a standardised format also helps reviewing PRs as we can be sure the diffs presented are actual changes, not just stylistic choices. I also like the automate these kinds of formatting so that I can write code without worrying that I'm conforming to the style or not (thought I've found over the years, my natural writing style conforms more and more).

We already use [pre-commit](https://pre-commit.com/) to [auto-format our markdown and YAML files](https://github.com/2i2c-org/infrastructure/blob/master/.pre-commit-config.yaml), the latter being super important given that most of this repository is helm configuration.

I would like to propose adding [pyupgrade](https://github.com/asottile/pyupgrade), [black](https://black.readthedocs.io/en/stable/) and [flake8](https://flake8.pycqa.org/en/latest/) to our pre-commit config for autoformatting our Python files as well.

**Proposed solution**

I would like to add the below dependencies to our existing pre-commit config:

```yaml
- repo: https://github.com/asottile/pyupgrade
  rev: v2.31.0
  hooks:
  - id: pyupgrade
- repo: https://github.com/psf/black
  rev: 21.12b0
  hooks:
  - id: black
- repo: https://github.com/pycqa/flake8
  rev: 4.0.1
  hooks:
  - id: flake8
    args: ["--ignore=E501,W503"]
```

**What's the value and who would benefit**

I think this would allow us to maintain tidier code that is easier to work on as a team. Having it applied by pre-commit means that there's no additional overhead to engineers, they push their code as-is and pre-commit will format it.

@consideRatio and I would greatly benefit from implementing this standardised formatting before we take a deep dive into the deployer code as we work towards https://github.com/2i2c-org/infrastructure/issues/879 (and related issues).

We only have five open PRs at the minute and only one of them touches any Python files (and only two commented lines) so there will be a very small chance of introducing merge conflicts if we take this action soon.

### Updates and ongoing work

- [x] Gain approval from the team to implement this
- [x] https://github.com/2i2c-org/infrastructure/pull/939
- [ ] https://github.com/2i2c-org/team-compass/pull/357

consideRatio:
I'm very happy about having both `pyupgrade` and `black` in a pre-commit config!

[...]

PR #939, opened 2022-01-21
Autoformatting Python code using pre-commit hooks

sgibson91:
Enable auto-formatting of Python using pyupgrade, black and flake8 hooks in our pre-commit config.

This PR also applies the changes of those hooks

This is a first step towards https://github.com/2i2c-org/infrastructure/issues/938. Remaining todo to close that issue is to document these hooks in the team-compass

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -9,6 +9,3 @@
 # I101: Imported names are in the wrong order. Should be
 max-line-length = 88
 ignore = E, C, W, E402, F821, I100, I101, D400
-exclude =
-    .github/,
-    .cache
--- a/.flake8
+++ b/.flake8
@@ -4,10 +4,11 @@
 # W: style warnings
 # C: complexity
 # E402: module level import not at top of file
+# F821: undefined name 'c' in hub image files
 # I100: Import statements are in the wrong order
 # I101: Imported names are in the wrong order. Should be
 max-line-length = 88
-ignore = E, C, W, E402, I100, I101, D400
+ignore = E, C, W, E402, F821, I100, I101, D400
 exclude =
     .cache,
     .github
--- a/.flake8
+++ b/.flake8
@@ -6,6 +6,7 @@
 # E402: module level import not at top of file
 # I100: Import statements are in the wrong order
 # I101: Imported names are in the wrong order. Should be
+max-line-length = 88
 ignore = E, C, W, E402, I100, I101, D400
 exclude =
     .cache,
--- a/.flake8
+++ b/.flake8
@@ -4,8 +4,8 @@
 # W: style warnings
 # C: complexity
 # E402: module level import not at top of file
-# F821: undefined name 'c' in hub image files
 # I100: Import statements are in the wrong order
 # I101: Imported names are in the wrong order. Should be
 max-line-length = 88
-ignore = E, C, W, E402, F821, I100, I101, D400
+ignore = E, C, W, E402, I100, I101, D400
+builtins = c
--- a/.flake8
+++ b/.flake8
@@ -10,5 +10,5 @@
 max-line-length = 88
 ignore = E, C, W, E402, F821, I100, I101, D400
 exclude =
-    .cache,
-    .github
+    .github/,
+    .cache
--- a/.flake8
+++ b/.flake8
@@ -6,5 +6,5 @@
 # E402: module level import not at top of file
 # I100: Import statements are in the wrong order
 # I101: Imported names are in the wrong order. Should be
-ignore = E, C, W, E402, I100, I101, D400
+ignore = E, C, W, I100, I101, D400
 builtins = c
--- a/.flake8
+++ b/.flake8
@@ -6,6 +6,5 @@
 # E402: module level import not at top of file
 # I100: Import statements are in the wrong order
 # I101: Imported names are in the wrong order. Should be
-max-line-length = 88
 ignore = E, C, W, E402, I100, I101, D400
 builtins = c
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -16,30 +16,28 @@ from ruamel.yaml import YAML
 from utils import decrypt_file
 
 # Without `pure=True`, I get an exception about str / byte issues
-yaml = YAML(typ='safe', pure=True)
+yaml = YAML(typ="safe", pure=True)
 
 
 class Cluster:
     """
     A single k8s cluster we can deploy to
     """
+
     def __init__(self, spec):
         self.spec = spec
-        self.hubs = [
-            Hub(self, hub_yaml)
-            for hub_yaml in self.spec['hubs']
-        ]
-        self.support = self.spec.get('support', {})
+        self.hubs = [Hub(self, hub_yaml) for hub_yaml in self.spec["hubs"]]
+        self.support = self.spec.get("support", {})
 
     @contextmanager
     def auth(self):
-        if self.spec['provider'] == 'gcp':
+        if self.spec["provider"] == "gcp":
             yield from self.auth_gcp()
-        elif self.spec['provider'] == 'aws':
+        elif self.spec["provider"] == "aws":
             yield from self.auth_aws()
-        elif self.spec['provider'] == 'azure':
+        elif self.spec["provider"] == "azure":
             yield from self.auth_azure()
-        elif self.spec['provider'] == 'kubeconfig':
+        elif self.spec["provider"] == "kubeconfig":
             yield from self.auth_kubeconfig()
         else:
             raise ValueError(f'Provider {self.spec["provider"]} not supported')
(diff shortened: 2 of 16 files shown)


deployer/hub.py
```diff
@@ -507,32 +567,49 @@ def deploy(self, auth_provider, secret_key, skip_hub_health_test=False):
...
-                'helm', 'upgrade', '--install', '--create-namespace', '--wait',
-                '--namespace', self.spec['name'],
-                self.spec['name'], os.path.join('hub-templates', self.spec['template']),
+                "helm",
+                "upgrade",
+                "--install",
+                "--create-namespace",
+                "--wait",
+                "--namespace",
+                self.spec["name"],
```
sgibson91, line 601:
```suggestion
                f"--namespace={self.spec['name']}",
```

deployer/hub.py
```diff
@@ -507,32 +567,49 @@ def deploy(self, auth_provider, secret_key, skip_hub_health_test=False):
...
                 # we should put the config from config/hubs last.
-                '-f', generated_values_file.name,
-                '-f', values_file.name,
-                '-f', secret_values_file.name,
+                "-f",
+                generated_values_file.name,
+                "-f",
+                values_file.name,
+                "-f",
+                secret_values_file.name,
```
sgibson91, line 612:
```suggestion
                f"--values={generated_values_file.name}",
                f"--values={values_file.name}",
                f"--values={secret_values_file.name}",
```

terraform/azure/proxycommand.py
```diff
@@ -12,33 +12,44 @@
...
+            "run",
+            "--image",
+            POD_IMAGE,
+            "--command",
+            "--wait",
+            POD_NAME,
+            "--",
+            "/bin/sh",
+            "-c",
+            SCRIPT,
```
sgibson91, line 42:
```suggestion
            f"--image={POD_IMAGE}",
            "--command",
            "--wait",
            POD_NAME,
            "--",
            "/bin/sh",
            f"-c={SCRIPT}",
```

sgibson91:
I am not sure why flake8 is still failing on the `.github` folder when it is listed in the `.flake8` config file

deployer/hub.py
```diff
@@ -55,71 +53,100 @@ def ensure_docker_credhelpers(self):
...
+                "--install",
+                "--create-namespace",
+                "--namespace",
+                "cert-manager",
+                "cert-manager",
+                "jetstack/cert-manager",
+                "--version",
+                cert_manager_version,
+                "--set",
+                "installCRDs=true",
```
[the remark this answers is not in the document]
consideRatio: Yeah or just `--set=installCRDs=true` would work I think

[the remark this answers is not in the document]
consideRatio: The flag parser is typically smart enough to realize that only the first `=` is meant for the flag, and take the rest as a string.

Yepp - verified that works fine.

deployer/hub.py
```diff
@@ -55,71 +53,95 @@ def ensure_docker_credhelpers(self):
...
+                "helm",
+                "upgrade",
+                "--install",
+                "--create-namespace",
+                "--namespace=cert-manager",
+                "cert-manager",
+                "jetstack/cert-manager",
+                f"--version={cert_manager_version}",
+                "--set",
+                "installCRDs=true",
```
sgibson91, line 115:
```suggestion
                "--set=installCRDs=true",
```

.flake8
```diff
@@ -4,10 +4,11 @@
 # W: style warnings
 # C: complexity
 # E402: module level import not at top of file
+# F821: undefined name 'c' in hub image files
```
consideRatio, line 7: I think F821 can be really helpful.

I suggest instead we declare the following in the flake8 config on a line below.

```
builtins = c
```

sgibson91: Done in [`8543569`](https://github.com/2i2c-org/infrastructure/pull/939/commits/85435698922bb7262a38006f1ab37b8e5c93eff5)

sgibson91:
pre-commit is now green! 🙌🏻

.flake8
```diff
@@ -6,7 +6,6 @@
 # E402: module level import not at top of file
 # I100: Import statements are in the wrong order
 # I101: Imported names are in the wrong order. Should be
+max-line-length = 88
```
yuvipanda, line 9: I'd love for us to *not* have a max-line-length enforced by flake8. Perhaps we can copy https://github.com/jupyterhub/jupyterhub/blob/main/.flake8?

sgibson91: This is so that flake8 is compatible with blacks autoformatting, which will default to 88. I was running off this suggestion https://github.com/2i2c-org/infrastructure/issues/938#issuecomment-1017609298

terraform/azure/proxycommand.py
```diff
@@ -31,15 +31,13 @@ def delete_pod():
...
-            POD_IMAGE,
+            f"--image={POD_IMAGE}",
             "--command",
             "--wait",
             POD_NAME,
             "--",
             "/bin/sh",
-            "-c",
-            SCRIPT,
+            f"-c={SCRIPT}",
```
consideRatio, line 40: This will break. It depends on what CLI. I tried `sh -c=echo` and i got an error about it.

sgibson91: Ok, will revert

terraform/azure/proxycommand.py
```diff
@@ -12,33 +12,42 @@
...
+        [
+            "kubectl",
+            "run",
+            f"--image={POD_IMAGE}",
+            "--command",
+            "--wait",
+            POD_NAME,
+            "--",
+            "/bin/sh",
+            f"-c={SCRIPT}",
```
sgibson91, line 40:
```suggestion
            "-c",
            SCRIPT,
```

deployer/hub.py
```diff
@@ -294,10 +285,8 @@ def auth_azure(self):
                         "az",
                         "aks",
                         "get-credentials",
-                        "--name",
-                        cluster,
-                        "--resource-group",
-                        resource_group,
+                        f"--name={cluster}",
+                        f"--resource-group={resource_group}",
```
consideRatio, line 288: Do you have `az` installed? If so, can you trial if it can parse flags like `--something=value`?

I've tried it works for kubectl / helm

terraform/azure/proxycommand.py
sgibson91: Done in [`c717bb3`](https://github.com/2i2c-org/infrastructure/pull/939/commits/c717bb3e943193ab2c1a8f9b5ba767b264bf7dbb)

deployer/hub.py
sgibson91: Yep, works for `az`

.flake8
consideRatio: @yuvipanda this is to be compliant with black, flake8 will not enforce anything harder - this is a relaxation I think.

sgibson91: Unless this line means we are automatically ignoring `E501` errors irrespective of line-length?

https://github.com/2i2c-org/infrastructure/pull/939/files#diff-6951dbb399883798a226c1fb496fdb4183b1ab48865e75eddecf6ceb6cf46442R10

consideRatio: Yeah hmm...

Well since we have a very relaxed config saying....
```
# Ignore style and complexity
# E: style errors
# W: style warnings
# C: complexity
ignore = E, C, W, ...
```

The max-line-length may be entirely irrelevant, while it could be relevant if we went with a minimal config to ensure it was compatible with black. This was the example they suggested in blacks doc.

```
[flake8]
max-line-length = 88
extend-ignore = E203
```

---

Since we now have a lot of ignored things making the config relaxed, I think the line length detail is irrelevant so +1 for removing it unless we tighten up the flake8 config.

sgibson91: I think we should remove `max-line-length` and keep the other ignores we currently have. Will do that now.

consideRatio: @sgibson91 I think you can also remove E402 since its included in E btw

sgibson91: Done and done!

yuvipanda: Yay! Thank you for the explanation too :)

consideRatio (approved):
LGTM!

deployer/hub.py
```diff
@@ -149,147 +170,176 @@ def auth_aws(self):
...
                     with open(decrypted_key_abspath) as f:
                         creds = json.load(f)
 
-                    os.environ['AWS_ACCESS_KEY_ID'] = creds["AccessKey"]['AccessKeyId']
-                    os.environ['AWS_SECRET_ACCESS_KEY'] = creds["AccessKey"]['SecretAccessKey']
-
-                os.environ['KUBECONFIG'] = kubeconfig.name
+                    os.environ["AWS_ACCESS_KEY_ID"] = creds["AccessKey"]["AccessKeyId"]
+                    os.environ["AWS_SECRET_ACCESS_KEY"] = creds["AccessKey"][
+                        "SecretAccessKey"
```
damianavila, line 197: This one is weird... but I guess I need to let it go 😜

damianavila (approved):
LGTM, there are things that look weird to me but I know it is a totally subjective view 😉

deployer/hub.py
consideRatio: Yikes, yeah, that was a ugly outcome.

sgibson91: Yeah, it's not always perfect, but I guess this what we sacrifice for automation. At least you never have to _write_ it that way 😝

choldgraf (approved):
This looks good to me - agreed some of the outcomes are a bit awkward, but such is a life of `black`. I'm +1 on merging this before it starts to create conflicts!

PR #941, opened 2022-01-21
Document how to enable CILogon for a hub

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/315

I added a commit that enables CILogon for the staging hub, just as an example. I deployed the change manually however if anybody wants to test tat, but will be overwritten when a new deploy from the CI will happen.

On the staging hub right now we have:
- staff 2i2c emails are admins
- anyone with an `@2i2c.org` and `@berkeley.edu` address is allowed to login the hub.

### DIFF
--- a/docs/howto/configure/auth-management.md
+++ b/docs/howto/configure/auth-management.md
@@ -169,7 +169,7 @@ Presently, this involves a few more manual steps than the `auth0` setup describe
 
 ## CILogon
 
-For communities that require authenticating users against their campus identity providers, [CILogon](https://www.cilogon.org) can be used. CILogon authentication is managed through [auth0](https://auth0.com), similar with how the Google and GitHub authentication is setup.
+[CILogon](https://www.cilogon.org) allows us to authenticate users against their campus identity providers. It is managed through [auth0](https://auth0.com), similar to Google and GitHub authentication.
 
 ```{note}
   The `USERNAME_KEY` used for the CILogon login is the email, so the JupyterHub username will be set to the email address associated with the CILogon connection and not the CILogon user_id!
--- a/docs/howto/configure/auth-management.md
+++ b/docs/howto/configure/auth-management.md
@@ -172,7 +172,7 @@ Presently, this involves a few more manual steps than the `auth0` setup describe
 [CILogon](https://www.cilogon.org) allows us to authenticate users against their campus identity providers. It is managed through [auth0](https://auth0.com), similar to Google and GitHub authentication.
 
 ```{note}
-  The `USERNAME_KEY` used for the CILogon login is the email, so the JupyterHub username will be set to the email address associated with the CILogon connection and not the CILogon user_id!
+The JupyterHub username will be the email address that users provide when authenticating in CILogon connection. It will not be the CILogon `user_id`! This is because the `USERNAME_KEY` used for the CILogon login is the email address.
 ```
 
 To enable CILogon authentication:
--- a/docs/howto/configure/auth-management.md
+++ b/docs/howto/configure/auth-management.md
@@ -180,7 +180,7 @@ The JupyterHub username will be the email address that users provide when authen
 
 To enable CILogon authentication:
 
-1. Explicitly list CILogon as the type of connection we want for a hub, via `auth0.connection`:
+1. List CILogon as the type of connection we want for a hub, via `auth0.connection`:
 
    ```yaml
    auth0:
--- a/docs/howto/configure/auth-management.md
+++ b/docs/howto/configure/auth-management.md
@@ -171,6 +171,9 @@ Presently, this involves a few more manual steps than the `auth0` setup describe
 
 [CILogon](https://www.cilogon.org) allows us to authenticate users against their campus identity providers. It is managed through [auth0](https://auth0.com), similar to Google and GitHub authentication.
 
+```{seealso}
+See the [CILogon documentation on `Auth0`](https://www.cilogon.org/auth0) for more configuration information.
+```
 ```{note}
 The JupyterHub username will be the email address that users provide when authenticating in CILogon connection. It will not be the CILogon `user_id`! This is because the `USERNAME_KEY` used for the CILogon login is the email address.
 ```
--- a/docs/howto/configure/auth-management.md
+++ b/docs/howto/configure/auth-management.md
@@ -167,7 +167,7 @@ Presently, this involves a few more manual steps than the `auth0` setup describe
 
 4. Run the deployer as normal to apply the config.
 
-# CILogon
+## CILogon
 
 For communities that require authenticating users against their campus identity providers, [CILogon](https://www.cilogon.org) can be used. CILogon authentication is managed through [auth0](https://auth0.com), similar with how the Google and GitHub authentication is setup.
 
--- a/docs/howto/configure/auth-management.md
+++ b/docs/howto/configure/auth-management.md
@@ -193,35 +193,37 @@ To enable CILogon authentication:
   Don't forget to allow login to the test user (`deployment-service-check`), otherwise the hub health check performed during deployment will fail.
   ```
 
-  ```yaml
-    config:
-      jupyterhub:
-        hub:
-          config:
-            Authenticator:
-              admin_users:
-                - user1@campus.edu
-                - user2@gmail.com
-                # This will be replaced with the staff's google addresses
-                - <staff_google_ids>
-              username_pattern: '^(.+@2i2c\.org|.+@campus\.edu|deployment-service-check)$'
-   ```
 
-  This example does two things:
+### Example config for CILogon
 
-  1. Sets the hub admins to be:
+The following configuration example shows off how to configure hub admins and allowed users:
 
-    - the 2i2c staff (identified through their 2i2c email address)
-    - one `@campus.edu` user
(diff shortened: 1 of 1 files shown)


damianavila (approved):
LGTM

config/hubs/2i2c.cluster.yaml
```diff
@@ -47,12 +47,11 @@ hubs:
...
-              allowed_users: &staging_users
-                # this will be replaced with the actual list of
-                # staff google ids by the deployer script
+              # this will be replaced with the actual list of
+              # staff google ids by the deployer script
+              admin_users: &staging_admins
                 - <staff_google_ids>
-                - colliand@gmail.com
-              admin_users: *staging_users
+              username_pattern: '^(.+@2i2c\.org|.+@berkeley\.edu|deployment-service-check)$'
```
GeorgianaElena, line 54: Do you think we should test/show off the CILogon feature on the staging hub or it's best if I deploy a new hub?

damianavila: I would be OK to have it on staging, @2i2c-org/tech-team WDYT?

choldgraf: I actually feel like we should have a dedicated hub to show off these kinds of things - the staging hub is meant for staging and trying out changes as part of our deployment pipeline, so I feel like we should reserve it for those purposes rather than showing off new features and such. What if we deployed a hub like `demo.2i2c.cloud` that was for demonstration purposes, and where we could change the config etc without worrying that it'll affect our other deployments?

docs/howto/configure/auth-management.md
```diff
@@ -166,3 +166,59 @@ Presently, this involves a few more manual steps than the `auth0` setup describe
     ```
 
 4. Run the deployer as normal to apply the config.
+
+# CILogon
```
choldgraf, line 170:
```suggestion
## CILogon
```

docs/howto/configure/auth-management.md
```diff
@@ -166,3 +166,59 @@ Presently, this involves a few more manual steps than the `auth0` setup describe
...
+
+For communities that require authenticating users against their campus identity providers, [CILogon](https://www.cilogon.org) can be used. CILogon authentication is managed through [auth0](https://auth0.com), similar with how the Google and GitHub authentication is setup.
```
choldgraf, line 172:
```suggestion
[CILogon](https://www.cilogon.org) allows us to authenticate users against their campus identity providers. It is managed through [auth0](https://auth0.com), similar to Google and GitHub authentication.
```

docs/howto/configure/auth-management.md
```diff
@@ -166,3 +166,59 @@ Presently, this involves a few more manual steps than the `auth0` setup describe
...
+
+```{note}
+  The `USERNAME_KEY` used for the CILogon login is the email, so the JupyterHub username will be set to the email address associated with the CILogon connection and not the CILogon user_id!
```
choldgraf, line 175:
```suggestion
The JupyterHub username will be the email address that users provide when authenticating in CILogon connection. It will not be the CILogon `user_id`! This is because the `USERNAME_KEY` used for the CILogon login is the email address.
```

docs/howto/configure/auth-management.md
```diff
@@ -166,3 +166,59 @@ Presently, this involves a few more manual steps than the `auth0` setup describe
     ```
 
 4. Run the deployer as normal to apply the config.
+
+# CILogon
+
+For communities that require authenticating users against their campus identity providers, [CILogon](https://www.cilogon.org) can be used. CILogon authentication is managed through [auth0](https://auth0.com), similar with how the Google and GitHub authentication is setup.
+
+```{note}
```
choldgraf, line 174:
````suggestion
```{seealso}
See the [CILogon documentation on `Auth0`](https://www.cilogon.org/auth0) for more configuration information.
```
```{note}
````

File: docs/howto/configure/auth-management.md [hunk 6]
```diff
@@ -166,3 +166,59 @@ Presently, this involves a few more manual steps than the `auth0` setup describe
...
+  The `USERNAME_KEY` used for the CILogon login is the email, so the JupyterHub username will be set to the email address associated with the CILogon connection and not the CILogon user_id!
+```
+
+To enable CILogon authentication:
+
+1. Explicitly list CILogon as the type of connection we want for a hub, via `auth0.connection`:
```
choldgraf, line 180:
```suggestion
1. List CILogon as the type of connection we want for a hub, via `auth0.connection`:
```

docs/howto/configure/auth-management.md
```diff
@@ -166,3 +166,59 @@ Presently, this involves a few more manual steps than the `auth0` setup describe
...
+
+   ```yaml
+   auth0:
+      connection: CILogon
+   ```
+
+2. Explicitly list *admin users* for a given hub and only allow the users that match specific identity providers to login into the hub.
```
choldgraf, line 187:
```suggestion
2. Add **admin users** to the hub by explicitly listing their email addresses. Add **allowed users** for the hub by providing a regex pattern that will match to an institutional email address. (see example below)
```

docs/howto/configure/auth-management.md
```diff
@@ -166,3 +166,59 @@ Presently, this involves a few more manual steps than the `auth0` setup describe
...
+
+  ```{note}
+  Don't forget to allow login to the test user (`deployment-service-check`), otherwise the hub health check performed during deployment will fail.
```
choldgraf, line 190: Isn't this something that we can automatically add at deploy time? It isn't configuration a community would ever want to copy over themselves if they deployed their own hub

docs/howto/configure/auth-management.md
```diff
@@ -166,3 +166,59 @@ Presently, this involves a few more manual steps than the `auth0` setup describe
...
+  2. Allows logging in only the users that match the `username_pattern` expression:
+
+    - all `@2i2c.org` email adresses
+    - all `@campus.edu` email addresses
+    - `user2@gmail.com`
+    - the test username, `deployment-service-check`
+
+  ```{note}
+  All the users listed under `admin_users` need to match the `username_pattern` expression otherwise they won't be allowed to login!
+  ```
```
choldgraf, line 224:
````suggestion

### Example config for CILogon

The following configuration example shows off how to configure hub admins and allowed users:

1. **Hub admins** are these explicit emails:
  - one `@campus.edu` user
  - one `@gmail.com` user
  - the 2i2c staff (identified through their 2i2c email address)

2. **Allowed users** are matched against a pattern, with a few specific addresses added in as well
  - all `@2i2c.org` email adresses
  - all `@campus.edu` email addresses
  - `user2@gmail.com`
  - the test username, `deployment-service-check`

```yaml
config:
  jupyterhub:
    hub:
      config:
        Authenticator:
          admin_users:
            - user1@campus.edu
            - user2@gmail.com
            # This will be replaced with the staff's google addresses
            - <staff_google_ids>
          username_pattern: '^(.+@2i2c\.org|.+@campus\.edu|deployment-service-check)$'
```


```{note}
All the users listed under `admin_users` need to match the `username_pattern` expression otherwise they won't be allowed to login!
```
````

choldgraf: I'm a bit confused where it says that `user2@gmail.com` is also allowed into the hub - I only see it defined in `admin_users`, not in `username_pattern`. Could you clarify?

choldgraf (commented):
This is a nice start! I added a few suggestions and comments in there, let me know what you think!

docs/howto/configure/auth-management.md
GeorgianaElena: Super nice rephrasing! Thanks @choldgraf :blossom:

docs/howto/configure/auth-management.md
GeorgianaElena: I meant to add the `user2` in the username pattern, but apparently didn't. I edited you commit suggestion to contain the `user2` example.

config/hubs/2i2c.cluster.yaml
GeorgianaElena: What if we used the cloudbank demo hub? I don't think that's used right now
https://github.com/2i2c-org/infrastructure/blob/e8e80aeb4b9955f86bf9ca4d0a01d4ff1ff1ff58/config/hubs/cloudbank.cluster.yaml#L189

docs/howto/configure/auth-management.md
GeorgianaElena: Yeah, so if I remember correctly from when I struggled to have a nice way to add the staff by default to the admin&allowed lists, I don't think there's a straightforward way of doing this. And I don't think we should stress out about this as this `username_pattern` thing it's already pretty ugly as it is.

We should be able to get away from this if we manage to automatize the CILogon client creation and move out from Auth0 entirely.

GeorgianaElena:
Thanks for the feedback @damianavila and @choldgraf! I decided to drop the commit that enabled CILogon on the staging hub and instead let this PR be just about the docs so we don't block on that.

The plan is to open a follow-up PR to deploy a new 2i2c demo hub to show off the CILogon auth instead and we can discuss more there. Hope that's ok. I will go ahead and merge this now since I have an appoval :rocket:  Thank you!

PR #942, opened 2022-01-24
Tidy up some extra scripts

sgibson91:


extra_scripts/README.md
```diff
@@ -0,0 +1,9 @@
+# Extra Scripts
```
GeorgianaElena, line 1: <3 for documenting these.

GeorgianaElena (approved):
> I don't believe we should keep scripts/files that are not workflow configurations in the .github/workflows directory.

Makes sense! I'm +1 on moving it to the scripts dir.

LGTM :rocket:

consideRatio (approved):

choldgraf (approved):
LGTM - thanks!

PR #947, opened 2022-01-25
Enable pre-pullers on UToronto hubs

yuvipanda:
pre-puller is necessary as the image is pretty big, and
pulling during first user spawn might cause timeouts.
Only required on staging hub though, as they share the same
cluster, and staging is always deployed before prod

### DIFF
--- a/config/hubs/utoronto.cluster.yaml
+++ b/config/hubs/utoronto.cluster.yaml
@@ -40,6 +40,15 @@ hubs:
           # Trailing slash is important!
           baseShareName: /2i2cutorontohubstorage/homes/
       jupyterhub:
+        # pre-puller is necessary as the image is pretty big, and
+        # pulling during first user spawn might cause timeouts.
+        # Only required on staging hub though, as they share the same
+        # cluster, and staging is always deployed before prod
+        prePuller:
+          continuous:
+            enabled: false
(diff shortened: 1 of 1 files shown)


sgibson91 (approved):

config/hubs/utoronto.cluster.yaml
```diff
@@ -40,6 +40,15 @@ hubs:
...
       jupyterhub:
+        # pre-puller is necessary as the image is pretty big, and
+        # pulling during first user spawn might cause timeouts.
+        # Only required on staging hub though, as they share the same
+        # cluster, and staging is always deployed before prod
+        prePuller:
+          continuous:
+            enabled: false
+          hook:
+            enabled: false
```
damianavila, line 51: Wait, if we want to enable the continuous and hook prepullers (I would agree), should not be both `true` instead of `false`?? Am I missing something?

damianavila: @yuvipanda @sgibson91 ⬆️

yuvipanda: Whoops, great catch @damianavila. Fixed in https://github.com/2i2c-org/infrastructure/pull/949

PR #948, opened 2022-01-25
Fix a broken line of code

sgibson91:
Remove print_colour from "Hub health check failed" message. This line will break when called since there are extra args.


sgibson91:
Gonne merge this one in as it's a small change and a bugfix.

Issue #908, opened 2022-01-02
[Incident] OpenScapes unauthenticated users and CPU usage spike


choldgraf:
# Summary

Over the past 2 weeks there has been a spike in usage of the OpenScapes hub. User pods have been increasingly steadily over time. Because nodes scale with users on a 1-1 ratio, this has resulted in a large spike in nodes as well.

## Hub information

- Hub URL: openscapes.2i2c.org

## Timeline (if relevant)

All times in US/Pacific

### 2021-01-02 11:00AM

OpenScapes emails 2i2c support saying that they've gotten an extremely large AWS bill for December.

They reported **several users who they did not add to the hub, and who had GitHub accounts created in the last 2 weeks**.

The hub was shut down as a precautionary measure.

### 12:01pm

A look at the grafana logs showed that around the 21st the hub users started going up steadily:

![image](https://user-images.githubusercontent.com/1839645/147888814-71af4c06-bcd4-4f7e-8b73-d486028a1ae1.png)

It seems like many of these users were maxing out the CPU:

![image](https://user-images.githubusercontent.com/1839645/147889033-d8863dfb-a018-41f2-aeda-7d4aa56ef809.png)

Though the plot for users with high CPU usage is broken so this is hard to confirm:

![image](https://user-images.githubusercontent.com/1839645/147889028-cb809693-08ef-43c5-8ffa-ed3dc92e5b5b.png)

EDIT: the users have mining software in their filesystems:

![image](https://user-images.githubusercontent.com/1839645/147900227-afd038cc-e1c6-4ad0-8fd7-5fe490a56f74.png)

### 12:40pm

Found the reason that unauthorized users can access:

**Any github user that logs into this hub is authorized**. A new GitHub account was created from scratch, and attempted to log-in, and was able to do so successfully despite not being added to the hub's user list.

Noticed that the openscapes hub does not explicitly have an allowed_usersconfig set:

- here's the 2i2c gke hub
  
  https://github.com/2i2c-org/infrastructure/blob/b6ab94a3e029a4418e137a434b69d3db1999f96e/config/hubs/2i2c.cluster.yaml#L90-L96

- here's the openscapes hub

  https://github.com/2i2c-org/infrastructure/blob/b6ab94a3e029a4418e137a434b69d3db1999f96e/config/hubs/openscapes.cluster.yaml#L149-L154

### 21:00

The hub is made inaccessible via this PR: https://github.com/2i2c-org/infrastructure/pull/911

### 21:25

[...]

PR #950, opened 2022-01-26
Rename tests dir so that it matches the template

GeorgianaElena:
Otherwise no tests actually run for the basehub hubs...
https://github.com/2i2c-org/infrastructure/blob/37e487d8e063d3a031f7360f4147c2b1622f02ce/deployer/tests/test_hub_health.py#L9

I stumbled upon this by accident when trying to check if it's possible not to have the test fake user listed under `username_pattern` as part of https://github.com/2i2c-org/infrastructure/pull/941#discussion_r791263845, similar with how mills has it https://github.com/2i2c-org/infrastructure/blob/37e487d8e063d3a031f7360f4147c2b1622f02ce/config/hubs/cloudbank.cluster.yaml#L366


sgibson91 (approved):

GeorgianaElena:
Also, I believe that we should add the `deployment-service-check` user to the list of allowed users of each hub, otherwise the health checks won't succeed.

I believe the only reason it works right now (because I tested locally with the staging hub and the test user gets created successfully), it's because the test user is still registered in the hub from a previous time when it was allowed.

Screenshot from the staging hub admin panel:
test-user

sgibson91:
Ah, nvm. I was going to suggest that if we programmatically add `deployment-service-check` to `allowed_users` in the deployer, we could also plug this security gap https://github.com/2i2c-org/infrastructure/issues/908

choldgraf:
really nice catch @GeorgianaElena :-)

PR #951, opened 2022-01-26
Shorten mills username_pattern

GeorgianaElena:
This allows everyone with a 2i2c address to login into the mills hub instead of listing the entire tech team one by one in an effort to de-uglify a bit the `username_pattern` and make it more understandable.

### DIFF
--- a/config/hubs/cloudbank.cluster.yaml
+++ b/config/hubs/cloudbank.cluster.yaml
@@ -363,7 +363,7 @@ hubs:
                 - jpercy@berkeley.edu
                 - akonrad@mills.edu
                 - wang@mills.edu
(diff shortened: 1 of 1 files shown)


consideRatio (approved):
:heart: oh I really appreciate when things are simplified like this :)

I guess a question of accepting all 2i2c people vs only people in the tech team should be considered as part of reviewing this PR. I lean towards thinking its okay, but I don't know really.

damianavila (approved):
I agree with @consideRatio's considerations about the distinction between the tech team and others with 2i2c email addresses, but for now, it is OK.

GeorgianaElena:
> but for now, it is OK.

I think so too!

Thanks @consideRatio and @damianavila :blossom:

PR #953, opened 2022-01-28
Rename hub-templates folder to helm-charts

sgibson91:
This PR renames our `hub-templates` directory to be called `helm-charts`. It also updates our docs to keep language consistent and the deployer to work with the new filepath. I did a test deploy against pangeo-hubs staging, and all worked.

The broken link check test will resolve itself once this is merged.


docs/topic/hub-templates.md
```diff
@@ -1,47 +1,48 @@
...
-The graphic below, shows the relationship between the hub templates and the other
+The graphic below, shows the relationship between the hub helm charts and the other
 config files and how they are merged together when deploying a pilot hub.
 
 ```{figure} ../images/config-flow.png
 ```
 % The editable version of the diagram is here: https://docs.google.com/presentation/d/1WZKTe5TSDU-5zA4NnNEPsKfgjaBzUqMgdPLTnz-Yb94/edit?usp=sharing
 
-Currently there are three hub templates available:
+Currently there are three hub helm charts available:
```
consideRatio, line 15:
```suggestion
Currently there are two hub helm charts available:
```

Unrelated suggestion, can be ignored.

consideRatio (commented):
This looks great and thorough! I note how you have also managed to spot changes in terminology when "template" or similar were instead instead of "chart".

I spotted one `hub-template` reference in a github issue template that needs updating still - fix that and go for merge makes perfect sense to me! This is one of those changes that influences so much that its relevant to get merged quickly.

EDIT: Actually, search for "hub template" as well and there are two instances more needing updates.

sgibson91:
I update the GitHub issue template, but rather than "helm chart" I used "hub type" instead. This is because the people who will be filling out that template most likely won't know (and arguably, don't need to know) what a helm chart is, and I didn't want to add unnecessary confusion for them.

@consideRatio for the instances of the `hub template` that appear in comments of the deployer as such, is it ok if I tackle those when I change the `template` key in the cluster config to `helm_chart` instead? As that PR will touch a lot more of the code in the deployer. I believe I have only touched what is required here to keep the deployer functional.

consideRatio (approved):
Nice, this LGTM! I suggest we go for merge now as we've agreed on the kind of change already.

choldgraf (approved):
Nice! I like how you've also changed the diagrams (and it makes me happy that we had documented where those diagrams were so they could be changed by others :-) ).

I agree w/ @consideRatio - we should probably try to increase the review/merge velocity on this set of PRs since it'll introduce a lot of semi-breaking changes that will be come stale quickly. I'm +1 on merging after quick QA reviews from team members, since we've already discussed the decision a good amount!

sgibson91:
Thanks @choldgraf! I would like to have one set of eyes that are not me or @consideRatio for these PRs I think. Only because we've both done really deep thinking on this topic and it would be nice to have verification from a team member who's a little more removed just to confirm our implementation details aren't bonkers! (as opposed to the general concept)

choldgraf:
Good point @sgibson91 I trust y'alls judgment on the right cadence to shoot for!

damianavila:
> I would like to have one set of eyes that are not me or @consideRatio for these PRs I think.

Belated LGTM 😉

PR #954, opened 2022-01-28
Rename hub-templates.md to hub-helm-charts.md

sgibson91:
Whoops! I forgot this in #953


sgibson91:
I will merge this without review as it only affects our docs

PR #955, opened 2022-01-31
Use helm_chart key instead of template in deployer

sgibson91:
This PR changes the `template` key in the deployer and cluster config files to `helm_chart`, in keeping with the new folder name. I have ran the modified deployer against the 2i2c staging hub and it executed successfully.

PR #958, opened 2022-01-31
Setup a demo hub for University of Texas

yuvipanda:
- Each user pod gets a postgres container as a 'sidecar', listening
  on localhost only. So each user gets their own postgres database
  with full rights! There is no persistent storage for each database,
  just an emptyDir. So all data is discarded when user's server stops.
- New image has been setup, deployed to https://quay.io/repository/2i2c/utexas-image?tab=info
  from https://github.com/2i2c-org/utexas-image
- Users can connect to the database on the terminal with
  ```
  psql -h localhost
  ```
- Python connections can be made with:

  ```
  import psycopg2
  conn = psycopg2.connect(host='localhost')
  ```
- https://pypi.org/project/ipython-sql/ is also installed!

Ref https://github.com/2i2c-org/leads/issues/52

### DIFF
--- a/config/hubs/2i2c.cluster.yaml
+++ b/config/hubs/2i2c.cluster.yaml
@@ -452,7 +452,7 @@ hubs:
               admin_users: *aup_users
   - name: utexas-demo
     domain: utexas-demo.pilot.2i2c.cloud
-    template: basehub
+    helm_chart: basehub
     auth0:
       connection: github
     config:
--- a/config/hubs/2i2c.cluster.yaml
+++ b/config/hubs/2i2c.cluster.yaml
@@ -450,3 +450,82 @@ hubs:
                 - swalker
                 - shaolintl
               admin_users: *aup_users
+  - name: utexas-demo
+    domain: utexas-demo.pilot.2i2c.cloud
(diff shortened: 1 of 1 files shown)


config/hubs/2i2c.cluster.yaml
```diff
@@ -450,3 +450,87 @@ hubs:
                 - swalker
                 - shaolintl
               admin_users: *aup_users
+  - name: utexas-demo
+    domain: utexas-demo.pilot.2i2c.cloud
+    template: basehub
```
sgibson91, line 455:
```suggestion
    helm_chart: basehub
```

See https://github.com/2i2c-org/infrastructure/pull/955

sgibson91 (approved):
One change to make, otherwise LGTM!

choldgraf:
One thought - we did not add any members of the U Texas community to this hub as an admin. I think at a minimum we should add `@jameshowison` so that he can access the hub 👉  https://github.com/2i2c-org/infrastructure/pull/960

damianavila:
Belated, LGTM 😉 
This is great, @yuvipanda!

PR #959, opened 2022-01-31
Move support chart under helm-charts directory

sgibson91:
This PR moves the `support` helm chart under the `helm-charts` directory and updates the deployer to work with the new filepath. I ran the `deploy-support` command against the 2i2c cluster and it executed successfully.


consideRatio (changes requested):
Searching for `support/` I found some missed references in .sops.yaml and the documentation that needs updating still.

sgibson91:
Thanks @consideRatio - updated!

consideRatio (approved):
Nice, LGTM!

PR #960, opened 2022-01-31
Add @jameshowison as an admin on the UT hub

choldgraf:
This adds @jameshowison as an admin on the UT hub that @yuvipanda just deployed.

### DIFF
--- a/config/hubs/2i2c.cluster.yaml
+++ b/config/hubs/2i2c.cluster.yaml
@@ -480,7 +480,6 @@ hubs:
                 - <staff_github_ids>
                 - jameshowison
               admin_users: *utexas_demo_users
-
         singleuser:
           storage:
             extraVolumes:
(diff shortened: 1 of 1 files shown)


config/hubs/2i2c.cluster.yaml
```diff
@@ -478,7 +478,9 @@ hubs:
             Authenticator:
               allowed_users: &utexas_demo_users
                 - <staff_github_ids>
+                - jameshowison
               admin_users: *utexas_demo_users
+
```
choldgraf, line 483:
```suggestion
```

colliand:
Ah, I misinterpreted. I thought I was to add James....

choldgraf:
That's fine too - we've just often explicitly listed the community representatives in the config and let them add others. If you've already added them then this will not change the hub itself, only make it more explicit

yuvipanda (approved):
tyvm!

PR #963, opened 2022-02-01
Rename config/hubs dir to config/clusters

sgibson91:
This PR renames the `config/hubs` and `secrets/config/hubs` folders to `config/clusters` and `secrets/config/clusters` respectively. The deployer and documentation have been updated to reflect the new path. I ran the deployer against the 2i2c staging hub and it executed successfully.

Broken link check will resolve itself upon merge.


docs/howto/operate/new-cluster/aws.md
```diff
@@ -204,6 +204,6 @@ Get the address the hub should use for connecting to NFS with
 ## Add the cluster to be automatically deployed
 
 The [CI deploy-hubs
-workflow](https://github.com/2i2c-org/pilot-hubs/blob/e96e7bcded187870dc2e07d6626de8a12586ed32/.github/workflows/deploy-hubs.yaml#L31-L36)
+workflow](https://github.com/2i2c-org/infrastructure/blob/master/.github/workflows/deploy-hubs.yaml#L31-L36)
```
GeorgianaElena, line 207: I think in jupyterhub repos the HEAD commit is being used instead of specifying a branch name when referencing files in the repo. It might be a good idea to do so too?

```suggestion
workflow](https://github.com/2i2c-org/infrastructure/blob/HEAD/.github/workflows/deploy-hubs.yaml#L31-L36)
```

GeorgianaElena (approved):
Thanks for pushing this forward @sgibson91 :rocket:

docs/howto/operate/new-cluster/aws.md
sgibson91: In this particular instance, I don't think it matters since `HEAD` always refers to the latest commit on the default branch, i.e., `master`. Making this change would only save us work if we decided to rename the default branch of the repo.

That being said, I do think it is a good idea! But I would prefer to tackle it in a new PR as I'll likely do `s/master/HEAD/g` across the whole docs repo :)

sgibson91: Thank you!

consideRatio (commented):
All changes LGTM and the changes seem complete!! :tada: :rocket:

choldgraf:
this one has bugged me forever! :-)

damianavila:
Super!!

PR #966, opened 2022-02-01
Replace instances of blob/master with tree/HEAD in links in our docs

### LINKED ISSUE #964: Replace github links referencing a default branch to HEAD
### Description of problem and opportunity to address it

**Context to understand the problem**
A github URL can be made using HEAD instead of main/master/or-some-other-default-branch-name. We should go for HEAD to simplify renaming the default branch and overall spread knowledge of this practice which is helpful to migrate from master to main in general.

**Proposed solution**
Search and replace.

_No response_

### Updates and ongoing work

_No response_

sgibson91:
fixes #964 

I only updated links to our own resources (this repo) or resources we also have a hand in managing (e.g. z2jh)

I also believe the `tree/HEAD` syntax is required over `blob/HEAD`

- `https://github.com/2i2c-org/infrastructure/blob/HEAD` https://github.com/2i2c-org/infrastructure/blob/HEAD -> 404
- `https://github.com/2i2c-org/infrastructure/tree/HEAD` https://github.com/2i2c-org/infrastructure/tree/HEAD -> resolves fine


docs/howto/operate/grafana.md
```diff
@@ -80,7 +80,7 @@ The key you create needs admin permissions.
 
 Encrypt and store this key using `sops` in `secrets/config/clusters/<cluster>.yaml` under `grafana_token` key.
 
-This key will be used by the [`deploy-grafana-dashboards` workflow](https://github.com/2i2c-org/infrastructure/blob/HEAD/.github/workflows/deploy-grafana-dashboards.yaml) to deploy some default grafana dashboards for JupyterHub using [`jupyterhub/grafana-dashboards`](https://github.com/jupyterhub/grafana-dashboards).
+This key will be used by the [`deploy-grafana-dashboards` workflow](https://github.com/2i2c-org/infrastructure/tree/HEAD/.github/workflows/deploy-grafana-dashboards.yaml) to deploy some default grafana dashboards for JupyterHub using [`jupyterhub/grafana-dashboards`](https://github.com/jupyterhub/grafana-dashboards).
```
consideRatio, line 83: I'm a bit confused on the difference between blob and head. Both links work, but if you use `/tree/HEAD` you are redirected to `/blob/<commit hash>` while if you use `/blob/HEAD` you remain fixed and it works.

This LGTM as it is, no change needed, but off topic: do you know something about it?

sgibson91: Hmmm, I updated the top comment with my thoughts. I wonder if `blob` is required for viewing files and `tree` required for viewing directories? Let's see what the linkcheck test says...

docs/reference/ci-cd.md
```diff
@@ -1,10 +1,10 @@
 # Our CI/CD system
 
 The best place to learn about the latest state of our *automatic* hub deployment
-is to look at [the `deploy-hubs.yaml` GitHub workflow](https://github.com/2i2c-org/pilot-hubs/blob/master/.github/workflows/deploy-hubs.yaml).
+is to look at [the `deploy-hubs.yaml` GitHub workflow](https://github.com/2i2c-org/pilot-hubs/tree/HEAD/.github/workflows/deploy-hubs.yaml).
```
consideRatio, line 4:
```suggestion
is to look at [the `deploy-hubs.yaml` GitHub workflow](https://github.com/2i2c-org/infrastructure/tree/HEAD/.github/workflows/deploy-hubs.yaml).
```

consideRatio (approved):
Thanks for picking up this also, LGTM!

docs/howto/operate/grafana.md
consideRatio: > Hmmm, I updated the top comment with my thoughts. I wonder if blob is required for viewing files and tree required for viewing directories? Let's see what the linkcheck test says...

Ahh haha yeah, actually, the behavior i wrote about is reversed for directories. I guess they support blob/tree for both directories/files, but they do `blob` and `tree` depending on what kind of view they render (directory viewing kind of view for tree / file viewing kind of view for blob). If they switch from blob -> tree or tree -> blob, they pin the reference HEAD to a specific commit.

sgibson91: So kinda doesn't matter, we get redirected to the right place anyway :D

GeorgianaElena (approved):
Thank you @sgibson91 :heart: !

GeorgianaElena:
> I also can't connect so maybe that site just happens to be down right now?

I think so too! https://www.gnu.org doesn't work either.

sgibson91:
Thanks for checking @GeorgianaElena! I'll just merge then and hopefully GNU sorts itself out 😄

docs/howto/operate/grafana.md
consideRatio: My summary becomes:

The choice is harmless, but you get another HTTP 301 redirect to follow using `/blob/HEAD` or `/tree/HEAD` when referencing the wrong associated type. `/tree` goes with directory views, `/blob` goes with file views.

PR #969, opened 2022-02-02
Tidyup-deployer

sgibson91:


deployer/__main__.py
```diff
@@ -19,6 +19,26 @@
...
+    # Validate our config with JSON Schema first before continuing
+    validate(cluster_name)
+
+    config_file_path = (
+        Path(os.getcwd()) / "config/clusters" / f"{cluster_name}.cluster.yaml"
+    )
+    with open(config_file_path) as f:
+        cluster = Cluster(yaml.load(f))
+
+    # Cluster.auth() method is has the context manager decorator so cannot call
```
sgibson91, line 36:
```suggestion
    # Cluster.auth() method has the context manager decorator so cannot call
```

deployer/__main__.py
```diff
@@ -19,6 +18,26 @@
 yaml = YAML(typ="safe", pure=True)
 
 
+def auth(cluster_name):
+    """
+    Quickly gain command-line access to a cluster using deployer credentials
+    """
```
consideRatio, line 24: Can you add to this docstring a bit? Some questions left unanswered to me are:

- Is this a "call once at any time to update the current kubeconfig file with credentials to a k8s cluster"?
- Is it supposed to be used as a CLI command only, or as a helper by the python code overall?

deployer/__main__.py
```diff
@@ -38,9 +57,10 @@ def deploy_support(cluster_name):
             cluster.deploy_support()
 
 
-def deploy_jupyterhub_grafana(cluster_name):
+def deploy_grafana_dashboards(cluster_name):
     """
-    Deploy grafana dashboards for operating a hub
+    Deploy grafana dashboards to a cluster that provide useful metrics
+    for operating a JupyterHub
```
consideRatio, line 63: Can you clarify what dashboards are deployed? This is https://github.com/jupyterhub/grafana-dashboards right?

deployer/__main__.py
```diff
@@ -199,33 +233,76 @@ def validate(cluster_name):
...
+        parents=[base_parser],
+        help="Deploy grafana dashboards to a cluster for monitoring JupyterHubs. deploy-support must be run first!",
+    )
 
-    deploy_grafana_parser.add_argument("cluster_name")
+    # Auth subcommand
+    auth_parser = subparsers.add_parser(
+        "auth",
+        parents=[base_parser],
+        help="Authenticate against a cluster using deployer credentials to gain quick command line access",
```
consideRatio, line 303: Authenticate against a cluster sounds like its actually setting something up against the remote cluster. I think this command setups credentials for k8s cluster access, but I'm not sure what it does. Is it only modifying the kubeconfig (i think so)? Is it influencing environment variables (i don't think so)? I hope to see have this be more explicit on what is done.

consideRatio (commented):
This looks very nice overall! I would like some additional clarity on what the `auth` command does in practice.

I also wonder about the naming of this command/function. When you do `gcloud auth login` you verify you are you as a google user. When you do `gcloud container clusters get-credentials <cluster name>` you update the current kubeconfig with credentials to access the k8s clsuter.

I think a name like `setup-cluster-credentials` would be more easy to think about than the minimal `auth` which makes me associate to quite a lot of things that may be unrelated to what it really does. I get thinking about things like auth against auth0, auth against google for access to our specific two-eye-two-see project, auth against some cloud provider related to a cluster, setup credentials in the current kubeconfig against a k8s cluster.

sgibson91:
Thanks @consideRatio - I pushed some updates

deployer/__main__.py
```diff
@@ -18,9 +18,14 @@
 yaml = YAML(typ="safe", pure=True)
 
 
-def auth(cluster_name):
+def setup_cluster_credentials(cluster_name):
     """
-    Quickly gain command-line access to a cluster using deployer credentials
+    Quickly gain command-line access to a cluster by udpating the current
+    kubeconfig file to use the deployer's access credentials for the named
+    cluster.
```
consideRatio, line 25:
```suggestion
    Quickly gain command-line access to a cluster by updating the current
    kubeconfig file to include the deployer's access credentials for the named
    cluster and mark it as the cluster to work against by default.
```

consideRatio (commented):
With the suggested changes, this LGTM! I love it! Knowing that this kind of functionality is available makes me fare less afraid of looking into some issue somewhere.

deployer/__main__.py
```diff
@@ -19,6 +18,31 @@
...
+        Path(os.getcwd()) / "config/clusters" / f"{cluster_name}.cluster.yaml"
+    )
+    with open(config_file_path) as f:
+        cluster = Cluster(yaml.load(f))
+
+    # Cluster.auth() method has the context manager decorator so cannot call
+    # it like a normal function
+    with cluster.auth():
+        return None
+
```
damianavila, line 44: This is interesting but I am not sure how it is expected to work, @sgibson91. 
You call it from the CLI and then the kubeconfig is updated with the deployer credentials, right?
But when the context manager finishes it should remove the environment variables or restore the kubeconfig (ie. https://github.com/2i2c-org/infrastructure/blob/master/deployer/hub.py#L226-L232, except in this case: https://github.com/2i2c-org/infrastructure/blob/master/deployer/hub.py#L147-L162, where we are not getting the kubeconfig to the original state). Am I probably missing something here?

damianavila:
Sorry for the belated review, left a tiny comment above 😉

sgibson91:
Screenshot 2022-02-04 at 10 08 09

(For some reason I couldn't add this image to the thread above)

@damianavila It seems you're right - which is annoying, this would be a really useful feature.

sgibson91:
I am trying to use a boolean flag to allow us to override this behaviour _only_ when the setup-cluster-credentials command is called, but it's not working.

Attempted diff:

```diff
diff --git a/deployer/__main__.py b/deployer/__main__.py
index 52fc7c8..2a8a003 100644
--- a/deployer/__main__.py
+++ b/deployer/__main__.py
@@ -39,7 +39,7 @@ def setup_cluster_credentials(cluster_name):
 
     # Cluster.auth() method has the context manager decorator so cannot call
     # it like a normal function
-    with cluster.auth():
+    with cluster.auth(clear_kubeconfig=False):
         return None
 
 
diff --git a/deployer/hub.py b/deployer/hub.py
index 7b07ced..e2dcba4 100644
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -30,13 +30,13 @@ def __init__(self, spec):
         self.support = self.spec.get("support", {})
 
     @contextmanager
-    def auth(self):
+    def auth(self, clear_kubeconfig=True):
         if self.spec["provider"] == "gcp":
-            yield from self.auth_gcp()
+            yield from self.auth_gcp(clear_kubeconfig)
         elif self.spec["provider"] == "aws":
-            yield from self.auth_aws()
+            yield from self.auth_aws(clear_kubeconfig)
         elif self.spec["provider"] == "azure":
-            yield from self.auth_azure()
+            yield from self.auth_azure(clear_kubeconfig)
         elif self.spec["provider"] == "kubeconfig":
             yield from self.auth_kubeconfig()
         else:
@@ -161,7 +161,7 @@ def auth_kubeconfig(self):
             os.environ["KUBECONFIG"] = decrypted_key_path
             yield
 
-    def auth_aws(self):
+    def auth_aws(self, clear_kubeconfig):
         """
         Reads `aws` nested config and temporarily sets environment variables
         like `KUBECONFIG`, `AWS_ACCESS_KEY_ID`, and `AWS_SECRET_ACCESS_KEY`
@@ -224,14 +224,15 @@ def auth_aws(self):
 
                 yield
             finally:
-                if orig_kubeconfig is not None:
-                    os.environ["KUBECONFIG"] = orig_kubeconfig
-                if orig_access_key_id is not None:
-                    os.environ["AWS_ACCESS_KEY_ID"] = orig_access_key_id
-                if orig_kubeconfig is not None:
-                    os.environ["AWS_SECRET_ACCESS_KEY"] = orig_secret_access_key
-
-    def auth_azure(self):
+                if clear_kubeconfig:
+                    if orig_kubeconfig is not None:
+                        os.environ["KUBECONFIG"] = orig_kubeconfig
+                    if orig_access_key_id is not None:
+                        os.environ["AWS_ACCESS_KEY_ID"] = orig_access_key_id
+                    if orig_kubeconfig is not None:
+                        os.environ["AWS_SECRET_ACCESS_KEY"] = orig_secret_access_key
+
+    def auth_azure(self, clear_kubeconfig):
         """
         Read `azure` nested config, login to Azure with a Service Principal,
         activate the appropriate subscription, then authenticate against the
@@ -292,10 +293,10 @@ def auth_azure(self):
 
                 yield
             finally:
-                if orig_kubeconfig is not None:
+                if (clear_kubeconfig) and (orig_kubeconfig is not None):
                     os.environ["KUBECONFIG"] = orig_kubeconfig
 
-    def auth_gcp(self):
+    def auth_gcp(self, clear_kubeconfig):
         config = self.spec["gcp"]
         key_path = config["key"]
         project = config["project"]
@@ -332,7 +333,7 @@ def auth_gcp(self):
 
                 yield
             finally:
-                if orig_kubeconfig is not None:
+                if (clear_kubeconfig) and (orig_kubeconfig is not None):
                     os.environ["KUBECONFIG"] = orig_kubeconfig
```

sgibson91:
I'll remove the function for now https://github.com/2i2c-org/infrastructure/pull/976

sgibson91:
This got reimplemented and fixed in https://github.com/2i2c-org/infrastructure/pull/977

Issue #956, opened 2022-01-31
 Logging out of uoft hub does not actually log you out


GeorgianaElena:
### Description of problem and opportunity to address it

**Context to understand the problem**
When logging out of the UToronto hub, you're not actually being logged out, so a second attempt to login won't show you the University of Toronto login service page and you'll be logged in automatically.

**Problem or idea**
Once the `logout` button in hub is pressed, the user should be taken to the UofT login service screen.

**Proposed solution**
We had a similar issue with the Auth0 Authenticator before https://github.com/2i2c-org/infrastructure/issues/422 that we managed to solve by enabling setting up a custom `logout_url` for any of the hub OAuthenticators types. We did setup a `logout_url` for the Auth0 authenticator: https://github.com/2i2c-org/infrastructure/blob/066f166de3b40af7dfcec9565a788163db5416a8/deployer/auth.py#L164-L167
But since the toronto hub uses the [AzureAdOAuthenticator](https://github.com/2i2c-org/infrastructure/blob/master/config/hubs/utoronto.cluster.yaml#L125), we need to figure out the logout_url for that and configure it. (Maybe some useful [docs](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#send-a-sign-out-request))

**What's the value and who would benefit**
This would allow uoft users to do completely logout their hubs and close their sessions.

GeorgianaElena:
Reiterating on [this comment from the PR](https://github.com/2i2c-org/infrastructure/pull/972#issuecomment-1028980925):

> Also, there's something that's bugging me about this logout on multiple levels process arrow_down

> Should we in fact logout users out of their SSO?

> Isn't the purpose of having Single sign-on to help with logging into apps that use this scheme with the same account a user is logged into their SSO? I'm a bit confused about all these levels of logging out and what is and is not expected confused

This is how https://github.com/2i2c-org/infrastructure/pull/972 is behaving on staging:
![utoronto-logout](https://user-images.githubusercontent.com/7579677/152505616-b890d40c-ed89-42e8-9577-a0878848a494.gif)

So even if you choose **"Keep me signed in"**, you won't be.

**HOWEVER**
Without https://github.com/2i2c-org/infrastructure/pull/972, if you choose to **not** be kept signed in, you will be kept signed in still.

So, it looks like either option breaks some assumptions.

PR #972, opened 2022-02-02
Specify the logout_redirect_url for the utoronto hubs

GeorgianaElena:
This PR uses the `logout_redirect_url` config option of OAuthenticator to log users out of Azure AD, otherwise, state is not cleared and a user that has recently logged out of their hub, won't be presented with the option to choose what UofT account to login with and be automatically logged in with whatever account they used before. 

*Note:* We're doing something similiar for the other hubs that use Auth0 https://github.com/2i2c-org/infrastructure/blob/066f166de3b40af7dfcec9565a788163db5416a8/deployer/auth.py#L164-L167

Fixes https://github.com/2i2c-org/infrastructure/issues/956

### DIFF
--- a/config/clusters/utoronto.cluster.yaml
+++ b/config/clusters/utoronto.cluster.yaml
@@ -154,3 +154,4 @@ hubs:
               login_service: "University of Toronto ID"
               oauth_callback_url: https://jupyter.utoronto.ca/hub/oauth_callback
               tenant_id: 78aac226-2f03-4b4d-9037-b46d56c55210
+              logout_redirect_url: https://login.microsoftonline.com/common/oauth2/logout?post_logout_redirect_uri=https://jupyter.utoronto.ca
--- a/config/clusters/utoronto.cluster.yaml
+++ b/config/clusters/utoronto.cluster.yaml
@@ -127,6 +127,7 @@ hubs:
               login_service: "University of Toronto ID"
               oauth_callback_url: https://staging.utoronto.2i2c.cloud/hub/oauth_callback
               tenant_id: 78aac226-2f03-4b4d-9037-b46d56c55210
+              logout_redirect_url: https://login.microsoftonline.com/common/oauth2/logout?post_logout_redirect_uri=https://staging.utoronto.2i2c.cloud
   - name: prod
     domain: jupyter.utoronto.ca
     helm_chart: basehub


sgibson91 (approved):
Looks good @GeorgianaElena!

Since the only difference is the hub domain appended to the logout URL, I wonder if setting this value is something we can automate in the deployer when using CILogon?

consideRatio:
> Since the only difference is the hub domain appended to the logout URL, I wonder if setting this value is something we can automate in the deployer when using CILogon?

Hmmm, I'm quite happy about seeing us break the DRY principle in configuration if it favors reduced code complexity which I think is more challenging to maintain sustainably. I'm also thinking that hardcoding assumptions about URLs etc may be too specific to a certain identity provider, or overall make the deployer script box of logic more opaque overall.

sgibson91:
We already do it for Auth0 though https://github.com/2i2c-org/infrastructure/blob/master/deployer/hub.py#L444-L456

GeorgianaElena:
> We already do it for Auth0 https://github.com/2i2c-org/infrastructure/blob/master/deployer/hub.py#L444-L456

@sgibson91 is right, we do it for Auth0! In fact, the entire config for Auth0Oauthenticator [is set by the deployer](https://github.com/2i2c-org/infrastructure/blob/master/deployer/hub.py#L462-L464) using the `auth.py` module:
https://github.com/2i2c-org/infrastructure/blob/4e13e5ad8db5b8ee4cf496d25cb4b132f3de8d8d/deployer/auth.py#L157-L169

But since only the toronto hub is using the AzureAdOAuthenticator, I figured it's ok to hardcode a couple of URLs for start. What do you think?

GeorgianaElena:
Also, there's something that's bugging me about this logout on multiple levels process :arrow_down: 

*Should we in fact logout users out of their SSO?*

 Isn't the purpose of having `Single sign-on` to help with logging into apps that use this scheme with the same account a user is logged into their SSO? I'm a bit confused about all these levels of logging out and what is and is not expected :confused:

sgibson91:
@GeorgianaElena I agree. Let's leave the decision to automate this or not until after we decide where CILogon in our authentication story (will it 100% replace Auth0 or not)

damianavila (approved):
LGTM

GeorgianaElena:
Since I have all of your approvals, I will go ahead and merge this :rocket: 
There is still however [this aspect](https://github.com/2i2c-org/infrastructure/issues/956#issuecomment-1029805453) that's bugging me, but I guess that if users seems to be ok with the new logout behavior, we're ok.

Thanks everyone!

Issue #957, opened 2022-01-31
 Logging out of 2i2c demo hub does not actually log you out


GeorgianaElena:
### Description of problem and opportunity to address it

**Context to understand the problem**
When logging out of the new 2i2c demo hub demo.pilot.2i2c.cloud, you're not actually being logged out on an Identity Provider level, so a second attempt to login might show you the CILogon login page, but you won't be able to change the IDP id.

**Problem or idea**
We're currently logging out users at a hub and Auth0 session level, which is enough for hub not using CILogon. But with CILogon we're adding another auth proxy level that so we need to support logging out from the IDP too.

**Proposed solution**
I believe we should implement the logout at an Identity Provider Session Layer too (https://auth0.com/docs/authenticate/login/logout).

**What's the value and who would benefit**
This would allow users to do change the IDP, or account when logging in using CILogon.

GeorgianaElena:
Unfortunately, I couldn't wasn't able to find a way to logout users out of CILogon through Auth0. The solution suggested in the [Auth0 docs](https://auth0.com/docs/authenticate/login/logout/log-users-out-of-idps), to use the `federated` query param didn't work not matter how hard I tried to hack into it. I believe it is because there isn't federated logout support in Auth0 for CILogon -> https://auth0.com/docs/authenticate/login/logout/log-users-out-of-idps#federated-logout-support.

### Possible workarounds
1. Ask users go to `https://cilogon.org/logout` if they want to use other an account from the same Identity Provider they have used befor
2. When closing the user's Auth0 session, instead of redirecting them back to the hub home page, redirect them to the CILogon logout page. This means however that they would need to navigate themselves back the hub home page.
https://github.com/2i2c-org/infrastructure/blob/066f166de3b40af7dfcec9565a788163db5416a8/deployer/auth.py#L154

damianavila:
Mmm.. both workarounds feel broken but I think I would probably go with option 1 because it feels less disruptive, IMHO.

choldgraf:
Agree w/ @damianavila

GeorgianaElena:
Closed by https://github.com/2i2c-org/infrastructure/pull/974

Issue #973, opened 2022-02-03
Should a 403 on login clear oauth state?


GeorgianaElena:
### Description of problem and opportunity to address it

**Context to understand the problem**
When a users logins into a hub using an account that isn't allowed, they will get a 403 error message that looks like this and it's configured from https://github.com/2i2c-org/pilot-homepage/blob/master/templates/error.html
![403-error-message](https://user-images.githubusercontent.com/7579677/152321718-feeb79dd-2d4c-4a38-87e9-4d2553c0226b.png)

**Problem**
If a user that got a 403 want to try login with another account, then they must navigate themselves to the hub logout endpoint `htttps://<hub-address>/hub/logout` that will clear any cookies saved, otherwise they will just be redirected to the `403` page until *I think* the `oauthenticator_state` cookie expires.

**How to reproduce video:**
![403-error-reproduce](https://user-images.githubusercontent.com/7579677/152323475-0fdf5a6e-7ba6-4463-9694-4e56f301a537.gif)

**Cookies for the request that resolves to a 403**
![cookies-403](https://user-images.githubusercontent.com/7579677/152322608-a242985e-6955-4255-9209-8e48bcccd034.png)

choldgraf:
Ah interesting - I ran into the same behavior as well. In my opinion, if you try to log-in but you hit an "unauthorized" error, you should immediately be logged out. At a minimum, I think there should be a link displayed like "To log in with a different account, click here".

Is there a way we can change this behavior here, or does it require a change in JupyterHub?

damianavila:
> In my opinion, if you try to log-in but you hit an "unauthorized" error, you should immediately be logged out. 

I am a little bit worried about automatically redirecting to logout and "missing" the error...

> At a minimum, I think there should be a link displayed like "To log in with a different account, click here".

I like that idea (probably with another message).

> Is there a way we can change this behavior here, or does it require a change in JupyterHub?

Maybe a button in the very same template referenced above and pointing to `https://<hub-address>/hub/logout`?

choldgraf:
> Maybe a button in the very same template referenced above and pointing to https://<hub-address>/hub/logout?

If this were possible, I think it would be nice. Just a quick "here's a next step to take" message would be a big help at reducing confusion, I think

GeorgianaElena:
Thanks @damianavila and @choldgraf! I really like the idea of guiding users to the hub logout endpoint through a button so
[...]

PR #974, opened 2022-02-03
DOCS: Document logging out of CILogon

GeorgianaElena:
Does @2i2c-org/tech-team agree that adding this note to the docs is acceptable for fixing https://github.com/2i2c-org/infrastructure/issues/957? 

Note [this comment in the original issue](https://github.com/2i2c-org/infrastructure/issues/957#issuecomment-1028036063) that gives more details about it.


GeorgianaElena:
>Can we maybe rephrase this along the lines of "If a user mistakenly provides an incorrect email when logging in, they will need to... before doing...". Or maybe "If a user wishes to change which email address they login with,...". I'm having trouble keeping track between the "first" and "second" attempts.

Sure! There's a bit of a mess in my head right now :exploding_head: 

Two thoughts:

1. I'm a bit reserved using `email address` to refer to the user account.

So, with CILogon, the login process looks like this:

Hub login page :arrow_right: Auth0 Login screen presenting the CILogon option :arrow_right: CILogon IDP selection page :arrow_right: **IDP login page** :arrow_right: hub access

If on the screen in bold, the same identity provider is used, let's say Google, then a logout must happen before using another email account. But if one chooses the berkeley Identity Provider, then they will be able to login with their berkeley email address without logging out. 

2. If this phrase `If a user mistakenly provides an incorrect email when logging in` means that the incorrect email resolved to a `403` error then this is what happens https://github.com/2i2c-org/infrastructure/issues/973 and a CILogon logout won't solve things.

Hope I didn't create more confusion :sweat:

sgibson91:
> 1. I'm a bit reserved using `email address` to refer to the user account.

Ok, let's definitely say something general like "account information"

> If this phrase `If a user mistakenly provides an incorrect email when logging in` means that the incorrect email resolved to a 403 error then this is what happens https://github.com/2i2c-org/infrastructure/issues/973 and a CILogon logout won't solve things.

Ah, so do we have to do something in the hub at this point? (I will check the issue in case you've answered it there)

sgibson91 (commented):
I like [this](https://github.com/2i2c-org/infrastructure/pull/974/commits/c562d67b9fc82aedcf93ceb48a8fcb7c8a343dc2), I feel it's much easier to follow. Thank you! 🙌🏻

GeorgianaElena:
Thanks @sgibson91! I tried to rephrase the note a bit. LTMK if I managed to be more clear and shorter than in the [novel above](https://github.com/2i2c-org/infrastructure/pull/974#issuecomment-1028996147) :sweat:

GeorgianaElena:
You already answered :heart: Thank you!!!!

choldgraf:
So just to make sure I understand, there are two "things to log out of":

**If a user logs into the hub with CILogon, but they use the wrong institutional address**. Then they'll need to log out of two things:

- First, log out of JupyterHub so that they don't keep hitting 403 errors. To do this, go to `https://{hub-name}/hub/logout`
- Second, log out of CILogon so that they're asked to select a new institution. To do this, go to `https://cilogon.org/logout`

When that is done, they will be able to "start fresh" and choose a new CILogon provider and/or account name.

Is that right? If so, I think we should also document this in our user docs as well, since I suspect this will happen to people.

damianavila:
> Is that right? If so, I think we should also document this in our user docs as well, since I suspect this will happen to people.

I agree with this thought.

GeorgianaElena:
> If a user logs into the hub with CILogon, but they use the wrong institutional address. Then they'll need to log out of two things:
> First, log out of JupyterHub so that they don't keep hitting 403 errors. To do this, go to https://{hub-name}/hub/logout

Yep. Checkout https://github.com/2i2c-org/pilot-homepage/pull/8 for a possible "documentation" of this step

> Second, log out of CILogon so that they're asked to select a new institution. To do this, go to https://cilogon.org/logout

Almost. So, they're asked to select an institution regardless of they go to https://cilogon.org/logout. But if they want to use the same institution, but just another address, then in order to be asked which institutional address to use, they will need to go to https://cilogon.org/logout.

> Is that right? If so, I think we should also document this in our user docs as well, since I suspect this will happen to people.

What do you think if we go with the button in https://github.com/2i2c-org/pilot-homepage/pull/8 for the first one and the note in this PR for the second one? Do you think we need to add more details in other places of the docs?

choldgraf:
I took a pass at the language to try and clarify things a little bit in these docs - however I think we will need to iterate a little bit because the instructions didn't quite work for me, so I likely got something wrong :-)

Could you take a look at my latest commit and let me know what I messed up? I think that your strategy sounds good to me as long as we can document the full "how to log out" process here as a start. In the future if the button isn't enough and people are still confused, we can make an addition to the user docs

docs/howto/configure/auth-management.md
```diff
@@ -232,3 +223,26 @@ config:
...
+
+### Switching user accounts or institutions
+
+By default, logging in with a particular user account will persist your credentials in future sessions.
+This means that you'll automatically re-use the same institutional and user account if you try to log back in.
+
+**To switch user accounts**, a user can go to the URL endpoing `https://{hub-name}/hub/logout`.
+The next time they go to the hub's landing page, they'll be asked to re-authenticate.
+
+**To switch CILogon institutions**, a user must go to the [CILogon logout page](https://cilogon.org/logout) and click the button to log out of their institutional account. When they try to log back in they should be directed to a page to select institutions once again.
```
GeorgianaElena: I believe these are actually the other way around. In practice I've havn't seen switching institution require logging out of CILogon.

GeorgianaElena (commented):
Thanks for clarifying the docs @choldgraf. I left a couple of comments with my thoughts, but also added [a commit](https://github.com/2i2c-org/infrastructure/pull/974/commits/7d48028325bfaa971a0d81ac44df5f6ec55f6e10) to modify things a bit again. Sorry for not providing suggestions directly, but I wanted to add images too to make things more clear. LTMK what you think and if they work

choldgraf:
Ah thanks for those updates @GeorgianaElena - I think they are good! I pushed a commit with some minor cleanups and clarifications as well. What do you think?

However, following the instructions didn't quite work for me. Here are the steps I followed and what happened. Can you help me debug?

1. Logged-in to demo.pilot.2i2c.cloud with Berkeley credentials
2. 403 page
2. Tap button to log out
3. Take me back to Auth0 login page
4. Choose CILogon
5. Straight to Berkeley log-in, cannot choose Identity Provider
6. Tried going to https://cilogon.org/logout/ and logging out
7. Same thing happened again

I would have expected that step 4 would have then asked me to **choose an identity provider**, but this didn't happen. Can you clarify?

GeorgianaElena:
> 6. Straight to Berkeley log-in, cannot choose Identity Provider

I was able to reproduce your workflow if I check the `remember` box. Did you checked the remember box here?
remeber

If so, then probably you need to clear the cookies. It doesn't have to do with the logout process, but it's probably worth mentioning in the docs here.

> ! I pushed a commit with some minor cleanups and clarifications as well. What do you think?

I don't see any new commit from you @choldgraf :eyes: Maybe you forgot to push it?
https://github.com/2i2c-org/infrastructure/pull/974/commits

GeorgianaElena:
I clarified the docs a bit more and mentioned this scenario when clearing the cookies is required

choldgraf:
@GeorgianaElena the cookie fix worked! Thanks for clarifying, and sorry that I forgot to push my commit before! I've just pushed another commit to clean up the language a bit. I added `sphinx-copybutton` while I was at it :-)

docs/howto/configure/auth-management.md
```diff
@@ -175,31 +175,33 @@ Presently, this involves a few more manual steps than the `auth0` setup describe
...
-Some notions about CILogon authentication worth mentioning:
+### Key terms
 
-1. **Identity Provider**.
-   This is the authentication service available through the CILogon connection.
+Some key terms about CILogon authentication worth mentioning:
 
-   When a user logs in via CILogon, they are first presented with a list of various institutions and organization that they may choose from (e.g. `UC Berkeley` or `Australia National University`).
+Identity Provider
+: The authentication service available through the CILogon connection.
```
GeorgianaElena, line 183: Had no idea about this syntax, but super cool :+1:

choldgraf (approved):
This looks good to me now!

choldgraf:
OK I am merging this one in! I think these are nice changes and we can continue updating them as we use CILogon more and learn!

PR #976, opened 2022-02-04
Remove setup-cluster-credentials deployer command

sgibson91:
As pointed out by @damianavila in #969, this doesn't work since the context manager completes by unsetting the kubeconfig :cry:


yuvipanda:
You can probably make it work by adding a `subprocess.check_call(os.environ['SHELL'], '-l')` inside the contextmanager. This will spawn a new shell with all the env vars (including KUBECONFIG) inherited, and once you quit that shell the python program will resume as usual.

sgibson91:
Screenshot 2022-02-04 at 10 56 00

\o/

PR #977, opened 2022-02-04
Ensure kubeconfig does not get reset when running setup-cluster-credentials in the deployer

sgibson91:
- Reverts https://github.com/2i2c-org/infrastructure/pull/976
- Spawns a new shell that inherits the env vars, including KUBECONFIG. You can the perform kubectl commands against your chosen cluster as normal. The python program continues as normal once the shell is quit

Is there a nice way that I can quit the shell without needing to remember to run `exit` after I'm done? Where's the best place to document this?

### DIFF
--- a/deployer/__main__.py
+++ b/deployer/__main__.py
@@ -43,6 +43,10 @@ def use_cluster_credentials(cluster_name):
         # This command will spawn a new shell with all the env vars (including
         # KUBECONFIG) inherited, and once you quit that shell the python program
         # will resume as usual.
+        # TODO: Figure out how to change the PS1 env var of the spawned shell
+        # to change the prompt to f"cluster-{cluster.spec['name']}". This will
+        # make it visually clear that the user is now operating in a different
+        # shell.
         subprocess.check_call([os.environ["SHELL"], "-l"])
 
 
--- a/deployer/__main__.py
+++ b/deployer/__main__.py
@@ -44,7 +44,6 @@ def use_cluster_credentials(cluster_name):
         # KUBECONFIG) inherited, and once you quit that shell the python program
         # will resume as usual.
         subprocess.check_call([os.environ["SHELL"], "-l"])
-        return None
 
 
 def deploy_support(cluster_name):
--- a/deployer/__main__.py
+++ b/deployer/__main__.py
@@ -18,13 +18,13 @@ from utils import decrypt_file, update_authenticator_config, print_colour
 yaml = YAML(typ="safe", pure=True)
 
 
-def setup_cluster_credentials(cluster_name):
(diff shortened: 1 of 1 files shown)


yuvipanda:
This is great :D

Hitting ctrl-d also will close the shell, but not sure how to enforce that. Perhaps we can modify the `PS1` environment variable to make a note? This is what conda / venv does...

deployer/__main__.py
```diff
@@ -18,6 +18,35 @@
...
+        cluster = Cluster(yaml.load(f))
+
+    # Cluster.auth() method has the context manager decorator so cannot call
+    # it like a normal function
+    with cluster.auth():
+        # This command will spawn a new shell with all the env vars (including
+        # KUBECONFIG) inherited, and once you quit that shell the python program
+        # will resume as usual.
+        subprocess.check_call([os.environ["SHELL"], "-l"])
+        return None
```
yuvipanda, line 47: why explicitly `return None` here? Isn't that default python behavior?

deployer/__main__.py
```diff
@@ -278,6 +307,13 @@ def main():
         parents=[base_parser],
         help="Deploy grafana dashboards to a cluster for monitoring JupyterHubs. deploy-support must be run first!",
     )
+
+    # setup-cluster-credentials subcommand
+    setup_cluster_credentials_parser = subparsers.add_parser(
+        "setup-cluster-credentials",
```
yuvipanda, line 313: Maybe consider calling it `use-cluster-credentials` or some such? `setup` makes it sound like it'll set up for general use by all shells, rather than just the shell we spawn.

deployer/__main__.py
sgibson91: Yeah, it can probably be removed now we have some actual code in this block. But when I first wrote this, I realised that `cluster.auth()` is a context manager so I needed to wrap _something_ and couldn't just leave it blank.

deployer/__main__.py
sgibson91: Done in [`b2c17aa` (#977)](https://github.com/2i2c-org/infrastructure/pull/977/commits/b2c17aa909a4434dccee637df5e59f06205269c5)

deployer/__main__.py
sgibson91: Done in [`7152233` (#977)](https://github.com/2i2c-org/infrastructure/pull/977/commits/715223389e7c13192a1eecb92a0cb5349ffd82d5)

yuvipanda (approved):
thanks a lot, @sgibson91

sgibson91:
Thanks for the pointers @yuvipanda 🎉

Issue #968, opened 2022-02-02
[New Hub] University of Texas


colliand:
### Hub Description

This is a first pilot hub for the University of Texas. The hub will be administered by James Howison ( @jameshowison ) and used by students in his course. The course is focused on databases so is deployed with a "database sidecar" strategy as suggested by @yuvipanda. 

### Community Representative(s)

@jameshowison

### Important dates

- 2021-01-29 Hub was made available
- 2021-02-09 Class starts (correct @jameshowison?(

### Target start date

2021-02-01

### Preferred Cloud Provider

No preference (default)

### Do you have your own billing account?

- [ ] Yes, I have my own billing account.

### Hub Authentication Type

GitHub Authentication (e.g., @mygithubhandle)

### Extra features you'd like to enable

- [ ] Dedicated Scalable Dask Cluster with [Dask Gateway](https://gateway.dask.org/)

### Other relevant information

James Howison has some questions about the database sidecars.

### Hub ID

utexas-demo

### Hub Cluster

pilot

### Hub URL

https://utexas-demo.pilot.2i2c.cloud/

### Hub Type

basehub (with database sidecar)

### Tasks to deploy the hub

- [ ] Engineer who will deploy the hub is assigned
- [ ] Deploy information filled in above
- [ ] Initial Hub deployment PR: <link to PR>
- [ ] Administrators able to log on
- [ ] Community Representative satisfied with hub environment
- [ ] Hub now in steady-state

jameshowison:
I'm checking it all out and getting acquainted.  Thanks!

This is an introductory class, so in the past I've used phpmyadmin for students to browse tables and to create their first tables.  That avoids having to teach the CREATE syntax and browsing the tables (rather than using a SELECT * query) helps build analogies to Excel, helping to scaffold students.

But phpmyadmin had its own issues, exposing lots of complexity. Looking at pgadmin as the equivalent in the postgres world seems even more complexity (or maybe that's just me adjusting to postgres rather than MySQL/MariaDB (which is fine, I'd been meaning to do that for years anyway).

For browsing, any thoughts on this:
https://github.com/pbugnion/jupyterlab-sql

colliand:
I am cutting a pasting comments from written by @yuvipanda on this hub from another channel:

@colliand I've now set up https://utexas-demo.pilot.2i2c.cloud/ with the postgres setup!

[...]

PR #978, opened 2022-02-04
Add pgadmin4 for UTexas hub

yuvipanda:
- Sets it up as a sidecar, so initial user setup is done easily.
  When https://github.com/yuvipanda/jupyter-pgadmin4-proxy actually
  works, we might be able to get rid of the sidecar.
- Sets the default username and password to a well known string,
  and protection is offered by jupyter-server-proxy instead.
- Upgrade to a more recent version of postgres while we are at it
 - Once deployed, user can go to https://utexas-demo.pilot.2i2c.cloud/hub/user-redirect/proxy/absolute/5050/
    (note the trailing slash) and login with jovyan@jovyan.org / jovyan
    as username / password. Proper shortcuts will require
    fixing https://github.com/yuvipanda/jupyter-launcher-shortcuts a
    bit

### DIFF
--- a/config/clusters/2i2c.cluster.yaml
+++ b/config/clusters/2i2c.cluster.yaml
@@ -508,22 +508,22 @@ hubs:
             - name: pgadmin4
               image: dpage/pgadmin4
               env:
-              # Users use this to login to pgadmin4 itself
-              # Access to that is secured with jupyter-server-proxy
-              - name: PGADMIN_DEFAULT_EMAIL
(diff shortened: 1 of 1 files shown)


yuvipanda:
I think I very clearly need to write documentation on using sidecars for these things in z2jh. I think it should probably just live in the z2jh docs.

sgibson91 (approved):
LGTM!

PR #979, opened 2022-02-04
Provide persistent storage for postgres db

yuvipanda:
- For each user, we create a 1Gi persistent standard non-ssd
  disk, and put the postgresql data on it. This way, it survives
  user pod restarts!
- 1Gi seems to be the smallest individual disk we can get. We
  can resize it up automatically if needed.
- Code adapted from common.yaml in
  https://github.com/berkeley-dsep-infra/datahub/pull/2651.


sgibson91 (approved):
LGTM!

PR #982, opened 2022-02-07
Update machine type for GPU node pools

sgibson91:
This PR changes the machine type for the GPU node pools to a family we have quota for, according to https://github.com/2i2c-org/infrastructure/issues/930#issuecomment-1028205052

It also removes the node taint and uses a label to identify gpu nodes instead


sgibson91:
This has already been deployed successfully

PR #984, opened 2022-02-07
Rename utexas-demo to utexas

yuvipanda:
This will lose the old hub, but that's ok.


yuvipanda:
This has been deployed.

choldgraf (approved):

damianavila:
> This will lose the old hub, but that's ok.

Was this hub eventually removed/deleted from the cluster?

Issue #962, opened 2022-02-01
[New Hub] University of Washington ICESAT Hackweeks


choldgraf:
### Hub Description

This is a hub for a team at the University of Washington for several upcoming ICESAT Hackweeks.

The environment should be pretty standard for a research workflow - the most complex bits here are the GitHub Teams authentication, and deploying this in an AWS cluster.

### Community Representative(s)

- @scottyhq
- ???

### Important dates

- **First Hackweek**: March 22-25, 2022
- **Hub Shutdown**: June 1, 2022

### Target start date

ASAP (so that they can git it a whirl and play around with the setup)

### Preferred Cloud Provider

Amazon Web Services

### Preferred Location of the Cloud Resources

us-west-2

### Do you have your own billing account?

- [X] Yes, I have my own billing account.

### Hub Authentication Type

Other (may not be possible, please specify in comments)

### Hub image service

**FIND OUT**. They wish to auto-deploy a Docker image, but not sure if they have one yet.

### Hub image

**FIND OUT**. They wish to auto-deploy a Docker image, but not sure if they have one yet.

### Extra features you'd like to enable

- [ ] Dedicated Scalable Dask Cluster with [Dask Gateway](https://gateway.dask.org/)

### Other relevant information

**Authentication**: They'd like GitHub Teams authentication within the following GitHub organization: https://github.com/ICESAT-2HackWeek

**Per-user resources**
- RAM: 16GB
- CPU: 4
- Home storage: 10GB

### Hub Type

Research Hub (though scalable dask cluster isn't needed)

### Tasks to deploy the hub

- [x] Engineer who will deploy the hub is assigned
- [x] Deploy information filled in above
- [x] Create AWS account using the credits voucher
- [x] Make sure all 2i2c members have access to this AWS account
- [x] Setup infrastructure (EKS, EFS, IAM) on this AWS account
- [x] Deploy support chart on this cluster
- [x] Deploy staging and prod hub on this cluster
- [x] Setup appropriate GitHub Auth

choldgraf:
Hey @scottyhq - would you mind clarifying some of the empty boxes above so that we can get this hub deployed?

consideRatio:
For the credit voucher granting AWS credits, see https://2i2c.freshdesk.com/a/tickets/73.

yuvipanda:
[...]

PR #986, opened 2022-02-08
Remove allowed_users key from uwhackweeks

sgibson91:
When using GitHub Org or Teams authentication, you **MUST NOT** provide the `allowed_users` key in the helm chart config. Since enabling `allowed_users` will block access to anyone who has not been added via the admin panel **even if** they are valid members of the team. In this case, it is **NOT TRUE** that anyone on GitHub will be able to login  without the `allowed_users` key.

This behaviour is documented in the Note block under step 3 in our docs here: https://infrastructure.2i2c.org/en/latest/howto/configure/auth-management.html#native-jupyterhub-oauthenticator-for-github-orgs-and-teams

This PR should resolve the authentication issues discussed by @scottyhq here: https://github.com/2i2c-org/infrastructure/issues/962#issuecomment-1032871676

### DIFF
--- a/config/clusters/uwhackweeks.cluster.yaml
+++ b/config/clusters/uwhackweeks.cluster.yaml
@@ -156,11 +156,9 @@ hubs:
                   - read:user
                   - read:org
               Authenticator:
-                admin_users: &users
+                admin_users:
                   - <staff_github_ids>
                   - scottyhq
-                # Without this, any GitHub user can authenticate
-                allowed_users: *users
       dask-gateway:
         extraConfig:
           idle: |


sgibson91:
This has been deployed, waiting for Scott to confirm this resolves his issue

consideRatio (approved):

choldgraf:
Are these the kinds of things we could add checks for in a validation step? Like

```
if (github teams auth):
  ...make sure that `allowed_users` is not set, raise error if it is
elif (normal auth0 auth):
  ...make sure that `allowed_users` *is* set, raise error if it is not
```

sgibson91:
@choldgraf Absolutely

consideRatio:
Nice catch @sgibson91!

Failed to make a comment as intended alongside the approval - it would be good to have an inline comment about not setting allowed_users. In the past, we learned we sometimes require it to be set alongside admin users and now we sometimes must not have it - tricky. I like the pattern of having a note about this inline in the config, even though it could be documented centrally. It could be a reference to that central documentation as well. Just some reminder note where its relevant.

PR #987, opened 2022-02-09
Check for homepage template branch name including clustername

GeorgianaElena:
This changes `templates-sync` behavior so that it searches for a https://github.com/2i2c-org/pilot-homepage branch that matches `<cluster-name>-<hub-name>` to update the homepage, before using the main branch.

Before this it would only matched `<hub-name>` which is not ok, since we have multiple hubs with the same name (staging, prod).

### DIFF
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -400,8 +400,8 @@ class Hub:
                                 dedent(
                                     f"""\
                                     while true; do git fetch origin;
-                                    if [[ $(git ls-remote --heads origin {self.spec["name"]} | wc -c) -ne 0 ]]; then
-                                        git reset --hard origin/{self.spec["name"]};
+                                    if [[ $(git ls-remote --heads origin {self.cluster}-{self.spec["name"]} | wc -c) -ne 0 ]]; then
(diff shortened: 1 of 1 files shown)


deployer/hub.py
```diff
@@ -400,8 +400,8 @@ def get_generated_config(self, auth_provider: KeyProvider, secret_key):
                                 dedent(
                                     f"""\
                                     while true; do git fetch origin;
-                                    if [[ $(git ls-remote --heads origin {self.spec["name"]} | wc -c) -ne 0 ]]; then
-                                        git reset --hard origin/{self.spec["name"]};
+                                    if [[ $(git ls-remote --heads origin {self.cluster}-{self.spec["name"]} | wc -c) -ne 0 ]]; then
```
sgibson91, line 403: I might be wrong here, but should this be `self.cluster.spec["name"]`?

GeorgianaElena: You're absolutely right! Thanks @sgibson91! I just added a commit to address it.

sgibson91 (approved):
🚀

Issue #879, opened 2021-12-07
CI/CD infrastructure for hubs that is automated, tested, and parallelized 


choldgraf:
_This is a pitch for a multi-sprint project - it uses a slightly different structure so that we can test out how it feels!_

### Problem statement

We have some basic hub deployment infrastructure set up. However this is missing a few important pieces of functionality:

- We cannot deploy hubs in parallel, which means that one hub can bottleneck all others, and that updating one hub means updating all other hubs
- Our testing infrastructure can be flaky, which makes it hard to distinguish a "true failure" from a "false failure"
- Our configuration lumps all hubs into a single config file, which makes it harder to find and refer to one hub's configuration.

### Proposed solution

We should improve our hub CI/CD infrastructure with a few specific focus areas:

- Hubs should be deployed in parallel with one another in CI (not the deployer)
- We should have automated testing infrastructure that will prevent hub deployment if it fails
- Configuration should be done via a GitHub repository, and deployment should be automatically triggered

Here's a rough idea of how this could work:

- Each cluster has a folder that contains configuration for each hub on the cluster
- Each hub gets its own configuration file, which is a valid and standalone Z2JH config.
  - Maybe there is another shared configuration file that each hub can draw from for common settings across hubs.
- Adding a new hub's configuration will trigger the deployment of a JupyterHub with that configuration on that cluster.
- We deploy staging hubs for each hub. We run some kind of test on the staging hub for common workflows, and if it fails then the production hub for that community does not deploy.
- We have a "source of truth" for the hubs that we're running that is easy to glance through, which provides important information needed to maintain them, as well as a rough status of each.

### Guidance for implementation

A few things that *must* exist in the final product:

- **Canary deployments**: Deploying a hub should be stopped if an automated test does not pass.

**Things to test for**

A few ideas to include in testing infrastructure:

- Unauthorized users cannot log in
- Authorized users can spin up an interactive server properly
- Run a notebook designed for the base environment top to bottom without errors
- Run a notebook designed for a Dask Hub environment top to bottom without errors

**No gos / out-of-scope**

- This should not become a deep dive into deploying Kubernetes on different cloud providers, unless we *really* think there's something crucial to solve there that would enable the deployment of hubs. Instead we should just assume there's a k8s cluster available and focus this on deploying hubs.

**Related issues**

These are related issues, and some (or all) may be in-scope to close as part of this project

- [ ] https://github.com/2i2c-org/infrastructure/issues/586

#### 2022-01-19

[...]

PR #988, opened 2022-02-09
Move the management of staff IDs from the deployer to the basehub chart

sgibson91:
This addresses the block I came up against for the refactor here https://github.com/2i2c-org/infrastructure/issues/879#issuecomment-1032649498 by implementing a solution proposed here https://github.com/2i2c-org/infrastructure/issues/879#issuecomment-1032902227

#### Some questions

- What is the difference between `c.JupyterHub` and `c.Authenticator`?
- In the docs, using `c.Authenticator.admin_users` says that this list is also added to the `allowed_users` set https://jupyterhub.readthedocs.io/en/stable/getting-started/authenticators-users-basics.html#configure-admins-admin-users Is that the same behaviour for `c.JupyterHub`? In cases when we don't want to set `allowed_users`, e.g. providing Teams authentication through `GitHubOAuthenticator`, how do we avoid blocking access to members by not defining `allowed_user`?

#### TODOs

- I have only updated the hubs on the 2i2c pilots cluster so far. Wanted to get a sanity check before I did too much work.


docs/topic/config.md
```diff
@@ -32,11 +32,7 @@ basis in the config for the hub under [`config/clusters`](https://github.com/2i2
...
-When a new hub is created, we use one of these two placeholders to add 2i2c staff to the hub:
 
-- `<staff_github_ids>` - If the hub uses GitHub authentication
-- `<staff_google_ids>` - If the hub uses Google OAuth authentication
+When a hub is deployed, some custom code is run as part of `jupyterhub_config.py` and the usernames corresponding to the correct staff list are added to the hub's Admin and Allowed users.
 
-When a hub is deployed, these placeholders are removed, and the usernames corresponding to the correct staff list are added to the hub's Admin and Allowed users.
-
-You can find the list of staff usernames at [`config/clusters/staff.yaml`](https://github.com/2i2c-org/infrastructure/tree/HEAD/config/clusters/staff.yaml).
+You can find the list of staff usernames in [the `basehub` helm chart](https://github.com/2i2c-org/infrastructure/tree/HEAD/helm-chart/basehub/values.yaml#L52) along with the [custom code](https://github.com/2i2c-org/infrastructure/tree/HEAD/helm-chart/basehub/values.yaml#L392) used to inject them.
```
sgibson91, line 38:
```suggestion
You can find the list of staff usernames in [the `basehub` helm chart](https://github.com/2i2c-org/infrastructure/tree/HEAD/helm-charts/basehub/values.yaml#L52) along with the [custom code](https://github.com/2i2c-org/infrastructure/tree/HEAD/helm-charts/basehub/values.yaml#L392) used to inject them.
```

helm-charts/basehub/values.yaml
```diff
@@ -369,3 +389,10 @@ jupyterhub:
           }
 
           c.KubeSpawner.environment.update(env)
+      07-2i2c-add-staff-user-ids-to-admin-users: |
+        from z2jh import get_config
+        add_staff_user_ids_to_admin_users = get_config("custom.2i2c.add_staff_user_ids_to_admin_users", False)
+        if add_staff_user_ids_to_admin_users:
+            user_id_type = get_config("custom.2i2c.add_staff_user_ids_of_type")
+            staff_user_ids = get_config(f"custom.2i2c.staff_{user_id_type}_ids", [])
+            c.JupyterHub.admin_users.extend(staff_user_ids)
```
consideRatio, line 398: Most of the time, we need to add these to allowed_users as well i think - right? If not always, maybe we should have another boolean flag.

GeorgianaElena:
> What is the difference between c.JupyterHub and c.Authenticator

I believe with the first one, you configure the hub class traitlets  vs Authenticator ones.

It might be, that at first, `admin_users` was a hub config option and it later moved to the authenticator class. But,
`c.JupyterHub.admin_users` should be deprecated now, so we should use `c.Authenticator.admin_users` I believe
https://jupyterhub.readthedocs.io/en/stable/api/app.html?highlight=admin_users#jupyterhub.app.JupyterHub.admin_users

GeorgianaElena:
> In the docs, using c.Authenticator.admin_users says that this list is also added to the allowed_users set https://jupyterhub.readthedocs.io/en/stable/getting-started/authenticators-users-basics.html#configure-admins-admin-users Is that the same behaviour for c.JupyterHub? In cases when we don't want to set allowed_users, e.g. providing Teams authentication through GitHubOAuthenticator, how do we avoid blocking access to members by not defining allowed_user?

I had no idea about this info being in the docs :confused: Hmm, I opened this https://github.com/jupyterhub/jupyterhub/issues/3525 some time ago specifically because I thought this is not true. I will double check with what happens in code, but I remember not finding any place where admins are allowed by default.
 Also, I believe in practice, you need to explicitly add the admins to the allowed list. This is what we've done for the 2i2c hubs. I might be wrong though

sgibson91:
@GeorgianaElena I did run this against 2i2c-staging and was still able to login fine and still had admin access. I'm not sure if that was because I was already in the hub database though?

helm-charts/basehub/values.yaml
sgibson91: We might be able to do something like:

```
authenticator_class = get_config("hub.config.JupyterHub.authenticator_class")
if authenticator_class == "github":
    # Unset allowed_users
else:
    c.Authenticator.extend(staff_user_ids)
```

sgibson91: Since we explicitly set the authenticator class when using GitHub Teams auth

https://github.com/2i2c-org/infrastructure/blob/e28ec98cd6f61e410d1e25b966bd303cd84bff22/config/clusters/pangeo-hubs.cluster.yaml#L84

sgibson91: It is always otherwise set to the following

https://github.com/2i2c-org/infrastructure/blob/e28ec98cd6f61e410d1e25b966bd303cd84bff22/deployer/hub.py#L459-L461

GeorgianaElena:
> @GeorgianaElena I did run this against 2i2c-staging and was still able to login fine and still had admin access. I'm not sure if that was because I was already in the hub database though?

Nooooo! I just tested this with a dummy hub and it indeed works... I'm very confused right now... sorry for spreading confusion

sgibson91:
> sorry for spreading confusion

Don't be sorry - I think there's a lot of confusion on this topic!

GeorgianaElena:
Ok, I think I have it!

Admin users are indeed added by default to the list of allowed users. But it's not done the regular way, but through the database (in order to persist changes).

What happens is that at startup, the [hub writes](https://github.com/jupyterhub/jupyterhub/blob/3ed345f4960f083b2d35c72b330254aa3cf31fe9/jupyterhub/app.py#L1944-L1946) both the [admins](https://github.com/jupyterhub/jupyterhub/blob/main/jupyterhub/app.py#L1901) and [allowed users list](https://github.com/jupyterhub/jupyterhub/blob/main/jupyterhub/app.py#L1928) to the database. The authenticator then updates this data base and uses it to update its own allowed users list.

I need to update that confusing jupyterhub issue that I opened too. Thanks for helping figuring this out @sgibson91 :heart:

consideRatio:
@GeorgianaElena aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaahhhhhhhhhhhhhhhHHH!!!

Then I won't post this issue I was writing...

---

### Must admin_users be part of allowed_users when allowed_users is set?

In [the documentation]() it sais that:

> Users in the admin set are automatically added to the user `allowed_users` set, if they are not already present.

I'm not sure I understand what that means. I have some confusion points:

1. If `allowed_users` is an empty set, it means all users will be allowed. Adding the `admin_users` set to the `allowed_users` set would restrict permissions because its no longer an empty set.
2. It looks like a user in the `admin_users` set won't lead to an outcome as if you were in the `allowed_users` set, so, you seem to be required to be part of both `allowed_admins` and `allowed_users`.
   - The [`check_allowed` function](https://github.com/jupyterhub/jupyterhub/blob/3ed345f4960f083b2d35c72b330254aa3cf31fe9/jupyterhub/auth.py#L413-L430), has a trivial logic: if `c.Authenticator.allowed_users` is falsy, all users will be allowed. If it's truthy, only the users listed in `allowed_users` are allowed.
   - The [`get_authenticated_user` function](https://github.com/jupyterhub/jupyterhub/blob/main/jupyterhub/auth.py#L496-L515) doesn't seem to decide if a user is allowed based on being part of `admin_users` or not, it only influences a boolean flag about if the user is an admin or not assuming the user was part in the first place.
    - The [`allowed_users` definition](https://github.com/jupyterhub/jupyterhub/blob/3ed345f4960f083b2d35c72b330254aa3cf31fe9/jupyterhub/auth.py#L113-L126) has a docstring that doesn't mention anything about being combined with `admin_users` in some way.

### Steps to take

- [x] Verify the issue by configuring `c.Authenticator.allowed_users = {"normal-user"}` and `c.Authenticator.admin_users = {"admin-user"}` and trying to login as the `admin-user`.
  ```python
  # it seems this allows the `admin-user` to access the jupyterhub after all...
  # ... why?
  c.JupyterHub.authenticator_class = "dummy"
  c.Authenticator.admin_users = {"admin-user"}
  c.Authenticator.allowed_users = {"normal-user"}
  ```
- [ ] Figure out why or remain confused.

sgibson91:
@GeorgianaElena You are a super star!!! 🤩

sgibson91:
So actually, instead setting `allowed_users: *admin_users`, we could actually leave `allowed_users` as an empty list and admins + those added via the UI would still have access?

sgibson91:
> I believe an empty allowed_list in the config means everyone has access. So, at least one user needs to be set through the config file for this to work. But I will double-check since I might be wrong again 😅

I was trying to work out if `empty` and `unset` had different meanings in this case

GeorgianaElena:
So, if `allowed_users` is not set in the config, then I believe everyone is allowed, regardless of what's added through the UI.

https://github.com/jupyterhub/jupyterhub/blob/bb5ec39b2fcdd3dd71ecd5ee6bf89bccaecae503/jupyterhub/app.py#L1978-L1980

> I was trying to work out if set but empty and unset had different meanings in this case

I just tried with `c.Authenticator.allowed_users = set()` and I still see this msg in the logs:

```
[I 2022-02-09 17:48:35.348 JupyterHub app:1925] Not using allowed_users. Any authenticated user will be allowed.
```

sgibson91:
Good to know, thank you

helm-charts/basehub/values.yaml
consideRatio: Superb comment!

config/clusters/2i2c.cluster.yaml
```diff
@@ -29,6 +29,9 @@ hubs:
         branch: gh-pages
       jupyterhub:
         custom:
+          2i2c:
+            add_staff_user_ids_to_admin_users: true
+            add_staff_user_ids_of_type: "google"
```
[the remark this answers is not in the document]
sgibson91: I think we can also have checks that give useful feedback about this as well once we have schema validation files for basehub and daskhub

helm-charts/basehub/values.yaml
```diff
@@ -369,3 +389,22 @@ jupyterhub:
...
+            # Check what authenticator class is set. If it's "github", we assume
+            # GitHub Orgs/Teams is being used for auth and unset allowed_users
+            # so valid members are not refused access.
+            # FIXME: This should be handled in basehub's schema validation file
+            # so that we get useful feedback about config. But at time of writing,
+            # it doesn't have one! Issue to track the creation of such files is:
+            # https://github.com/2i2c-org/infrastructure/issues/937
+            authenticator_class = get_config("hub.config.JupyterHub.authenticator_class")
+            if authenticator_class == "github":
+                c.Authenticator.allowed_users = set()
```
consideRatio, line 410: I figure we should have a print statement here as well about this being done, so we can catch onto this from the logs. I'm not confident on how logging should be done from a `hub.extraConfig` snippet - should `print` be used? I'll look it up and provide a commit for this as well.

helm-charts/basehub/values.yaml
```diff
@@ -43,6 +43,26 @@ jupyterhub:
         projectId:
       scratchBucket:
         enabled: false
+    2i2c:
+      # Should 2i2c engineering staff user IDs be injected to the admin_users
+      # configuration of the JupyterHub's authenticator by our custom
+      # jupyterhub_config.py snippet as declared in hub.extraConfig?
+      add_staff_user_ids_to_admin_users: false
+      add_staff_user_ids_of_type: ""
```
GeorgianaElena, line 51: Maybe name it as the var in the extra config? For me it wasn't clear exactly what this is until I got the extra config snippet.

```suggestion
      user_id_type: ""
```

GeorgianaElena (approved):
This looks good to me and in my opinion the changes here are a fair tradeof :rocket: to unblock this awesome restructuring effort! Thank you @sgibson91 <3

damianavila:
Really nice PR, folks! Thanks for finally clarifying the admin > allowed confusion, @GeorgianaElena.
Belated L_GREAT_TM 😉.

Issue #991, opened 2022-02-09
[New Hub] Australia National University education pilot


choldgraf:
### Hub Description

A team at the Australia National University is interested in partnering with 2i2c to host a JupyterHub for a class this semester. The environment shouldn't be too complicated, and it can run on our shared GKE cluster.

The class is for econometrics and will have about 150 students in it (though likely not all will log on at once).

Opportunity issue: https://github.com/2i2c-org/leads/issues/69

### Community Representative(s)

- @mmcky will be the community representative on this one 👋 

### Important dates

Class start: Feb 21st (more regular use on the 28th)
Class stop: June 21st

### Target start date

Feb 14th if possible, to have time to experiment

### Preferred Cloud Provider

No preference (default)

### Do you have your own billing account?

- [ ] Yes, I have my own billing account.

### Hub Authentication Type

Other (may not be possible, please specify in comments)

### Extra features you'd like to enable

- [ ] Dedicated Scalable Dask Cluster with [Dask Gateway](https://gateway.dask.org/)

### Other relevant information

**Authentication**: They'd prefer CILogon if possible. (it's not in our dropdown menu yet)
**Environment**: @mmcky should set up an environment repo by [following the instructions in this environment template](https://github.com/2i2c-org/hub-user-image-template/blob/main/README.md)

### Hub ID

anu

### Hub Cluster

`pilot.2i2c.cloud`

### Hub URL

`anu.pilot.2i2c.cloud`

### Hub Type

`educational`

### Tasks to deploy the hub

- [x] Engineer who will deploy the hub is assigned
- [x] Deploy information filled in above
- [x] Initial Hub deployment PR: https://github.com/2i2c-org/infrastructure/pull/993
- [x] Administrators able to log on
- [x] Community Representative satisfied with hub environment
- [x] Hub now in steady-state

choldgraf:
Hey @mmcky - could you help us fill in the information above to the best of your knowledge? That will help folks set up a hub for you.

mmcky:
hey @choldgraf for sure. I'll update this now.

mmcky:
> Configuration Repo

I have setup: https://github.com/QuantEcon/jupyterhub-anu-emet2007 and will populate it with configuration as I learn about the setup.

choldgraf:
> OK whoever wrote the docs are absolute legends 🙏

That would be @GeorgianaElena :-D

mmcky:
[...]

PR #993, opened 2022-02-10
Add ANU hub config

GeorgianaElena:
This adds the config for the ANU hub. Reference https://github.com/2i2c-org/infrastructure/issues/991

### DIFF
--- a/config/clusters/2i2c.cluster.yaml
+++ b/config/clusters/2i2c.cluster.yaml
@@ -534,7 +534,6 @@ hubs:
         hub:
           config:
             Authenticator:
-              allowed_users:
               username_pattern: '^(.+@2i2c\.org|.+@anu\.edu\.au|deployment-service-check)$'
   - name: utexas
     domain: utexas.pilot.2i2c.cloud
--- a/config/clusters/2i2c.cluster.yaml
+++ b/config/clusters/2i2c.cluster.yaml
@@ -132,6 +132,7 @@ hubs:
           config:
             Authenticator:
               # Only usernames matching this regex will be allowed to login into the hub
+              # ref: https://jupyterhub.readthedocs.io/en/stable/api/auth.html#jupyterhub.auth.Authenticator.username_pattern
               username_pattern: '^(.+@2i2c\.org|deployment-service-check)$'
   - name: ohw
     domain: ohw.pilot.2i2c.cloud
@@ -538,6 +539,7 @@ hubs:
               admin_users:
                 - matthew.mckay@anu.edu.au
               # Only usernames matching this regex will be allowed to login into the hub
+              # ref: https://jupyterhub.readthedocs.io/en/stable/api/auth.html#jupyterhub.auth.Authenticator.username_pattern
               username_pattern: '^(.+@2i2c\.org|.+@anu\.edu\.au|deployment-service-check)$'
   - name: utexas
     domain: utexas.pilot.2i2c.cloud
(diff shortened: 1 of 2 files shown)


GeorgianaElena:
I've just deployed this manually :rocket: I was able to login, checked out the admin list which seemed ok and also started RStudio successfully (haven't run anything there though).

So, I'll go ahead and merge this now so that it can be tested more thoroughly!

config/clusters/2i2c.cluster.yaml
```diff
@@ -495,6 +495,48 @@ hubs:
...
+            limit: 2
+          image:
+            name: quay.io/mmcky/anu-emet2007
+            tag: b7937f446fe6
+        hub:
+          config:
+            Authenticator:
+              admin_users:
+                - matthew.mckay@anu.edu.au
+              username_pattern: '^(.+@2i2c\.org|.+@anu\.edu\.au|deployment-service-check)$'
```
sgibson91, line 539: Do we need to set `allowed_users` here?

GeorgianaElena: No, we don't need the `allowed_users` list when `username_pattern` is used. Basically, `username_pattern` does the same thing like `allowed_users`, but using a regex, rather that a specific list that should be matched. Not sure what happens if you specify both.

sgibson91: Ah great! Could you add an inline comment explaining that please?

GeorgianaElena: But since admins need to match the pattern to be allowed access, I imagine any user in the `allowed_users` list would have to match too. So, I believe `username_pattern` takes precedence over both.

sgibson91: All I'm really after is clarity on when we do use `allowed_users` or not. Since we know it's a security risk for the majority of our hubs, but it can block access in some cases, e.g., GitHub Orgs/Teams Auth. So if we're _not_ explicitly listing `allowed_users`, I think it's a good idea to add a short comment explaining why so other engineers don't add it by mistake thinking their plugging a security leak :)

consideRatio: :tada: excellent, I lacked comprehension about that as well and have become very cautious about anything with these settings, so this clarification in review + inline comment is very appreciated!

GeorgianaElena: Done! Thank you :D

config/clusters/2i2c.cluster.yaml
```diff
@@ -131,6 +131,8 @@ hubs:
         hub:
           config:
             Authenticator:
+              # Only usernames matching this regex will be allowed to login into the hub
+              # ref: https://jupyterhub.readthedocs.io/en/stable/api/auth.html#jupyterhub.auth.Authenticator.username_pattern
```
sgibson91, line 135:
```suggestion
              # We do not define allowed_users here since only usernames matching this regex
              # will be allowed to login into the hub. Ref: https://jupyterhub.readthedocs.io/en/stable/api/auth.html#jupyterhub.auth.Authenticator.username_pattern
```

sgibson91 (approved):

sgibson91:
@GeorgianaElena I pushed a commit to make the comment a little bit more explicit - but otherwise this is great! Let me know what you think!

GeorgianaElena:
That's perfect! Thanks @sgibson91! I just propagated your suggestion to all of the comments

PR #995, opened 2022-02-11
Add extra env vars to GPU profile

sgibson91:
This is a fix so that tensorflow can also see the GPUs


sgibson91:
Screenshot 2022-02-11 at 11 58 45

Hooray!

Will self-merge this since it works and it's already deployed

PR #997, opened 2022-02-11
Hardcode the docs service into basehub

sgibson91:
Previously, the deployer needed to read the helm chart values of a hub, generate a token for the docs service, and insert this into the rest of the helm chart values. This became tricky when we decided to split each hub into a set of files that uniquely describe it since the deployer would then have to read in the helm chart values files and find which one, if any, the docs-service was defined in.

However, this PR to jupyterhub https://github.com/jupyterhub/jupyterhub/pull/3531 means that the z2jh helm chart can find previously generated api tokens for services, or create them if they don't exist. Hence there is no more need to generate a token within the deployer. We can hardcode the docs-service into basehub's default values the same way we do with the configurator. We do not need to specify the full namespace of the URL so long as the proxy pod is in the same namespace.

### DIFF
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -498,18 +498,6 @@ class Hub:
             "services", {}
         )["hub-health"] = {"apiToken": hub_health_token, "admin": True}
 
-        docs_token = hmac.new(
-            secret_key, f'docs-{self.spec["name"]}'.encode(), hashlib.sha256
-        ).hexdigest()
-        if (
-            "docs_service" in self.spec["config"].keys()
-            and self.spec["config"]["docs_service"]["enabled"]
-        ):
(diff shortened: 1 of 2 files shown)


sgibson91:
Hmmm, I deployed this to 2i2c staging but something isn't right

Screenshot 2022-02-11 at 14 11 32

choldgraf (approved):
This looks good to me - I don't understand the details of the implementation super well, but I think the new configuration setup is more intuitive and simpler 👍

sgibson91:
Screenshot 2022-02-11 at 14 16 42

Fixed! I misnamed the service in basehub - commit incoming!

GeorgianaElena (approved):
Yay :rocket:

consideRatio (commented):
Looks great, nice!

PR #999, opened 2022-02-14
Modernize RTD config

consideRatio:


GeorgianaElena (approved):
LGTM :rocket:

.readthedocs.yaml
```diff
@@ -1,11 +1,18 @@
...
+  tools:
+    # NOTE: Update the python version hand in hand with the requirement
+    #       specification in docs/environment.yml
+    python: "3.8"
 
-python:
-  version: 3.8
-  install: []
+sphinx:
+  configuration: docs/source/conf.py
```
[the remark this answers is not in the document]
consideRatio: Oooh nice catch!! Thanks!

damianavila:
The latest failure seems relevant... but at the same time, it seems the file is readable (cat cmd worked) so maybe this is just a hiccup from the RTD run?

.readthedocs.yaml
```diff
@@ -1,11 +1,18 @@
+# Configuration on how ReadTheDocs (RTD) builds our documentation
+# ref: https://readthedocs.org/projects/2i2c-pilot-hubs/
```
choldgraf, line 2:
```suggestion
# ref: https://readthedocs.org/projects/2i2c-pilot-hubs
```

just to re-trigger a RTD build

.readthedocs.yaml
```diff
@@ -1,11 +1,18 @@
...
-conda:
-  environment: docs/environment.yml
-
 build:
-  image: latest
+  os: ubuntu-20.04
+  tools:
+    # NOTE: Update the python version hand in hand with the requirement
+    #       specification in docs/environment.yml
+    python: "3.8"
```
choldgraf, line 12:
```suggestion
```

choldgraf: maybe we shouldn't specify python?

choldgraf:
Argh, I tried a few things but they didn't work, so reverted back to @consideRatio's original version.

I kinda feel like we are triggering a bug in `readthedocs`. The command that is failing is:

```
python env create --quiet --name 999 --file docs/environment.yml
```

I think that they want to be doing `conda env create` instead...I have no idea why `python` is being used.

sgibson91:
@choldgraf yeah, I'd say that "python" should be a "conda" instead for the command to make sense.

choldgraf:
ah, found it: https://stackoverflow.com/questions/69774313/why-does-rtd-run-python-env-create-instead-of-conda-env-create-when-building

.readthedocs.yaml
```diff
@@ -1,11 +1,18 @@
...
```
choldgraf, line 12:
```suggestion
    # NOTE: This must be mambaforge or it will mistakenly try to invoke `python` to do a conda install
    # ref: https://stackoverflow.com/questions/69774313/why-does-rtd-run-python-env-create-instead-of-conda-env-create-when-building
    python: "mambaforge-4.10"

```

.readthedocs.yaml
```diff
@@ -1,11 +1,18 @@
...
-conda:
-  environment: docs/environment.yml
-
 build:
-  image: latest
+  os: ubuntu-20.04
+  tools:
+    # NOTE: This must be mambaforge or it will mistakenly try to invoke `python` to do a conda install
+    # ref: https://stackoverflow.com/questions/69774313/why-does-rtd-run-python-env-create-instead-of-conda-env-create-when-building
+    python: "mambaforge-4.10"
```
damianavila, line 12: We may want to still use miniconda here instead of mambaforge... I am not against mamba, it is great! But we do not use it anywhere else in the codebase, AFAIK. But we do use conda.
I may be splitting hairs here but I aim for env management consistency, ja ja 😜

choldgraf, line 12:
```suggestion
    python: "miniforge-4.10"
```

choldgraf: sounds good, changed to conda

.readthedocs.yaml
```diff
@@ -1,11 +1,18 @@
...
-conda:
-  environment: docs/environment.yml
-
 build:
-  image: latest
+  os: ubuntu-20.04
+  tools:
+    # NOTE: This must be mambaforge or it will mistakenly try to invoke `python` to do a conda install
+    # ref: https://stackoverflow.com/questions/69774313/why-does-rtd-run-python-env-create-instead-of-conda-env-create-when-building
+    python: "miniforge-4.10"
```
consideRatio, line 12:
```
Problem in your project's configuration. Invalid "build.tools.python": expected one of (2.7, 3.6, 3.7, 3.8, 3.9, 3.10, pypy3.7, miniconda3-4.7, mambaforge-4.10, 3), got miniforge-4.10
```

consideRatio:
```suggestion
    python: "miniconda3-4.7"
```

consideRatio:
```suggestion
    python: "mambaforge-4.10"
```

consideRatio: I've seen a transition to mambaforge across many repos. In the dask ecosystem, pangeo, jupyter/docker-stacks, and repo2docker use micromamba now which is more experimental overall I think.

+1 for going for mambaforge over miniforge at this point in general, but it really doesn't matter really.

damianavila: I would prefer to keep miniconda until we have a consolidated story around mamba (which I believe we should eventually migrate).

damianavila: So committing you first option 😉

choldgraf:
@damianavila that Sphinx error is my fault, because we should be making releases of that theme instead of pinning to master ;-) I'll fix tomorrow

.readthedocs.yaml
```diff
@@ -1,11 +1,18 @@
...
 version: 2
 
-conda:
-  environment: docs/environment.yml
-
 build:
-  image: latest
+  os: ubuntu-20.04
+  tools:
+    # NOTE: This must be mambaforge or it will mistakenly try to invoke `python` to do a conda install
```
choldgraf, line 10:
```suggestion
    # NOTE: This must be a miniconda or it will mistakenly try to invoke `python` to do a conda install
```

choldgraf (approved):
I fixed the bug in our theme, and re-built with RTD. The build is now passing! Here's a link: https://readthedocs.org/projects/2i2c-pilot-hubs/builds/16221520/

The reason our CI/CD says it's failing is because somehow a duplicated build was triggered in RTD, but it should work fine now.

Issue #1001, opened 2022-02-14
Docs-service errors showing in logs


GeorgianaElena:
### Description of problem and opportunity to address it

When checking hub logs, I noticed there are lots of error messages related to a failed connection to the docs service

First error message is this one:
```
[I 2022-02-14 12:08:47.784 JupyterHub app:2798] Adding external service docs at http://docs-service
[E 2022-02-14 12:08:48.715 JupyterHub app:2825] Cannot connect to external service docs at http://docs-service. Is it running?
```

Then a bunch of:

```
[E 2022-02-14 12:09:48.967 JupyterHub utils:81] Unexpected error connecting to docs-service:80 [Errno -2] Name or service not known
```

### Implementation guide and constraints

This is probably happening since we moved the service config to the basehub chart. 
The *staging* hub has the service configured, so there aren't any error messages there.

We probably need to either find a way to mark the hubs we don't need this service deployed to or just enable the docs for all the hubs.

PR #1002, opened 2022-02-14
Move the docs service out of the basehub chart

GeorgianaElena:
The hub checks if the services are up periodically. So this generates a lot of messages in the logs of the hubs not enabling/configuring this service.

Fixes https://github.com/2i2c-org/infrastructure/issues/1001


consideRatio:
Is it correct that the hub starts anyhow, making the error harmless?

I think the docs service that ships with basehub - should be configured integrated with jupyterhub (here, to let jupyterhubs proxy pod proxy traffic onwards). But, of course, users of the basehub helm chart should be allowed to opt out of the docs service - and if they do, they shouldnt get errors about it in logs - so, it should be properly disabled.

What i think we should do:
1. Make sure we don't configure tech details related to the basehub charts internal workings in the end user of the basehub chart's config, such as the network destination of the docs service. I think this is what this PR does currently.
2. Make sure that a enabled flag for the basehub chart is respected - not making the jupyterhub show errors about this. To do this, i suggest hub.extraConfig is used to check our config, and if enabled/disabled docs service, ensure we dont set c.JupyterHub.services to include the docs service at all.

Outcome: error goes away, and we avoid needing all hubs wanting to use the docs service to include tech details like url: http://docs-service

helm-charts/basehub/values.yaml
```diff
@@ -409,3 +414,8 @@ jupyterhub:
             if authenticator_class == "github" and c.Authenticator.allowed_users:
                 print("WARNING: hub.config.JupyterHub.authenticator_class was set to github and c.Authenticator.allowed_users was set, custom 2i2c jupyterhub config is now resetting allowed_users to an empty set.")
                 c.Authenticator.allowed_users = set()
+      08-add-docs-service-if-enabled: |
+        from z2jh import get_config
+
+        if get_config("custom.docs_service.enabled"):
+          c.JupyterHub.services.append({"name": "docs", "url": "http://docs-service"})
```
consideRatio, line 421: I don't think its an error to use two spaces to indent in Python, but let's use four consistently.

```suggestion
            c.JupyterHub.services.append({"name": "docs", "url": "http://docs-service"})
```

helm-charts/basehub/values.yaml
```diff
@@ -30,6 +30,9 @@ inClusterNFS:
   enabled: false
   size: 100Gi
 
+docs_service:
+  enabled: false
+
```
consideRatio, line 35: This is the default value for use here:

https://github.com/2i2c-org/infrastructure/blob/5d316ac8ad1cc2faf3b5a1323214f353d5c911e4/helm-charts/basehub/templates/docs-service-deployment.yaml#L1-L4

We could also make it so, that we rely directly on `jupyterhub.custom.docs_service.enabled`, making us not duplicate code. If we do need to configure this in two locations, we should have inline comments about the need to update one if the other.

I note that there are other configuration values of relevance to docs_service, such as...

https://github.com/2i2c-org/infrastructure/blob/5d316ac8ad1cc2faf3b5a1323214f353d5c911e4/helm-charts/basehub/templates/docs-service-deployment.yaml#L26-L34

So, they need a home as well, maybe put them under `jupyterhub.custom.docs_service` as well? I'm not sure what I think makes most sense.

Could you as part of this PR add the default values for any configuration related to the docs_service (`enabled`, `branch`, `repo` --- I figure they are `false`, `""`, and `""`)

GeorgianaElena: > So, they need a home as well, maybe put them under jupyterhub.custom.docs_service as well? I'm not sure what I think makes most sense.

I asked the same thing in a comment above :sweat_smile: Let's go for that since we both thought about it.

> Could you as part of this PR add the default values for any configuration related to the docs_service (enabled, branch, repo --- I figure they are false, "", and "")

Sure!

helm-charts/basehub/values.yaml
GeorgianaElena: The thing is that [most of the code](https://github.com/2i2c-org/infrastructure/blob/master/helm-charts/basehub/values.yaml#L328) in `hub.extraConfig` is indented using two spaces, only a few use four. But agree, there should be consistency there. Do you prefer to change it all to four or two?

consideRatio: Ah, I suggest we ignore this for now then and in a standalone PR for the future maybe go make all of them indent like python code almost always is indented - with four spaces.

GeorgianaElena:
> Whitespace chomping is a technical detail of relevance in templates of various kinds, I think they exist for jinja templates as well but I'm not sure. For an example of how they work etc, see this snippet [from the Helm documentation](https://helm.sh/docs/chart_template_guide/control_structures/):

Thank you @consideRatio for sharing your knowledge about this :heart:

GeorgianaElena:
Thanks a lot for catching this @consideRatio! All references should now be updated :rocket:

PR #1003, opened 2022-02-15
Fix pgadmin4 for U Texas

yuvipanda:
pgadmin4 was using end user IP in its session cookie for
extra security - however this doesn't work in kubernetes
type environments with a lot of dynamic IPs. The documentation
suggests we turn this off, and we do.


GeorgianaElena (approved):
Makes sense to me! thanks @yuvipanda :rocket:

yuvipanda:
Thanks for the review, @GeorgianaElena!

PR #1005, opened 2022-02-17
Update jhub-client version

GeorgianaElena:
The new version comes with [a few improvements](https://github.com/Quansight/jhub-client/compare/v0.1.2...v0.1.4), that could help the current effort of improving the testing infra.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,5 +3,5 @@ chartpress
 six # temp workaround for docker==5.0.0, used by chartpress, that failed to depend on six
 ruamel.yaml
 auth0-python
-jhub-client==0.1.2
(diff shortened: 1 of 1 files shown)


damianavila:
@GeorgianaElena, looking at the diff link you posted above I do not see something new and relevant for us, am I missing something?

GeorgianaElena:
Nope! Shouldn't be anything of relevance (apart from the newlines stripping when comparing expected cell results, maybe).
But since I've noticed new versions have been released, I figured we should be up to date with it, considering we're currently updating our testing infrastructure.

yuvipanda (approved):
Thanks, @GeorgianaElena!

Issue #998, opened 2022-02-14
Persistent CI/CD failure for uwhackweeks test: dask_test_notebook.ipynb


consideRatio:
### Description of problem and opportunity to address it

As can be seen in https://github.com/2i2c-org/infrastructure/runs/5181343521?check_suite_focus=true, the uwhacksweek deployment fail.

The details of the failure is hidden though. It is related to a failure in the [dask_test_notebook.ipynb](https://github.com/2i2c-org/infrastructure/blob/master/deployer/tests/test-notebooks/daskhub/dask_test_notebook.ipynb) that is executed.

```
======================================================================= FAILURES ========================================================================
___________________________________________________________________ test_hub_healthy ____________________________________________________________________

hub_url = 'https://staging.uwhackweeks.2i2c.cloud', api_token = '009d53ad7606a562239aeaf39fce5a8774b2155004a623e07433faff387a7ecd'
notebook_dir = PosixPath('/home/erik/dev/2i2c-org/infrastructure/deployer/tests/test-notebooks/daskhub')

    @pytest.mark.asyncio
    async def test_hub_healthy(hub_url, api_token, notebook_dir):
        try:
            print(f"Starting hub {hub_url} health validation...")
            for root, directories, files in os.walk(notebook_dir, topdown=False):
                for name in files:
                    print(f"Running {name} test notebook...")
                    test_notebook_path = os.path.join(root, name)
                    await check_hub_health(hub_url, test_notebook_path, api_token)
            print(f"Hub {hub_url} is healthy!")
        except Exception as e:
            print(
                f"Hub {hub_url} not healthy! Stopping further deployments. Exception was {e}."
            )
>           raise (e)

deployer/tests/test_hub_health.py:76: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
deployer/tests/test_hub_health.py:70: in test_hub_healthy
    await check_hub_health(hub_url, test_notebook_path, api_token)
deployer/tests/test_hub_health.py:44: in check_hub_health
    await execute_notebook(
../../../py/lib/python3.8/site-packages/jhub_client/execute.py:115: in execute_notebook
[...]

PR #1006, opened 2022-02-17
Update uwhackweeks hub

GeorgianaElena:
- This updates uwhackwweks `funded_by` section as noticed by @damianavila [here](https://github.com/2i2c-org/infrastructure/pull/981#discussion_r808598988)
- Changes the helm chart used by the hub to `basehub` instead of `daskhub` (see [this comment](https://github.com/2i2c-org/infrastructure/issues/998#issuecomment-1041818555) for context)
_Note: Not sure about this last one. Waiting for confirmation if the uwhackweeks hub wasn't meant to be a daskhub as it's my understanding from the initial issue_

### DIFF
--- a/config/clusters/uwhackweeks/staging.values.yaml
+++ b/config/clusters/uwhackweeks/staging.values.yaml
@@ -32,7 +32,7 @@ basehub:
             url: https://2i2c.org
           funded_by:
             name:
-            url: https://www.openscapes.org/
+            url: https://icesat-2.hackweek.io
     singleuser:
       serviceAccountName: cloud-user-sa
       defaultUrl: /lab


damianavila (approved):
LGTM (modulo the base hub is confirmed, otherwise the URL change LGTM).

damianavila:
> Changing daskhub to basehub is a bit of a destructive change, as you're modifying what chart will be installed. I am reasonably convinced it'll just fail, and if not produce some inconsistent state.

Mmm... that is a good point given that the hub is already running and ready to work, AFAIK from #962.
But at the same time, it itches me a little bit that we are deploying resources we are not going to use.
This happened as well with openscapes...
I guess we should improve our process in the gathering of information from the community representative so we do not repeat ourselves on this sort of error.

damianavila (changes requested):
Let's keep the daskhub since changing to basehub could be a destructive action.

GeorgianaElena:
Thanks everyone for the feedback! I've dropped the commit that changed the template type and plan to merge this soon!

So, if we ever want to change the type of a running hub (basehub/daskhub), we first need to tear it down, and just deploy a new one with the desired base chart, right?

damianavila:
> So, if we ever want to change the type of a running hub (basehub/daskhub), we first need to tear it down, and just deploy a new one with the desired base chart, right?

That seems to be a reasonable recommendation (modulo you migrate the user content from the old daskhub to the new base hub, I guess).

sgibson91:
> modulo you migrate the user content from the old daskhub to the new base hub, I guess

If we're redeploying into the same cluster (which I think in most cases we would be) I _think_ we would only need to migrate the hub database, and then it should reuse the NFS storage automatically

Issue #990, opened 2022-02-09
[Decommission Hub] Justice Innovation Lab


choldgraf:
### Summary

The Justice Innovation Lab has three hubs running as well as a dedicated cluster. They no longer have use for these hubs, and so we can decommission them.

Hub URLs:

- https://justiceinnovationlab.pilot.2i2c.cloud/
- https://staging.justiceinnovationlab.2i2c.cloud/
- https://justiceinnovationlab.2i2c.cloud/

### Info

- **Community Representative:** @JILPulvino
- **Link to New Hub issue:** https://github.com/2i2c-org/infrastructure/issues/288
- **Proposed end date:** Feb 14th

#### Phase I

- [x] Confirm with Community Representative that the hub is no longer in use and it's safe to decommission
- [x] Confirm if there is any data to migrate from the hub before decommissioning
  - [ ] If yes, confirm where the data should be migrated to -> we don't delete the CR
  - [ ] Confirm a 2i2c Engineer has access to the destination in order to complete the data migration

#### Phase II - Hub Removal

1. Remove 2i2c hub
- [ ] (Optional) Migrate data from the hub
- [x] Remove hub entry from the appropriate `*.cluster.yaml` file
- [x] Remove the hub deployment
  - `helm --namespace HUB_NAME delete HUB_NAME`
  - `kubectl delete namespace HUB_NAME`
- [x] Delete the hub's authentication application on auth0 or GitHub

2. Remove justiceinnovationlab hubs (staging and prod)
- [ ] (Optional) Migrate data from the hub
- [ ] Remove hub entry from the appropriate `*.cluster.yaml` file
- [ ] Remove the hub deployment
  - `helm --namespace HUB_NAME delete HUB_NAME`
  - `kubectl delete namespace HUB_NAME`
- [ ] Delete the hub's authentication application on auth0 or GitHub

#### Phase III - Cluster Removal

_This phase is only necessary for single hub clusters._

- [x] Run `terraform plan -destroy` and `terraform apply` to destroy the cluster
- [x] Remove the following files from the repository:
  - The associated `*.cluster.yaml` file
  - The associated CI deployer key in `secrets/`
  - Remove the name of the cluster from [CI](https://github.com/2i2c-org/infrastructure/tree/HEAD/.github/workflows/deploy-hubs.yaml)

choldgraf:
[...]

PR #1007, opened 2022-02-17
Remove justiceinnovationlab hub from the 2i2c cluster

GeorgianaElena:
First part of decommissioning the justiceinovationlab hubs :arrow_right: the 2i2c hub
Reference https://github.com/2i2c-org/infrastructure/issues/990

Still need to:
- [ ] Delete helm release
- [ ] Delete k8s namespace
- [ ] Delete auth0 client

Wondering if we should also backup the data here? I don't think so, because I believe at some point it was migrated to the new cluster and hub, and since we're keeping the Container Registry there, then this won't be needed.


damianavila:
> I don't think so, because I believe at some point it was migrat

I would agree with that...

damianavila (approved):
Config removal LGTM.

sgibson91:
> Wondering if we should also backup the data here?

Nope, I asked in the issue and I think all they want to keep is the ACR :)

sgibson91 (approved):

PR #1017, opened 2022-02-22
Bump version of configurator

yuvipanda:
Brings in https://github.com/yuvipanda/jupyterhub-configurator/pull/7

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -283,7 +283,7 @@ jupyterhub:
           - --Configurator.config_file=/usr/local/etc/jupyterhub-configurator/jupyterhub_configurator_config.py
     image:
       name: quay.io/2i2c/pilot-hub
-      tag: "0.0.1-n1159.h5b045cd"
+      tag: "0.0.1-n2184.h4ff07d9"
     nodeSelector:
       hub.jupyter.org/node-purpose: core
     networkPolicy:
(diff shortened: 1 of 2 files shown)


yuvipanda:
Will need to be decoupled from the hub image soon, so we can iterate on the configurator faster / better.

GeorgianaElena (approved):
:crossed_fingers:

damianavila (approved):

choldgraf:
any reason this is still in draft mode? when will it be time to merge this?

damianavila:
@yuvipanda, do you want to do anything else here before migrating this one off the draft state?

yuvipanda:
@damianavila i think this won't actually work - right now, configurator ships with built js in the repo (https://github.com/yuvipanda/jupyterhub-configurator/blob/main/jupyterhub_configurator/static/index.js) and I think the last PR didn't actually commit the built JS

GeorgianaElena:
@yuvipanda, @damianavila , sorry about missing to add that file to the PR. I opened https://github.com/yuvipanda/jupyterhub-configurator/pull/12 to address this.

damianavila:
Do we want to also merge https://github.com/yuvipanda/jupyterhub-configurator/pull/11 before updating this one with a new tag?

consideRatio:
I'll pre-emptively close this as it seems no longer relevant.

PR #1020, opened 2022-02-22
Clarify docs on changing terraform workspaces

sgibson91:
This PR updates some language on changing terraform workspaces to be less confusing and corrects some bash commands.


GeorgianaElena (approved):
Thank you @sgibson91 :rocket: :blossom:

Issue #1012, opened 2022-02-21
Automatically update our dependencies using dependabot


sgibson91:
### Description of problem and opportunity to address it

**Context to understand the problem**

We maintain a lot of tools in this repository that have dependencies. Remembering to update them for security or bugfix reasons is often not at the top of our to-do lists and only gets there as a result of a wider problem.

**Proposed solution**

[Dependabot](https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/) is a tool (supported by GitHub) that can watch select package ecosystems for upgrades to listed dependencies and generate PRs to bump them over a given frequency. The generated PRs come with changelogs and can be managed using a range of `@dependabot` commands (that are also explained in the PR) and we can choose to ignore a given dependency bump if needed.

**What's the value and who would benefit**

We would no longer need to "remember" to make PRs like this https://github.com/2i2c-org/infrastructure/pull/1005 while still retaining flexibility on when we upgrade.

Dependabot also tracks more than just Python dependencies! It can also bump GitHub Action versions, terraform modules, and Docker base images https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#package-ecosystem

### Implementation guide and constraints

Folk are concerned that dependabot is noisy and so we should phase the rollout of dependabot managing our dependencies so we don't get an influx of of Pull Requests.

- [x] @sgibson91 will create a PR that will enable dependabot update only for the pip dependencies in our `docs` folder (since those are regularly built/tested by RTD)
   - Ensure frequency is set to weekly
   - Ignore patch updates to further reduce noise
- [x] We monitor the PRs and see how noisy we feel dependabot is being
- [ ] Decide whether to apply this to Terraform as well

choldgraf:
I'm +1 on this in theory, but have one concern to figure out: my experience with dependabot in other repositories is that it tends to generate a decent amount of noise. Maybe this is less of a problem for us because we don't have a *ton* of dependencies, but I've seen a lot of repositories with large numbers of un-merged dependabot PRs for patch versions and such. To that extent, what if we also adopt a practice like:

```
Any dependabot PR that bumps a minor or patch version should be merged quickly by any team member, so long as the tests are passing. For major version bumps, we should notify the @2i2c-org/tech-team and give a day or two for others to respond before merging.
```

This will require to believe in our testing infrastructure, but I think that this is a target we should be shooting for anyway).

[...]

PR #1022, opened 2022-02-23
Add opinionated dependabot config

sgibson91:
This adds a configuration file to enable dependency bumps for our `docs` folder via dependabot. In #1012, folks were concerned about noise and so this is a really opinionated config and those trade-offs are described in comments at the top and throughout the config file.

There will be an initial flurry of PRs when we merge this and dependabot stabilises, after that it should only run weekly on Mondays. I'm happy to take on the work of reviewing/merging those initial PRs. One reason we chose the `docs` folder is because readthedocs builds/tests our docs in every PR, and so we will quickly be able to tell if we run into issues.


choldgraf (approved):
This looks good to me - let's give it a shot!

Just a note that I discovered you can tell Dependabot to manually run. Just go to the dependabot tab of the repo (on the book theme it is https://github.com/executablebooks/sphinx-book-theme/network/updates) and you can select subsets of ecosystems to run on.

This is particularly useful to get past that initial flurry of PRs (since the bot will only keep 5 open at a time)

sgibson91:
@choldgraf Good tip! Unfortunately, that link is a 404 for me?

choldgraf:
Ah it may only be available for repo admins. Try the same url pattern with our repo?

sgibson91:
Yeah I found it. But what I'm confused about is that there is seemingly no way to navigate there via the UI - you have to go straight to the URL. The left-hand-side options also look like the old settings menu style. Maybe they're in the middle of a refresh and it'll get ported over soon?

choldgraf:
I believe you can get there from the `dependencies` link in the dropdown menu for the repo to the right:

image

then there's a dependabot tab on the next page

sgibson91:
Oof, all that stuff that used to be findable along the top and now they've hidden. Thanks GitHub 👍🏻 

Complaining aside, thank you for walking me through that

choldgraf:
it took me a few tries to find it as well haha, agreed

Issue #1021, opened 2022-02-22
[Incident] Kernel restarts and internal gateway errors for for U. Texas hub


choldgraf:
# Summary

We recently deployed the UT Austin hub, which is running a database sidecar for each student (https://github.com/2i2c-org/infrastructure/issues/968). Since the hub began use, we've gotten reports of intermittent kernel crashes affecting students. This has been difficult to reproduce, but has been happening consistently.

There was originally a FreshDesk ticket about this here: https://2i2c.freshdesk.com/a/tickets/84?lang=en

We thought we had resolved it by increasing the memory limits for each user session, but these issues have persisted. Here are the two major things that we've found:

1. Users experiencing "kernel died" errors sporadically
2. Users getting "internal gateway errors" when using pgadmin (https://github.com/2i2c-org/infrastructure/issues/968#issuecomment-1048229587)

Here's an issue where we describe some of the problems people have, and how to reproduce them:

https://github.com/2i2c-org/infrastructure/pull/1008#issuecomment-1047372853

## Hub information

- Hub URL: https://utexas.pilot.2i2c.cloud/

## Timeline (if relevant)

All times in US/Pacific.

### 2022-02-17

- @jameshowison reports that student kernels are dying via FreshDesk

### 2022-02-18

We conclude that this may be a problem with memory limits students are hitting during queries, and bump their guarantee here: https://github.com/2i2c-org/infrastructure/pull/1008

### 2022-02-21

We get reports that many students are still running into this problem. It is not reliable to replicate, but people are consistently hitting these errors.

ref: https://github.com/2i2c-org/infrastructure/pull/1008#issuecomment-1047040355

### TODO: add more as new developments happen

---

# After-action report

These sections should be filled out once we've resolved the incident and know what happened.
They should focus on the knowledge we've gained and any improvements we should take.

## What went wrong

Things that could have gone better. Ideally these should result in concrete
action items that have GitHub issues created for them and linked to under
Action items. 

## Where we got lucky

These are good things that happened to us but not because we had planned for them.

## Action items

These are only sample subheadings. Every action item should have a GitHub issue
(even a small skeleton of one) attached to it, so these do not get forgotten. These issues don't have to be in `infrastructure/`, they can be in other repositories.

### Process improvements

1. {{ summary }} [link to github issue]
2. {{ summary }} [link to github issue]

### Documentation improvements

1. {{ summary }} [link to github issue]
2. {{ summary }} [link to github issue]

### Technical improvements

[...]

PR #1023, opened 2022-02-23
Bump postgres resource limit for utexas

yuvipanda:
Increase postgresql limits by a bit - it is possible that
the kernel deaths are caused by the postgres container hitting
its memory limit and dying. The graph in
https://github.com/2i2c-org/infrastructure/issues/1021#issuecomment-1048540620
makes it look like it isn't hitting the limit, but perhaps node
memory pressure is making it get killed just past the request
(which was 64MB) - so the bump might still help.


sgibson91 (approved):
LGTM!

PR #1025, opened 2022-02-23
Update FIXME comment in deployer

sgibson91:
Add a link to an upstream issue which, upon resolution, will assist in
us cleaning up code


sgibson91:
I will self-merge this as it only an update to a comment

PR #1027, opened 2022-02-24
Secret config refactor and relative path discovery of files in the deployer

sgibson91:
### Summary

This PR moves around various config files, both encrypted and not, in order to align with the conventions laid out here https://github.com/2i2c-org/infrastructure/issues/879#issuecomment-1048722191

It also aims to break as many path dependencies as possible in the deployer. I think the only ones left are:

- those under `shared/deployer`: will be fixed when we make the deployer an installable package and this folder ships as included datafiles
- the location of the helm-charts themselves: ideas welcome on that!

### Changes in this PR

- Bundle deployer specific files ready for packaging
  - Move config/secrets.yaml and config/clusters/schema.yaml into a
    /shared/deployer folder ready for packaging the deployer library
  - Rename these files to be more descriptive and follow our new secret
    naming convention
    - `config/secrets.yaml` --> `shared/deployer/enc-auth0-credentials.secret.yaml`
    - `config/clusters/schema.yaml` --> `shared/deployer/cluster.schema.yaml`
  - Update the deployer to recognise these new filepaths
- Rename support secrets file and update deployer
  - The support chart has a secrets file. Rename it to our new convention.
    - `helm-charts/support/secrets.yaml` --> `helm-charts/support/enc-support.secrets.yaml`
  - Update the deployer script to recognise the new filepath
- Add function to discover cluster.yaml filepath
  - Add a function that will return the absolute path to a cluster.yaml
  file for a named cluster
  - The function will verify that ONE AND ONLY ONE cluster.yaml file
  exists per cluster dir
  - The function will verify that the name of the cluster defined in the
  cluster.yaml file matches the name of the parent dir, raising a
  warning if not. (Moves code previously implemented in `__main__.py` https://github.com/2i2c-org/infrastructure/pull/1011)
- Update decrypt_file func to verify encryption
  - Function now checks for the 'enc-' prefix we use to identify
  encrypted files
  - If the sops key is not found for a file, raise an error and recommend
  regenerating the file
  - Rename the function to verify_and_decrypt_file and update throughout
  deployer
- API tokens for grafana used to deploy the `jupyterhub/grafana-dashboards` repo that were previously stored in a `secrets/config/clusters/*.cluster.yaml` file have been moved to a `config/clusters/*/enc-grafana-token.secret.yaml` file
- Encrypted credential files that grant the deployer `kubectl` access to given cluster have been renamed
  - `secrets/<cluster-name>.{{ json | yaml }}` --> `config/clusters/<cluster-name>/enc-deployer-credentials.secret.{{ json | yaml }}`
  - The deployer has been updated to discover these keys relative to the `cluster.yaml` file
  - `cluster.schema.yaml` file updated to note these paths are now relative to the `cluster.yaml` file
- `.sops.yaml` config file updated: will recognise any file with `secret` anywhere in its filepath to be handled by `sops`
- `.gitignore` file updated to allow cleaning of ignored, unencrypted secrets
- `.prettierignore` file updated to not have multiple auto-formatters acting on secrets file (should have done this earlier in this PR but I didn't realise it was configured in this way - sorry!)
- Add a function to check for files that _absolutely have to exist_ for deployment to be successful (mostly files listed under `helm_chart_values_files`) and raise an error if they were not found at the expected location
- Hubs that had secret helm chart values containing, e.g., credentials for 0Auth apps had this config moved to a `config/clusters/*/enc-*.secret.values.yaml` file and this file is then explicitly listed under `helm_chart_values_files` in the `cluster.yaml` file
  - The deployer was updated to decrypt these filepaths when generating the list of values files to parse
  - This change is limited to the pangeo-hubs, utoronto and uwhackweeks clusters
- Explicitly allow for overriding domain names: We had this in place for the howard hub on the cloudbank cluster, I think by request to prevent students cheating by blocking access to the hub, and it relied on a `secret/config/clusters/cloudbank.cluster.yaml`
  - I have introduced a `domain_override_file` key for the hub items in `cluster.schema.yaml`. This provides the filepath _relative to the cluster.yaml file_ that contains the domain we want to override with. If we don't want this to be discoverable, it should be sops-encrypted, but that is not enforced in the code.
  - The deployer has been updated to handle this additional key and setting the override domain if found
- The `/secrets` directory has been removed including `secrets/config/clusters/schema.yaml` and any use of that validation file in the deployer
- Deleted credential files for clusters we don't run any more: `hackathon-alpha.json` and `pangeo-181919.json`

### TODOS

**DOCUMENT ALL THE THINGS!**

However, I thought it would be wise to do that in a second PR since this one is already mega.

I also think some broken links in docs will be generated on merge, but again, I'll fix those in a separate PR.


.prettierignore
```diff
@@ -3,9 +3,7 @@
 
 # It would be trouble if we had two automatic formatters on the same files. This
 # could happen if we would auto-format the sops generated files with prettier.
-secrets/**/*
-**/secrets.yaml
-!secrets/config/clusters/schema.yaml
+enc-*secret*
```
sgibson91, line 6:
```suggestion
enc-*
*secret*
```

deployer/hub.py
```diff
@@ -579,20 +588,17 @@ def deploy(self, auth_provider, secret_key, skip_hub_health_test=False):
...
             ]
 
             # Add on the values files
             cmd.extend(values_files)
 
-            # Add on the secret file
-            cmd.append(f"--values={secret_values_file.name}")
-
-            print_colour(f"Running {' '.join(cmd)}")
+            print_colour(f"Running {' '.join([str(c) for c in cmd])}")
```
sgibson91, line 601:
```suggestion
            # join method will fail on the PosixPath element if not transformed
            # into a string first
            print_colour(f"Running {' '.join([str(c) for c in cmd])}")
```

deployer/hub.py
```diff
@@ -522,54 +527,57 @@ def deploy(self, auth_provider, secret_key, skip_hub_health_test=False):
...
-            subprocess.check_call(["helm", "dep", "up", "daskhub"])
-        os.chdir("..")
-
-        # Check if this cluster has any secret config. If yes, read it in.
-        secret_config_path = Path(os.getcwd()).joinpath(
-            "secrets", "config", "clusters", f'{self.cluster.spec["name"]}.cluster.yaml'
-        )
+        # Support overriding domains in case we want to block access to a hub
+        #
+        # Check if this hub has an override file. If yes, apply overrides.
```
consideRatio, line 532: I think this configuration is too specific and we should have a more generalized strategy to support secret cluster.yaml configuration. But, I also think we should iterate on that later and get this PR merged sooner rather than later. Due to that, I made the comment more explicit to help me understand it better and added a fixme that represents what I think we should do in the long term.

```suggestion
        # Support overriding domain configuration in the loaded cluster.yaml via
        # a cluster.yaml specified enc-<something>.secret.yaml file that only
        # includes the domain configuration of a typical cluster.yaml file.
        #
        # Check if this hub has an override file. If yes, apply override.
        #
        # FIXME: This could could be generalized so that the cluster.yaml would allow
        #        any of this configuration to be specified in a secret file instead of a
        #        publicly readable file. We should not keep adding specific config overrides
        #        if such need occur but instead make cluster.yaml be able to link to
        #        additional secret configuration.
```

deployer/hub.py
```diff
@@ -341,9 +345,10 @@ class Hub:
     A single, deployable JupyterHub
     """
 
-    def __init__(self, cluster, spec):
+    def __init__(self, cluster, spec, config_path):
```
consideRatio, line 348: If this is the constructor for `Hub` class and we have a `Cluster` class - what does `config_path` refer to?

Here is a snippet from the Cluster class' constructor.

```python
    def __init__(self, spec, config_path):
        self.spec = spec
        self.config_path = config_path
        self.hubs = [Hub(self, hub_yaml, config_path) for hub_yaml in self.spec["hubs"]]
```

What do you think about being more verbose, and let `config_path` be named `cluster_config_path`, or perhaps rely on the cluster object like `self.cluster.config_path` etc?

deployer/hub.py
```diff
@@ -24,9 +28,10 @@ class Cluster:
     A single k8s cluster we can deploy to
     """
 
-    def __init__(self, spec):
+    def __init__(self, spec, config_path):
         self.spec = spec
-        self.hubs = [Hub(self, hub_yaml) for hub_yaml in self.spec["hubs"]]
+        self.config_path = config_path
+        self.hubs = [Hub(self, hub_yaml, config_path) for hub_yaml in self.spec["hubs"]]
```
consideRatio, line 34: Mentioning `hub_yaml` is a bit misleading here I think, since has already been loaded as a dictionary from a YAML string. The distinction I'm making is that YAML implies the content is a string, but it doesn't appear to be a valid YAML string but a Python dictionary.

```suggestion
        self.hubs = [Hub(self, hub_spec, config_path) for hub_spec in self.spec["hubs"]]
```

consideRatio (commented):
Beautiful!! This looks very solid to me and if it deploys without erroring like this I'd say this is ready to be merged!

I made some comments with smaller refactoring followups etc that overall isn't blocking.

Feel free to ping me on slack to get expedited feedback on any response/update with regards to this PR!

GeorgianaElena (approved):
Thanks for all this work @sgibson91 :rocket:

deployer/hub.py
sgibson91: Yeah, I went back and forth on this change and whether to make it general or not. On the one hand, we only do this for one hub, on the other hand, now we have an explicit list of helm chart values files (including secret ones), there's not much left at the cluster.yaml level we'd want to explicitly override.

- For `auth0`, we override by setting `auth0.enabled = false` and provide alternative auth config using a JupyterHub native OAuthenticator in the helm chart values, which would be explicitly listed
- For the hub `name` (helm namespace and release name), I'm not sure why we would ever want to override this?
- `helm_chart` we can't override without deleting a release and reinstalling

So I agree it's specific, but I don't know if we'd get much more benefit from a completely generalised version of this?

config/clusters/cloudbank/cluster.yaml
```diff
@@ -55,6 +55,7 @@ hubs:
       - elcamino.values.yaml
   - name: howard
     domain: howard.cloudbank.2i2c.cloud
+    domain_override_file: enc-howard.domain-override.secret.yaml
```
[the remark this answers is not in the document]
sgibson91: In the docs PR (to be done separately to this one), I was going to write up proper documentation about this. It's on my todo list for this week.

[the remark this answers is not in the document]
GeorgianaElena: Super! Thank you!

deployer/hub.py
sgibson91: Yeah, this telling to how inexperienced I am at working with Python classes 😁 Relying on the Cluster object to propagate the path the the cluster.yaml file didn't occur to me as a possibility

sgibson91: Done in [bacd6e0](https://github.com/2i2c-org/infrastructure/pull/1027/commits/bacd6e04b17131a1501c38bed0e687e8ef3fe01c)

sgibson91:
I just successfully ran this against the 2i2c staging hub so I will hit merge and we can iterate on other things elsewhere!

consideRatio:
@sgibson91 wiee nice! I read your comments on my comments and it all makes sense to me! :+1: for the merge!

sgibson91:
There was some bugs on merge

1. The farallon deployer credentials file wasn't named correctly: https://github.com/2i2c-org/infrastructure/pull/1036
2. For pangeo-hubs, uwhackweeks, and utoronto, there is an issue reading in the secret helm chart values files. Error:

```
1 / 2: Deploying hub staging...
Getting updates for unmanaged Helm repositories...
...Successfully got an update from the "https://jupyterhub.github.io/helm-chart/" chart repository
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "kvaps" chart repository
...Successfully got an update from the "jetstack" chart repository
Update Complete. ⎈Happy Helming!⎈
Saving 1 charts
Downloading jupyterhub from repo https://jupyterhub.github.io/helm-chart/
Deleting outdated charts
Running helm upgrade --install --create-namespace --wait --namespace=staging staging /Users/sgibson/source/github/2i2c-org/pilot-hubs/helm-charts/basehub --values=/var/folders/hb/3kw2n13d17l7rnljn9jgplr00000gn/T/tmp14nt1lt2 --values=/Users/sgibson/source/github/2i2c-org/pilot-hubs/config/clusters/utoronto/staging.values.yaml --values=/var/folders/hb/3kw2n13d17l7rnljn9jgplr00000gn/T/tmpwru2e5dp
Error: open /var/folders/hb/3kw2n13d17l7rnljn9jgplr00000gn/T/tmpwru2e5dp: no such file or directory
  File "/usr/local/Caskroom/miniconda/base/envs/pilot-hubs/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/Caskroom/miniconda/base/envs/pilot-hubs/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/sgibson/source/github/2i2c-org/pilot-hubs/deployer/__main__.py", line 324, in <module>
    main()
  File "/Users/sgibson/source/github/2i2c-org/pilot-hubs/deployer/__main__.py", line 307, in main
    deploy(
  File "/Users/sgibson/source/github/2i2c-org/pilot-hubs/deployer/__main__.py", line 215, in deploy
    hub.deploy(k, SECRET_KEY, skip_hub_health_test)
  File "/Users/sgibson/source/github/2i2c-org/pilot-hubs/deployer/hub.py", line 611, in deploy
    subprocess.check_call(cmd)
  File "/usr/local/Caskroom/miniconda/base/envs/pilot-hubs/lib/python3.9/subprocess.py", line 373, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['helm', 'upgrade', '--install', '--create-namespace', '--wait', '--namespace=staging', 'staging', PosixPath('/Users/sgibson/source/github/2i2c-org/pilot-hubs/helm-charts/basehub'), '--values=/var/folders/hb/3kw2n13d17l7rnljn9jgplr00000gn/T/tmp14nt1lt2', '--values=/Users/sgibson/source/github/2i2c-org/pilot-hubs/config/clusters/utoronto/staging.values.yaml', '--values=/var/folders/hb/3kw2n13d17l7rnljn9jgplr00000gn/T/tmpwru2e5dp']' returned non-zero exit status 1.
```

.gitignore
```diff
@@ -1,4 +1,21 @@
 # Added by 2i2c
+#
+# To clean up ignored files you can use: git clean -Xfd
+# Delete untracked files (-X), with required confirmation (-f), recursively (-d).
+
+# This section handles sensitive content that is or isn't encrypted in our repo.
+#
+# Ignore files (but not directories) with secret in their name
+*secret*
```
choldgraf, line 9: I was gonna note that we might hit situations where a community has the word "secret" in it, but then I saw that this only affects files and not folders, so I think we are OK there. (that is clever!)

choldgraf:
I just did a `tree` of the `config/` directory, and I am really liking it :-) very easy to find where a hub's configuration is and glance at the contents of folders.

Two quick questions, not sure if this is a good place for feedback so direct me elsewhere if you like)

1. In some cases, there is a `staging.values.yaml` but no corresponding `prod.values.yaml` (openscapes is one-such config). For example:
   
	  ```
	      ├── openscapes
	      │   ├── cluster.yaml
	      │   ├── enc-deployer-credentials.secret.json
	      │   └── staging.values.yaml
	  ```
	  
	  What's up with that? Does this mean that the prod/staging configs are identical, and if we want to have staging-specific config we need to create a new file?
2. For multi-hub clusters, I found it a bit easy to lose the cluster credentials files when they are sorted alphabetically (e.g. you might see files that start with `elcamino`, `enc-deployer`, `howard`). What do you think about appending a prefix to the meta config files (even if it's just `_` or something) so they are naturally grouped?

sgibson91:
> 1. What's up with that? Does this mean that the prod/staging configs are identical, and if we want to have staging-specific config we need to create a new file?

Yes. Though I think you mean, if we want to have _prod_-specific config. See the `pangeo-hubs` directory as an example of this. This is basically replacing the anchoring behaviour we were using before to prevent repeated config. Also because of how helm merges config, we can very simply change one value for a prod deployment vs staging deployment very easily, without a lot of copy-pasta.

> 2\. For multi-hub clusters, I found it a bit easy to lose the cluster credentials files when they are sorted alphabetically (e.g. you might see files that start with `elcamino`, `enc-deployer`, `howard` next to each other, but even though `enc-deployer` is a different kind of config). What do you think about appending a prefix to the meta config files (even if it's just `_` or something) so they are naturally grouped?

Don't know. Happy for others in @2i2c-org/tech-team to chime in here

choldgraf:
After reflection, unless somebody things that prepending a common prefix for "meta configuration" is really important, we should just go with the status quo for a bit and see if it feels confusing to people over time. I don't wanna prematurely optimize :-)

sgibson91:
@choldgraf The next step is to configure the deployer to be able to generate a list of input files from a cluster-name hub-name pairing, and the reverse (working towards optimising CI). So perhaps the answer will be "don't look up the folder, use this deployer function instead"

damianavila:
Lovely work @sgibson91, thanks!!!

PR #1028, opened 2022-02-24
Fix typo in dependabot config

sgibson91:
update-types should be of type array


sgibson91:
Will self-merge this as it's just a typo

Issue #967, opened 2022-02-01
Investigate using CILogon instead of Auth0


GeorgianaElena:
As part of https://github.com/2i2c-org/infrastructure/issues/315, we decided to Investigate a deeper integration with CILogon - to replace our current Auth0 setup.

## Current setup:
CILogon is configured as an Auth0 social connection as described [here](https://www.cilogon.org/auth0) and then registered as a [CILogon](https://cilogon.org/oauth2/register) client. This means, that for any 2i2c hubs that enable login through CILogon (like the [2i2c demo hub](demo.pilot.2i2c.cloud)) the authentication flow is:

**Hub login page** :arrow_right: Auth0 **Login screen presenting the CILogon option** :arrow_right: **CILogon IDP selection page** :arrow_right: **IDP login page** :arrow_right: hub access

## Context
Jim from the CILogon project suggested that it would be possible to use a paid CILogin API to automate client creation. We would require a 12.5k USD/year subscription for access to this API. This also comes with other features described here https://www.cilogon.org/subscribe. Docs about the API here https://cilogon.github.io/oa4mp/server/manuals/dynamic-client-registration.html.

## Would using CILogon instead of Auth0 simplify our login process + make it more flexible?
- CILogon is a very cool project and is generally more aligned with 2i2c mission and needs (focused on research and academia , while Auth0 is more enterprise oriented and logging in through social connections)
- We won't need to maintain both Auth0 setup and subscription and the CILogon one (GitHub & Google auth can be achieved through CILogon)
- We can use the [JupyterHub CILogon oauthenticator](https://github.com/jupyterhub/oauthenticator/blob/main/oauthenticator/cilogon.py), so we we'd be able to specify list of [allowed idps](https://github.com/jupyterhub/oauthenticator/blob/main/oauthenticator/cilogon.py#L89) instead of using `username_pattern` to provide authentication by a given domain.
**HOWEVER**
- The cheapest CILogon  subscription that allows automating client creation seems expensive (probably also because we dont yet need all the cool features the come with [the subscription](https://www.cilogon.org/subscribe))
- We would need to change the current infra for creating OAuth2 Clients from [`auth.py`](https://github.com/2i2c-org/infrastructure/blob/master/deployer/auth.py) - that uses the Auth0 API to using the CILogon REST API.

## Does CILogon have all of the same features that Auth0 does? What are the differences?
[...]

PR #1033, opened 2022-02-28
Save cilogon administrative client creds

GeorgianaElena:
I've filled in the form in https://cilogon.org/oauth2/admin-register (`2i2c` client name and I used the support email) and it generated this creds. Once the client gets approved we should be able to use them. But until then, let's store them somewhere safe (the old ones will probably become obsolete, but let's keep them both until we figure out the new auth infra ?? )

Ref: https://github.com/2i2c-org/infrastructure/issues/967


sgibson91:
Hey @GeorgianaElena, this PR will create a conflict with https://github.com/2i2c-org/infrastructure/pull/1027

I suggest:

1. Wait til #1027 is merged
2. Rename `shared/deployer/enc-auth0-credentials.secret.yaml` to something sensible: `shared/deployer/enc-auth-provider-credentials.secret.yaml`?
3. Add the CILogon credentials to that file

Does that sound OK?

GeorgianaElena:
@sgibson91, of course!

GeorgianaElena:
I've rebased the PR @sgibson91. Also, this new client has recently been approved by CILogon folks :rocket: We just need to figure out how to use it

sgibson91 (approved):
@GeorgianaElena I pushed a commit to update the hard-coded path in the deployer, otherwise this is great - thank you!!

choldgraf (approved):
Just to confirm - this is embedding the CILogon API credentials in our secrets, so that we could re-use them in subsequent iterations in setting up CILogon with our stack. Is that right? That sounds good to me!

PR #1035, opened 2022-02-28
Pin gcp provider versions

yuvipanda:
Otherwise the versions on different people's machines might not
always match - I think my `terraform plan` was behaving differently
than it does on other people's.


sgibson91 (approved):
Thanks @yuvipanda!

yuvipanda:
Thanks, @sgibson91!

PR #1042, opened 2022-03-01
Use ExitStack to generate a variable number of context managers

sgibson91:
This allows us to persist temporary files that contain decrypted helm chart values files and fixes the bug raised here https://github.com/2i2c-org/infrastructure/issues/879#issuecomment-1054444375

### DIFF
--- a/deployer/__main__.py
+++ b/deployer/__main__.py
@@ -93,12 +93,6 @@ def deploy_grafana_dashboards(cluster_name):
         "enc-grafana-token.secret.yaml"
     )
 
-    # Check the secret file exists before continuing
-    if not os.path.exists(grafana_token_file):
-        raise FileExistsError(
-            f"File does not exist! Please create it and try again: {grafana_token_file}"
-        )
-
     # Read the cluster specific secret config file
     with verify_and_decrypt_file(grafana_token_file) as decrypted_file_path:
         with open(decrypted_file_path) as f:
--- a/deployer/__main__.py
+++ b/deployer/__main__.py
@@ -93,7 +93,7 @@ def deploy_grafana_dashboards(cluster_name):
         "enc-grafana-token.secret.yaml"
     )
 
-    # Read the cluster specific secret config file
+    # Read the cluster specific secret grafana token file
     with verify_and_decrypt_file(grafana_token_file) as decrypted_file_path:
         with open(decrypted_file_path) as f:
             config = yaml.load(f)
--- a/deployer/__main__.py
+++ b/deployer/__main__.py
@@ -13,7 +13,7 @@ import shutil
 from auth import KeyProvider
 from hub import Cluster
 from utils import (
-    verify_and_decrypt_file,
+    get_decrypted_file,
     print_colour,
     find_absolute_path_to_cluster_file,
 )
@@ -94,7 +94,7 @@ def deploy_grafana_dashboards(cluster_name):
     )
 
     # Read the cluster specific secret grafana token file
-    with verify_and_decrypt_file(grafana_token_file) as decrypted_file_path:
+    with get_decrypted_file(grafana_token_file) as decrypted_file_path:
         with open(decrypted_file_path) as f:
             config = yaml.load(f)
 
@@ -169,7 +169,7 @@ def deploy(cluster_name, hub_name, skip_hub_health_test, config_path):
     # Validate our config with JSON Schema first before continuing
     validate(cluster_name)
 
-    with verify_and_decrypt_file(config_path) as decrypted_file_path:
+    with get_decrypted_file(config_path) as decrypted_file_path:
         with open(decrypted_file_path) as f:
             config = yaml.load(f)
 
--- a/deployer/utils.py
+++ b/deployer/utils.py
@@ -142,7 +142,6 @@ def get_decrypted_files(files, abspath):
         yield [
             stack.enter_context(verify_and_decrypt_file(abspath.joinpath(f)))
             for f in files
-            if assert_file_exists(abspath.joinpath(f))  # Check the file exists
         ]
 
 
--- a/deployer/utils.py
+++ b/deployer/utils.py
@@ -87,8 +87,10 @@ def verify_and_decrypt_file(encrypted_path):
     filename = os.path.basename(encrypted_path)
     _, ext = os.path.splitext(filename)
 
-    # Our convention is that encrypted secrets in the repository begin with "enc-",
-    # so first we check for that
+    # Our convention is that encrypted secrets in the repository begin with "enc-" and include
+    # "secret" in the filename, so first we check for that. We use an 'or' conditional here since
+    # we want to catch files that contain "secret" but do not have the "enc-" prefix and ensure
+    # they are encrypted, raising an error if not.
     if filename.startswith("enc-") or ("secret" in filename):
         # We must then determine if the file is using sops
         # sops files are JSON/YAML with a `sops` key. So we first check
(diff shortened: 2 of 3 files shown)


sgibson91:
I tested this by deploying to pangeo-hubs staging and it works!

deployer/utils.py
```diff
@@ -130,6 +135,21 @@ def verify_and_decrypt_file(encrypted_path):
...
+    """
+    This is a context manager that when entered provides a list of
+    file paths, where temporary files have been created if needed for
+    files that were encrypted and first need to be decrypted.
+    """
+    with ExitStack() as stack:
+        yield [
+            stack.enter_context(verify_and_decrypt_file(abspath.joinpath(f)))
+            for f in files
+            if check_file_exists(abspath.joinpath(f))  # Check the file exists
```
consideRatio, line 149: Hmmm, this is always true or error - so no filtering is done. You are not really checking you are asserting.

What do you think about renaming the function to be "assert" file exist maybe? That would read more clear to me because I read this as you may end up filter out files that doesn't exist which to me should instead raise an error.

I also wonder, if we have a function named "verify_and_decrypt_file" - it becomes very surprising if that would just cleanly return a file path that didn't exist - but only verify that encrypted files wasn't decrypted already.

I considered if "verify" should be included in the function name - it is describing what it does in a way, but fail to do so in clear enough detail to really help me understand. If we think it is helpful, then in a way, `get_decrypted_files` should perhaps be named similarly, because it also verifies files in a way after all - but it relies on the helper functions - both for encrypted/not-encrypted files.

Concrete ideas:
1. Rename `check_file_exists` to `assert_file_exists`.
2. Let `assert_file_exists` run for the `verify_and_decrypt_file` function to assert a file to exist no matter what as its first verification check.

deployer/__main__.py
```diff
@@ -93,12 +93,6 @@ def deploy_grafana_dashboards(cluster_name):
         "enc-grafana-token.secret.yaml"
     )
 
-    # Check the secret file exists before continuing
-    if not os.path.exists(grafana_token_file):
-        raise FileExistsError(
-            f"File does not exist! Please create it and try again: {grafana_token_file}"
-        )
-
     # Read the cluster specific secret config file
```
sgibson91, line 96:
```suggestion
    # Read the cluster specific secret grafana token file
```

deployer/utils.py
sgibson91: Done in most recent commits

deployer/utils.py
```diff
@@ -130,6 +131,20 @@ def verify_and_decrypt_file(encrypted_path):
         return
 
 
+@contextmanager
+def get_decrypted_files(files, abspath):
+    """
+    This is a context manager that when entered provides a list of
+    file paths, where temporary files have been created if needed for
+    files that were encrypted and first need to be decrypted.
```
GeorgianaElena, line 139:
```suggestion
    This is a context manager that when entered provides a list of
    file paths to temporary files that hold the decrypted version of
    encrypted files in `files`.
```
Question: if any file in the `files` list is not encrypted, what happens? Do we raise an error, or we just leave it as it is and add it to the list of decrypted files?

deployer/utils.py
```diff
@@ -11,16 +11,13 @@
...
     such as, files listed in the `helm_chart_values_file` key in the `cluster.yaml` file
 
     Args:
-        filepath (str): Absolute path to the file that is to be checked for existence
-
-    Returns:
-        bool: Returns True if the function has not raised an error
+        filepath (str): Absolute path to the file that is to be asserted for existence
     """
     if not os.path.exists(filepath):
```
consideRatio, line 22: I think `os.path.exists` is a bug we _maybe_ never would have noticed, because it would accept a directory path as well. Since the function is named `assert_file_exists` rather than `assert_path_exists` or similar.

```suggestion
    if not os.path.isfile(filepath):
```

deployer/utils.py
sgibson91: https://github.com/2i2c-org/infrastructure/blob/cc32b3c71b532ee83d3a63939cbeb55afbd6e8af/deployer/utils.py#L89-L116

deployer/utils.py
```diff
@@ -88,6 +83,7 @@ def verify_and_decrypt_file(encrypted_path):
             decrypted contents. Unless the file is not valid JSON/YAML or does not have
             the prefix `enc-`, then we return the original, encrypted path.
     """
+    assert_file_exists(encrypted_path)
```
consideRatio, line 86: The functions parameter is named encrypted_path and has a docstring that doesn't indicate it would accept a non-encrypted file.

I think it makes sense to update the function name/parameter name in line with `get_decrypted_files` to for example `get_decrypted_file` and clarify that it will: a) return a file path as it is if its not encrypted or b) return a temporary decrypted file if it was encrypted, and c) run verifications that the file exists and wasn't already decrypted if it had certain naming conventions to indicate it should be encrypted.

deployer/utils.py
GeorgianaElena: @sgibson91, I believe we should modify the function description a bit. Otherwise we're describing what `verify_and_decrypt` does. 

```suggestion
    This is a context manager that combines multiple `verify_and_decrypt_file`
    context managers that open and/or decrypt the files in `files`.
```

deployer/utils.py
```diff
@@ -83,12 +83,13 @@ def verify_and_decrypt_file(encrypted_path):
...
     filename = os.path.basename(encrypted_path)
     _, ext = os.path.splitext(filename)
 
     # Our convention is that encrypted secrets in the repository begin with "enc-",
     # so first we check for that
```
sgibson91, line 91:
```suggestion
    # Our convention is that encrypted secrets in the repository begin with "enc-" and include
    # "secret" in the filename, so first we check for that. We use an 'or' conditional here since
    # we want to catch files that contain "secret" but do not have the "enc-" prefix and ensure
    # they are encrypted, raising an error if not.
```

deployer/utils.py
```diff
@@ -95,23 +99,23 @@ def verify_and_decrypt_file(encrypted_path):
...
                 try:
-                    encrypted_data = json.load(f)
+                    content = json.load(f)
                 except json.JSONDecodeError:
-                    yield encrypted_path
+                    yield original_filepath
                     return
 
-        if "sops" not in encrypted_data:
+        if "sops" not in content:
```
consideRatio, line 118: I think right now, this function assumes our encrypted files are .json or .yaml - which is what we so far need. It doesn't indicate anything about this assumption in the docstring though, and it could be something of relevance in the future.

I think adding a hard failure with a FIXME note on being passed seemingly encrypted files that aren't .yml/.yaml/.json could make sense at this point.

sgibson91: I feel like we are approaching feature creep on this PR now. This was opened as a bug fix so we can reenable deployments to three clusters that were broken after merging #1027. General improvements like this can be addressed in a new PR.

GeorgianaElena (approved):
LGTM! :rocket:

deployer/utils.py
consideRatio: Yepp =/ I felt that too, sorry - let's go for a merge!

consideRatio (approved):

sgibson91:
Thank you @GeorgianaElena and @consideRatio for your reviews! I did one last manual deploy to pangeo-hubs staging to check we are still working and will now hit merge!

damianavila:
Nice, @sgibson91. TIL about ExitStack!

PR #1043, opened 2022-03-01
Disable auth0 for uwhackweeks hub

GeorgianaElena:
Since the uwhackweeks hub uses the JupyterHub GitHubOAuthenticator and not auth0, then I believe auth0 should be disabled.
https://github.com/2i2c-org/infrastructure/blob/HEAD/config/clusters/uwhackweeks/staging.values.yaml#L113-L115

Not sure exactly why it works right now, but maybe it's because of how the config gets merged and overwritten?

### DIFF
--- a/config/clusters/uwhackweeks/cluster.yaml
+++ b/config/clusters/uwhackweeks/cluster.yaml
@@ -34,9 +34,7 @@ hubs:
     domain: staging.uwhackweeks.2i2c.cloud
     helm_chart: daskhub
     auth0:
-      # connection update? Also ensure the basehub Helm chart is provided a
-      # matching value for jupyterhub.custom.2i2c.add_staff_user_ids_of_type!
-      connection: github
+      enabled: false
     helm_chart_values_files:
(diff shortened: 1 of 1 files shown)


choldgraf:
Also just a note that the health check for this hub is failing right now (https://github.com/2i2c-org/infrastructure/runs/5376070604?check_suite_focus=true) no idea if it's related to auth0 or not, but just noting in case it's relevant.

sgibson91:
@choldgraf That is because of this issue https://github.com/2i2c-org/infrastructure/issues/998

damianavila:
> We should investigate how that was allowed to happen

Is there any follow-up issue opened for this investigation, @sgibson91 @GeorgianaElena?

sgibson91:
> Is there any follow-up issue opened for this investigation, @sgibson91 @GeorgianaElena?

Yes https://github.com/2i2c-org/infrastructure/issues/1125

damianavila:
Thanks!

PR #1044, opened 2022-03-01
Update our documentation

sgibson91:
This PR refreshes our documentation to include all the changes that happened in #1027

Issue #716, opened 2021-09-28
Add a conditional check for whether to run `helm dep up` in our deploy scripts


choldgraf:
### Description

In closing https://github.com/2i2c-org/pilot-hubs/issues/556, we've added an automatic step to run `helm dep up` as part of our deploys, to ensure that we don't have outdated local charts. One drawback is that this slows down the deploys a bit, even though this step may not always be needed.

To resolve this, we can add a conditional check to decide if the local version of the chart is out of date with latest version we need. Only if the answer is "yes" do we run `helm dep up`.

It's unclear how much benefit this would give us, since the deployment process already takes some time. So let's **time box this and decide by January 2022 whether we really need this change**.

### Value / benefit

The main benefit is that this would save us some time when doing deploys, in cases where we unnecessarily run `helm dep up`.

sgibson91:
Yeah, this is because when I was working in the deployer, I ended up with a circular dependency when I separated out various functions into their own files: https://github.com/2i2c-org/infrastructure/pull/1054

Issue #556, opened 2021-07-27
Automate deleting old charts and updating dependencies


GeorgianaElena:
### Summary

When trying to manually deploy a new 2i2c hub, I ran into a bug caused by our rename from `base-hub` to `basehub` from some time ago and the presence of an outdated chart with the old name.

To prevent this from repeating, when manually deploying a hub, we should automatically:
- if there's a version mismatch
  - delete the `charts/` dir
- run `helm dep up`
- continue deploying the hub/s 

### Benefit

This will reduce engineer confusion if there's a mismatch between local charts + whatever is being used in our deployer script. We've already had a few reports where this bug bit people and made them lose a few hours so this would resolve that.

### Implementation details

This will be an update to our `deployer.py` tool here: https://github.com/2i2c-org/pilot-hubs/tree/master/deployer

We should first just automatically run `helm dep up` as a part of our deploys, as a stop-gap solution to this problem. After that, we should decide how to add in a check for whether we *need* to run `helm dep up`.

### Tasks

- [x] https://github.com/2i2c-org/pilot-hubs/pull/705
- [x] Decide whether to add the conditional check as part of this issue or to open a new medium-term issue

damianavila:
OK, this one has bitten me again I am looking at it a little bit more closely...

The first item is proposing:

> if there's a version mismatch
> delete the charts/ dir

When I look into the `helm dep list` in my local repo for the basehub and daskhub charts:

```
helm dep list
NAME      	VERSION	REPOSITORY                              	STATUS       
jupyterhub	1.1.3  	https://jupyterhub.github.io/helm-chart/	wrong version
```

```
helm dep list
NAME        	VERSION	REPOSITORY                              	STATUS       
basehub     	0.1.0  	file://../basehub                       	wrong version
dask-gateway	0.9.0  	https://dask.org/dask-gateway-helm-repo/	ok     
```      

This is great because we see there is a wrong version!
But if I try to get the output as json

```
helm dep list --output json
Error: unknown flag: --output
```

Buuu...
And a big PR for that seems to be stale: https://github.com/helm/helm/pull/8077

Since we are unconditionally updating the support chart here: https://github.com/2i2c-org/pilot-hubs/blob/master/deployer/hub.py#L109

[...]

PR #1045, opened 2022-03-02
Chart schemas, deployer script validation, and ci integration

### LINKED ISSUE #279: Helm template validation in our CI/CD
###  Description

Rendering Helm templates is a cheap and easy test that I think would help us catch some potential problems.

We could also use the `kubeval` plugin for helm or the more rigorous `--validate` flag, but the latter requires access to a k8s cluster - either the one we actually use or a local CI cluster. These help us validate the rendered templates themselves.

### Benefit

Doing this would ensure that we don't have any configuration in our charts that is going to clash with what Kubernetes expects - by validating quickly we can fail early rather than later on as a part of a deployment.

### Implementation details

We could add a GitHub action that does a simple [`helm template` command](https://helm.sh/docs/helm/helm_template/). This wouldn't be a rigorous validation, but would be a way to quickly check that the configuration was reasonable and didn't have errors. 

We could use dummy values such as `proxy.secretToken` if needed to render without our secret config.

For a more rigorous test, we could also investigate using [the `helm kubeval` plugin](https://github.com/instrumenta/helm-kubeval). 

Here's how it is used in the mybinder.org deployment:

https://github.com/jupyterhub/mybinder.org-deploy/blob/2603737e02c6a32632a06ab0bad6608235cc86c0/.github/workflows/lint-validate.yml#L93-L136

### Tasks to complete

- [x] Decide which of the two approaches above is best to implement
- [ ] Implement the ideas in https://github.com/2i2c-org/infrastructure/issues/279#issuecomment-978019478
- [ ] Add GitHub action to do this!

### LINKED ISSUE #937: Define values.schema.yaml for each Helm chart we manage to validate its passed values
### Background

The `helm` CLI can automatically validate passed values (`--values my-prod-config.yaml --values my-common-config.yaml`) against a JSON schema file (`values.schema.json`) whenever one use it to do `helm template`, `helm install`, `helm upgrade`.

I'd like us to make use of that as well, just like we have done in the official JupyterHub Helm chart. It allows us to catch several configuration errors quickly.

In practice, working with a `.json` file is troublesome compared to working with a `.yaml` file. Due to that, I suggest we define a `values.schema.yaml` file and automatically convert it to a `values.schema.json` file that isn't git version controlled but automatically generated by a Python script whenever we need it.

### Tasks

- [ ] Define a values.schema.yaml file for each Helm chart we manage
- [ ] Have a script to generate a values.schema.json file based on this
- [ ] .gitignore relevant values.schema.json
- [ ] Make the deployer script able to do a schema validation given passed values for a hub

### Related
- Z2JH's schema file: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/jupyterhub/schema.yaml
- Z2JH Python script to generate a values.schema.json from a yaml file: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/tools/generate-json-schema.py

### LINKED ISSUE #994: basehub: ensure default values are set for all values used by templates
The basehub helm chart makes use of values as seen in the templates.

https://github.com/2i2c-org/infrastructure/blob/fcd83405e42847b7b322d7d91b94e84a21c0d2b1/helm-charts/basehub/templates/docs-service-deployment.yaml#L1-L4

Not all values, for example these, are not provided with a default value in the basehub helm charts values.yaml file. This has a few downsides.

1. We are forced to write something like this ...
   ```yaml
   {{ if .Values.docs_service }}
   {{ if .Values.docs_service.enabled }}
   ```
   ... instead of just ...
   ```yaml
   {{ if .Values.docs_service.enabled }}
   ```
   ... because we have to check for the existance of a mapping/dictionary named docs_service
2. We are making it hard for anyone looking at values.yaml to understand what is configurable.
   This would be helpful to have for anyone configuring the helm chart, and if we are to create a schema for the helm chart - then we need to define it for the available options - see #937

---

### Action point
- [ ] Go through all basehub's templates and verify that values.yaml has an entry for all referenced configuration.

consideRatio:
## Summary

- Closes #937 by adding values.schema.yaml files, which if converted to values.schema.json files will be recognized by the `helm` CLI that will validate any passed helm chart values.
- Adds helm chart schema validation of our clusters' hubs' configuration by trialing the `helm template` command against the chart in question with the non-encrypted helm chart values files.
- Closes #279 by adding a workflow to trial our deployer scripts `validate` command by using the non-encrypted helm chart values files.
- Closes #994 by manually ensuring and using the new schemas to ensure we don't have Helm chart values our templates rely on, but that we don't provide default values for in the Helm chart's values.yaml file
- Cleans up helm chart configuration that wasn't correct, as identified by the schema files
- Cleaned up logic associated with and content of requirements.txt and dev-requirements.txt
  - `requirements.txt` is now purely dependencies for the deployer script
  - `dev-requirements.txt` is now purely dependencies for scripts in the extra_scripts folder etc
  - Both files have inline comments to clarify the situation
  
## Test failure unrelated

The test failure is unrelated and is resolved by #1044.

### DIFF
--- a/deployer/hub.py
+++ b/deployer/hub.py
@@ -1,26 +1,28 @@
-from auth import KeyProvider
 import hashlib
 import hmac
 import json
 import os
-import sys
 import subprocess
+import sys
 import tempfile
 from contextlib import contextmanager, redirect_stderr, redirect_stdout
-from textwrap import dedent
 from pathlib import Path
+from textwrap import dedent
 
 import pytest
 from ruamel.yaml import YAML
 
+from auth import KeyProvider
 from utils import (
     get_decrypted_file,
-    print_colour,
     get_decrypted_files,
+    prepare_helm_charts_dependencies_and_schemas,
+    print_colour,
 )
 
 # Without `pure=True`, I get an exception about str / byte issues
 yaml = YAML(typ="safe", pure=True)
+helm_charts_dir = Path(__file__).parent.parent.joinpath("helm-charts")
 
 
 class Cluster:
@@ -526,6 +528,8 @@ class Hub:
         """
         Deploy this hub
         """
+        prepare_helm_charts_dependencies_and_schemas()
+
         # Support overriding domain configuration in the loaded cluster.yaml via
         # a cluster.yaml specified enc-<something>.secret.yaml file that only
         # includes the domain configuration of a typical cluster.yaml file.
@@ -549,16 +553,6 @@ class Hub:
             self.spec["domain"] = domain_override_config["domain"]
 
         generated_values = self.get_generated_config(auth_provider, secret_key)
-
-        # Ensure helm charts are up to date
-        helm_charts_dir = (Path(__file__).parent.parent).joinpath("helm-charts")
-        subprocess.check_call(
-            ["helm", "dep", "up", helm_charts_dir.joinpath("basehub")]
-        )
-        if self.spec["helm_chart"] == "daskhub":
-            subprocess.check_call(
-                ["helm", "dep", "up", helm_charts_dir.joinpath("daskhub")]
-            )
 
         with tempfile.NamedTemporaryFile(
             mode="w"
--- a/.github/actions/deploy/action.yml
+++ b/.github/actions/deploy/action.yml
@@ -19,7 +19,6 @@ runs:
     - name: Setup dependencies
       run: |
         python3 -m pip install -r requirements.txt
-        python3 -m pip install -r dev-requirements.txt
       shell: bash
     - name: Deploy support components
       run: |
--- a/.github/workflows/deploy-grafana-dashboards.yaml
+++ b/.github/workflows/deploy-grafana-dashboards.yaml
@@ -22,7 +22,6 @@ jobs:
       - name: Setup dependencies
         run: |
           python3 -m pip install -r requirements.txt
-          python3 -m pip install -r dev-requirements.txt
           sudo apt install jsonnet
 
       - name: Setup gcloud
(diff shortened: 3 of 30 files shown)


deployer/hub.py
```diff
@@ -550,16 +554,6 @@ def deploy(self, auth_provider, secret_key, skip_hub_health_test=False):
...
-        # Ensure helm charts are up to date
-        helm_charts_dir = (Path(__file__).parent.parent).joinpath("helm-charts")
-        subprocess.check_call(
-            ["helm", "dep", "up", helm_charts_dir.joinpath("basehub")]
-        )
-        if self.spec["helm_chart"] == "daskhub":
-            subprocess.check_call(
-                ["helm", "dep", "up", helm_charts_dir.joinpath("daskhub")]
-            )
-
```
sgibson91, line 562: I would strongly prefer us not to remove this block. It was added because the tech team kept running into issues that needed a run of `helm dep up` locally and we kept 'forgetting', leading to lost time trying to debug when this was all we needed to run. There are also no steps in CI currently to run this step, so this removal equates to "never run `helm dep up` ever".

Therefore I would like us to keep this block here for now and open a discussion on where to put it in the future (if this is not the correct place).

refs: https://github.com/2i2c-org/infrastructure/issues/556 and https://github.com/2i2c-org/infrastructure/issues/716

sgibson91: Ignore this comment - I just found where you moved it to!

sgibson91 (approved):

sgibson91:
I am going to merge this now as I'd like to use some of the functionality added in the deployer in https://github.com/2i2c-org/infrastructure/pull/1047 to validate our support chart 🎉

sgibson91:
I updated all the clusters/hubs helm charts locally and all is fine - so any failures in CI leftover are flakey tests

damianavila:
Lovely work, @consideRatio!

PR #1047, opened 2022-03-02
Support values refactor

sgibson91:
- Move support values or each cluster from the `cluster.yaml` file to a `support.values.yaml` file
- Adds `.helmignore`, `values.schema.yaml` to `support` chart
- Updates `cluster.schema.yaml`
- Refactors `deploy-support` and `deploy-grafana-dashbaords` to handle being passed files instead of straight values
- Adds a `validate_support_config` function when `deploy-support` is called

### DIFF
--- a/deployer/__main__.py
+++ b/deployer/__main__.py
@@ -279,9 +279,7 @@ def validate_support_config(cluster_name):
     with open(config_file_path) as f:
         cluster = Cluster(yaml.load(f), config_file_path.parent)
 
-    print_colour(
-        f"Validating non-encrypted support values files for {cluster_name}..."
-    )
+    print_colour(f"Validating non-encrypted support values files for {cluster_name}...")
 
     cmd = [
         "helm",
--- a/deployer/__main__.py
+++ b/deployer/__main__.py
@@ -104,16 +104,20 @@ def deploy_grafana_dashboards(cluster_name):
             f"`grafana_token` not provided in secret file! Please add it and try again: {grafana_token_file}"
         )
 
-    # Get the url where grafana is running from the cluster config
+    # FIXME: We assume grafana_url and uses_tls config will be defined in the first
+    #        file listed under support.helm_chart_values_files.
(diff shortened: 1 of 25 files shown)


consideRatio:
@sgibson91 i made a PR to this PR, see https://github.com/sgibson91/pilot-hubs/pull/135

consideRatio:
I'd also want to see that we make sure that the CI system validates the support chart on PRs making changes.

sgibson91:
> I'd also want to see that we make sure that the CI system validates the support chart on PRs making changes.

@consideRatio the last 2 commits I pushed should address this

consideRatio:
@sgibson91 nice!

Hey btw, you may want to start using pre-commit locally to avoid needing to get pushed commits to your PRs. To do that, `pre-commit install --install-hooks` - and you will never make a commit that doesn't please pre-commit again in this repo :)

consideRatio (approved):
Wieeeeee nice this LGTM!

consideRatio:
@sgibson91 it seems like RTD builds are triggered but failing to start. I suspect their CI service is down atm or experiencing troubles. This can be merged as it is without the RTD CI check in my mind though.

deployer/__main__.py
```diff
@@ -355,6 +391,7 @@ def main():
         )
     elif args.action == "validate":
         validate_cluster_config(args.cluster_name)
+        validate_support_config(args.cluster_name)
```
choldgraf, line 394: Seeing these three `validate_` functions makes me happy :-)

sgibson91:
I will check back in on the CI deployment after I've been to the gym 💪🏻

deployer/__main__.py
sgibson91: Only useful if they work though 😉 https://github.com/2i2c-org/infrastructure/pull/1048#discussion_r817946077

consideRatio: @sgibson91 we just need `validate_validation_config()` as well :D #1049

PR #1048, opened 2022-03-02
Fix our list of running hubs

### LINKED ISSUE #1009: Update/Fix running hubs page in docs
I understand this change as we will now render this part without the `Name` column - at the same time, this reduced the complexity by not entangling configuration significantly.

I support this change, but I think it would be nice to retain this. Perhaps by adding a field to cluster.yaml - which indeed may need to be duplicated as well. Hmm... I propose we move onwards but add a reference to this comment in the todo-list associated with this PR for later consideration.

image

_Originally posted by @consideRatio in https://github.com/2i2c-org/infrastructure/pull/985#discussion_r804882713_

sgibson91:
fixes #1009

### DIFF
--- a/config/clusters/meom-ige/cluster.yaml
+++ b/config/clusters/meom-ige/cluster.yaml
@@ -7,6 +7,7 @@ gcp:
   zone: us-central1-b
 hubs:
   - name: staging
+    display_name: "SWOT Ocean Pangeo Team (staging)"
     domain: staging.meom-ige.2i2c.cloud
     helm_chart: daskhub
     auth0:
@@ -19,6 +20,7 @@ hubs:
       # that you intend for these files to be applied in this order.
       - staging.values.yaml
   - name: prod
+    display_name: "SWOT Ocean Pangeo Team (staging)"
     domain: meom-ige.2i2c.cloud
     helm_chart: daskhub
     auth0:
--- a/shared/deployer/cluster.schema.yaml
+++ b/shared/deployer/cluster.schema.yaml
@@ -175,6 +175,11 @@ properties:
             description: |
               Name of the hub. This will be used to determine
               the kubernetes namespace the hub is deployed into.
+           display_name:
+             type: string
+             description: |
+               A long form name for the hub to display on our website. This is usually
+               the name of the community we run the hub for.
           domain:
             type: string
             description: |
--- a/config/clusters/2i2c/cluster.yaml
+++ b/config/clusters/2i2c/cluster.yaml
@@ -18,6 +18,7 @@ support:
               - grafana.pilot.2i2c.cloud
 hubs:
   - name: staging
+    display_name: "2i2c staging"
     domain: staging.pilot.2i2c.cloud
     helm_chart: basehub
     auth0:
@@ -30,6 +31,7 @@ hubs:
       # that you intend for these files to be applied in this order.
       - staging.values.yaml
   - name: dask-staging
+    display_name: "2i2c dask staging"
     domain: dask-staging.pilot.2i2c.cloud
     helm_chart: daskhub
     auth0:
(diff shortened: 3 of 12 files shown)


config/clusters/meom-ige/cluster.yaml
```diff
@@ -19,6 +20,7 @@ hubs:
       # that you intend for these files to be applied in this order.
       - staging.values.yaml
   - name: prod
+    display_name: "SWOT Ocean Pangeo Team (staging)"
```
sgibson91, line 23:
```suggestion
    display_name: "SWOT Ocean Pangeo Team (prod)"
```

sgibson91:
Look @consideRatio! A pre-commit free PR! 😂 (for now...)

shared/deployer/cluster.schema.yaml
```diff
@@ -175,6 +175,11 @@ properties:
             description: |
               Name of the hub. This will be used to determine
               the kubernetes namespace the hub is deployed into.
+          display_name:
```
consideRatio, line 178: With the schema, you can force a display_name to be set.

I note that this schema is somewhat relaxed with regards to what is set here overall. For example, we of course require a name for the hub, but we don't declare we do.

I noted that you wrote fallback logic for display_name to rely on name - what about instead forcing a display name to be explicitly set, as well as a name, by updating this schema between line 171 and 172 to have a `required` field that sais `name`, `display_name` and other things be required?

consideRatio: Welcome to the JSONSchema human parsing trials!!!

> We already have this?

Actually no, that you link is for the root path of cluster.yaml, not a requirement for fields in the items part of the `hubs` field.

consideRatio: OMG @sgibson91 I THINK WE NEED VALIDATION OF OUR SCHEMA!!!! AAAAAAAAAARGH! :D

I think the required part you linked, won't do anything. You declare it for the `type: array`, but in reality we need it for the items in the array, not the array.

sgibson91: Sometimes the easiest way to test is to unintentionally break. You are right.

Screenshot 2022-03-02 at 19 43 24

Screenshot 2022-03-02 at 19 44 03

Note: I initially tried removing the `name` key and we only raise a Python error because of this line:

https://github.com/2i2c-org/infrastructure/blob/60dc9ff22415a1984cbfd5d6e3b0171303386423/deployer/__main__.py#L242

sgibson91: [bd22ab8](https://github.com/2i2c-org/infrastructure/pull/1048/commits/bd22ab8f2786aee96fa42f5c94f129980a1e269a)

sgibson91:
So close to being a pre-commit-free PR!

shared/deployer/cluster.schema.yaml
sgibson91: Now let's watch all the validation tests fail... 😭

consideRatio:
hahahah oh yes I was just rejoice over a99f9f9 the tireless service pre-commit.ci provides :D

I've confirmed on the RTD preview that this looks good.

I think it could be good to clarify for anyone via the PR title/description that this is about docs rather than core functionality. If you know a PR is about docs, it becomes less critical to review for observed regressions unrelated to docs and of course critical to review for something related to regressions in docs.

consideRatio (approved):

shared/deployer/cluster.schema.yaml
consideRatio: ;) If they fail, they fail very gracefully though - it makes me happy to see such failures.

sgibson91:
Thank you for your review efforts @consideRatio!

PR #1054, opened 2022-03-03
Restructuring the deployer

sgibson91:
### Summary

I recently came across [this blog post](https://breadcrumbscollector.tech/stop-naming-your-python-modules-utils/) demonstrating how having a `utils.py` can be dangerous for a project as it becomes a "dumping ground for code that doesn't obviously belong". Since the recent refactor generated quite a few additions to `utils.py` and other scripts in this module, and I will be adding more as I begin to parallelise the CI/CD, I thought it best to do a clean up before I start.

The main point of this PR is to move code around rather than change the behaviour. I have tried to group similar functions together in a single script that has a descriptive name. So for example, all of our functions related to finding file paths and reading files in now live in a file called `file_acquisition.py`. I have also split up `__main__.py` into `cli.py` (containing all the `argparse` logic) and `deploy_actions.py` (that contains the functions that are called by the cli subcommands). Another thing I did was move the `Cluster` class from `hub.py` into `cluster.py` (not least because `from hub import Cluster` was just wrong 😆 ). I'm hoping this will lead to shorter files that are easier to parse because the content is all related.

I also applied a convention of using a `_` prefix in the names of functions that are not used outside of the file in which they are defined. However, I didn't apply this convention to the class methods as I _believe_ prefixing class methods with `_` has a different meaning than for functions so I left alone.

I removed the `unset_env_var` method from the `Hub` class because it wasn't actually used anywhere!

### Circular import

Upon undertaking this restructure, I unwittingly brought about a circular import which can be described as follows:

1. `__main__.py` imports `cli`
2. `cli.py` imports `cluster`
3. `cluster.py` imports `hub`
4. `hub.py` imports `config_validation`
5. `config_validation.py` imports `cluster`

The problem lying between steps 3 and 5, where running step 3 pauses to process imports, but finds itself being imported back by whatever it was pausing to import - indirectly via `hub` --> `config_validation`.

I broke this circular import by applying commit [7d10d5b](https://github.com/2i2c-org/infrastructure/pull/1054/commits/7d10d5bb6928e6700c58c67b5f6bfed6bf87fe54) which removes a call to `prepare_helm_charts_dependencies_and_schemas` from the `Hub.deploy()` method.

`prepare_helm_charts_dependencies_and_schemas` is what runs `helm dep up` and generates JSON schemas for us to validate against. However the below function, calls `validate_hub_config`, which also calls `prepare_helm_charts_dependencies_and_schemas` before we then call `Hub.deploy()`.

https://github.com/sgibson91/pilot-hubs/blob/5c80e1b61efd37b31eb5a6b0035f75cd2ce95e7a/deployer/deploy_actions.py#L164-L210

So, we were in the position that we would actually run `helm dep up` twice for each deploy: once to validate config and again before the actual deploy. The commit I linked means that we only run `helm dep up` once (via a validation function) and hence break the circular import.

So here is the itch to scratch per se with this PR:

1. we may deploy without running `helm dep up` and generating a schema json file
2. but we likely never run into that issue, because we always call a validation function first which generates said schema file
3. the validation function did a little more than indicated by it's name

How do people feel about this? Too much like splitting hairs? Something we should think more on to maintain confidence in our infrastructure? I don't really know.


deployer/hub.py
```diff
@@ -5,347 +5,22 @@
 import subprocess
 import sys
 import tempfile
-from contextlib import contextmanager, redirect_stderr, redirect_stdout
+import pytest
```
consideRatio, line 8: PEP8 is a style guide describing how to format imports that we don't enforce but like following. It does help me I think - separating the std. lib imports from non-std lib imports and local imports --- that's a great convention i think.

You typically don't have a line separator between import and from statements, but put the from statements after the direct imports.

> Imports should be grouped in the following order:
>
>1. Standard library imports.
>2. Related third party imports.
>3. Local application/library specific imports.
>
>You should put a blank line between each group of imports.

https://www.python.org/dev/peps/pep-0008/#imports

consideRatio (approved):
I reviewed functions locations etc, but assumed there wasn't changes besides relocating and function renaming etc.

This LGTM!

sgibson91:
Oh @consideRatio, I did add that "explicit fail if an encrypted file isn't valid YAML/JSON" with a FIXME comment you requested in the other PR to `get_decrypted_file` in `file_acquistition.py`. I probably wasn't being my most creative for writing an error message at the time though.

damianavila:
Lovely PR, @sgibson91! Thanks.

PR #1057, opened 2022-03-07
Remove unused hackathon-alpha project tfvars

yuvipanda:


choldgraf (approved):
Is this for the original paleohackweek hub? Agree we can remove this one since they're on the shared GKE cluster now

PR #1059, opened 2022-03-08
Use google-github-actions/auth to authenticate against Google Cloud in CI/CD

sgibson91:
fixes #903 

As reported in #903, the functionality of the google-github-actions/gcloud-setup action changed such that authenticating using service account keys was being deprecated and the google-github-actions/auth action should be used instead. This PR adds in the step to authenticate to GCP using the google-github-actions/auth action in our `deploy-hubs.yaml` and `deploy-grafana-dashboards.yaml` workflow files. A few other minor changes come with this PR also:

- Versions of the google-github-actions/auth and google-github-actions/gcloud-setup actions have been pinned to the latest version releases. Updating these versions will be handled by dependabot.
- An extra "decision" step has been added to the `deploy-hubs.yaml` workflow to shorten if statements and improve readability
- Some unnecessary path triggers were removed from `deploy-hubs.yaml` also: config/secrets.yaml no longer exists, and dev-requirements.txt does not contain packages required for deployment

PR #1063, opened 2022-03-08
Use google-github-actions/auth to authenticate against Google Cloud in CI/CD

### LINKED ISSUE #903: Warnings about Service Account Key for gcloud in CI
### Description

CI/CD is producing the following warning when we login to gcloud:

```
"service_account_key" has been deprecated. Please switch to using google-github-actions/auth which supports both Workload Identity Federation and Service Account Key JSON authentication. For more details, see google-github-actions/setup-gcloud#authorization
```

I think this is related to https://cloud.google.com/blog/products/identity-security/enabling-keyless-authentication-from-github-actions

### Value / benefit

- Keeping our CI/CD pipeline up to date will reduce future toil
- Less noise

### Implementation details

_No response_

### Tasks to complete

- [ ] https://github.com/2i2c-org/infrastructure/pull/1059

### Updates

_No response_

sgibson91:
> This is a reimplementation of #1059 

fixes #903 

As reported in #903, the functionality of the google-github-actions/gcloud-setup action changed such that authenticating using service account keys was being deprecated and the google-github-actions/auth action should be used instead. This PR adds in the step to authenticate to GCP using the google-github-actions/auth action in our `deploy-hubs.yaml` and `deploy-grafana-dashboards.yaml` workflow files. A few other minor changes come with this PR also:

- Versions of the google-github-actions/auth and google-github-actions/gcloud-setup actions have been pinned to the latest version releases. Updating these versions will be handled by dependabot.
- An extra "decision" step has been added to the `deploy-hubs.yaml` workflow to shorten if statements and improve readability
- Some unnecessary path triggers were removed from `deploy-hubs.yaml` also: config/secrets.yaml no longer exists, and dev-requirements.txt does not contain packages required for deployment

### DIFF
--- a/.github/workflows/deploy-grafana-dashboards.yaml
+++ b/.github/workflows/deploy-grafana-dashboards.yaml
@@ -25,7 +25,7 @@ jobs:
           sudo apt install jsonnet
 
       - name: Setup gcloud
-        uses: google-github-actions/setup-gcloud@master
+        uses: google-github-actions/setup-gcloud@v0.5.1
         with:
           version: "290.0.1"
           # This is used for KMS only
--- a/.github/workflows/deploy-grafana-dashboards.yaml
+++ b/.github/workflows/deploy-grafana-dashboards.yaml
@@ -24,14 +24,16 @@ jobs:
           python3 -m pip install -r requirements.txt
           sudo apt install jsonnet
 
+      - name: Authenticate to Google Cloud
+        uses: google-github-actions/auth@v0.6.0
+        with:
+          project_id: two-eye-two-see
+          credentials_json: ${{ secrets.GCP_KMS_DECRYPTOR_KEY }}
+
       - name: Setup gcloud
         uses: google-github-actions/setup-gcloud@v0.5.1
         with:
           version: "290.0.1"
-          # This is used for KMS only
-          project_id: two-eye-two-see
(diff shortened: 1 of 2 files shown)


consideRatio (approved):
Still looks good to me - hope this works!

If it does work, I'm not confident it was because we set the `project_id` explicitly, but if it was because we removed the `export_default_credentials` from the `google-github-actions/setup-gcloud` action.

I think perhaps `project_id` isn't required, it depends I guess. See [this documentation](https://github.com/google-github-actions/auth#other-inputs):

> `project_id`: (Optional) Custom project ID to use for authentication and exporting into other steps. If unspecified, the project ID will be extracted from the Workload Identity Provider or the Service Account Key JSON.

sgibson91:
I wonder if we need to replace `export_default_credentials` with `create_credentials_file`? Documented here: https://github.com/google-github-actions/auth#other-inputs

consideRatio:
> I wonder if we need to replace `export_default_credentials` (in setup-gcloud) with `create_credentials_file` (in auth)? Documented here: https://github.com/google-github-actions/auth#other-inputs

It defaults to true. It is a flag to help you opt-out rather than opt-in I'm thinking.

> `create_credentials_file`: (Optional) If true, the action will securely generate a credentials file which can be used for authentication via gcloud and Google Cloud SDKs in other steps in the workflow. __The default is true.__

PR #1069, opened 2022-03-09
Additional safeguards against accidental cluster deletion

yuvipanda:
I ran `terraform workspace new` to create a new workspace,
and somehow assumed it also switched me to the new workspace -
which it *does not*. So just running `terraform apply` almost
deleted a bunch of running clusters.

This lifecycle hook will *prevent* any of these from being
deleted ever. To delete these resources (as part of a decomission),
you'll need to edit the files to set it to false. This is a manual
step, but very appropriate for irreversible data loss.

### DIFF
--- a/terraform/aws/efs.tf
+++ b/terraform/aws/efs.tf
@@ -45,6 +45,12 @@ resource "aws_efs_file_system" "homedirs" {
   tags = {
     Name = "hub-homedirs"
   }
+
+  lifecycle {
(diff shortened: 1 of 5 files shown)


sgibson91:
We probably need a "delete a cluster" version of [these docs](https://infrastructure.2i2c.org/en/latest/howto/operate/delete-hub.html) that documents this manual step

sgibson91 (approved):

PR #1072, opened 2022-03-09
Bump utoronto image tag

sgibson91:
This PR bumps the U of T image tag to the most recent build to include some packages they requested via support.

@yuvipanda can you check the configurator on prod to make sure it won't override this tag please?

PR #1073, opened 2022-03-09
Bump tag of utoronto image

yuvipanda:
Brings in https://github.com/2i2c-org/utoronto-image/pull/19
and https://github.com/2i2c-org/utoronto-image/pull/17

### DIFF
--- a/config/clusters/utoronto/staging.values.yaml
+++ b/config/clusters/utoronto/staging.values.yaml
@@ -76,7 +76,7 @@ jupyterhub:
             subPath: "{username}"
     image:
       name: quay.io/2i2c/utoronto-image
-      tag: f3069be5964c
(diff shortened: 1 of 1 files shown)


yuvipanda:
Dupe of #1072

config/clusters/utoronto/staging.values.yaml
```diff
@@ -76,7 +76,7 @@ jupyterhub:
             subPath: "{username}"
     image:
       name: quay.io/2i2c/utoronto-image
-      tag: f3069be5964c
+      tag: 18b55a7bcb96
```
consideRatio, line 79: Warning that without quotes, this could cause problem every 0.35% of the times when the hex string is entirely numerical.

PR #1078, opened 2022-03-10
Allow logins with rmbl.org accounts

yuvipanda:
Users should select 'Microsoft' as the way to login,
and try this out.


choldgraf (approved):
nice, LGTM!

Issue #1055, opened 2022-03-03
Upgrade our hubs to Z2JH 2 / JupyterHub 3.0


choldgraf:
### Background and proposal

Our `basehub` chart defines the version of Zero to JupyterHub that we pull in here:

https://github.com/2i2c-org/infrastructure/blob/6da898b2892c9e81b6afa91d1ade3d2a0e56a472/helm-charts/basehub/Chart.yaml#L1-L15

However, Z2JH 2.0 will be released relatively soon, and is going to bring with it **JupyterHub 2.x**.

We should upgrade to Z2JH 2.0 as long as we're confident that it will be fit for production in our hubs. However, there may be some coordination that we need to take with either our `basehub` etc helm charts, or with the user images that we're serving.

This is an issue to track that effort so that we know what needs to be done!

### Implementation guide and constraints

[Here's the issue tracking the release of Z2JH 2.0](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2386). I believe that @consideRatio has been putting a lot of time into this one (thanks 🙏🙏🙏) and maybe he can advise on progress as we near a release. Looking at that issue, it seems quite close.

[Here's an example of a problem that occurred because somebody bumped `jupyterhub-singleuser` too early, on a hub that did not yet have JupyterHub 2.0](https://github.com/pangeo-data/pangeo-docker-images/issues/272#issuecomment-1058390957).

 

### Updates and ongoing work

- [x] Wait for Z2JH 2.0 to be released (and/or help make it happen!) (ref: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2386)
- [ ] Compile a list of things we should double-check in our hub configuration to make sure an upgrade won't break things.
- [ ] Make the necessary upgrades in our config + any community hub configs.
- [ ] Hope that everything doesn't explode.
- [ ] Check if https://github.com/2i2c-org/infrastructure/issues/1643 is actionable now

consideRatio:
> We should upgrade to Z2JH 2.0 as long as we're confident that it will be fit for production in our hubs. However, there may be some coordination that we need to take with either our basehub etc helm charts, or with the user images that we're serving.

To ensure user images have JupyterHub 2 installed ahead of time is a very big deal of the 2i2c prep i think.

consideRatio:
[...]

PR #1079, opened 2022-03-10
Upgrade to z2jh 1.2.0

yuvipanda:
This is the latest stable version. We want to eventually
move to 2.0 (with some breaking changes) but let's get to
the latest non 2.0 stable version first

Ref https://github.com/2i2c-org/infrastructure/issues/1055

### DIFF
--- a/helm-charts/basehub/Chart.yaml
+++ b/helm-charts/basehub/Chart.yaml
@@ -11,5 +11,5 @@ dependencies:
     # images/hub/Dockerfile, and will also involve manually building and pushing
     # the Dockerfile to https://quay.io/2i2c/pilot-hub. Details about this can
     # be found in the Dockerfile's comments.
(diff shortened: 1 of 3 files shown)


consideRatio (approved):

GeorgianaElena (approved):

choldgraf (approved):

yuvipanda:
Looked at the very nicely maintained changelog here: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/CHANGELOG.md#120---2021-11-04. Seems good :) I'm testing this on staging, and then deploying after.

yuvipanda:
When upgrading z2jh, it is important to check if there is anything we need to check in the images used for it to work with the new version of z2jh. 1.1.x to 1.2.x doesn't have any such changes though.

PR #1085, opened 2022-03-11
Pin sops to 3.7.1 in GitHub actions

yuvipanda:
sops 3.7.2, released yesterday, had a corrupted linux
binary https://github.com/mozilla/sops/issues/1023.
This causes all our actions needing sops to fail.
We pin to 3.7.1 until sops upstream fixes it
https://github.com/mdgreenwald/mozilla-sops-action/issues/105#issuecomment-1064018419

PR #1086, opened 2022-03-11
Revert "Upgrade to z2jh 1.2.0"

yuvipanda:
Reverts 2i2c-org/infrastructure#1079

Reverting since sops's latest release is broken. Fixed by
https://github.com/2i2c-org/infrastructure/pull/1085, but we'll need to revert the original 
upgrade PR, then revert the revert to actually run the deploy - as our CD only deploys
when files are touched rather than at all times in a reconcilation loop.


yuvipanda:
Going to self-merge, as this is a revert of an approved PR (https://github.com/2i2c-org/infrastructure/pull/1079). Our CD is broken so this won't actually deploy anything.

PR #1089, opened 2022-03-11
Full Cilogon integration

GeorgianaElena:
This PR adds another way to use CILogon auth for our hubs, one that doesn't require relying on Auth0.

It adds a script, `cilogon_app.py` that: 
* uses the administrative client credentials that we received from CILogon to  `create/udpate/delete/get` approved OAuth applications for our hubs.
* stores and ecrypts the credentials of each hub in `config/clusters/<cluster_name>/enc-<hub_name>.secret.values.yaml

The process to activate CILogon for a hub using the credentials generated above is very similar to the current process of using GitHubOAuthenticator.

Also, this PR enables CILogon for our `2i2c dask-staging` hub and `2i2c demo` hub. It also adds credentials for the 2i2c staging hub also, but without activating it for this hub.

The  [`dask-staging`](https://dask-staging.pilot.2i2c.cloud) hub uses this, so you can play with it. It works from what I've tested so far.

Reference https://github.com/2i2c-org/infrastructure/issues/967


config/clusters/2i2c/enc-cilogon-clients.secret.yaml
```diff
@@ -0,0 +1,16 @@
+2i2c-dask-staging:
```
consideRatio, line 1: This seems confusing to me. Should this be named like that?

config/clusters/2i2c/cluster.yaml
```diff
@@ -30,7 +30,11 @@ hubs:
     auth0:
       # connection update? Also ensure the basehub Helm chart is provided a
       # matching value for jupyterhub.custom.2i2c.add_staff_user_ids_of_type!
-      connection: google-oauth2
+      enabled: false
+      # connection: google-oauth2
+    cilogon:
```
choldgraf, line 35: I like that we are allowing them to co-exist for a bit! As long as we don't have too much headaches maintaining both, this should give us time to try out CILogon without disrupting our current workflows as much

deployer/cilogon_auth.py
```diff
@@ -0,0 +1,193 @@
+import base64
```
choldgraf, line 1: since we now have several auth-related python modules, maybe we should nest these all under `deployer/auth/`?

choldgraf (commented):
This looks nice to me! A few comments and questions. I'll let others on the @2i2c-org/tech-team give a more technical review!

- Could we add the start of some lightweight docs for how we'd use CILogon auth? I think it could be helpful exercise to think about the changes from a user's perspective. It doesn't need to be totally complete since I know you'd like to get the PR in quickly, but something to help people start would be useful.
- Given what you've seen, are there any major decision-points that you can think of where we'd want to use Auth0 vs. CILogon?
- Have you encountered any major downsides or concerns about this approach? Or have you encountered any major benefits?

config/clusters/2i2c/enc-cilogon-clients.secret.yaml
GeorgianaElena: This name comes from https://github.com/2i2c-org/infrastructure/blob/6bf2e153256fd15b2793ea8229e1da2b75efe7cc/deployer/hub.py#L135.

We had this convention before too https://github.com/2i2c-org/infrastructure/blob/dd1517d2414f6f49ba22d325074bf488908dbb0e/deployer/hub.py#L131-L133

This is how the 2i2c Auth0 applications are called https://manage.auth0.com/dashboard/us/2i2c/applications.

This client name gets written to this encrypted file [here](https://github.com/2i2c-org/infrastructure/blob/6bf2e153256fd15b2793ea8229e1da2b75efe7cc/deployer/cilogon_auth.py#L160-L170). The reason we are storing these for cilogon and we didn't for Auth0 is because cilogon retrieves details about a client only if provided with the client's id and not its name (`2i2c-dask-staging` is such a name). Also, it doesn't retrieve you the client's secret afterwards, only when it gets created.

deployer/cilogon_auth.py
sgibson91: Yes, I would be +1 on this too. _But_ happy for that to be done in a separate PR so as to not slow this one down, so long as we track it.

GeorgianaElena: I've added two lists in the top level comments to track what I believe should be happening as part of this PR and what should be their own separate issues after we merge this. WDYT?

choldgraf: Looks good to me! For the docs, I think the minimal set would be the steps a team member would need to take to:

- Set up CILogon auth with a hub (that used this approach instead of `auth0`)
- Any common next steps in setting up auth (e.g. white-list one or more organizations, or specific users...basically turning our big "regexes" into more structured config)

yuvipanda (changes requested):
Thank you so much for working on this, @GeorgianaElena!!!!

The primary reason the auth0 stuff was baked into the deployer was so we won't have to keep secret files in the repo. However, given that is not possible with CILogon (we can not retrieve secret key multiple times), I think we should instead change this to something where the script just spits out appropriate sops encrypted files as `enc-<hub-name>.secret.values.yaml` file (just using the CILogon authenticator), which we can just check-in. This also simplifies the deployer (once we can get rid of auth0 at least), and enhances right to replicate.

So we could have something like a new-hub script or cookie-cutter that talks to the CILogon API, generates the appropriate secrets and puts it in place. I think the code for this already exists here.

Once we move off auth0 completely, we can just get rid of the dynamic auth part of the deployer completely.

GeorgianaElena:
Thanks for the feedback @yuvipanda. I just pushed some changes with the cilogon client provider turned into a script.
It's not ready, but I would appreciate an early feedback so that I don't get into another rabbit hole.

Also, weird thing, I just got a 500 error with a message saying that the cilogon credentials I was trying to use don't have an "admin" role. @choldgraf do you mind asking the CILogon folks if they've revoked our admin credentials? I'm afraid they think we've abused their system since I've created quite a few dummy clients while testing the implementation 😕

choldgraf:
@GeorgianaElena will send them an email today!

GeorgianaElena:
Thanks a lot @choldgraf :blossom:

yuvipanda:
@GeorgianaElena i haven't done a deep dive into the code but overall this seems the right direction to me!

GeorgianaElena:
I have marked this as ready for review again!

I still want to add some things like:
- [ ] ability to delete a client/secret pair (i.e. OAuth application)
- [ ] docs about the `cilogon_app.py` script
- [ ] update docs on how to update/delete a hub that's using CILogon using this script

But I don't think it should block a first review iteration.

Also, note that the pre-commit is also failing on the main branch, so it shouldn't be because of the changes here.

docs/howto/configure/auth-management.md
```diff
@@ -227,7 +221,96 @@ User account
...
+   ```
+
+5. **Edit the non-secret config under `config/clusters/<cluster_name>/<hub_name>.values.yaml`.**
+   You should make sure the matching hub config takes one of the following forms.
+
+   ```{warning}
+   When using this method of authentication, make sure to remove the `allowed_users` key from the config.
+   This is because this key will block any user not listed under it **even if** they are valid members of the the organisation or team you are authenticating against.
+
+   Also, the `admin_users` list need to match `allowed_idps` currently. TODO: we should fix this upstream.
```
yuvipanda, line 280: Is there an issue for this upstream? If not, let's file one and link to it here?

yuvipanda (changes requested):
This is pretty awesome! I left some minor style and follow-up comments, but I think this is ready to merge as is once those are fixed!

config/clusters/2i2c/enc-cilogon-clients.secret.yaml
consideRatio: Ah thanks for the thorough comment and references @GeorgianaElena! Sorry for the slow followup on this! I see now that the structure has changes, so I figure this may be a resolved comment.

---

I think a key confusion point for me was regarding that this was a .YAML file that we load, but it wasn't a helm chart values file and didn't seem to be a cluster.yaml like file either - so then it seemed like there wasn't a schema to validate this configuration.

I think that for any configuration it is probably good if we avoid having keys that contain data by themselves though, so instead of having a name as a key like...

```yaml
this-is-some-cluster-hub-name: some value
```

We could have something like....

```yaml
clusterHub:
  name: this-is-some-cluster-hub-name
  someProperty: some-value
```

This latter would be easy to have a schema for, and its data would be described in the YAML structure etc.

consideRatio: I see https://github.com/2i2c-org/infrastructure/pull/1089#pullrequestreview-912171302 which is related, and yes - this seems to be resolved!

docs/howto/configure/auth-management.md
```diff
@@ -227,7 +221,96 @@ User account
...
+      python3 ./deployer/cilogon_auth.py create 2i2c dask-staging daskhub https://dask-staging.2i2c.cloud/hub/oauth_callback
+      ```
+   - If successfull, the script will have created a secret values file under `config/clusters/<cluster_name>/enc-<hub_name>.secret.values.yaml`. This file 
+   holds the encrypted OAuth client id and secret that have been created for this hub.
+   - The unecrypted file contents should look like this:
+      ```yaml
+        jupyterhub:
+          hub:
+            config:
+              GitHubOAuthenticator:
```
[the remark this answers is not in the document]
GeorgianaElena: Fixed in https://github.com/2i2c-org/infrastructure/pull/1089/commits/6819a9e6a13fb9e3bbdf7fdda21809ad86cf9c68

docs/howto/configure/auth-management.md
```diff
@@ -227,7 +221,96 @@ User account
...
+    jupyterhub:
+      hub:
+        config:
+          JupyterHub:
+            authenticator_class: cilogon
+          CILogonOAuthenticator:
+            oauth_callback_url: https://{{ HUB_DOMAIN }}/hub/oauth_callback
+            username_claim: USERNAME_KEY
+            scope:
+              - openid
```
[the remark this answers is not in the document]
GeorgianaElena: Done in https://github.com/2i2c-org/infrastructure/pull/1089/commits/249e8a62aed3eadd5e97f569f750fae3848a26f9

docs/howto/configure/auth-management.md
GeorgianaElena: Done in https://github.com/2i2c-org/infrastructure/pull/1089/commits/d607b333a09c0fd61fd5bf535fad1dc9fcbdf303

deployer/cilogon_app.py
```diff
@@ -0,0 +1,387 @@
...
+        return requests.post(url, json=data, headers=headers, timeout=self.timeout)
+
+    def _get(self, url, params=None):
+        headers = self.base_headers.copy()
+
+        timeout = time.time() + 10
+        t = 0.05
+
+        # Allow up to 10s of retrying getting the response
+        while time.time() < timeout:
```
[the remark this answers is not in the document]
GeorgianaElena: Thanks @yuvipanda! I agree with the points above and fixed them in https://github.com/2i2c-org/infrastructure/pull/1089/commits/2c49f3cd4d81114b9b1ce47f2b6096f0c43633e3

GeorgianaElena:
This should be now ready for a new review 👀 or a merge

docs/howto/configure/auth-management.md
```diff
@@ -227,7 +221,97 @@ User account
...
+              - email
+              - org.cilogon.userinfo
+            allowed_idps:
+              - 2i2c.org
+              - IDP
+    ```
+  - Check the [CILogon scopes section](https://www.cilogon.org/oidc#h.p_PEQXL8QUjsQm) to checkout available values for `USERNAME_KEY` claim.
```
yuvipanda, line 303:
```suggestion
  - Check the [CILogon scopes section](https://www.cilogon.org/oidc#h.p_PEQXL8QUjsQm) to checkout available values for `USERNAME_KEY` claim. 
     Be careful selecting a value, as once picked this can not be easily changed! It should ideally be something
     that the end user *can not change*, as otherwise they can change it after logging in to access other
     users' servers! 
```

yuvipanda:
@GeorgianaElena i've left some suggested changes in https://github.com/GeorgianaElena/pilot-hubs/pull/2! I'm going to now test deploying this on the demo hub.

yuvipanda:
https://github.com/GeorgianaElena/pilot-hubs/pull/3 is me trying to deploy this to the demo hub, but ran into problems :( I've detailed them in the PR. I think if we can resolve this, this is good to go.

This is such a huge step forward, *thank you* for working on this @GeorgianaElena!

yuvipanda:
The test failure is a link to a file introduced in this PR, so is alright to let fail.

docs/howto/configure/auth-management.md
```diff
@@ -292,16 +292,24 @@ The steps to enable the JupyterHub CILogonOAuthenticator for a hub are simmilar
...
-            scope:
-              - openid
-              - email
-              - org.cilogon.userinfo
             allowed_idps:
               - 2i2c.org
               - IDP
     ```
-  - Check the [CILogon scopes section](https://www.cilogon.org/oidc#h.p_PEQXL8QUjsQm) to checkout available values for `USERNAME_KEY` claim.
-  - Per [CILogon's suggestion]((https://www.cilogon.org/oidc#h.p_PEQXL8QUjsQm)), please use the same list of scopes in this example when enabling CILogon for a hub.
```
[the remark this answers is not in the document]
GeorgianaElena: @consideRatio, they were added to the config directly. See https://github.com/GeorgianaElena/pilot-hubs/pull/2#discussion_r834383513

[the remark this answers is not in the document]
consideRatio: Oh nevermind, i see that there are default scopes provided in the basehub helm chart - okay!

Overall I value duplicating configuration if it helps me understand what is actually configured in a single place. For example, I may look at this configuration and be surprised that it works - should it have some scopes specified? To understand it properly, I'd need to know if scopes are appended by `hub.extraConfig` or via `hub.config` as specified in the basehub helm chart default values etc.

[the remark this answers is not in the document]
consideRatio: @GeorgianaElena @yuvipanda what do you think about where we put config in a situation like this? Do you value needing to provide less config and avoiding duplicating configuration across files more than you value the complexity of dispersing the config in different places?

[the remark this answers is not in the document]
yuvipanda: @consideRatio Thank you for raising this question! I definitely value needing to provide less config and avoiding duplicating configuration across files more than you value the complexity of dispersing the config in different places? I think multiple helm chart values files that cascade is a very important part of keeping config concise. I'll write up a doc on 'how to find what config', which might help - as for example, you might sometimes need to look at defaults in z2jh or even in kubespawner if necessary!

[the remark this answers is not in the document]
yuvipanda: I opened https://github.com/2i2c-org/infrastructure/issues/1144, @consideRatio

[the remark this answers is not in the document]
consideRatio: Thank you Yuvi, I'll mark this as resolved to not bog down this PR with long term deliberations!

yuvipanda (approved):
@GeorgianaElena this is super awesome, am very excited to get this to land!

1. https://github.com/GeorgianaElena/pilot-hubs/pull/4 keeps status quo for demo hub, and we can figure out what to do with IDPs
2. https://github.com/GeorgianaElena/pilot-hubs/pull/5 removes some extra config
3. Can you update the PR description here too?

I think you can merge after these two are done.

docs/howto/configure/auth-management.md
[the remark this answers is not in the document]
yuvipanda: Turns out you don't actually need to set scopes at all, as these are the same that's [the default](https://github.com/jupyterhub/oauthenticator/blob/8c1cf0c616d5e19d1f93238673f17db2daf57fa0/oauthenticator/cilogon.py#L69) set by CILogonOAuthenticator! https://github.com/GeorgianaElena/pilot-hubs/pull/5 removes these.

GeorgianaElena:
Merging this now! THANK YOU! 🚀

damianavila:
Lovely piece of work, @GeorgianaElena!!

Issue #328, opened 2021-03-27
Cloud usage monitoring and alerting infrastructure and process


yuvipanda:
### Description of problem and opportunity to address it

**Problem description**
In #908 we ran into a case where a user was abusing the JupyterHub for crypto mining. This resulted in a lot of stress and high costs for the hub's community. Part of the problem was that we did not detect the mining activity for several weeks. This activity was basically:

- The steady creation of new users on the hub
- Each user maxing out their CPU and never shutting down their session

**Proposed solution**
We should create a mechanism for automatically monitoring statistics around hub usage, and triggering notifications that suggest something nefarious is happening. Ideally, this would be a single process for all of our clusters, not one process for each cluster.

We need a quick way to:

1. Keep an eye on all these projects *in one place*
2. Have automated alerts for abnormal costs
3. Do rounds of cost optimizations

**What's the value and who would benefit**
This would allow us to minimize the risk of abuse if somebody *did* try to use a hub for the wrong purposes. It would give our team more confidence that something isn't happening without us knowing about it, and would give communities more confidence that they won't have an unexpected spike in their cloud bill.

### Implementation guide and constraints

A rough idea of what to try:

- Set up a Grafana dashboard that aggregates activity across all of our clusters (this will be tricky because the Prometheus instances are *private* for our clusters, not public like the Binder ones).
- Define a few metrics that are particularly useful for identifying abuse and problematic abnormal behavior. For example, here are two images from the openscapes grafana that were particularly useful:

  - **Users over time**
  <img src="https://user-images.githubusercontent.com/1839645/148128329-1d8bd83c-3a2e-4b83-91ae-91ff33769bd5.png" width=400 />

  - **CPU usage histogram over time**
  <img src="https://user-images.githubusercontent.com/1839645/148128328-58d190e3-99ce-4590-984d-258e2c46cd79.png" width=400 />
  
  - And noting `5xx` errors from user pods in general is a good indication that something is wrong.
- Define some thresholds for these metrics, and create a reporting mechanism to ping `support@2i2c.org` when it thinks something problematic is going on.

### Issues where we have been bitten by this

- https://github.com/2i2c-org/infrastructure/issues/1081
- https://github.com/2i2c-org/infrastructure/issues/908

#### 2022-01-06

@GeorgianaElena is going to work on these things for **one week**:

- [x] How complex will it be to aggregate feeds from each cluster's Prometheus?
[...]

PR #1091, opened 2022-03-12
Expose prometheus with basic auth

yuvipanda:
This is the beginning of implementing idea 1 from the
list @georiganaelena made in
https://github.com/2i2c-org/infrastructure/issues/328#issuecomment-1016601966.

We have one prometheus running per cluster, but manage many clusters.
A single grafana that can connect to all such prometheus clusters
will help with monitoring as well as reporting. So we need to expose
it as securely as possible to the external world, as it can contain
private information.

In this case, we're using https + basic auth provided by
nginx-ingress
(https://kubernetes.github.io/ingress-nginx/examples/auth/basic/)
to safely expose prometheus to the outside world. We can then
use a grafana that knows these username / passwords to access this
prometheus instance. Each cluster needs its own username / password
(generated with pwgen 64 1), so users in one cluster can not access
prometheus for another cluster.

Ref https://github.com/2i2c-org/infrastructure/issues/328


yuvipanda:
I used a tiny helper script to generate these encrypted secret files:

```
#!/bin/bash
set -euo pipefail

F="${1}/enc-support.secret.yaml"
echo "prometheusAuth:" > $F
echo "  username: $(pwgen 64 1)" >> $F
echo "  password: $(pwgen 64 1)" >> $F
sops -i -e $F
echo "Done $F"%   
```

Then ran `find . -type d | xargs -L1 ./gen.bash  ` to run it for all dirs inside config/clusters.

However, meom-ige and farallon still don't have a support cluster, so this can't be used there yet.

yuvipanda:
Note - I've already run `deploy-support` for all these :)

consideRatio:
Hmmm, the support chart could also do the trick jupyterhub does and generate secrets in a Secret file - if prometheus could accept accessing the content from there.

It may not be qhat makes sense, just floating the idea to ensure it is considered as an option. Someplace, the grafana server would need to have access to those secrets in other clusters so it would probably be relevant to have centrally still.

yuvipanda:
@consideRatio yeah, that's the other option - but that would require giving kubernetes API access to every single cluster to the centralized grafana, which I would very much prefer to not do.

deployer/cluster.py
```diff
@@ -8,7 +8,7 @@
 
 from hub import Hub
 from utils import print_colour
-from file_acquisition import get_decrypted_file
+from file_acquisition import get_decrypted_file, get_decrypted_files
```
consideRatio, line 11:
```suggestion
from file_acquisition import get_decrypted_files
```

deployer/cluster.py
```diff
@@ -73,53 +73,54 @@ def deploy_support(self):
...
-        subprocess.check_call(
-            [
-                "helm",
-                "repo",
-                "add",
-                "jetstack",
-                cert_manager_url,
-            ]
-        )
+        # print_colour("Adding cert-manager chart repo...")
```
consideRatio, line 76: Reminder to add this back

deployer/cluster.py
```diff
@@ -224,7 +224,7 @@ def auth_aws(self):
                     os.environ["KUBECONFIG"] = orig_kubeconfig
                 if orig_access_key_id is not None:
                     os.environ["AWS_ACCESS_KEY_ID"] = orig_access_key_id
-                if orig_kubeconfig is not None:
+                if orig_secret_access_key is not None:
```
consideRatio, line 227: A small bugfix, nice!

helm-charts/support/values.schema.yaml
```diff
@@ -69,6 +70,22 @@ properties:
         properties:
           enabled:
             type: boolean
+  prometheusAuth:
```
consideRatio, line 73:
```suggestion
  # prometheusAuth is _not a dependent helm chart_. It is values directly
  # coupled with this helm chart and are influencing the rendering of templates
  # we provide as part of this helm chart.
  #
  # Due to that, we maintain a strict schema here but not elsewhere.
  #
  prometheusAuth:
```

helm-charts/support/values.yaml
```diff
@@ -24,6 +29,13 @@ prometheus:
   pushgateway:
     enabled: false
   server:
+    ingress:
+      annotations:
+        nginx.ingress.kubernetes.io/auth-type: basic
+        nginx.ingress.kubernetes.io/auth-secret: prometheus-ingress-auth
+        nginx.ingress.kubernetes.io/auth-realm: "Authentication Required"
+        kubernetes.io/ingress.class: nginx
+        cert-manager.io/cluster-issuer: letsencrypt-prod
```
consideRatio, line 38: This setup enables a nice security feature I think.

By using a netpol for prometheus internally that requires traffic to arrive from the ingress controller pods which in turn enforce passwords, we would have locked down access to prometheus internally and externally.

helm-charts/support/values.yaml
```diff
@@ -24,6 +29,13 @@ prometheus:
   pushgateway:
     enabled: false
   server:
+    ingress:
+      annotations:
+        nginx.ingress.kubernetes.io/auth-type: basic
+        nginx.ingress.kubernetes.io/auth-secret: prometheus-ingress-auth
```
consideRatio, line 35: TODO: This secret needs to be created and consume the values passed.

helm-charts/support/values.yaml
```diff
@@ -1,3 +1,8 @@
+prometheusAuth:
```
consideRatio, line 1: The naming `prometheusAuth` is helpful, but I don't think it is helpful enough. We who use this configuration this often need to know more technical details and understand what this influence. I think a more detailed description is favorable, hiding less of the technical details and by that also giving more insights into what is done by configuring this.

I got confused about `prometheusAuth` as if it was how you configured prometheus to require authentication, but in reality the strategy is coupled with Prometheus creating an Ingress resource to be managed by nginx-ingress which in turn is hardcoded to reference a k8s Secret named `prometheus-ingress-auth` for username/password credentials required for requests.

I think this name should reflect that this is related to a) prometheus authentication, b) it is a k8s Secret being created, c) it is referenced by an Ingress resource, and d) the ingress resource is managed by nginx-ingress.

`prometheusAuthIngressSecret` perhaps?

consideRatio (commented):
This is a nice beginning!

### My understanding summarized

1. We assume we have `prometheus` subdomains for each cluster point to an `nginx-ingress` controllers in each cluster.
2. Letting the support chart create a k8s Secret to hold username/password for prometheus access via support chart values.
3. Letting the support chart's dependent prometheus Helm chart create an k8s Ingress resource that references the created k8s Secret with username/password. This reference is a annotation that an nginx-ingress controller understands: we wish to require basic authentication using the provided username/password in the k8s Secret
4. Later we manually (?) configure a central Grafana instance through its web UI to have many data sources, one for each cluster. Following this, we can allow dashboards there to choose to inspect any given cluster or have some dashboards that summarize the state of all clusters.

### A few action points

- [x] Add a k8s Secret template in the support chat
- [ ] Do work to ensure we get the relevant new sub-domain names get CNAME entries in the DNS servers we control, I figure they should point to the primary domain name we use which also points to the ingress controllers IP.
- [ ] Point out a specific Grafana instance to be the central instance we use (the one in the 2i2c cluster right?)

config/clusters/azure.carbonplan/support.values.yaml
```diff
@@ -1,3 +1,17 @@
+prometheusAuth:
+  enabled: true
+
+prometheus:
```
consideRatio, line 4: There is a CI error about this. There are two separate entries for `prometheus` in this file, where one below configures the resource requests.

deployer/cluster.py
yuvipanda: Done!

helm-charts/support/values.schema.yaml
yuvipanda: Done!

helm-charts/support/values.yaml
yuvipanda: @consideRatio oh yeah, I just forgot to git add that file! done :)

helm-charts/support/values.yaml
yuvipanda: @consideRatio flawless logic and very well written feedback! I've renamed it to be prometheusIngressAuthSecret (since it's the prometheusIngress we are providing auth to, not prometheus itself).

config/clusters/azure.carbonplan/support.values.yaml
yuvipanda: Thanks for catching it! Fixed! All hail CI!

deployer/cluster.py
yuvipanda: ty :)

helm-charts/support/templates/prometheus-htpasswd.yaml
```diff
@@ -0,0 +1,10 @@
+
+{{- if .Values.prometheusIngressAuthSecret.enabled }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: prometheus-ingress-auth-basic
+type: Opaque
+data:
+  auth: {{ htpasswd .Values.prometheusIngressAuthSecret.username .Values.prometheusIngressAuthSecret.password | b64enc | quote}}
+{{- end }}
```
consideRatio, line 10: - Remove blank new line in the beginning of file
- Add new line in end of file
- Add right whitespace chomp on the helm template expression before the first content has been rendered as a rule of thumb to ensure we don't render a blank line as the first thing as our output of the rendered helm template.

```suggestion
{{- if .Values.prometheusIngressAuthSecret.enabled -}}
apiVersion: v1
kind: Secret
metadata:
  name: prometheus-ingress-auth-basic
type: Opaque
data:
  auth: {{ htpasswd .Values.prometheusIngressAuthSecret.username .Values.prometheusIngressAuthSecret.password | b64enc | quote}}
{{- end }}

```

helm-charts/support/values.yaml
```diff
@@ -24,6 +29,19 @@ prometheus:
   pushgateway:
     enabled: false
   server:
+    ingress:
+      annotations:
+        # Annotations required to enable basic authentication for any ingress
+        # into prometheus server from the outside world - this secret is
+        # created via the prometheus-htpasswd.yaml file, controlled by config
+        # under prometheusIngressBasicAuth. Ingress is not enabled by default,
```
consideRatio, line 37:
```suggestion
        # under prometheusIngressAuthSecret. Ingress is not enabled by default,
```

helm-charts/support/values.yaml
```diff
@@ -24,6 +29,19 @@ prometheus:
   pushgateway:
     enabled: false
   server:
+    ingress:
+      annotations:
+        # Annotations required to enable basic authentication for any ingress
+        # into prometheus server from the outside world - this secret is
+        # created via the prometheus-htpasswd.yaml file, controlled by config
```
consideRatio, line 36: I don't understand that "this secret is created via the prometheus-htpasswd.yaml file, can you elaborate?

To clarify my confusion, I'm thinking `"The referenced k8s Secret is created by the Helm chart via config under prometheusIngressBasicAuth"`, and I lack knowledge about a `prometheus-htpasswd.yaml` file. I figure the k8s Secret will be consumed by the ingress controller independent of any prometheus logic.

yuvipanda: @consideRatio https://github.com/2i2c-org/infrastructure/pull/1091/files#diff-7042b93ab64b7c665e0172a141323f9e75f60d0fc7d06fce61dad6bc4ba41fff is prometheus-htpasswd.yaml, the actual file that is creating the k8s secret.

yuvipanda: Clarified by renaming file as well as saying `templates/<filename>`

consideRatio: Oh when you said "the file that is creating the k8s secret" I got thinking something like `kubectl create secret generic --from-file ...`. But I understand it is a Helm chart template file that renders the k8s Secret resource! I figure if we clarify that it is the file mentioned is a helm template, it would be a lot clearer to me.

Can we also name the helm chart template file with a suffix of `secret.yaml`? It is very common practice in helm charts to name template files in a way that includes the k8s resource type, and it has been written in the Helm docs in the past at least that they suggest one resource per template file. Currently I see they have updated [their docs](https://helm.sh/docs/chart_template_guide/getting_started/) to say the less strict advice of:

> TIP: Template names do not follow a rigid naming pattern. However, we recommend using the suffix .yaml for YAML files and .tpl for helpers.

yuvipanda: @consideRatio ah, I'm not a big fan of adding the type of resource being created to the name of the file, preferring to signal intent instead. Just naming it the same as resource type works great when you're deploying a single application, or when it's under a subdirectory whose name explains what is being deployed. I'd like to not have both a name that explains intent and a suffix. It also reminds me a lot of [hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation). I'd prefer to keep it out if that's ok with you.

I've reworded the comment to make it clearer both where the secret is generated, as well as what feeds content in there.

helm-charts/support/templates/prometheus-ingress-auth.yaml
```diff
@@ -0,0 +1,10 @@
+
+{{- if .Values.prometheusIngressAuthSecret.enabled }}
```
consideRatio, line 2:
```suggestion
{{- if .Values.prometheusIngressAuthSecret.enabled -}}
```

helm-charts/support/templates/prometheus-ingress-auth.yaml
```diff
@@ -0,0 +1,10 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: prometheus-ingress-auth-basic
+type: Opaque
+data:
+  auth: {{ htpasswd .Values.prometheusIngressAuthSecret.username .Values.prometheusIngressAuthSecret.password | b64enc | quote}}
+{{- end }}
```
consideRatio, line 10:
```suggestion
{{- end }}

```

consideRatio (commented):
This LGTM, if needed, lets iterate from here to not slow down too much!

I think the key points now with relation to this PR is that this goes with the creation of additional domain names etc and those aspects may need documentation updates.

helm-charts/support/values.yaml
consideRatio: Oh okay - It gives me some itch as the convention is strong and has never been broken in the helm chart's I recall seeing besides the occational `rbac.yaml` file that includes multiple rbac related resource kinds - which in its name indicates what kind of resources it includes though.

Let's go with this though if you feel strongly about it!

yuvipanda: @consideRatio i've made a directory called prometheus-ingress-auth and moved this to a secret.yaml file there. I think this should both satisfy your itch as well as my desire to avoid hungarian notation.

consideRatio: :D You are a kind person Yuvi!

helm-charts/support/templates/prometheus-htpasswd.yaml
yuvipanda: Done!

helm-charts/support/templates/prometheus-ingress-auth.yaml
yuvipanda: Done, I think!

helm-charts/support/templates/prometheus-ingress-auth.yaml
yuvipanda: Done!

helm-charts/support/values.yaml
yuvipanda: @consideRatio 
image

yuvipanda:
I've added docs on what needs to happen with DNS, and I think it's already happened for all the clusters mentioned here.

docs/howto/operate/grafana.md
```diff
@@ -28,9 +28,13 @@ These tools are [`ingress-nginx`](https://kubernetes.github.io/ingress-nginx/),
 In the `infrastructure` repo, the full filepath should be: `config/clusters/<cluster_name>/support.values.yaml`.
 
 Add the following helm chart values to your `support.values.yaml` file.
-`<grafana-domain>` should follow the pattern `grafana.<cluster_name>.2i2c.cloud`.
+`<grafana-domain>` should follow the pattern `grafana.<cluster_name>.2i2c.cloud`,
+and `<prometheus-domain>` should follow the pattern `prometheus.<cluster_name>.2i2c.cloud`.
 
 ```yaml
+prometheusEngressAuthSecret:
```
consideRatio, line 35:
```suggestion
prometheusIngressAuthSecret:
```

docs/howto/operate/grafana.md
```diff
@@ -39,8 +43,40 @@ grafana:
...
+        - secretName: prometheus-tls
+          hosts:
+            - <prometheus-domain>
+```
+
+#### Create a `enc-support.secret.yaml` file
+
+Only 2i2c staff + our centralized grafana should be able to access the
+prometheus data on a cluster from outside the cluster. The [basic auth](https://kubernetes.github.io/ingress-nginx/examples/auth/basic/)
+feature of nginx-ingress is used to restrict this. A `enc-support.seecret.yaml`
```
consideRatio, line 63: I think it will be critical that this is named `enc-support.secret.values.yaml` --- that it includes `values.yaml` in the end to be compatible with https://github.com/2i2c-org/infrastructure/pull/1093/files#diff-6d3cb43608c85f4e234a8decb5a6dc2f7f62e5606891b377941a201fd8222d25R77.

```suggestion
feature of nginx-ingress is used to restrict this. A `enc-support.secret.values.yaml`
```

docs/howto/operate/grafana.md
```diff
@@ -50,6 +86,7 @@ Note this filepath is _relative_ to the location of your `cluster.yaml` file.
 support:
   helm_chart_values_files:
     - support.values.yaml
+    - enc-support.secret.yaml
```
consideRatio, line 89: TODO: Systematic rename of this file to `enc-support.secret.values.yaml`

helm-charts/support/values.yaml
```diff
@@ -24,6 +29,20 @@ prometheus:
   pushgateway:
     enabled: false
   server:
+    ingress:
+      annotations:
+        # Annotations required to enable basic authentication for any ingress
+        # into prometheus server from the outside world. This secret is
+        # created via templates/prometheus-ingress-auth/secret.yaml file in the support chart,
+        # and the contents are # controlled by config under prometheusIngressAuthSecret.
```
consideRatio, line 37:
```suggestion
        # and the contents are controlled by config under prometheusIngressAuthSecret.
```

consideRatio (approved):
With these comments resolved, this LGTM! Nice work on this @yuvipanda!

docs/howto/operate/grafana.md
yuvipanda: I thought the secret.yaml is an alternative to .values.yaml, specifying what they are. Why have both?

yuvipanda: Either way, happy to just do this for now and move the conversation about conventions elsewhere :)

docs/howto/operate/grafana.md
yuvipanda: Just done.

deployer/cluster.py
yuvipanda: Done!

deployer/cluster.py
```diff
@@ -112,14 +112,15 @@ def deploy_support(self):
...
         support_dir = (Path(__file__).parent.parent).joinpath("helm-charts", "support")
-        subprocess.check_call(["helm", "dep", "up", support_dir])
-
-        support_secrets_file = support_dir.joinpath("enc-support.secret.yaml")
-        # TODO: Update this with statement to handle any number of context managers
-        #       containing decrypted support values files. Not critical right now as
-        #       no individual cluster has specific support secrets, but it's possible
-        #       to support that if we want to in the future.
-        with get_decrypted_file(support_secrets_file) as secret_file:
+        # subprocess.check_call(["helm", "dep", "up", support_dir])
```
GeorgianaElena, line 115: I believe this also needs to be restored?

deployer/cluster.py
```diff
@@ -73,43 +73,43 @@ def deploy_support(self):
...
+            [
+                "helm",
+                "upgrade",
+                "--install",
+                "--create-namespace",
+                "--namespace=cert-manager",
+                "cert-manager",
+                "jetstack/cert-manager",
+                f"--version={cert_manager_version}",
+                "--set=installCRDs=true",
```
consideRatio, line 107: I'll open a separate PR with this change after this is merged to avoid merge conflicts.

This is an optimization to not need to: a) add a named helm chart repo, b) updating all helm chart repos, c) referencing a helm chart repo via named helm chart repo. Instead one would simply name the helm chart and declare the repo, which is fetched just in time of the upgrade.

```suggestion
        print_colour("Provisioning cert-manager...")
        subprocess.check_call(
            [
                "helm",
                "upgrade",
                "cert-manager",  # given release name (aka. installation name)
                "cert-manager",  # helm chart to install
                f"--repo={cert_manager_url}",
                "--install",
                "--create-namespace",
                "--namespace=cert-manager",
                f"--version={cert_manager_version}",
                "--set=installCRDs=true",
```

deployer/cluster.py
consideRatio: Nice catch!!

consideRatio:
```suggestion
        subprocess.check_call(["helm", "dep", "up", support_dir])
```

deployer/cluster.py
yuvipanda: Perfect! Would definitely <3 a new PR!

yuvipanda:
I think what's left here is to automate adding all these as a data source to one centralized grafana.

deployer/cluster.py
yuvipanda: aha! restored, thank you :)

yuvipanda:
I think this is ready to go!

docs/howto/operate/grafana.md
sgibson91: All values files can be secret, but not all secrets need to be values files. So I think we should adopt a `.secret.values.yaml` convention to indicate that this is a secret values file. See the notes box in the docs here: https://infrastructure.2i2c.org/en/latest/topic/config.html#id1

helm-charts/support/values.yaml
consideRatio: I opened https://github.com/2i2c-org/infrastructure/issues/1101 about this.

deployer/cluster.py
consideRatio: Retained memory of doing this in #1100

yuvipanda:
Still need to write code that'll populate central grafana with datasources, but let's get this merged until then.

damianavila:
> Still need to write code that'll populate central grafana with datasources

I think the upcoming work is actually described here: https://github.com/2i2c-org/infrastructure/issues/328#issuecomment-1082542191, let me know if you disagree.

Btw, thanks for all this work, @yuvipanda!!

Issue #1031, opened 2022-02-24
[Incident] Dask Gateway fails with Unavailability of Compute Engine Resources - turned out to be a transient issue


choldgraf:
# Summary

### Background

The Pangeo community recently tried to update their user environment image via the Configurator. When user sessions started with the new image, Dask Gateway was no longer functional.

When the image was bumped to `2022.02.04` then running the following code:

```
from dask_gateway import GatewayCluster
cluster = GatewayCluster()
cluster.scale(20)
```

Would not function properly. Looking at some of the logs in the GCP console, we saw these errors:

<img src="https://user-images.githubusercontent.com/1839645/155581517-8aedb94f-2526-49c6-8dc3-dec7b4807019.png" width=300 />

### Resolution

Ultimately, we realized that this was likely a **transient issue** due to Google Cloud's resource limits being reached, and the fact that we are not paying for dedicated guaranteed resources.

### Relevant information

- Hub URL: http://us-central1-b.gcp.pangeo.io/
- FreshDesk ref: https://2i2c.freshdesk.com/a/tickets/82
- Pangeo image repo is here: https://github.com/pangeo-data/pangeo-docker-images
- A diff of the repo changes between these two versions: https://github.com/pangeo-data/pangeo-docker-images/compare/2021.10.19...2022.02.04

### Actions to resolve

We'll need to collaborate with folks in the Pangeo community who are familiar with the user environment images (maybe that is @scottyhq and @rabernat ? Please cc others who may have insight). We could then do some combination of:

- Look at the diff in the user image, and try to spot anything that would obviously stop Dask Gateway from scaling: https://github.com/pangeo-data/pangeo-docker-images/compare/2021.10.19...2022.02.04
- If that doesn't work, then work backwards with _earlier releases_ of the image, running the code snippet above to see if it works. When we find an image environment that works, then we can figure out the exact diff that broke it.
- [x] Work with members of the Pangeo team to identify the problem and resolve it
- [x] Incident has been dealt with or is over
- [x] Sections below are filled out
- [x] Incident title and after-action report is cleaned up
- [x] All actionable items above have linked GitHub Issues

# Timeline

All times in US/Pacific.

### 2022-02-17 - First reports

- @rabernat reported that the Dask Gateway cluster was not scaling properly
- After some investigation, we realized that this happened just after bumping the user image to `2022.02.04`.
- Reverting the image made things work again, but now we cannot update the user image without breaking things.

[...]

PR #1092, opened 2022-03-14
Use common.values.yaml files where config is shared between hubs on a cluster

sgibson91:
Previously, when we had staging/prod hub pairs, our workflow was to define all config in `staging.values.yaml`, define config that only applies to the `prod` hub in `prod.values.yaml` and then parse only the staging file when we upgrade staging, and **both** the staging and prod files when upgrading prod.

I think this was probably a little conceptually confusing, and definitely caused confusion for hub admins trying to locate config. See https://github.com/2i2c-org/infrastructure/issues/1031#issuecomment-1065516737

This PR refactors helm chart values files so that we now define shared config between paired hubs in a `common.values.yaml` and then provide hub-specific tweaks in a `staging.values.yaml` or `prod.values.yaml` file. Hopefully, this will reduce confusion in the future.

### DIFF
--- a/docs/topic/config.md
+++ b/docs/topic/config.md
@@ -67,7 +67,7 @@ A hub's helm chart values file can be encrypted as well, following the naming co
 ```
 
 Where we run dedicated clusters that only host a `staging` and `prod` hub, we aggregate all helm chart values shared by each hub into a `common.values.yaml` file, and then describe the helm chart values specific to either `staging` or `prod` with a `staging.values.yaml` or `prod.values.yaml` file respectively.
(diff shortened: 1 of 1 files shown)


GeorgianaElena (approved):
+1 on having the common config into its own file. Thanks @sgibson91 🚀

choldgraf:
Agree this is a conceptually simpler approach, and matches other patterns I've seen in projects in this ecosystem 👍

damianavila:
Super, @sgibson91!
I totally agree the previous iteration was confusing.

PR #1093, opened 2022-03-14
Add logic to the deployer to decide which clusters/hubs require a helm upgrade

sgibson91:
This PR adds a series of functions to the deployer that will analyse a set of filepaths (ideally that have been added/changed by a GitHub PR, this input is not provided in this PR) and decide which hubs/clusters require a helm upgrade and whether that upgrade is for the *hub helm charts or the support helm chart

It's logic works like so:

**Hubs:**

- If a `*.values.yaml` file has been added/changed, upgrade the appropriate hub on the appropriate cluster
- If a `cluster.yaml` file has been added/changed, upgrade all hubs on that cluster
- If the `basehub` or `daskhub` charts have been changed, upgrade all hubs on all clusters

**Support:**

- If a `support.values.yaml` or `cluster.yaml` file has changed, upgrade support for that cluster
- If the support helm chart has changed, upgrade support on all clusters

Some extra filepaths are also watched to maintain the functionality we have currently, such as the deployer and the files under `.github/actions/deploy`.

The output is two lists of dictionaries that each represent a set of matrix jobs to be read by a second GitHub Actions job that will run each dictionary in parallel. One dictionary represents all the support upgrade jobs, the other all the hub upgrade jobs. A `--pretty-print` flag has been provided that will print these jobs in a human-readable format by using tables from the `rich` package. It is intended to run this logic in PRs using the pretty print option so engineers can check which jobs will be run before merging.

The support jobs have the keys `provider` (the cloud provider) and `cluster_name` (the name of the cluster). The hub jobs have the same keys plus an additional `hub_name` key (the name of a hub). These will be fed to the `deployer deploy` command to begin a deployment.

I have also added some tests to ensure all this new logic is behaving how we expect it to. I've added an environment variable that is set to `True` when we run the tests via `pytest` this allows us to ignore `cluster.yaml` files nested under `tests/` when we are doing live deployments, and exclusively consider them when we are running the test suite. I have added a new GitHub Actions workflow file to run these new tests when changes are made to the deployer.

**NOTE:** This PR is not yet implementing the changes in GitHub Actions to actually deploy hubs in parallel. This is only the Python logic that will underpin such a workflow.


deployer/helm_upgrade_decision.py
```diff
@@ -0,0 +1,358 @@
...
+    Args:
+        input_file_list (list[str]): A list of files that have been added or modified
+            in a GitHub Pull Request
+    Returns:
+        list[path obj]: A list of unique filepaths to cluster folders
+        set[str]: A set of all files matching the pattern "*/cluster.yaml"
+        set[str]: A set of all files matching the pattern "*/*.values.yaml"
+        set[str]: A set of all files matching the pattern "*/*support*.values.yaml"
+    """
+    patterns_to_match = ["*/cluster.yaml", "*/*.values.yaml", "*/*support*.values.yaml"]
```
consideRatio, line 77: With this, we will start relying on (but not enforcing here) the previously non-strict naming convention of having `values.yaml` in the end of all values files.

Adding this strict filter pattern makes me think we should enforce `values.yaml` suffix in `helm_chart_values_files`.

I'm brainstorming an idea that one could not need a filter pattern at all by detecting/inspecting all clusters and inspecting their helm_chart_values_yaml files etc for relevant files to observe for changes. Given how this implementation is now though, I suggest its pair with enforcement of the files listed in `helm_chart_values_files` will lead to matching the patterns defined here.

deployer/helm_upgrade_decision.py
```diff
@@ -0,0 +1,358 @@
...
+from rich.table import Table
+
+yaml = YAML(typ="safe", pure=True)
+
+# Determine if we are running a test or not. We set this env var to true in the
+# pytest.ini file so it is set when the package is tested.
+test_env = os.getenv("RUN_ENV", False)
+
+
+def discover_modified_common_files(modified_paths: list):
```
consideRatio, line 20: I'm not sure what this function should be named, but I'm uneasy that it returns a critical decision if to upgrade or not and it's not something conveyed by the name.

image

deployer/helm_upgrade_decision.py
```diff
@@ -0,0 +1,358 @@
...
+    """There are certain common files which, if modified, we should upgrade all hubs
+    and/or all clusters appropriately. These common files include the helm charts we
+    deploy, as well as the GitHub Actions and deployer package we use to deploy them.
+    Args:
+        modified_paths (list[str]): The list of files that have been added or modified
+            in a given GitHub Pull Request.
+    Returns:
+        upgrade_all_clusters (bool): Whether or not all clusters should be upgraded
+            since the support chart has changed
```
consideRatio, line 29: I'm not confident the logic of upgrading all clusters and helm charts in them makes sense just because we change the support chart we deploy on all/most clusters, but that is probably matching the current behavior we have so :+1: for now!

deployer/helm_upgrade_decision.py
```diff
@@ -0,0 +1,358 @@
...
+    been modified.
+    Args:
+        hub_matrix_jobs (list[dict]): A list of dictionaries which describe the set of
+            matrix jobs required to update only the hubs on clusters whose config has
+            been modified.
+        support_matrix_jobs (list[dict]):  A list of dictionaries which describe the
+            set of matrix jobs required to update only the support chart on clusters
+            whose config has been modified.
+    """
+    with open(os.getenv("GITHUB_ENV"), "a") as f:
```
consideRatio, line 325:
```suggestion
    # A GitHub Workflow's subsequent steps environment variables is influenced by
    # what is written to a file at the `GITHUB_ENV` location.
    #
    # For more details, see: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable
    #
    with open(os.getenv("GITHUB_ENV"), "a") as f:
```

choldgraf:
This is super exciting, thanks @sgibson91 !! It would be great if we can get more thoughts from the @2i2c-org/tech-team since this will be a big change to our CI/CD setup.

A few guiding principles thoughts from me:

- I like @consideRatio's comments about choosing function names that are tightly scoped and convey the most important information about what a function does
- If there are places where we might optimize behavior about the upgrade logic, I suggest that (unless it is an obvious and easy fix) we create issues to tackle them in follow-up PRs, and make sure this PR isn't making it _impossible_ to make those improvements in the future.

A quick question, I noticed:

> This PR is not yet implementing the changes in GitHub Actions to actually deploy hubs in parallel

Is the plan to build the GitHub Action that runs these commands in this PR as well, or is that a follow-up PR?

sgibson91:
> Is the plan to build the GitHub Action that runs these commands in this PR as well, or is that a follow-up PR?

My intention was for it to be a follow-up

consideRatio:
> If there are places where we might optimize behavior about the upgrade logic, I suggest that (unless it is an obvious and easy fix) we create issues to tackle them in follow-up PRs, and make sure this PR isn't making it impossible to make those improvements in the future.

Absolutely agree! I think @sgibson91 has done wonders breaking apart a very large kind of change into a sequence of changes in PRs. Feel free to proceed as needed so that work doesn't get stuck as large changes often can get when trying to address everything all at once!

Thank you @sgibson91 for running this ultra-marathon!!

deployer/helm_upgrade_decision.py
yuvipanda: Can you help me understand why the .values.yaml prefix is needed, vs just a .yaml?

yuvipanda: Actually, just ignore that question :) I don't think that's particularly important at this point, we can always re-evaluate later if needed.

consideRatio: We have `helm_chart_values_files` configuration, and then we have naming conventions. With this PR we currently end up requiring both explicit configuration and following the naming convention, but we currently don't enforce the naming convention.

Overall, I appreciate if we follow a naming convention of having `values` be included in all helm chart values files because it simply clarifies something that isn't obvious. Anything including `secret` doesn't by any mean say to me that _this is a helm chart values file_ so I appreciate the verbosity and convention of including `values` in the name of a helm chart values file.

If we adopt this PRs currently logic that relies on file name patterns rather than reading the explicit configuration in `helm_chart_values_files` for each cluster to do the smart CI/CD logic, that is an acceptable compromise as long as the convention is enforced as well.

sgibson91: This convention also helps us differentiate between other yaml files we have in the repo, like GitHub Actions workflow files.

sgibson91:
Folks, I tried to do a `git rebase` to drop a commit which was causing a merge conflict(and didn't really need to be in this PR, related to broken links in the docs) but the rebase has brought in other commits from https://github.com/2i2c-org/infrastructure/pull/1092, see below 😭 

Screenshot 2022-03-15 at 11 06 33

 I apologise for this faux pas and would appreciate any learning resources to help me actually use `git rebase` effectively without me messing up PRs like this 😢

consideRatio:
@sgibson91 I think `git rebase` has been tricky to learn more about, and I feel that I've only managed to incrementally add understanding over time - but, the biggest increments of understanding came from watching two amazing video-focused courses:
1. https://www.pluralsight.com/courses/how-git-works
2. https://www.pluralsight.com/courses/master-git

Thanks for clarifying what was part of this PR and what wasn't!

.github/workflows/test-deployer-code.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - uses: actions/setup-python@v3
+        with:
+          python-version: "3.9"
+      - name: Install dependencies
+        run: |
+          python -m pip install -U pip
+          python -m pip install -r requirements.txt
+      - name: Run tests
+        run: |
+          python -m pytest -vvv
```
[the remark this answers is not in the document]
sgibson91: IIUC, using `python -m <package_name>` is a convention that ensures that the package you are invoking uses the installation of python on your path (most likely your intended version) as opposed to any other python installations on the system. And this convention is most helpful when there are multiple virtual environments at play. And I think it's safe to assume that there are probably a few virtual environments that are hidden from us active on a GitHub Actions runner.

So, my conclusion is that it's probably not a requirement to have it this way, and we'd probably be fine without it, but it also doesn't harm us either.

.github/workflows/test-deployer-code.yaml
```diff
@@ -31,4 +31,4 @@ jobs:
           python -m pip install -r requirements.txt
       - name: Run tests
         run: |
-          python -m pytest -vvv --color=always
+          python -m pytest -vvv --color=yes
```
consideRatio, line 34: Aaargh woops, I always mistake `git diff --color=always` with `pytest --color=yes`...

In this case, I even did a manual check to verify I got it right --- the issue was that i verified that `always` should be used for `git diff`, not for `pytest` ^^

deployer/helm_upgrade_decision.py
```diff
@@ -0,0 +1,358 @@
...
+        else:
+            print(f"No support defined for cluster: {cluster_config.get('name', {})}")
+
+    return matrix_jobs
+
+
+def update_github_env(hub_matrix_jobs, support_matrix_jobs):
+    """Update the GITHUB_ENV environment with two new variables describing the matrix
+    jobs that need to be run in order to update the support charts and hubs that have
+    been modified.
```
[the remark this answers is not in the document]
sgibson91: So the decision is made to print rich tables if EITHER the `--pretty-print` argument is passed OR `GITHUB_ENV` is found in the environment:

https://github.com/2i2c-org/infrastructure/pull/1093/files#diff-a3ba5c5035ddef3b5c64742f652b5c0e14075ed68b82ee882c884d514b259817R267-R268

> I'm also thinking perhaps we should fail loudly and early if we end up running functions that assume we are doing it as part of a GitHub workflow execution - but aren't.

So we don't fail, but instead print out human-readable output instead of trying to write to the GITHUB_ENV file

deployer/cli.py
```diff
@@ -92,6 +100,24 @@ def main():
         parents=[base_parser],
         help="Modify the current kubeconfig with the deployer's access credentials for the named cluster",
     )
+
+    # generate-helm-upgrade-jobs subcommand
+    # This subparser does not depend on the base parser.
+    generate_helm_upgrade_jobs_parser = subparsers.add_parser(
+        "generate-helm-upgrade-jobs",
+        help="Generate a set of matrix jobs to perform a helm upgrade in parallel across clusters and hubs",
```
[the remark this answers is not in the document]
sgibson91: Done [b4b0287](https://github.com/2i2c-org/infrastructure/pull/1112/commits/b4b028760cf2f98f57f66e2154b6e5fd9db0c082)

deployer/helm_upgrade_decision.py
```diff
@@ -0,0 +1,358 @@
...
+        "requirements.txt",
+        ".github/workflows/deploy-hubs.yaml",
+    ]
+    # If this filepath has changes, we should update the support chart on all clusters
+    support_chart_filepath = "helm-charts/support/*"
+
+    # Discover if the support chart has been modified
+    support_matches = []
+    support_matches.extend(fnmatch.filter(modified_paths, support_chart_filepath))
+    upgrade_all_clusters = len(support_matches) > 0
```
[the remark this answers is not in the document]
sgibson91: [d843e43](https://github.com/2i2c-org/infrastructure/pull/1112/commits/d843e43ad4531cbf3d61c66f4eacb44577222344)

.github/workflows/test-deployer-code.yaml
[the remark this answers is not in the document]
sgibson91: Hmm, maybe it was important after all? https://github.com/2i2c-org/infrastructure/runs/5555841473?check_suite_focus=true

deployer/helm_upgrade_decision.py
sgibson91: Done [e25d452](https://github.com/2i2c-org/infrastructure/pull/1112/commits/e25d4522e767b6c360d2af093507852aad92f6b0)

sgibson91:
I'm closing this PR in favour of https://github.com/2i2c-org/infrastructure/pull/1112 I've done my best to copy across unresolved conversations

PR #1096, opened 2022-03-14
Rename pilot to docs

choldgraf:
Renames the use of "pilot" to "docs" where applicable, and cleans up some other small docs things.

ref: https://github.com/2i2c-org/docs/issues/125

### DIFF
--- a/docs/howto/operate/new-hub/aws.md
+++ b/docs/howto/operate/new-hub/aws.md
@@ -22,10 +22,10 @@ existing cluster.
 
 Follow the steps outlined in [](new-hub:deploy) with the following modifications:
 
-1. Generate a new values file for your hub if there is not an existing one.
+1. Generate a new configuration file for your hub if there is not an existing one.
 
    You can use one of the existing hub values files as a "template" for your hub (for
-   example, [here is the Farallon Institute staging hub values file](https://github.com/2i2c-org/infrastructure/tree/HEAD/config/clusters/farallon/staging.values.yaml)).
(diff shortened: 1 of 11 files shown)


docs/howto/operate/new-hub/aws.md
```diff
@@ -22,10 +22,10 @@ existing cluster.
 
 Follow the steps outlined in [](new-hub:deploy) with the following modifications:
 
-1. Generate a new values file for your hub if there is not an existing one.
+1. Generate a new configuration file for your hub if there is not an existing one.
 
    You can use one of the existing hub values files as a "template" for your hub (for
-   example, [here is the Farallon Institute staging hub values file](https://github.com/2i2c-org/infrastructure/tree/HEAD/config/clusters/farallon/staging.values.yaml)).
+   example, [here are the configuration files for the Farallon Institute cluster](https://github.com/2i2c-org/infrastructure/blob/master/config/clusters/farallon/common.values.yaml)).
```
consideRatio, line 28: If we want to switch from master to main, we can use `HEAD` for now and it will work no matter what to reference the default branch.

```suggestion
   example, [here are the configuration files for the Farallon Institute cluster](https://github.com/2i2c-org/infrastructure/blob/HEAD/config/clusters/farallon/common.values.yaml)).
```

choldgraf: great point

consideRatio:
Oh no, now I got relaxed and suddenly it was in the infrastructure repo =/

choldgraf:
haha I think this is fine - the PR is only touching docs and tests are happy. IMO we should just try to be reasonable and use our best judgment about whether merging a PR is controversial or not. IMO in this case it's OK

PR #1097, opened 2022-03-15
Prevent unencrypted files from being checked in

yuvipanda:
I wrote https://github.com/yuvipanda/pre-commit-hook-ensure-sops
to prevent possible unencrypted files from being checked-in! This
reduces need to keep unencrypted files on disk at rest in our
local checkouts as well.

Any file with the word 'secret' in its path will be checked to
make sure it is encrypted with sops. Exclusions will have to
be explicitly listed in our .pre-commit-config.yaml file - I'd
rather we be over-aggressive with this than not, given the
possible dire consequences of a credential leak.

I moved the sops encrypted ssh keys we use for our eksctl / kops
clusters into a directory named 'secret' so they are also checked


yuvipanda:
I also want to get the pre-commit hook itself moved into the 2i2c org - https://github.com/yuvipanda/pre-commit-hook-ensure-sops. What do folks think?

choldgraf:
I think this is a clever idea to keep us from accidentally committing something we don't want to!

sgibson91 (approved):
🎉🎉🎉

yuvipanda:
Thanks, @sgibson91!

PR #1098, opened 2022-03-15
Use sops --in-place encryption

yuvipanda:
This prevents the possibility of unencrypted secrets left
in our current working checkout. .gitignore prevents these
from being committed, but it's nice to not have those lying
around regardless.

Relies on https://github.com/2i2c-org/infrastructure/pull/1097 to
prevent unencrypted files from being committed.

### DIFF
--- a/docs/howto/configure/auth-management.md
+++ b/docs/howto/configure/auth-management.md
@@ -73,7 +73,7 @@ Presently, this involves a few more manual steps than the `auth0` setup describe
    - The authorisation callback URL is the homepage url appended with `/hub/oauth_callback`. For example, `staging.pilot.2i2c.cloud/hub/oauth_callback`.
    - Once you have created the OAuth app, make a new of the client ID, generate a client secret and then hold on to these values for a future step
 
-2. **Create or update the appropriate secret config file under `config/clusters/<cluster_name>/<hub_name>.secret.values.yaml`.**
+2. **Create or update the appropriate secret config file under `config/clusters/<cluster_name>/enc-<hub_name>.secret.values.yaml`.**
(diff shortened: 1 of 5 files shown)


yuvipanda:
What do you think of this, @sgibson91?

yuvipanda:
@sgibson91 I was hoping https://github.com/2i2c-org/infrastructure/pull/1097 will mean we don't have to keep unencrypted data on our local disks, even if they're encrypted in git. So `--in-place` will mean that the unencrypted copy is less likely to lie around, compared to using `--output`. 

And yeah, the logic for erroring out in the deployer doesn't need `enc-` I think.

sgibson91:
To be clear, I think I'm happy with any naming convention that allows us to be strict about checking secret files for encryption. My personal worry is that it'll be janky for me to rely on the pre-commit hook _solely_ to do this.

yuvipanda:
@sgibson91 yeah, given the high stakes nature of this, I'm happy for us to have both the pre-commit hook as well as the enc- convention + gitignore in place.

sgibson91:
I will approve a PR that adds the use of the pre-commit hook to the docs as a complementary method for ensuring secrets are encrypted

yuvipanda:
@sgibson91 https://github.com/2i2c-org/infrastructure/pull/1097 is that PR I think? It only adds the pre-commit hook, doesn't touch anything that already exists.

sgibson91:
I meant documenting it's use in the files we were changing here. So adding a section to this page describing the hook, how to install/use it and where it's maintained https://infrastructure.2i2c.org/en/latest/topic/secrets.html

PR #1099, opened 2022-03-15
Keep at least one node up at all times for uwhackweeks

yuvipanda:
After making this change, I did:

```
jsonnet uwhackweeks.jsonnet > uwhackweeks.yaml
eksctl scale nodegroup --config-file=uwhackweeks.yaml
```

This brought up a node!

I authenticated via https://2i2c.awsapps.com/start#/,
and the AWS SSO setup.

Ref https://2i2c.freshdesk.com/a/tickets/100

### DIFF
--- a/eksctl/uwhackweeks.jsonnet
+++ b/eksctl/uwhackweeks.jsonnet
@@ -16,10 +16,10 @@ local namespaces = ['staging', 'prod'];
 // A `node.kubernetes.io/instance-type label is added, so pods
 // can request a particular kind of node with a nodeSelector
 local notebookNodes = [
-    { instanceType: "m5.large" },
-    { instanceType: "m5.xlarge" },
-    { instanceType: "m5.2xlarge" },
-    { instanceType: "m5.8xlarge" },
+    { instanceType: "m5.large", minSize: 1 },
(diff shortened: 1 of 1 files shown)


yuvipanda:
I've already deployed this.

consideRatio (approved):
LGTM I think, and given that it is already deployed and functional it feels kind of safe for me to approve!

sgibson91:
Thank you @yuvipanda!

damianavila:
Can we capture this process in some documentation bits?

For instance, it is important to highlight you are passing a config file to the `eksctl scale` command which is different from 
`eksctl create` command referenced here (https://infrastructure.2i2c.org/en/latest/howto/operate/new-cluster/aws.html#create-the-cluster) even when we use the same config file!
AFAIK (and I might miss something) the `scale` command is only looking at the scale-relevant pieces from that config file.

`eksctl` does NOT do cluster reconciliation as terraform yet, so this might confuse people, maybe?

yuvipanda:
@damianavila yeah I agree that our eksctl process is not documented well at all. Needs fixing

yuvipanda:
@damianavila follow-up docs in https://github.com/2i2c-org/infrastructure/pull/1138!

PR #1104, opened 2022-03-15
Revert the explicit raising of an error if an encrypted file is not valid JSON or YAML

sgibson91:
I implemented this on an assumption that has since been shown to be false in https://github.com/yuvipanda/pre-commit-hook-ensure-sops/issues/1 So perhaps it is no longer required?

### DIFF
--- a/deployer/file_acquisition.py
+++ b/deployer/file_acquisition.py
@@ -3,7 +3,6 @@ Functions related to finding and reading files. Checking files exist, finding th
 absolute paths, decrypting and reading encrypted files when needed.
 """
 import os
-import json
 import warnings
 import subprocess
 import tempfile
--- a/deployer/file_acquisition.py
+++ b/deployer/file_acquisition.py
@@ -107,26 +107,19 @@ def get_decrypted_file(original_filepath):
         # if the file is valid JSON/YAML, and then if it has a `sops` key
         with open(original_filepath) as f:
 
-            # FIXME: Right now we expect encrypted files to be JSON or YAML files, and
-            #        so we fail if these are not valid JSON/YAML. However in the
-            #        future, we may want to support encrypted files of other types
-            #        and we should update this section accordingly.
-            #
             # Support the (clearly wrong) people who use .yml instead of .yaml
             if ext == ".yaml" or ext == ".yml":
                 try:
                     content = yaml.load(f)
                 except ScannerError:
-                    raise ScannerError(
-                        "We expect encrypted files to be valid JSON or YAML files."
-                    )
+                    yield original_filepath
+                    return
             elif ext == ".json":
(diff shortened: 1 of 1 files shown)


deployer/file_acquisition.py
```diff
@@ -107,26 +107,19 @@ def get_decrypted_file(original_filepath):
         # if the file is valid JSON/YAML, and then if it has a `sops` key
         with open(original_filepath) as f:
 
-            # FIXME: Right now we expect encrypted files to be JSON or YAML files, and
-            #        so we fail if these are not valid JSON/YAML. However in the
-            #        future, we may want to support encrypted files of other types
-            #        and we should update this section accordingly.
-            #
             # Support the (clearly wrong) people who use .yml instead of .yaml
             if ext == ".yaml" or ext == ".yml":
```
yuvipanda, line 111: All JSON is valid YAML as well by design, so you can actually just use the YAML parser here to load files unconditionally, without having to rely on extensions. If a YAML parser can't read it, then it's not encrypted by sops.

sgibson91: Done!

consideRatio (approved):
:tada:

sgibson91:
Uh-oh!

```
ERROR: gcloud crashed (ValueError): ('Unexpected credentials type', u'ENC[AES256_GCM,data:aRNvFJjtkTPJfckQAK7B,iv:pzD0TlNSO9CpeGkHUQmPfW6hSVM5rs7fNUX/79NdLU4=,tag:6X2G1z4Z4qOHM84yevLm7g==,type:str]', 'Expected', 'service_account')
```

deployer/file_acquisition.py
```diff
@@ -104,29 +103,14 @@ def get_decrypted_file(original_filepath):
...
-                try:
-                    content = json.load(f)
-                except json.JSONDecodeError:
-                    raise json.JSONDecodeError(
-                        "We expect encrypted files to be valid JSON or YAML files."
-                    )
+            try:
+                content = yaml.load(f)
+            except ScannerError:
+                yield original_filepath
```
consideRatio, line 112: Hmmm, it should not yield when the erroring - right? previously it raised an error on scanner error.

consideRatio: In this case, we matched the file path name pattern - and should error rather than yield. Later we shouldn't.

```python
    else:
        # For a file that does not match our naming conventions for secrets, yield the
        # original path
        yield original_filepath
```

sgibson91: I think I got confused when undoing my work in this commit https://github.com/2i2c-org/infrastructure/commit/7c676214729c81a5f5e0a7c59c15cc4799b52b17

PR #1106, opened 2022-03-15
Remove reliance on `enc-` prefix when decrypting secret files

sgibson91:
This was identified as redundant in https://github.com/2i2c-org/infrastructure/pull/1098#issuecomment-1067869844

### DIFF
--- a/deployer/file_acquisition.py
+++ b/deployer/file_acquisition.py
@@ -141,7 +141,7 @@ def get_decrypted_file(original_filepath):
             yield f.name
 
     else:
-        # The file does not have "secret" in it's name, therefore does not need to be
+        # The file does not have "secret" in its name, therefore does not need to be
         # decrypted. Yield the original filepath unchanged.
         yield original_filepath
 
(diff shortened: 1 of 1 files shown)


deployer/file_acquisition.py
```diff
@@ -143,8 +141,8 @@ def get_decrypted_file(original_filepath):
             yield f.name
 
     else:
-        # For a file that does not match our naming conventions for secrets, yield the
-        # original path
+        # The file does not have "secret" in it's name, therefore does not need to be
```
sgibson91, line 144:
```suggestion
        # The file does not have "secret" in its name, therefore does not need to be
```

yuvipanda:
@sgibson91 based on https://github.com/2i2c-org/infrastructure/pull/1098#issuecomment-1067885768, if you feel strongly about keeping `enc-` we can keep it!

sgibson91:
@yuvipanda I think the logic here is actually separate to our discussion on naming conventions. If we're looking for "secret" in the name, then we pick up all files, whether or not they have the `enc-` prefix. The important bit is to raise an error if we don't find a `sops` key in the content, as that points to leaked credentials.

yuvipanda:
Ah right - you can have a secret file that isn't encrypted and doesn't have enc- in your local checkout, but this code only hits things that are actually read by the deployer, and hence likely to be committed! makes sense.

yuvipanda (approved):

yuvipanda:
Thanks for explaining it to me, @sgibson91!

PR #1107, opened 2022-03-15
Revert "Revert the explicit raising of an error if an encrypted file is not valid JSON or YAML"

sgibson91:
Reverts 2i2c-org/infrastructure#1104


sgibson91:
Will merge without review since this reverts a PR that broke CI

PR #1110, opened 2022-03-15
Update prettier config to reformat secret files

sgibson91:
Hopefully the comment is self-explanatory. Related to #1109


sgibson91:
Hmmm, no. I feel this changes too much. I will close and think about with a better config.

PR #1111, opened 2022-03-15
Update utoronto user image to accomodate latest fixes

GeorgianaElena:
Particularly this one https://github.com/2i2c-org/utoronto-image/pull/17.

- Reference https://2i2c.freshdesk.com/a/tickets/65
- Available image tags https://quay.io/repository/2i2c/utoronto-image?tab=tags

cc @damianavila

### DIFF
--- a/config/clusters/utoronto/common.values.yaml
+++ b/config/clusters/utoronto/common.values.yaml
@@ -77,7 +77,7 @@ jupyterhub:
             subPath: "{username}"
     image:
       name: quay.io/2i2c/utoronto-image
-      tag: 18b55a7bcb96
+      tag: a716955d6147
   hub:
     db:
       pvc:
--- a/config/clusters/utoronto/common.values.yaml
+++ b/config/clusters/utoronto/common.values.yaml
@@ -77,7 +77,7 @@ jupyterhub:
             subPath: "{username}"
     image:
       name: quay.io/2i2c/utoronto-image
-      tag: a716955d6147
+      tag: 3aa146dc095f
   hub:
     db:
       pvc:
--- a/config/clusters/utoronto/common.values.yaml
+++ b/config/clusters/utoronto/common.values.yaml
@@ -77,7 +77,7 @@ jupyterhub:
             subPath: "{username}"
     image:
       name: quay.io/2i2c/utoronto-image
-      tag: 3aa146dc095f
+      tag: d422076c3695
   hub:
     db:
       pvc:


choldgraf:
Should we try this on `staging` first, before merging it to prod as well, to confirm that the fix works as-expected? I'm happy to go with your preference, just noting it here in case we didn't consider this.

GeorgianaElena:
You're totally right @choldgraf! We used to do this a lot when the toronto hub was in a separate repo.

I'll deploy this manually to the staging hub then ask for feedback in the freshdesk ticket about how the staging hub behaves.
What do you think @damianavila?

sgibson91:
> I'll deploy this manually to the staging hub then ask for feedback in the freshdesk ticket about how the staging hub behaves.

I think this would be a good practice to adopt for any hub image we more actively maintain

GeorgianaElena:
It might be that the last line of `Rprofile.site` gets ignored if there's no trailing new line 🤯 https://yihui.org/en/2018/04/rprofile-trailing-newline ?

sgibson91:
> It might be that the last line of `Rprofile.site` gets ignored if there's no trailing new line 🤯 [yihui.org/en/2018/04/rprofile-trailing-newline](https://yihui.org/en/2018/04/rprofile-trailing-newline) ?
> 
> I opened [2i2c-org/utoronto-image#26](https://github.com/2i2c-org/utoronto-image/pull/26) to try this.

This is _crazy_, but an excellent catch!!

choldgraf:
> It might be that the last line of Rprofile.site gets ignored if there's no trailing new line

lol this is ridiculous, thanks for investigating @GeorgianaElena , I guess the lesson to learn is "never expect that infrastructure will behave in a sensible manner" :-P

damianavila:
@GeorgianaElena, see my comment here: https://github.com/2i2c-org/utoronto-image/pull/26#issuecomment-1069699057

It seems your wonderful catch solves the env variable issue but wondering if you can actually load the dataset after that...

GeorgianaElena:
@damianavila, I just tried again and it seems to be working now. Wondering if the URL fetch error is was network-related and just transient? Do you want to ask Nathan on freshdesk to try out on staging hub https://staging.utoronto.2i2c.cloud and then we'll know for sure if there's yet another error we need to deal with 🙄 

loaded

damianavila:
I have already pinged them on the support ticket: https://2i2c.freshdesk.com/a/tickets/65?note=80121282739
I will report back when I hear something from them.

GeorgianaElena:
I just checked and they've reported seeing the `https` error (the one that somehow disappeared for me when I restarted the RStudio session 🙄 ). I've opened https://github.com/2i2c-org/utoronto-image/pull/27 hoping to be the fix of this.

GeorgianaElena:
So the PR linked above didn't solve the issue :( 

I also tried to uninstall miniconda using:
```
reticulate::miniconda_uninstall()
```

Still investigating.

GeorgianaElena:
I give up for tonight 😪 

A few things worth noting:

### Premise
1. I have two utoronto ids that I can use to login into the toronto hub at the moment (one of them was created so I can have access to the Azure subsription). The interesting part is that for the account I choose to install miniconda as part of this installation prompt presents the same issue reported in the support ticket and the other account just works.

staging-miniconda-offer

2. From inside a R notebook, both accounts work and the datasets get downloaded successfully.

### Leads and what I've tried
So it looks like installing miniconda and creating that reticulate env just messed with the location of ssl/openssl/certificates libs for python, basically what urllib needs to perform https requests (urllib is what keras uses in the back)

So I've tried to:
- bump `reticulate` version
- uninstall miniconda from within reticulate hoping this would undo the situation. It didn't :|
- look at PATH`, `LD_LIBRRY_PATH` and have them be the same like on the account that works
- search for local caches that could store some old config
- various random solutions I found on the internet

So far nothing worked, so giving up for today.

### Conclusion

Our current setup should work or any user that hasn't installed miniconda from reticulate. New users that load the reticulate lib for the first time shouldn't be prompted with that option (this we have fixed).

We just need to figure out how to fix currently broken  setups.

damianavila:
Thanks for this whole investigation, @GeorgianaElena!
Do you want to pair next week at some time so we can debug this together?
Without real access I can not do too much now... so maybe I can be a second pair of eyes with some conda experience 😉?

choldgraf:
@GeorgianaElena and @damianavila you both deserve lots of appreciation for all of the work you've put into this one. I just want to recognize that this issue is super unsustainable for you both, and we will work to avoid never-ending support situations like this in the future. 🙏🙏🙏

GeorgianaElena:
> Do you want to pair next week at some time so we can debug this together?

YES please! Let's sync on slack about this 🚀

damianavila:
OK, we met with @GeorgianaElena, tried a few things, and ended the meeting without hopes although we mostly agree this is probably an OpenSSL mismatch.

Fetching the latest image @GeorgianaElena was working on finally allowed me to reproduce the issue! So after the meeting, I kept debugging on the 2i2c staging hub 🤦:wink:.

It seems RStudio is loading the **system** OpenSSL stuff at the time to initialize...
```
$ ldd /usr/lib/rstudio-server/bin/rserver | grep ssl        libssl.so.1.1 => /usr/lib/x86_64-linux-gnu/libssl.so.1.1 (0x00007f0413833000)
```
and the system version is `1.1.1f`
```
$ dpkg -l | grep openssl
ii  libcurl4-openssl-dev:amd64           7.68.0-1ubuntu2.7                   amd64        development files and documentation for libcurl (OpenSSL flavour)
ii  openssl                              1.1.1f-1ubuntu2.12                  amd64        Secure Sockets Layer toolkit - cryptographic utility
```
But the **conda base** environment has another version, `1.1.1l`:
```
$ conda list openssl
# packages in environment at /opt/conda:
#
# Name                    Version                   Build  Channel
openssl                   1.1.1l               h7f98852_0    conda-forge
pyopenssl                 19.1.0                     py_1    conda-forge
$ openssl version
OpenSSL 1.1.1l  24 Aug 2021
```

So my theory was:
When R studio starts the session, it loads OpenSSL 1.1.1f stuff, and then, when you try to load the dataset, it uses that version but the python side of things is waiting for the 1.1.1l version living in the conda environment, then it fails!

To test if this specific mismatch was causing the problem... I went the hacky ninja path...
```
mv /opt/conda/lib/libssl.so.1.1 /opt/conda/lib/libssl.so.1.1.backup
ln -s /usr/lib/x86_64-linux-gnu/libssl.so.1.1 /opt/conda/lib/libssl.so.1.1
```
and it worked!

Screen Shot 2022-03-22 at 12 01 33

Although I am not sure about the other implications for that hacky symlink...

The bottom problem here is the fact we are installing RStudio "systemwide" instead of installing it on the conda environment... I am not sure if that is different in repo2docker (or even possible, although I see conda packages for Rstudio in anaconda.org).

Well, I guess we can keep the _status quo_ but we would need to make sure openssl versions on the different environments are actually the same to prevent this sort of mismatch, maybe that is our next step here?

GeorgianaElena:
> There is one more thing I don't have a clear picture about... why the mismatch manifests itself after installing miniconda

@damianavila, I double checked this morning loading an entirely new dataset using keras, using the account that never installed miniconda and it actually **failed**.

So miniconda might have nothing to do with this after all. And it probably worked  before because I had the datasets already downloaded from when I've tried it in using the R kernel? This also probably means the current keras setup we have won't work in any circumstances.

not-working

Sorry for putting you on a false lead!

damianavila:
https://github.com/2i2c-org/utoronto-image/pull/28 nor https://github.com/2i2c-org/utoronto-image/pull/29 work as expected... grrr...
The next step would be to just go wild with the "hack-ish" solution I ran above (based on symbolic links) after making sure we are not breaking other stuff (and providing users a "simple" way to remove the link if they see something weird in their workflows).
Open to other's thoughts...

yuvipanda:
@damianavila at this point I think we should say 'this is too broken, sorry we can not support this use case. You are welcome to try it out yourself and hack on the image'.

damianavila:
OK, let me summarize the whole situation and my next step around this one:

1. This PR is promoting the image obtained after https://github.com/2i2c-org/utoronto-image/pull/27 was merged.

This PR is NOT solving the `openssl` issue when you use RStudio, just updating the reticulate package, but it is the latest merged PR on the image repo, then we should sync that state with the current state in the cluster (which is what this PR is aiming for). So, I am advocating merging this PR.

2. We have already deployed TF+Keras with previous deployed images (and we probably have users for those packages), so we can not remove them. In fact, TF + Keras is fully functional from the Python side AND from the R side if you use the R kernel inside Jupyter instead of RStudio!

3. We just need to communicate to Toronto the underlying `openssl` issue with RStudio, the fact we tried multiple things that did not work, and the fact we are not comfortable supporting a hack involving symbolic links (and "invite" them to use the R-kernel inside the notebook to work around the issue at the time to download dataset or mentioning the symbolic link hack if they wish to go that route).

@2i2c-org/tech-team, further thoughts? Am I missing something else? Or do you want to present this to Toronto in a different way? Feedback is welcome, I would like to answer them before tomorrow's EOD if that is possible...

yuvipanda:
This sounds good to me, @damianavila. Also I didn't realize that starting it from Jupyter notebook fixes it! I don't think we should suggest they use the symlink hack because I'm 100% convinced that will cause difficult to debug issues elsewhere. Otherwise sounds great to me.

choldgraf:
Just so I understand the situation:

- It is possible to install Tensorflow and Keras in the same environment, in both Python and R library form.
- These can both be accessed via a Jupyter interface.
- Neither can be accessed via an RStudio interface.

And for that last point, is it related to *either* Tensorflow *or* Keras, or the combination of both of them specifically?

damianavila:
For the readers/followers, I have addressed @choldgraf's questions in a 1:1 meeting. A detailed explanation of the whole problem situation lives here: https://github.com/2i2c-org/utoronto-image/issues/6#issuecomment-1091954425

We will use that content to create a post in the Jupyter discourse forum so we share the experience with others in the community and with UToronto as well.

damianavila (approved):
@GeorgianaElena, this one LGTM (we should sync the state in the UToronto repo with the current state in the cluster).

GeorgianaElena:
@damianavila, I see other tags on quay that were pushed after the one in this PR
utornto-tags

But the sha  of the latest commit on the main branch is still this one. so I assume those tags are from PRs that weren't merged, right?

damianavila:
> so I assume those tags are from PRs that weren't merged, right?

Yep, that is correct... we are pushing from the PRs as well because that image fails to be built and tested in Binder.
You can see here the latest merge to main: https://github.com/2i2c-org/utoronto-image/actions/workflows/build.yaml
And it is the very same hash: https://github.com/2i2c-org/utoronto-image/commit/d422076c3695e44f09a959b14cb10f89b6e5f538

damianavila:
>Update: The health check failed on CI, but when I ran it locally to inspect the logs, it passed, so it was something transient I guess.

Thanks for your help here, @GeorgianaElena!!

PR #1112, opened 2022-03-15
(Attempt 2) Add logic to the deployer to decide which clusters/hubs require a helm upgrade

sgibson91:
Reconstruction of #1093 

This PR adds a series of functions to the deployer that will analyse a set of filepaths (ideally that have been added/changed by a GitHub PR, this input is not provided in this PR) and decide which hubs/clusters require a helm upgrade and whether that upgrade is for the *hub helm charts or the support helm chart. The output is two lists of dictionaries that each represent a set of matrix jobs to be read by a second GitHub Actions job that will run each dictionary in parallel. One dictionary represents all the support and staging hub upgrade jobs, the other all the prod hub upgrade jobs.

It's logic works like so:

1. Read in a list of added/modified filepaths from a GitHub Pull Request 
  - **If filepaths under `helm-charts/support` change we default to upgrading support on all clusters**
  - **If filepaths belonging to the deployer, our GitHub Actions infrastructure, or the basehub/daskhub helm charts change, we default to upgrading all hubs on all clusters**
2. Get a set of paths to all cluster.yaml files in the repo
3. Begin generating list of matrix jobs to upgrade hub helm charts:
    1. For each cluster.yaml file located in 2, read in the list of hubs
    2. For each hub listed in that cluster.yaml file, get the list of helm chart values files and inspect whether any of these files are present in the original list of added/modified files parsed to the deployer
    3. If the inspection in ii yields True, add a job description for that hub to the list of matrix jobs. Else pass.
    4. If the cluster.yaml file itself has been modified, we skip steps ii and iii and instead generate jobs to upgrade all hubs on the given cluster
4. Begin generating a list of matrix jobs to upgrade support helm charts
	1. For each cluster.yaml file located in 2, verify the `support` config exists in the config. Else pass.
	2. Inspect whether the support helm chart values files listed in the support config are present in the original list of added/modified files parsed to the deployer
	3. If the inspection ii. yields True, add a job description for that cluster to the list of matrix jobs. Else pass.

We then run the generated lists to ensure each job conforms to the following structure:

- `support_and_staging_matrix_jobs` to upgrade the support chart and staging deployment, if required.

  Its structure is:

  ```yaml
  {
    "provider": str                     # The cloud provider the given cluster runs on
    "cluster_name": str                 # The name of the given cluster
    "upgrade_support": str(bool)        # String representation of a Boolean. Whether to upgrade the support chart for this cluster.
    "reason_for_support_redeploy": str  # Why the support chart needs redeploying
    "upgrade_staging": str(bool)        # String representation of a Boolean. Whether to upgrade the staging hub for this cluster.
    "reason_for_staging_redeploy": str  # Why the staging hub needs redeploying
  }
  ```

  **Note:** These must be string representation of Booleans for the GitHub Action runner to parse them correctly

- `prod_hub_matrix_jobs` to upgrade the production hubs.

  Its structure is:

  ```yaml
  {
    "provider": str             # The cloud provider the given cluster runs on
    "cluster_name": str         # The name of the given cluster
    "hub_name": str             # The name of a given prod hub on the given cluster that requires upgrading
    "reason_for_redeploy": str  # Why this prod hub needs redeploying
  }
  ```

5. **If we are running in a GitHub Actions environment as determined by the presence of the `CI` environment variable**, set the matrix job lists as job output variables. We cannot use `GITHUB_ENV` here because
  - Different jobs run on different GitHub Action runners, even if they are defined in the same GitHub Action workflow file
  - Environment variables are written to a file located at `GITHUB_ENV` and this is not shared between jobs since it is local to the runner
  - Hence `GITHUB_ENV` is not available to downstream jobs
6. **If we are not running in a GitHub Actions environment as determined by the absence of the `CI` environment variable or the presence of the `--pretty-print` flag**, generate tables describing all the jobs to be run using `rich`

A `--pretty-print` flag has been provided that will print these jobs in a human-readable format by using tables from the `rich` package. It is intended to run this logic in PRs using the pretty print option so engineers can check which jobs will be run before merging. These tables will include a column stating the reason for redeploy. An example of the output is below.

Screenshot 2022-03-29 at 13 46 36

I have also added some tests to ensure all this new logic is behaving how we expect it to and added a new GitHub Actions workflow file to run these new tests when changes are made to the deployer.

**NOTE:** This PR is not yet implementing the changes in GitHub Actions to actually deploy hubs in parallel. This is only the Python logic that will underpin such a workflow. An example workflow file can be viewed here:

* https://github.com/sgibson91/helm-upgrade-decision-logic/blob/main/.github/workflows/helm-upgrade-decision-logic.yaml

and the corresponding Actions runs are here:

* https://github.com/sgibson91/helm-upgrade-decision-logic/actions/runs/2058557284


deployer/helm_upgrade_decision.py
```diff
@@ -0,0 +1,367 @@
...
+        set[str]: A set of all files matching the pattern "*/cluster.yaml"
+        set[str]: A set of all files matching the pattern "*/*.values.yaml"
+        set[str]: A set of all files matching the pattern "*/*support*.values.yaml"
+    """
+    patterns_to_match = ["*/cluster.yaml", "*/*.values.yaml", "*/*support*.values.yaml"]
+    cluster_filepaths = []
+
+    # Identify cluster paths amongst target paths depending on the files they contain
+    for pattern in patterns_to_match:
+        cluster_filepaths.extend(fnmatch.filter(added_or_modified_files, pattern))
```
sgibson91, line 84: _Original discussion:_ https://github.com/2i2c-org/infrastructure/pull/1093#discussion_r826182520

`patterns_to_matc`h lists three patterns, but the third is matching a subset of the second pattern, making it redundant it seems.

The same patterns show up again later when its relevant to separate the general values file paths from the support file paths. This seems fine, but it breaks DRY principle a bit and could be refactored to something like `support_values_pattern` and `values_pattern` or `non_support_values_pattern`.

deployer/helm_upgrade_decision.py
```diff
@@ -0,0 +1,367 @@
...
+        added_or_modified_files (list[str]): A list of files that have been added or
+            modified in a GitHub Pull Request
+
+    Returns:
+        list[path obj]: A list of unique filepaths to cluster folders
+        set[str]: A set of all files matching the pattern "*/cluster.yaml"
+        set[str]: A set of all files matching the pattern "*/*.values.yaml"
+        set[str]: A set of all files matching the pattern "*/*support*.values.yaml"
+    """
+    patterns_to_match = ["*/cluster.yaml", "*/*.values.yaml", "*/*support*.values.yaml"]
```
sgibson91, line 79: _Original discussion:_ https://github.com/2i2c-org/infrastructure/pull/1093#discussion_r826196267

> With this, we will start relying on (but not enforcing here) the previously non-strict naming convention of having values.yaml in the end of all values files.
> 
> Adding this strict filter pattern makes me think we should enforce `values.yaml` suffix in `helm_chart_values_files`.
> 
> I'm brainstorming an idea that one could not need a filter pattern at all by detecting/inspecting all clusters and inspecting their `helm_chart_values_yaml` files etc for relevant files to observe for changes. Given how this implementation is now though, I suggest its pair with enforcement of the files listed in `helm_chart_values_files` will lead to matching the patterns defined here.

deployer/helm_upgrade_decision.py
```diff
@@ -0,0 +1,367 @@
...
+def discover_modified_common_files(modified_paths: list):
+    """There are certain common files which, if modified, we should upgrade all hubs
+    and/or all clusters appropriately. These common files include the helm charts we
+    deploy, as well as the GitHub Actions and deployer package we use to deploy them.
+    Args:
+        modified_paths (list[str]): The list of files that have been added or modified
+            in a given GitHub Pull Request.
+    Returns:
+        upgrade_all_clusters (bool): Whether or not all clusters should be upgraded
+            since the support chart has changed
```
sgibson91, line 29: _Original discussion:_ https://github.com/2i2c-org/infrastructure/pull/1093#discussion_r826210449

> I'm not confident the logic of upgrading all clusters and helm charts in them makes sense just because we change the support chart we deploy on all/most clusters, but that is probably matching the current behavior we have so 👍 for now!
> 
> There is a comment below:
> 
> ```
>     # If this filepath has changes, we should update the support chart on all clusters
>     support_chart_filepath = "helm-charts/support/*"
> ```
> 
> Hmmm, perhaps the `upgrade_all_clusters` was really `upgrade_support_chart_in_all_clusters` or similar?

deployer/helm_upgrade_decision.py
```diff
@@ -0,0 +1,367 @@
...
+        else:
+            print(f"No support defined for cluster: {cluster_config.get('name', {})}")
+
+    return matrix_jobs
+
+
+def update_github_env(hub_matrix_jobs, support_matrix_jobs):
+    """Update the GITHUB_ENV environment with two new variables describing the matrix
+    jobs that need to be run in order to update the support charts and hubs that have
+    been modified.
```
sgibson91, line 320: _Original discussion:_ https://github.com/2i2c-org/infrastructure/pull/1093#discussion_r826233431

> [consideRatio](https://github.com/consideRatio) [21 hours ago](https://github.com/2i2c-org/infrastructure/pull/1093#discussion_r826233431)
> Can you rephrase this a bit to be clearer on how and why it's done? My understanding is that GITHUB_ENV is a path to a file, which if we write to, influences the environment as observed by a GitHub Workflow's Job's subsequent steps.
> 
> I'm also thinking perhaps we should fail loudly and early if we end up running functions that assume we are doing it as part of a GitHub workflow execution - but aren't.
> 
> From [docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables):
> 
> GITHUB_ACTIONS | Always set to true when GitHub Actions is running the workflow. You can use this variable to differentiate when tests are being run locally or by GitHub Actions.
>  
> Member
> Author
> @[sgibson91](https://github.com/sgibson91) sgibson91 [1 hour ago](https://github.com/2i2c-org/infrastructure/pull/1093#discussion_r827048183) • 
> So the decision is made to print rich tables if EITHER the --pretty-print argument is passed OR GITHUB_ENV is found in the environment:
> 
> [sgibson91/pilot-hubs@5b8d720/deployer/deploy_actions.py#L267-L272](https://github.com/sgibson91/pilot-hubs/blob/5b8d720c39a20fa1b09de4f026f7d172c5fe4eaa/deployer/deploy_actions.py#L267-L272)
> 
> I'm also thinking perhaps we should fail loudly and early if we end up running functions that assume we are doing it as part of a GitHub workflow execution - but aren't.
> 
> So we don't fail, but instead print out human-readable output instead of trying to write to the GITHUB_ENV file

deployer/helm_upgrade_decision.py
sgibson91: @consideRatio For this, I really wish `fnmatch.filter()` would output _two_ lists - one with filenames that match, and the other with ones that don't. Maybe we need to write something more bespoke here? 😕

deployer/helm_upgrade_decision.py
sgibson91: @consideRatio I think I avoided "reading all cluster.yaml files first" as it felt a bit too invasive. Instead, I used the filtering to find unique cluster folders that had additions/modifications and just read those.

deployer/helm_upgrade_decision.py
```diff
@@ -0,0 +1,367 @@
...
+from rich.table import Table
+
+yaml = YAML(typ="safe", pure=True)
+
+# Determine if we are running a test or not. We set this env var to "test" in the
+# pyproject.toml file so it is set when the package is tested using pytest.
+test_env = os.getenv("RUN_ENV", False)
+
+
+def discover_modified_common_files(modified_paths: list):
```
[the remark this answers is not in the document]
sgibson91: @consideRatio I'm not confident I can think of smaller units for this function 😕

deployer/cli.py
```diff
@@ -16,6 +17,13 @@
 )
 
 
+def _convert_string_to_list(full_str: str) -> list:
+    """
+    Take a SPACE-DELIMITED string and split it into a list
+    """
+    return full_str.split(" ")
```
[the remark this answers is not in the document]
consideRatio: Sorry for not making it clear!

Yes i was thinking to either directly reference str.split instead of the new custom function which then is deleted, or to clarify it makes sense to have a dedicated function that does something slightly different (narrowing the split function to only accepts spaces)

consideRatio:
I got thinking that the review notes I wrote related a higher structural level, so with this comment I wanted to brainstorm the concrete implementation to organize state and logic that I believed could help develop and maintain this quite advanced feature.

# Constraints

### Algorithm input

1. What clusters.yaml files we manage in the git repo, and its defined helm chart installations and associated files
2. What files have been modified (and in what way: added/changed/deleted etc)

### Algorithm output

1. For each cluster: what helm chart release that needs to be upgraded, and for what reasons etc.

# Implementation idea

Maybe we could use a Python class to represent relevant info input, calculated output, logic to calculate the output etc. The class should then be scoped to only focus on the _re-deploy or not_ decisions we need to make.

- During class object initialization it should:
  - be passed information about changed files
  - be passed information about where cluster.yaml files are to be located
  - process all cluster.yaml files
  - update object state of things to consider
    - all clusters, their support charts, hubs, and `helm_chart_values_files`
    - all other files of relevance
  - process state to deduce what to do with each cluster / hub and why
- After initialization it should be able to:
 - Emit a log friendly report for an individual cluster
  - Emit a log friendly report for all clusters
  - Emit JSON for use by a github workflow's job's `strategy.matrix` field with relevant information
    I made [this map](https://mermaid.live/edit#pako:eNqVU8FO4zAQ_ZWRr7SVWnrKAWmhe1gOaKVFWqQEVZN42gQc29jjRRXi39cpcdWkCJWcHM9782bmed5EZSSJTGyUea1qdAz314WG-PlQbh3aGgpxE5wjzeDZIdN2l8GKrDI7qFTwTM5Do8GiQ6VITaAMDI4Uoad9xNNLIF2RL8QodZUX4q9xz516IR7H0fXTPL815YncZ8A1dlD4h65Bbowek-BHz_oAr-f5L-05VhzTWGscP8J0epWiiwF2ccDWpNrUG8wHlMsB5fJzymJAWQ4oy3w2m_VFkpZfHk4s-u2MNZ7kuR6hllCHcnB7Yo_MkznjictzrZHfsUZ-aY0cWJP-8lXjLXJVwx289tX6CRDGGzaxvY9MGAeD20ZvL6wzEirTll3wpIRz551U48ST6qGtXgkuYK-VvD9Se0ij5rgA-zEeOu65PSkuxb73MMYdZz5sznHNYiJaci02Mm73W3ddCK6pjegsHiVtMCju6niP0GBlfDU_ZcPGiWyDytNEYGDzZ6crkbELlECrBuMQ2h71_h_K4HB4) to overview the kind of logic we would need to get a functional and efficient CI system based on GitHub Actions. For that map, we would need the following information to be summarized when we emit a `strategy.matrix` JSON blob.
    1. A list of clusters that needs to be at least partially re-deployed
    2. For each cluster: a boolean if the support chart is to be re-deployed
    3. For each cluster: a set of _staging/prod combinations_ in the cluster that should be re-deployed in parallell

__Appreciation for well defined small scoped tasks composed to a larger algorithm__

I have a vague idea that a hub can be concluded to need be re-deployed for multiple reasons, where each reason could be a small scoped check. That would allow us to review pieces of logic in separation and allow us to have nice logs on why we make various decisions on re-deploying or not.

Example reasons/checks for re-deploying a hub in a cluster: "because the deployer script changed", "because a secret feeding into generated config changed", "because of a manual override", "because the following values files changed for the hub", "because the helm chart it depends on changed"
  
# Final thoughts

This to be a complicated feature!!! The last thing I want for my review notes to bog you down @sgibson91, but that is something I fear I may end up doing now. I suggest you see my notes as a source of potential inspiration but otherwise move on with what you think makes sense.

Btw, I'll also need to focus a bit more on z2jh helm chart release + JMTE needs so I may not be as active with review a while now.

deployer/cli.py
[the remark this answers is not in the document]
consideRatio: I think `str.split` is a function on the class `str` that has the same function signature I think.

```
In [1]: str.split("ab cd")
Out[1]: ['ab', 'cd']
```

The comment about `str.split` was just a non-important detail - please proceed in any way!

deployer/helm_upgrade_decision.py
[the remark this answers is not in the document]
consideRatio: Neither could I! Let's resolve this comment without taking any action!

I think perhaps a function with a large scope of work that is tricky to describe could be avoided. Thinking of that made 
me consider using a Class to keep state etc. A class can have a scoped purpose, which means its methods come bundled with a context making them easier to understand, and they also have access to a lot of state in the instance that could help reduce the complexity of a standalone function I think.

I wrote about that in a large standalone comment below.

deployer/helm_upgrade_decision.py
consideRatio: I suggest marking this as resolved without action.

It was related to decision logic, and I think as long as the decision logic made is easy to overview, whatever we do is fine.

deployer/helm_upgrade_decision.py
consideRatio: I figure two strategies are possible and plausible:
a) to rely on assumptions about files' paths and names
b) to rely on explicit configuration

I'm leaning towards thinking b) could be better with ideas like:
- Letting a computer read all cluster.yaml files seems like a suitable chore for a computer
- Looking for specific filename patterns adds logic we didn't really need as we had explicit configuration to identify the relevant files
- Looking for relevant files with new logic based on conventions seems to force us to following them while we previously just recommended them, but forcing isn't enforcing, so now we also need to enforce they are followed so our logic doesn't break.
  - File names
  - Where files are located
     For example I think we could use a file shared by multiple clusters even it wasn't recommended, but now I think all files must be located under the cluster associated directly

deployer/helm_upgrade_decision.py
consideRatio: Now I think of it as the code accomplishes the intentions of the function, so at this point I figure let's wait with adjusting low level tech details like this.

deployer/helm_upgrade_decision.py
consideRatio: This was entirely about code comprehension and can be ignored as well.

The key point of confusion I needed clarify in the docstring was about GITHUB_ENV: a technical environment variable used in GitHub workflows. It includes the path to a file. If that file is appended with environment variables in a `HELLO=VALUE` format for example, they will be set in a subsequent step of the github workflow's job we are currently executing in.

sgibson91:
> For each cluster: a set of staging/prod combinations in the cluster that should be re-deployed in parallell

I actually think this would fail for the 2i2c and cloudbank clusters. Because they only have one staging hub and many unrelated prod hubs. If we have one job that deploys to both _the same_ staging hub and different prod hubs in parallel, then we will run into helm issues of `a helm upgrade is already in process` which will require manual intervention via a `helm rollback`.

Instead, I propose two jobs: job1 to upgrade the staging hub on any given cluster, job2 to upgrade all the prod hubs on any given cluster in parallel. job2 needs job1 to complete successfully.

consideRatio:
> I actually think this would fail for the 2i2c and cloudbank clusters. Because they only have one staging hub and many unrelated prod hubs. If we have one job that deploys to both the same staging hub and different prod hubs in parallel, then we will run into helm issues of a helm upgrade is already in progress which will require manual intervention via a helm rollback.

Okay sigh =/ If i recall correctly, we cannot be fully dynamic, but I think that given some fixed structure, we can populate it. The challenge is understanding what fixed structure meets the needs then.

I want us to aim for the best result, where everything is as parallelized as possible and where everything runs to completion without being blocked something unrelated.

For example: I don't think a staging/prod combo should need to wait for another unrelated staging hub to be deployed or because another unrelated staging hub failed. To accomplish that where each deployment is a separate job in GitHub Workflows we need to rely on some tricks like [considered here, where dispatching workflows](https://mermaid.live/edit#pako:eNqVU8FO4zAQ_ZWRr5RKLZxyWAkoB_awQgKJlRJUTeJpk13HNvYYVCH-HaeNqyZFqOSUzLw3b2Ze5l1URpLIxEqZt6pGx_B4XWiIjw_l2qGtoRA3wTnSDJ4dMq03GSzIKrOBSgXP5Dw0Giw6VIrUBMrA4EgRetpmPL0E0hX5QoxKV3khnoz736kX4nmcXf6b5b9NeST3FXCJHRRe0TXIjdFjElz1rB14OcvvtOfYcSxjrXH8DOfnv1J2PsDO99iaVJtmg9mAcjGgXHxNmQ8olwPKZT6dTvsmSctvX44sunfGGk_yVI9QS6hDOYge2SPzZM544_JUa-RPrJHfWiMH1qSvfNF4i1zV8Afe-m79BAhjhE0cb1cJ42Jw3ej1mXVGQmXasksetXDqvpNq3HhS3Y_VK8EZbLWS9wdqf9OqOR7Ado37iXtuT4pHsZ09jHGHlfeXc9izmIiWXIuNjNf93oULwTW1EZ3FV0krDIq7Pj4iNFgZ_5pb2bBxIluh8jQRGNg8bHQlMnaBEmjRYFxCuwt-fAJbCXAt) are used for each thread of jobs that can run in parallel.

deployer/helm_upgrade_decision.py
```diff
@@ -0,0 +1,368 @@
...
+    and/or all clusters appropriately. These common files include the helm charts we
+    deploy, as well as the GitHub Actions and deployer package we use to deploy them.
+
+    Args:
+        modified_paths (list[str]): The list of files that have been added or modified
+            in a given GitHub Pull Request.
+
+    Returns:
+        upgrade_support_on_all_clusters (bool): Whether or not all clusters should be
+            upgraded since the support chart has changed
```
sgibson91, line 31:
```suggestion
        upgrade_support_on_all_clusters (bool): Whether or not all clusters should have
            their support chart upgraded since has changes
```

deployer/deploy_actions.py
```diff
@@ -208,3 +217,52 @@ def deploy(cluster_name, hub_name, skip_hub_health_test, config_path):
...
+    """Analyse added or modified files from a GitHub Pull Request and decide which
+    clusters and/or hubs require helm upgrades to be performed for their *hub helm
+    charts or the support helm chart.
+
+    Args:
+        changed_filepaths (list[str]): A list of files that have been added or
+            modified by a GitHub Pull Request
+        pretty_print (bool, optional): If True, output a human readable table of jobs
+            to be run using rich. If False, output a list of dictionaries to be
+            passed to a GitHub Actions matrix jobs. Defaults to False.
```
sgibson91, line 232:
```suggestion
            passed to a GitHub Actions matrix job. Defaults to False.
```

deployer/helm_upgrade_decision.py
```diff
@@ -0,0 +1,368 @@
...
+        cluster_info = {
+            "cluster_name": cluster_config.get("name", {}),
+            "provider": cluster_config.get("provider", {}),
+        }
+
+        # Double-check that support is defined for this cluster.
+        support_config = cluster_config.get("support", {})
+        if support_config:
+            if upgrade_support_on_all_clusters:
+                # We know we're upgrading all hubs, so just add the hub name to the list
```
sgibson91, line 288:
```suggestion
                # We know we're upgrading support on all clusters, so just add the cluster name to the list
```

deployer/helm_upgrade_decision.py
```diff
@@ -0,0 +1,368 @@
...
+        else:
+            print(f"No support defined for cluster: {cluster_config.get('name', {})}")
+
+    return matrix_jobs
+
+
+def update_github_env(hub_matrix_jobs, support_matrix_jobs):
+    """Update the GITHUB_ENV environment with two new variables describing the matrix
+    jobs that need to be run in order to update the support charts and hubs that have
+    been modified.
```
sgibson91, line 321:
```suggestion
    been modified.

```

deployer/helm_upgrade_decision.py
sgibson91: My most recent work removes the majority of the file filtering I was doing and instead does the following:

- From the input list of added/modified files, get unique cluster filepaths and verify these are cluster filepaths by checking for the existence of a cluster.yaml file
- For each identified unique cluster filepath, read in the cluster.yaml file and get the list of helm chart values files (whether for a hub or support)
- Inspect whether any of these files pulled from cluster.yaml appear in the original list of added/modified files parsed from the command line and decide whether that hub needs upgrading or support for that cluster needs upgrading based on the results of this inspection

We only read in _all_ cluster.yaml files if core infrastructure has changed, random e.g., the daskhub helm chart

consideRatio:
> > For example: I don't think a staging/prod combo should need to wait for another unrelated staging hub to be deployed or because another unrelated staging hub failed.
> 
> I would like that too, but I also doubt it's possible.

I'm quite confident it is possible, I thought about this a lot when I came up with the mindmap thingy. I think almost all our strategies will have some limitations, but I think that strategy would enable a reasonable experience for us admins at the same time as it would support the full parallelism we want.

Sadly, we can't support any directed acyclic graph (DAG) of ordering of or re-deployments, but we can support a few structures and let some jobs be skipped if not needed. For example, all jobs would fit in:

1. Deploy support chart if needed
1. Deploy shared staging chart if needed
1. Deploy X sets of staging/prod combinations (where staging is optional)

sgibson91:
I would love to see an example GitHub Actions workflow config to ease my mind! I thought we established this wasn't possible because you cannot set job IDs dynamically and matrix jobs share the same job ID?

consideRatio:
@sgibson91 yeah I think this should be piloted without any other complexity to consider to verify it really does work and the UX of dispatching workflows to run is acceptable. I'm really low on time in general... I feel torn between many major items of work, already behind with z2jh release and JMTE staging hub needs etc.

sgibson91:
Ok, well I'll work on what I believe is possible in the current realm of GitHub Actions. I don't disagree with your mindmap per se - I just don't think GitHub Actions is technically capable of running it the way you want it to.

yuvipanda:
The code looks great to me, @sgibson91! Do you want to get this merged and then fiddle with GitHub actions declarations separately, or add that on to this PR?

yuvipanda:
The code looks great to me, @sgibson91! Do you want to get this merged and then fiddle with GitHub actions declarations separately, or add that on to this PR?

sgibson91:
Thanks @yuvipanda I think I will do a few more tweaks now that I know you don't want to run support in parallel. But thank you for the clear path forward it helps a lot with how I can resolve this one!

yuvipanda:
@sgibson91 thank you for your patience, and thank you also for surfacing this as needing broader input!

deployer/helm_upgrade_decision.py
```diff
@@ -0,0 +1,407 @@
...
+
+    Args:
+        added_or_modified_files (list[str]): A list of files that have been added or
+            modified in a GitHub Pull Request
+
+    Returns:
+        list[path obj]: A list of unique filepaths to cluster folders
+    """
+    # Get absolute paths
+    if test_env == "test":
```
yuvipanda, line 83: This introduces a global variable here. Given that this function is being called only from one place, i'd suggest either passing in `is_test` or `exclude_test` as a parameter to this function, or doing the filtering in the calling site itself. Global variables can become very difficult to remove after you add them.

yuvipanda (commented):
Couple of style comments.

The thorough use of inline comments + the detailed description in the PR body was extremely helpful for me. Thank you, @sgibson91

sgibson91:
@yuvipanda I've done some more work on this and updated the top comment. I believe it is now ready for review.

GeorgianaElena (approved):
🎉 🎉 🎉 for the deployer tests, comments, docs and the pretty printing of the jobs.
Awesome work @sgibson91

deployer/helm_upgrade_decision.py
```diff
@@ -10,12 +10,21 @@
 from rich.console import Console
 from rich.table import Table
 from ruamel.yaml import YAML
-from utils import print_colour
+
+# This try/except block is here because pytest wants to import print_colour from
+# deployer.utils, whereas the deployer wants to call it directly from utils. There is no
+# way to fix this for one without breaking it for the other until we make the deployer
+# an actual pip installable package. See the below issue for discussion on this topic:
+# https://github.com/2i2c-org/infrastructure/issues/970
```
yuvipanda, line 18: \o/ great comment!

yuvipanda (approved):
I think this is great progress and good to go! (3a) from https://github.com/2i2c-org/infrastructure/issues/1131 needs to be done but we can do that in a separate PR.

Thank you so much for working on this marathon, @sgibson91! And thanks a lot for all the guidance and review, @consideRatio :)

sgibson91:
Thank you everyone! 🎉

choldgraf:
Just a note that I think the table output of changes to be made is pretty damn cool 😎

PR #1113, opened 2022-03-15
Attempting to fix login issues to uwhackweeks staging hub

sgibson91:
ref: https://2i2c.freshdesk.com/a/tickets/104

- Tighten up auth scopes on uwhackweeks
    - I wasn't sure if the mix of orgs and teams was confusing things
so I restricted us just to teams.
- Regenerate client secrets for uwhackweeks staging github oauth app
  - Since the issue was isolated to staging, not prod, I tried
regenerating the client secret for the github oauth app and
revoked all active user tokens to force people to log back in
again in case that helped the situation.

### DIFF
--- a/config/clusters/uwhackweeks/common.values.yaml
+++ b/config/clusters/uwhackweeks/common.values.yaml
@@ -113,10 +113,9 @@ basehub:
           authenticator_class: github
         GitHubOAuthenticator:
           allowed_organizations:
-            - 2i2c-org
+            - 2i2c-org:tech-team
             - ICESAT-2HackWeek:jupyterhub-2022
           scope:
-            - read:user
             - read:org
         Authenticator:
           # This hub uses GitHub Orgs/Teams auth and so we do not set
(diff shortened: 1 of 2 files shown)


choldgraf:
Probably worth noting that this brings it more in-line with how the pangeo hubs are configured:

https://github.com/2i2c-org/infrastructure/blob/87ee7de3de3903e62e8f196b1dad63aab0c8fb8a/config/clusters/pangeo-hubs/common.values.yaml#L42-L59

sgibson91:
So, because I revoked the tokens and forced the auth flow again Scott noticed that the GitHub OAuth App had not been granted permission to the ICESat org and that is why Authentication was failing. This PR is probably not strictly _required_ but I'll leave it open in case @2i2c-org/tech-team feel it is useful.

choldgraf:
I believe this has been deployed already, yes? If so we should merge it in right?

sgibson91:
@choldgraf yes it has been deployed, but the actual issue ended up being outside our config. So if this was not merged and we reverted to the old config, we don't lose anything.

choldgraf:
Sounds good - I will leave that decision up to others, I don't have strong preferences either way, but I think we should either close this PR or merge it soon :-)

yuvipanda (approved):

sgibson91:
uwhackweeks always fails in ci/cd atm so I will redeploy manually too

PR #1114, opened 2022-03-15
Add notes about authentication with github orgs

choldgraf:
As a follow up to https://github.com/2i2c-org/infrastructure/pull/1113 and https://2i2c.freshdesk.com/a/tickets/104 we discovered that a confusing situation can arise where an organization administrator has not granted our GitHub Org/Teams authenticator app the proper permissions. In this case, they'll get an unauthorized `403` error, and the solution is to grant the auth app access. This adds docs to describe this situation so it's more debuggable later.

### DIFF
--- a/docs/howto/configure/auth-management.md
+++ b/docs/howto/configure/auth-management.md
@@ -1,10 +1,16 @@
 # Manage authentication
 
+This section describes how to set up various authentication providers for the 2i2c JupyterHubs.
+
 ## Auth0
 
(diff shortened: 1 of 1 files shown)


choldgraf:
cc @sgibson91 - maybe you can confirm that my description of the problem is actually correct? :-D

sgibson91 (approved):
This looks good to me

choldgraf:
yay, thanks!

choldgraf:
yay, thanks!

PR #1121, opened 2022-03-15
Add user image repo refs for a few communities

choldgraf:
This adds user image links for a few images that we have recently had to interact with. Just a short-term improvement until we can improve this in a more sustainable fashion in the future.

### DIFF
--- a/config/clusters/uwhackweeks/common.values.yaml
+++ b/config/clusters/uwhackweeks/common.values.yaml
@@ -53,7 +53,7 @@ basehub:
             - name: home
               mountPath: /home/jovyan/shared
               subPath: _shared
-      # User image repo: https://github.com/ICESAT-2HackWeek/jupyter-image-2020
+      # User image repo: https://github.com/ICESAT-2HackWeek/website2022
       image:
         name: quay.io/uwhackweek/icesat2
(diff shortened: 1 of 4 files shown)


config/clusters/uwhackweeks/common.values.yaml
```diff
@@ -53,6 +53,7 @@ basehub:
             - name: home
               mountPath: /home/jovyan/shared
               subPath: _shared
+      # User image repo: https://github.com/ICESAT-2HackWeek/jupyter-image-2020
```
choldgraf, line 56: @scottyhq is this correct? I couldn't find any other user image for the hackweek though this one seems to be updated quite a while ago

scottyhq: we actually have a monorepo going this time around that includes content and image in one place https://github.com/ICESAT-2HackWeek/website2022

choldgraf, line 56:
```suggestion
      # User image repo: https://github.com/ICESAT-2HackWeek/website2022
```

sgibson91 (approved):

GeorgianaElena (approved):
Good idea 🚀

PR #1122, opened 2022-03-15
Simplify incident response template

choldgraf:
We've had some recent incidents, and I found that it was quite noisy to have a huge incident report with "empty" fields waiting to be filled in. During the incident itself, you want to focus effort on fixing the problem itself, and having all the extra "after action report" stuff was making it difficult to focus on this.

So, this PR rearranges our incident template to try and boost the signal-to-noise a little bit. Here are the major changes:

- The after-action report is now hidden under a `details` tab, and is meant to by copy/pasted into the issue after the incident is resolved, so that we can focus on it at the right time.
- Moved the "timeline" into the after action report, with the idea that we link to comments in the issue w/ timestamps, rather than manually noting when various things happened. Also simplified it a bit.

[Here's what the new template will look like when rendered](https://github.com/2i2c-org/infrastructure/blob/f49db9ffbb0548b7c4127cd45468a17ea0ab9ffe/.github/ISSUE_TEMPLATE/3_incident-report.md)

### DIFF
--- a/.github/ISSUE_TEMPLATE/3_incident-report.md
+++ b/.github/ISSUE_TEMPLATE/3_incident-report.md
@@ -6,19 +6,39 @@ labels: ["type: Hub Incident", "support"]
 assignees: ""
 ---
 
-# Summary
+# Context
 
-<!-- 
-Quick summary of the problem. Update this section as we learn more, answering:
+## Summary
 
-- what user impact was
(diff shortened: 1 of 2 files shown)


sgibson91 (approved):

GeorgianaElena (approved):
🚀

PR #1058, opened 2022-03-07
Add a separate nodepool for paleohack event

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/1032

This reverts commit 20969eb31b62b9dd85cebc6080a6e8c516a3a047.

- I've manually scaled to 20 nodes in the GUI
- Enable the pre-puller so nodes get the latest image pre-pulled for faster user starts
- Specify the image in the config here, so the pre-puller works

PR #1123, opened 2022-03-16
Revert "Add a separate nodepool for paleohack event"

yuvipanda:
Reverts 2i2c-org/infrastructure#1058

Event is done!

### DIFF
--- a/config/clusters/2i2c/paleohack2021.values.yaml
+++ b/config/clusters/2i2c/paleohack2021.values.yaml
@@ -25,11 +25,6 @@ jupyterhub:
         funded_by:
           name: "NSF Paleo Perspectives on Climate Change program"
           url: "https://www.nsf.gov/funding/pgm_summ.jsp?pims_id=5750"
-  prePuller:
-    continuous:
-      enabled: true
-    hook:
-      enabled: true
   singleuser:
     memory:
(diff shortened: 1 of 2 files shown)


sgibson91 (approved):
🎉

config/clusters/2i2c/paleohack2021.values.yaml
```diff
@@ -39,9 +34,7 @@ jupyterhub:
       limit: 2
     image:
       name: quay.io/2i2c/paleohack-2021
-      tag: 9d557294938e
-    nodeSelector:
-      2i2c.org/community: paleo
+      tag: 7534858b1098
```
GeorgianaElena, line 37: Does their user image need to change back too?

GeorgianaElena (approved):

config/clusters/2i2c/paleohack2021.values.yaml
sgibson91: I think you are right and this is probably a bug caused by just hitting the "revert PR" button

yuvipanda:
Great catch, @sgibson91 and @GeorgianaElena! Done :)

yuvipanda:
terraform applied!

PR #1026, opened 2022-02-24
Use OPTIMIZE_UTILIZATION cluster autoscaler profile

yuvipanda:
GKE supports setting a cluster autoscaler profile that
optimizes utilization[1]. This changes the behavior of
the autoscaler *and* the scheduler, packing pods in tighter
into nodes rather than 'spreading' them around. This is
what the z2jh user-scheduler primarily does. We can actually
do away with the user-scheduler if we use the OPTIMIZE_UTILIZATION
autoscaler profile - Berkeley has been running under this setup
for a few months now. This reduces how much resources our core
nodes need, and the GKE provided scheduler is also faster.
This will also prevent issues like
https://github.com/2i2c-org/infrastructure/issues/1004.

[1]: https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-autoscaler#autoscaling_profiles

PR #1126, opened 2022-03-16
Disable user scheduler almost everywhere

yuvipanda:
After https://github.com/2i2c-org/infrastructure/pull/1026,
all our GCP hubs use the 'Optimize Utilization' scheduler.
It functions pretty similar to the z2jh custom user scheduler,
and is generally faster & has less overhead. So we want to use
that wherever possible.

In other cloud providers, most of the hubs we have operate on
a one node per user model - a user gets all the resources on a
single node. The only hub that falls outside this is UToronto. this PR
disables user scheduler by default, but enables it on just on UToronto.

utoronto should be explicitly enabled when both these conditions are met:

1. The Cloud provider is *not* GKE
2. The user profiles available pack multiple pods into the same node, rather than
    providing one dedicated node per user.

### DIFF
--- a/config/clusters/pangeo-hubs/common.values.yaml
+++ b/config/clusters/pangeo-hubs/common.values.yaml
@@ -10,6 +10,9 @@ basehub:
       # Name of Google Filestore share
       baseShareName: /homes/
   jupyterhub:
+    scheduler:
+      userScheduler:
+        enabled: true
     proxy:
       https:
         enabled: false
--- a/config/clusters/pangeo-hubs/common.values.yaml
+++ b/config/clusters/pangeo-hubs/common.values.yaml
@@ -10,9 +10,6 @@ basehub:
       # Name of Google Filestore share
       baseShareName: /homes/
   jupyterhub:
-    scheduling:
-      userScheduler:
-        enabled: true
     proxy:
       https:
         enabled: false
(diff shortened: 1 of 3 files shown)


config/clusters/pangeo-hubs/common.values.yaml
```diff
@@ -10,6 +10,9 @@ basehub:
       # Name of Google Filestore share
       baseShareName: /homes/
   jupyterhub:
+    scheduling:
+      userScheduler:
+        enabled: true
```
choldgraf, line 15: doesn't this hub run on GCP? Why do we need the user scheduler given that this hub is (I assume) using the `Optimize Utilization` setup you describe?

choldgraf (approved):
looks good to me - anything that can simplify the setup and push complexity onto the cloud provider instead of our infrastructure sounds like a win to me (provided it is reliable and not super expensive!)

config/clusters/pangeo-hubs/common.values.yaml
yuvipanda: @choldgraf great catch! Fixed.

GeorgianaElena (approved):
🚀  Monday launch 🚀

damianavila (approved):
LGTM

Issue #1105, opened 2022-03-15
Move off manually maintained NFS servers on GCP


yuvipanda:
### Context

On pilot-hubs cluster as well as cloudbank (I think?) user home directories are on a hand-rolled NFS VM. This is problematic - when it fills up, it is a bit complex to resize, and while resizing a PD is fairly standard, I don't think we have it documented.

We should just switch over to paying money for [Google Filestore](https://cloud.google.com/filestore).

This is one of those issues that isn't an issue until it is, and when it is an issue, it's a big issue.

### Next steps

- [x] Figure out which clusters are using hand-rolled NFS
- [x] Modify appropriate cluster tfvars to enable google filestore creation
- [x] Migrate data from current NFS server to google filestore
- [ ] Decomission old NFS VM

yuvipanda:
Only the shared 2i2c cluster on GCP is still using a manual NFS server now. Let's get rid of it before it goes critical and requires someone with XFS knowledge to fix.

yuvipanda:
rsync takes about 44 minutes now.

yuvipanda:
While doing https://github.com/2i2c-org/infrastructure/pull/2672 I discovered that the cloudbank cluster *also* uses a manually setup NFS server! Good catch, and I'll eventually migrate that one too. This will make our cluster design more uniform everywhere.

yuvipanda:
Am running the same copy steps for cloudbank hubs as well now.

yuvipanda:
Everything on the 2i2c shared cluster has been migrated, and I've shut down the original VM! I'll clean it up in a few days if everything is alright.

yuvipanda:
Copy completed on cloudbank for in about 17h. Will run rsync and prep

yuvipanda:
rsync takes about 50m to complete now! I'll just leave it running in a loop, and find an opportune moment to deploy.

yuvipanda:
still waiting for an opportune moment without any skyline users.

yuvipanda:
Cloudbank fully migrated, and the NFS server there has been shut down! I'll wait for a week then kill it.


I've also deleted the disk of the 2i2c shared cluster, although a snapshot remains. I've left the NFS server on still, just in case we need to bring it back up. I'll kill it in a week.

consideRatio:
Wieeee thank you for working this @yuvipanda!!!! I feel a sense of relief with fewer cluster/hub specific exceptions to consider!

[...]

Issue #1102, opened 2022-03-15
Move pilot-hubs cluster to a regional k8s cluster for better availability


yuvipanda:
The pilot-hubs cluster (Running everything under *.pilot.2i2.clcoud) is using a simple, single apiserver zonal k8s cluster. While this is mostly ok, this can mean some downtime and reduced API reliability. It also means no 0 downtime k8s apiserver upgrades, so any upgrades of the apiserver (which happen automatically!) will cause outages.

A regional cluster is highly available, and has 3 apiservers running in a HA configuration. Gives us more reliability. https://cloud.google.com/kubernetes-engine/docs/concepts/regional-clusters has more info.

https://2i2c.freshdesk.com/a/tickets/102 is probably caused by intermittent slowness in k8s apiserver response.

choldgraf:
@yuvipanda +1 from me. Even if it costs marginally more, but increases the reliability and stability of the service, or gives us extra flexibility for upgrades, I think that's preferable.

yuvipanda:
This hit the neurohackademy hub today, as a transient k8s master failure 'borked' the hub. It failed with:

```
[I 2022-07-30 02:24:19.658 JupyterHub log:189] 200 POST /hub/api/users/nicobruno92/activity (nicobruno92@10.0.14.61) 86.49ms
ERROR:asyncio:Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7f8f3d70d6a0>
[E 2022-07-30 02:24:20.351 JupyterHub reflector:351] Watching resources never recovered, giving up
    Traceback (most recent call last):
      File "/usr/local/lib/python3.9/site-packages/kubespawner/reflector.py", line 285, in _watch_and_update
        resource_version = await self._list_and_update()
      File "/usr/local/lib/python3.9/site-packages/kubespawner/reflector.py", line 228, in _list_and_update
        initial_resources_raw = await list_method(**kwargs)
      File "/usr/local/lib/python3.9/site-packages/kubernetes_asyncio/client/api_client.py", line 185, in __call_api
        response_data = await self.request(
      File "/usr/local/lib/python3.9/site-packages/kubernetes_asyncio/client/rest.py", line 193, in GET
        return (await self.request("GET", url,
      File "/usr/local/lib/python3.9/site-packages/kubernetes_asyncio/client/rest.py", line 177, in request
        r = await self.pool_manager.request(**args)
[...]

Issue #989, opened 2022-02-09
Stop using `pilot` in `foo.pilot.2i2c.cloud` for new hubs


choldgraf:
### Description of problem and opportunity to address it

**Context to understand the problem**
Currently, the hubs on our shared cluster follow the pattern `<hubname>.pilot.2i2c.cloud`.

However, we are past our pilot phase, and calling it a `pilot` may cause organizations to make assumptions about the quality or reliability of our hubs. Moreover, `pilot` isn't particularly descriptive, it's more of an implementation details from 2i2c's perspective rather than something that a hub's user cares about.

**Proposed solution**
We should instead use the pattern `<foo>.hub.2i2c.cloud`. This will be:

- more memorable
- a clear indication that there's a hub at the URL
- less tied to the "phase of 2i2c's service"

### Implementation guide and constraints

- We should probably set up a redirect so that `*.2i2c.cloud` redirects to `*.hub.2i2c.cloud`
- One reason we might **not** want to do this is if we explicitly want to have different hubs on different parts of our infrastructure (e.g. `alpha.2i2c.cloud` as well). But thus far I don't think we're doing this.
- As part of this, we should also rename `github.com/2i2c-org/pilot` to `https://github.com/2i2c-org/user-guide` or something like this.

choldgraf:
Could I get some feedback from the @2i2c-org/tech-team on this one? I don't think that making this change will be too difficult and the longer we keep it as `pilot`, the more annoyance there will be if/when we change it.

yuvipanda:
Another option is to just move them to <hub-name>.2i2c.cloud.

choldgraf:
@yuvipanda I'm +1 on that as well! I defer to the preference/judgment of others here, just think that we should move away from the word `pilot`

sgibson91:
No pushback from me about removing `pilot` - I suspect no one has had the time to action this yet :D

consideRatio:
@choldgraf is there any remaining action points besides this for this issue then?

choldgraf:
@consideRatio I think it would be "documenting the new practice of not using `pilot` in our new hub setup docs, to formalize this plan from our team". Does that make sense?

consideRatio:
@choldgraf can I assign you to fix that? I note that you have done a lot of work recently to migrate docs across the 2i2c org

choldgraf:
@consideRatio sure thing 👍 

Ah one question I thought up for redirects. What about:

[...]

PR #1129, opened 2022-03-17
Move a few hubs off *.pilot.2i2c.cloud

yuvipanda:
To signal that this is reasonably stable infrastructure,
we're removing 'pilot' from a few domain names. I've setup
a wildcard domain *.2i2c.cloud to point to the pilot-hubs
cluster's nginx-ingress service IP, so merging this would
just switch out these 3 hubs to get rid of the .pilot
part of their domain.

This does mean our 'primary' cluster becomes a bit more
important, so we should make it a little more resilient.
See https://github.com/2i2c-org/infrastructure/issues/1105
and https://github.com/2i2c-org/infrastructure/issues/1102

Ref https://github.com/2i2c-org/infrastructure/issues/989

### DIFF
--- a/config/clusters/2i2c/cluster.yaml
+++ b/config/clusters/2i2c/cluster.yaml
@@ -12,7 +12,7 @@ support:
 hubs:
   - name: staging
     display_name: "2i2c staging"
-    domain: staging.pilot.2i2c.cloud
+    domain: staging.2i2c.cloud
     helm_chart: basehub
     auth0:
       # connection update? Also ensure the basehub Helm chart is provided a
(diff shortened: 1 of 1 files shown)


yuvipanda:
Let's try check nobody is using the demo hub before switching?

Also, @GeorgianaElena, this might affect your testing!

choldgraf (approved):
Looks good to me - do we need to do anything with the DNS in namecheap?

yuvipanda:
@choldgraf already did the thing with namecheap (setting wildcard DNS - documented in #1091)

PR #1124, opened 2022-03-16
Switch to using pd-balanced for all user & dask nodes

yuvipanda:
We were using standard disk to save costs, but that brings with
it much slower node startup time, as images being pulled take time.
pd-balanced is a newer alternative to pure SSD disks that is
not as expensive, but provides much better performance than pd-standard.

I think the extra cost is worth the performance on all these
cases.

This was based on feedback on new node spinup performance in 
https://github.com/2i2c-org/infrastructure/issues/991#issuecomment-1068644915

Issue #1046, opened 2022-03-02
Transition GCP config-connector to terraform - cleanup `cloudResources`


consideRatio:
## Updated issue

The "config connector" system enabled GKE clusters to define resources of custom types representing GCP resources, where GCP then would create these for us. In practice it enabled [our chart's to emit k8s resources](https://github.com/2i2c-org/infrastructure/blob/HEAD/helm-charts/basehub/templates/cloud-resources/gcp/) that led to the creation of GCP resources outside the k8s cluster, such as a GCP service account and a GCP storage bucket.

We have now transitioned to create such service accounts and buckets via terraform instead, but some clusters still rely on the config connector system. If we can transition to using terraform entirely, we would avoid the complexity of retaining config connector related config that is only relevant to some clusters, and we would avoid the quite significant CPU/Memory overhead requested inside the k8s cluster by the config connector system.

### Action points

From Yuvi's comment https://github.com/2i2c-org/infrastructure/issues/1046#issuecomment-1308202723

* Make sure the cloudResources config is removed from config for:
  * [x]  pangeo-hubs/common
  * [x]  2i2c/dask-staging
  * [x]  2i2c/ohw
  * [x]  2i2c/catalyst-cooperative
* Remove set of resources from our config, under `helm-charts/`
* Remove custom code we have in our `values.yaml` to support this
* Remove terraform code that supported setting up config-connector

## Original issue

<details>

In #1045 I removed this section.

https://github.com/2i2c-org/infrastructure/blob/689db10eee7ac806aa95d0a927953900b98b10fe/helm-charts/daskhub/values.yaml#L1-L6

It wasn't functional and instead silently ignored. It should be located under `basehub.jupyterhub.custom.cloudResources.scratchBucket.enabled` to be functional and not ignored.

In #1045 I opted to not change any behavior but instead remove the misconfiguration that was silently ignored and open this issue to raise awareness about the situation: daskhub won't create a scratch bucket by default etc, and you must enable this for each hub that wants it.

I lean towards thinking that any custom resources that one may want, should be explicit configuration for a cluster's hubs rather than default configuration. If there is agreement about that, there may be no action point at all but to close this issue.

## Action points

Outlined by Yuvi in https://github.com/2i2c-org/infrastructure/issues/1046#issuecomment-1308202723

</details>

sgibson91:
[...]

Issue #669, opened 2021-09-10
Debug why helm chart/config connector doesn't create a new service account


sgibson91:
### Description

While working on https://github.com/2i2c-org/pilot-hubs/pull/662, it came to light that this should already be automated as part of our helm chart by including the appropriate config in the hub config file. However, when I tried to enable this for the Pangeo cluster, the service account (`staging-user-sa@{{ pangeo project ID }}.iam.gserviceaccount.com`) doesn't get created as expected. We need to work out why so we can automate this.

Annoyingly, there's no error message to help debug - I know it's not working because I can see from the Google Cloud IAM page that the service account is not being created.

For now, requester pays access for Pangeo is manually enabled so this isn't a blocking issue.

### Value / benefit

Automating this deployment will save us a lot of manual busywork :)

### Tasks to complete

- [x] Try the deploying the config to a cluster in `two-eye-two-see-sandbox`. This will help us figure out if it's a bug due to the special-casing of Pangeo or if it's a more general broken piece of config
- [ ] Set up this infrastructure with Terraform as a stop-gap solution (rather than using config connector)
- [ ] IDENTIFY AND SQUASH THE BUG! Update this with more details when they're available
- [ ] ...
- [ ] Document how the config snippet that enables requester pays access

### Updates

- **2021-09-10:** I tried deploying this in `two-eye-two-see-sandbox` and the `test-user-sa@two-eye-two-see-sandbox.iam.gserviceaccount.com` service account is not created there either. So this is a general problem.
- **2021-11-24**: Decided to use terraform rather than Config Connector: https://github.com/2i2c-org/infrastructure/issues/669#issuecomment-978039992

sgibson91:
I tried deploying this in two-eye-two-see-sandbox and the test-user-sa@two-eye-two-see-sandbox.iam.gserviceaccount.com service account is not created there either. So this is a **general problem**, not Pangeo-specific.

damianavila:
Looking quickly at the pangeo yaml files:

* https://github.com/2i2c-org/pilot-hubs/blob/master/config/hubs/pangeo-hubs.cluster.yaml
* https://github.com/2i2c-org/pilot-hubs/blob/master/config/hubs/pangeo-181919.cluster.yaml

I do not see `scratchBucket` stuff 
```
scratchBucket:
  enabled: true
```
[...]

PR #1074, opened 2022-03-09
Add LEAP hub

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/1050

PR #1130, opened 2022-03-18
Setup workload identity via terraform

yuvipanda:
The GKE config connector was helpful in letting us deploy
Google Cloud Service Accounts with permissions for cloud storage
directly just from helm. However, it has been difficult to debug,
and in https://github.com/2i2c-org/infrastructure/issues/669
we decided to move away from it and towards creating these
cloud resources via Terraform.

This commit adds:
- Terraform code that will create a Google Service Account,
  bind it to a given Kubernetes Service Account, for a list of
  hub namespaces passed in. This means that some hub initial deployments
  now *can not be done just with CD*, but need manual work with
  terraform. I think this would be any hub that wants to use
  requestor pays or scratch buckets. This would need to be
  documented.
- Move meom-ige to use this new scheme. metadata concealment
  (https://cloud.google.com/kubernetes-engine/docs/how-to/protecting-cluster-metadata#concealment)
  which is what we were using earlier as alternative to config-connector
  + workload identity, is no longer supported by the terraform
  google provider. In b7b42ced82a1b3ad9d6936fa78c5a099ce5004f3,
  we changed the default from 'SECURE' to 'UNSPECIFIED', but
  it looks like 'UNSPECIFIED' really means 'use workload identity'
  haha. When https://github.com/2i2c-org/infrastructure/pull/1124 was
  deployed to meom-ige yesterday, it seems to have enabled workload
  identity, causing cloud access to stop working, leading to
  https://2i2c.freshdesk.com/a/tickets/107. Further investigation on
  what happened here is needed, but I've currently fixed it by
  just deploying this change for meom-ige.
- Documentation is provided for each 'feature', and how 2i2c engineers can
  enable these on a per-hub basis. They are designed to be 'feature flags'
  in terraform wherever possible. This still involves copying some terraform
  outputs to helm values files unfortunately, as we use helm to create namespaces
  (and terraform doesn't have a way to say 'create this namespace if required',
  unlike eksctl)
- Add a 'features/' subdirectory to documentation, that lists specifically just
   hub features, with a corresponding entry in the topic page. Will slowly move
   other documentation about features to this structure.
- Rename current `cloud-access.md` document to clarify it's about auth
  for 2i2c engineers, not about accessing cloud features from the hub
- Remove scratch bucket from the 3 hubs that have it on the 2i2c cluster -
  dask-staging, ohw and catalyst-cooperative. None of these actually have
  used their bucket. requestor_pays is enabled for all of them.

TODO:
- [x] Separate requestor pays & scratch bucket access into feature flags
- [x] Document how to create new hubs that have either of these features
- [x] Figure out how to deal with creating namespace for the secret to live in.
       Perhaps we can move the Kubernetes secret creation away from terraform
       and into helm to simplify this?
- [ ] Apply this to existing pangeo-hubs
- [x] Apply this to hubs that use this right now in 2i2c cluster

Ref https://github.com/2i2c-org/infrastructure/issues/669
Ref https://github.com/2i2c-org/infrastructure/issues/1046


yuvipanda:
This was caught by @sgibson91 in #1074, but then https://2i2c.freshdesk.com/a/tickets/107 happened today as well so I had to fix *that*. Automating this seemed the best way to go forward.

yuvipanda:
Looking at the cloud buckets in https://console.cloud.google.com/storage/browser?forceOnBucketsSortingFiltering=false&project=two-eye-two-see&prefix=&forceOnObjectsSortingFiltering=false, none of them actually have any content in them! So nobody is actually using scratch buckes in the default pilot-hubs cluster.

sgibson91:
Thank you for your work following this up @yuvipanda! :raised_hands:

terraform/gcp/buckets.tf
```diff
@@ -12,10 +12,32 @@ resource "google_storage_bucket" "user_buckets" {
   labels = {}
 }
 
-resource "google_storage_bucket_iam_member" "member" {
+locals {
+  # Nested for loop, thanks to https://www.daveperrett.com/articles/2021/08/19/nested-for-each-with-terraform/
+  bucket_permissions = distinct(flatten([
+    for hub_name, permissions in var.hub_cloud_permissions : [
+      for bucket_name in permissions.bucket_admin_access : {
```
GeorgianaElena, line 19: This is an interesting syntax

GeorgianaElena (commented):
I really love the level of documentation in this PR and I'm very grateful for understanding what's happening here with little knowledge of the terraform syntax. 🌼 Thank you, Yuvi!

helm-charts/basehub/values.yaml
```diff
@@ -1,3 +1,7 @@
+userServiceAccount:
+  enabled: true
```
sgibson91, line 2: Should this default to true? I'm not sure if it's better to explicitly enable features or rely on default config

helm-charts/basehub/values.yaml
```diff
@@ -1,3 +1,10 @@
+# We define a service account that is attached by default to all Jupyter user posd
```
sgibson91, line 1:
```suggestion
# We define a service account that is attached by default to all Jupyter user pods
```

yuvipanda:
pangeo-hubs I've preserved current access, as deploying this there would require #1153 to be done too.

Issue #1128, opened 2022-03-17
[Incident] uwhackweeks hub not spawning due to AWS quota limits


yuvipanda:
### Summary

https://2i2c.freshdesk.com/a/tickets/109 reports that users can't spin up new nodes, even the smallest. This is due to the AWS account running out of quota.

### Impact on users

The hub is unusable for new users.

### Important information

- Hub URL: https://uwhackweeks.2i2c.cloud/
- Support ticket ref: https://2i2c.freshdesk.com/a/tickets/109
- 
### Tasks and updates

- [x] Discuss and address incident, leaving comments below with updates
- [x] Incident has been dealt with or is over
- [ ] Copy/paste the after-action report below and fill in relevant sections
- [ ] Incident title is discoverable and accurate
- [ ] All actionable items in report have linked GitHub Issues

<details>
<summary>After-action report template</summary>

```
# After-action report

These sections should be filled out once we've resolved the incident and know what happened.
They should focus on the knowledge we've gained and any improvements we should take.

## Timeline

_A short list of dates / times and major updates, with links to relevant comments in the issue for more context._

All times in {{ most convenient timezone}}.

- {{ yyyy-mm-dd }} - [Summary of first update](link to comment)
- {{ yyyy-mm-dd }} - [Summary of another update](link to comment)
- {{ yyyy-mm-dd }} - [Summary of final update](link to comment)


## What went wrong

_Things that could have gone better. Ideally these should result in concrete
action items that have GitHub issues created for them and linked to under
Action items._

- Thing one
- Thing two

## Where we got lucky

_These are good things that happened to us but not because we had planned for them._

- Thing one
- Thing two

## Follow-up actions

_Every action item should have a GitHub issue (even a small skeleton of one) attached to it, so these do not get forgotten. These issues don't have to be in `infrastructure/`, they can be in other repositories._

### Process improvements

1. {{ summary }} [link to github issue]
2. {{ summary }} [link to github issue]

### Documentation improvements

1. {{ summary }} [link to github issue]
2. {{ summary }} [link to github issue]

### Technical improvements

1. {{ summary }} [link to github issue]
2. {{ summary }} [link to github issue]
```

</details>

yuvipanda:
https://us-east-1.console.aws.amazon.com/support/home?region=us-east-1#/case/?displayId=9786816411&language=en is the case where I've asked for a bump in the quota.

choldgraf:
[...]

PR #1132, opened 2022-03-18
Temporarily remove large node options from uwhackweeks

yuvipanda:
Our quota is severely constrained - 64 vCPUs only right now!
This increases the number of users who can get on this
temporarily. Using 2xlarge instances helps with improving
number of users who can fit in in our total 64 vCPUs.

Ref https://github.com/2i2c-org/infrastructure/issues/1128


yuvipanda:
I've deployed this, and don't want other deploys to accidentally overwrite this deployed config.

Issue #1103, opened 2022-03-15
Transient issues spinning up servers related to pod reflector errors


sgibson91:
### Context

We often get support tickets along the lines of "can't spin up a user server" that appear to be transient in nature. Upon inspection of the logs, we see `pod reflector errors` which can be caused either by a k8s master API outage, or a race condition in the hub.

I have opened an issue in the jupyterhub/grafana-dashboards repo to ask for the k8s master API stats to be included since this will help us debug these types of issue: https://github.com/jupyterhub/grafana-dashboards/issues/34

Deleting the hub pod and allowing it to be recreated should help things if it was a race condition in the hub too.

Specifically, the pilot-hubs cluster is zonal, not regional, which means it's k8s master API is not highly available and is therefore more prone to these issues. See https://github.com/2i2c-org/infrastructure/issues/1102

### Actions and updates

- [ ] Resolve https://github.com/2i2c-org/infrastructure/issues/1055
- [ ] Check whether this solved our problem (which will probably require just noticing whether this behavior pops up over time)

yuvipanda:
I've bounced (aka restarted) all the hub pods on the pilot-hubs cluster with this line:

```
kubectl get ns | choose 0 | rg -v aup | xargs -L1  kubectl  delete pod -l component=hub -n    
```

[choose](https://github.com/theryangeary/choose) is an awesome alternative to `cut` or `awk`, and rg is [ripgrep](https://github.com/BurntSushi/ripgrep) an awesome alternative to grep. I skipped `aup` because I had already restarted that hub pod, and didn't want to do it again.

yuvipanda:
In this case, I think restarting the hub was the right call - I think we were just seeing https://github.com/jupyterhub/kubespawner/pull/525 again. I see the following line in the log:

```
[C 2022-03-15 03:26:39.979 JupyterHub spawner:2222] Pods reflector failed, halting Hub.
```

But the hub does not halt :) 

I think what happened is:

1. k8s api had some downtime / latency issues
2. hub's reflectors failed multiple times during this, causing kubespawner to decide it needs to shut itself down
3. This triggers https://github.com/jupyterhub/kubespawner/pull/525, leaving the hub in a zombie state.

Restarting the pods fixes this, and upgrading to latest z2jh (which has https://github.com/jupyterhub/kubespawner/pull/525 in it) will fix this for good.

yuvipanda:
I think upgrading z2jh will help fix this.

[...]

PR #1137, opened 2022-03-19
Bump up kubespawner version

yuvipanda:
https://github.com/2i2c-org/infrastructure/issues/1103 is happening
more frequently now, and it's a hard fail - many users just can
not start up their servers. The z2jh upgrade will involve more
work (https://github.com/2i2c-org/infrastructure/issues/1055),
so let's just bump up kubespawner in our custom hub image until then.
    
Fixes https://github.com/2i2c-org/infrastructure/issues/1103

### DIFF
--- a/docs/howto/customize/custom-image.md
+++ b/docs/howto/customize/custom-image.md
@@ -1,5 +1,5 @@
 (customize/custom-image)=
-# Use a custom hub image
+# Use a custom user image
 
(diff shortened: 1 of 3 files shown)


choldgraf (commented):
LGTM but I'm confused why it says the `sops` encryption pre-commit hook is failing 🤔

yuvipanda:
@choldgraf fixed in

sgibson91 (approved):

PR #1140, opened 2022-03-22
Clarify github organizations list

choldgraf:
This clarifies that all of the organizations a user has access to will be listed when they try to authenticate with the GitHub Organization authenticator app. It tries to reduce confusion associated with users seeing a long list of orgs.

cc @scottyhq

### DIFF
--- a/docs/howto/configure/auth-management.md
+++ b/docs/howto/configure/auth-management.md
@@ -187,6 +187,7 @@ Once the OAuth callbacks have been set up following the steps above, somebody wi
 The first time that somebody tries to log on to the hub with this authentication set up, they should be presented with a page that asks them to grant access to various GitHub organizations.
 For each user, GitHub will list _all organizations to which they can control authorization_.
 **They must grant access to the organization that is added to `allowed_organizations` above**, but do not need to grant access to any other organizations.
+In this case, "granting access" only means that the JupyterHub can view whether a user is a member of the GitHub organization.
 
 If administrators report a `403 forbidden` error when they try to log in to the hub, there is a good chance that they did _not grant access_ to the right GitHub organization.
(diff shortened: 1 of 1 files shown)


docs/howto/configure/auth-management.md
```diff
@@ -185,7 +185,8 @@ There are also some steps that a Community Representative will need to take to s
 Once the OAuth callbacks have been set up following the steps above, somebody with administrator permissions for the GitHub organization needs to grant access to the OAuth app that we have created.
 
 The first time that somebody tries to log on to the hub with this authentication set up, they should be presented with a page that asks them to grant access to various GitHub organizations.
-**They must grant access to the organization that is being used to authenticate**.
+For each user, GitHub will list _all organizations to which they can control authorization_.
+**They must grant access to the organization that is added to `allowed_organizations` above**, but do not need to grant access to any other organizations.
```
yuvipanda, line 189: Let's note that 'granting access' just lets us know they belong to the org? We don't get much other access

scottyhq:
I still don't understand why some orgs are automatically confirmed, others have a 'Request' option and others have a 'Grant' option: 
image

docs/howto/configure/auth-management.md
choldgraf, line 189:
```suggestion
**They must grant access to the organization that is added to `allowed_organizations` above**, but do not need to grant access to any other organizations.
In this case, "granting access" only means that the JupyterHub can view whether a user is a member of the GitHub organization.
```

choldgraf: how's that @yuvipanda ?

yuvipanda (approved):
Good to go, although I'd also like to know the answer to @scottyhq's question

sgibson91:
I feel like the only way we'll get an answer is by asking GitHub Auth people 😅

PR #1141, opened 2022-03-23
Disable write access for non-admin users to ~/shared

yuvipanda:
Fixes https://2i2c.freshdesk.com/a/tickets/112

### DIFF
--- a/config/clusters/uwhackweeks/common.values.yaml
+++ b/config/clusters/uwhackweeks/common.values.yaml
@@ -62,7 +62,7 @@ basehub:
           - name: home
             mountPath: /home/jovyan/shared
             subPath: _shared
-            readOnly: false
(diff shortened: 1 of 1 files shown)


yuvipanda:
I've deployed this already as the hub has users on it already. Works well.

Need more docs for our shared directory setup.

damianavila:
This is a nice quick fix, but we should eventually remove this section and leave the base stuff to take over, IMHO: https://github.com/2i2c-org/infrastructure/blob/master/helm-charts/basehub/values.yaml#L182-L186. And maybe that is the part of the pieces we need to document 😉.

yuvipanda:
@damianavila yeah my hope is that this can go under https://github.com/2i2c-org/infrastructure/tree/master/docs/howto/features

PR #1143, opened 2022-03-24
Pin setup-gcloud action to @v0

yuvipanda:
@master is broken, and they explicitly ask you to switch
to using @v0: https://github.com/2i2c-org/infrastructure/runs/5668729514?check_suite_focus=true

https://github.com/google-github-actions/setup-gcloud#-notices
has migration info

### DIFF
--- a/.github/workflows/deploy-grafana-dashboards.yaml
+++ b/.github/workflows/deploy-grafana-dashboards.yaml
@@ -25,7 +25,7 @@ jobs:
           sudo apt install jsonnet
 
       - name: Setup gcloud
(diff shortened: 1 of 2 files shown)


choldgraf (approved):
LGTM

sgibson91:
@yuvipanda do we also want to disable dependabot updates for this action? Or is the problem "just don't use master"?

yuvipanda:
@sgibson91 the action was moving branches from 'master' to 'main' upstream in a few weeks, so they actively explicitly broke master with a useful error message asking people to move to v0. Don't think we need to change anything dependabotwise

Issue #1125, opened 2022-03-16
Implement a check for multiple authentication types in the deployer


sgibson91:
### Background and proposal

**Context**
Most of our hubs are deployed with auth0 apps, and the relevant information is pulled programmatically via the auth0 api and added into our hub config within the deployer. However, sometimes we use JupyterHub's native OAuthenticator in order to provide authentication against GitHub orgs or teams where appropriate.

In https://github.com/2i2c-org/infrastructure/pull/1043, it was discovered that the `uwhackweeks` hubs that were using GitHub authentication, also had the `auth0.enabled` flag set to `true` which would have triggered the deployer to also setup an auth0 app. This passed silently.

**Proposed solution**
We should design and implement a check in the deployer that alerts us via a warning or failure if this configuration occurs again.

PR #1150, opened 2022-03-28
Validate authenticator config before deploying

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/1125

### DIFF
--- a/deployer/config_validation.py
+++ b/deployer/config_validation.py
@@ -154,7 +154,7 @@ def validate_support_config(cluster_name):
         print_colour(f"No support defined for {cluster_name}. Nothing to validate!")
 
 
-def validate_auth0_config(cluster_name, hub_name):
+def assert_single_auth_method_enabled(cluster_name, hub_name):
     """
     For each hub of a specific cluster, it validates that when an authenticator
     other than Auth0 is enabled, then `auth0` is explicitly disabled.
--- a/deployer/config_validation.py
+++ b/deployer/config_validation.py
@@ -156,7 +156,8 @@ def validate_support_config(cluster_name):
 
 def validate_authenticator_config(cluster_name, hub_name):
     """
-    Validates the provided authenticator config for each hub of a specific cluster.
+    For each hub of a specific cluster, it validates that when an authenticator
+    other than Auth0 is enabled, then `auth0` is explicitly disabled.
     """
     _prepare_helm_charts_dependencies_and_schemas()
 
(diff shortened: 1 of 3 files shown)


deployer/config_validation.py
```diff
@@ -152,3 +152,47 @@ def validate_support_config(cluster_name):
                 sys.exit(1)
     else:
         print_colour(f"No support defined for {cluster_name}. Nothing to validate!")
+
+
+def validate_authenticator_config(cluster_name, hub_name):
+    """
+    Validates the provided authenticator config for each hub of a specific cluster.
```
choldgraf, line 159: Could you add a few more comments throughout this function to in plain english what is being checked/validated in each major step?

choldgraf, line 159: It seems like this is primarily being used to check whether `auth0` has been disabled if another authenticator is being used, is that right? If so, I think we could make the function description or name more specific. Though if we intend to expand the kind of validation we do in the future then it might be fine.

choldgraf (commented):
A few quick comments about adding extra comments etc, but it is looking good to me!

GeorgianaElena:
Thanks @choldgraf! I've added some comments and changed the function name.

sgibson91 (commented):
@GeorgianaElena I realise you've already changed the function name once, but I feel this is more of an assert function than validation. What do you think of `assert_single_auth_method_enabled`?

GeorgianaElena:
`assert_single_auth_method_enabled` -> it's just perfect 💯

deployer/deploy_actions.py
```diff
@@ -16,6 +16,7 @@
     validate_cluster_config,
     validate_hub_config,
     validate_support_config,
+    validate_auth0_config,
```
GeorgianaElena, line 19:
```suggestion
    assert_single_auth_method_enabled,
```

deployer/config_validation.py
```diff
@@ -154,7 +154,7 @@ def validate_support_config(cluster_name):
         print_colour(f"No support defined for {cluster_name}. Nothing to validate!")
 
 
-def validate_auth0_config(cluster_name, hub_name):
+def assert_single_auth_method_enabled(cluster_name, hub_name):
     """
     For each hub of a specific cluster, it validates that when an authenticator
     other than Auth0 is enabled, then `auth0` is explicitly disabled.
```
sgibson91, line 160:
```suggestion
    For each hub of a specific cluster, it asserts that only a single auth
    method is enabled. An error is raised when an authenticator
    other than Auth0 is enabled and `auth0` is not explicitly disabled.
```

What do you think?

sgibson91 (approved):
🚀

Issue #818, opened 2021-11-09
Parallelize the deployment of hubs on a cluster


choldgraf:
### Description

We should deploy our hubs in parallel during CI/CD. This would make the deployment of each hub faster and remove unnecessary interdependence between hubs.

### Problem and value in solving it

Deploying in serial has a few shortcomings:

- Hubs at the end of the queue must wait for previous hubs to finish deploying before they are deployed. This is a problem if we really need a hub at the end of a queue to be deployed quickly (like during an event)
- If a hub's deployment fails (e.g., because of #586 ) then all subsequent hubs will *not* be deployed at all, even if they should be.

Parallelizing our hub deploys should make the whole thing faster, and also removes the dependence of one hub's deployment on all of the other hub deployments.

### Design constraints

- **Hubs should deploy in parallel**.
- **Failure for one hub should not stop deployment on other hubs**.
- **A hub should only deploy if configuration relevant to it has been changed**.
- **Use staging hub tests to make sure we are ready to deploy to prod**. While we want to parallelize the deployment of community hubs, we should continue deploying *staging hubs* first and intentionally block deployment for a given hub if the staging hub fails to pass its tests.

### Implementation ideas

- We could investigate chained GitHub actions jobs to define the dependence between a staging deploy + tests, and a subsequent deploy to prod at the hub-level

### Related issues

- https://github.com/2i2c-org/infrastructure/issues/586
- https://github.com/2i2c-org/infrastructure/issues/857
- https://github.com/2i2c-org/infrastructure/issues/279

damianavila:
One thing we have been discussing on Slack that is relevant here is the idea of "maybe selectively deploying hubs only if certain config changes"

@sgibson91 replied
> I think this would be really hard to achieve as it would likely involve reading the YAML in the workflow

I agree with that idea unless we make changes to be granular...

> Wondering if more granularity would help us here, I mean... instead of having a big config file for the cluster, let's have a directory to represent the cluster plus 1) a cluster-specific yaml file and 2) several hub-specific yaml files for each staging/prod pair.

> See the pre-existing discussion we had here: https://github.com/2i2c-org/infrastructure/issues/124

> With more granularity, it would be definitely easier to come up with a proper matrix that selectively deploys on each hub pair accordingly to the specific config we are touching... in addition, it actually helps with our R2R story because the config is actually "isolated" from others if they want to migrate.

And @choldgraf replied

> another benefit would be that you could create a permalink to somebody's hub configuration

[...]

Issue #1131, opened 2022-03-18
Discussion on strategy to parallelize deployments focused on GitHub workflows/jobs


consideRatio:
### Summary

We are trying to decide how we should structure our GitHub Actions workflow to deploy hubs in parallel. This will determine what inputs into the `strategy.matrix` fields will be required and thus will unblock #1112 

- Should the support chart and `cert-manager` be deployed sequentially with the hubs or can it be done in parallel? What are the implications of either scenario?
- We want failed deploys to staging to prevent deploys to prod hubs (canary deploys). Should a failure for staging on _one_ cluster stop prod deploys to _all other_ clusters? If no, what's the simplest way of avoiding this pattern?

---

These are the strategies we consider in GitHub's CI system on how to deploy multiple hubs in parallel etc.

These strategies influences the logic we may want to embed in the deployer script. Due to that, I think now is the time to clarify a strategy to try use in the GitHub CI context.

## Current situation

We have [a GitHub workflow](https://github.com/2i2c-org/infrastructure/blob/master/.github/workflows/deploy-hubs.yaml) with a single job definition running multiple times (`strategy.matrix`), once per cluster. For each cluster, it unconditionally re-deploys everything in sequence.

## Discussion goals

- Clarify strategies we consider
- Reduce the amount of strategies we consider
- Decide to go for a specific strategy

If we manage to reach these goals I think development/review of #1112 will be easier.

## Alternative strategies

In the comments below are strategies I've tried to make explicit to some degree, with a focus on what is done in GitHub's workflow files as that influences what we need from the deployer script.

A key strategy to be aware about is that we can dispatch separate workflows to run in order to run something in parallel. The key downside of doing this is that GitHub's UI don't know we have dispatched a workflow and it will show up decoupled from the workflow that dispatched it to run. To see what I mean with this practically, you can inspect [this workflow execution](https://github.com/consideRatio/testing-gh-workflows/runs/5584159429?check_suite_focus=true) that finished after dispatching more workflows as can be seen [in the list showing all workflows](https://github.com/consideRatio/testing-gh-workflows/actions).

consideRatio:
## Single workflow: pre-planner -> cluster setup -> parallel hub deployment

This strategy would have a single GitHub workflow with a few jobs to execute in sequence, most of them running with multiple variations using `strategy.matrix`.

**Job 1 - The pre-planner job**
[...]

PR #1159, opened 2022-04-01
Parallelise our CI/CD deployment of hubs

sgibson91:
fixes https://github.com/2i2c-org/infrastructure/issues/818

### Summary

This PR implements a first pass of parallelised deployment of hubs in our CI/CD as discussed in https://github.com/2i2c-org/infrastructure/issues/1131 There is still work to be done to make this match the vision in that issue, but I think this is a good start and we can iterate in future PRs.

### What has changed

- I have renamed `.github/actions/deploy` to `.github/actions/setup-deploy`. It now contains all the steps required to be ready to deploy a hub to a cluster, many of which we had originally defined `.github/workflows/deploy-hubs.yaml`, but no longer performs the deploy itself.
- Parallelised jobs have been added to `.github/workflows/deploy-hubs.yaml`.
  - The `generate-jobs` job generates all the matrix jobs required as defined by the deployer logic implemented in https://github.com/2i2c-org/infrastructure/pull/1112. In PRs, this job provides a rich output table of jobs.
  - The `upgrade-support-and-staging` job upgrades the support chart and/or the staging hub on each cluster that require it in parallel. This job requires `generate-jobs` to complete before running.
  - The `upgrade-prod-hubs` job then upgrades every production hub that requires it in parallel. This job requires both `generate-jobs` and `upgrade-support-and-staging` to complete before running.

### Failing test

The message appearing in the failed test is being tracked here: https://github.com/jitterbit/get-changed-files/issues/42

But I actually think this will resolve itself once this PR is merged. I opened a test PR to my test repo via a fork and all worked fine: https://github.com/sgibson91/helm-upgrade-decision-logic/pull/6


.github/workflows/deploy-hubs.yaml
```diff
@@ -6,121 +6,131 @@ on:
...
+  pull_request:
+    branches:
+      - master
+    paths:
+      - deployer/**
+      - requirements.txt
+      - .github/workflows/deploy-hubs.yaml
+      - .github/actions/setup-deploy/**
+      - "**.yaml"
+  workflow_dispatch:
```
sgibson91, line 24:
```suggestion
```

.github/actions/setup-deploy/action.yaml
```diff
@@ -0,0 +1,58 @@
...
+      shell: bash
+
+    - name: Setup gcloud auth for docker
+      # FIXME: Add more auth providers & registries here as needed
+      run: |
+        gcloud auth configure-docker us-central1-docker.pkg.dev
+      shell: bash
+
+    - name: Setup helm
+      uses: azure/setup-helm@v2.0
```
consideRatio, line 34: Note that `helm` is bundled in the github environment by default now, they may lag behind a patch version or two, but stay up to date in general.

Not having it here could perhaps save a few seconds and some PRs about bumping this via dependabot etc.

.github/workflows/deploy-hubs.yaml
```diff
@@ -6,121 +6,130 @@ on:
...
+          python deployer deploy ${{ matrix.jobs.cluster_name }} staging
+
+      - name: Upgrade dask-staging hub on cluster ${{ matrix.jobs.cluster_name }}
+        if: matrix.jobs.upgrade_staging == 'true' && matrix.jobs.cluster_name == '2i2c'
+        run: |
+          python deployer deploy ${{ matrix.jobs.cluster_name }} dask-staging
+
+  # TODO: Add in "job 3a" where the matrix for prod jobs is edited to exclude any clusters
+  # that failed the preceding upgrade-support-and-staging job. See discussion in:
+  # https://github.com/2i2c-org/infrastructure/issues/1131
```
consideRatio, line 111: Note that I sadly think this may require hardcoded outputs for each cluster - something we were able to avoid above.

Well, hmmm... Maybe one could upload/download artifacts to communicate between jobs, but that is the last hope I have for a workaround about this.

.github/workflows/deploy-hubs.yaml
```diff
@@ -6,121 +6,130 @@ on:
...
+    steps:
+      - name: Setup deploy for ${{ matrix.jobs.cluster_name }}
+        uses: ./.github/actions/setup-deploy
         with:
-          cluster: ${{ matrix.cluster_name }}
-        env:
-          TERM: xterm
+          provider: ${{ matrix.provider }}
+
+      - name: Upgrade  and test ${{ matrix.jobs.hub_name }} on cluster ${{ matrix.jobs.cluster_name }}
```
consideRatio, line 133:
```suggestion
      - name: Upgrade and test ${{ matrix.jobs.hub_name }} on cluster ${{ matrix.jobs.cluster_name }}
```

.github/actions/setup-deploy/action.yaml
consideRatio: Yepp!

Note though that the Python versions are available ahead of time and setup-python makes the chosen version installed in <1 second by changing symlinks, so its a choice of the system-installed version - unless a specific patch version of python is chosen that wasnt available.

sgibson91:
I don't think the prettier pre-commit failure is related to what I've changed here, but it's really hard to tell from the logs. There's no spaces between the filenames and so it's just one giant line - very difficult to parse 😕

.github/workflows/deploy-hubs.yaml
```diff
@@ -6,121 +6,129 @@ on:
...
       - requirements.txt
-      - dev-requirements.txt
-      - config/secrets.yaml
-      - config/clusters/**
       - .github/workflows/deploy-hubs.yaml
-      - .github/actions/deploy/**
+      - .github/actions/setup-deploy/**
+      - "**.yaml"
+    tags:
+      - "**"
```
sgibson91, line 14:
```suggestion
```

.github/workflows/deploy-hubs.yaml
```diff
@@ -6,121 +6,127 @@ on:
...
-              - "helm-charts/**"
-              - "requirements.txt"
-              - "config/secrets.yaml"
-              - ".github/workflows/deploy-hubs.yaml"
-              - ".github/actions/deploy/*"
-
-      - name: Check which cluster directory has changes (if any)
-        uses: dorny/paths-filter@v2
-        id: config_files
+          fetch-depth: 0
```
GeorgianaElena, line 39: The docs say that `fetch-depth: 0` is to fetch all history for all branches and tags. But I don't think I understand why we need this. Was this the default behavior in v2?

.github/workflows/deploy-hubs.yaml
```diff
@@ -6,121 +6,127 @@ on:
...
+          python deployer deploy ${{ matrix.jobs.cluster_name }} dask-staging
+
+  # TODO: Add in "job 3a" where the matrix for prod jobs is edited to exclude any clusters
+  # that failed the preceding upgrade-support-and-staging job. See discussion in:
+  # https://github.com/2i2c-org/infrastructure/issues/1131
+
+  upgrade-prod-hubs:
+    runs-on: ubuntu-latest
+    needs: [generate-jobs, upgrade-support-and-staging]
+    # Only run this job on pushes to the default branch and when the job output is not
```
GeorgianaElena, line 113: I think this is referring to the `generate-jobs` job?
```suggestion
    # Only run this job on pushes to the default branch and when the `generate-jobs` job output is not
```

.github/workflows/deploy-hubs.yaml
```diff
@@ -6,121 +6,127 @@ on:
...
-          (steps.config_files.outputs.hub_config == 'true')
-        uses: ./.github/actions/deploy
+          python deployer deploy-support ${{ matrix.jobs.cluster_name }}
+
+      - name: Upgrade staging hub on cluster ${{ matrix.jobs.cluster_name }}
+        if: matrix.jobs.upgrade_staging == 'true'
+        run: |
+          python deployer deploy ${{ matrix.jobs.cluster_name }} staging
+
+      - name: Upgrade dask-staging hub on cluster ${{ matrix.jobs.cluster_name }}
```
GeorgianaElena, line 101: I think we should make it clear this is just a special case?

```suggestion
      - name: Upgrade dask-staging hub on cluster ${{ matrix.jobs.cluster_name }} if it exists
```

GeorgianaElena (commented):
This work that has been done around parallelizing hub deployment is just awesome 🚀 ❤️ 
Thanks @sgibson91!
I left a few questions, but I believe we should merge this and see how it goes. However, I believe that others on the @2i2c-org/tech-team should should also have a look before since it's a pretty big change.

.github/workflows/deploy-hubs.yaml
sgibson91: Ah, that change is related to this issue with the jitterbit/get-changed-files action https://github.com/jitterbit/get-changed-files/issues/24#issuecomment-1047442168

sgibson91, line 39:
```suggestion
          # Fetch-depth is required for the jitterbit/get-changed-files action
          # https://github.com/jitterbit/get-changed-files/issues/24#issuecomment-1047442168
          fetch-depth: 0
```

GeorgianaElena: Thanks @sgibson91 🌼 It makes sense now

consideRatio (approved):
Nice work @sgibson91 this LGTM!!!

I've added a few comments focused on documentation that I think are quite important for maintenance sustainability but not technically critical - from the technical standpoint is approved as it is!

sgibson91:
Note: we may need to address https://github.com/2i2c-org/infrastructure/issues/1171 first

.github/actions/setup-deploy/action.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+    - name: Get gcloud info
+      run: gcloud info
+      shell: bash
+
+    - name: Setup gcloud auth for docker
+      # FIXME: Add more auth providers & registries here as needed
+      run: |
+        gcloud auth configure-docker us-central1-docker.pkg.dev
+      shell: bash
```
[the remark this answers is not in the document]
sgibson91: Perhaps @yuvipanda can answer this question?

.github/workflows/deploy-hubs.yaml
```diff
@@ -6,121 +6,129 @@ on:
...
-      - name: Setup sops
-        if: |
-          (steps.base_files.outputs.files == 'true') ||
-          (steps.config_files.outputs.hub_config == 'true')
-        uses: mdgreenwald/mozilla-sops-action@v1
+          format: space-delimited
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Generate matrix jobs
+        if: github.event_name != 'pull_request'
```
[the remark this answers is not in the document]
sgibson91: It is serving a purpose, we run a different last step in this job if triggered by a PR.

When triggered by a PR, we pretty-print the matrix jobs by running this step:

https://github.com/2i2c-org/infrastructure/pull/1159/files#diff-b04e1f867bb3bf0588d7d7e8f3337f0206eebc2337a56cc569d3146a4a2edc0fR65-R71

When triggered by a push to the default branch, the matrix jobs are sent to the next jobs via `::set-output`:

https://github.com/2i2c-org/infrastructure/pull/1159/files#diff-b04e1f867bb3bf0588d7d7e8f3337f0206eebc2337a56cc569d3146a4a2edc0fR61-R63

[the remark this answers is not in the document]
consideRatio: Ah, the "pretty print matrix jobs" step below also has `if: github.event_name != 'pull_request'` rather than `if: github.event_name == 'pull_request'`.

I think it can be nice to have a pretty printed summary alongside doing or not doing `::set-output` no matter what btw. If easy to adjust, I suggest to make us do `::set-output` + `--pretty-print` at all times and then like we have it now - we ignore any outputs if we have a pull_request event later.

.github/workflows/deploy-hubs.yaml
```diff
@@ -6,121 +6,129 @@ on:
...
-
-      - name: Setup sops
-        if: |
-          (steps.base_files.outputs.files == 'true') ||
-          (steps.config_files.outputs.hub_config == 'true')
-        uses: mdgreenwald/mozilla-sops-action@v1
+          format: space-delimited
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Generate matrix jobs
```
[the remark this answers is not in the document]
sgibson91: Does [874fd0d](https://github.com/2i2c-org/infrastructure/pull/1159/commits/874fd0d188c88d64027ce9cdc8bb69ffd997601f) also address this?

.github/workflows/deploy-hubs.yaml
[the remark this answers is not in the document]
sgibson91: The following commits should have addressed this:

- [f7dadad](https://github.com/2i2c-org/infrastructure/pull/1159/commits/f7dadadf3868b6ecb969f62026663edf99ed8b7b)
- [4028548](https://github.com/2i2c-org/infrastructure/pull/1159/commits/4028548b4c49205dbaa3ed8587cfa0e190cf6949)
- [c388e6e](https://github.com/2i2c-org/infrastructure/pull/1159/commits/c388e6e984f79fa1b9e9c4e3e0fc78a66dc0e3b0)

.github/actions/setup-deploy/action.yaml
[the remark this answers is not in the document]
consideRatio: This is the commit, and it hasn't been changed since that: https://github.com/2i2c-org/infrastructure/commit/7aff26ea76777af0ef8aa9aef08c9fede824b9ba

I assume that we had previously built images and such before, but now we don't as part of deploy.

The image built seem to have been the custom 2i2c image, which is now available under quay.io publicly without need for authentication: https://github.com/2i2c-org/infrastructure/pull/925/files

I think we are safe to remove this section.

sgibson91:
@consideRatio I propose we start our co-working tomorrow by hitting merge on this, fix any bugs that happen to arise, and then decide where to go with the rest of our time from there

consideRatio:
Wieee! London collaboration merge!

PR #1162, opened 2022-04-02
Fix grafana documentation code

choldgraf:
In trying to debug the Carbon Plan outage today I noticed that the Grafana URL was missing from our "list of running hubs" page. This PR makes a few tweaks to that code so that we now correctly pick up the Grafana URL for carbonplan, Pangeo, and other hubs with multiple config files.


GeorgianaElena (approved):
🚀

damianavila (approved):

PR #1164, opened 2022-04-03
Bump user memory guarantee to 2GB

### LINKED ISSUE #1163: [U Toronto] Increase user memory guarantee to 2GB
### Context

The University of Toronto hub originally had 2GB of RAM for each user, and when we moved it onto our shared infrastructure, this got changed to 1GB instead of 2GB (because that's our default). They have reports from several users that their sessions are getting killed due to OOM, and their community representative requested that we bump it up to 2GB per user.

### Proposal

We should increase the requested memory per user on the Toronto hub to be 2GB.

I think this'll need to be a change somewhere around here:

https://github.com/2i2c-org/infrastructure/blob/master/config/clusters/utoronto/common.values.yaml

### Updates and actions

_No response_

choldgraf:
closes #1163


yuvipanda (approved):
We can fiddle with overcommit ratios later, as there's a conversation being had with the rest of the university around budgets.

PR #1176, opened 2022-04-09
Add tutorial on setting up your local environment

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/1167

Issue #1167, opened 2022-04-03
Major documentation structure for helping our team debug issues


choldgraf:
### Context

In our latest incident on CarbonPlan @yuvipanda and I had some conversation about ways to lower the barrier for people to debug our cloud infrastructure when incidents occur. We agreed it would be helpful to have some basic documentation to help our team members get started with debugging common things.

### Proposal

We've got an outline of major areas of documentation to write to make it easier for people to use our docs in the debugging/operations process. Here's that document:

https://hackmd.io/omFosDsjS3-2UEtAIZstcA

Here are the documents outlined there that we should create

- [x] https://github.com/2i2c-org/infrastructure/pull/1176
- [x] https://github.com/2i2c-org/infrastructure/issues/1891
- [x] https://github.com/2i2c-org/infrastructure/issues/1144
- [x] https://github.com/2i2c-org/infrastructure/issues/1890
- [x] https://github.com/2i2c-org/infrastructure/issues/1893
- [x] https://github.com/2i2c-org/infrastructure/issues/1892
- [x] https://github.com/2i2c-org/infrastructure/issues/1892
- [ ] Howto: Enable shared directory
- [x] https://github.com/2i2c-org/infrastructure/issues/996#issuecomment-1308246131
- [ ] https://github.com/2i2c-org/infrastructure/issues/2272

We should flesh out the major sections in that HackMD, and update this issue as we do so with refs to PRs that implement things. We can close the issue once we've got docs that cover the major parts of that HackMD.

### Updates and actions

- [x] https://github.com/2i2c-org/infrastructure/pull/1177
- [x] https://github.com/2i2c-org/infrastructure/pull/1176
- [x] #1826

yuvipanda:
A more structured way to think about this is in terms of the 'objects' we have and the 'actions' that can be performed on them.

A starter pack would be:

| Object | Verbs |
| - | - |
| Cloud account | Provision, authenticate, view web console, etc |
| Kubernetes cluster | Authenticate, explore objects in `kubectl`, look at logs, etc |
| Hub | View full config of, change config of, create, decomission, deploy change to, look at logs for, find out community rep, etc |
| User image | Create repo, setup image, debug changes, etc |

A lot of these have detailed documentation and trianing provided elsewhere (`kubectl` for example), so we should strike a balance between writing our own custom docs, linking to upstream docs, and improving upstream docs.

yuvipanda:
[...]

PR #1177, opened 2022-04-09
Reorganizing how-to section

choldgraf:
As a complement to https://github.com/2i2c-org/infrastructure/pull/1176 and https://github.com/2i2c-org/infrastructure/issues/1167, @yuvipanda and I discussed that one challenge with the How-To section is that the top-level sections are very generic and hard to semantically distinguish (e.g. what is the difference between "configure" and "customize"?)

So, this PR re-organizes the how-to section with the following goals:

- Do not change the content, just rearrange the pages and `toctree` objects + some page titles
- The top-level sections should be more specific
- Sub-sections should be naturally-grouped based on the top-level sections

Hopefully this will make it easier to quickly find information, and also more obvious when / how to add new information to these howto sections.

Here's a pic of the new `howto` section:

image

Old version for comparison:

image


yuvipanda (approved):

damianavila (approved):
LGTM, it actually makes things more discoverable, IMHO.

choldgraf:
thanks all :-)

PR #1183, opened 2022-04-12
ci: fix syntax error in workflow related to an if statement

consideRatio:


consideRatio:
@sgibson91 approves :)

image

PR #1184, opened 2022-04-12
Transform input strings into absolute Posix Paths

sgibson91:
Ensures that comparisons are performed on the same object type


sgibson91:
@consideRatio verbally approved this. My phone is too far away for a joke :P

Issue #1024, opened 2022-02-23
Generate token for Health Check service in the z2jh helm chart instead of in the deployer


sgibson91:
### Background and proposal

**Context**

We currently have two functions in the deployer that generates helm values on-the-fly that get parsed to the `helm upgrade` command:

- [`get_generated_config`](https://github.com/2i2c-org/infrastructure/blob/aa580cb0c28f4ef6599b6ba1a1a3fae9b1ac33f2/deployer/hub.py#L348-L466)
- [`apply_hub_helm_chart_fixes`](https://github.com/2i2c-org/infrastructure/blob/aa580cb0c28f4ef6599b6ba1a1a3fae9b1ac33f2/deployer/hub.py#L479-L518)

In `apply_hub_helm_chart_fixes`, [an API token is generated with admin privileges](https://github.com/2i2c-org/infrastructure/blob/aa580cb0c28f4ef6599b6ba1a1a3fae9b1ac33f2/deployer/hub.py#L492-L499). The deployer script later [uses this API token in the testing phase](https://github.com/2i2c-org/infrastructure/blob/aa580cb0c28f4ef6599b6ba1a1a3fae9b1ac33f2/deployer/hub.py#L600-L627). It uses the token to create a singleuser server on which some notebooks are executed top-to-bottom to ensure the hub is working at a basic level.

**Problem statement:** The more config that is not explicitly stated in a helm chart values file, and is instead generated on-the-fly in the deployer's "black box", the more difficult our deployment process becomes for our engineers to understand, and the more difficult a deployment is to replicate for a customer choosing to leave 2i2c and take their hub with them.

**Proposed solution**

The z2jh helm chart has the ability to automatically generate tokens (if they do not already exist) for registered services and store them as kubernetes secrets. The deployer, using service account credentials, will have permission to read kubernetes secrets in order to retrieve the token and execute its tests as normal. This would reduce the amount of on-the-fly-generated config produced by the deployer's "black box" and explicitly define more config in the form of helm chart values, or direct configuration in the basehub helm chart. Hence improving the comprehensibility of our deployment process and simplifying the amount of tasks the deployer carries out.

consideRatio:
@sgibson91 :100: :+1: this would make me very happy to see.

[...]

PR #1185, opened 2022-04-12
Separate hub health check into it's own function/cli subcommand

sgibson91:
### Motivation

Previously, the hub health check ran immediately after a helm upgrade of a hub.

Given that our tests are often flakey since they trigger a spin up of a new node, we would like to retry our tests and give the cluster time to provision the node before the tests fail.

We cannot retry tests in this way easily if we are locked into doing a helm upgrade.

We therefore separate running the hub health check from the deploy function into its own function and subcommand.

### What has changed

A new deployer subcommand called `run-hub-health-check` which takes a cluster name and hub name as arguments and runs our health checks against them.

To retain the previously behaviour, users can run:

```bash
python deployer deploy CLUSTER_NAME HUB_NAME && python deployer run-hub-health-check CLUSTER_NAME HUB_NAME
```

The GitHub Actions workflow file has been updated to run the health check in a new step.

We read the service API token required to interact with the JupyterHub API directly from the Kubernetes secret it is stored in, instead of needing to propagate the state of on-the-fly-generated config across functions. Related discussion to this point in https://github.com/2i2c-org/infrastructure/issues/1024. We have not changed how this API token is generated when we deploy a new/existing hub though.

### DIFF
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -114,10 +114,20 @@ jobs:
         run: |
           python deployer deploy ${{ matrix.jobs.cluster_name }} staging
 
+      - name: Run health check for staging hub on cluster ${{ matrix.jobs.cluster_name }}
+        if: matrix.jobs.upgrade_staging == 'true'
+        run: |
+          python deployer run-hub-health-check ${{ matrix.jobs.cluster_name }} staging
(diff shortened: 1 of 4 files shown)


damianavila:
Nice idea!!

consideRatio (approved):

PR #1190, opened 2022-04-13
Split tests of dask compute from dask worker scaling

sgibson91:
### Motivation

In our previous test dask notebook, we made sure that dask could be imported and the arrays could be used, then we ensured a dask worker can be scaled up using dask-gateway.

However, scaling up a dask worker cluster was one source of of flakiness in our hub health check since we sometimes would have to wait for a node to scale up. It was also quite wasteful in terms of time and resources to spin up a cluster every time we run our hub health check in CI.

This PR therefore splits those two aspects of checking daskhub capabilities into two notebooks and we provide functionality to be able to run the scaling check locally by providing a flag to the `run-hub-health-check` deployer subcommand.

### What has changed

- There are now two notebooks under `deployer/tests/test-notebooks/daskhub`: one that tests dask compute, and another that tests dask-gateway scaling
- The `--check-dask-scaling` flag to the `run-hub-health-check` subcommand can be provided to run the scaling tests, but by default this won't run, particularly in CI

### Issues discovered by not solved by this PR

We use `jhub-client` to execute local notebooks on a JupyterHub singleuser server. The `execute_notebook` function creates a new server each time it is called and attempts to clean up after itself (we also attempt to clean up too by calling `delete_server` and `delete_user`. However, when a delete server request is sent to the JupyterHub API, it can produce two response codes:

- 202: server is taking a while to stop
- 204: server is stopped

`jhub-client` does not check which response code is returned, just that _something_ is returned. This means when running multiple notebooks back-to-back, we run the risk of failing because a server for the deployment-health-service already exists.

We have introduced a pause in the script to give the JupyterHub some extra time to resolve itself as a temporary workaround. An upstream issue describing the problem is here: https://github.com/Quansight/jhub-client/issues/14


deployer/tests/test-notebooks/daskhub/dask_test_notebook.ipynb
```diff
@@ -1,121 +1,63 @@
...
+            "execution_count": 6,
+            "metadata": {},
+            "outputs": [],
+            "source": [
+                "import dask.array as da"
+            ]
+        },
+        {
+            "cell_type": "code",
+            "execution_count": 7,
```
sgibson91, line 14:
```suggestion
            "execution_count": null,
```

deployer/tests/test-notebooks/daskhub/dask_test_notebook.ipynb
```diff
@@ -1,121 +1,63 @@
...
+            "metadata": {},
+            "outputs": [],
+            "source": [
+                "a = da.ones(shape=(1000, 1000), chunks=(500, 500))"
+            ]
+        },
+        {
+            "cell_type": "code",
+            "execution_count": 8,
```
sgibson91, line 23:
```suggestion
            "execution_count": null,
```

sgibson91:
@consideRatio verbally approved this for merge after the cycle meeting

PR #1192, opened 2022-04-13
ci: enable support/staging failures to not block other clusters prod deploy

### LINKED ISSUE #1131: Discussion on strategy to parallelize deployments focused on GitHub workflows/jobs

consideRatio:
This makes us skip deploying prod hubs after a staging fails. Note that i think it also fixes a bug where prod deployment would continue even if staging failed as we had an if conditional that disregarded that.

This implementation is the job3a from https://github.com/2i2c-org/infrastructure/issues/1131#issuecomment-1072573736. Constraints that led to the hardcoding you see of cluster names is reported already to https://github.community/t/bug-jobs-output-should-return-a-list-for-a-matrix-job/128626/32?u=consideratio.

Closes #1131.

---

This code is not tested in this context, so there are very possibly small tweaks needed following merge.

### DIFF
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -87,7 +87,29 @@ jobs:
   # upgrade_staging for each cluster that requires it.
   upgrade-support-and-staging:
     runs-on: ubuntu-latest
-    needs: generate-jobs
+    needs: [generate-jobs]
+
+    # We declare outputs indicating the job failed status of a specific job
+    # variation. We are currently required to do this in a hardcoded fashion,
+    # see this post for feature requests for this to be improved:
+    # https://github.community/t/bug-jobs-output-should-return-a-list-for-a-matrix-job/128626/32?u=consideratio
+    #
+    # IMPORTANT: names can include alphanumerics, '-', and '_', but not '.', so
+    #            we replace '.' for '-' in cluster names.
+    #
+    outputs:
+      failure_2i2c: ${{ steps.declare-failure-status.outputs.failure_2i2c }}
+      failure_azure-carbonplan: ${{ steps.declare-failure-status.outputs.failure_azure-carbonplan }}
+      failure_carbonplan: ${{ steps.declare-failure-status.outputs.failure_carbonplan }}
+      failure_cloudbank: ${{ steps.declare-failure-status.outputs.failure_cloudbank }}
+      failure_farallon: ${{ steps.declare-failure-status.outputs.failure_farallon }}
+      failure_leap: ${{ steps.declare-failure-status.outputs.failure_leap }}
+      failure_meom-ige: ${{ steps.declare-failure-status.outputs.failure_meom-ige }}
+      failure_openscapes: ${{ steps.declare-failure-status.outputs.failure_openscapes }}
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-hubs.yaml
```diff
@@ -144,28 +166,74 @@ jobs:
...
+              for prod_job in prod_hub_matrix_jobs
+              if outputs[f"failure_{prod_job['cluster_name']}"] != "true"
+          ]
+
+          print(f"::set-output name=filtered-prod-hub-matrix-jobs::{json.dumps(filtered_prod_hub_matrix_jobs)}")
+
+  # This job upgrades production hubs on clusters in parallel, if required. This
+  # job needs both the `filter-generate-jobs` to have completed to provide its
+  # output `filtered-prob-hub-matrix-jobs`. It is a list of dictionaries with
+  # the keys cluster_name, provider, and hub+name for each production hub that
```
sgibson91, line 221:
```suggestion
  # the keys cluster_name, provider, and hub_name for each production hub that
```

typo

sgibson91 (approved):
LGTM!

consideRatio:
Thanks @sgibson91!!

consideRatio:
Omg I just realized I dare merge without a big stress building up - the health check etc is more reliable now!!!

PR #1193, opened 2022-04-14
Use minSize property for scaling eksctl nodes

yuvipanda:
We were unnecessarily introducing a new minNodes property
when we could just use the eksctl native minSize property


damianavila (approved):
LGTM, although I think we should have a follow-up PR with a different on-the-fly approach that does not kill live pods (or alternatively, figure out why this approach is actually killing them as experienced by @yuvipanda on Slack: https://2i2c.slack.com/archives/C028WU9PFBN/p1649940996067329?thread_ts=1649691374.360779&cid=C028WU9PFBN).

PR #1196, opened 2022-04-14
Remove unnecessary paths from workflow triggers

sgibson91:
Logic to setup clusters for a deploy is no longer contained within `.github/workflows/deploy-hubs.yaml` so I believe it is no longer useful to trigger deploys based on changes to that file


sgibson91:
@consideRatio I requested your review but this may be something you'd like to merge after you've debugged/iterated on the job 3a work

consideRatio (approved):

consideRatio:
Nice!

PR #1212, opened 2022-04-15
Cache pip's install cache between jobs in a workflow run, stop re-install of gcloud, pin helm's version

### LINKED ISSUE #1203: ci: consider optimizing workflow execution speed by caching pip deps _within the workflow_ specifically
I've ended up with too much debugging hurdles to want to recommend caching python dependencies for optimized speed in workflows - if - the cache isn't cleared very regularly.

In this case, we have a LOT of jobs within the same workflow, starting out with the workflow from the first `generate-jobs` job. @sgibson91 suggested we cache the installed deployer script pip deps there and then re-use them for all later jobs in the workflows - which are a lot.

Like this, we will always run new workflows with fresh dependencies and avoid complexity of outdated caches while at the same time avoiding re-installing the pip dependencies in very many jobs to follow.

consideRatio:
We run an [action run ~50 times on each deploy workflow execution](https://github.com/2i2c-org/infrastructure/actions/runs/2170081750), so the value of optimizing this is can avoid wasting compute/energy and our time waiting for it.

Closes #1203

## How does actions/cache action work?

- Actions can have a "post job" hook. I think an action that failed to restore cache already saved from another job will use the "post job" hook to upload files instead for others to use. This means, that saving of cache isn't done when the action is called during a step in a job, but at job completion - but restoring of cache will be done when the cache action is called if a cache is found.
- I assume the cache is stored someplace with GitHub's object storage.
- The configured `key` helps decide if a cache should be used, and if so, what version of the cache. If a key is the same all the time, it will re-use a cache forever unless cleared for other reasons. But, we don't want to re-use a cache nonstop, so we specify a key that is unique for the workflow "run" - meaning for each time it is originally triggered (but a cache is reused between re-attempts of the same workflow).

## What did I do and why?

- EDIT: I had written that python dependencies were cached, but it's really only the pip's install cache that is itself cached. This will still speed up execution as installing from pip's install cache is faster than re-building wheels from source, but it's not as good performance improvement as I originally thought we could accomplish.

  I setup caching of python dependencies installed for the deployer script, which is relevant in two locations: .github/workflows/deploy-hubs.yaml and .github/actions/setup-deploy/action.yaml
- I've reordered steps in .github/actions/setup-deploy/action.yaml to delay exposing credentials more directly
- EDIT: I've added back `helm` (re-)install with the efficient `azure/setup-helm` action using a github cache mechanism internally and pinned the `helm` version.

  I've removed the (re-)install of `gcloud` and `helm`, which shouldn't affect us as its already [available from the GitHub's virtual environment](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md). I motivate this with two pro-points and a con-point:
  - con: we can't pin `gcloud` / `helm` any more, but I'm not sure its a net positive to do that. We had `gcloud` pinned, but doesn't have it pinned any more with this PR.
  - pro: we reduce risk of issues with a github action being malicious in these workflows
  - pro: we maybe save up to a few seconds by removing this per job execution
  
## Further optimizations

- We could have a) setup-python b) cache deps c) install deps run in a dedicated action, like that we could avoid duplicating some code in both generate-jobs job in the deploy workflow and in our setup-deploy/action.yaml action used by the other deploy workflow jobs.
- We could cache kops, but its not needed for `generate-jobs` job and only used in a few situations, therefore its a bit trickier to cache its installation. Also, the impact is lower as we only use `kops` once or twice or so anyhow. Further we move away from it, so I suggest we don't do this.

### DIFF
--- a/.github/actions/setup-deploy/action.yaml
+++ b/.github/actions/setup-deploy/action.yaml
@@ -75,7 +75,20 @@ runs:
       with:
         python-version: "3.9"
 
-    - name: Setup deployer script Python dependencies
+    - name: Attempt to restore cached Python dependencies
+      uses: actions/cache@v3
+      id: cache-pip
+      with:
+        path: ~/.cache/pip
+        # key determines if we define or re-use an existing cache or not, and we
+        # define a string composed of parts to ensure we only cache within a
+        # workflow run attempt, not between workflow runs or separate attempts
+        # of the runs.
+        key: "${{ github.run_id }}-${{ github.run_attempt }}"
+
+    - name: Install deployer script's Python dependencies
+      if: steps.cache-pip.outputs.cache-hit != 'true'
       run: |
-        python -m pip install -r requirements.txt
+        pip install -r requirements.txt
+        pip list
       shell: bash
--- a/.github/actions/setup-deploy/action.yaml
+++ b/.github/actions/setup-deploy/action.yaml
@@ -17,9 +17,10 @@
 #
 name: "Setup the deployer script for use to deploy"
 description: >-
-  Setups the deployer script by loading credentials and installing relevant tools
-  needed to interact with encrypted files, kubernetes clusters, and container
-  registries.
+  Setups the deployer script by loading credentials and installing library
+  dependencies and relevant tools needed to interact with encrypted files,
+  kubernetes clusters, and container registries. Tools already available in the
+  github virtual environments are not re-installed (gcloud, helm).
 
 # inputs configuration reference:
 # https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#inputs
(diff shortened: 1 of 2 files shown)


yuvipanda (changes requested):
This is super damn pretty cool!

However, I'd like us to pin helm - helm upgrades across releases can leave things in an inconsistent state, and I don't want that changing without explicit action on our part. gcloud doesn't matter as much though.

consideRatio:
Thanks @yuvipanda for the review! I've addressed your comment and fixed this to have a chance of functioning by correcting a misunderstanding. See ff82a50 about that.

yuvipanda (approved):
Thanks @consideRatio! Let's see how it goes :)

consideRatio:
Success! The duration dropped from min-max of 47-82 seconds to a min-max of 17-25 seconds for the setup-deploy/action.yaml file, and generate-jobs just invested 2 seconds extra to upload the pip cache of 11 MB!

yuvipanda:
Thank you, @consideRatio! I'm always inspired by your commitment to pushing through long term gains.

PR #1217, opened 2022-04-19
DOC: Fix action link

choldgraf:


choldgraf:
If the linkcheck job passes, I'll merge this one in since it's just fixing a broken link

Issue #1216, opened 2022-04-18
Protect cloudbank demo hub from crypto miners


yuvipanda:
### Context

Saturday Apr 16 we got an email from GCP saying they detected crypto mining in our cloudbank hub. I investigated and made notes in https://hackmd.io/3_UBXSW3T4Om3wv_buyZ-g. I kicked out the miner after gathering evidence. Evidence gathering process as well as evidence itself is in that hackmd.

Currently, cloudbank demo hub is set to allow logins via username / password managed via auth0. This is open to anyone. Let's lock it down to only allowing institutional logincs via CILogon.

### Proposal

- [ ] Switch off from Auth0 to CILogon
- [ ] Disable all known 'open account creation' providers in CILogon - Microsoft, Google, GitHub, ORCID
- [ ] Bring the hub back online

PR #1218, opened 2022-04-19
Restrict cloudbank demo hub to edu & 2i2c users

yuvipanda:
- Move cloudbank hub from username / password auth0 auth to
  CILogon
- Restrcit login to folks with emails either of 2i2c.org or any
  email ending in .edu. This doesn't catch *all* educational users,
  but I think it should catch everyone that cloudbank tries to reach.
- Switches @ericvd-ucb's admin email to use his berkeley.edu email,
  as gmail.com emails are no longer allowed

Fixes https://github.com/2i2c-org/infrastructure/issues/1216

### DIFF
--- a/config/clusters/cloudbank/cluster.yaml
+++ b/config/clusters/cloudbank/cluster.yaml
@@ -95,16 +95,13 @@ hubs:
     domain: demo.cloudbank.2i2c.cloud
     helm_chart: basehub
     auth0:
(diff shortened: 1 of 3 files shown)


yuvipanda:
I've deployed this and it works ok.

damianavila (approved):
LGTM

PR #1228, opened 2022-04-22
Fix the datasource updater

GeorgianaElena:
Spotted a few errors in this script when updating the datasources again


yuvipanda (approved):
ty

Issue #1233, opened 2022-04-25
Bug in deploy-hubs.yaml causing upgrade-prod-hubs job to wait for pending jobs that won't arrive


sgibson91:
## Current understanding

- `upgrade-support-and-staging` job is allowed to fail with the next jobs in the workflow allowed to execute
- However, we need to use `always()` to ensure the next jobs _do_ actually execute and aren't skipped due to some unknown magic happening in GitHub Actions
- Using `always()` alone meant that the `upgrade-prod-hubs` job would be left in a "Waiting for pending jobs" state, even when no jobs were coming because the `prod-hubs-matrix` was an empty list.
- We add a job status check of `filter-generate-jobs` to the conditional of `upgrade-prod-hubs` to ensure this job _only runs if `filter-generate-jobs` was successful` and the rest of the condition is also respected too.
- The overall status of the workflow will always be failed if any of the `upgrade-support-and-staging` matrix jobs fail - even if the rest of the jobs successfully complete.

### Context

We have a bug, probably in the `if` statement for the `upgrade-prod-hubs` job, causing that job to register as failed instead of skipped. See screenshot below.

<img width="1900" alt="Screenshot 2022-04-25 at 13 29 43" src="https://user-images.githubusercontent.com/44771837/165089432-2b2f5a01-8377-4a96-bde4-01580b6e30c9.png">

The output of `generate-jobs` for this run was nothing, so I'm certain the job should have been skipped.

<img width="714" alt="Screenshot 2022-04-25 at 13 32 13" src="https://user-images.githubusercontent.com/44771837/165089612-2350b149-b567-4738-b72f-ae2a2ebe9800.png">

Run in question: https://github.com/2i2c-org/infrastructure/actions/runs/2219349118

`if` block in question: https://github.com/2i2c-org/infrastructure/blob/82c550c818924930b9b17550e3d9b309979d8742/.github/workflows/deploy-hubs.yaml#L209-L212

sgibson91:
My instinct is maybe the `always()` is being evaluated first?

choldgraf:
I think it is pretty cool that this nifty output table makes it easier to debug these kinds of things.

sgibson91:
I did some experimenting over here: https://github.com/sgibson91/testing-gh-actions/blob/main/.github/workflows/chained-failure.yaml

[...]

PR #1234, opened 2022-04-25
Check filter-generate-jobs job completed successfully before trying to run upgrade-prod-hubs job

sgibson91:
Adds a statement to check that the `filter-generate-jobs` job completed successfully before trying to execute `upgrade-prod-hubs` job. This should prevent the job always trying to run, even with invalid input.

fixes https://github.com/2i2c-org/infrastructure/issues/1233

### DIFF
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -248,7 +248,11 @@ jobs:
     # Only run this job on pushes to the default branch and when the `generate-jobs` job output is not
     # an empty list
     #
+    # always() is added as it seems to be a magic function to ensure a job can
+    # run at all without being skipped if a previous job has had any failures.
+    #
     if: |
+      always() &&
       (github.event_name == 'push' && contains(github.ref, 'master')) &&
       needs.filter-generate-jobs.outputs.filtered-prod-hub-matrix-jobs != '[]'
     strategy:
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -248,11 +248,7 @@ jobs:
     # Only run this job on pushes to the default branch and when the `generate-jobs` job output is not
     # an empty list
     #
-    # always() is added as it seems to be a magic function to ensure a job can
-    # run at all without being skipped if a previous job has had any failures.
-    #
     if: |
-      always() &&
       (github.event_name == 'push' && contains(github.ref, 'master')) &&
       needs.filter-generate-jobs.outputs.filtered-prod-hub-matrix-jobs != '[]'
     strategy:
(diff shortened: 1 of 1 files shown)


consideRatio:
@sgibson91 yikes `always()` is messy and badly documented.

I think though that by not having it here, a failure in one of the several staging jobs will make the deploy to prod hub job be skipped due to previous failures.

1. We need `always()` because other kinds of status check, which kind of does the same thing, doesn't make the job run if there is a previous failure - because of undocumented magic.
2. We want to not run if `filter-generate-jobs` errored still though, so we may want to add a explicit status check about the outcome of that job specifically if needed, or make our check for outputs be smarter than `` != `[]` `` or adding a default value to our output so that we don't end up with a blank string making this become truthy when the output is entirely unset.

If you want to accomplish `(i) filter-generate-jobs has successfully completed (satisfied by the needs block),`

sgibson91:
> I think though that by not having it here, a failure in one of the several staging jobs will make the deploy to prod hub job be skipped due to previous failures.

I think this is not true because `upgrade-staging-and-support` is not listed in `upgrade-prod-hubs`'s `needs` block. If it *were* then I believe your assessment is true. But as it is, I think it only cares about the status of `filter-generate-jobs` which has retained it's `always()` statement

consideRatio:
### Related

- About always(): https://docs.github.com/en/actions/learn-github-actions/expressions#always
- About more explicit checks alongside `always()` etc.: https://docs.github.com/en/actions/learn-github-actions/expressions#evaluate-status-explicitly

.github/workflows/deploy-hubs.yaml
```diff
@@ -248,11 +248,7 @@ jobs:
     # Only run this job on pushes to the default branch and when the `generate-jobs` job output is not
     # an empty list
     #
-    # always() is added as it seems to be a magic function to ensure a job can
-    # run at all without being skipped if a previous job has had any failures.
-    #
     if: |
```
consideRatio, line 251:
```suggestion
    # always() is added as it seems to be a magic function to ensure a job can
    # run at all without being skipped if a previous job has had any failures. By
    # using always() we need to ensure filter-generate-jobs wasn't skipped or
    # failed explicitly now though.
    if: |
      always() && needs.filter-generate-jobs.result == 'success' &&
```

sgibson91:
Ugh, yeah this is not tidy at all.

I tested over here: https://github.com/sgibson91/testing-gh-actions/blob/main/.github/workflows/chained-failure.yaml

It also seems that the overall status will always be failed even though the workflow completed right through to the end. Which I don't really like, but I guess we need to know about support/staging failures somehow.

Screenshot 2022-04-26 at 13 41 33

sgibson91:
This is much more sensible!

Screenshot 2022-04-26 at 14 21 41

consideRatio:
@sgibson91 thanks for following up and fixing this!

Issue #1146, opened 2022-03-24
[LEAP Hub] Allow for profile list options based on GitHub team membership


choldgraf:
### Background and proposal

The LEAP Hub has different sub-communities within its user community. Some of them require access to more complex infrastructure than others, and they'd like to offer different environments to some users over others.

Specifically, they'd like this breakdown of **github team membership** and **machine types** (ref: https://github.com/2i2c-org/infrastructure/issues/1050#issuecomment-1077938916):

| tier | privileges |
|--|--|
| Public tier `leap-stc/leap-pangeo-users` | Access to "Small" and "Medium" machine types |
| Research tier `leap-stc/leap-pangeo-research` | Access to all machine types plus GPU option | 

So from a user's perspective, the end-result they would like is something like:

- A user logs in to the hub, authenticating with their GitHub handle
- At the profile list page, they are presented with a different list of options based on their GitHub team membership
- They can then select an option like normal and proceed.

### Implementation guide and constraints

@consideRatio provided a nice description of the solution in [this Discourse comment](https://discourse.jupyter.org/t/tailoring-spawn-options-and-server-configuration-to-certain-users/8449#problem-1-adjust-presented-server-options-based-on-user-2). 

Here's the major parts of this approach:

> # Problem 1 - Adjust presented server options based on user
> What if you want to present different server options to different users, based on some property of the user? Just setting `c.KubeSpawner.profile_list` / `singleuser.profileList` won't do because then all users would get the same options.
> 
> The solution to this is to update `c.KubeSpawner.options_form` with a custom function that first updates the `the_users_spawner_instance.profile_list` based on logic that involve the user information.
> 
> #### Common section for examples below
> The examples below assume we can request a "scope" called "gpu_access", which is supposed to return a "claim" with the same name as a true or false value.
> 
> ```python
> # This snipped assumes a OAuthenticator based Authenticator, and the scopes will
> # vary and this is just an example.
> #
> # - openid is common for OpenID Connect based authentication
> # - profile, email are both commonly available scopes to request
> # - gpu_access is entirely custom
> #
> c.OAuthenticator.scopes = ["openid", "profile", "email", "gpu_access"]
> ```
> 
> #### Solution to problem 1
> To present JupyterHub server spawn options based on the user
> 
> ```python
> # To adjust the spawn options presented to the user, we must create a custom
> # options_form function, and this example demonstrates how!
> #
> #
> #
> # profile_list (KubeSpawner class) can be configured as a convenience to
[...]

Issue #1050, opened 2022-03-02
[New Hub] LEAP Pangeo


choldgraf:
### Hub Description

LEAP Pangeo is an extension of the Pangeo project to new communities around research and education with Machine Learning. The hub's environment will be nearly identical to the Pangeo Hubs, and run on GKE, though the setup might be slightly different and we should get clarifications from @rabernat.

- This website has a lot of useful information about the project: https://leap-stc.github.io/leap-pangeo/architecture.html
- Lead ref: https://github.com/2i2c-org/leads/issues/63

### Community Representative(s)

@rabernat

Not sure if there are others serving as leads on the project.

### Important dates

- **Required start date**: March 14th
- **Target start date**: ASAP - they would like to get this running whenever we can get the hub set up
- **Any important dates for usage**: Not that I know of

### Hub Authentication Type

Other (may not be possible, please specify in comments)

### Hub logo information

**TODO**: @rabernat does this look correct?

- **URL to Hub Image**: https://leap-stc.github.io/_static/LEAP_logo.png
- **URL for Image Link**: https://leap-stc.github.io/index.html

### Hub user image

**TODO**: @rabernat can you advise here? Is this the Pangeo user image?

- Repository for user image: { REPO LINK IF IT EXISTS }
- User image registry: { REGISTRY IF ONE ALREADY EXISTS }
- User image tag and name: { NAME AND TAG IF IT EXISTS }

### Extra features you'd like to enable

**TODO**: @rabernat does it need to be in a specific data center?

- [ ] Specific cloud provider or datacenter: 
- [x] Dedicated Kubernetes cluster
- [x] Scalable Dask Cluster
- [x] GPUs available to users

### Other relevant information

**There is a GCP billing account with credits for this hub**. It is under the 2i2c.org GCP organization. Here are the details:

- Name: community-LEAP-NSF
- ID: 01A164-923D17-3199D9

### Hub URL

leap.pangeo.2i2c.cloud

### Hub Type

daskhub

### Tasks to deploy the hub

- [x] Engineer who will deploy the hub is assigned
- [x] Deploy information filled in above
- [x] Initial Hub deployment PR: #1074
- [x] Administrators able to log on
- [ ] Community Representative satisfied with hub environment
- [ ] Hub now in steady-state

### Follow-up issues

- https://github.com/2i2c-org/infrastructure/issues/1146
- https://github.com/2i2c-org/infrastructure/issues/1083
[...]

PR #1239, opened 2022-04-25
Restrict access to profiles based on GH team membership

yuvipanda:
profile_list is now dynamically generated, based on the GH teams
user is a part of. This list of teams is refreshed only during login -
so user needs to log out and log back in to see new teams! This also
means that users removed from teams on GH will still have access to
the profiles until they are logged out from the admin panel too (to
be fixed)

This approach is taken over customizing options_form to protect
against users just bypassing the options form and using the API
directly to spawn servers.

Deployed to the leap hub, except 'large' & 'huge' is only available to
leap-stc:leap-pangeo-research members, not to leap-stc:leap-pangeo-users
members - based on https://github.com/2i2c-org/infrastructure/issues/1050#issuecomment-1077938916

Brings in https://github.com/jupyterhub/oauthenticator/pull/498,
we can update to a newer oauthenticator once that is merged.

Fixes https://github.com/2i2c-org/infrastructure/issues/1146

TODO:
- [ ] Carve out an exception for deployment-service-check user
- [x] Document how to enable this feature

### DIFF
--- a/config/clusters/leap/common.values.yaml
+++ b/config/clusters/leap/common.values.yaml
@@ -36,6 +36,7 @@ basehub:
       allowNamedServers: true
       config:
         Authenticator:
+          enable_auth_state: true
           # This hub uses GitHub Teams auth and so we don't set
           # allowed_users in order to not deny access to valid members of
           # the listed teams. These people should have admin access though.
@@ -44,6 +45,7 @@ basehub:
         JupyterHub:
           authenticator_class: github
         GitHubOAuthenticator:
+          populate_teams_in_auth_state: true
           allowed_organizations:
             - leap-stc:leap-pangeo-users
             - 2i2c-org:tech-team
@@ -67,6 +69,9 @@ basehub:
         - display_name: "Small"
           description: 5GB RAM, 2 CPUs
           default: true
+          allowed_teams:
+            - leap-stc:leap-pangeo-users
+            - 2i2c-org:tech-team
           kubespawner_override:
             mem_limit: 7G
             mem_guarantee: 4.5G
@@ -74,6 +79,9 @@ basehub:
               node.kubernetes.io/instance-type: n1-standard-2
         - display_name: Medium
           description: 11GB RAM, 4 CPUs
+          allowed_teams:
+            - leap-stc:leap-pangeo-users
+            - 2i2c-org:tech-team
           kubespawner_override:
             mem_limit: 15G
             mem_guarantee: 11G
@@ -81,6 +89,9 @@ basehub:
               node.kubernetes.io/instance-type: n1-standard-4
         - display_name: Large
           description: 24GB RAM, 8 CPUs
+          allowed_teams:
+            - leap-stc:leap-pangeo-research
+            - 2i2c-org:tech-team
           kubespawner_override:
             mem_limit: 30G
             mem_guarantee: 24G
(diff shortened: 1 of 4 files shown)


config/clusters/leap/common.values.yaml
```diff
@@ -67,27 +69,39 @@ basehub:
...
+            - 2i2c-org:tech-team
           kubespawner_override:
             mem_limit: 15G
             mem_guarantee: 11G
             node_selector:
               node.kubernetes.io/instance-type: n1-standard-4
         - display_name: Large
           description: 24GB RAM, 8 CPUs
+          allowed_teams:
+            - leap-stc:leap-pangeo-research
```
rabernat, line 93:
```suggestion
            - leap-stc:leap-pangeo-education
            - leap-stc:leap-pangeo-research
```

rabernat (commented):
This is amazing!

yuvipanda:
I think this is ready to go!

sgibson91 (approved):
Nice!

docs/howto/configure/auth-management.md
```diff
@@ -193,6 +193,77 @@ If administrators report a `403 forbidden` error when they try to log in to the
...
+In addition, we can allow people access to specific profiles based on their GitHub Teams membership!
+This only works if the hub is already set to allow people only from certain GitHub organizations
+to log in.
+
+The key `allowed_teams` can be set for any profile definition, with a list of GitHub
+teams (formatted as `<github-org>:<team-name>`) that will get access to that profile. Users
+need to be a member of any one of the listed teams for access. The list of teams a user
+is part of is fetched at login time - so if the user is added to a GitHub team, they need
+to log out and log back in to the JupyterHub (not necessarily to GitHub!) to see the new
+profiles they have access to. To remove access to a profile from a user, they have to be
```
choldgraf, line 211: We should document this in docs.2i2c.org so that others know this is possible. I think we should also flag this as a `BETA` feature that may change at any moment (e.g. if we want to generalize this to non-github, or if we want to change the behavior around things like the need to delete users)

helm-charts/basehub/values.yaml
```diff
@@ -441,3 +441,74 @@ jupyterhub:
...
+            if spawner.authenticator.populate_teams_in_auth_state == False:
+              return original_profile_list
+
+            auth_state = yield spawner.user.get_auth_state()
+
+            if not auth_state or "teams" not in auth_state:
+              if spawner.user.name == 'deployment-service-check':
+                # For our hub deployer health checker, ignore all this logic
+                print("Ignoring allowed_teams check for deployment-service-check")
+                return original_profile_list
```
sgibson91, line 477: Now that we own https://github.com/deployment-service-check/ maybe we should add that account to our GitHub org and tech team and remove this logic? I wonder if we can close the security backdoor by treating `deployment-service-check` as an actual member of our team?

yuvipanda: @sgibson91 we couldo that if we can make that user authenticate using the oauth flow - although last i looked that was pretty difficult :(

docs/howto/configure/auth-management.md
yuvipanda: @choldgraf opened https://github.com/2i2c-org/infrastructure/issues/1245 to track.

helm-charts/basehub/values.yaml
sgibson91: Ah okay

helm-charts/images/hub/Dockerfile
```diff
@@ -18,6 +18,9 @@ RUN pip install --no-cache git+https://github.com/yuvipanda/jupyterhub-configura
 # z2jh 1.2.x ships with kubespawner 1.1.0, so we just do a little bump
 RUN pip install --no-cache --upgrade jupyterhub-kubespawner==1.1.2
 
+# Brings in https://github.com/jupyterhub/oauthenticator/pull/498
```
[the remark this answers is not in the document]
yuvipanda: Done! https://github.com/2i2c-org/infrastructure/issues/1252

Issue #1237, opened 2022-04-25
GPU option for LEAP and M2LInES Hubs


rabernat:
### Context

The LEAP and M2LInES Hubs are for machine learning. Machine learning libraries work great on GPUs. We have a lot of credits to spend. So we would to have an option to attach a GPU to the user's server at startup. I don't have strong feelings on which GPU. Not the most expensive or fancy. Any GPU will be much better than the CPU.

The `ml-image` from https://github.com/pangeo-data/pangeo-docker-images/ should contain GPU-enabled ML libraries such as tensorflow.

GPUs should always be optional and not on by default. 

### Proposal

I propose to modify the profile list for both hubs to add a GPU option. This is linked closely to #1146. For LEAP, we would want to restrict the option only to the research tier.

yuvipanda:
We have quota, but I asked for a bump anyway.

However, we're in us-central1-b and womp womp, GPUs are only available in us-central1-a and us-central1-c lololol https://cloud.google.com/compute/docs/gpus/gpu-regions-zones

Given nobody is using these yet, let's move this right now - otherwise it'll be a massive PITA

yuvipanda:
The quota request was denied, so we need to get some more usage up before we can request more. I think m2lines has 8 GPUs and leap has 1.

yuvipanda:
I'm redeploying the m2lines cluster now as it hasn't had any usage.

yuvipanda:
@rabernat should GPU be available for all authenticated users in m2lines, or should it be restricted to a certain team as is on LEAP?

yuvipanda:
With #1244, this is done!

I heard from @rabernat that we don't need any team-based customization for m2lines!

I opened https://github.com/2i2c-org/infrastructure/issues/1250 to deal with documentation, so closing this \o/

sgibson91:
We were already tracking docs related to GPUs in https://github.com/2i2c-org/infrastructure/issues/996 so I closed #1250 as a duplicate :)

PR #1244, opened 2022-04-26
Enable GPUs for m2lines hub

yuvipanda:
- Move entire cluster + filestore from us-central1-b to us-central1-c,
  as us-central1-b does not have any GPUs! https://cloud.google.com/compute/docs/gpus/gpu-regions-zones
  has list of zones with GPUs. This was a full redeployment, but our
  terraform is now good enough that this went without a hitch. No data
  migration was needed thankfully.
- Allow specifying GPU info in terraform variables, so we can have GPU
  nodepools
- Install stable nvidia GPU driver from
  https://cloud.google.com/kubernetes-engine/docs/how-to/gpus#installing_drivers,
  in a similar manner to how we install driver for Azure.
- Set PATH and LD_LIBRARY_PATH for the ml-notebook, as that is needed to
  get GPU drivers to work. This should ideally be in the upstream pangeo
  ml-notebook, so PR sent: https://github.com/pangeo-data/pangeo-docker-images/pull/311

Ref https://github.com/2i2c-org/infrastructure/issues/1237


yuvipanda:
I've deployed this on the m2lines hub at m2lines.2i2c.cloud, as I had to run `terraform apply` and move the entire cluster to a different zone to get GPUs.

Good for review now!

sgibson91 (approved):
LGTM!

PR #1246, opened 2022-04-26
Replace terraform lint GitHub Action with pre-commit config

sgibson91:


sgibson91:
Yeah, don't think terraform is installed in the pre-commit.ci env so I gave up 😆

Issue #1248, opened 2022-04-26
[Incident] m2lines hub unavailable for a short period of time


yuvipanda:
### Summary

From https://2i2c.freshdesk.com/a/tickets/125

> Around 4:40 ET, everything stopped.  I refreshed my browser and just saw a plain text message "service unavailable". This lasted for about 5 minutes. Then it came back. The notebook and cluster were still alive. I carried on.

> In the end, it was not that big of a problem. But it was very disruptive. All of the new hub users who are about to hit this hub will get very frustrated if they have the same experience.

### Important information

- Hub URL: {{ INSERT HUB URL HERE }}
- Support ticket ref: {{ INSERT SUPPORT REF HERE }}

### Tasks and updates

- [ ] Discuss and address incident, leaving comments below with updates
- [ ] Incident has been dealt with or is over
- [ ] Copy/paste the after-action report below and fill in relevant sections
- [ ] Incident title is discoverable and accurate
- [ ] All actionable items in report have linked GitHub Issues

<details>
<summary>After-action report template</summary>

```
# After-action report

These sections should be filled out once we've resolved the incident and know what happened.
They should focus on the knowledge we've gained and any improvements we should take.

## Timeline

_A short list of dates / times and major updates, with links to relevant comments in the issue for more context._

All times in {{ most convenient timezone}}.

- {{ yyyy-mm-dd }} - [Summary of first update](link to comment)
- {{ yyyy-mm-dd }} - [Summary of another update](link to comment)
- {{ yyyy-mm-dd }} - [Summary of final update](link to comment)


## What went wrong

_Things that could have gone better. Ideally these should result in concrete
action items that have GitHub issues created for them and linked to under
Action items._

- Thing one
- Thing two

## Where we got lucky

_These are good things that happened to us but not because we had planned for them._

- Thing one
- Thing two

## Follow-up actions

_Every action item should have a GitHub issue (even a small skeleton of one) attached to it, so these do not get forgotten. These issues don't have to be in `infrastructure/`, they can be in other repositories._

### Process improvements

1. {{ summary }} [link to github issue]
2. {{ summary }} [link to github issue]

### Documentation improvements

1. {{ summary }} [link to github issue]
2. {{ summary }} [link to github issue]

### Technical improvements

1. {{ summary }} [link to github issue]
2. {{ summary }} [link to github issue]
```

</details>

yuvipanda:
[...]

PR #1249, opened 2022-04-27
Added Canada College to cloudbank

sean-morris:
- Added Canada College to the cloudbank hub configuration

PR #1251, opened 2022-04-27
Move m2lines & LEAP hubs to regional  HA clusters

yuvipanda:
These are less prone to k8s API failures, like those reported in https://github.com/2i2c-org/infrastructure/issues/1248

We now default to provisioning regional HA clusters - these cost ~70$ a month,
but I think the added reliability is worth it

Ref https://github.com/2i2c-org/infrastructure/issues/1102

### DIFF
--- a/docs/howto/operate/new-cluster/new-cluster.md
+++ b/docs/howto/operate/new-cluster/new-cluster.md
@@ -46,8 +46,11 @@ See the [variables file](https://github.com/2i2c-org/infrastructure/tree/HEAD/te
 Example `.tfvars` file:
 
 ```
-prefix     = "my-awesome-project"
-project_id = "my-awesome-project-id
+prefix           = "my-awesome-project"
+project_id       = "my-awesome-project-id
+zone             = "us-central1-c"
+region           = "us-central1"
+regional_cluster = true
 ```
 ````
 
--- a/docs/howto/operate/new-cluster/new-cluster.md
+++ b/docs/howto/operate/new-cluster/new-cluster.md
@@ -30,6 +30,15 @@ The _minimum_ inputs this file requires are:
   Primary identifier to 'group' together resources.
 - `project_id`: GCP Project ID to create resources in.
   Should be the id, rather than display name of the project.
+- `regional_cluster`: Set to true to provision a [GKE Regional
+  Highly Available cluster](https://cloud.google.com/kubernetes-engine/docs/concepts/regional-clusters).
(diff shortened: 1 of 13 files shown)


yuvipanda:
You can't 'upgrade' from a non-HA cluster to a HA cluster, needs a redeploy. I've redeployed both the LEAP and m2lines clusters :)

docs/howto/operate/new-cluster/new-cluster.md
```diff
@@ -30,6 +30,16 @@ The _minimum_ inputs this file requires are:
...
+- `regional_cluster`: Set to true to provision a [GKE Regional
+  Highly Available cluster](https://cloud.google.com/kubernetes-engine/docs/concepts/regional-clusters).
+  Costs ~70$ a month, but worth it for the added reliability for most
+  cases except when cost saving is an absolute requirement. Defaults
+  to `true`.
+- `zone`: Zone where cluster nodes and filestore for home directory
+  are created.
+- `region`: Region where cluster master (if `regional_cluster` is
+  `true`) is run, as well as any storage buckets created with
+  `user_buckets`.
```
GeorgianaElena, line 42: What do you think about adding these extra inputs (`regional_cluster`, `zone`, `region`) to the example below too: 

https://github.com/2i2c-org/infrastructure/blob/1b45e2fe7f1dce300be12996622e3786aeccc054/docs/howto/operate/new-cluster/new-cluster.md?plain=1#L46-L52

Also, I just noticed there are some extra  ` ```` ` there too

GeorgianaElena (approved):
Thanks @yuvipanda! I've left a small, docs-related comment, but this looks good to me 🚀

docs/howto/operate/new-cluster/new-cluster.md
sgibson91: 'the extra `````` matches this line, so it is correct :) https://github.com/2i2c-org/infrastructure/blob/1b45e2fe7f1dce300be12996622e3786aeccc054/docs/howto/operate/new-cluster/new-cluster.md#L26

GeorgianaElena: Ah, right! Thanks @sgibson91 <3

yuvipanda: @GeorgianaElena added these additional variables to the example!

yuvipanda:
Linkcheck failed with failure for canada.cloudbank.2i2c.cloud, which is from #1249. I'll investigate separately!

Issue #1261, opened 2022-05-02
[UTexas] Sidecar database container malfunction for an individual user


consideRatio:
A user in the UTexas hub has their sidecar PostgreSQL container failing to start or similar: https://2i2c.freshdesk.com/a/tickets/127

## Action points

- Investigate and try to figure out the reason for the failure
- If possible solve it
- Communicate back

GeorgianaElena:
### Update
Checking the logs of the `postrgres` container for the user that reported issues showed the following error:

```
 PANIC:  could not write to file "pg_wal/xlogtemp.14": No space left on device
```

Which probably means that the `postgres` server can't start because of the lack of space, which seems to be confirmed by the fact that that the `postgress-{username}` volume is almost full.

### Possible solutions

1. Manually resize the disk for this particular user to fix the issue -> but this would create inconsistencies with the infra repo

2. Ask the user to delete some data from the db to free up some space

3. Resize the db disks for all the users from 1GB to 2GB ([code ref](https://github.com/2i2c-org/infrastructure/blob/master/config/clusters/2i2c/utexas.values.yaml#L50))

consideRatio:
I think asking the representative makes sense here. I see their options as: 
1. they can ask their user to clear some space on their disk,
2. or, they ask that we increase limits for all users to X GB from 1GB.

The option of asking us to manage insividual users disk space seems possibly unsustainable and is an option ill suggest we dont propose at this point.

Long term, allowing users etc to see their storage space quota etc would be very helpful. Then, a user may figure out right away something would be wrong due to max storage.

yuvipanda:
We mount the postgres database storage in the user's container too - https://github.com/2i2c-org/infrastructure/blob/f52a643e8e27ddbd2cbf0a64334fd025f36f5ce8/config/clusters/2i2c/utexas.values.yaml#L79. They *should* have enough permissions to delete it all themselves. `df -h` on a terminal *should* show them quota info but I agree this should be far more visible.

choldgraf:
Note that @jameshowison actually suggested in the FreshDesk ticket that they'd need significantly more than 2GB - more like 10GB I believe. So it sounds like the limit that's there just about an order of magnitude too small?

yuvipanda:
Looking at https://cloud.google.com/compute/disks-image-pricing#disk, it's about 0.04GB per GB per month. So if there are ~40 students, and we give them all 10G that's about a 16$ increase in monthly cost. So let's just do it?

[...]

PR #1262, opened 2022-05-02
Resize utexas postgres volumes to be 10G

yuvipanda:
k8s should dynamically resize the volumes the next time
the user starts their server

Fixes https://github.com/2i2c-org/infrastructure/issues/1261

### DIFF
--- a/config/clusters/2i2c/utexas.values.yaml
+++ b/config/clusters/2i2c/utexas.values.yaml
@@ -47,6 +47,8 @@ jupyterhub:
             )
 
         # 1Gi seems the smallest PVC you can make - anything smaller than that is rounded up to 1Gi
+        # This size can be increased, and users will get the newer sized disk when they start
+        # their servers next - https://kubernetes.io/docs/concepts/storage/persistent-volumes/#expanding-persistent-volumes-claims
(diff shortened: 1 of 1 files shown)


config/clusters/2i2c/utexas.values.yaml
```diff
@@ -46,8 +46,8 @@ jupyterhub:
                 annotations=annotations
             )
 
-        # 1Gi seems the smallest PVC you can make - anything smaller than that is rounded up to 1Gi
-        make_db_pvc = partial(make_extra_pvc, 'postgres-storage', 'postgres-{username}', 'standard-rwo', '1Gi')
+        # 10Gi seems the smallest PVC you can make - anything smaller than that is rounded up to 1Gi
```
consideRatio, line 49:
```suggestion
        # 1Gi seems the smallest PVC you can make - anything smaller than that is rounded up to 1Gi
```

consideRatio (commented):
Briefly reviewed this from mobile device, is this going to apply changes to pvcs already created, and if so, does the pvcs allow getting their size increased as sometimes needs to be excplicitly allowed by a flag on the storageclass or PV that was created for the pvc if i recall correctly?

yuvipanda:
@consideRatio yeah this is the standard gcp storage class that supports automatic resizing

yuvipanda:
I've added a comment about PVC expansion to the code too!

Issue #1268, opened 2022-05-04
generate-jobs said deployments should happen, why was the subsequent workflow jobs get skipped?


consideRatio:
This is a followup to figure out why #1239 didn't trigger deployments of staging/prod, and practically, why the workflow's jobs associated with that was skipped.

![image](https://user-images.githubusercontent.com/3837114/166693869-92e2104e-8e3d-4e5e-873d-602371da8d6e.png)

What we know is that when we look at "View details" on the merged PR, we see that the deployment jobs etc didn't trigger even though this PR changed the following relevant files...
- helm-charts/basehub/values.yaml
- config/clusters/leap/common.values.yaml
... and that the deployer script correctly decided that many hubs should have been re-deployed.

![image](https://user-images.githubusercontent.com/3837114/166694432-9ebba71e-52f9-4a7e-9634-048357602476.png)

To inspect the workflow run, see: https://github.com/2i2c-org/infrastructure/runs/6168152254?check_suite_focus=true

---

Why was the staging/filter/prod jobs skipped?

The check for the support job looks like this, and seems sound to me. It seems as if the outputs was set, to `"[]"`, but not to the actual generate-jobs concluded changes. Could that be the case? Could the generate-jobs manage to print the list clearly, but fail to set the output properly?

```yaml
    # Only run this job on pushes to the default branch and when the job output is not
    # an empty list
    if: |
      (github.event_name == 'push' && contains(github.ref, 'master')) &&
      needs.generate-jobs.outputs.support-and-staging-matrix-jobs != '[]'
```

Looking into the deployer code responsible to set the output, I see no reason for an issue, everything looks good. I'm clueless why the deploy steps were skipped.

https://github.com/2i2c-org/infrastructure/blob/2af57d0e3d1aa2ec98e69504338fda84899d6506/deployer/deploy_actions.py#L330-L339

My hypothesis of what has gone wrong is that something in the deployer script made the support output be set to `[]` for some reason, even though we printed it successfully to be a long list of hubs to deploy. I assume that this change must stem from something quite recent as well, as I'm quite confident we had this functional before.

sgibson91:
You are looking at the wrong action logs. You are looking at the actions that ran _in the PR_ and staging/filter/prods get skipped during the PR. If you look at the logs from the merge commit on `master`, you can see that a lot of failures happened and the PR wasn't deployed to prod hubs at all until I ran the deployer locally against master yesterday

[...]

PR #1272, opened 2022-05-04
Add workflow that comments a link to the correct actions location when a PR is merged

sgibson91:
In https://github.com/2i2c-org/infrastructure/issues/1268 it became clear that GitHub's UI is confusing when trying to distinguish between GitHub Action workflows that ran in a PR vs those that run in `master` after a PR is merged.

This PR adds a new GitHub Action workflow that will comment on a merged PR with a link to the `deploy-hubs` workflow running in `master` after a PR that will trigger that PR is merged.

Workflow failure is unrelated.

### DIFF
--- a/.github/workflows/comment-test-link-merged-pr.yaml
+++ b/.github/workflows/comment-test-link-merged-pr.yaml
@@ -47,6 +47,8 @@ jobs:
     # on the master branch.
     runs-on: ubuntu-latest
     needs: [get-pr-number]
+    permissions:
+      issues: write
     if: needs.get-pr-number.outputs.continue-workflow == 'True'
     steps:
       - name: Comment on merged PR with GitHub Actions link
--- a/.github/workflows/comment-test-link-merged-pr.yaml
+++ b/.github/workflows/comment-test-link-merged-pr.yaml
@@ -15,7 +15,7 @@ jobs:
   # trigger this workflow in a Pull Request context, it is not readily available to us.
   # Instead, we extract the PR number from the message of the head commit in the push
   # event payload. In the case of a merged PR, this message will be of the form:
-  # 'Merged pull request #XXX from owner/branch'
+  # 'Merge pull request #XXX from owner/branch'
   # This job sets two outputs: continue-workflow (bool, whether or not to run the next
   # job) and pr-number (str)
   get-pr-number:
--- a/.github/workflows/comment-test-link-merged-pr.yaml
+++ b/.github/workflows/comment-test-link-merged-pr.yaml
@@ -1,0 +1,59 @@
+name: Comment GitHub Actions link on a merged PR
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - deployer/**
+      - requirements.txt
+      - .github/actions/setup-deploy/**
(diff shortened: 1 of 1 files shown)


.github/workflows/comment-test-link-merged-pr.yaml
```diff
@@ -0,0 +1,70 @@
...
+  add-pr-comment:
+    # This job depends on the get-pr-number job to have completed successfully and set
+    # the continue-workflow variable to True. It consumes the pr-number variable to
+    # comment on that Pull Request with a link to the deploy-hubs.yaml wiorkflow running
+    # on the master branch.
+    runs-on: ubuntu-latest
+    needs: [get-pr-number]
+    if: needs.get-pr-number.outputs.continue-workflow == 'True'
+    steps:
+      - name: Comment on closed PR with GitHub Actions link
```
sgibson91, line 52:
```suggestion
      - name: Comment on merged PR with GitHub Actions link
```

sgibson91:
Example of the workflow in action

- https://github.com/sgibson91/testing-gh-actions/pull/17

consideRatio (approved):
Nice, should be very safe to merge no matter what!

It probably would be a good idea if you defined the permissions required for this to function explicitly in the 2i2c org as well. The GITHUB_TOKEN should IMO have read access only by default like we have it in the jupyterhub org, unsure if its the case in the 2i2c org though.

I guess the permissions we require will be... ([based on this](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/)):

```yaml
  permissions:
    pull-requests: write
```

.github/workflows/comment-test-link-merged-pr.yaml
```diff
@@ -0,0 +1,70 @@
...
+          else:
+              print(f"::set-output name=pr-number::{None}")
+
+  add-pr-comment:
+    # This job depends on the get-pr-number job to have completed successfully and set
+    # the continue-workflow variable to True. It consumes the pr-number variable to
+    # comment on that Pull Request with a link to the deploy-hubs.yaml wiorkflow running
+    # on the master branch.
+    runs-on: ubuntu-latest
+    needs: [get-pr-number]
```
consideRatio, line 49:
```suggestion
    needs: [get-pr-number]
    permissions:
      pull-requests: write
```

sgibson91: I _think_ it might actually need to be `issues` because, according to the GitHub REST API, a PR is just a special type of issue. And since we're only commenting, not updating the PR-ness, then I think `issues` is right. For example, we use the `issues` endpoint to create the comment here

https://github.com/2i2c-org/infrastructure/pull/1272/files#diff-e38f6f1c50d8c5fc35bba540c9ba5d99d4a1e817453e5229bda0de776fb1895bR61

sgibson91, line 49:
```suggestion
    needs: [get-pr-number]
    permissions:
      issues: write
```

consideRatio: Ah nice catch!

Issue #1271, opened 2022-05-04
Policy for keeping k8s clusters updated to somewhat a similar version


consideRatio:
@GeorgianaElena just ran into failures because the k8s client (`kubectl`) used in the github virtual environment was bumped to 1.24.0, and since the version skew policy they have declares that it should be compatible with one minor above and two minor below, we can expect errors if `kubectl` interacts with a k8s cluster versioned <=1.21.

The resolution for now is probably is to pin the kubectl version to 1.23, which is compatible with 1.24 that is the recently released version of k8s, but also lower versions all the way down to 1.21 at least.

I'm not sure if there is a policy on upgrading the k8s clusters, but we should probably have one so that we don't fall behind too far where some clusters have very old versions and others have very new.

consideRatio:
Closing as duplicate of #1271 with more concrete ideas

sgibson91:
> Closing as duplicate of #1271 with more concrete ideas

But... this is issue 1271? What was the duplicated issue?

consideRatio:
Woops thanks @sgibson91, I meant #412!

PR #1273, opened 2022-05-04
CI: pin kubectl version

GeorgianaElena:
Temporary fix until we figure out https://github.com/2i2c-org/infrastructure/issues/1271

### DIFF
--- a/.github/actions/setup-deploy/action.yaml
+++ b/.github/actions/setup-deploy/action.yaml
@@ -81,7 +81,6 @@ runs:
     - uses: azure/setup-kubectl@v2.0
       with:
         version: "1.23"
-      id: install
 
     # This step could be cached, but we only do it for a few jobs in the
     # workflow where this action is used
--- a/.github/actions/setup-deploy/action.yaml
+++ b/.github/actions/setup-deploy/action.yaml
@@ -80,7 +80,7 @@ runs:
     # See https://github.com/2i2c-org/infrastructure/issues/1271.
     - uses: azure/setup-kubectl@v2.0
       with:
-        version: '1.23'
+        version: "1.23"
       id: install
 
     # This step could be cached, but we only do it for a few jobs in the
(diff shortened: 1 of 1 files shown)


.github/actions/setup-deploy/action.yaml
```diff
@@ -76,6 +76,13 @@ runs:
         # behalf.
         version: v3.8.2
 
+    # Pin kubectl version to 1.23 otherwise interactions with k8s clusters versioned <=1.21 won't work.
+    # See https://github.com/2i2c-org/infrastructure/issues/1271.
+    - uses: azure/setup-kubectl@v2.0
+      with:
+        version: "1.23"
+      id: install
```
consideRatio, line 84:
```suggestion
```

The workflow has a jobs, the jobs has steps, this step has an id set to install. This is only relevant if we want to reference the outputs of this step or similar later, which we currently don't. Due to that, let's not add an id for this step.

consideRatio:
I don't think it closes #1271, it only mitigates the issue that makes us want to have a policy and procedure to upgrade our k8s clusters to a somewhat modern and cohesive version. I'm thinking for example that perhaps we should ensure all our clusters has version 1.21+ or similar at this point in time when 1.24 is out.

.github/actions/setup-deploy/action.yaml
GeorgianaElena: Oh, right! Copy-pasta gone wrong 😅 Thanks @consideRatio

GeorgianaElena:
> I don't think it closes https://github.com/2i2c-org/infrastructure/issues/1271, it only mitigates the issue that makes us want to have a policy and procedure to upgrade our k8s clusters to a somewhat modern and cohesive version. I'm thinking for example that perhaps we should ensure all our clusters has version 1.21+ or similar at this point in time when 1.24 is out.

Good point @consideRatio! I've updated the top comment so that it won't close the issue

PR #1277, opened 2022-05-04
Update CI triggers to be more robust than **.yaml

sgibson91:
We wanted to exclude full scale redeployments if Our GitHub Actions workflows files changed, but the `**.yaml` path actually included these. I've updated the path triggers to only include `config/clusters/**` and `helm-charts/**` since changes on these paths should trigger large scale redeployments.


sgibson91:
l'll self-merge this as I don't believe it's controversial - it will only reduce the number of times deploy-hubs.yaml will be triggered

PR #1283, opened 2022-05-06
Add a README to document the deployer

sgibson91:
This PR adds a README to document the deployer module. It's very rough and mostly includes the CLI help info right now.

### DIFF
--- a/deployer/README.md
+++ b/deployer/README.md
@@ -2,6 +2,10 @@
 
 ## Installing dependencies
 
+```bash
+pip install -r requirements.txt
+```
+
 ## Functions
 
 ```bash
@@ -122,4 +126,11 @@ positional arguments:
 optional arguments:
   -h, --help            show this help message and exit
   --check-dask-scaling  For daskhubs, optionally check that dask workers can be scaled
+```
+
+### Running Tests
+
+```bash
+pip install -r dev-requirements.txt
+python -m pytest -vvv
 ```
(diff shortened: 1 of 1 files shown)


deployer/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+This section describes the functions the `deployer` can carry out, their purpose, and their command line usage.
+
+**Note:** The `deployer` must currently be invoked from the root of this repository, i.e.:
+
+```bash
+$ pwd
+[...]/infrastructure/deployer
+$ cd .. && pwd
+[...]/deployer
```
consideRatio, line 26:
```suggestion
[...]/infrastructure
```

consideRatio (approved):
Lgtm! Great work documenting this! I've reasoned a bit below but i suggest not acting on anything unless it resonates clearly with you as well.

---

Can the help strings you write be part of the help text of the CLI as well btw, im not sure, hmmm? If so, maybe we want that to be part of the CLI emitted help.

I realize one aspect i hadn't considered when we discussed this IRL was that all help isnt provided by a single call with -h, only the help on that parser level and not for subcommands etc. Due to that, having a readme.md capture the help description of the CLI is less easy, and its more tricky to keep the help text updated.

Ideally in my mind, we would have a readme including the full help text of the CLI and all subcommands to avoid repeating ourselves and also having most documentation available in both markdown and via cli execution.

I really appreciate co-locating a readme that at least points to where documentation is unless its contains it directly as it does now which is also great i think!

GeorgianaElena (approved):
Thanks for adding these docs @sgibson91 ✨  🚀

PR #1285, opened 2022-05-09
Don't set node_locations for zonal clusters

yuvipanda:
Without this, terraform apply fails with:

> for multi-zonal clusters, node_locations should not contain the primary 'zone'


GeorgianaElena (approved):

yuvipanda:
Thanks, @GeorgianaElena

PR #1286, opened 2022-05-09
Set GPU config for all terraform workspaces

yuvipanda:
- Since we don't use terraform nested defaults, we have to explicitly say
   'no gpu' for all nodes that don't have a GPU - otherwise terraform apply won't work
- Explicitly set GPU taint for GPU nodes, as terraform otherwise sets this implicitly, freaks
  out on each `apply` and wants to recreate the entire nodepool


GeorgianaElena (approved):
> we have to explicitly say 'no gpu' for all nodes that don't have a GPU - otherwise terraform apply won't work

:(

LGTM 🚀

PR #1289, opened 2022-05-09
Create a BinderHub helm chart for deployment onto 2i2c's infrastructure

sgibson91:
This PR adds a `binderhub` helm chart with dask-gateway enabled under our `helm-charts` folder


helm-charts/binderhub/values.yaml
```diff
@@ -0,0 +1,560 @@
...
+        nodeSelector:
+          hub.jupyter.org/node-purpose: core
+        resources:
+          requests:
+            memory: 64Mi
+          limits:
+            memory: 1Gi
+    singleuser:
+      {{- if .Values.dask-gateway.enabled }}
+      # Almost everyone using dask by default wants JupyterLab
```
yuvipanda, line 58: Unfortunately we can't template values.yaml files :(

sgibson91:
Ok, I've gone a different route. Two helm charts: binderhub and binderdask. Binderdask has the dask-gateway dependency. Yes, this sucks, but I also don't know what else to do given that (i) helm charts aren't composable, and (ii) we don't know what the future of BinderHub looks like yet given the upcoming GESIS work.

sgibson91:
I think validating that these charts are up-to-date with basehub/daskhub will be less filled with logic acrobatics as well with this approach.

sgibson91:
@yuvipanda ok, made that change!

yuvipanda (approved):
This LGTM!

helm-charts/binderhub/values.yaml
```diff
@@ -0,0 +1,522 @@
+binderhub:
+  config:
+    BinderHub:
+      use_registry: true
```
sgibson91, line 4:
```suggestion
      use_registry: true
      build_image: "quay.io/jupyterhub/repo2docker:2022.02.0-30.gc4e01b3"
```

Add repo2docker image tag

sgibson91:
Just pushed an extra commit to include the repo2docker tag. will merge now.

PR #1292, opened 2022-05-10
Block auth for test user

GeorgianaElena:
- fixes https://github.com/2i2c-org/meta/issues/321
- adds the test username to the list of users that are not allowed to login using the authentication flow
- the test hub service can still create this user, and start a server for it
- I have checked how this will behave for `username_pattern` and `blocked_users` takes priority over `username_pattern` in the auth flow, and it works as expected
- we're blocking this username for all hubs, no matter the authenticator used, since it doesn't hurt, simplifies the implementation, and it will be useful if we ever want to do domain stripping for the hubs using the email claim

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -251,6 +251,10 @@ jupyterhub:
                 matchLabels:
                   app.kubernetes.io/component: traefik
   hub:
+    config:
+      Authenticator:
+        blocked_users:
+          - deployment-service-check
     extraFiles:
       configurator-schema-default:
         mountPath: /usr/local/etc/jupyterhub-configurator/00-default.schema.json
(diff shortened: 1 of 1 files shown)


helm-charts/basehub/values.yaml
```diff
@@ -251,6 +251,10 @@ jupyterhub:
                 matchLabels:
                   app.kubernetes.io/component: traefik
   hub:
+    config:
+      Authenticator:
+        blocked_users:
```
yuvipanda, line 256: Let's add a comment here on why we have this?

yuvipanda (approved):
AMAZING FIND!!!!

helm-charts/basehub/values.yaml
GeorgianaElena: Done in https://github.com/2i2c-org/infrastructure/pull/1292/commits/0b56d64e06167b08d24b18270495d84359944952

GeorgianaElena:
Thanks for brainstorming with me on this one @sgibson91 ❤️

Issue #1229, opened 2022-04-23
Allow configurable retention policies in terraform created scratch buckets


yuvipanda:
### Context

We currently have a 7-day delete policy for scratch buckets created using config-connector (https://github.com/2i2c-org/infrastructure/blob/master/helm-charts/basehub/templates/cloud-resources/gcp/storage-bucket.yaml#L10-L16) but don't have a way to automate that when scratch buckets are created via terraform. We should!

### Proposal

- [ ] Expand the `user_buckets` terraform variable to allow specifying per-bucket retention policies
- [ ] Apply it to current scratch buckets (mostly unused) that are created with terraform

PR #1297, opened 2022-05-11
Provide 7 day retention policy for terraform created buckets

yuvipanda:
Applied to m2lines and LEAP, which didn't have retention
policies set yet.

Fixes https://github.com/2i2c-org/infrastructure/issues/1229


sgibson91 (approved):
LGTM!

Issue #1295, opened 2022-05-11
[LEAP] Inconsistent scratch bucket environment variables


GeorgianaElena:
Reported in https://2i2c.freshdesk.com/a/tickets/137

### Scenario

Running:

```print(os.environ['SCRATCH_BUCKET'])``` prints ➡️ `gcs://leap-scratch/user`
```print(os.environ['PANGEO_SCRATCH'])``` prints ➡️ `gs://leap-pangeo-scratch/user`

### Context

According to ⬇️ they should be the same
https://github.com/2i2c-org/infrastructure/blob/4638da28c347cf0b434eae33a203c6ff9c8341ca/config/clusters/leap/common.values.yaml#L60-L61

But they're not because of https://github.com/2i2c-org/infrastructure/blob/841fc8555ac4ec46505e800a8b74ddc06be46117/config/clusters/leap/prod.values.yaml#L8

### Problem 
PANGEO_SCRATCH var [is deprecated](https://github.com/2i2c-org/infrastructure/blob/841fc8555ac4ec46505e800a8b74ddc06be46117/docs/topic/features.md#scratch-buckets-on-google-cloud-storage), so I don't think this inconsistency should cause issues, but we should be consistent.

PR #1298, opened 2022-05-11
Set PANGEO_SCRATCH for leap and m2lines hubs

yuvipanda:
While we consider PANGEO_SCRATCH deprecated, it's still
set in the upstream Pangeo images
(https://github.com/pangeo-data/pangeo-docker-images/blob/1939869ce8e3b3accb65a0fdf40ec7eedc7f7235/pytorch-notebook/start#L6
and similar), so let's keep it for now.

This commit removes the setup in common.yaml for leap, as
staging and prod have separate scratch buckets

Fixes https://github.com/2i2c-org/infrastructure/issues/1295

### DIFF
--- a/config/clusters/leap/prod.values.yaml
+++ b/config/clusters/leap/prod.values.yaml
@@ -6,6 +6,7 @@ basehub:
     singleuser:
       extraEnv:
         SCRATCH_BUCKET: gcs://leap-scratch/$(JUPYTERHUB_USER)
+        PANGEO_SCRATCH: gcs://leap-scratch/$(JUPYTERHUB_USER)
     hub:
       config:
         GitHubOAuthenticator:
--- a/config/clusters/m2lines/prod.values.yaml
+++ b/config/clusters/m2lines/prod.values.yaml
@@ -6,6 +6,7 @@ basehub:
     singleuser:
       extraEnv:
         SCRATCH_BUCKET: gcs://m2lines-scratch/$(JUPYTERHUB_USER)
+        PANGEO_SCRATCH: gcs://m2lines-scratch/$(JUPYTERHUB_USER)
     hub:
       config:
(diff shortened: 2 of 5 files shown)


GeorgianaElena (approved):

config/clusters/leap/prod.values.yaml
```diff
@@ -6,6 +6,7 @@ basehub:
     singleuser:
       extraEnv:
         SCRATCH_BUCKET: gcs://leap-scratch/$(JUPYTERHUB_USER)
+        PANGEO_SCRATCH: gcs://leap-scratch/$(JUPYTERHUB_USER)
```
choldgraf, line 9:
```suggestion
        # Pangeo defines this variable in their image, so we also use it here+staging to avoid confusion.
        PANGEO_SCRATCH: gcs://leap-scratch/$(JUPYTERHUB_USER)
```

config/clusters/m2lines/prod.values.yaml
```diff
@@ -6,6 +6,7 @@ basehub:
     singleuser:
       extraEnv:
         SCRATCH_BUCKET: gcs://m2lines-scratch/$(JUPYTERHUB_USER)
+        PANGEO_SCRATCH: gcs://m2lines-scratch/$(JUPYTERHUB_USER)
```
choldgraf, line 9:
```suggestion
        # Pangeo defines this variable in their image, so we also use it here+staging to avoid confusion.
        PANGEO_SCRATCH: gcs://m2lines-scratch/$(JUPYTERHUB_USER)
```

choldgraf (commented):
WDYT about this quick context?

choldgraf:
oops looks like this got merged seconds before my comments - ah well, doesn't need to be a blocker

PR #1299, opened 2022-05-11
Use location instead of zone for filestore

yuvipanda:
Removes this deprecation warning:

╷
│ Warning: Argument is deprecated
│
│   with google_filestore_instance.homedirs,
│   on storage.tf line 4, in resource "google_filestore_instance" "homedirs":
│    4:   zone    = var.zone
│
│ Deprecated in favor of location.


yuvipanda:
no-op otherwise

sgibson91 (approved):

Issue #1252, opened 2022-04-27
Move to using OAuthenticator from upstream


yuvipanda:
### Context

With https://github.com/2i2c-org/infrastructure/pull/1239, we are using https://github.com/jupyterhub/oauthenticator/pull/498 even though it isn't merged yet. We should get that merged, and switch to using OAuthenticator from upstream or from a release.

### Proposal

- [x] https://github.com/jupyterhub/oauthenticator/pull/498 get this merged
- [ ] Modify our hub image to use the OAuthenticator version from upstream rep
- [ ] Write docs on how to customize hub image

PR #1303, opened 2022-05-11
Use OAuthenticator from upstream

GeorgianaElena:
Closes https://github.com/2i2c-org/infrastructure/issues/1252

https://github.com/jupyterhub/oauthenticator/pull/498 has been merged and although there isn't a new release for `oauthenticator`, until then we can still use the upstream repo.

### DIFF
--- a/docs/topic/hub-image.md
+++ b/docs/topic/hub-image.md
@@ -1,0 +1,59 @@
+# Hub Image
+
+The 2i2c hubs use a custom hub image that is defined in the [helm-charts directory of the infrastructure repository](https://github.com/2i2c-org/infrastructure/tree/HEAD/helm-charts/images/hub).
+
+This custom hub image is built on top of the [jupyterhub/k8s-hub](https://hub.docker.com/r/jupyterhub/k8s-hub) Docker image and configured based on the needs of 2i2c hubs. This allows adding and configuring other packages like the [jupyterhub-configurator](https://github.com/yuvipanda/jupyterhub-configurator) or using specific versions of the spawner and authenticator. More information about this custom image can be found in the [Dockerfile](https://github.com/2i2c-org/infrastructure/blob/HEAD/helm-charts/images/hub/Dockerfile) itself.
+
+The 2i2c custom hub image is lives at [Quay.io](https://quay.io/repository/2i2c/pilot-hub?tab=tags).
+
+## Updating the hub image
+
+When this hub image needs to be updated, the steps to take are:
+
+1. Update the [Dockerfile]((https://github.com/2i2c-org/infrastructure/blob/HEAD/helm-charts/images/hub/Dockerfile) with any changes wanted
+
+2. Commit the changes
+
+   ```bash
+   git add helm-charts/images/hub/Dockerfile
+   git commit
+   ```
+
+   ```{note}
+   The commit SHA with be used to generate the image tag.
+   ```
+
+3. Rebuild the Docker image and push it to the [Quay.io registry](https://quay.io/repository/2i2c/pilot-hub)
+ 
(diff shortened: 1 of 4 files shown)


yuvipanda:
Thanks for working on this, @GeorgianaElena! I think you need to run `chartpress --push` from inside the `helm-charts` directory to build the image and update the tag.

GeorgianaElena:
oops! Thanks @yuvipanda! I've pushed the image now and updated the tag.

Chartpress also updated the basehub version in https://github.com/2i2c-org/infrastructure/blob/master/helm-charts/basehub/Chart.yaml#L7, but that didn't seem to be using the previous tag.

I don't remember if this was intended or it is an error (I feel there was a discussion about his at some point)?

yuvipanda (approved):
This LGTM, @GeorgianaElena! chartpress updates the basehub version, but I think we should discard that - otherwise, it may cause problems with the daskhub dependency mechanism.

Doesn't have to block this PR, but can you document how to rebuild the hub image too?

docs/topic/hub-image.md
```diff
@@ -0,0 +1,59 @@
...
+
+This custom hub image is built on top of the [jupyterhub/k8s-hub](https://hub.docker.com/r/jupyterhub/k8s-hub) Docker image and configured based on the needs of 2i2c hubs. This allows adding and configuring other packages like the [jupyterhub-configurator](https://github.com/yuvipanda/jupyterhub-configurator) or using specific versions of the spawner and authenticator. More information about this custom image can be found in the [Dockerfile](https://github.com/2i2c-org/infrastructure/blob/HEAD/helm-charts/images/hub/Dockerfile) itself.
+
+The 2i2c custom hub image is lives at [Quay.io](https://quay.io/repository/2i2c/pilot-hub?tab=tags).
+
+## Updating the hub image
+
+When this hub image needs to be updated, the steps to take are:
+
+1. Update the [Dockerfile]((https://github.com/2i2c-org/infrastructure/blob/HEAD/helm-charts/images/hub/Dockerfile) with any changes wanted
```
GeorgianaElena, line 13:
```suggestion
1. Update the [Dockerfile](https://github.com/2i2c-org/infrastructure/blob/HEAD/helm-charts/images/hub/Dockerfile) with any changes wanted
```

GeorgianaElena:
Thanks @yuvipanda! I've just added some docs and I'll merge the PR now 🚀

docs/topic/hub-image.md
```diff
@@ -0,0 +1,59 @@
...
+
+     ```bash
+     pip install chartpress
+     ```
+
+     This package is also listed under `dev-requirements.txt`, so it should be present if you've installed the dev dependencies.
+
+   - Run chartpress to build the image, push it to the registry and update the basehub helm chart to use the updated image tag
+
+     ```bash
```
choldgraf, line 46: extra kudos to @GeorgianaElena for documenting these :-)

helm-charts/images/hub/Dockerfile
```diff
@@ -19,7 +19,7 @@ RUN pip install --no-cache git+https://github.com/yuvipanda/jupyterhub-configura
 RUN pip install --no-cache --upgrade jupyterhub-kubespawner==1.1.2
 
 # Brings in https://github.com/jupyterhub/oauthenticator/pull/498
-RUN pip install --no-cache --upgrade git+https://github.com/yuvipanda/oauthenticator@7f3fdc0a14d06f1a081c23b1ceb7060a940d11f8
+RUN pip install --no-cache --upgrade git+https://github.com/jupyterhub/oauthenticator@d189037fe6b57453a3ae733a60740c50f241e713
```
damianavila, line 22: @GeorgianaElena, can we create a follow-up issue to use a released version when that one is released? I have been bitten before when I used unreleased versions...

Issue #1305, opened 2022-05-11
dask-gateway doesn't respect worker_cores option


yuvipanda:
From https://2i2c.freshdesk.com/a/tickets/141, no matter what option worker_cores is set to, it can't seem to be more than 1!

PR #1306, opened 2022-05-11
Remove fudging of worker_cores

yuvipanda:
min(worker_cores / 2, 1) is always going to be 1 for worker_cores > 2!

Not sure why this was here, but let's just pass through worker_cores
directly to the pod definition

Fixes https://github.com/2i2c-org/infrastructure/issues/1305


choldgraf:
this feels like a configuration equivalent of a [useless machine](https://www.youtube.com/watch?v=kproPsch7i0) 😅

sgibson91 (approved):

Issue #1278, opened 2022-05-05
Deploy the support charts in the remaining three clusters


GeorgianaElena:
### Context

We still have a few clusters where we don't deploy the supports charts to. This means that these clusters don't have grafana dashboards and can't be monitored through the central grafana at https://grafana.pilot.2i2c.cloud

### Proposal

Deploy the support chart for these clusters following the docs at https://infrastructure.2i2c.org/en/latest/howto/operate/grafana.html#deploy-the-support-chart

Note: we will need to tune their resource requests to match the smaller clusters.

### Updates and actions

Ref: https://github.com/2i2c-org/infrastructure/issues/328

List of clusters without the support chart:
- [x] leap -> is only missing the basicAuth prometheus setup https://github.com/2i2c-org/infrastructure/pull/1310
- [x] meom-ige -> https://github.com/2i2c-org/infrastructure/pull/1320
- [ ] farallon -> will be decommissioned, no need to deploy the support chart anymore https://github.com/2i2c-org/infrastructure/issues/1454

sgibson91:
LEAP has support deployed :) One down already, way to go!

https://github.com/2i2c-org/infrastructure/blob/d4ed3007bd7f51425fe9a15b36c4cfd6e7932a97/config/clusters/leap/cluster.yaml#L8-L10

sgibson91:
Ah good catch!

damianavila:
Related: https://github.com/2i2c-org/infrastructure/issues/594

GeorgianaElena:
Since farallon (last cluster without a support chart deployed) will be decommissioned soon, https://github.com/2i2c-org/infrastructure/issues/1454, I'll close this issue as completed.

damianavila:
Awesome work, @GeorgianaElena!

PR #1310, opened 2022-05-16
Expose leap prometheus isntance through basicAuth

GeorgianaElena:
Followed the steps in https://github.com/2i2c-org/infrastructure/pull/1091 for the leap cluster.

Reference https://github.com/2i2c-org/infrastructure/issues/1278#issuecomment-1118561470

### DIFF
--- a/config/clusters/leap/support.values.yaml
+++ b/config/clusters/leap/support.values.yaml
@@ -2,6 +2,9 @@ nvidiaDevicePlugin:
   gke:
     enabled: true
 
+prometheusIngressAuthSecret:
+  enabled: true
+
 grafana:
   ingress:
     hosts:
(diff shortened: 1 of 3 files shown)


yuvipanda (approved):

config/clusters/leap/support.values.yaml
```diff
@@ -11,6 +11,14 @@ grafana:
         hosts:
           - grafana.leap.2i2c.cloud
 prometheus:
+  ingress:
```
damianavila, line 14: I think you need the following piece at the top (to be congruent with other clusters) of the file:
```
prometheusIngressAuthSecret:
  enabled: true
```

In addition and related to this specific section, I think this snippet goes under the `server` key, right?

GeorgianaElena: Absolutely! Thanks for catching this @damianavila!

GeorgianaElena: Done in https://github.com/2i2c-org/infrastructure/pull/1310/commits/d40080b9364a5904ad3057eaa0a37c0f05e86299

consideRatio: Nice catch @damianavila! I think d40080b does the trick based on inspection of https://github.com/prometheus-community/helm-charts/blob/b73814645a57e2b13673b2f67dfbcdfd799ee946/charts/prometheus/templates/server/ingress.yaml#L2.

This would have been caught by promethues helm chart if they provided a strict schema!

consideRatio (approved):
LGTM!

consideRatio:
Nice, I note how this also didn't trigger a production hub re-deployment - which it shouldn't either I think! https://github.com/2i2c-org/infrastructure/actions/runs/2337748885

image

config/clusters/leap/support.values.yaml
damianavila: > This would have been caught by promethues helm chart if they provided a strict schema!

Should we open a follow-up issue for that, @consideRatio?

PR #1316, opened 2022-05-17
Cleanup base values.yaml files a little

yuvipanda:
- Use hub.config instead of extraConfig for some simple config
  settings.
- Remove proxy.https in the base, as it is disabled by default. We
  only use it in farallon and meom-ige, and so the config is shifted
  there.
- Remove a config override on daskhub that was config as basehub

### DIFF
--- a/config/clusters/farallon/common.values.yaml
+++ b/config/clusters/farallon/common.values.yaml
@@ -90,6 +90,8 @@ basehub:
         type: LoadBalancer
       https:
         enabled: true
+        letsencrypt:
+          contactEmail: yuvipanda@gmail.com
       chp:
         nodeSelector: {}
         tolerations:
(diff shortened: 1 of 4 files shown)


config/clusters/farallon/common.values.yaml
```diff
@@ -90,6 +90,8 @@ basehub:
         type: LoadBalancer
       https:
         enabled: true
+        letsencrypt:
+          contactEmail: yuvipanda@gmail.com
```
sgibson91, line 94: I wonder if we should change this to `support@2i2c.org` to remove you as a bottle neck?

sgibson91 (approved):

config/clusters/farallon/common.values.yaml
yuvipanda: @sgibson91 we definitely should but I didn't want to do that as part of this change!  As we move these two clusters to using support chart (#1278) this should just go away.

yuvipanda:
Tested this on the staging hub, worked alright!

Issue #1309, opened 2022-05-13
[New Hub] University of Washington - NASA SnowEx Hackweek 2022


damianavila:
### Hub Description

The hub has the following needs:
- Deployed in AWS us-west-2 datacenter
- Support up to 100 simultaneous users
- Minimum of 4 CPU, 16GB RAM, 10GB home directory per user
- GPU node group with NVIDIA K80, T4, or V100 GPU
- Read-write access to S3 scratch bucket
- Gate usage with GitHub Teams within Organization authentication https://github.com/orgs/snowex-hackweek/teams/participants-2022 
- Auto-deploy single docker image
https://quay.io/repository/uwhackweek/snowex 

### Community Representative(s)

@scottyhq

### Important dates

- **Required start date**: Jun 11th, 2022
- **Target start date**: Jun 1st, 2022
- **Any important dates for usage**: 
  - 2i2c engineer on-hand for same-day answering questions or troubleshooting during week-long main event Pacific Time 8-4 PT (July 11 - July 15)
  - Hub will remain accessible through September 1, 2022 or when budget runs out, whichever comes first

### Hub Authentication Type

GitHub Authentication (e.g., @mygithubhandle)

### Hub logo information

- **URL to Hub Image**: {{ URL HERE }}
- **URL for Image Link**: {{ URL HERE }}

### Hub user image

- Repository for user image: { REPO LINK IF IT EXISTS }
- User image registry: https://quay.io/repository/uwhackweek/snowex 
- User image tag and name: { NAME AND TAG IF IT EXISTS }

### Extra features you'd like to enable

- [x] Specific cloud provider or datacenter: AWS
- [ ] Dedicated Kubernetes cluster
- [x] Scalable Dask Cluster

### Other relevant information

GPU support in AWS is not yet available in our infra and it should be developed for this Hub.

### Hub URL

`TBD.TBD.2i2c.cloud`

### Hub Type

daskhub

### Tasks to deploy the hub

- [x] GPU setup on eksctl
- [x] S3 scratch bucket setup
- [x] GitHub Teams based authentication setup
- [x] Deploy on existing uwhackweeks cluster

yuvipanda:
I've updated the issue with the TODOs that need to happen.

I'm going to use the existing uwhackweeks infra to prototype GPU support as well the scratch bucket support.

damianavila:
> I've asked for an increase in quota on the uwhackweeks AWS account for GPU instances, and can proceed onces that comes through.

[...]

PR #1322, opened 2022-05-24
Add scratch bucket functionality for AWS

yuvipanda:
- Bump up AWS provider version, as there had been a few deprecations
  in the IAM resources
- Mimic the GCP setup as much as possible
- Write some docs on how to enable this for S3
- Setup IRSA with Terraform so users can be granted IAM roles
- Use the uwhackweeks to test

Ref https://github.com/2i2c-org/infrastructure/issues/1309

### DIFF
--- a/docs/howto/features/cloud-access.md
+++ b/docs/howto/features/cloud-access.md
@@ -28,10 +28,9 @@ to map a particular [Kubernetes Service Account](https://kubernetes.io/docs/task
 to a particular [AWS IAM Role](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles.html).
 All pods using the Kubernetes Service Account (user's jupyter notebook pods
 as well as dask worker pods)
-will have the permissions assigned to the Google Cloud Service Account.
-This Google Cloud Service Account is managed via terraform.
+will have the permissions assigned to the AWS IAM Role.
+This AWS IAM Role is managed via terraform.
 
-On
 
 (howto:features:cloud-access:access-perms)=
 ## Enabling specific cloud access permissions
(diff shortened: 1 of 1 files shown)


yuvipanda:
This is good to go I think!

yuvipanda:
@scottyhq can you test out the SCRATCH_BUCKET stuff on the uwhackweeks hub as well? This will make sure that when we deploy it for snowex it'll be all good.

scottyhq:
> can you test out the SCRATCH_BUCKET stuff on the uwhackweeks hub as well?  

Thanks! Just did a quick test with AWS CLI and things seem to work well! We'll see if it's gone in a week :) 
```bash
aws s3 cp scratch-bucket-test.txt $SCRATCH_BUCKET/ 
upload: ./scratch-bucket-test.txt to s3://uwhackweeks-scratch/scottyhq/scratch-bucket-test.txt
```

scottyhq:
By the way, i've found for hackweeks people often want to get a bunch of local data (laptop, labgroup computer) into the cloud and the scratch bucket is a pretty great way to kick the tires. 

A trick i figured out for last year's event is getting temporary session token from the hub: (copied from https://github.com/snowex-hackweek/jupyterhub/issues/9#issuecomment-875058825)
```bash
aws sts assume-role-with-web-identity \
 --role-arn $AWS_ROLE_ARN \
 --role-session-name $JUPYTERHUB_CLIENT_ID \
 --web-identity-token file://$AWS_WEB_IDENTITY_TOKEN_FILE \
 --duration-seconds 1000 > /tmp/irp-cred.txt

export AWS_REGION="us-west-2"
export AWS_ACCESS_KEY_ID="$(cat /tmp/irp-cred.txt | jq -r ".Credentials.AccessKeyId")"
export AWS_SECRET_ACCESS_KEY="$(cat /tmp/irp-cred.txt | jq -r ".Credentials.SecretAccessKey")"
export AWS_SESSION_TOKEN="$(cat /tmp/irp-cred.txt | jq -r ".Credentials.SessionToken")"
rm /tmp/irp-cred.txt
```

Do you think it's worth documenting this in https://docs.2i2c.org/en/latest/user/storage.html?highlight=scratch#scratch-bucket ? It'd be pretty cool if there were a file menu to 'get session token' instead of a hacky script, do you know of something like this ?

damianavila:
@yuvipanda, do you have any thoughts about this one? I presume the permissions you get with these temporary credentials only give you access to the scratch bucket on S3 and anything else, am I correct?

scottyhq:
>  It'd be pretty cool if there were a file menu to 'get session token' instead of a hacky script, do you know of something like this ?

@yuvipanda and I (mostly Yuvi :) ) put together this proof of concept jupyter server extension to give Hub users temporary credentials via an API endpoint. It's still a hacky script behind the scenes but works! I probably won't work on it much in the near future but wanted to link here in case people want to develop it further and try out: https://github.com/scottyhq/jupyter-cloud-scoped-creds

Issue #1279, opened 2022-05-05
Metrics and Reporting for LEAP Hub


rabernat:
### Context

In https://github.com/2i2c-org/infrastructure/issues/1050#issuecomment-1064450254 I listed several important requirements for the new LEAP hub. That issue has been closed with some of those points unresolved, so I am raising this issue to continue tracking.

I cannot overemphasize **how concerned the LEAP leadership are about runaway cloud costs**. 2i2c is holding all of our Google Credits, which means we are placing a huge amount of trust in 2i2c to steward this resource on behalf of our project. LEAP leadership is concerned with the following scenarios:
- Illegitimate use of the hub (e.g. bitcoin mining)
- Disproportionate resource use by individuals, leading to unexpected charges
- Too quick burn rate of credits, meaning we would run out before 5 years are up

We need a plan to address these concerns before LEAP leadership are comfortable opening up the hub to users. So therefore, this issue has the most urgent priority for LEAP, above all other ongoing technical development.

### Proposal

I propose that 2i2c build a reporting system to deliver the following information in the form of a **weekly email** and / or **interactive dashboard**.

For each of the following breakdowns:
- Project total
- By GitHub group (group membership used to authorize LEAP users)
- By individual users

Report the following information:
- Total CPU compute cost consumed
- Total GPU compute cost consumed
- Total storage cost consumed

For the following periods
- weekly total
- monthly total
- net total to date

A LEAP person (probably me to start) will review this report on a weekly basis to ensure that no anomalous costs are occurring. If anomalous costs are found, these reports will allow us to trace them to specific users and intervene.

Additionally, we would like to request _a standing monthly meeting between LEAP and a 2i2c representative to review costs_. 

---

I understand that these reports are not currently part of 2i2c services. I see three possible ways forward:
- You can start building this reporting system now without any additional overhead.
- You could build this reporting system _if we pay more money_.
- You will never build this reporting system.

It would be great if someone could identify which of these pathways is most likely. I need to convey a response to the LEAP leadership asap.

---

Linked issues:
- https://github.com/2i2c-org/infrastructure/issues/730
- https://github.com/2i2c-org/infrastructure/issues/328

yuvipanda:
Since the goal is to avoid overrunning costs, I think a progressive approach would be:

1. Figure out a weekly report that just lists total cloud spend *that week*
2. Have tools in place that let us investigate usage when we need to.

[...]

PR #1323, opened 2022-05-24
Add the username as label to dask worker and dask scheduler pods

GeorgianaElena:
This just copies @yuvipanda's code in https://github.com/yuvipanda/pangeo-cloud-federation/commit/fdda146c714cdf932fc18d38835eaca33f68fa2f

The notebook pods already have this label, added by kubespawaner

Reference https://github.com/2i2c-org/infrastructure/issues/1279


yuvipanda (approved):
yay!

PR #1331, opened 2022-05-25
Properly escape usernames and use them for dask pod labels

GeorgianaElena:
Second try to https://github.com/2i2c-org/infrastructure/pull/1323


helm-charts/daskhub/values.yaml
```diff
@@ -157,20 +157,44 @@ dask-gateway:
     extraConfig:
       optionHandler: |
         from dask_gateway_server.options import Options, Integer, Float, String, Mapping
+        import string
+
+        # Escape a string to be dns-safe in the same way that KubeSpawner does it.
+        # Reference https://github.com/jupyterhub/kubespawner/blob/main/kubespawner/spawner.py#L1828-L1835
```
consideRatio, line 163:
```suggestion
        # Reference https://github.com/jupyterhub/kubespawner/blob/616f72c4aee26c3d2127c6af6086ec50d6cda383/kubespawner/spawner.py#L1828-L1835
```

I made this a permalink as otherwise this comment may point to the wrong lines over time.

image

GeorgianaElena: Perfect! Thank you! ✨

consideRatio (approved):
Perfect!

PR #1332, opened 2022-05-25
Add dependabot config for pip

sgibson91:
This PR updates our dependabot config to also check our pip dependencies


choldgraf (approved):
I'm +1 on giving it a shot ! In my experience the pip dependency detector is a bit unpredictable but I may have configured things wrong

sgibson91:
I will merge this and then force a run so we are up-to-date :)

PR #1335, opened 2022-05-26
Add snowex hub to uwhackweeks cluster

yuvipanda:
- Moves current hackweeks hub (which is really icesat hackweek hub)
  config out of common and into staging / prod.yaml
- Add new config for snowex hackweek
- Add scratch bucket for snowex hackweek

Ref https://github.com/2i2c-org/infrastructure/issues/1309


sgibson91 (approved):
❄️

Issue #1329, opened 2022-05-25
[New Hub] Jack Eddy Symposium


damianavila:
### Hub Description

The request is for the launch or use of a research hub with Dask. (Daniel Marsh, one of the co-organizers, wishes to share tutorial notebooks. He plans to use "intake-esm to access all the CMIP6 climate runs which are hosted in zarr format on aws. We can pull down all the time series data and regress it against the solar forcing used to derive the solar response."

### Community Representative(s)

@colliand, can you give us the contact information for the community representatives.
Looking at the [lead issue](https://github.com/2i2c-org/leads/issues/77), it seems the contacts would be Daniel Marsh and/or Ryan McGranaghan. If that is the case, do you have their contact information? Any GitHub handles?

### Important dates

- **Required start date**: Jun 1st
- **Target start date**: ASAP
- **Any important dates for usage**: 
  - Start date for the event: June 6
  - End date for the event: June 10
  - The active times for the event (e.g., 9am to 5pm US/Pacific) 8 AM MT - 6 PM MT
  - How many people will attend the event? ~100-150
  - Do you need to need hub infrastructure to be [pre-initialized before the event] **Not a requirement, but a nice-to-have**

### Hub Authentication Type

Other (may not be possible, please specify in comments)

### Hub logo information

- **URL to Hub Image**: https://cpaess.ucar.edu/sites/default/files/meetings/2022/nasa-jack-eddy-symposium-2022-860.jpg
- **URL for Image Link**: {{ URL HERE }}

### Hub user image

- Repository for user image: { REPO LINK IF IT EXISTS }
- User image registry: { REGISTRY IF ONE ALREADY EXISTS }
- User image tag and name: { NAME AND TAG IF IT EXISTS }

### Extra features you'd like to enable

- [x] Specific cloud provider or datacenter: AWS
- [x] Dedicated Kubernetes cluster
- [x] Scalable Dask Cluster

### Other relevant information

I presume GitHub auth would be OK but we need to confirm it.
It seems they might need to interact with datasets on AWS, so it might make sense to deploy in that cloud provider...
From the lead description, it is not clear to me how this hub will be paid so there is some stuff to figure it out in case we need a new AWS land to deploy into...

### Hub URL

jackeddy.2i2c.cloud

### Hub Type

daskhub

### Tasks to deploy the hub

- [x] Engineer who will deploy the hub is assigned
- [x] Deploy information filled in above
- [x] Initial Hub deployment PR: <link to PR>
[...]

PR #1337, opened 2022-05-26
Add jackeddy hub

yuvipanda:
- Put in GCP as the CMIP6 data desired is on GCP as well.
  However, I thin GCP's CMIP6 dataset is more limited than AWS' -
  we cna move this to AWS if needed.
- SCRATCH_BUCKET is setup
- Secure push to GitHub via gh-scoped-creds is setup
- GitHub teams is used for auth
- Default memory limits are 1Gi Limit + 256Mi Guarantee. I *think*
  this won't be enough - let's customize this as needed.
- I tried to use https://gallery.ecr.aws/q3h7b4o8/helio-notebook-py
  built from https://git.mysmce.com/heliocloud/heliocloud-docker-images,
  but it doesn't work - primarily because it has JupyterHub 2.0 in
  there, and we're still at 1.5.  Need to debug that. In the meantime,
  just using the default pangeo image.

Ref https://github.com/2i2c-org/infrastructure/issues/1329

### DIFF
--- a/config/clusters/2i2c/cluster.yaml
+++ b/config/clusters/2i2c/cluster.yaml
@@ -166,3 +166,12 @@ hubs:
       # to the helm upgrade command in, and that has meaning. Please check
       # that you intend for these files to be applied in this order.
       - utexas.values.yaml
+  - name: jackeddy
+    display_name: "Jack Eddy Symposium"
+    domain: jackeddy.2i2c.cloud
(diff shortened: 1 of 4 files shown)


yuvipanda:
Deployed and ready for testing at https://jackeddy.2i2c.cloud/

sgibson91 (approved):

PR #1339, opened 2022-05-30
Automate bumping of Pangeo image tags

sgibson91:
Add workflow file to bump image tags for pangeo

### DIFF
--- a/.github/workflows/bump-image-tags.yaml
+++ b/.github/workflows/bump-image-tags.yaml
@@ -3,7 +3,7 @@ name: Bump Image Tags
 on:
   workflow_dispatch:
   schedule:
-    - cron: "0 0 * * 1"  # Run at 00:00 UTC every Monday
+    - cron: "0 0 * * 1" # Run at 00:00 UTC every Monday
 
 env:
   team_reviewers: 2i2c-org:tech-team
--- a/.github/workflows/bump-image-tags.yaml
+++ b/.github/workflows/bump-image-tags.yaml
@@ -23,9 +23,11 @@ jobs:
             values_paths: ".basehub.jupyterhub.singleuser.image"
           - name: "m2lines common singleuser/PyTorch/Tensorflow"
             config_path: "config/clusters/m2lines/common.values.yaml"
+            # If the ordering of profileList changes, update the index in this expression
             values_paths: ".basehub.jupyterhub.singleuser.image .basehub.jupyterhub.singleuser.profileList[4].kubespawner_override.image .basehub.jupyterhub.singleuser.profileList[5].kubespawner_override.image"
           - name: "leap common singleuser/pangeo-ml-notebook"
             config_path: "config/clusters/leap/common.values.yaml"
+            # If the ordering of profileList changes, update the index in this expression
             values_paths: ".basehub.jupyterhub.singleuser.image .basehub.jupyterhub.singleuser.profileList[4].kubespawner_override.image"
 
     steps:
--- a/.github/workflows/bump-image-tags.yaml
+++ b/.github/workflows/bump-image-tags.yaml
@@ -21,6 +21,12 @@ jobs:
           - name: "pangeo-hubs common singleuser image"
             config_path: "config/clusters/pangeo-hubs/common.values.yaml"
             values_paths: ".basehub.jupyterhub.singleuser.image"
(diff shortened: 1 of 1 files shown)


.github/workflows/bump-image-tags.yaml
```diff
@@ -0,0 +1,43 @@
...
+        include:
+          # For each new config_path to monitor, add a new item in this matrix.
+          # The Action can read multiple paths to images, but not multiple config files.
+          - name: "pangeo-hubs common singleuser image"
+            config_path: "config/clusters/pangeo-hubs/common.values.yaml"
+            values_paths: ".basehub.jupyterhub.singleuser.image"
+
+    steps:
+      # We want tests to be run on the Pull Request that gets opened by the next step,
+      # so we generate a token from a GitHub App that would allow this.
```
sgibson91, line 27:
```suggestion
      # so we generate a token from a GitHub App that would allow this.
      By default, secrets.GITHUB_TOKEN is prevented from triggering
      other workflows.
```

.github/workflows/bump-image-tags.yaml
```diff
@@ -0,0 +1,45 @@
+name: Bump Image Tags
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * 1" # Run at 00:00 UTC every Monday
+
+env:
+  team_reviewers: 2i2c-org:tech-team
```
sgibson91, line 9:
```suggestion
  team_reviewers: 2i2c-org/tech-team
```

.github/workflows/bump-image-tags.yaml
```diff
@@ -0,0 +1,51 @@
...
+      #
+      # Action: https://github.com/marketplace/actions/github-app-token
+      - name: Fetch a token from GitHub App
+        id: generate_token
+        uses: tibdex/github-app-token@v1.5
+        with:
+          app_id: ${{ secrets.APP_ID }}
+          private_key: ${{ secrets.PRIVATE_KEY }}
+
+      - name: "Bump image tags: ${{ matrix.name }}"
```
sgibson91, line 45:
```suggestion
      # Action repo: https://github.com/sgibson91/bump-jhub-image-action
      - name: "Bump image tags: ${{ matrix.name }}"
```

.github/workflows/bump-image-tags.yaml
```diff
@@ -0,0 +1,52 @@
...
+      matrix:
+        include:
+          # For each new config_path to monitor, add a new item in this matrix.
+          # The Action can read multiple paths to images, but not multiple config files.
+          - name: "pangeo-hubs common singleuser image"
+            config_path: "config/clusters/pangeo-hubs/common.values.yaml"
+            values_paths: ".basehub.jupyterhub.singleuser.image"
+          - name: "m2lines common singleuser/PyTorch/Tensorflow"
+            config_path: "config/clusters/m2lines/common.values.yaml"
+            values_paths: ".basehub.jupyterhub.singleuser.image .basehub.jupyterhub.singleuser.profileList[4].kubespawner_override.image .basehub.jupyterhub.singleuser.profileList[5].kubespawner_override.image"
```
yuvipanda, line 26: Can you add a comment to the values files here that if the ordering changes, we should also update the files here?

Separately, a feature request for the upstream action is more native support for profile_list, where it can then read image_override and do the bumps accordingly :)

yuvipanda (approved):
one suggestion for comment pointers, and one upstream feature request. Once the comment is done, please do merge!

.github/workflows/bump-image-tags.yaml
sgibson91: So we can do `...profileList[*]` which will return everything under the profileList key, but then we would need Python code to loop through each item and check that the kubespawner_override key is present and that in turn has the image key present. It's doable; I just don't know which is better: Relying on "magic" that the code will just "figure it out" or providing explicit/specific input? If we made this change and something went wrong with the profileList bumping, then the user would have to debug both their JMESPath query _and_ the source code. Whereas currently the only mistake that can be made right now is the user provides an incorrect JMESPath query.

sgibson91: > 'anything with an image key under profileList'

I'll actually dig into this a little bit more

Issue #1336, opened 2022-05-26
[Decommission Hub] uwhackweeks.2i2c.cloud


yuvipanda:
### Summary
Based on https://github.com/2i2c-org/infrastructure/issues/1309#issuecomment-1136277855, we should turn down the ICESat Hackweek hub at uwhackweeks.2i2c.cloud on June 1.

### Info

- **Community Representative:** 
- **Link to New Hub issue:** 
- **Proposed end date:** June 1

#### Phase I

- [x] Confirm with Community Representative that the hub is no longer in use and it's safe to decommission
- [x] Confirm if there is any data to migrate from the hub before decommissioning
  - [ ] If yes, confirm where the data should be migrated to
  - [ ] Confirm a 2i2c Engineer has access to the destination in order to complete the data migration

#### Phase II - Hub Removal

- [ ] (Optional) Migrate data from the hub
- [x] Remove the appropriate `config/clusters/<cluster_name>/<hub_name>.values.yaml` file
- [x] Remove the associated hub entry from the `config/clusters/<cluster_name>/cluster.yaml file.
- [x] Remove the hub deployment
  - `helm --namespace HUB_NAME delete HUB_NAME`
  - `kubectl delete namespace HUB_NAME`
- [x] Delete the hub's authentication application on auth0 or GitHub

yuvipanda:
@scottyhq just to confirm that on June 1, I'll turn off the hub at uwhackweeks.2i2c.cloud *and* delete all the user home directories.

scottyhq:
>  confirm that on June 1, I'll turn off the hub at uwhackweeks.2i2c.cloud and delete all the user home directories. 

That is correct!

scottyhq:
One other request @yuvipanda it would be great to have a plot / summary of the total cloud costs for this event (basically this figure extended through June 1 https://github.com/2i2c-org/infrastructure/issues/1142#issuecomment-1078544468). Cheers!

GeorgianaElena:
The hub was deleted and the user home dirs deleted 🚀 Closing this.

PR #1386, opened 2022-06-06
Remove uwhackweeks.2i2c.cloud

GeorgianaElena:
- Removes config of the prod uwhackweeks.2i2c.cloud hub
- I kept the staging to help with testing the snowex one. Is this ok?
- I believe we also need to remove the scratch bucket for this hub, right? (we should add this to the decommission steps if so) https://github.com/2i2c-org/infrastructure/blob/16bd879f40c0cc7f81da9629c98368c80cac42de/terraform/aws/projects/uwhackweeks.tfvars#L11-L13

Manual todos:
- [ ] delete home dirs
- [ ] delete github oauth app
- [ ] Remove the hub deployment
      - `helm --namespace uwhackweeks delete prod`
      - `kubectl delete namespace prod`

Ref: https://github.com/2i2c-org/infrastructure/issues/1336


yuvipanda (approved):
Yay! We'll also need to delete the home directories tho. I think keeping the staging hub is fine. I think what was at 'uwhackweeks.2i2c.cloud' is really the ICESat hackweek hub, and we'll run more of these in the future here.

GeorgianaElena:
I'm currently deleting the home dirs of the users

damianavila:
> I'm currently deleting the home dirs of the users

@GeorgianaElena, can you confirm when this is done so I push this one to the "Done" column? Thanks!

GeorgianaElena:
Home dirs are gone 🚀

PR #1388, opened 2022-06-06
Increase prometheus server limits for the 2i2c cluster

GeorgianaElena:
The `prometheus-server` pod gets `OOMKilled` currently so we don't have dashboards for the 2i2c cluster.

Currently, the sizes are:

```
Limits:
      cpu:     1
      memory:  2Gi
    Requests:
      cpu:        200m
      memory:     512Mi
```

@damianavila, please feel free to merge this is you think it's LGTM to use the grafana dashbords today


damianavila:
Thanks for taking care of this one @GeorgianaElena!

PR #1390, opened 2022-06-07
Enable 0 user placeholders when using continuous prePuller

yuvipanda:
Without this, looks like prePuller tries to use the priority
set for placeholder and the daemonset fails.

Daemonset fails with:

> Error creating: pods "continuous-image-puller-" is forbidden: no PriorityClass with name jackeddy-user-placeholder-priority was found

### DIFF
--- a/config/clusters/2i2c/jackeddy.values.yaml
+++ b/config/clusters/2i2c/jackeddy.values.yaml
@@ -110,7 +110,8 @@ basehub:
             readOnly: true
     scheduling:
       userPlaceholder:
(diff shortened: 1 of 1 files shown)


sgibson91 (approved):

sgibson91:
If you rebase your branch against master, the generate-jobs CI/CD job should work

damianavila:
I feel this is probably a good candidate for the troubleshooting section: https://infrastructure.2i2c.org/en/latest/topic/troubleshooting.html

PR #1394, opened 2022-06-07
[BUGFIX] Add environment so secrets are accessible

sgibson91:
This ensures the bump helm versions workflow runs in the correct environment with access to the required secrets


sgibson91:
Minor change so will self-merge

PR #1397, opened 2022-06-07
[BUGFIX] Fix team name for review requests

sgibson91:
This fixes a bug where a review from `2i2c-org/tech-team` was not being requested. This is because the API expects only the `team_slug` which (after a lot of unhelpful documentation and defaulting to experimentation) is apparently _just_ `tech-team`, not `2i2c-org/tech-team`.


sgibson91:
Will self-merge as this is a small change.

Issue #1319, opened 2022-05-18
Move to using an OAuthenticator release


GeorgianaElena:
### Context

We're now using an unreleased version of OAuthenticator and pinning to this commit on the main branch: https://github.com/jupyterhub/oauthenticator/commit/60ddc4fe3ab06b19c6876992679c1fd2ec1c64dd.

### Proposal

We should the latest OAuthenticator version that has the commit mentioned above once that gets released.

### Updates and actions

- [x] help out with the release of a new OAuthenticator version -> https://github.com/jupyterhub/team-compass/issues/518
- [ ] use that version here

PR #1403, opened 2022-06-08
Bump oauthenticator

GeorgianaElena:
This bumps the oauthenticator version to 15.0.0. However, doing this revealed some shortcomings and this PR contains some workarounds for those.

### OAuthenticator

Some upstream issues that are relevant and provide more info:

- https://github.com/jupyterhub/oauthenticator/issues/514
- https://github.com/jupyterhub/oauthenticator/issues/515

### CILogon

There is a possible issue with CILogon using an old OIDC endpoint for google (and open an issue upstream about this) (Old endpoint is http://google.com/accounts/o8/id, and new endpoint should be https://accounts.google.com/o/oauth2/v2/auth).

Look how for shown idps, the correct url is https://github.com/2i2c-org/infrastructure/blob/6debc6f283276922cc9d69fb40e34ce8694a31b0/config/clusters/2i2c/anu.values.yaml#L42

But the idp field of the `userinfo` response
https://github.com/jupyterhub/oauthenticator/blob/f3af268474ecce4fc2a9eb789c2e06cd692375fa/oauthenticator/cilogon.py#L306

is an old endpoint https://github.com/2i2c-org/infrastructure/blob/6debc6f283276922cc9d69fb40e34ce8694a31b0/config/clusters/2i2c/anu.values.yaml#L44

Opened an issue upstream about this one. https://github.com/ncsa/OA4MP/issues/45

### TODO list
- [x] use oauthenticator 15.0.1 release instead of 15.0.0
Deploy and test these changes on:
- [x] 2i2c-staging
- [x] 2i2c-demo
- [x] leap-staging
- [x] utoronto-staging
Maybe:
- [x] 2i2c-anu
- [x] 2i2c-dask-staging

Closes https://github.com/2i2c-org/infrastructure/issues/1319

### DIFF
--- a/config/clusters/2i2c/anu.values.yaml
+++ b/config/clusters/2i2c/anu.values.yaml
@@ -51,4 +51,4 @@ jupyterhub:
         # We only want 2i2c and anu users to sign up
         username_pattern: '^(.+@2i2c\.org|.+@anu\.edu\.au|deployment-service-check)$'
         # Delete any prior existing users in the db that don't pass username_pattern
-        # delete_invalid_users: true
+        delete_invalid_users: true
--- a/config/clusters/2i2c/anu.values.yaml
+++ b/config/clusters/2i2c/anu.values.yaml
@@ -46,7 +46,7 @@ jupyterhub:
               username_claim: "email"
           https://idp2.anu.edu.au/idp/shibboleth:
             username_derivation:
-            username_claim: "email"
+              username_claim: "email"
       Authenticator:
         # We only want 2i2c and anu users to sign up
         username_pattern: '^(.+@2i2c\.org|.+@anu\.edu\.au|deployment-service-check)$'
--- a/config/clusters/2i2c/anu.values.yaml
+++ b/config/clusters/2i2c/anu.values.yaml
@@ -48,7 +48,7 @@ jupyterhub:
             username_derivation:
             username_claim: "email"
       Authenticator:
-        # We only want 2i2c and any users to sign up
+        # We only want 2i2c and anu users to sign up
         username_pattern: '^(.+@2i2c\.org|.+@anu\.edu\.au|deployment-service-check)$'
         # Delete any prior existing users in the db that don't pass username_pattern
         # delete_invalid_users: true
--- a/config/clusters/2i2c/anu.values.yaml
+++ b/config/clusters/2i2c/anu.values.yaml
@@ -44,23 +44,9 @@ jupyterhub:
           http://google.com/accounts/o8/id:
               username_derivation:
                 username_claim: "email"
-                action: "strip_idp_domain"
-                # Use a hack here to allow leaving usernames unchanged.
-                # https://github.com/jupyterhub/oauthenticator/blob/42bebe989ec10e8a439bc80a011fbe64cc01669f/oauthenticator/cilogon.py#L341
-                # Since we're only allowing 2i2c.org domains and anu.edu.au, then this will fail
-                # and usernames will be left unchanged.
-                # FIXME when https://github.com/jupyterhub/oauthenticator/issues/514 lands.
-                domain: "gmail.com"
           https://idp2.anu.edu.au/idp/shibboleth:
                 username_derivation:
                 username_claim: "email"
-                action: "strip_idp_domain"
-                # Use a hack here to allow leaving usernames unchanged.
-                # https://github.com/jupyterhub/oauthenticator/blob/42bebe989ec10e8a439bc80a011fbe64cc01669f/oauthenticator/cilogon.py#L341
-                # Since we're only allowing 2i2c.org domains and anu.edu.au, then this will fail
-                # and usernames will be left unchanged.
-                # FIXME when https://github.com/jupyterhub/oauthenticator/issues/514 lands.
-                domain: "gmail.com"
       Authenticator:
         # We only want 2i2c and any users to sign up
         username_pattern: '^(.+@2i2c\.org|.+@anu\.edu\.au|deployment-service-check)$'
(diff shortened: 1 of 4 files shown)


helm-charts/images/hub/Dockerfile
```diff
@@ -18,9 +18,8 @@ RUN pip install --no-cache git+https://github.com/yuvipanda/jupyterhub-configura
 # z2jh 1.2.x ships with kubespawner 1.1.0, so we just do a little bump
 RUN pip install --no-cache --upgrade jupyterhub-kubespawner==1.1.2
 
-# Brings in https://github.com/jupyterhub/oauthenticator/pull/498 and
-# https://github.com/jupyterhub/oauthenticator/pull/504
-RUN pip install --no-cache --upgrade git+https://github.com/jupyterhub/oauthenticator@60ddc4fe3ab06b19c6876992679c1fd2ec1c64dd
+# Latest version comes with some breaking changes https://oauthenticator.readthedocs.io/en/latest/migrations.html#migrating-cilogonoauthenticator-to-version-15-0-0
```
GeorgianaElena, line 21:
```suggestion
# Latest version comes with some breaking changes https://oauthenticator.readthedocs.io/en/latest/migrations.html#migrating-cilogonoauthenticator-to-version-15-0
```

config/clusters/2i2c/anu.values.yaml
```diff
@@ -34,10 +34,21 @@ jupyterhub:
...
+            username_derivation:
+              username_claim: "email"
+          https://idp2.anu.edu.au/idp/shibboleth:
+            username_derivation:
+            username_claim: "email"
+      Authenticator:
+        # We only want 2i2c and anu users to sign up
+        username_pattern: '^(.+@2i2c\.org|.+@anu\.edu\.au|deployment-service-check)$'
+        # Delete any prior existing users in the db that don't pass username_pattern
+        # delete_invalid_users: true
```
GeorgianaElena, line 54:
```suggestion
        delete_invalid_users: true
```

Without this, hub logs contain errors about previous usernames being in the database and not matching the regex. For example, from when we were using our @gmail.com addresses instead of our 2i2c ones. I believe it's a good rule of thumb to delete users that are not valid anymore?

Error from logs is:
```
[E 2022-06-09 15:16:08.554 JupyterHub app:1884] Error adding user georgiana.dolocan@gmail.com already in db
      File "/usr/local/lib/python3.8/dist-packages/jupyterhub/app.py", line 1880, in init_users
        f = self.authenticator.add_user(user)
      File "/usr/local/lib/python3.8/dist-packages/jupyterhub/auth.py", line 619, in add_user
        raise ValueError("Invalid username: %s" % user.name)
    ValueError: Invalid username: georgiana.dolocan@gmail.com
    
[W 2022-06-09 15:16:08.554 JupyterHub app:1891] 
    You can set
        c.Authenticator.delete_invalid_users = True
    to automatically delete users from the Hub database that no longer pass
    Authenticator validation,
    such as when user accounts are deleted from the external system
    without notifying JupyterHub.
```

GeorgianaElena:
This should now be ready for review. See the initial comment about follow-up issues. Also, I left a comment inline about deleting invalid users when using `username_pattern` that would love your thoughts on.

config/clusters/2i2c/anu.values.yaml
sgibson91: +1

helm-charts/images/hub/Dockerfile
```diff
@@ -18,9 +18,8 @@ RUN pip install --no-cache git+https://github.com/yuvipanda/jupyterhub-configura
 # z2jh 1.2.x ships with kubespawner 1.1.0, so we just do a little bump
 RUN pip install --no-cache --upgrade jupyterhub-kubespawner==1.1.2
 
-# Brings in https://github.com/jupyterhub/oauthenticator/pull/498 and
-# https://github.com/jupyterhub/oauthenticator/pull/504
-RUN pip install --no-cache --upgrade git+https://github.com/jupyterhub/oauthenticator@60ddc4fe3ab06b19c6876992679c1fd2ec1c64dd
+# Latest version comes with some breaking changes https://oauthenticator.readthedocs.io/en/latest/migrations.html#migrating-cilogonoauthenticator-to-version-15-0
+RUN pip install --no-cache --upgrade oauthenticator==15.0.1
```
sgibson91, line 22: What is the effect of having the `--upgrade` flag but a strictly pinned package here? I feel like either the flag has no effect and maybe we should remove it, or it could have strange effects? Hmmm....

sgibson91 (approved):
LGTM - I'm +1 on deleting invalid users from the hub db and have a pip-related question.

helm-charts/images/hub/Dockerfile
GeorgianaElena: Hmm, I think you're right and this flag might be not be needed in this case of pinned pkgs. But I don't believe it does any harm, but rather does nothing at the moment.

I think we can remove it. The only reason I can think of that would be in favor of keeping it is if we ever decide to not use a pinned version of these pkgs and we forget to add the flag, but both things seem unlikely.

Do you think I should remove it as part of this PR? (it's also used when installing `kubespawner` so that would need to be removed too) or in its own thing?

sgibson91: >The only reason I can think of that would be in favor of keeping it is if we ever decide to not use a pinned version of these pkgs and we forget to add the flag

I am +1 on this.

I am fine keeping the flag if we are confident it does no harm. But just wanted to check :)

GeorgianaElena:
I've deployed this change to all of the hubs mentioned in the top comment and I was able to successfully login and start a server on those afterwards, so merging it 🚀 Thank you ✨

PR #1408, opened 2022-06-09
Bugfixing the bump image tags workflow

sgibson91:
- An extra quotation mark caused a json.load failure
- Assigning team reviewers is currently broken, so disable that for now


sgibson91:
Will self-merge since this is a bugfix

PR #1417, opened 2022-06-14
Create support link that will appear on new issue page

sgibson91:
We get a few questions opened as issues on this repo instead of being sent to our support email. This PR adds a config file under ISSUE_TEMPLATE which will add a new option to our "New Issues" page that redirects to our support docs. Inspired by the work @GeorgianaElena did here: https://github.com/jupyterhub/.github/blob/master/.github/ISSUE_TEMPLATE/config.yml 💖 Thank you!

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yaml
+++ b/.github/ISSUE_TEMPLATE/config.yaml
@@ -1,0 +1,4 @@
+contact_links:
+  - name: Looking for Support?
+    url: https://docs.2i2c.org/en/latest/support.html
+    about: Ask for help through our support process


sgibson91:
Another topic would be if we could automate something to send a similar message to Discussions, e.g., https://github.com/2i2c-org/infrastructure/discussions/1416

(I would actually be +1 on turning off Discussions for this repo as I don't think we use them, but probably a good topic to discuss at a team meeting.)

GeorgianaElena (approved):
LGTM :rocket:

sgibson91:
I apologise for not doing the work in finding a suitable emoji 😂

choldgraf:
love it 👍

PR #1420, opened 2022-06-15
Enable cryptnono on all our clusters

yuvipanda:
https://github.com/yuvipanda/cryptnono/ is now deployed
on mybinder.org. We've already had a few incidents with
cryptomining (OpenScapes, cloudbank, pangeo binder). This
should help with that.

### DIFF
--- a/helm-charts/support/Chart.yaml
+++ b/helm-charts/support/Chart.yaml
@@ -35,3 +35,10 @@ dependencies:
     version: 1.3.1
     repository: https://kvaps.github.io/charts
     condition: nfs-server-provisioner.enabled
+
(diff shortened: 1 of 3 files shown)


choldgraf:
We should def document this, it is a big feature IMO ! Is that ok or would that not be a good idea from a "security through obscurity POV"?

yuvipanda:
@choldgraf we should definitely document this!

choldgraf:
I gave a first shot at this here:

- https://github.com/2i2c-org/docs/pull/147

would love feedback!

damianavila:
Let's keep this one in the back of our heads in case it is related to the #1423 incident.

PR #1421, opened 2022-06-15
Bump up cryptnono new versions as they come

yuvipanda:


sgibson91:
https://github.com/sgibson91/bump-helm-deps-action/pull/102 ensures that this works! 🎉

PR #1424, opened 2022-06-16
Post hub deployment plan as a comment on PRs

### LINKED ISSUE #1269: When generate-jobs is run in a PR, add the rich table to the PR as a comment
https://github.com/2i2c-org/infrastructure/blob/2af57d0e3d1aa2ec98e69504338fda84899d6506/.github/workflows/deploy-hubs.yaml#L93-L95

### Benefit

Adding the table of hubs that will be upgraded to a PR raises the visibility of what effect merging a PR may have on live infrastructure (engineers won't have to go looking for it in the actions logs) and will help engineers reason about whether the correct infrastructure is being touched by their changes.

### Things to consider

- The comment should be done in a "Netlify way", i.e., if a comment already exists on the PR, update it instead of posting a new one
- If the table is empty (i.e. the two JSON outputs are `[]`), we should not post a comment at all to reduce noise

sgibson91:
fixes #1269 

This PR adds a workflow job that will post a table of the hub deployment plan (which clusters will have their support chart and/or staging hub upgraded, and which prod hubs will be upgraded) as a comment to a Pull Request that triggers the deploy-hubs.yaml workflow file. This will make it obvious which hubs, if any, will be upgraded when the PR is merged, without having to go through the GitHub Actions logs.

* The comment is posted in a "Netlify way", i.e., update an existing comment rather than posting a new one, and doesn't post a comment at all if no matrix jobs are generated
* Unfortunately, rich didn't have a nice way to get a Markdown table from what is rendered by rich (see https://github.com/Textualize/rich/discussions/1254). I first experimented with the [text](https://github.com/sgibson91/testing-gh-actions/pull/19#issuecomment-1154342947) and html output, but they weren't formatted very nicely. So I ended up using a new package called py-markdown-table to get something that will render nicely in comments.
* There are some changes to how the dictionaries are constructed in the Python file. This is a noop change, but ensures that when the dictionaries are rendered as tables, the order of the columns (or dictionary keys) makes sense to a human reading the table.

### DIFF
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -46,12 +46,14 @@ jobs:
   # hub upgraded; and the second set of jobs describe which production hubs require
   # upgrading. These two lists are set as job outputs using ::set-output to be consumed
   # by the later jobs. They are also pretty-printed in a human-readable format to the
-  # logs.
+  # logs, and converted into Markdown tables for posting into GitHub comments.
   generate-jobs:
     runs-on: ubuntu-latest
     outputs:
       support-and-staging-matrix-jobs: ${{ steps.generate-jobs.outputs.support-and-staging-matrix-jobs }}
       prod-hub-matrix-jobs: ${{ steps.generate-jobs.outputs.prod-hub-matrix-jobs }}
+      md-table-support-staging: ${{ steps.generate-jobs.outputs.md-table-support-staging }}
+      md-table-prod: ${{ steps.generate-jobs.outputs.md-table-prod }}
 
     steps:
       - name: Checkout repo
@@ -91,9 +93,6 @@ jobs:
           format: space-delimited
           token: ${{ secrets.GITHUB_TOKEN }}
 
-      # TODO: Post the pretty print output of this step as a comment on the Pull Request.
-      # Do this in a Netlify-style way where if a comment already exists, it is updated
-      # instead of generating a new comment.
       - name: Generate matrix jobs
         id: generate-jobs
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy-hubs.yaml
```diff
@@ -288,3 +288,111 @@ jobs:
...
+  # variables.
+  post-deployment-plan:
+    # Only run this job on Pull Requests and when the job matrices are not empty
+    if: github.event_name == 'pull_request' &&
+      needs.generate-jobs.outputs.support-and-staging-matrix-jobs != '[]' &&
+      needs.generate-jobs.outputs.prod-hub-matrix-jobs != '[]'
+    # Grant GITHUB_TOKEN just enough permissions to read issues and comment on pull requests
+    permissions:
+      issues: read
+      pull-requests: write
```
sgibson91, line 311:
```suggestion
    # Grant GITHUB_TOKEN just enough permissions to check issues and comment on pull requests
    permissions:
      issues: write
      pull-requests: write
```

sgibson91:
Currently debugging why posting the comment doesn't work with a "Resource not accessible by integration" error when the same workflow [over here](https://github.com/2i2c-org/infrastructure/blob/master/.github/workflows/comment-test-link-merged-pr.yaml) works without issue

sgibson91:
So since I couldn't figure out why using GITHUB_TOKEN wasn't working (even though it should and works _totally fine_ for [this workflow](https://github.com/2i2c-org/infrastructure/blob/master/.github/workflows/comment-test-link-merged-pr.yaml)!!!), I switched to using the token generator app instead.

In [`12bd857` (#1424)](https://github.com/2i2c-org/infrastructure/pull/1424/commits/12bd8575999c185956704ce70828f7135f59342d), I ported over the step that would generate the token and pass it to the Python script and set the environment to the one that contains the appropriate secrets for the generator app. However, the Actions are saying the required inputs are not being provided to that step... even though they are?!

I have come too the conclusion that it is too hot for GitHub Actions to be working correctly right now. Will come back to this later when I am less mad.

sgibson91:
Actually, this may be because the PR is opened from a fork 😕

sgibson91:
Yep - confirmed this doesn't work across forks: https://github.com/sgibson91/testing-gh-actions/pull/22

.github/workflows/deploy-hubs.yaml
```diff
@@ -92,14 +92,32 @@ jobs:
...
-      # instead of generating a new comment.
+      # This step will create a comment-body.txt file containing the jobs to be run in a
+      # Markdown table format to be posted on a Pull Request, if this job is triggered
+      # by one
       - name: Generate matrix jobs
         id: generate-jobs
         run: |
           python deployer generate-helm-upgrade-jobs "${{ steps.changed-files.outputs.added_modified }}"
 
+      # The comment-deploy-ment-plan-pr.yaml workflow won't have the correct context to
```
sgibson91, line 103:
```suggestion
      # The comment-deployment-plan-pr.yaml workflow won't have the correct context to
```

sgibson91:
I followed examples in the below blog and I have now got this PR to work from forks. There's a little more overhead, but at least it's secure. Here is a PR demonstrating the workflow works: https://github.com/sgibson91/testing-gh-actions/pull/27

* https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

.github/workflows/comment-deployment-plan-pr.yaml
```diff
@@ -0,0 +1,130 @@
...
+  workflow_run:
+    workflows:
+      - "Deploy and test hubs"
+    types:
+      - "completed"
+
+jobs:
+  # This job downloads artifacts generated by the "Deploy and test hubs" workflow.
+  # The artifacts generated are two files: one containing the Pull Request number of the
+  # run, and a second containing Markdown tables detailing which hubs will upgraded as
```
GeorgianaElena, line 13:
```suggestion
  # run, and a second containing Markdown tables detailing which hubs will be upgraded as
```

.github/workflows/comment-deployment-plan-pr.yaml
```diff
@@ -0,0 +1,130 @@
...
+  # a result of merging a Pull Request. This Markdown content will be posted to the
+  # Pull Request as a comment. This means that no-one has to go searching through
+  # GitHub Actions logs to see if the hubs they intended to be upgraded, will be
+  # upgraded. The job posts the comment in a 'Netlify style', i.e., if a comment has
+  # already been posted (identified as having the 'github-actions[bot]' author and
+  # '<!-- deployment-plan -->' string in the body), the existing comment will be
+  # updated rather than posting a new one.
+  post-deployment-plan:
+    runs-on: ubuntu-latest
+    # Only run this job if the wroflow that triggered it was from a Pull Request and
```
GeorgianaElena, line 23:
```suggestion
    # Only run this job if the workflow that triggered it was from a Pull Request and
```

deployer/utils.py
```diff
@@ -32,3 +35,103 @@ def print_colour(msg: str, colour="green"):
...
+        "upgrade_support": "Upgrade Support?",
+        "reason_for_support_redeploy": "Reason for Support Redeploy",
+        "upgrade_staging": "Upgrade Staging?",
+        "reason_for_staging_redeploy": "Reason for Staging Redeploy",
+        "hub_name": "Hub Name",
+        "reason_for_redeploy": "Reason for Redeploy",
+    }
+
+    # A dictionary to convert row values when they are Boolean
+    row_converter = {
```
GeorgianaElena, line 65:
```suggestion
    boolean_converter = {
```

I feel like this is more descriptive for what it does?

GeorgianaElena (approved):
I think this is really great @sgibson91 ✨ Thanks for making this information easier to find

Also, I love how everything is documented in comments <3. Do you think it's also worth adding a small section about the workflow here and how all the pieces interact in the docs at https://infrastructure.2i2c.org/en/latest/reference/ci-cd.html?

deployer/utils.py
sgibson91: Agreed! Done in [`fe9be10` (#1424)](https://github.com/2i2c-org/infrastructure/pull/1424/commits/fe9be101f722c03e901e41bf5706fd1ff9adfa33)

sgibson91:
> I love how everything is documented in comments <3.

❤️ 

> Do you think it's also worth adding a small section about the workflow here and how all the pieces interact in the docs at [infrastructure.2i2c.org/en/latest/reference/ci-cd.html](https://infrastructure.2i2c.org/en/latest/reference/ci-cd.html)?

Yes! I'll do that in a follow-up PR :)

Issue #1423, opened 2022-06-16
[Incident] MEOM hub is unreachable


choldgraf:
### Summary

The MEOM hub is currently un-reachable, and there are a number of Kubernetes errors on the cluster.

### Impact on users

The hub is totally unreachable. It returns `503 service temporarily unavailable`.

### Important information

- Hub URL: https://meom-ige.2i2c.cloud/
- Support ticket ref: https://2i2c.freshdesk.com/a/tickets/150
- Slack discussion: https://2i2c.slack.com/archives/C028WU9PFBN/p1655370839409829

### Tasks and updates

- [ ] Discuss and address incident, leaving comments below with updates
- [ ] Incident has been dealt with or is over
- [ ] Copy/paste the after-action report below and fill in relevant sections
- [ ] Incident title is discoverable and accurate
- [ ] All actionable items in report have linked GitHub Issues

<details>
<summary>After-action report template</summary>

```
# After-action report

These sections should be filled out once we've resolved the incident and know what happened.
They should focus on the knowledge we've gained and any improvements we should take.

## Timeline

_A short list of dates / times and major updates, with links to relevant comments in the issue for more context._

All times in {{ most convenient timezone}}.

- {{ yyyy-mm-dd }} - [Summary of first update](link to comment)
- {{ yyyy-mm-dd }} - [Summary of another update](link to comment)
- {{ yyyy-mm-dd }} - [Summary of final update](link to comment)


## What went wrong

_Things that could have gone better. Ideally these should result in concrete
action items that have GitHub issues created for them and linked to under
Action items._

- Thing one
- Thing two

## Where we got lucky

_These are good things that happened to us but not because we had planned for them._

- Thing one
- Thing two

## Follow-up actions

_Every action item should have a GitHub issue (even a small skeleton of one) attached to it, so these do not get forgotten. These issues don't have to be in `infrastructure/`, they can be in other repositories._

### Process improvements

1. {{ summary }} [link to github issue]
2. {{ summary }} [link to github issue]

### Documentation improvements

1. {{ summary }} [link to github issue]
2. {{ summary }} [link to github issue]

### Technical improvements

1. {{ summary }} [link to github issue]
2. {{ summary }} [link to github issue]
```

</details>

choldgraf:
# What we have tried so far

(times CET)

### 11am

Found this error in `templates-clone` init container:

```
Cloning into '/srv/repo'...
[...]

PR #1425, opened 2022-06-17
Switch meom-ige core node to n1-highmem-2

yuvipanda:
g1-small is too small apparently - CPU usage was too high,
and the nodes kept going into NotReady.

Ref https://github.com/2i2c-org/infrastructure/issues/1423


GeorgianaElena (approved):

GeorgianaElena:
https://github.com/2i2c-org/infrastructure/pull/1426 should fix the failing linkcheck.

PR #1428, opened 2022-06-17
[BUGFIX] Ensure images_info is a list

sgibson91:


sgibson91:
Will self-merge as this is a small change to fix a bug

PR #1434, opened 2022-06-17
Add/update grafana tokens

GeorgianaElena:
Manually triggering https://github.com/2i2c-org/infrastructure/actions/runs/2515304896 revealed that some clusters were missing grafana tokens to be used by that action.

This adds those tokens and also updates the token for one cluster (uwhackweeks I believe) that had a token, but didn't have enough permissions (the action might need the token to have `admin` permissions not just `editor`, as a 403 error was returned by the `grafana-dashboards` deployer)?


sgibson91:
> the action might need the token to have `admin`

That seems to be correct based on https://infrastructure.2i2c.org/en/latest/howto/operate/grafana.html#setting-up-grafana-dashboards

sgibson91 (approved):

PR #1446, opened 2022-06-22
Update troubleshooting docs

sgibson91:
Reflect changes in our file configuration in the troubleshooting guide.

- CI/CD will no longer _always_ block upgrades to hubs, only if the support chart itself needs upgrading
- Refactor the config to match what is presently in the repo


docs/topic/troubleshooting.md
```diff
@@ -34,18 +34,15 @@ This should revert the Helm deployment to the previous revision (the one just be
 
 Prometheus collects Kubernetes metrics for our grafana charts and often requires a lot of memory during startup.
 If the server runs out of memory during the startup process, this presents itself as a "Timed out waiting for condition" error from `helm` on the command line, and `OOMKilled` events in the prometheus server pod events before the pod reverts to `CrashLoopBackOff` status.
```
sgibson91, line 36:
```suggestion
If the server runs out of memory during the startup process, this presents itself as a "Timed out waiting for condition" error from `helm` on the command line, and `OOMKilled` events in the prometheus server pod events before the pod reverts to `CrashLoopBackOff` status.
This might also block our CI/CD system from updating the hubs on the cluster if the support chart requires an upgrade after a Pull Request is merged.
```

Issue #1437, opened 2022-06-20
[Monitoring] Enable GitHub OAuth2 authentication for Grafana


GeorgianaElena:
### Context

Our current monitoring setup implies a Grafana for each cluster, with the on deployed for the ` 2i2c` cluster, being the "central" one, and gathering info about all other clusters.

This gives us the possibility to allow access to the cluster specific grafanas to the hub admins or community representatives of that specific hub/cluster, without too much work in restricting access to only specific datasources (clusters) to different teams of people.

One example, where defining such access could be useful, is https://github.com/2i2c-org/infrastructure/issues/1279.

Our current workflow for accessing these grafanas is using basic auth and described in more details in the docs here https://infrastructure.2i2c.org/en/latest/howto/operate/grafana.html?highlight=grafana#access-hub-grafana-dashboards. 

So having access to these grafanas currently means having access to the sops security keys, stored in `two-eye-two-see` Google project.

In order to give access to people outside of 2i2c, within the current setup, we would need to create new users, possible with only `viewer` roles and find a secure way to share an initial password with them. Adding a user can be done through the Grafana UI, but also using their REST API (so there's a possibility for automation of this task).

**However** I believe enabling GitHub authentication for Grafana might be a more appropriate way of achieving the same thing.

### Proposal

I propose enabling GitHub OAuth2 authentication of the grafanas deployed by 2i2c and use this docs to allow access to different GitHub orgs or teams https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/github.

*Note:* Using other OAuth2 providers could be a valid option too?

yuvipanda:
> However I believe enabling GitHub authentication for Grafana might be a more appropriate way of achieving the same thing.

+1.

The other alternative to explore for per-cluster Grafanas is to allow authentication with JupyterHub itself. This is possible because Grafana supports generic OAuth2 authentication (https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/generic-oauth/) and JupyterHub (with a service token) can provide that.

PR #1451, opened 2022-06-22
Add config for 2i2c grafana github auth

GeorgianaElena:
- [x] created a GitHub OAuth app and stored the secretes
- [x] added some initial config that should allow 2i2c org team member to sign in with GitHub in the 2i2c central grafana
- [x] deploy and test if this works

Closes https://github.com/2i2c-org/infrastructure/issues/1437


yuvipanda (approved):
https://github.com/berkeley-dsep-infra/datahub/blob/staging/support/secrets.yaml is similar config from berkeley, and it works well.

GeorgianaElena:
I've actually already deployed this and I was able to use my GitHub account to login into the central grafana.
I plan to merge this and then I will come back with another PR to enable GitHub auth for the leap grafana and one to add docs about this.

Issue #1225, opened 2022-04-21
Howto about migrating a user's directory to another username


damianavila:
### Context

This is coming from a Freshdesk support ticket: https://2i2c.freshdesk.com/a/tickets/122

@consideRatio shared a k8s Pod manifest to perform the migration: https://2i2c.slack.com/archives/C028WU9PFBN/p1650560320277699

@yuvipanda shared some edits on the manifest here: https://2i2c.slack.com/archives/C028WU9PFBN/p1650562288977739?thread_ts=1650560320.277699&cid=C028WU9PFBN

### Proposal

Port all this info into a howto section in our infra docs.

### Updates and actions

* [ ] who is willing (and has the time) to write the docs, @consideRatio WDYT? Maybe @yuvipanda can review the PR in-depth since he used it?

consideRatio:
I hope it is okay that I say no to help out further with this.

I'm generally conflicted about how to spend time, but I think I've in generally surpassed the amount of time I can justify spending with 2i2c given my statement of work with JMTE and failing to catch up with JMTE needs.

damianavila:
> I hope it is okay that I say no to help out further with this.

It is totally OK! Thanks for letting us know!

PR #1459, opened 2022-06-23
Add deployer command to open a shell with user homedirs

yuvipanda:
Fixes https://github.com/2i2c-org/infrastructure/issues/1225

Example:

```
❯ python3 deployer exec-homes-shell 2i2c staging                                                                                      2 ↵
Activated service account credentials for: [pilot-hubs-cd-sa@two-eye-two-see.iam.gserviceaccount.com]
Fetching cluster endpoint and auth data.
kubeconfig entry generated for pilot-hubs-cluster.
If you don't see a command prompt, try pressing enter.
root@shell:/# 
root@shell:/# cd home
root@shell:/home# ls
0000-2d0002-2d8055-2d2870  damianavila-402i2c-2eorg           georgianaelena                              yuvipanda-402i2c-2eorg
_shared                    deployment-2dservice-2dcheck       georgianaelena-402i2c-2eorg                 yuvipanda-40gmail-2ecom
blablabla                  dolocansimona-40gmail-2ecom        oauth2-7corcid-7c0000-2d0002-2d8055-2d2870
choldgraf-40gmail-2ecom    georgiana-2edolocan-40gmail-2ecom  sgibson-402i2c-2eorg
root@shell:/home# 
```

### DIFF
--- a/deployer/README.md
+++ b/deployer/README.md
@@ -150,6 +150,28 @@ optional arguments:
   -h, --help    show this help message and exit
 ```
 
+### `exec-homes-shell`
+
+This function gives you a shell with the home directories of all the
+users on the given hub in the given cluster mounted under `/home`.
+Very helpful when doing (rare) manual operations on user home directories,
+such as renames.
+
+When you exit the shell, the temporary pod spun up is removed.
+
+**Command line usage:**
+
+```bash
+usage: deployer exec-homes-shell [-h] cluster_name hub_name
+
+positional arguments:
+  cluster_name  The name of the cluster to perform actions on
+  hub_name      The deployed hub whose home directories are to be examined
+
+optional arguments:
+  -h, --help    show this help message and exit
+```
+
 ### `generate-helm-upgrade-jobs`
 
 This function consumes a list of files that have been added or modified, and from that deduces which hubs on which clusters require a helm upgrade, and whether the support chart also needs upgrading.
--- a/deployer/cli.py
+++ b/deployer/cli.py
@@ -12,6 +12,7 @@ from deploy_actions import (
     deploy,
     deploy_grafana_dashboards,
     deploy_support,
+    exec_homes_shell,
     generate_helm_upgrade_jobs,
     run_hub_health_check,
     use_cluster_credentials,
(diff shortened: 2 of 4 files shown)


GeorgianaElena (approved):
Love it! ✨ This could also be used when decommissioning a hub and deleting user's home dirs (I plan to test it with `anu` 😅 )

yuvipanda:
@GeorgianaElena ah, great use case! To ensure we don't accidentally delete the wrong hub homes, I've added the name of the hub to the pod name. This puts the name of the hub in the default prompt, so it looks like:

```
root@staging-shell:/#
```

GeorgianaElena:
Amazing <3 LGTM 🚀

sgibson91:
> puts the name of the hub in the default prompt

Can we add this magic so the name of the cluster is added to the prompt for use-cluster-credentials? :D I think there's an in-line TODO comment but I didn't know how to do it

yuvipanda:
@sgibson91 oh yeah i can take that on!

yuvipanda:
@sgibson91 so i poked at this and as you mentioned it's difficult because if we pass `-l` to shells, they set `PS1` and override ours. 

How about using something like https://github.com/jonmosco/kube-ps1 or (what I use) https://starship.rs/. This lets us put the current active kubectl context. Mine looks like this:

image

I think that might be the way to go, not PS1 :( 

If we want to use PS1, we should have have `use-cluster-config` print `PS1=<PS1>`, and then execute it as `source $(<use-cluster-config command)` - this is how `minikube` sets env vars.

sgibson91:
I use oh-my-zsh which does this as well. I guess we're just relying on everyone having some kind of terminal setup that shows this kind of context and remembering to exit the shell created after running use-cluster-credentials.

yuvipanda:
@sgibson91 yeah :(

sgibson91:
Can we get the next shell to print a message before handing over the prompt? Something like "Hello, you are now using CLUSTER's credentials! You can make kubectl calls against this cluster as normal. Remember to exit it once you're done!"

I'm thinking similar to how the terminal app prints "Last login:" before I get my prompt

damianavila:
Lovely stuff, @yuvipanda. Thanks!!

PR #1462, opened 2022-06-24
Allow deleting cilogon oauth apps using just the client id

GeorgianaElena:
While removing the anu hub, I realized that if you delete the file where the cilogon client credentials are stored, you won't be able to delete the cilogon client with only the cluster and the hub name.

This PR adds the option to delete a cilogon client by its id directly.

In this case you'll need to get all clients with:
```bash
python3 deployer/cilogon_app.py get-all
```
And then identitify the client of the hub and delete based on its id with:
```bash
python3 deployer/cilogon_app.py delete --id cilogon:/client_id/<id>
```

### DIFF
--- a/deployer/cilogon_app.py
+++ b/deployer/cilogon_app.py
@@ -294,16 +294,20 @@ class CILogonClientProvider:
         )
         print(self.admin_client.get(client_id))
 
-    def delete_client(self, cluster_name, hub_name):
-        config_filename = self._build_config_filename(cluster_name, hub_name)
-
-        client_id = self._load_client_id(config_filename)
-
-        # No client has been found
+    def delete_client(self, cluster_name, hub_name, client_id=None):
         if not client_id:
-            return
+            if not cluster_name or not hub_name:
+                print("Please provide either the client id to delete or or the cluster and hub name.")
+                return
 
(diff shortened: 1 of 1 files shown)


deployer/cilogon_app.py
```diff
@@ -294,16 +294,22 @@ def get_client(self, cluster_name, hub_name):
         )
         print(self.admin_client.get(client_id))
 
-    def delete_client(self, cluster_name, hub_name):
-        config_filename = self._build_config_filename(cluster_name, hub_name)
+    def delete_client(self, cluster_name, hub_name, client_id=None):
+        if not client_id:
+            if not cluster_name or not hub_name:
+                print(
+                    "Please provide either the client id to delete or or the cluster and hub name."
```
sgibson91, line 301:
```suggestion
                    "Please provide either the client id to delete or the cluster and hub name."
```

sgibson91 (commented):
@GeorgianaElena can we add some documentation for this great feature please?

- The deployer REAMDE: https://github.com/2i2c-org/infrastructure/blob/master/deployer/README.md
- Somewhere in the docs regarding our decommissioning process? I think having it in the issue template might be too bulky, but also I'm not sure so happy to have it there if there isn't a sensible place in our docs

sgibson91 (approved):

PR #1463, opened 2022-06-24
Decomission info

GeorgianaElena:
Update delete hub docs to be up to date with the template and reference it from the template.

### DIFF
--- a/docs/howto/hubs/delete-hub.md
+++ b/docs/howto/hubs/delete-hub.md
@@ -2,36 +2,82 @@
 
 If you'd like to delete a hub, there are a few steps that we need to take:
 
-1. **Remove the hub values file**.
-   Without removing the values file from the repository first, the hub could be redeployed by any merged PR that triggers our CI/CD pipeline.
-   Open a PR that removes the appropriate `config/clusters/<cluster_name>/<hub_name>.values.yaml` file, and removes the associated hub entry from the `config/clusters/<cluster_name>/cluster.yaml` file.
-   Steps 2 and 3 can be actioned while this PR is reviewed and merged.
+## 1. Manage existing data
+The existing data should either be migrated to another place or should be deleted, depending on what has been aggreed to with the Community Representative.
 
-2. **Backup the hub database**. Backup the `jupyterhub.sqlite` db off the hub.
+If the data should be migrated from the hub before decommissioning, then make sure that a 2i2c Engineer has access to the destination in order to complete the data migration.
 
-3. **Backup the home directory contents**.  Especially if we think that
-   users will want this information in the future (or if we plan to re-deploy a
-   hub elsewhere).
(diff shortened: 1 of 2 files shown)


docs/howto/hubs/delete-hub.md
```diff
@@ -2,36 +2,83 @@
...
+
+And then identitify the client of the hub and delete based on its id with:
+
+```bash
+$ python3 deployer/cilogon_app.py delete --id cilogon:/client_id/<id>
+```
+
+```bash
+$ python3 deployer/cilogon_app.py delete <cluster-name> <hub-name>
+```
```
sgibson91, line 84: What's the difference between these two commands?

Do I have to find the ID at all or would providing the cluster and hub name achieve the same thing without running the `get-all` command?

sgibson91 (approved):
This is great! Thank you! Just a couple of comments

docs/howto/hubs/delete-hub.md
GeorgianaElena: So, it's possible to delete an app using just the cluster and hub name, but only if the relevant `config/clusters/<cluster_name>/enc-<hub_name>.values.yaml` hasn't been deleted,  because this is the file where the client id is stored.

Since this step is to be taken after the deletion of that file, the above is not possible anymore.

**But** I think we could move this step before the one that instructs deleting the file, right? Do think that would make more sense? we do get rid of the extra step of retrieving all the clients

**Also**, having both commands here was totally a mistake, so in case we decide not to re-order the steps, last cmd mus go:
````suggestion
```bash
$ python3 deployer/cilogon_app.py delete --id cilogon:/client_id/<id>
```
````

File: docs/howto/hubs/delete-hub.md [hunk 2]
```diff
@@ -2,36 +2,83 @@
...
+## 1. Manage existing data
+The existing data should either be migrated to another place or should be deleted, depending on what has been aggreed to with the Community Representative.
+
+If the data should be migrated from the hub before decommissioning, then make sure that a 2i2c Engineer has access to the destination in order to complete the data migration.
+
+The sub-sections below cover both scenarios.
+### 1.1. Migrate data
+
+#### 1.1.1 Backup the hub database
+Backup the `jupyterhub.sqlite` db off the hub.
```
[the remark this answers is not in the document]
GeorgianaElena:
```suggestion
[Backup the `jupyterhub.sqlite` database](https://infrastructure.2i2c.org/en/latest/howto/hubs/move-hub.html#transfer-the-jupyterhub-database) off the hub.
```

docs/howto/hubs/delete-hub.md
sgibson91: I think we pick whichever workflow feels best for these docs, and then document the difference between the two workflows in the deployer readme :)

PR #1464, opened 2022-06-24
Document the scripts in the deployer module

GeorgianaElena:


deployer/README.md
```diff
@@ -216,6 +214,62 @@ optional arguments:
   --check-dask-scaling  For daskhubs, optionally check that dask workers can be scaled
 ```
 
+## Sub-scripts
```
sgibson91, line 217: Soooo, it kind of niggles at me that we have scripts in the deployer that aren't called by some `python deployer {COMMAND}` function (ie. they don't have an entry in deploy_actions.py). Especially considering we have an extra_scripts dir for some utility style scripts.

What are your thoughts on this? Is it worth having them as sub-scripts, or should we either move them to extra_scripts or make them full commands of the deployer? Maybe the extra_scripts dir has outlived its usefulness and those should be migrated here?

I feel like we need a strategy here otherwise we're going to have scripts spread out across the repo and called differently and things will get confusing.

(This comment doesn't have to block this PR btw, but I think it's an important thing for us to think about.)

yuvipanda: I think putting them into `deployer_actions` is the way to go.

GeorgianaElena:
@sgibson91 and @yuvipanda, are you ok with me merging this PR and track updating "sub-scripts" to deployer actions in a different issue + PR?

sgibson91 (approved):
Yes!

yuvipanda (approved):
Yes absolutely! Not a blocker for this at all.

Issue #1384, opened 2022-06-03
[EVENT] Jack Eddy Symposium [DECOMMISSION SCHEDULED]


damianavila:
### Summary

The [3rd Jack Eddy Symposium](https://cpaess.ucar.edu/meetings/eddy-symposium-2022) will take place June 6 - 11 in Vail CO. The conference honors the late [Jack Eddy](https://en.wikipedia.org/wiki/John_A._Eddy), an astronomer who pioneered the field of heliophysics and demonstrated that the variability of the Sun can be seen in signals (e.g. tree rings with carbon dating) on Earth. Jack Eddy approached his research using interdisciplinary tools and appears to have inspired a next generation of Sun-Earth-Climate researchers.

Website: https://cpaess.ucar.edu/meetings/eddy-symposium-2022

 The new hub was deployed here: https://github.com/2i2c-org/infrastructure/issues/1329

### Event Info

- **Community Representative:** @dan800
- **Event begin:** June 6th
- **Event end:** June 10th
- **Active times:** 8 AM MT - 6 PM MT
- **Number of attendees:** ~45
- [**Hub Events Calendar**](https://calendar.google.com/calendar/u/2?cid=Y19rdDg0c2g3YW5tMHNsb2NqczJzdTNqdnNvY0Bncm91cC5jYWxlbmRhci5nb29nbGUuY29t)

### Hub info

- **Hub URL**: https://jackeddy.2i2c.cloud/
- **Hub decommisioned after event?**: TBD (@dan800, can you confirm?)

### Task List

**Before the event**

- [x] Dates confirmed with the community representative and added to Hub Events Calendar.
- [x] Quotas from the cloud provider are high-enough to handle expected usage.
- [x] **One week before event** Hub is running.
- [x] Confirm with Community Representative that their workflows function as expected.
  - <details>
    <summary>👉Template message to send to community representative</summary>
    
    ```
    Hey {{ COMMUNITY REPRESENTATIVE }}, the date of your event is getting close!
    
    Could you please confirm that your hub environment is ready-to-go, and matches your hub's infrastructure setup, by ensuring the following things:
    - [ ] Confirm that the "Event Info" above is correct
    - [ ] On your hub: log-in and authentication works as-expected
    - [ ] `nbgitpuller` links you intend to use resolve properly
    - [ ] Your notebooks and content run as-expected
    ```
    
    </details>  
[...]

PR #1466, opened 2022-06-24
Decomission jackeddy hub

yuvipanda:
Fixes https://github.com/2i2c-org/infrastructure/issues/1384


damianavila (approved):
Code removal and terraform output, both LGTM.

Issue #919, opened 2022-01-04
Deploy and operate a BinderHub for Pangeo


choldgraf:
### Description / problem to solve

**Problem description**
The Pangeo BinderHub has been down for about a month ([due to crypto mining](https://github.com/pangeo-data/pangeo-binder/issues/195), but also because it did not have operational support to keep it going sustainably). The Pangeo community made heavy of use their Binder deployment, and it powered a lot of reproducible sharing (e.g., via [gallery.pangeo.io](https://gallery.pangeo.io).).

**Proposed solution**
We should deploy a BinderHub on the 2i2c deployment infrastructure that can live in parallel to the JupyterHub we run for the Pangeo community. We'll need to make a few modifications to their setup (including using up-to-date binderhub versions and locking down auth more reliably).

**What's the value and who would benefit**
This would allow the Pangeo community to re-gain the use of their BinderHub, which would benefit many people!

### Implementation guide and constraints

There are a few things that we should consider here:

- We'll need to update our configuration and CI/CD infrastructure to be able to deploy a BinderHub chart, since we're currently assuming JupyterHub.
- We want to make sure we don't deploy a BinderHub and immediately run into the same crypto issues that Pangeo ran into before

Here's a GitHub issue where @scottyhq describes the environment that was available on the Pangeo BinderHub: https://github.com/pangeo-data/pangeo-binder/issues/195#issuecomment-989107771

### Updates and ongoing work

Here are a few major issues that would need to be tackled as part of this effort:

- [ ] https://github.com/2i2c-org/infrastructure/issues/857 (so that we can generalize our config beyond just JupyterHub)
- [ ] https://github.com/2i2c-org/infrastructure/issues/879 (if we want to strengthen the CI/CD infrastructure in general before deploying new things)
- [ ] https://github.com/2i2c-org/infrastructure/issues/863

choldgraf:
cc @rabernat and @sgibson91 - is there anything major here that I am missing? I believe that @sgibson91 is working with @consideRatio on https://github.com/2i2c-org/infrastructure/issues/857 right now, which is laying the foundation to letting us deploy BinderHubs from the Pangeo cluster CI/CD.

sgibson91:
@choldgraf I think this is a really nice outline of the work that needs to be done to get us into a position where we are ready to deploy a BinderHub. I'm happy with how this is and add to the list of tasks as an when they arise

sgibson91:
[...]

PR #1467, opened 2022-06-24
Make our binderhub/ helm-chart a little more generic

yuvipanda:
- Move pangeo-specific messages into their own file
- Rename binder secret file to match name of deployment

Ref https://github.com/2i2c-org/infrastructure/issues/919


config/clusters/2i2c/cluster.yaml
```diff
@@ -47,7 +47,8 @@ hubs:
       # matching value for jupyterhub.custom.2i2c.add_staff_user_ids_of_type!
       enabled: false
     helm_chart_values_files:
-      - enc-binder.secret.values.yaml
+      - binder-staging.values.yaml
```
sgibson91, line 50:
```suggestion
```

This file doesn't exist anymore - _unless_ you want to migrate what you've removed from the helm chart into that file?

yuvipanda: @sgibson91 whoops, forgot to `git add`! Perils of relying so heavily on `git commit -p`... Added now!

sgibson91 (approved):

PR #1475, opened 2022-06-27
Reduce number of times we refresh helm repos

yuvipanda:
We refresh the 'local' helm repo cache only once instead of
three times for every deploy. This significantly speeds up
deployment, especially on slower internet connections.

### DIFF
--- a/deployer/config_validation.py
+++ b/deployer/config_validation.py
@@ -53,15 +53,15 @@ def _prepare_helm_charts_dependencies_and_schemas():
 
     daskhub_dir = helm_charts_dir.joinpath("daskhub")
     _generate_values_schema_json(daskhub_dir)
-    subprocess.check_call(["helm", "dep", "up", daskhub_dir])
+    subprocess.check_call(["helm", "dep", "up", daskhub_dir, "--skip-refresh"])
 
     binderhub_dir = helm_charts_dir.joinpath("binderhub")
     _generate_values_schema_json(binderhub_dir)
-    subprocess.check_call(["helm", "dep", "up", binderhub_dir])
+    subprocess.check_call(["helm", "dep", "up", binderhub_dir, "--skip-refresh"])
 
     support_dir = helm_charts_dir.joinpath("support")
     _generate_values_schema_json(support_dir)
-    subprocess.check_call(["helm", "dep", "up", support_dir])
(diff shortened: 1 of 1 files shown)


sgibson91:
We need at least one call to helm repo update without --skip-refresh otherwise CI/CD is failing - because those machines probably don't have the cached versions of the charts like we do on our local ones.

sgibson91:
Can we add this as a CLI option instead? Seems CI is breaking with the --skip-refresh option

yuvipanda:
ooh, that's interesting - I wonder why CI is breaking! My understanding was that it was refreshing the system's helm repos, and so this shouldn't happen.

consideRatio:
This can be reproduced locally like this.

```
# clear helm cache
rm ~/.cache/helm/repository/*

# create a dummy helm chart
cd /tmp
helm create test
cd test

# append a remote dependency
cat <<EOF >> Chart.yaml
dependencies:
  - name: jupyterhub
    version: "1.2.0"
    repository: "https://jupyterhub.github.io/helm-chart"
EOF

# reproduce situation
helm dep up . --skip-refresh
```

My understanding is that helm will require a index.yaml from someplace, and that is what is being skipped refreshing. But, if that isn't refreshed and assumed to be in a cache, and we have the error.

What we may have expected was that `--skip-refresh` would still fetch the index.yaml file from the provided repo, but not do `helm repo update`.

I suspect this won't lead to a change in behavior in helm/helm if reported, but personally on my computer, I've stopped using `helm repo add` etc, and instead relies on `--repo=...` as part of `helm upgrade`. Like that, my `helm repo list` is blank, and the process of updating named helm chart repositories index.yaml files is instant.

sgibson91:
I'm going to mark this as draft until it's ready to be worked on again

consideRatio:
For reference, this change relates to a deep rabbit hole with many connections, for example Chart.yaml / Chart.lock, `helm dep up` / `helm dep build`, etc.

Related:
- https://github.com/helm/chart-releaser-action/issues/74#issuecomment-1313931556
- https://github.com/helm/helm/issues/9903#issuecomment-1315488423

yuvipanda:
I don't think I'll have time to come back to this one so closing for now :(

PR #1476, opened 2022-06-28
Add terraform config for linked-earth cluster

sgibson91:
related to #1418 

Output of terraform apply:

```
Terraform will perform the following actions:

  # google_container_cluster.cluster will be created
  + resource "google_container_cluster" "cluster" {
      + cluster_ipv4_cidr           = (known after apply)
      + datapath_provider           = (known after apply)
      + default_max_pods_per_node   = (known after apply)
      + enable_binary_authorization = false
      + enable_intranode_visibility = (known after apply)
      + enable_kubernetes_alpha     = false
      + enable_l4_ilb_subsetting    = false
      + enable_legacy_abac          = false
      + enable_shielded_nodes       = true
      + enable_tpu                  = false
      + endpoint                    = (known after apply)
      + id                          = (known after apply)
      + initial_node_count          = 1
      + label_fingerprint           = (known after apply)
      + location                    = "us-central1"
      + logging_service             = (known after apply)
      + master_version              = (known after apply)
      + monitoring_service          = (known after apply)
      + name                        = "linked-earth-cluster"
      + network                     = "default"
      + networking_mode             = (known after apply)
      + node_locations              = [
          + "us-central1-c",
        ]
      + node_version                = (known after apply)
      + operation                   = (known after apply)
      + private_ipv6_google_access  = (known after apply)
      + project                     = "linked-earth-hubs"
      + remove_default_node_pool    = true
      + self_link                   = (known after apply)
      + services_ipv4_cidr          = (known after apply)
      + subnetwork                  = (known after apply)
      + tpu_ipv4_cidr_block         = (known after apply)

      + addons_config {
          + cloudrun_config {
              + disabled           = (known after apply)
              + load_balancer_type = (known after apply)
            }

          + config_connector_config {
              + enabled = (known after apply)
            }

          + dns_cache_config {
              + enabled = (known after apply)
            }

          + gce_persistent_disk_csi_driver_config {
              + enabled = (known after apply)
            }

          + gcp_filestore_csi_driver_config {
              + enabled = (known after apply)
            }

          + horizontal_pod_autoscaling {
              + disabled = true
            }

          + http_load_balancing {
              + disabled = true
            }

          + istio_config {
              + auth     = (known after apply)
              + disabled = (known after apply)
            }

          + kalm_config {
              + enabled = (known after apply)
            }

          + network_policy_config {
              + disabled = (known after apply)
            }
        }

      + authenticator_groups_config {
          + security_group = (known after apply)
        }

[...]

Issue #1418, opened 2022-06-14
[New Hub] PaleoCube and PaleoHack hubs


colliand:
### Hub Description

2i2c has been engaged by members of the paleoclimate community. Deborah Khider (@khider) has been our principal contact in establishing this connection. This engagement extends previous work we've done to support the PaleoHackWeeks. We will learn more from Deborah about the community and their needs as we roll out the service. Here are a few links that provide some background information. 

[LinkedEarth](http://linked.earth/)

[Deborah's home page](https://earth.usc.edu/~khider/)

### Community Representative(s)

@khider

### Important dates

- **Target start date**: June 30, 2022
- **Required start date**: July 15, 2022
- **Any important dates for usage**: _???_

### Hub Authentication Type

CILogon (e.g., username@institution.org)

### Hub logo information

- **URL to Hub Image**: https://github.com/LinkedEarth/Logos/blob/master/linkedearth_hub_logo.png?rgh-link-date=2022-06-15T16%3A21%3A32Z
- **URL for Image Link**: https://linked.earth/

### Hub user image

- Repository for user image: https://quay.io/repository/linkedearth/pyleoclim
- User image registry: quay.io
- User image tag and name: `latest` tags

Once again, seeking input from @khider! 

Does your community have a Docker Image that specifies the software environment you wish to use in your hub(s)? If so, please share a link and we will review and take steps to set things up using that environment. If not, we can proceed with a Pangeo image and help you and your community adapt the environment to better meet your needs. 

### Extra features you'd like to enable

- [x] Specific cloud provider or datacenter: GCP, us-central1-c
- [x] Dedicated Kubernetes cluster
- [x] Scalable Dask Cluster

### Hub URL

linkedearth.2i2c.cloud

### Hub Type

daskhub

### Tasks to deploy the hub

- [x] Engineer who will deploy the hub is assigned
- [x] Deploy information filled in above
- [x] Initial cluster deployment PR: https://github.com/2i2c-org/infrastructure/pull/1476
- [x] Initial Hub deployment PR: https://github.com/2i2c-org/infrastructure/pull/1476
- [x] Administrators able to log on
- [x] Community Representative satisfied with hub environment
- [x] Hub now in steady-state

damianavila:
We already have a live PaleoHack hub: https://github.com/2i2c-org/infrastructure/blob/master/config/clusters/2i2c/paleohack2021.values.yaml

Is this a request for a new one, @colliand?

[...]

PR #1478, opened 2022-06-28
Deploying helm components to linked earth cluster

sgibson91:
- Deploy the support chart
- Deploy two daskhubs: staging and prod

related https://github.com/2i2c-org/infrastructure/issues/1418

(This PR is branched from https://github.com/2i2c-org/infrastructure/pull/1476 and should be rebased against master before merging)


sgibson91:
Currently blocked deploying the support chart due to following error message:

> Error: unable to build kubernetes objects from release manifest: resource mapping not found for name: "support-prometheus-server" namespace: "support" from "": no matches for kind "Ingress" in version "networking.k8s.io/v1beta1"

yuvipanda:
my guess is it's because the cluster is new enough that [networking.k8s.io/v1beta1](http://networking.k8s.io/v1beta1) is removed, and it needs to be [networking.k8s.io/v1](http://networking.k8s.io/v1)

yuvipanda:
So this might need a bump up of the prometheus chart (we're currently at 13.x, latest is 15.10.1). Some labels changed, so we might need https://github.com/jupyterhub/grafana-dashboards/pull/42 to be merged. This would impact prometheus on all clusters.

sgibson91:
Commits are getting messy from working on two branches that are related but one isn't merged. I'm closing this and will just add all config to #1476

Issue #1381, opened 2022-06-01
Deployment of cert-manager and dask-gateway require `kubectl apply -f <crd manifests>` step when version bumped


consideRatio:
Helm charts that include CRDs such as cert-manager and dask-gateway should typically by upgraded by doing `kubectl apply -f <crd manifests>` before being upgraded.

The "default upgrade procedure" of cert-manager is to do this `kubectl apply` step manually first, and users of dask-gateway upgrading from 0.9.0 to 2022.4.0 were supposed to do that as well at least for that release, probably in general as well.

For cert-manager, this can be done like in mybinder.org-deploy: https://github.com/jupyterhub/mybinder.org-deploy/blob/2db9dbdc8f51356713219dd2ddf72d0bb36723fb/deploy.py#L221-L253, and the same procedure can be used for dask-gateway.

## Action point

- [ ] Ensure automation managing dask-gateway includes `kubectl apply` of CRDs
   - https://github.com/2i2c-org/infrastructure/pull/1486
- [x] Ensure automation managing cert-manager includes `kubectl apply` of CRDs
   - https://github.com/2i2c-org/infrastructure/pull/1479

sgibson91:
I wasn't sure what the manifest URL for dask-gateway was and couldn't find an example in the docs, so I only updated cert-manager for now, but happy to do dask-gateway as well.

consideRatio:
@sgibson91 I made a note in the changelog for 2022.4.0 when an update was needed: https://gateway.dask.org/changelog.html#breaking-changes

consideRatio:
Feel free to ping me for review @sgibson91!

PR #1479, opened 2022-06-28
Automate installation of CRDs for cert-manager

sgibson91:
- Adds a CLI flag to deploy-support to set which version of cert-manager to install (this was previously hard-coded into the deployer)
- Add a subprocess.check_call command to install cert-managers CRDs before installing/upgrading the chart

related https://github.com/2i2c-org/infrastructure/issues/1381

### DIFF
--- a/deployer/README.md
+++ b/deployer/README.md
@@ -103,13 +103,15 @@ This command only needs to be run once per cluster, not once per hub.
 **Command line usage:**
 
 ```bash
-usage: python deployer deploy-support [-h] cluster_name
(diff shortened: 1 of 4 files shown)


consideRatio (approved):
This will be helpful when upgrading cert-manager beyond 1.3.1 to recent versions of 1.8 something I think.

The mybinder.org-deploy issue about cert-manager upgrade is here https://github.com/jupyterhub/mybinder.org-deploy/issues/2183.

EDIT: opened https://github.com/2i2c-org/infrastructure/issues/1480 about upgrading cert-manager.

PR #1482, opened 2022-06-28
Upgrade prometheus to latest

yuvipanda:
Required for compatibility with k8s 1.22

Ref https://github.com/2i2c-org/infrastructure/pull/1478#issuecomment-1168588736

### DIFF
--- a/helm-charts/support/Chart.yaml
+++ b/helm-charts/support/Chart.yaml
@@ -7,7 +7,7 @@ dependencies:
   # Prometheus for collection of metrics.
   # https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus
   - name: prometheus
-    version: 13.8.0
+    version: 15.10.1
     repository: https://prometheus-community.github.io/helm-charts
 
   # Grafana for dashboarding of metrics.


sgibson91 (approved):
Let's try it and see what breaks in grafana-land

yuvipanda:
Yay!

Trying this out in meom-ige, I get:

```
Error: UPGRADE FAILED: cannot patch "support-kube-state-metrics" with kind Deployment: Deployment.apps "support-kube-state-metrics" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/instance":"support", "app.kubernetes.io/name":"kube-state-metrics"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

yuvipanda:
since kube-state-metrics is actually stateless, I did k -n support delete deployment support-kube-state-metrics.

This helped deploy complete! Now let's look at the damage on grafana...

yuvipanda:
So I had to manually kill a deployment and service object to get this upgrade to work. I wrote this small throwaway script:

```bash
#!/bin/bash
set -euo pipefail
CLUSTER="${1}"
echo "kubectl -n support delete deployment support-kube-state-metrics && kubectl -n support delete service support-prometheus-node-exporter" | python3 deployer use-cluster-credentials $CLUSTER
python3 deployer deploy-support ${CLUSTER
```

And then ran it with `ls config/clusters | xargs -L1 ./wat.bash`. 

With that, https://github.com/2i2c-org/infrastructure/actions/runs/2579393900 passed!

sgibson91:
That is some seriously cool use of use-cluster-credentials!!

yuvipanda:
@sgibson91 thank you for building it!!!!

PR #1484, opened 2022-06-29
Add James to our staff lists for hub access

sgibson91:
This will grant @jmunroe access to 2i2c hubs as an admin

related https://github.com/2i2c-org/meta/issues/337


sgibson91:
I will self-merge this without review since it is a small update

PR #1486, opened 2022-06-29
Automatically install dask-gateway CRDs for daskhubs

### LINKED ISSUE #1381: Deployment of cert-manager and dask-gateway require `kubectl apply -f <crd manifests>` step when version bumped

sgibson91:
This PR adds a --dask-gateway-version CLI flag to the deploy action so that the appropriate CRDs for the dask-gateway helm chart can be installed automatically. It defaults to 2022.6.1, which is the current dask-gateway version we are running.

A deprecated CLI flag has also been removed.

fixes #1381

### DIFF
--- a/deployer/cli.py
+++ b/deployer/cli.py
@@ -65,9 +65,6 @@ def main():
         help="The hub, or list of hubs, to install/upgrade the helm chart for",
     )
     deploy_parser.add_argument(
-        "--skip-hub-health-test", action="store_true", help="Bypass the hub health test"
-    )
-    deploy_parser.add_argument(
         "--config-path",
         help="File to read secret deployment configuration from",
         # This filepath is relative to the PROJECT ROOT
--- a/deployer/cli.py
+++ b/deployer/cli.py
@@ -73,6 +73,8 @@ def main():
     deploy_parser.add_argument(
         "--dask-gateway-version",
         type=str,
+        # This version must match what is listed in daskhub's Chart.yaml file
+        # https://github.com/2i2c-org/infrastructure/blob/HEAD/helm-charts/daskhub/Chart.yaml#L14
         default="2022.6.1",
         help="For daskhubs, the version of dask-gateway to install for the CRDs. Default: 2022.6.1",
     )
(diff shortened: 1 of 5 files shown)


consideRatio (approved):

deployer/cli.py
```diff
@@ -64,15 +64,18 @@ def main():
...
-    deploy_parser.add_argument(
-        "--skip-hub-health-test", action="store_true", help="Bypass the hub health test"
-    )
     deploy_parser.add_argument(
         "--config-path",
         help="File to read secret deployment configuration from",
         # This filepath is relative to the PROJECT ROOT
         default="shared/deployer/enc-auth-providers-credentials.secret.yaml",
     )
+    deploy_parser.add_argument(
```
yuvipanda, line 73: Can we make a note in daskhub/Chart.yaml where we specify the daskhub version that if that is bumped, this default also needs to be changed?

yuvipanda (changes requested):
One suggestion, consider it approved after that.

helm-charts/daskhub/Chart.yaml
```diff
@@ -7,6 +7,9 @@ dependencies:
   - name: basehub
     version: "0.1.0"
     repository: file://../basehub
+    # If bumping the version of dask-gateway, please also bump the default version set
+    # in the deployer's CLI
+    # https://github.com/2i2c-org/infrastructure/blob/HEAD/deployer/cli.py#L73-L78
```
sgibson91, line 12:
```suggestion
    # https://github.com/2i2c-org/infrastructure/blob/HEAD/deployer/cli.py#L73-L80
```

Issue #1480, opened 2022-06-28
Update cert-manager to modern version


consideRatio:
We use cert-manager 1.3.1 currently, but we should upgrade to a somewhat modern version to avoid getting issues with things. For example, note this comment from the 1.5.0 cert-manager changelog.

> The recent Kubernetes 1.22 release has removed a number of deprecated APIs. You can read the official blog post [Kubernetes API and Feature Removals In 1.22](https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/) to learn more about it. If you intend to upgrade to Kubernetes 1.22, you must upgrade to cert-manager 1.5.

We upgraded things in mybinder.org-deploy, and the issue tracking that was https://github.com/jupyterhub/mybinder.org-deploy/issues/2183. Looking at the release notes from 1.3.1 upwards, is it sufficient to upgrade using `kubectl apply -f <cert-manager crds>`?

https://cert-manager.io/docs/release-notes/

- cert-manager 1.6.0 use of v1alphaX resources provided by cert-manager should be replaced by v1 resources, if we provide for example a ClusterIssuer of a resource type v1alpha, it must be ugpraded.
- cert-manager 1.8.0 validates a field previously not validated, we don't use such invalid fields right? See https://cert-manager.io/docs/release-notes/release-notes-1.8/#breaking-changes-you-must-read-this-before-you-upgrade.

I think we can upgrade to 1.8.x without issues, but I suggest we upgrade once first to to 1.6.3, and then keep going to 1.8.2, because I suspect without clear motivation based on recollections and reading the changelog notes that it can help us avoid issues.

sgibson91:
I needed to bump cert-manager in order to unblock #1476 (and the swathe of chart bumps that come with it). I went straight from 1.3.1 to 1.8.2 with no issues in commit [`2435282` (#1476)](https://github.com/2i2c-org/infrastructure/pull/1476/commits/2435282bb06a048ca66c4fdccc071d1925163666)

sgibson91:
I believe this was closed with the merge of #1491

PR #1491, opened 2022-06-30
Bump cert-manager to v1.8.2

yuvipanda:
Required for k8s v1.22

related #1480

Done by @sgibson91


yuvipanda:
Going to merge this so as to not leave clusters in an in-between state. Tested on 2i2c cluster, seems ok.

PR #1492, opened 2022-06-30
Tuskegee: Modify username_pattern

sean-morris:
Temporary Addition of gmail.com to username_pattern for
Tuskegee University


sean-morris:
I did not notice the changes that I was writing over in username_pattern

PR #1493, opened 2022-06-30
Tuskegee: modified username_pattern

sean-morris:
- I change the ercvd from gmail to berkeley so could remove the specific
username_pattern entry
- I temporarily added @gmail.com to
username_pattern so Tuskegee users could use the hub
while we work out the cilogon issues


yuvipanda (approved):
We generally don't want to allow arbitrary gmail users all the time, but given there's an event tomorrow morning I think this is ok. Let's revert it by next week though?

sean-morris:
> We generally don't want to allow arbitrary gmail users all the time, but given there's an event tomorrow morning I think this is ok. Let's revert it by next week though?

Yes for sure

PR #1498, opened 2022-07-04
CSU  hub: set user_name pattern to only edu

sean-morris:
This is intended to serve CSU's; I changed the username pattern to include any edu.

### DIFF
--- a/config/clusters/cloudbank/cluster.yaml
+++ b/config/clusters/cloudbank/cluster.yaml
@@ -269,11 +269,10 @@ hubs:
     auth0:
       # connection update? Also ensure the basehub Helm chart is provided a
       # matching value for jupyterhub.custom.2i2c.add_staff_user_ids_of_type!
-      connection: password
-      password:
-        database_name: database-csu
+      enabled: false
     helm_chart_values_files:
       # The order in which you list files here is the order the will be passed
       # to the helm upgrade command in, and that has meaning. Please check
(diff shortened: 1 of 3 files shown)


GeorgianaElena (changes requested):
@sean-morris, I believe that right now this is the only hub left that works with username & password authentication through `auth0`. Is this something that's still wanted?

If not, I believe in addition to only allow `.edu` addresses maybe we can also switch this hub to CILogon so that institutionals identity providers can be used too (example PR that does the same thing https://github.com/2i2c-org/infrastructure/pull/1218)

For context, the `cloudbank demo` hub used to have the same config and we changed it to cilogon + `.edu` allow list because of https://github.com/2i2c-org/infrastructure/issues/1216.

sean-morris:
> @sean-morris, I believe that right now this is the only hub left that works with username & password authentication through `auth0`. Is this something that's still wanted?
> 
> If not, I believe in addition to only allow `.edu` addresses maybe we can also switch this hub to CILogon so that institutionals identity providers can be used too (example PR that does the same thing #1218)
> 
> For context, the `cloudbank demo` hub used to have the same config and we changed it to cilogon + `.edu` allow list because of #1216.

perfect yes. I am with you. It should be correct for cilogon now. Thanks

yuvipanda (approved):
Try it out!

sean-morris:
> Try it out!

Thanks!

PR #1504, opened 2022-07-07
debugging ci: why is `helm dep up . --skip-refresh` failing?

consideRatio:
- Reduce number of times we refresh helm repos
- debugging


consideRatio:
Conclusion reported in https://github.com/2i2c-org/infrastructure/pull/1475#issuecomment-1177691512

PR #1514, opened 2022-07-12
Add Jim as admin to hubs

GeorgianaElena:


damianavila:
Wondering if we would need to do the same with @jmunroe (most likely yes...).

Issue #1505, opened 2022-07-07
[New Hub] NeuroHackademy 2022


damianavila:
### Hub Description

A general description of the hub requested lives here: https://github.com/2i2c-org/infrastructure/issues/1300#issue-1232600616

### Community Representative(s)

@arokem

### Important dates

- **Target start date**: July 15th
- **Required start date**: July 22th
- **Any important dates for usage**: 
  - Event begin: July 25th
  - Event end: August 5th

### Hub Authentication Type

GitHub Authentication (e.g., @mygithubhandle)

### Hub logo information

- **URL to Hub Image**: https://user-images.githubusercontent.com/118582/177910655-a2403aeb-9c08-44fd-94a8-543b5070f6a9.png
- **URL for Image Link**: https://neurohackademy.org/ (todo: confirm this is it)

### Hub user image

- Repository for user image: https://github.com/neurohackademy/hub-image-2022
- User image registry: quay.io
- User image tag and name: quay.io/arokem/nh-jhub-2022

### Extra features you'd like to enable

- [x] Specific cloud provider or datacenter: GCP or AWS is preferable.
- [ ] Dedicated Kubernetes cluster
- [ ] Scalable Dask Cluster

### Other relevant information

More detailed information about the requirements:
* General needs + some tiny specifics: https://github.com/2i2c-org/infrastructure/issues/1300#issue-1232600616
* More specifics: https://github.com/2i2c-org/infrastructure/issues/1300#issuecomment-1128126597
* Template customization:  https://github.com/2i2c-org/infrastructure/issues/1300#issuecomment-1145530640
* Zoom links under auth: https://github.com/2i2c-org/infrastructure/issues/1500

### Hub URL

neurohackademy.2i2c.cloud

### Hub Type

basehub

### Tasks to deploy the hub

- [x] Engineer who will deploy the hub is assigned
- [x] Deploy information filled in above
- [x] Initial Hub deployment PR: https://github.com/2i2c-org/infrastructure/pull/1507
- [x] Administrators able to log on
- [x] Community Representative satisfied with hub environment
- [x] Hub now in steady-state

damianavila:
Event issue ref: https://github.com/2i2c-org/infrastructure/issues/1300

damianavila:
@arokem, do you have any preference for the name of the hub? Any logo image link that you want to share with us?

damianavila:
[...]

PR #1518, opened 2022-07-13
Fix git sync of static content repo and make lab default for neurohackademy

GeorgianaElena:
Ref https://github.com/2i2c-org/infrastructure/issues/1505#issuecomment-1183295359


yuvipanda (approved):
Thanks!

PR #1519, opened 2022-07-13
Tweak dask-gateway on LEAP hub

yuvipanda:
- Provide more memory & cpu for the dask-scheduler pod. When
  we were running a larger cluster, the scheduler kept dying
  and providing strange errors. Increasing resources helps us
  get bigger clusters.
- Tweak the size of nodes for dask. Dask, at least as used in
  LEAP, is primarily memory bound and not cpu-bound. This helps
  us use resources more efficiently.
- Bump up the prometheus resource requests so it stops crashing
- Add another label to record from pods, required for reporting on
  dask-gateway

Worked with @TomNicholas and @rabernat for bringing up a big cluster
to crunch through some compute for Tom's talk today.

The other thing we discovered is that by default, dask-gateway
sets number of *threads* to number of CPUs, which actually causes
a lot of GIL contention in our workflows. Need to carefully tune
and investigate that.

### DIFF
--- a/config/clusters/leap/common.values.yaml
+++ b/config/clusters/leap/common.values.yaml
@@ -145,8 +145,8 @@ dask-gateway:
     backend:
       scheduler:
         cores:
-          request: 0.8
-          limit: 1
+          request: 1
+          limit: 2
         memory:
-          request: 1G
-          limit: 2G
+          request: 4G
+          limit: 4G
--- a/config/clusters/leap/support.values.yaml
+++ b/config/clusters/leap/support.values.yaml
@@ -16,9 +16,12 @@ prometheus:
           hosts:
             - prometheus.leap.2i2c.cloud
     resources:
+      requests:
+        cpu: 3
+        memory: 20Gi
       limits:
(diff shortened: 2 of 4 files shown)


consideRatio:
For heavy computations, i think a sane option could be to co-locate scheduler pod with the user pod. I think its really complicated on how to handle the scheduler resource allocation - noi option seems great.

yuvipanda:
@consideRatio so right now the scheduler pods land in the same nodes as the notebook nodes. I agree ideally they'd be *inside* the notebook pod, but unfortunately it looks like that isn't possible with dask-gateway?

consideRatio:
No i dont think so, but co-locating on the same node would be possible with some tricks i figure.

yuvipanda:
@consideRatio ya i'd love to figure that out, but I don't think there's a really easy way to do that unfortunately. Regardless, we definitely need to bump up the limits for larger computations as otherwise the OOM killer gets it.

consideRatio (approved):

Issue #1517, opened 2022-07-13
Static web content URLs behave differently with trailing slash or not `/`


GeorgianaElena:
The static web content is available at https://staging.2i2c.cloud/textbook/ but not at https://staging.2i2c.cloud/textbook (note the missing trailing slash)

I would have expected a redirect from https://staging.2i2c.cloud/textbook to https://staging.2i2c.cloud/textbook/ and then see the content.

Maybe something is missing from the nginx config?

damianavila:
cc @yuvipanda

yuvipanda:
Fixed in #1520 :)

PR #1520, opened 2022-07-13
Redirect slashless static sites to the slashful version

yuvipanda:
slash slash slashes slash slash

Fixes https://github.com/2i2c-org/infrastructure/issues/1517

### DIFF
--- a/helm-charts/basehub/templates/static/configmap.yaml
+++ b/helm-charts/basehub/templates/static/configmap.yaml
@@ -9,9 +9,15 @@ data:
   nginx.conf: |
     server {
       listen 8080;
-      location {{ .Values.staticWebsite.ingress.path }} {
+      # We require path to *not* have a trailing slash
+      location {{ .Values.staticWebsite.ingress.path }}/ {
         index index.html;
         alias /srv/content/repo;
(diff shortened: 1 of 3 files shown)


helm-charts/basehub/templates/static/configmap.yaml
```diff
@@ -9,9 +9,15 @@ data:
   nginx.conf: |
     server {
       listen 8080;
-      location {{ .Values.staticWebsite.ingress.path }} {
+      # We require path to *not* have a trailing slash
```
consideRatio, line 12:
```suggestion
      # We enforce path to *not* have a trailing slash via the helm chart schema
```

consideRatio (approved):
Great to see an example on how this is done!!

GeorgianaElena (approved):
🚀

yuvipanda:
I love writing nginx config! Been so long...

yuvipanda:
fun fact - https://github.com/yuvipanda/jupyterhub-nginx-chp was one of my first JupyterHub projects, that was dynamically doing nginx config as a replacement for configurable-http-proxy

PR #1528, opened 2022-07-15
Fix slash redirect and 403 nginx dir indexing

GeorgianaElena:
Fixes slash redirect and 403 error when accessing the web content (ref https://github.com/2i2c-org/infrastructure/issues/1505#issuecomment-1185160478). (follow-up to https://github.com/2i2c-org/infrastructure/pull/1520)

- absolute redirects need to be turned off, otherwise, the redirect from slashless version to slash version is to an absolute address of the nginx server, which doesn't make any sense from the outside network ([reference](https://serverfault.com/questions/905657/nginx-redirects-to-http-after-port-in-redirect-off))
- the 403 showing this in the logs:
   ```
    directory index of "/srv/content/repo" is forbidden
   ```
   
   I guess, adding the trailing slash to the repo content directory fixed this by generating the correct path to the index file.
   
P.S. > why are the nginx config + errors are confusing 💥


consideRatio (approved):
LGTM, nice work figuring these details out!!

PR #1538, opened 2022-07-19
Debugging deployment plan workflow

sgibson91:
It's currently broken 😢 trying to figure out why

Adding a debugging statement to the workflow


sgibson91:
Self merging this so we actually see the output of the statements in the logs

PR #1540, opened 2022-07-19
[debugging] Rewrite deployment plan comment workflow entirely in Python

sgibson91:
We will end up having error messages directly from the GitHub API this
way which will be easier to debug if something is still not working

### DIFF
--- a/.github/workflows/comment-deployment-plan-pr.yaml
+++ b/.github/workflows/comment-deployment-plan-pr.yaml
@@ -66,7 +66,7 @@ jobs:
               run_id=run_id,
               per_page=100,
           )
-          print(f"Does all artifacts have something in it?: {bool(all_artifacts)}")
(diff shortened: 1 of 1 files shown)


.github/workflows/comment-deployment-plan-pr.yaml
```diff
@@ -32,61 +32,70 @@ jobs:
...
+
+          # List all artifacts for the workflow run
+          all_artifacts = paged(
+              api.actions.list_workflow_run_artifacts,
+              owner=owner,
+              repo=repo,
+              run_id=run_id,
+              per_page=100,
+          )
+          print(f"Does all artifacts have something in it?: {bool(all_artifacts)}")
```
sgibson91, line 69:
```suggestion
          print(f"Does all_artifacts have something in it?: {bool(all_artifacts)}")
```

sgibson91:
Self-merging to see if this unbreaks the workflow

PR #1541, opened 2022-07-19
[ci/cd debugging] Properly enumerate through the list of workflow artifacts for the run

sgibson91:
Fixes a bug where `all_artifacts` cannot be subscripted


sgibson91:
Self-merged to continue debugging

PR #1542, opened 2022-07-19
[ci/cd debugging]

sgibson91:
Output current working directory and the contents after downloading the archive

Trying to establish where it is saved to


sgibson91:
self-merging to continue debugging

PR #1543, opened 2022-07-19
[ci/cd debugging] Properly handle zip download from a URL

sgibson91:
I switch to using the requests package here to properly download a zip file from a URL

🤞🏻


sgibson91:
Will self-merge to see if it works/what other bugs crop up

PR #1545, opened 2022-07-19
[ci/cd debugging] Remove dedent when creating comment body

sgibson91:
It didn't work as expected and the resulting comment looked messy. Hoping pre-commit won't try to override this 🤞🏻


sgibson91:
self-merging as this will fix a bug

PR #1550, opened 2022-07-20
Actually upgrade the binder-staging hub on the 2i2c cluster

sgibson91:
Since it has `staging` in it's name, the binder-staging hub was not being upgraded since we don't generate explicit jobs for multiple staging hubs on the same cluster because the 2i2c cluster is the only one with multiple staging hubs running on it.

We add a step so that binder-staging is upgraded and update the docs to reflect this.

### DIFF
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -201,7 +201,7 @@ jobs:
         run: |
           python deployer deploy ${{ matrix.jobs.cluster_name }} dask-staging
 
-      - name: Upgrade dask-staging hub on cluster ${{ matrix.jobs.cluster_name }} if it exists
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-hubs.yaml
```diff
@@ -201,6 +201,11 @@ jobs:
         run: |
           python deployer deploy ${{ matrix.jobs.cluster_name }} dask-staging
 
+      - name: Upgrade dask-staging hub on cluster ${{ matrix.jobs.cluster_name }} if it exists
```
sgibson91, line 204:
```suggestion
      - name: Upgrade binder-staging hub on cluster ${{ matrix.jobs.cluster_name }} if it exists
```

yuvipanda (approved):

Issue #1536, opened 2022-07-19
Automatically run script to clone a git repo on user server startup


damianavila:
### Context

This is a requirement for the upcoming event #1300.

In https://2i2c.freshdesk.com/a/tickets/155, Ariel Rokem asks that the hub is configured in a way so that user pods starting will automatically run a script doing a certain thing - namely using gitpuller CLI from nbgitpuller python package to clone a fresh repo into a folder.
It may merit some discussion on how to do that technically. 

### Proposal 
@consideRatio describes how he did it earlier in https://github.com/2i2c-org/infrastructure/issues/1505#issuecomment-1183375652 - using a k8s lifecycle postStart hook and a script he had mounted via the Helm chart but now can be provided via singleuser.extraFiles functionality as hub specific configuration.

### Updates and actions

This can be approached in various ways, depending on how we expect/want to use in the future, in the context of other hubs too. The list below aggregates multiple great options from people that we can choose from:

➡️ 1.  We provide script files via singleuser.extraFiles, and we configure a postStart hook to run one script on startup
➡️ 2.  We configure an initContainer to run the script
➡️ 3.  We provide script files via singleuser.extraFIles, but we don't configure a postStart hook to run one script on startup
➡️ 4. We ask that Ariel provide script files via the Dockerfile and that users are provided instructions to run a script from the terminal and/or use a nbgitpuller link to get setup initially, as that de-couples 2i2c from adding customization for this hub beyond what has been done for other hubs.
➡️ 5. We ask that Ariel provide script files via the Dockerfile and run a [Jupyter Docker Stack start-up hook](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#startup-hooks) to execute it
➡️ 6. We make the defaultUrl be a pointer to the nbgitpuller URL? That way, users can login to the hub and it'll treat them exactly like they clicked the nbgitpuller url
➡️ 7.

consideRatio:
The approach of using a postStart lifecycle hook is a bit problematic and I wanted to highlight that so it doesn't come as a surprise later.

If a [k8s `postStart` lifecycle hook](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/) fail, it means pod startup will be considered failed. So, if the script does something like accessing a github repo, and github is intermittently down - the script can fail, and the user pod startup fails.

Due to this, use of `postStart` lifecycle hooks should be adopted with some caution.

[...]

PR #1552, opened 2022-07-21
[neurohackademy] Set defaultUrl to be a nbgitpuller link instead

GeorgianaElena:
Addresses https://2i2c.freshdesk.com/a/tickets/155, 

Reference: https://github.com/2i2c-org/infrastructure/issues/1536

**Todo:** 
- [x] make sure nbgitpuller is installed in the user image

### DIFF
--- a/config/clusters/2i2c/neurohackademy.values.yaml
+++ b/config/clusters/2i2c/neurohackademy.values.yaml
@@ -35,8 +35,7 @@ jupyterhub:
           name: The National Institutes of Health grant 2R25MH112480-06
           url: https://reporter.nih.gov/search/ydTvTwXxk0yd6eGdRznbLQ/project-details/10409452
   singleuser:
-    # Pulls content from https://github.com/neurohackademy/nh2022-curriculum using nbgitpuller then opens up /lab
-    defaultUrl: https://neurohackademy.2i2c.cloud/hub/user-redirect/git-pull?repo=https%3A%2F%2Fgithub.com%2Fneurohackademy%2Fnh2022-curriculum&urlpath=lab%2Ftree%2Fnh2022-curriculum%2F&branch=main
+    defaultUrl: /git-pull?repo=https%3A%2F%2Fgithub.com%2Fneurohackademy%2Fnh2022-curriculum&urlpath=lab%2Ftree%2Fnh2022-curriculum%2F&branch=main
     # User image: https://quay.io/repository/arokem/nh-jhub-2022?tab=tags
     image:
       name: quay.io/arokem/nh-jhub-2022
@@ -57,6 +56,33 @@ jupyterhub:
           http://github.com/login/oauth/authorize:
             username_derivation:
               username_claim: "preferred_username"
+     extraFiles:
+      configurator-schema-default:
+        data:
+          properties:
+            Spawner.default_url:
+              type: string
+              title: Default User Interface
+              enum:
+                - "/tree"
+                - "/lab"
+                - "/git-pull?repo=https%3A%2F%2Fgithub.com%2Fneurohackademy%2Fnh2022-curriculum&urlpath=lab%2Ftree%2Fnh2022-curriculum%2F&branch=main"
+              default: "/lab"
+              enumMetadata:
+                interfaces:
+                  - value: "/tree"
+                    title: Classic Notebook
+                    description:
+                      The original single-document interface for creating
+                      Jupyter Notebooks.
+                  - value: "/lab"
+                    title: JupyterLab
+                    description: A Powerful next generation notebook interface
+                  - value: "/git-pull?repo=https%3A%2F%2Fgithub.com%2Fneurohackademy%2Fnh2022-curriculum&urlpath=lab%2Ftree%2Fnh2022-curriculum%2F&branch=main"
+                    title: Pull curriculum repo and redirect to /lab
+                    description: Use ngbitpuller to pull https://github.com/neurohackademy/nh2022-curriculum and redirect to /lab afterwards
+
+
     services:
       dex:
         url: http://dex:5556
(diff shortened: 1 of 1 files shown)


config/clusters/2i2c/neurohackademy.values.yaml
```diff
@@ -35,8 +35,9 @@ jupyterhub:
           name: The National Institutes of Health grant 2R25MH112480-06
           url: https://reporter.nih.gov/search/ydTvTwXxk0yd6eGdRznbLQ/project-details/10409452
   singleuser:
+    # Pulls content from https://github.com/neurohackademy/nh2022-curriculum using nbgitpuller then opens up /lab
+    defaultUrl: https://neurohackademy.2i2c.cloud/hub/user-redirect/git-pull?repo=https%3A%2F%2Fgithub.com%2Fneurohackademy%2Fnh2022-curriculum&urlpath=lab%2Ftree%2Fnh2022-curriculum%2F&branch=main
```
GeorgianaElena, line 39: Not sure if this should be a full formed url or just a redirect path?

yuvipanda: @GeorgianaElena I think this is the first time anyone is trying this ever! Can you try it out on staging to see how it goes?

GeorgianaElena:
I managed to get this to work on staging. Try it out before its gone 😄 https://staging.2i2c.cloud

There were couple of bumps making this to work:
### 1. defaultUrl needs to be a path, not a full formed URL, so its
  - `/git-pull?repo=https%3A%2F%2Fgithub.com%2Fneurohackademy%(....)`
  and not
  - https://neurohackademy.2i2c.cloud/hub/user-redirect/git-pull?repo=https%3A%2F%2Fgithub.com%2Fneurohackademy(...)`
 
### 2. [If there is a default set in the schema for available options in the configurator, it will always override the config elsewhere in our YAML files](https://infrastructure.2i2c.org/en/latest/topic/config.html#inheritance-of-configuration)

   - So, in order for `singleuser.defaultUrl` to take effect, the default interface set through the configurator schema, needs to not be set.
   - An alternative to this would what I did in https://github.com/2i2c-org/infrastructure/pull/1552/commits/343909e38c1bad60f94af8b0c00fa6dabc7d189d, which is to overwrite the configurator's schema and make the new defaultUrl option show up in the actual configurator.
   
I believe this is ready to be tried out on the neurohackademy hub too.

config/clusters/2i2c/neurohackademy.values.yaml
```diff
@@ -35,8 +35,9 @@ jupyterhub:
           name: The National Institutes of Health grant 2R25MH112480-06
           url: https://reporter.nih.gov/search/ydTvTwXxk0yd6eGdRznbLQ/project-details/10409452
   singleuser:
+    # Pulls content from https://github.com/neurohackademy/nh2022-curriculum using nbgitpuller then opens up /lab
```
consideRatio, line 38:
```suggestion
    # Pulls content from https://github.com/neurohackademy/nh2022-curriculum using nbgitpuller then opens up /lab
    # This configuration needs to be paired with overriding schema configuration for jupyterhub-configurator, see the
    # hub.extraFiles.configurator-schema-default entry below!
```

config/clusters/2i2c/neurohackademy.values.yaml
```diff
@@ -56,6 +57,12 @@ jupyterhub:
           http://github.com/login/oauth/authorize:
             username_derivation:
               username_claim: "preferred_username"
+    extraFiles:
+      configurator-schema-default:
```
consideRatio, line 61:
```suggestion
      # We override the configuration of jupyterhub-configurator as we have configured singleuser.defaultUrl
      # above, and no longer wants the user to override this
      configurator-schema-default:
```

consideRatio (commented):
If jupyterhub-configurator's schema had to be updated alongside the change of `/git-pull` as default url, I figure it kind of have to be mentioned in a comment in this configuration to manage the complexity growth.

An option to overriding the schema, could be to relax the schema and use the jupyterhub-configurator to configure this directly instead.

config/clusters/2i2c/neurohackademy.values.yaml
GeorgianaElena:
```suggestion
      # We disable the defaultUrl set by the jupyterhub-configurator schema,
      # otherwise it would override the singleuser.defaultUrl set above.
      # This is because the jupyterhub-configurator schema gets loaded after the hub
      # yaml config and overrides any config set before.
      # Ref: https://infrastructure.2i2c.org/en/latest/topic/config.html#inheritance-of-configuration
      configurator-schema-default:
```

GeorgianaElena: Actually, setting `defaultUrl` to null in configurator's schema doesn't disable the configurator for changing the interface. You can still do that.

It just doesn't set a default again(https://github.com/2i2c-org/infrastructure/blob/a3cf596562e374aa01df8cc70d5a26b4817111a7/helm-charts/basehub/values.yaml#L189) overriding whatever we set in the hub's config file.

I added a suggestion above with what the explanation I think it should be. It also has a link to our docs where there a list with the order in which each config file gets loaded and can override each other. Do you think that is enough?

config/clusters/2i2c/neurohackademy.values.yaml
GeorgianaElena: I edited your comment above @consideRatio. Hope that's ok. WDYT?

config/clusters/2i2c/neurohackademy.values.yaml
consideRatio: Thanks for clarifying how things work! This is absolutely adequate, the key part in my suggestions were to somehow link together the configuration so that its not two separate parts that have a clear connection - but that connection isn't seen by readers of the config.

GeorgianaElena: Makes sense! Thanks @consideRatio ✨

GeorgianaElena:
Thanks for the review @consideRatio! Do you think this looks good enough for a merge?

consideRatio (approved):
Absolutely go for it @GeorgianaElena!

Issue #1555, opened 2022-07-22
[BUG] Central grafana datasource script behaves strangely


sgibson91:
### Context

<img width="1556" alt="Screenshot 2022-07-22 at 14 31 37" src="https://user-images.githubusercontent.com/44771837/180460754-b4d62a93-b1d9-4060-8888-5589e7c26be7.png">

When I've added new clusters to the central grafana, I get strange error messages from the script. First run through says the new cluster was added successfully but there was also an error, though no error messages appear. The second run through shows that the cluster was indeed added, but again says there was an error even though there were no error messages?

damianavila:
cc @GeorgianaElena in case she has a quick hint about it!

PR #1556, opened 2022-07-22
Print error message only if there were exceptions

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/1555


sgibson91 (approved):
🚀 Thank you @GeorgianaElena !

Issue #1515, opened 2022-07-12
[New Hub] OceanHackWeek 2022


colliand:
### Hub Description

OceanHackWeek 2022 will take place August 15-19. This event is under the umbrella of eScience Institute.

### Community Representative(s)

- @paigem
- @ocefpaf
- @abkfenris

### Important dates

- **Target start date**:  July 25th, 2022
- **Required start date**:  August 1, 2022
- **Any important dates for usage**: August 8-26, 2022

### Hub Authentication Type

GitHub Authentication (e.g., @mygithubhandle)

### Hub logo information

- **URL to Hub Image**: {{ URL HERE }}
- **URL for Image Link**: {{ URL HERE }}

### Hub user image

- Repository for user image: { REPO LINK IF IT EXISTS }
- User image registry: [ghcr.io/oceanhackweek/python:e7303ce](https://github.com/oceanhackweek/jupyter-image/pkgs/container/python/29849868?tag=e7303ce) and [ghcr.io/oceanhackweek/r:e7303ce](https://github.com/oceanhackweek/jupyter-image/pkgs/container/r/29849834?tag=e7303ce).
- User image tag and name: { NAME AND TAG IF IT EXISTS }

### Extra features you'd like to enable

- [ ] Specific cloud provider or datacenter: 
- [ ] Dedicated Kubernetes cluster
- [x] Scalable Dask Cluster

### Other relevant information

- around 100 participants or so, and 30ish organizers/instructors/helpers
- Github teams based auth
   - team that should have access:
      - @oceanhackweek/participants_2022
      - @oceanhackweek/ohw22-organizers
   - [ ] todo: check if sub-teams get access too
- have different Python/R profiles
- [ ] todo: figure out bumping the image tag given different profiles 
- [x] todo: check if there can be different defaultUrls for different profiles

### Hub URL

oceanhackweek.2i2c.cloud

### Tasks to deploy the hub

- [x] Engineer who will deploy the hub is assigned
- [x] Deploy information filled in above
- [x] Initial Hub deployment PR: https://github.com/2i2c-org/infrastructure/pull/1560
- [x] Administrators able to log on
- [x] Community Representative satisfied with hub environment
- [ ] Hub now in steady-state

colliand:
An email  was sent to hello@2i2c.org by Alex Kerney today. I've transcribed some of the content into the hub request fields above. Further details need to be processed. I invite 2i2c Engineering Team to consider the questions posed below. 

---

[...]

PR #1560, opened 2022-07-25
Update ohw hub for this year's event

GeorgianaElena:
- [x] Github teams based auth
   team that should have access:
    - [x] @oceanhackweek/participants_2022
    - [x] @oceanhackweek/ohw22-organizers
    - [x] check if sub-teams get access too: so far it looks like people were able to login. Not 100% it was because they were authorized from last year's event, but will keep an eye on it.
- [x] have different Python/R profiles
- [x] figure out bumping the image tag given different profiles -> we should probably use https://github.com/sgibson91/bump-jhub-image-action
- [x] check if there can be different defaultUrls for different profiles
- [x] make sure the user image uses at least jupyterhub 2.3.1 otherwise default_url won't work
- [x] why is this error still showing up in the hub logs:
   ```
    WARNING: hub.config.JupyterHub.authenticator_class was set to github and c.Authenticator.allowed_users was set, custom 2i2c jupyterhub config is now resetting allowed_users to an empty set.
  ```
- [ ] remove existing auth0 oauth2 app for ohw hub

Reference: https://github.com/2i2c-org/infrastructure/issues/1515

### DIFF
--- a/config/clusters/2i2c/cluster.yaml
+++ b/config/clusters/2i2c/cluster.yaml
@@ -76,13 +76,12 @@ hubs:
     domain: ohw.pilot.2i2c.cloud
     helm_chart: daskhub
     auth0:
-      # connection update? Also ensure the basehub Helm chart is provided a
-      # matching value for jupyterhub.custom.2i2c.add_staff_user_ids_of_type!
-      connection: github
+      enabled: false
     helm_chart_values_files:
       # The order in which you list files here is the order the will be passed
       # to the helm upgrade command in, and that has meaning. Please check
       # that you intend for these files to be applied in this order.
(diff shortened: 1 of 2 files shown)


GeorgianaElena:
This PR is missing some things, like the different images for the profiles and a better thinking about profile resources.

But since the images are not available, I set them to be the 2i2c default user image and I was thinking to merge this if it seems appropriate so that we make sure the authentication and profile works. WDYT?

sgibson91:
> But since the images are not available, I set them to be the 2i2c default user image and I was thinking to merge this if it seems appropriate so that we make sure the authentication and profile works.

I endorse this!

sgibson91 (approved):

PR #1563, opened 2022-07-26
Move github authenticator allowed list validation to the validate workflow

GeorgianaElena:
While investigating why this warning message was present in the logs for https://github.com/2i2c-org/infrastructure/pull/1560 even though an allowed_users list was not set, I noticed the fixme comment 😅

~Also, the `auth0` config was never validated either because we were never calling the assert func in the validate step.~ 
We were actually calling it in the `deploy` function.

So this PR:
- changes the assert auth0 config method be a a generic `validate_authenticator_config`
- this func does two things now:
  -  It asserts that only a single auth method is enabled.
       An error is raised when an authenticator other than Auth0
       is enabled and `auth0` is not explicitly disabled.
  - It asserts that when the JupyterHub GitHubOAuthenticator is used,
       then `Authenticator.allowed_users` is not set.
       An error is raised otherwise.
       
I remember last time there were some objections against naming this function `validate_...` when what it does is to `assert`. So, happy to consider changing this if people thing that this should instead be two `assert` functions.

### DIFF
--- a/deployer/cli.py
+++ b/deployer/cli.py
@@ -7,6 +7,7 @@ from config_validation import (
     validate_cluster_config,
     validate_hub_config,
     validate_support_config,
+    validate_authenticator_config,
 )
 from deploy_actions import (
     deploy,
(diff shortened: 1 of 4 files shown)


sgibson91 (approved):
Thanks for this work @GeorgianaElena

PR #1564, opened 2022-07-26
Bump AWI CIROH user image tag

sgibson91:
Bumps the image tag to the most recently published one which has a working Linux desktop feature


sgibson91:
I will self-merge this one as it is only an image tag bump

PR #1572, opened 2022-07-28
Update ohw allowed_organizations list and user images

GeorgianaElena:
Ref https://github.com/2i2c-org/infrastructure/issues/1515#issuecomment-1196617428


GeorgianaElena:
I have already manually deployed this

sgibson91 (approved):

PR #1583, opened 2022-08-01
Upgrade oauthenticator version and docs

GeorgianaElena:
- Upgrade and pin `oauthenticator` version to latest commit
- Update docs about building hub image on an M1 arch
- Update docs about cilogon oathenticator config to match current version of `oauthenticator` we're using
- Update two hubs (`dask-staging` and `2i2c-demo`) to use latest feature of `oauthenticator`, i.e. `allowed_idps.idp.allowed_domains`. 
   
  There are other hubs that could use this instead of the `username_pattern` we were using (esp the `cloudbank` hubs), but I wasn't sure which identity providers they wished to use specifically, so I didn't change them. But I suggest using this pattern from now on.

Ref: https://github.com/2i2c-org/infrastructure/pull/1579#discussion_r934573485

### DIFF
--- a/docs/howto/configure/auth-management.md
+++ b/docs/howto/configure/auth-management.md
@@ -438,34 +438,34 @@ The steps to enable the JupyterHub CILogonOAuthenticator for a hub are simmilar 
         Example config:
 
         ```yaml
-        jupyterhub:
-          hub:
-            config:
-              JupyterHub:
-                authenticator_class: cilogon
-              CILogonOAuthenticator:
-                oauth_callback_url: https://{{ HUB_DOMAIN }}/hub/oauth_callback
-                admin_users:
-                  - admin@anu.edu.au
-                # Only show the option to login with Google and ANU
-                shown_idps:
-                  - https://idp2.anu.edu.au/idp/shibboleth
-                  - https://accounts.google.com/o/oauth2/auth
-                allowed_idps:
-                  http://google.com/accounts/o8/id:
-                    username_derivation:
-                      username_claim: "email"
-                      action: "prefix"
-                      prefix: "2i2c"
-                    allowed_domains:
-                      - "2i2c.org"
-                  https://idp2.anu.edu.au/idp/shibboleth:
-                    username_derivation:
-                      username_claim: "email"
-                      action: "strip_idp_domain",
-                      domain: "anu.edu.au"
-                    allowed_domains:
-                      - "anu.edu.au"
+          jupyterhub:
+            hub:
+              config:
+                JupyterHub:
+                  authenticator_class: cilogon
+                CILogonOAuthenticator:
+                  oauth_callback_url: https://{{ HUB_DOMAIN }}/hub/oauth_callback
+                  admin_users:
+                    - admin@anu.edu.au
+                  # Only show the option to login with Google and ANU
+                  shown_idps:
+                    - https://idp2.anu.edu.au/idp/shibboleth
+                    - https://accounts.google.com/o/oauth2/auth
+                  allowed_idps:
+                    http://google.com/accounts/o8/id:
+                      username_derivation:
+                        username_claim: "email"
+                        action: "prefix"
+                        prefix: "2i2c"
+                      allowed_domains:
+                        - "2i2c.org"
+                    https://idp2.anu.edu.au/idp/shibboleth:
+                      username_derivation:
+                        username_claim: "email"
+                        action: "strip_idp_domain",
+                        domain: "anu.edu.au"
+                      allowed_domains:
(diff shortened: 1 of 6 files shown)


docs/howto/configure/auth-management.md
```diff
@@ -364,48 +364,115 @@ The steps to enable the JupyterHub CILogonOAuthenticator for a hub are simmilar
...
+                  allowed_domains:
+                    - "2i2c.org"
+                https://idp2.anu.edu.au/idp/shibboleth:
+                  username_derivation:
+                    username_claim: "email"
+                    action: "strip_idp_domain",
+                    domain: "anu.edu.au"
+                  allowed_domains:
+                    - "anu.edu.au"
+      ```
```
sgibson91, line 469:
```suggestion
          - only shows Shibboleth, the ANU identity provider, and Google as the possible login options though CILogon
          - sets `2i2c.org` as the only domain that can login to the hub using Google
          - sets `anu.edu.au` as the only domain that can login to the hub using Shibboleth
          - adds a `2i2c:` prefix to the usernames logging in through Google. The hub usernames are the `email` addresses of these accounts, as specified through `username_claim`.
          - strips the `@anu.edu.au` domain from the usernames logging in through Shibboleth. The hub usernames are the `email` addresses of these accounts, as specified through `username_claim`.

      ```{note}
      To get the value of the key that must go in the `allowed_idp` dict for a specific IdP, go to https://cilogon.org/idplist and get the value of the `EntityID` key of the desired institutional IdP.
      ```

      Example config:

      ```yaml
      jupyterhub:
        hub:
          config:
            JupyterHub:
              authenticator_class: cilogon
            CILogonOAuthenticator:
              oauth_callback_url: https://{{ HUB_DOMAIN }}/hub/oauth_callback
              admin_users:
                - admin@anu.edu.au
              # Only show the option to login with Google and Shibboleth
              shown_idps:
                - https://idp2.anu.edu.au/idp/shibboleth
                - https://accounts.google.com/o/oauth2/auth
              allowed_idps:
                  username_derivation:
                    username_claim: "email"
                    action: "prefix"
                    prefix: "2i2c"
                  allowed_domains:
                    - "2i2c.org"
                https://idp2.anu.edu.au/idp/shibboleth:
                  username_derivation:
                    username_claim: "email"
                    action: "strip_idp_domain",
                    domain: "anu.edu.au"
                  allowed_domains:
                    - "anu.edu.au"
      ```
```

I changed the wording slightly as it sounded like ANU were hosting their own identity provider, when I think they are actually subscribed to Shibboleth (as many universities are)

docs/topic/hub-image.md
```diff
@@ -47,6 +47,16 @@ When this hub image needs to be updated, the steps to take are:
      chartpress --push
      ```
 
+     ````{note}
+     If you are on macOs wit M1, you need to run chartpress with [docker buildx](https://docs.docker.com/build/buildx/) under the hood and specify which platform to use, i.e. `amd64`.
```
sgibson91, line 51:
```suggestion
     If you are on macOs with M1, you need to run chartpress with [docker buildx](https://docs.docker.com/build/buildx/) under the hood and specify which platform to use, i.e. `amd64`.
```

helm-charts/images/hub/Dockerfile
```diff
@@ -6,6 +6,11 @@
 # quay.io container registry credentials configured to have access to
 # https://quay.io/repository/2i2c/pilot-hub.
 #
+# Note: if on macOs wit M1, you need to run chartpress with docker buildx under the hood and specify
```
sgibson91, line 9:
```suggestion
# Note: if on macOs with M1, you need to run chartpress with docker buildx under the hood and specify
```

sgibson91 (approved):
LGTM! Thanks for this work!

docs/howto/configure/auth-management.md
GeorgianaElena: > I changed the wording slightly as it sounded like ANU were hosting their own identity provider, when I think they are actually subscribed to Shibboleth (as many universities are)

I believe that's ok. Whatever makes more sense for everybody. After reading the CILogon docs and refactoring the oauthenticators, I admit I am confused about the terminology they use and I realized there isn't necessarily a convention on identity provider vs entity id, the use of username claims vs username keys 🤯 etc
So thank you ✨

damianavila:
This PR is awesome, thanks for working on this one @GeorgianaElena.

> Upgrade and pin oauthenticator version to latest commit

Can we create an issue to follow up with a released version when it is released? I am always worried about using unreleased "versions" of upstream dependencies 😉  (sorry to be annoying with these recurrent requests!).

Issue #1576, opened 2022-07-29
[EVENT] OceanHackWeek 2022


GeorgianaElena:
### Event Info

- **Community Representative:** @abkfenris 
- **Event begin:** August 15 
- **Event end:** August 19 
- **Active times:**  todo
- **Number of attendees:** around 100 participants or so, and 30ish organizers/instructors/helpers 
- **Notes**: The undergraduate program starts a week earlier and goes a week later, so the event is longer (August 8-26), though that is going to be about a tenth of the users as we will have from 15-19 for the HackWeek itself.
- [**Hub Events Calendar**](https://calendar.google.com/calendar/u/2?cid=Y19rdDg0c2g3YW5tMHNsb2NqczJzdTNqdnNvY0Bncm91cC5jYWxlbmRhci5nb29nbGUuY29t)

### Hub info

- **Hub URL**: https://oceanhackweek.2i2c.cloud/user 
- **Hub decommisioned after event?**: todo

### Task List

**Before the event**

- [x] Dates confirmed with the community representative and added to Hub Events Calendar.
- [x] Quotas from the cloud provider are high-enough to handle expected usage.
- [x] **One week before event** Hub is running.
- [x] Confirm with Community Representative that their workflows function as expected.
  - <details>
    <summary>👉Template message to send to community representative</summary>
    
    ```
    Hey {{ COMMUNITY REPRESENTATIVE }}, the date of your event is getting close!
    
    Could you please confirm that your hub environment is ready-to-go, and matches your hub's infrastructure setup, by ensuring the following things:
    - [ ] Confirm that the "Event Info" above is correct
    - [ ] On your hub: log-in and authentication works as-expected
    - [ ] `nbgitpuller` links you intend to use resolve properly
    - [ ] Your notebooks and content run as-expected
    ```
    
    </details>  
- [x] **1 day before event**, either a separate nodegroup is provisioned for the event or the cluster is scaled up.

**During and after event**

- [x] Confirm event is finished.
- [x] Nodegroup created for the hub is decommissioned / cluster is scaled down.
- [x] `bump-ohw-image-tags.yaml` from https://github.com/2i2c-org/infrastructure/tree/master/.github/workflows is deleted
- [ ] Hub decommissioned (if needed).
- [x] Debrief with community representative.
  - <details>
    <summary>👉Template debrief to send to community representative</summary>
      
     ```
     Hey {{ COMMUNITY REPRESENTATIVE }}, your event appears to be over 🎉
     
[...]

PR #1588, opened 2022-08-03
Run ohw bump image tags workflow more frequently

GeorgianaElena:
Adds a new workflow to bump ohw image tags more frequently.

The current `bump-image-tags` workflow runs once every week, so this temporarily adds a new workflow that runs every 6 hours and checks for new image tags for the ohw hub. We should keep this probably only during the event https://github.com/2i2c-org/infrastructure/issues/1576.

I chose 6h as the interval since this means 4 times a day and it means about two times during someone's awake time. But feel free to suggest another value if it makes more sense.

TODO:
- [x] add removing this workflow on the todo list of the issue tracking the event once this PR is accepted
- [x] add this as a suggested practice in the event issue template for hubs that run multiple images per user profile and hence the configurator doesn't work for them. -> https://github.com/2i2c-org/infrastructure/issues/1590


sgibson91 (approved):
Looks good to me!

In the future, we could have an events workflow that has a matrix similar to the current workflow so we can easily add/remove entries as we need them maybe? Probably need to test this practice out for a few events before deciding if that's a real need or not.

PR #1591, opened 2022-08-04
Centralize the initContainer that fixes ownership

yuvipanda:
We were copy-pasting this, and it was causing shared-readwrite
to not be functional in some cases.

Fixes https://github.com/2i2c-org/infrastructure/issues/440

### DIFF
--- a/config/clusters/2i2c/utexas.values.yaml
+++ b/config/clusters/2i2c/utexas.values.yaml
@@ -89,6 +89,26 @@ jupyterhub:
           # So we put data in a subpath
           subPath: data
     initContainers:
+      # Repeat this here even though it's defined in basehub/values.yaml - otherwise
+      # tthe other initContainer we define here will overwrite this
+      - name: volume-mount-ownership-fix
+        image: busybox
+        command:
+          [
+            "sh",
+            "-c",
+            "id && chown 1000:1000 /home/jovyan && chown 1000:1000 /home/jovyan/shared-readwrite && ls -lhd /home/jovyan ",
+          ]
+        securityContext:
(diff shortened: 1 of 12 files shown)


yuvipanda:
image

Nice

config/clusters/2i2c/utexas.values.yaml
```diff
@@ -89,6 +89,26 @@ jupyterhub:
           # So we put data in a subpath
           subPath: data
     initContainers:
+      # Repeat this here even though it's defined in basehub/values.yaml - otherwise
+      # tthe other initContainer we define here will overwrite this
```
consideRatio, line 93: This addition is to 2i2c/utexas --- should it be centralized also?

yuvipanda: It needs to be repeated here, I left a comment explaining why there too - initContainers is a list so we can't just override it, as there is another initContainers here.

consideRatio: 👍👍👍

consideRatio (approved):

damianavila (approved):
This is exactly what I had in mind when I talked about this one at the cycle meeting.
Thanks for investigating it and fixing it, @yuvipanda!!

sgibson91:
I'll merge this one in since it has a few approvals

damianavila:
@yuvipanda, is there anything else you want to add here, otherwise, I think it is ready...

yuvipanda:
@damianavila ya just wanted to find a time when I've a few hours uninterrupted in case something goes wrong. Hopefully tomorrow.

PR #1593, opened 2022-08-04
Bump filestore capacity to match changes made through UI

sgibson91:
ref: https://2i2c.freshdesk.com/a/tickets/164

This PR is ensuring our Terraform state reflects changes made through the UI in response to the above freshdesk ticket

### DIFF
--- a/terraform/gcp/projects/pangeo-hubs.tfvars
+++ b/terraform/gcp/projects/pangeo-hubs.tfvars
@@ -11,7 +11,7 @@ config_connector_enabled = false
 
 # Setup a filestore for in-cluster NFS
 enable_filestore = true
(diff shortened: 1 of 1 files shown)


sgibson91:
I did a `--refresh-only` application of this config, so the state is now up-to-date

GeorgianaElena (approved):
Wondering if we should do the same for the neurohackademy nodepool size

damianavila:
> Wondering if we should do the same for the neurohackademy nodepool size

My inclination would be to say YES unless it is a transitory (event-related) mismatch? Mmm... not sure...
Wondering what other @2i2c-org/tech-team members think about it...

PR #1595, opened 2022-08-04
Update the docs about setting up terraform

GeorgianaElena:
I recently needed to re install and configure terraform and the `terraform init` command in the docs wasn't quite right (no workspaces were showing).

This updates the docs about ⬆️ and adds `terraform`as a requirement, to the initial setup guide.


sgibson91:
There are two things here and I'm not strongly opinionated on how we go about tackling them but have made some suggestions:

1) You probably needed to use the `-reconfigure` flag because there was an old `terraform.lock.hcl` file lying around in your directory. So rather than adding the flag as a "you must run it this way" thing, maybe a note saying something like "you might need this flag if you see `ERROR` and this is caused by the existence of the lock file"? (Deleting the lock file and generating a new one with `terraform init` is another way around needing the `-reconfigure` flag.)
2) The backends only apply to our GCP configuration atm and come from when I setup the pangeo-hubs cluster in Columbia's project. It was too difficult to be simultaneously authenticated to gcloud with both `@2i2c.org` and `@columbia.edu` emails and so I setup pangeo's own state file within their project. So we should note that if using terraform with Azure or AWS, a straightforward `terraform init` will suffice (no need to define a backend). And we should maybe move the `backends` folder under `terraform/gcp` to make this clear. Because if we end up needing another backend, it'll be because we've gotten into another Columbia-like situation and that would be a failure, IMO.

Just some thoughts and context around this, let me know if this resonates with you or not :)

GeorgianaElena:
Thanks @sgibson91! I've added a couple of commits to clarify the docs a bit more and moved the `backends` under `gcp`.

> 1. You probably needed to use the -reconfigure flag because there was an old terraform.lock.hcl file lying around in your directory. So rather than adding the flag as a "you must run it this way" thing, maybe a note saying something like "you might need this flag if you see ERROR and this is caused by the existence of the lock file"? (Deleting the lock file and generating a new one with terraform init is another way around needing the -reconfigure flag.)

The `-reconfigure` flag was  suggested in the docs actually. I didn't have any prior terraform files because I never used terraform on this laptop before. 
The problem was actually that the path to the file in `backend-config` implied running it from the wrong place, i.e. the `terraform` dir and not the `terraform/gcp`, the project I was trying to setup.

> The backends only apply to our GCP configuration atm and come from when I setup the pangeo-hubs cluster in Columbia's project. It was too difficult to be simultaneously authenticated to gcloud with both @2i2c.org and @columbia.edu emails and so I setup pangeo's own state file within their project. So we should note that if using terraform with Azure or AWS, a straightforward terraform init will suffice (no need to define a backend). And we should maybe move the backends folder under terraform/gcp to make this clear. Because if we end up needing another backend, it'll be because we've gotten into another Columbia-like situation and that would be a failure, IMO.

That makes sense 👍🏼 I moved the `backends` din under `gcp` and. updated the docs to match that a bit more.

docs/topic/terraform.md
```diff
@@ -8,12 +9,43 @@ The source files for each provider we currently deploy with terraform are under
 
 - [`terraform/gcp`](https://github.com/2i2c-org/infrastructure/tree/HEAD/terraform/gcp)
 - [`terraform/azure`](https://github.com/2i2c-org/infrastructure/tree/HEAD/terraform/azure)
+- [`terraform/aws`](https://github.com/2i2c-org/infrastructure/tree/HEAD/terraform/aws)
 
 Each provider directory also has a `projects` subdirectory where variables defining each cluster are stored.
 
+## Backends
+
+A [backend](https://www.terraform.io/language/settings/backends/configuration) defines where Terraform stores its state data files. Normally, terraform automatically manages state in the workspaces associated with your configuration, and there's no need to explicity define a backend.
```
sgibson91, line 18:
```suggestion
A [backend](https://www.terraform.io/language/settings/backends/configuration) defines where Terraform stores its state data files in cloud storage (we use a GCP bucket for the majority of our deployments).
Normally, we are using Terraform when logged in with our 2i2c accounts and there's no need to explicitly define a backend.
```

docs/topic/terraform.md
```diff
@@ -8,12 +9,43 @@ The source files for each provider we currently deploy with terraform are under
...
+
+But for some projects where we don't have access to using our 2i2c accounts, e.g. universities that require us to have specific university-affiliated identities, we can configure different backends to access the terraform state stored in those projects.
+Working this way saves us the pain of trying to work with terraform using two different authentications.
+
+These backend configs are stored in [`terraform/gcp/backends`](https://github.com/2i2c-org/infrastructure/tree/HEAD/terraform/gcp/backends)
```
sgibson91, line 23:
```suggestion
These backend configs are stored in [`terraform/gcp/backends`](https://github.com/2i2c-org/infrastructure/tree/HEAD/terraform/gcp/backends).
Currently, we only implement a choice of backend for GCP deployments.
```

docs/topic/terraform.md
```diff
@@ -8,12 +9,43 @@ The source files for each provider we currently deploy with terraform are under
...
+```bash
+cd terraform/gcp
+terraform init -backend-config=backends/pangeo-backend.hcl
+```
+````
+
 ```{note}
-Currently, only our GCP terraform config uses different backends.
-If you are working on an Azure cluster, you can use `terraform init` as normal.
+If prior backend data exists in a `terraform.lock.hcl`, you might see an `ERROR` when trying to initialize that backend. To reconfigure this backend, ignoring any saved configuration, add the `-reconfigure` flag to the init command.
```
sgibson91, line 48: Can we remember what the error actually was and add is here instead of `ERROR`

sgibson91 (approved):
Couple of suggestions but this LGTM! Thank you!

docs/topic/terraform.md
GeorgianaElena: I'm not sure I ever seen that error, hmm. I believe I always used the `-reconfigure` flag in my terraform init command, because I just copy pasted it from the docs, without putting any thought into it 😅 

Maybe it's this one I found on the internet (https://stackoverflow.com/questions/68159962/cannot-convert-terraform-remote-state-in-s3-backend-to-local-state-with-terrafo)?:

```
Error: Backend configuration changed
│ 
│ A change in the backend configuration has been detected, which may require migrating existing state.
│ 
│ If you wish to attempt automatic migration of the state, use "terraform init -migrate-state".
│ If you wish to store the current configuration with no changes to the state, use "terraform init -reconfigure".
```

sgibson91: Yes, I think that's the one! I think you could just swap out `ERROR` for `Error: Backend configuration changed` rather than paste the whole thing, as we are providing the information about using the `-reconfigure` flag as the solution

GeorgianaElena:
```suggestion
If prior backend data exists in a `terraform.lock.hcl`, you might see an `Error: Backend configuration changed` when trying to initialize that backend. To reconfigure this backend, ignoring any saved configuration, add the `-reconfigure` flag to the init command.
```

GeorgianaElena:
Thanks for the knowledge share @sgibson91 💖

PR #1599, opened 2022-08-05
OceanHackWeek image updates

abkfenris:
Updates to both the Python and R images for OceanHackWeek.

I was hoping to see the image updater (#1588) do it's thing, but it [didn't appear to recognize the new tags on the run after these were pushed](https://github.com/2i2c-org/infrastructure/runs/7683672070?check_suite_focus=true#step:4:19)?

### DIFF
--- a/config/clusters/2i2c/ohw.values.yaml
+++ b/config/clusters/2i2c/ohw.values.yaml
@@ -47,7 +47,7 @@ basehub:
           description: "~2 CPU, ~8G RAM"
           default: true
           kubespawner_override:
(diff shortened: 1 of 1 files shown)


GeorgianaElena (approved):
Thanks @abkfenris! I will look into what's going on with the action today.

GeorgianaElena:
I'm planning to use the latest tags pushed for testing, so I'll probably delay merging this a bit, now that the hub its not used.

abkfenris:
@GeorgianaElena thank you!

PR #1630, opened 2022-08-16
For version bumping workflows, add dry-run input for manual triggers

sgibson91:
This allows manual deploys of the actions that bump helm subchart versions or docker image tags to perform a dry-run, i.e., not open a Pull Request. This can be useful for testing. When dry-run is enabled, information about the bump will be printed to stdout.


sgibson91:
Merging because the upstream change I made (https://github.com/sgibson91/bump-jhub-image-action/pull/155) now means our workflow is broken: https://github.com/2i2c-org/infrastructure/runs/7857740467?check_suite_focus=true

PR #1633, opened 2022-08-17
Add workflow job to find specific workflow run URL when a PR is merged

sgibson91:
### Motivation

We currently have a workflow that posts this link on merged PRs that require a hub redeployment: https://github.com/2i2c-org/infrastructure/actions/workflows/deploy-hubs.yaml?query=branch%3Amaster However, this is a _generic_ link that shows the most recent runs of that workflow on that branch. If we were looking through old PRs, it would be difficult to associate a given PR with the specific workflow run since new merges would have triggered new runs which would have populated the top of the list.

This PR is adding a new job to the `comment-test-link-merged-pr.yaml` workflow file which uses the GitHub API to identify which _specific_ run was triggered by merging a PR and return the _specific_ URL, which will end in something like `.../actions/runs/<run_id>`. We list the workflow runs for the repo and filter the results down by the following criteria:

- event that triggered the workflow was `push`
- branch associated with the workflow is our default `master`
- the workflow was created "today" (where "today" is calculated dynamically when this workflow runs)
- the name of the workflow matches our target workflow, i.e., "Deploy and test hubs" (defined in the `deploy-hubs.yaml` file)
- the head commit message of this workflow and our target workflow match (In the previous `get-pr-number` job, we will have already established that the head commit contains a merge and extracted the PR number from that message)

### Overview of the whole workflow after this PR is merged

1. `get-pr-number` job:
   - Establishes that the head commit that triggered the workflow is a merge commit and extracts the number of the PR from the message
   - Sets output variables `continue-workflow` and `pr-number`
2. `get-workflow-url` job
   - Only runs if `get-pr-number` job completes successfully and `continue-workflow == True`
   - Locates the `.../actions/runs/<run_id>` URL associated with the PR identified in the previous job
   - Sets the URL as the output variable `workflow-url`
3. `add-pr-comment` job
   - Only runs if both `get-pr-number` and `get-workflow-url` complete successfully, `continue-workflow == True` and `workflow-url != False`
   - Consumes `pr-number` and `workflow-url` variables to post the URL as a comment on the PR

PR #1635, opened 2022-08-17
Iterate through the workflow_runs object instead

sgibson91:
Follow up to #1633 

Fixes a broken workflow, e.g., https://github.com/2i2c-org/infrastructure/actions/runs/2876499696

Example of functional workflow: https://github.com/sgibson91/testing-gh-actions/blob/main/.github/workflows/comment-link-to-merged-pr-2.yaml


sgibson91:
I will self-merge this since it should fix a broken workflow

Issue #1611, opened 2022-08-08
Deployer failing to create deployment-service-check user on openscapes


sgibson91:
### Context

The deployer is failing to create the deployment-service-check user (specifically on the staging hub) hence why CI/CD is failing.

```
========================================= FAILURES ==========================================
_____________________________________ test_hub_healthy ______________________________________

hub_url = 'https://staging.openscapes.2i2c.cloud'
api_token = 'xxxx'
notebook_dir = PosixPath('/Users/sgibson/source/github/2i2c-org/infrastructure/deployer/tests/test-notebooks/daskhub')
check_dask_scaling = False

    @pytest.mark.asyncio
    async def test_hub_healthy(hub_url, api_token, notebook_dir, check_dask_scaling):
        try:
            print(f"Starting hub {hub_url} health validation...")
            for root, directories, files in os.walk(notebook_dir, topdown=False):
                for i, name in enumerate(files):
                    # We only want to run the "scale_dask_workers.ipynb" file if the
                    # check_dask_scaling variable is true. We continue in the loop if
                    # check_dask_scaling == False when we iterate over this file.
                    if (not check_dask_scaling) and (name == "scale_dask_workers.ipynb"):
                        continue

                    print(f"Running {name} test notebook...")
                    test_notebook_path = os.path.join(root, name)
                    await check_hub_health(hub_url, test_notebook_path, api_token)

            print(f"Hub {hub_url} is healthy!")
        except Exception as e:
            print(
                f"Hub {hub_url} not healthy! Stopping further deployments. Exception was {e}."
            )
>           raise (e)

deployer/tests/test_hub_health.py:84:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
deployer/tests/test_hub_health.py:77: in test_hub_healthy
    await check_hub_health(hub_url, test_notebook_path, api_token)
deployer/tests/test_hub_health.py:45: in check_hub_health
    await execute_notebook(
/usr/local/Caskroom/miniconda/base/envs/infrastructure/lib/python3.10/site-packages/jhub_client/execute.py:153: in execute_notebook
    return await execute_code(hub_url, cells, **kwargs)
/usr/local/Caskroom/miniconda/base/envs/infrastructure/lib/python3.10/site-packages/jhub_client/execute.py:79: in execute_code
[...]

PR #1639, opened 2022-08-19
Fix spawning a server for the deployment-service-check user on openscapes hubs

GeorgianaElena:
Temporarily fixes https://github.com/2i2c-org/infrastructure/issues/1611
 
Hopefully, https://github.com/jupyterhub/kubespawner/pull/631 is the more sustainable solution, but until/if that gets merged/released, this PR hacks the test so that we're passing a dict of `user-options` when spawning a test server on the openscapes hub.

More details in [this comment](https://github.com/2i2c-org/infrastructure/issues/1611#issuecomment-1219466905).

### DIFF
--- a/config/clusters/openscapes/common.values.yaml
+++ b/config/clusters/openscapes/common.values.yaml
@@ -68,6 +68,7 @@ basehub:
         # on these nodes.
         - display_name: "Small: m5.large"
           description: "~2 CPU, ~8G RAM"
+          slug: "small"
           profile_options: &profile_options
             image:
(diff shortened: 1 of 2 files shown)


sgibson91 (approved):
Thanks for investigating this and preparing this fix!! I didn't have time to respond to your comment on the issue yesterday as I was going to suggest we add a `if: matrix.cluster_name != 'openscapes'` condition to the testing step in the prod job as it was technically easier. But since you've already done the work, let's merge this instead!

Issue #1636, opened 2022-08-17
[coudbank ccsf] RetroLab as Default Interface


shawnwiggins:
### Context

We made some updates to the CCSF JupyterHub image and now have RetroLab working, but we need support configuring the hub to default to RetroLab instead of JupyterLab since it is not a menu item on the Configurator.

- [Hub Image Repo](https://github.com/ccsf-math/jupyterhub)
- [CCSF Hub](https://ccsf.cloudbank.2i2c.cloud/)

This is just a small issue to help students see a less complex interface initially and still be able to utilize the JupyterLab framework until Notebook fully merges with Retro. Our Data Science courses start the week of Aug 29.

### Proposal

If possible, it would be great to have a RetroLab option on the Configurator.

shawnwiggins:
Thank you, @GeorgianaElena! @yuvipanda created this GitHub Issue for me, but I'll make sure to use that support page in the future. 👍🏻

PR #1645, opened 2022-08-22
Add retroLab as a configurator option and default to it for ccsf hub

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/1636


GeorgianaElena:
I've already deployed this and set retro as the default interface.

sgibson91 (approved):
LGTM!

PR #1653, opened 2022-08-25
Update Decommission Hub template

sgibson91:
Adds an item to delete the terraform workspace when removing a cluster


sgibson91:
I'll merge this without review because I don't think it's controversial

PR #1642, opened 2022-08-19
Enable selecting the environment to use for each machine profile on AWS carbonplan hubs

sgibson91:
fixes https://github.com/2i2c-org/infrastructure/issues/1640
fixes https://github.com/2i2c-org/infrastructure/issues/1641

PR #1670, opened 2022-08-31
Fix CarbonPlan hub health check failures

sgibson91:
After deploying #1642, I ran into the same error message I reported in #1611. Since both the CarbonPlan and Openscapes hubs allow choosing of environment images for each machine profile, I figured it was the same bug and applied the fix from #1639 to the CarbonPlan hubs.

### DIFF
--- a/config/clusters/carbonplan/common.values.yaml
+++ b/config/clusters/carbonplan/common.values.yaml
@@ -44,6 +44,7 @@ basehub:
         # on these nodes.
         - display_name: "Small: r5.large"
           description: "~2 CPU, ~15G RAM"
+          slug: "small"
           default: true
           profile_options: &profile_options
             image:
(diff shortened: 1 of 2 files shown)


GeorgianaElena (commented):
This looks to me like it should work 😕 I can try check the logs and look deeper into why this doesn't work tomorrow if that's ok.

sgibson91:
Which logs need to be inspected? I had `k logs -c hub hub-xxxx --follow` running while I ran the test, but it wasn't very informative, just showed the startup time increasing

sgibson91:
I've figured out the issue. We have this schema `[...].value.slug: value`. It wasn't working because I had something like: `[...].python_trace.slug: python` when it should have been: `[...].python.slug: python`. I can see the deployment-service-check pod pending now. I'll push a new commit.

sgibson91:
Wooo, passing locally now!

GeorgianaElena (approved):
🚀 🚀 🚀

PR #1685, opened 2022-09-06
Fix Callysto prod hub's display name

sgibson91:
This affects the [list of running hubs](https://infrastructure.2i2c.org/en/latest/reference/hubs.html) and just makes the table more readable


sgibson91:
I'll merge this without review because it is a minor change

sgibson91:
CI/CD failed on this PR, but we actually didn't need it to run since `display_name` only affects docs, not the actual deployment.

I opened an issue regarding why (a lot of) CI/CD failed as it's a recurring issue that I haven't been able to figure out yet

* https://github.com/2i2c-org/infrastructure/issues/1686

PR #1701, opened 2022-09-12
Change how `comment` variable is defined in `comment-deployment-plan-pr.yaml` workflow file

sgibson91:
In CI/CD, this was breaking due to [`comment` not being set](https://github.com/2i2c-org/infrastructure/runs/8297082132?check_suite_focus=true), which I don't believe should happen using the `next(...)` function. This PR makes the code look more like the block that extracts the artifact ID further up in the code in the hopes that this will alleviate this issue.

### DIFF
--- a/.github/workflows/comment-deployment-plan-pr.yaml
+++ b/.github/workflows/comment-deployment-plan-pr.yaml
@@ -124,25 +124,25 @@ jobs:
 
           # Find if a deployment-plan comment has been previously posted
           for page in issue_comments:
-              comment = next(
+              comment_id = next(
                   (
-                      comment
+                      comment["id"]
                       for comment in page
-                      if comment.user.login == "github-actions[bot]"
-                      and "<!-- deployment-plan -->" in comment.body
+                      if comment["user"]["login"] == "github-actions[bot]"
+                      and "<!-- deployment-plan -->" in comment["body"]
                   ),
-                  False,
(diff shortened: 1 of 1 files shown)


yuvipanda (approved):

GeorgianaElena (approved):
> In CI/CD, this was breaking due to [comment not being set](https://github.com/2i2c-org/infrastructure/runs/8297082132?check_suite_focus=true), which I don't believe should happen using the next(...)

Hmm, maybe if the iterator raises an exception, then default is not set and the var remains undefined?

Anyway, this looks LGTM 🚀

sgibson91:
Thanks @GeorgianaElena I'll bear that in mind if this doesn't resolve the issue :)

GeorgianaElena:
@sgibson91, I believe it is now complaining about `comment_id` https://github.com/2i2c-org/infrastructure/actions/runs/3061536788/jobs/4941424873 😢

sgibson91:
Bleugh! Thank you for bringing to my attention. Part of me wants to scrap using `ghapi` completely and just default to the `requests` library. `ghapi` had such potential, but it is poorly documented and obscures the error messages from the GitHub API itself.

damianavila:
> ghapi had such potential, but it is poorly documented and obscures the error messages from the GitHub API itself.

If it is biting us...

Issue #1439, opened 2022-06-20
[New Hub] Callysto


colliand:
### Hub Description

2i2c will work with @ianabc from PIMS and @byrcyb from Cybera (perhaps others?) to transition the current Callysto hub to a new service managed by 2i2c. 

The Callysto hub should be a "standard education hub" since it will be used by students and teachers in grades 5-12 who are making their initial forays into data-intensive studies.

### Community Representative(s)

@ianabc @byrcyb

### Important dates

_**Notes: target and required dates are outdated, we need to update them accordingly to new information and prioritization.**_

- **Target start date**: ~2022-07-08~ end of August?
- **Required start date**: ~2022-07-15~ mid September?
- **Any important dates for usage**: Fall term (end of September)

### Hub Authentication Type

Other (may not be possible, please specify in comments)

### Hub logo information

- **URL to Hub Image**: https://www.callysto.ca/wp-content/uploads/2022/08/Callysto-HUB_vertical.png
- **URL for Image Link**: https://www.callysto.ca

### Hub user image

- Repository for user image: { REPO LINK IF IT EXISTS }
- User image registry: https://hub.docker.com/repository/docker/callysto/2i2c
- User image tag and name: callysto/2i2c-notebook:latest

### Extra features you'd like to enable

- [x] Specific cloud provider or datacenter: must be in Canada; need input here from @byrcyb
- [ ] Dedicated Kubernetes cluster
- [ ] Scalable Dask Cluster

### Hub URL

<hub ID>.<hub cluster>.2i2c.cloud

### Tasks to deploy the hub

- [x] Engineer who will deploy the hub is assigned
- [x] Deploy information filled in above
- [x] Initial Hub deployment PR: https://github.com/2i2c-org/infrastructure/pull/1649
- [ ] Administrators able to log on
- [ ] Community Representative satisfied with hub environment
- [ ] Hub now in steady-state

damianavila:
@colliand request ack. I have added the issue to our backlog board and we will prioritize the hub deployment accordingly to available eng resources so we can deploy the hub in a timely manner.

@ianabc @byrcyb, we will ping you soon with some questions about the specific of the hub deployment.

yuvipanda:
Who would manage the GCP billing account?

byrcyb:
[...]

Issue #1697, opened 2022-09-09
Custom jupyter-server templates


GeorgianaElena:
### Context

Currently we are able to provide custom templates for the hubs through https://github.com/2i2c-org/default-hub-homepage.

But a recent request to customize the look of  the single user's server from https://github.com/2i2c-org/infrastructure/issues/1439#issuecomment-1238790781 generated this issue.

Do we want to provide customizations for [`jupyter-server` templates](https://github.com/jupyter-server/jupyter_server/tree/main/jupyter_server/templates)? If so, where should we host this custom templates:? 
- in a new repo similar to https://github.com/2i2c-org/default-hub-homepage
- or in the same one?

damianavila:
@GeorgianaElena come up with a compromised solution in #1710.

> But as I mentioned in the comments above, this implementation is a compromise. This was my own take on making this to work given the available time and feedback (at that time I considered this to be a time sensitive request, but turns out it wasn't).

In that PR I raised this comment:

> Regardless of the specifics of the implementation, I have the feeling we are entering into the user image territory... It should relatively easy for them to self-serve the customization of the templates on the user image itself, right?

When I saw the problem I was immediately thinking about customization of the templates at the user image level.
Now re-reading the first message of this thread, I understand there might be a second option to allow community customization but still own the template repo as we do with the [default-hub-homepage](https://github.com/2i2c-org/default-hub-homepage).

I am still not convinced about the second option because the `tree` and `notebook`, IMHO, belong to the user territory but happy to be convinced otherwise :wink.

GeorgianaElena:
It is possible to use [singleuser.extraFiles config](https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html#singleuser-extrafiles) to customize certain jupyter templates. That is not suitable for complex UIs, but since we weren't requested to make complex changes, let's close this and reopen/open a new issue when/if this will be needed.

PR #1710, opened 2022-09-15
Personalize callysto singleuser server page

GeorgianaElena:
This uses the [`singleuser.extraFiles` config](https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html#singleuser-extrafiles) to add a custom logo for the singleuser server dashboard and notebook, i.e. the Callysto one and hides the `Clusters` tab from the dashboard.

The Hub interface has been customized too through the https://github.com/2i2c-org/default-hub-homepage/tree/callysto-staging, but only for the staging hub at the moment. I will add a new branch `callysto-prod` to override the prod hub also.

Closes https://github.com/2i2c-org/infrastructure/issues/1439

Note that I will leave a couple of comments in the code to explain some of the assumptions/compromises I've made.

### DIFF
--- a/config/clusters/callysto/common.values.yaml
+++ b/config/clusters/callysto/common.values.yaml
@@ -33,7 +33,7 @@ jupyterhub:
       name: callysto/2i2c
       tag: 0.1.0
     extraFiles:
-      page.html:
+      tree.html:
         mountPath: /usr/local/share/jupyter/custom_template/tree.html
         stringData: |
           {% extends "templates/tree.html" %}
--- a/config/clusters/callysto/common.values.yaml
+++ b/config/clusters/callysto/common.values.yaml
@@ -32,6 +32,28 @@ jupyterhub:
     image:
       name: callysto/2i2c
       tag: 0.1.0
+    extraFiles:
+      page.html:
+        mountPath: /usr/local/share/jupyter/custom_template/tree.html
+        stringData: |
+          {% extends "templates/tree.html" %}
+          {% block header %}
+          <style>
+          .clusters_tab_link {
+            visibility: hidden;
+          }
+          </style>
+          {% endblock %}
+          {% block logo %}
+            <img src="https://www.callysto.ca/wp-content/uploads/2022/08/Callysto-HUB_horizontal.png" alt="CallystoHub"/>
+          {% endblock %}
+      notebook.html:
+        mountPath: /usr/local/share/jupyter/custom_template/notebook.html
+        stringData: |
+          {% extends "templates/notebook.html" %}
+          {% block logo %}
+            <img src="https://www.callysto.ca/wp-content/uploads/2022/08/Callysto-HUB_horizontal.png" alt="CallystoHub"/>
+          {% endblock %}
   hub:
     config:
       JupyterHub:
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -206,6 +206,9 @@ jupyterhub:
             # this is what allows us to shutdown servers
             # when people leave a notebook open and wander off
             cull_connected: true
+          NotebookApp:
+            extra_template_paths:
+              - /usr/local/share/jupyter/custom_template
     startTimeout: 600 # 10 mins, because sometimes we have too many new nodes coming up together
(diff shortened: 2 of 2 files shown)


config/clusters/callysto/common.values.yaml
```diff
@@ -32,6 +32,28 @@ jupyterhub:
     image:
       name: callysto/2i2c
       tag: 0.1.0
+    extraFiles:
+      tree.html:
```
GeorgianaElena, line 36: Normally, extending only the `page.html` template would have been enough since both the `tree` and the `notebook` templates are extending it themselves. But for some reason, it turned impossible to make it to work like this... I fear this is because of NotebookApp -> ServerApp? But I have tried everything with no luck, so this is the best version that's actually doing the job

config/clusters/callysto/common.values.yaml
```diff
@@ -32,6 +32,28 @@ jupyterhub:
...
+        mountPath: /usr/local/share/jupyter/custom_template/tree.html
+        stringData: |
+          {% extends "templates/tree.html" %}
+          {% block header %}
+          <style>
+          .clusters_tab_link {
+            visibility: hidden;
+          }
+          </style>
```
GeorgianaElena, line 45: Hmm, actually this one is only possible in "tree.html", so we would have needed two templates anyway

helm-charts/basehub/values.yaml
```diff
@@ -206,6 +206,9 @@ jupyterhub:
             # this is what allows us to shutdown servers
             # when people leave a notebook open and wander off
             cull_connected: true
+          NotebookApp:
+            extra_template_paths:
+              - /usr/local/share/jupyter/custom_template
```
GeorgianaElena, line 211: I'm setting this one in the `basehub` template because it does no harm if we set it here and we provide no custom templates.

The benefit of this being here is that we can inject `extraFiles` for other hubs too to extend the templates if need be in the future.

Or at least until will decide on a better strategy as part of https://github.com/2i2c-org/infrastructure/issues/1697 that might imply sync-ing the templates from a repo on the fly like we do with the hub templates.

sgibson91 (approved):
LGTM!

damianavila (commented):
Regardless of the specifics of the implementation, I have the feeling we are entering into the user image territory... It should relatively easy for them to self-serve the customization of the templates on the user image itself, right?

GeorgianaElena:
Thanks for the feedback @sgibson91, @damianavila!

> I have the feeling we are entering into the user image territory... It should relatively easy for them to self-serve the customization of the templates on the user image itself, right?

@damianavila , I don't think it has to do with user image, but it's instead a separate thing. And I agree this should be a feature that could be customized by the  community instead of us (similar to the hub templates customization).
But as I mentioned in the comments above, this implementation is a compromise. This was my own take on making this to  work given the available time and feedback (at that time I considered this to be a time sensitive request, but turns out it wasn't).

I suggest sharing any thoughts about possible ways to implement this in a more sustainable way, in #1697 and consider this an workaround until a clear path forward arises from #1697. WDYT?

In the meantime I will merge this since I have an approval and want to ship this for callysto as promised. But feel free to revert or share any other thoughts about the implementation specifics in #1697. Thanks!

damianavila:
> But as I mentioned in the comments above, this implementation is a compromise. This was my own take on making this to work given the available time and feedback (at that time I considered this to be a time sensitive request, but turns out it wasn't).

I understand, thanks for re-raising that point. I think it is a good compromise (you may notice I did not actually requested changes to the PR, I just dropped a general opinion.

> I suggest sharing any thoughts about possible ways to implement this in a more sustainable way, in https://github.com/2i2c-org/infrastructure/issues/1697 and consider this an workaround until a clear path forward arises from https://github.com/2i2c-org/infrastructure/issues/1697. WDYT?

LGTM, I will reference my comment on that issue!

> In the meantime I will merge this since I have an approval and want to ship this for callysto as promised. But feel free to revert or share any other thoughts about the implementation specifics in https://github.com/2i2c-org/infrastructure/issues/1697. Thanks!

No need to revert anything! Thanks again for working on this.

PR #1713, opened 2022-09-16
Keep a placeholder node around for UToronto Hub

yuvipanda:
Users were reporting slow starts and timeouts in the morning, as many users come on at the same time. This change keeps a placeholder *node* around, with a placeholder pod that will get displaced whenever a user needs that node. This should increase the odds of a new node being up by the time more than 1 node worth of users pop in.

Ref https://2i2c.freshdesk.com/a/tickets/201

### DIFF
--- a/config/clusters/utoronto/prod.values.yaml
+++ b/config/clusters/utoronto/prod.values.yaml
@@ -1,4 +1,15 @@
 jupyterhub:
+  scheduling:
+    userPlaceholder:
+      # Keep at least one spare node around
+      replicas: 1
+      resources:
+        requests:
+          # Each node on the UToronto cluster has 59350076Ki of RAM
+          # You can find this out by looking at the output of `kubectl get node <node-name> -o yaml`
+          # Look under `allocatable`, not `capacity`
+          # So even though this is under `userPlaceholder`, it really is operating as a `nodePlaceholder`
+          memory: 57350076Ki
   hub:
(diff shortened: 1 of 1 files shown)


config/clusters/utoronto/prod.values.yaml
```diff
@@ -1,4 +1,15 @@
...
+    userPlaceholder:
+      # Keep at least one spare node around
+      replicas: 1
+      resources:
+        requests:
+          # Each node on the UToronto cluster has 59350076Ki of RAM
+          # You can find this out by looking at the output of `kubectl get node <node-name> -o yaml`
+          # Look under `allocatable`, not `capacity`
+          # So even though this is under `userPlaceholder`, it really is operating as a `nodePlaceholder`
+          memory: 57350076Ki
```
damianavila, line 12: Did you end up with this number taking into consideration stuff running on those nodes by defaault or you are guess-estimating here? I am just curios... the approach LGTM.

damianavila (approved):

config/clusters/utoronto/prod.values.yaml
yuvipanda: @damianavila updated comment with how :)

damianavila: I suspected that was the case, nice to have it written! Thanks!

Issue #1718, opened 2022-09-19
Increase memory limit for AUP hub to 6GB


jmunroe:
### Context

Original ticket: https://2i2c.freshdesk.com/a/tickets/202

> I am from the American University of Paris and we have been using
successfully a JupyterHub provided by you.
> I am wondering if it might be possible to increase the memory limit of
2GB for 6 students during this semester.

I have re-engaged with AUP to work towards settting up a formal contract. On the assumption that we will be able to eventually come to agreement, we should increase the amount of RAM available right away to allow their students to not be impacted.

### Proposal

I leave it to engineering to decide if "6GB" is the right upper bound or if it should be a soft limit of 6GB and hard limit of 8GB when making the appropriate changes in the JupyterHub configuration files and the Terraform configuration files. 

### Updates and actions

[ ] Changes made in appropriate configuration files.
[ ] Update pushed to hub
[ ] Follow up with community partner and close the ticket

PR #1720, opened 2022-09-20
Increase mem limits for aup

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/1718, guarantees 6G per request and adds a stretch on the limit to 8G


damianavila (approved):
LGTM

PR #1725, opened 2022-09-21
Init `comment_id` in comment plan workflow

GeorgianaElena:
I believe an empty `issue_comments` might be what's causing the undefined `comment_id` issues?

Ref: https://github.com/2i2c-org/infrastructure/pull/1701

### DIFF
--- a/.github/workflows/comment-deployment-plan-pr.yaml
+++ b/.github/workflows/comment-deployment-plan-pr.yaml
@@ -121,7 +121,7 @@ jobs:
               issue_number=pr_number,
               per_page=100,
           )
-          
+
           # Otherwise, if no `issue_comments`, this will be undefined
           comment_id = None
           # Find if a deployment-plan comment has been previously posted
--- a/.github/workflows/comment-deployment-plan-pr.yaml
+++ b/.github/workflows/comment-deployment-plan-pr.yaml
@@ -121,7 +121,9 @@ jobs:
               issue_number=pr_number,
               per_page=100,
           )
-
+          
+          # Otherwise, if no `issue_comments`, this will be undefined
+          comment_id = None
           # Find if a deployment-plan comment has been previously posted
           for page in issue_comments:
               comment_id = next(


sgibson91:
I'll come back to this, but I'm not wholly convinced that predefining `comment_id` should be required given that we are using `next`. I understood it that it would use the default arg provided (second arg) if the first arg didn't return anything (and I believe that this understanding is consistent with how we use `next` in other parts of the codebase). And hence I believe there's something a little more sinister going on with `ghapi` in the loop that I'd like to investigate as well.

sgibson91:
https://docs.python.org/3/library/functions.html#next

GeorgianaElena:
Totally agree about the `next` functionality thing. But that's under a `for` statement:
```for page in issue_comments:```, and if `issue_comments` is None, than the `next` isn't  even called I beleive.

GeorgianaElena:
Sorry if I'm not making any sense btw. Wished we were in front of the same screen and debug this together ✨

sgibson91:
Yeah, I need to verify my beliefs of what `issue_comments` returns. For example, I believe that if an issue exists, then the top comment is also treated as an issue comment, so then `len(issue_comments) == 1` always and that singular item is a dict of metadata about the issue top comment. That is either not the case, or something else is going wrong meaning we get something that breaks the code saved to `issue_comments` instead.

sgibson91:
What you're saying makes sense, but I expect it's something sinister that is causing this issue that I want to investigate a little bit more and decide that this is indeed the correct fix.

I also think I'll strip out ghapi.

GeorgianaElena:
> I also think I'll strip out ghapi.

👍🏼 Let's close this then.

sgibson91:
So this workflow worked here: https://github.com/2i2c-org/infrastructure/pull/1727#issuecomment-1255529787 and it has multiple comments beforehand.

Hence I think the below statement is False.

> I believe that if an issue exists, then the top comment is also treated as an issue comment, so then `len(issue_comments) == 1` always and that singular item is a dict of metadata about the issue top comment.

And instead this statement is True.

> if `issue_comments` is None, than the `next` isn't even called I beleive.

Hence I believe we should reopen this PR and merge :)

Sorry for taking you around the houses @GeorgianaElena!!

GeorgianaElena:
No worries @sgibson91 and thanks for thinking about it still

sgibson91 (approved):

damianavila (approved):

Issue #1485, opened 2022-06-29
[New Hub] London Interdisciplinary School


choldgraf:
This is an issue to collect information and coordinate deploying a new hub! We should:

- [ ] Fill out the missing information below in collaboration with the Community Representative
- [ ] Assign an engineer to the hub and deploy it in the cloud

### Roles

- Hub shepherd: @choldgraf
- Hub engineer: `{{ @XXX }}`
- PR for deployment: `{{ XXX }}`

### About the community

**Community Representative(s)** ([reference docs](https://docs.2i2c.org/en/latest/about/service/team.html#community-roles))

- @LaCrecerelle
- @matthew-brett 

**Important Dates**
- **Target start date**: `{{ date }}` 
- **Required start date**: `{{ date }}` 
- **Any important dates for usage**: `{{ date }}` 

### About the infrastructure

**Hub Authentication Type** ([reference docs](https://docs.2i2c.org/en/latest/admin/howto/manage-users.html?highlight=authentication#authentication)): `{{ github / google / cilogon / etc }}`

**Hub Logo**

- URL to Image: `{{ URL }}`
- URL for image link: `{{ URL }}`

**User image** ([**reference docs**](https://docs.2i2c.org/en/latest/admin/howto/environment/hub-user-image-template-guide.html))

- Image repository: `{{ URL }}`
- Image registry: `{{ URL }}`
- Image tag and name: `{{ NAME AND TAG IF IT EXISTS }}`

**Extra features**

- Dedicated kubernetes cluster: `{{ YES / NO }}`
- Preferred cloud provider: No requirement, preferably not Google
- Scalable Dask Gateway Cluster: `{{ YES / NO }}`

### JupyterHub information

- Hub URL:` {{ <hub ID>.<hub cluster>.2i2c.cloud }}`
- Helm chart: `{{ basehub / daskhub }}`

### Deployment checklist

- [ ] Administrative setup for community is finished
- [ ] All information above is filled in
- [ ] Engineer who will deploy the hub is assigned
- [ ] Initial Hub deployment PR is added
- [ ] Administrators able to log on
- [ ] Community Representative satisfied with hub environment
- [ ] Hub now in steady-state

choldgraf:
ping to @LaCrecerelle and @matthew-brett - we can use this hub to coordinate around the information needed to deploy the hub in your community. There are a number of missing fields above, could you take a quick look and help us fill in the missing information? When that's done, we can get started

[...]

PR #1727, opened 2022-09-22
London Interdisciplinary School cluster and hub

GeorgianaElena:
- [ ] cluster
- [ ] hub

Closes https://github.com/2i2c-org/infrastructure/issues/1485


yuvipanda:
One question I have is - is this going to be a *shared 2i2c cluster* that has LIS, or something specific just to LIS? And if it is specific just to LIS, what does that mean for pricing? @jmunroe @damianavila might have answers?

If it's going to be a shared cluster, we should probably name this differently. Otherwise, the content of this looks great!

jmunroe:
> If it's going to be a shared cluster, we should probably name this differently. Otherwise, the content of this looks great!

From what I've seen been able to see, this is indeed a "shared" cluster but it just so happens to be the first one we are deploying in GCP europe-west2.  I don't want to complicate this deployment but could it be instead be:

`name: 2i2c-gcp-europe-west2`

and then list `lis-uk` as one of the hubs running on this new cluster?

yuvipanda:
Yeah, I'd suggest calling it 2i2c-uk or similar and treating that as a shared cluster.

GeorgianaElena:
I believed this will be a `lis` only hub, and the `shared` cluster was a misunderstanding because it wasn't clear that our shared cluster is deployed in the us.  But I will change the naming, sorry for the confusion.

GeorgianaElena:
Also, is everyone ok with having a `staging` hub on this cluster too?

damianavila:
My understanding here is that we are creating a new cluster **because** of a hard requirement, then this is effectively converting from a "hub in a shared cluster" into a "hub in a dedicate" one and this is why I believe it should have a markup as our documentation indicates.

I shared the same thoughts in the leads issue: https://github.com/2i2c-org/leads/issues/80#issuecomment-1255475254.

damianavila:
> this is why I believe it should have a markup as our documentation indicates.

Although if there were previous conversations and agreements around this during the contract negotiation, we should adhere to what we have agreed on regardless of my thoughts/opinions 😉.

yuvipanda:
I think there are two questions here:

1. Are we deploying a *new shared cluster* in the UK, for use by LIS and other potential UK users in the future? This won't be the first time someone wants data to stay just in the US
2. Are we deploying a *new cluster* just for LIS?

IMO, given the existing contract negotiation, the small size of the hub as well as the fact we probably will get other hub users in the future who want data to stay in the UK (or at least, *not* in the US), it makes sense to frame this as 'we are creating a new shared cluster in the UK, and LIS is the first hub on that'.

yuvipanda:
And +1 on a staging hub here.

GeorgianaElena:
Updates:

- I've teared down the previous gcp project and the lis cluster and created new ones under different names
- hope the naming is ok now
- didn't deploy a staging hub but will do so later
- I probably messed something in the support chart or dns settings because navigating to the new grafana shows a cert problem. Too tired to investigate right now, but at least the LIS hub is running again

config/clusters/2i2c-uk/support.values.yaml
```diff
@@ -0,0 +1,21 @@
+prometheusIngressAuthSecret:
+  enabled: true
+
+prometheus:
+  server:
+    ingress:
+      enabled: true
+      hosts:
+        - prometheus.community.uk.2i2c.cloud
```
yuvipanda, line 9: Let's get rid of the `.community` prefix, and just use `.uk.2i2c.cloud` both here and for the staging hub?

yuvipanda (approved):
One suggestion re: naming of domains, otherwise LGTM.

Thank you so much for working on this, @GeorgianaElena

config/clusters/2i2c-uk/support.values.yaml
GeorgianaElena: Done

GeorgianaElena:
Thank you all! Merging this now as complete, but will leave https://github.com/2i2c-org/infrastructure/issues/1485 open a little longer, until the authentication issue sorts out

PR #1735, opened 2022-09-28
Update pangeo images to latest versions

rabernat:
We have had several reports of performance issues on the Pangeo hubs that might be fixed by using a new image

- https://discourse.pangeo.io/t/dask-cluster-stays-idle-for-a-long-time-before-computing/2742
- https://discourse.pangeo.io/t/dask-not-completing-large-operations-on-sose-data/2788 

Some questions

- Is it possible to make an announcement to the hub users that the image is being updated?
- If this update breaks user code, how should we respond? It would be great to allow users to select their own images (a feature I've been requesting for a long time).
- Is there a way to automate or otherwise obliviate the need for me to make PRs like this every time an update is needed? The release cadence of Pangeo docker images is approximately [two per month](https://hub.docker.com/r/pangeo/ml-notebook/tags). This reflects the pace of innovation in the Pangeo software ecosystem.

### DIFF
--- a/config/clusters/leap/common.values.yaml
+++ b/config/clusters/leap/common.values.yaml
@@ -54,7 +54,7 @@ basehub:
     singleuser:
       image:
         name: pangeo/pangeo-notebook
-        tag: "2022.06.02"
+        tag: "2022.09.21"
       extraEnv:
         GH_SCOPED_CREDS_CLIENT_ID: "Iv1.0c7df3d4b3191b2f"
         GH_SCOPED_CREDS_APP_URL: https://github.com/apps/leap-hub-push-access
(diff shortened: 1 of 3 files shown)


yuvipanda:
> Is there a way to automate or otherwise obliviate the need for me to make PRs like this every time an update is needed? The release cadence of Pangeo docker images is approximately [two per month](https://hub.docker.com/r/pangeo/ml-notebook/tags). This reflects the pace of innovation in the Pangeo software ecosystem.

We have a way to automatically open PRs for new tags that @sgibson91 set up. Perhaps we can add a date filter to it, so it happens every time there's a date tagged release of the pangeo images?

sgibson91:
> Perhaps we can add a date filter to it, so it happens every time there's a date tagged release of the pangeo images?

I had already implemented that - I can dig the config out of the commits. This was all setup but I disabled it because I received feedback that changing the image under people wasn't favourable

sgibson91:
The action I wrote only automates this PR-opening process - so yes, you can rollback.

sgibson91:
Here's an example of the PRs it opens https://github.com/2i2c-org/infrastructure/pull/1629

rabernat:
To be clear, we would need _individual users_ to be able to roll back to earlier images, not the entire hub at once. The use case here is that a user has a project that depends crucially on a specific library version that is present in, say `2022.06` but not `2022.09`. Even if the default image goes forward to `2022.09`, that user needs to be able to choose `2022.06`  when they are booting their server.

There is an inherent tension between the desire to get the latest and greatest packages into the hub and the need to preserve stability for users. I'm just trying to brainstorm different ways we might achieve resolve this tension.

yuvipanda:
A dropdown that people can use to select from all available image tags, automatically populated, seems like it would be killer. We should actually be able to do that now https://github.com/jupyterhub/kubespawner/pull/607 has landed.

We'll have to make an API call to the dockerhub to get the list of tags, and provide them as options.

rabernat:
> We should actually be able to do that now

If this is the case, I would love to see it implemented asap. This has been one of our main feature requests from Pangeo world for 2.5 years (https://github.com/jupyterhub/kubespawner/issues/307).

yuvipanda:
@rabernat looks like I missed a tiny feature there when I implemented it, https://github.com/jupyterhub/kubespawner/pull/640 is needed to enable this. Tiny patch, hopefully it gets merged soon.

Issue #1742, opened 2022-10-04
Update image for lis hub


damianavila:
### Context

This request is coming through the following support ticket: https://2i2c.freshdesk.com/a/tickets/209.
Matthew already used the Configurator to update the image but he wants a more permanent change by now.

### Proposal

Change the current image source from:

`lisds/lisds-base:781af1c`

to:

`lisacuk/lishub-base:b183052`

And this is the relevant config: https://github.com/2i2c-org/infrastructure/blob/master/config/clusters/2i2c-uk/lis.values.yaml#L32-L35

sgibson91:
I opened a PR: https://github.com/2i2c-org/infrastructure/pull/1744

We could also ask if they would like these kinds of PRs to be auto-opened using the action: https://github.com/2i2c-org/infrastructure/blob/master/.github/workflows/bump-image-tags.yaml

damianavila:
> I opened a PR:

Thanks!!

> We could also ask if they would like these kinds of PRs to be auto-opened using the action: https://github.com/2i2c-org/infrastructure/blob/master/.github/workflows/bump-image-tags.yaml

Asked about it in the support ticket.

damianavila:
Let's ping them again.

GeorgianaElena:
Received a reply -> yes, enable it

sgibson91:
@GeorgianaElena I don't actually see their affirmative reply in https://2i2c.freshdesk.com/a/tickets/216 did this message come to you elsewhere?

GeorgianaElena:
It's in here https://2i2c.freshdesk.com/a/tickets/268
The replies broke the ticketing somehow 🤷🏼‍♀️

GeorgianaElena:
Maybe we can merge them? At least we'll have them in the same place

sgibson91:
Thanks! I merged them

consideRatio:
My understanding from reading https://2i2c.freshdesk.com/a/tickets/209 is that the configurator was presumed to only adjust the image for the individual user, and not the default image for all users. Due to that, Matthew wanted a more permant solution of helping other users use the new image as well - but in practice, it was already accomplished by the configurator.

I'll communicate that the configurator changes the default image for all users already.

PR #1744, opened 2022-10-04
Bump the image tag for the LIS hub image

sgibson91:
addresses #1742


yuvipanda (approved):
Can they use the configurator in the future?

sgibson91:
> Can they use the configurator in the future?

They already did but wanted this to be a permanent change. I noted in the issue that we should ask them if they would like the action enabled for this hub.

PR #1746, opened 2022-10-05
Keep 1y of prometheus data on shared cluster

yuvipanda:
James is using these for billing.

### DIFF
--- a/config/clusters/2i2c/support.values.yaml
+++ b/config/clusters/2i2c/support.values.yaml
@@ -3,7 +3,7 @@ prometheusIngressAuthSecret:
 
 prometheus:
   server:
-    retention: 180d
+    etention: 365d
     ingress:
       enabled: true
       hosts:
--- a/config/clusters/2i2c/support.values.yaml
+++ b/config/clusters/2i2c/support.values.yaml
@@ -3,6 +3,7 @@ prometheusIngressAuthSecret:
 
 prometheus:
   server:
+    retention: 180d
     ingress:
       enabled: true
       hosts:


config/clusters/2i2c/support.values.yaml
```diff
@@ -3,6 +3,7 @@ prometheusIngressAuthSecret:
 
 prometheus:
   server:
+    retention: 180d
```
sgibson91, line 6: I'm looking at this and thinking `180d` means "180 days", but the title of this PR indicates we want to save data for 1 year. So either I've misunderstood, or this is a typo?

choldgraf:
Could this come with an MVP amount of documentation to show team members how to actually access this data? Like the least amount of lines of code needed, nothing complex

Or if @jmunroe already knows how to do this then can we collaborate in a follow up PR to do this?

yuvipanda:
@choldgraf This doesn't actually change anything at all about what data is being collected or how Grafana does anything - currently we only keep 90 days of data, this extends it to 1 year. I demo'd this to @jmunroe yesterday during a call, and we realized we needed to fix this ASAP as currently this means there is no way to determine usage beyond a 'last 90 days' window in the shared cluster. Work on actual invoicing remains separate, and I think is being tracked in another issue (that I can't find now)

I'm going to just merge this now so we stop losing data

config/clusters/2i2c/support.values.yaml
yuvipanda: @sgibson91 yep, missed a push! fixed

PR #1748, opened 2022-10-06
Fix typo

yuvipanda:
Follow-up to https://github.com/2i2c-org/infrastructure/pull/1746


yuvipanda:
Merging with same rationale as https://github.com/2i2c-org/infrastructure/pull/1746#issuecomment-1270851929

PR #1749, opened 2022-10-07
Allow running deployer from any path

yuvipanda:
Previously, the deployer could only be run if the current working directory of the running it is the root of the repository. This commit removes that dependency, and allows the deployer to be called from any working directory.

### DIFF
--- a/deployer/deploy_actions.py
+++ b/deployer/deploy_actions.py
@@ -245,9 +245,8 @@ def generate_helm_upgrade_jobs(changed_filepaths):
     ) = discover_modified_common_files(changed_filepaths)
 
     # Convert changed filepaths into absolute Posix Paths
-    changed_filepaths = [
-        Path(os.getcwd()).joinpath(filepath) for filepath in changed_filepaths
-    ]
+    root_dir = Path(__file__).parent.parent
+    changed_filepaths = [root_dir.joinpath(filepath) for filepath in changed_filepaths]
 
     # Get a list of filepaths to all cluster.yaml files in the repo
     cluster_files = get_all_cluster_yaml_files()
--- a/deployer/helm_upgrade_decision.py
+++ b/deployer/helm_upgrade_decision.py
@@ -4,7 +4,6 @@ support helm chart upgrading depending on an input list of filenames that have b
 added or modified in a GitHub Pull Request.
 """
 import fnmatch
-import os
 from pathlib import Path
 
 from rich.console import Console
@@ -84,26 +83,16 @@ def get_all_cluster_yaml_files(is_test=False):
     Returns:
         set[path obj]: A set of absolute paths to all cluster.yaml files in the repo
     """
+    root_path = Path(__file__).parent.parent
     # Get absolute paths
     if is_test:
         # We are running a test via pytest. We only want to focus on the cluster
         # folders nested under the `tests/` folder.
-        cluster_files = [
-            filepath
-            for filepath in Path(os.getcwd()).glob("**/cluster.yaml")
-            if "tests" in str(filepath)
-        ]
+        return root_path.glob("tests/test-clusters/**/cluster.yaml")
     else:
         # We are NOT running a test via pytest. We want to explicitly ignore the
         # cluster folders nested under the `tests/` folder.
-        cluster_files = [
-            filepath
-            for filepath in Path(os.getcwd()).glob("**/cluster.yaml")
-            if "tests" not in str(filepath)
-        ]
-
-    # Return unique absolute paths
-    return set(cluster_files)
+        return root_path.glob("config/clusters/**/cluster.yaml")
 
 
 def generate_hub_matrix_jobs(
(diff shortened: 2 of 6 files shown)


deployer/helm_upgrade_decision.py
```diff
@@ -84,26 +83,16 @@ def get_all_cluster_yaml_files(is_test=False):
     Returns:
         set[path obj]: A set of absolute paths to all cluster.yaml files in the repo
```
GeorgianaElena, line 84: this doesn't return a `set` anymore, right?

deployer/helm_upgrade_decision.py
```diff
@@ -84,26 +83,16 @@ def get_all_cluster_yaml_files(is_test=False):
...
     if is_test:
         # We are running a test via pytest. We only want to focus on the cluster
         # folders nested under the `tests/` folder.
-        cluster_files = [
-            filepath
-            for filepath in Path(os.getcwd()).glob("**/cluster.yaml")
-            if "tests" in str(filepath)
-        ]
+        return root_path.glob("tests/test-clusters/**/cluster.yaml")
     else:
```
GeorgianaElena, line 92: this can be removed I believe

GeorgianaElena (commented):
I left a couple of inline comments, but this looks good to me. Thanks @yuvipanda

deployer/helm_upgrade_decision.py
sgibson91: We should do a run of `generate-helm-upgrade-jobs` locally to check as I used set operations throughout that logic and they _could_ break if this object is no longer a set. I'd hope that CI would've picked it up, but I wasn't able to write a test for everything.

sgibson91: I did a local test run and `generate-helm-upgrade-jobs` was fine so this property can be edited as needed

tests/test_helm_upgrade_decision.py
```diff
@@ -16,24 +16,23 @@
...
 def test_get_all_cluster_yaml_files():
     expected_cluster_files = {
-        Path(os.getcwd()).joinpath("tests/test-clusters/cluster1/cluster.yaml"),
-        Path(os.getcwd()).joinpath("tests/test-clusters/cluster2/cluster.yaml"),
+        root_path.joinpath("tests/test-clusters/cluster1/cluster.yaml"),
+        root_path.joinpath("tests/test-clusters/cluster2/cluster.yaml"),
     }
 
-    result_cluster_files = get_all_cluster_yaml_files(is_test=True)
+    result_cluster_files = set(get_all_cluster_yaml_files(is_test=True))
```
GeorgianaElena, line 28: @sgibson91, the result it's turned into a set here, so probably this is why the generate job is not complaining. But I still believe we should update the comment to match what's happening currently or return the set directly

sgibson91: That code link is from the test file, not the actual codebase, but probably something similar is happening in the codebase too (I'm finding this a difficult PR to review actually). But I agree, the comment should be updated now that I've checked that the logic isn't broken.

deployer/deploy_actions.py
```diff
@@ -245,9 +245,8 @@ def generate_helm_upgrade_jobs(changed_filepaths):
...
 
     # Convert changed filepaths into absolute Posix Paths
-    changed_filepaths = [
-        Path(os.getcwd()).joinpath(filepath) for filepath in changed_filepaths
-    ]
+    root_dir = Path(__file__).parent.parent
+    changed_filepaths = [root_dir.joinpath(filepath) for filepath in changed_filepaths]
 
     # Get a list of filepaths to all cluster.yaml files in the repo
     cluster_files = get_all_cluster_yaml_files()
```
sgibson91, line 252: @GeorgianaElena this is in the codebase, but `set()` is not being called on the result 🤔

tests/test_helm_upgrade_decision.py
GeorgianaElena: huh! You're right, I didn't notice it's in a test

yuvipanda:
@GeorgianaElena @sgibson91 so the output of `get_all_cluster_yaml_files` was always used in our code only to iterate, never for set operations. However, to make the overall change in this PR smaller, I've made it return a `set` again so the only code change here is to get rid of the `os.getcwd()` calls and not make any other changes. I hope that makes it easier to review too!

deployer/helm_upgrade_decision.py
yuvipanda: @GeorgianaElena good catch!

sgibson91 (approved):
Thanks @yuvipanda - I appreciate the simplification, was much easier to review as code changes were more side-by-side. We can make a follow-up PR to address the `set` thing if needed :)

GeorgianaElena (approved):
+1 on what @sgibson91 said ☝🏼
Thanks @yuvipanda!

yuvipanda:
Thanks a lot for your review, @sgibson91 and @GeorgianaElena!

sgibson91:
I think I hit a bug with this implementation when I accidentally tried to run the deployer from the `terraform/aws` folder

Screenshot 2022-10-14 at 16 07 52

PR #1750, opened 2022-10-07
Add simple HTTPS uptime checks for all hubs

yuvipanda:
- Uses GCP's [Uptime Checks](https://cloud.google.com/monitoring/uptime-checks) feature, as we get it for free.
- Performs a simple HTTPS check to the /hub/health endpoint of all the hubs we maintain.
- An alert is sent to our pagerduty channel 'Manged JupyterHubs' if any of them are down for 5 minutes
- Uses Terraform to set these up
- Documentation added

### DIFF
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -329,13 +329,25 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v3
 
-      - name: Setup deploy for ${{ matrix.jobs.cluster_name }} cluster
-        uses: ./.github/actions/setup-deploy
+      - uses: hashicorp/setup-terraform@v2
+
+      - name: Install sops
+        uses: mdgreenwald/mozilla-sops-action@v1
         with:
-          # Use GCP, as we use GCP uptime checks
-          provider: gcp
-          GCP_KMS_DECRYPTOR_KEY: ${{ secrets.GCP_KMS_DECRYPTOR_KEY }}
+          version: v3.7.2
+
+      - name: Setup sops credentials to decrypt repo secrets
+        uses: google-github-actions/auth@v0
+        with:
+          credentials_json: "${{ inputs.GCP_KMS_DECRYPTOR_KEY }}"
 
       - name: ensure uptime checks are set up
         run: |
-          python3 deployer ensure-uptime-checks
+          cd terraform/uptime-checks
+          # Decrypt the GCP ServiceAccount key with permissions to run terrafomr
+          sops -d enc-service-account-key.secret.json > service-account-key.json
+          export GOOGLE_APPLICATION_CREDENTIALS=service-account-key.json
+          # Setup Terraform
+          terraform init
+          # Run terraform automatically
+          terraform apply -auto-approve
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -356,7 +356,7 @@ jobs:
       - name: ensure uptime checks are set up
         run: |
           cd terraform/uptime-checks
-          # Decrypt the GCP ServiceAccount key with permissions to run terrafomr
+          # Decrypt the GCP ServiceAccount key with permissions to run terraform
           sops -d enc-service-account-key.secret.json > service-account-key.json
           export GOOGLE_APPLICATION_CREDENTIALS=service-account-key.json
           # Setup Terraform
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -318,3 +318,24 @@ jobs:
           max_attempts: 3
           command: |
             python deployer run-hub-health-check ${{ matrix.jobs.cluster_name }} ${{ matrix.jobs.hub_name }}
+
+  # This job should run after we deploy staging or prod hubs
+  # It makes sure there is an uptime check and an alerting policy for all our hubs
+  ensure-uptime-checks:
+    runs-on: ubuntu-latest
+    needs: [upgrade-support-and-staging, upgrade-prod-hubs]
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+
+      - name: Setup deploy for ${{ matrix.jobs.cluster_name }} cluster
+        uses: ./.github/actions/setup-deploy
+        with:
+          # Use GCP, as we use GCP uptime checks
+          provider: gcp
+          GCP_KMS_DECRYPTOR_KEY: ${{ secrets.GCP_KMS_DECRYPTOR_KEY }}
+
+      - name: ensure uptime checks are set up
+        run: |
+          python3 deployer ensure-uptime-checks
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -325,17 +325,29 @@ jobs:
     runs-on: ubuntu-latest
     needs: [upgrade-support-and-staging, upgrade-prod-hubs]
 
+    # always() is added as it seems to be a magic function to ensure a job can
+    # run at all without being skipped if a previous job has had any failures.
+    # https://docs.github.com/en/actions/learn-github-actions/expressions#always
+    # We want this job to run if *any* hub was deployed, production or staging.
+    if: |
+      always() && needs.filter-generate-jobs.result == 'success' &&
+      (github.event_name == 'push' && contains(github.ref, 'master')) &&
+      (needs.filter-generate-jobs.outputs.filtered-prod-hub-matrix-jobs != '[]' ||
+      needs.generate-jobs.outputs.support-and-staging-matrix-jobs != '[]')
+
     steps:
       - name: Checkout repo
         uses: actions/checkout@v3
 
+      # Uptime checks are set up and managed via terraform
       - uses: hashicorp/setup-terraform@v2
 
+      # We use sops to store encrypted GCP ServiceAccount Key that terraform uses
+      # to run, as well as PagerDuty config terraform uses
       - name: Install sops
         uses: mdgreenwald/mozilla-sops-action@v1
-        with:
-          version: v3.7.2
 
+      # Authenticate with the correct KMS key that sops will use.
       - name: Setup sops credentials to decrypt repo secrets
         uses: google-github-actions/auth@v0
         with:
--- a/deployer/README.md
+++ b/deployer/README.md
@@ -30,12 +30,13 @@ $ python deployer [sub-command]
 **Command line usage:**
 
 ```bash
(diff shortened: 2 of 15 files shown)


deployer/deploy_actions.py
```diff
@@ -466,3 +467,57 @@ def exec_homes_shell(cluster_name, hub_name):
...
+
+def ensure_uptime_checks(cluster_name: str, notification_channel_id: str):
+    """
+    Ensure uptime checks for *all* hubs are set up
+
+    cluster_name is the *cluster* that is in the *project* where we want to
+    create the uptime checks. We are using this as a convenience function to
+    authenticate to the appropriate GCP project.
+
+    notification_channel is the full id (of the form projects/<project-id>/notificationChannels/<id>)
```
GeorgianaElena, line 480:
```suggestion
    notification_channel_id is the full id (of the form projects/<project-id>/notificationChannels/<id>)
```

.github/workflows/deploy-hubs.yaml
```diff
@@ -318,3 +318,24 @@ jobs:
...
+      - name: Setup deploy for ${{ matrix.jobs.cluster_name }} cluster
+        uses: ./.github/actions/setup-deploy
+        with:
+          # Use GCP, as we use GCP uptime checks
+          provider: gcp
+          GCP_KMS_DECRYPTOR_KEY: ${{ secrets.GCP_KMS_DECRYPTOR_KEY }}
+
+      - name: ensure uptime checks are set up
+        run: |
+          python3 deployer ensure-uptime-checks
```
GeorgianaElena, line 341: Hmm, I think that without a cluster passed here, this ensures the checks are setup for the hubs in the 2i2c cluster, for every cluster deployed?

But it's not very clear who `--notification-channel-id` should be? Do we need to create  this channel manually for each gcp project?

.github/ISSUE_TEMPLATE/4_decommission-hub.md
```diff
@@ -38,6 +38,7 @@ Usually, it is because it was a hub that we created for a workshop/conference an
   - `helm --namespace HUB_NAME delete HUB_NAME`
   - `kubectl delete namespace HUB_NAME`
 - [ ] Delete the hub's authentication application on auth0, GitHub or CILogon
+- [ ] Remove the [uptime check](https://console.cloud.google.com/monitoring/uptime?referrer=search&project=two-eye-two-see) and associated alert policy for this hub
```
GeorgianaElena, line 41: I think we should also have instructions about the manual steps in the corresponding  "add hub" (or rather "add cluster") docs section?

*I think* but I'm not sure that the only manual step is to take per clsuter is to create the notification channel ?

GeorgianaElena (commented):
This is super cool to have

.github/workflows/deploy-hubs.yaml
sgibson91: Perhaps a simpler way to include this would be as a new step in the `upgrade-support-and-staging` and `upgrade-prod-hubs` jobs? They already consume the relevant cluster info and ensure we only run jobs for the clusters/hubs we need to

sgibson91: We may also need to maintain a mapping of `cluster name` -> `notification channel ID` somewhere for the script to read, unless we have a catch-all channel set as the default?

sgibson91: If we keep is as a separate job, we will need an `if: always()` statement like the above jobs because of the reasoning given in the comment

https://github.com/2i2c-org/infrastructure/blob/c52df2d99fa77c6eb6ffd82e6deda9e09eb89742/.github/workflows/deploy-hubs.yaml#L239-L245

yuvipanda:
@GeorgianaElena @sgibson91 thank you very much for the review and comments. It made me realize that the code and documentation as written was not easy to understand, so I've completely rewritten everything and taken a totally different approach. I'm hoping the documentation in https://github.com/2i2c-org/infrastructure/pull/1750/files#diff-f4ac06565f573ca00491b1116529d5b76e373e63f87defa64662235f8e79e739 should help answer all the questions raised in review.

.github/ISSUE_TEMPLATE/4_decommission-hub.md
yuvipanda: I've tried to answer this in https://github.com/2i2c-org/infrastructure/pull/1750/files#diff-f4ac06565f573ca00491b1116529d5b76e373e63f87defa64662235f8e79e739

.github/workflows/deploy-hubs.yaml
yuvipanda: I hope https://github.com/2i2c-org/infrastructure/pull/1750/files#diff-f4ac06565f573ca00491b1116529d5b76e373e63f87defa64662235f8e79e739 clarifies this a little bit more. Re: GitHub, it basically needs to run every time at the end if *any* hubs have been deployed I think. I *think* I got it with my new commits, but not 100% sure.

deployer/README.md
```diff
@@ -30,12 +30,13 @@ $ python deployer [sub-command]
 **Command line usage:**
 
 ```bash
-usage: deployer [-h] {deploy,validate,deploy-support,deploy-grafana-dashboards,use-cluster-credentials,generate-helm-upgrade-jobs,run-hub-health-check,exec-homes-shell} ...
+usage: deployer [-h]
```
yuvipanda: I just updated the output here, nothing material has changed

docs/topic/uptime-checks.md
```diff
@@ -0,0 +1,60 @@
+# Simple HTTPS Uptime Checks
+
+Ideally, when a hub is down, a machine alerts us - we do not have to wait for a user
+to report it to our helpdesk. While we aren't quite there, we currently have very simple
+uptime monitoring for all our hubs. [GCP Uptime Checks](https://cloud.google.com/monitoring/uptime-checks)
+are used (because they are free) to hit the `/hub/health` endpoint
+of the public URL of all our hubs. If these checks fail for 5 minutes, an alert is sent
+to our [PagerDuty](https://team-compass.2i2c.org/en/latest/projects/managed-hubs/incidents.html)
+notification channel.
```
GeorgianaElena, line 9: Maybe add a link to the docs?
```suggestion
[notification channel](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.notificationChannels).
```

GeorgianaElena (commented):
Thanks for re-working this again @yuvipanda to make it easier to understand.
Setting aside the fact that terraform config is more intimidating (for me) than python code, the new docs really clarified the questions I had.

Also, judging from the pageduty channel on slack, it looks like this new setup it working, right? 🚀

.github/workflows/deploy-hubs.yaml
```diff
@@ -318,3 +318,48 @@ jobs:
...
+      # Authenticate with the correct KMS key that sops will use.
+      - name: Setup sops credentials to decrypt repo secrets
+        uses: google-github-actions/auth@v0
+        with:
+          credentials_json: "${{ inputs.GCP_KMS_DECRYPTOR_KEY }}"
+
+      - name: ensure uptime checks are set up
+        run: |
+          cd terraform/uptime-checks
+          # Decrypt the GCP ServiceAccount key with permissions to run terrafomr
```
sgibson91, line 359:
```suggestion
          # Decrypt the GCP ServiceAccount key with permissions to run terraform
```

typo

terraform/uptime-checks/checks.tf
```diff
@@ -0,0 +1,81 @@
...
+}
+
+resource "google_monitoring_uptime_check_config" "hub_simple_uptime_check" {
+  for_each = { for h in local.hubs : h.domain => h }
+
+  display_name = "${each.value.domain} on ${each.value.cluster}"
+  timeout      = "30s"
+
+  # Check every minute
+  period = "300s"
```
sgibson91, line 20: This comment says "check every minute", but the period is 300 seconds (5 minutes) - which is it?

ETA: Judging by the comment on L61, the comment needs to be updated to "check every 5 mins"

docs/topic/uptime-checks.md
```diff
@@ -0,0 +1,60 @@
...
+instructions](https://cloud.google.com/monitoring/support/notification-options#pagerduty).
+
+```{note}
+This Notification Channel was created manually following instructions in the
+link provided, not automatically. We *must* automate this if we create another
+notification channel, but since a single notification channel is ok for how we
+use PagerDuty now, it is fine.
+```
+
+When any of the checks fail, they automatically open an Incident in the
```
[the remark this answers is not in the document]
sgibson91: I think this is the comment describing the behaviour, so we should make sure the docs accurately reflect that comment: https://github.com/2i2c-org/infrastructure/pull/1750/files#diff-7851a08e9136742f1f5264dc95b6a9a4c0550ceba1525e83a8d8f17167f4cfedR61

sgibson91 (commented):
Thank you @yuvipanda for working on this - this is much clearer to me now! I think there's still a few documentation points that need clarifying that Georgiana has pointed out too, but we are very close to a merge!

docs/topic/uptime-checks.md
```diff
@@ -0,0 +1,60 @@
...
+
+Uptime checks are *centralized* - they don't exist in the same project or cloud provider
+as the hubs they are checking, but in one centralized GCP project (`two-eye-two-see`). This
+has a few advantages:
+
+1. We do not have to implement the same functionality three times (one per cloud provider),
+   as we would have to if this were to exist in the same project as the hub.
+2. We use a centralized PagerDuty for alerting, so this Notification Channel can exist in
+   only one place. Setting up this channel requires secrets, and we'll have to either add
+   secret handling to our terraform code or re-implement setting this up in python 3 times.
```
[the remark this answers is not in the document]
yuvipanda: @GeorgianaElena ah yep! I’ll reword, this is left over from before the rewrite

docs/topic/uptime-checks.md
yuvipanda: Added a link!

yuvipanda:
@sgibson91 @GeorgianaElena thank you for your review! The document had a lot of leftovers from before the rewrite that I had completely missed, sorry about that. I have read through and clarified it, and also tweaked the alerting conditions to prevent alert fatigue. What do you think now?

docs/topic/uptime-checks.md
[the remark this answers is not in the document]
yuvipanda: I've tried to clarify this with a 'when are notifications triggered?' section in the doc

yuvipanda:
@sgibson91 I also really appreciate you specifically catching so many of my typos non-judgementally! I try to avoid them but alas...

sgibson91:
@yuvipanda Typing without typos is _hard_!! 😆

sgibson91 (approved):
🎉 🎉 🎉 🎉 🎉

Issue #629, opened 2021-08-23
Documenting use of nfs-server-provisioner


sgibson91:
### Description

We recently deployed `nfs-server-provisioner` in our support chart so that we can deploy in-cluster NFS servers for our hubs. This is **only used for Google Cloud** because its filesystem service is much more expensive than other cloud providers.

We should document this practice, as well as some other important information about the NFS servers. Here are a few questions that @sgibson91 brought up:

How is user data stored longer term when using the `nfs-server-provisioner` helm chart?

1. What are the common ways data loss can occur? Does deleting the PV delete the backing disk?
2. Can we resize them later if required? If so, how?
3. Will we have one disk per hub or one per cluster?
4. How can we do backups?

### Benefit

It is important that we document our team practices in general, but particular when they deviate from the "standard / expected" behavior. In this case, we are using managed services for filesystems in both AWS and Azure, but manually deploying NFS in Google Cloud. At the least, this should be documented so others aren't surprised.

More generally, as @sgibson91 notes:

> Losing user home dirs is always the worst case scenario, so we need to make sure we understand what is happening here.

Now that we are manually deploying NFS, it means we are on the hook for its behavior in a stronger way. It is really important that we understand what's going on because losing user data is a worst-case scenario.

### Implementation

I think this could be a "Filesystems and local storage" section in the Hub Engineer's guide: https://pilot-hubs.2i2c.org/en/latest/

choldgraf:
**Note**: this one is probably a little bit less impactful right now, since we are temporarily using Google FileStore instead of our in-cluster NFS server. So this support component isn't currently in-production (I believe). We should re-visit this one when we make a switch back to in-cluster NFS on Google Cloud.

Issue #1686, opened 2022-09-06
Workflows API rate limited in CI/CD?


sgibson91:
### Context

I think we are hitting some kind of rate limit in CI/CD. Every now and then the following error message appears.

<img width="529" alt="Screenshot 2022-09-06 at 20 24 35" src="https://user-images.githubusercontent.com/44771837/188721724-714e9b6f-b411-463f-8724-1b89ea17c0a8.png">

Although, one would expect that if it were a rate limiting issue, simply waiting an amount of time and rerunning the workflow should resolve it; however, that is not the case. Often pushing a new commit resolves it.

We mostly use the inbuilt `GITHUB_TOKEN` with workflow-appropriately scoped permissions via the `permissions` block. Except where a workflow would open a PR (e.g., the bump image tags workflow), which instead swaps `GITHUB_TOKEN` for a token generated by a GitHub App. (This is because we went tests to run on the opened PR, but this doesn't happen if it was opened by `GITHUB_TOKEN`.)

### Checkpoints

- [x] The immediate problem is gone because time passed and our limit was reset
- [x] Wait until https://github.com/2i2c-org/meta/issues/417 is resolved
- [ ] Re-ping github if we don't hear by 2022/11/21
- [ ] Further investigation is needed as per https://github.com/2i2c-org/infrastructure/issues/1686#issuecomment-1463985843

sgibson91:
Found this in the docs:

> When using GITHUB_TOKEN, the rate limit is 1,000 requests per hour per repository. For requests to resources that belong to an enterprise account on GitHub.com, GitHub Enterprise Cloud's rate limit applies, and the limit is 15,000 requests per hour per repository.

https://docs.github.com/en/rest/overview/resources-in-the-rest-api#requests-from-github-actions

sgibson91:
We may need to update _all_ our workflows so that they swap out `GITHUB_TOKEN` for a PAT from our GitHub App: https://github.com/organizations/2i2c-org/settings/apps/2i2c-token-generator-bot

This code already implements this: https://github.com/2i2c-org/infrastructure/blob/72ae1c824b9c6021fc3cd3597c187c549a5f90c5/.github/workflows/bump-helm-versions.yaml#L39-L50

docs: https://docs.github.com/en/developers/apps/building-github-apps/rate-limits-for-github-apps

sgibson91:
[...]

Issue #764, opened 2021-10-19
Enabling per hub storage quotas


sgibson91:
### October 2023 Update

Current proposals:

- Implementation: https://github.com/2i2c-org/infrastructure/issues/764#issuecomment-1273882247
- Alerting/monitoring: https://github.com/2i2c-org/infrastructure/issues/764#issuecomment-1735897743

Work already done:

- Prometheus exporter for dir sizes: https://github.com/2i2c-org/infrastructure/issues/764#issuecomment-1579158332 

---

### Description

Right now in our NFS servers, user home dirs expand until the whole NFS storage is consumed and then we either expand the storage further or delete stuff. This isn't necessarily a problem, except that the storage may not be equally distributed between hubs or even users! A slightly nicer solution may be to implement some storage quotas on a per user or per hub basis and then we can at least communicate with Community Representatives that that is the amount of storage they are allocated and can work from there.

This is a topic that has already had a lot of discussion in Pangeo https://github.com/pangeo-data/pangeo-cloud-federation/issues/654

I already did some work to [bring the `nfs-server-provisioner` helm chart into our infrastructure](https://github.com/2i2c-org/infrastructure/pull/613). We time-boxed the effort to get that working and instead fell back onto use Google Cloud Filestore on the Pangeo deployment. I believe achieving such quotas will involve more work in making the `nfs-server-provisioner` chart functional.

### Value / benefit

- Can more clearly communicate how much storage a given Hub Community can expect
- No single user or hub can bogart the NFS storage with lots of data

sgibson91:
@damianavila I wonder if we can bring this issue into our backlog following https://2i2c.freshdesk.com/a/tickets/171 Since disk space filling up is most often felt by hubs on shared clusters and feels most unfair that they may not have even been the ones to cause it.

damianavila:
I added it to the backlog and raised the priority! Thanks for bringing this one to attention, @sgibson91.

yuvipanda:
My current worry is that the nfs external provisoner seems a bit abandoned (https://github.com/kubernetes-sigs/nfs-ganesha-server-and-external-provisioner/issues/106). @consideRatio got some patches into it, so he might have an idea on how active it is?

The other problem I have with it is that it generates directory names with randomized chars, so if we lose the PVC objects (by recreating the k8s cluster, for example) we can no longer map the users to their home directories! currently that is not the case - we can recreate the k8s cluster and not lose any user data. This complicates backup quite a bit.

[...]

PR #1751, opened 2022-10-07
Remove nfs-server-provisioner

yuvipanda:
We don't actually use this anywhere, and I think we have decided to eat the cost of using a managed NFS service (AWS EFS, Google Filestore, etc) wherever possible to minimize effort it takes for us.

Fixes https://github.com/2i2c-org/infrastructure/issues/629

### DIFF
--- a/helm-charts/support/Chart.yaml
+++ b/helm-charts/support/Chart.yaml
@@ -30,12 +30,6 @@ dependencies:
     repository: https://kubernetes.github.io/autoscaler
     condition: cluster-autoscaler.enabled
 
-  # In-cluster NFS server provisioner
-  - name: nfs-server-provisioner
-    version: 1.3.1
-    repository: https://kvaps.github.io/charts
-    condition: nfs-server-provisioner.enabled
-
   # cryptnono, counters crypto mining
   # Source code: https://github.com/yuvipanda/cryptnono/
   - name: cryptnono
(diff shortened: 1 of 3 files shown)


sgibson91:
@yuvipanda If we don't investigate setting up this helm chart properly, how do you propose we go about solving https://github.com/2i2c-org/infrastructure/issues/764? I think this is a really critical issue that we need to push forward, particularly those on shared clusters. When we receive support requests regarding the disk being out of space, usually the next question is "Where can I trim my data down?" and the truth is that we have no way of tracking that. Right now, one user could consume the whole NFS storage and could be a user of a different hub to the person who reported the issue. It's systemically not fair.

yuvipanda:
@sgibson91 I agree it's very important to have such limits in place. Thank you for bringing that up. I've left some comments in https://github.com/2i2c-org/infrastructure/issues/764#issuecomment-1273882247 with alternative approaches. It feels the external provisioner is unmaintained upstream, which is what prompted me to make this change.

I'd also like for us to manage as little NFS as possible, as it has been the one piece of technology that has constantly been a major cause of stress in every piece of infra I've ever run. Right now at Berkeley we're spending the last 6 weeks trying to track down a linux kernel NFS bug that seems to cause hour long outages when nobody is looking...

sgibson91 (approved):
Thank you for the proposal on #764 @yuvipanda - I like it a lot and now I'm happy that we have a path forward and can remove this dependency!

sgibson91:
BTW, I have an issue open related to the errors we see in CI/CD in this PR: https://github.com/2i2c-org/infrastructure/issues/1686 I have no idea what's happening!

yuvipanda:
Thanks a lot for consistently thinking about our users, @sgibson91!

PR #1763, opened 2022-10-12
Update notebook for tensorflow in m2lines deplyoment

dhruvbalwada:
This change should fix the jax problems that are mentioned here: https://github.com/pangeo-data/pangeo-docker-images/issues/387#issuecomment-1276427334


yuvipanda:
Thanks @dhruvbalwada! I’ll merge once tests complete.

rabernat:
See also #1765 - this might not work unless we change the GPU type.

PR #1766, opened 2022-10-12
Provide better GPU support in m2lines & LEAP hubs

yuvipanda:
- Add T4 GPUs in addition to K80 GPUs currently available
- Use newer profile dropdowns to make choice of GPUs and images easier
- Switch to a newer GPU driver version to allow using newer CUDA, based on
  conversation in https://github.com/pangeo-data/pangeo-docker-images/issues/387

Ref https://github.com/2i2c-org/infrastructure/issues/1765
image

PR #1773, opened 2022-10-14
Add missing GKE nvidia helm template

yuvipanda:
Follow-up to https://github.com/2i2c-org/infrastructure/pull/1766, git add was missed


yuvipanda:
Going to self-merge this because it was a missed file in #1766. It breaks GPUs on leap and m2lines

PR #1779, opened 2022-10-15
Keep prometheus data for 1 year rather than 90 days

yuvipanda:
I think this is broadly useful for everyone, and we now have uptime checks for prometheuses so if they go down we can know.


sgibson91 (approved):

choldgraf:
This looks great to me - I still don't believe that we have "how to access and use prometheus data" documented anywhere. I opened up an issue to track this, because I think it's an important one to resolve if we want to make this data useful to our team:

- https://github.com/2i2c-org/infrastructure/issues/1785

Issue #1757, opened 2022-10-11
Restructure our infrastructure docs


sgibson91:
### Context

We are getting a lot of docs in infrastructure.2i2c.org now, and I think the "how-to" section is becoming a bit of a "catch-all" area, which ultimately makes the docs you need more difficult to find.

### Proposal

I'm proposing we refactor and split up the "how-to" section into the following, separate sections:

* **Hub Deployment Guide**
  * This contains all the documentation required for deploying hubs and enabling/configuring specific features
  * It should follow, more or less, the order of operations required to deploy a hub so an engineer can just step through the docs in order, i.e., start with "Create new cloud accounts" -> "Create a k8s cluster" -> "deploy support infrastructure" -> "deploy a hub" -> and then we get into the realm of turning on/off specific features so this part can be less structured I think
* **Site Reliability Engineering Guide**
  * This contains day-to-day tasks we do _outside_ of hub deployments, e.g., a lot of the content currently under ["Cloud and Kubernetes infrastructure"](https://infrastructure.2i2c.org/en/latest/howto/k8s/index.html) and ["Support"](https://infrastructure.2i2c.org/en/latest/howto/support/index.html)
  * When we have incidents and work out how to resolve them, the solutions should be documented under this guide for the next occurrence (where the solution is not a "this will never happen again" kind of fix, which obviously we should be aiming for)

What does @2i2c-org/tech-team think? I am happy to implement this if we have consensus.

### Updates and actions

New docs:

- [x] https://github.com/2i2c-org/infrastructure/issues/1757#issuecomment-1285659593
- [x] https://github.com/2i2c-org/infrastructure/issues/1757#issuecomment-1285890909 (#1806)

Common problems and their solutions (troubleshooting/support):

- [ ] ~~https://github.com/2i2c-org/infrastructure/issues/1757#issuecomment-1286117676 (https://github.com/2i2c-org/infrastructure/pull/1780#issuecomment-1282096496)~~
- [x] https://github.com/2i2c-org/infrastructure/issues/1757#issuecomment-1287027429
- [x] https://github.com/2i2c-org/infrastructure/issues/1757#issuecomment-1292666904 (https://github.com/2i2c-org/docs/issues/100#issuecomment-925962073)

yuvipanda:
I LIKE!

damianavila:
Yep, I like it as well!

sgibson91:
[...]

PR #1780, opened 2022-10-15
Reduce memory limit for pangeo-hubs prometheus

yuvipanda:
It was down because it was crashing trying to reload from WAL, not because it was running out of RAM. `kubectl top pod -n support` shows it using under 2G, so a 4G limit seems appropriate

### DIFF
--- a/config/clusters/pangeo-hubs/support.values.yaml
+++ b/config/clusters/pangeo-hubs/support.values.yaml
@@ -28,5 +28,6 @@ prometheus:
             - prometheus.gcp.pangeo.2i2c.cloud
     resources:
       limits:
+        # Increase limits here as this isn't too small a cluster
         cpu: 2
-        memory: 16Gi
+        memory: 4Gi


GeorgianaElena (approved):
Makes sense. Thanks @yuvipanda

sgibson91:
Do we have what "WAL" is documented anywhere?

yuvipanda:
@sgibson91 it's in https://prometheus.io/docs/prometheus/latest/storage/ and maybe https://en.wikipedia.org/wiki/Write-ahead_logging. I'm not sure where we can put it in our repo though.

yuvipanda:
I've only heard about the concept too until I started digging into the logs here!

sgibson91:
> @sgibson91 it's in [prometheus.io/docs/prometheus/latest/storage](https://prometheus.io/docs/prometheus/latest/storage/) and maybe [en.wikipedia.org/wiki/Write-ahead_logging](https://en.wikipedia.org/wiki/Write-ahead_logging). I'm not sure where we can put it in our repo though.

Maybe when I get around to the docs restructuring it could go in the SRE guide?

* https://github.com/2i2c-org/infrastructure/issues/1757

sgibson91:
I'm thinking something like: "Prometheus is down. How do I find out why and how do I fix it?" We should have steps to answer "is this WAL or memory or something else?" and then resolution steps for the known causes. E.g., if it's memory, bump it. We do have: https://infrastructure.2i2c.org/en/latest/topic/troubleshooting.html#prometheus-server-is-out-of-memory-oom

PR #1781, opened 2022-10-15
Bump up prometheus & grafana versions

yuvipanda:
Good to keep these up to date, and the problem with our pangeo-hubs prometheus was a [WAL replay](https://www.postgresql.org/docs/current/wal-intro.html) bug causing a crash - perhaps a newer version will not have that.


GeorgianaElena (approved):

sgibson91:
@yuvipanda this is also something we can configure the helm chart version bumper to take care of? https://github.com/2i2c-org/infrastructure/blob/master/.github/workflows/bump-helm-versions.yaml

yuvipanda:
@sgibson91 ooooh, yes actually let's!

PR #1782, opened 2022-10-16
Add budget alerts based on forecasts

yuvipanda:
Just for GCP, this will email support@2i2c.org whenever forecasted spend for the month goes over a $$$ we set. Currently, I've set the numbers arbitrarily based on median spend for the project in question based on looking at history of spends.

pangeo-hubs and cloudbank don't give us enough permissions to set up these alerts, so we don't.

### DIFF
--- a/docs/topic/budget-alerts.md
+++ b/docs/topic/budget-alerts.md
@@ -1,0 +1,35 @@
+(topic:budget-alerts)=
+# Cloud Billing Budget Alerts
+
+"I forgot to turn off my cloud resources, your honor" as a reason for declaring
+bankruptcy is second only to "The US healthcare system sucks, your honor" in the
+US court system. "How much is my cloud going to cost?" is a big anxiety for a lot
+of our users, and hence us. We set up billing alerts to help deal with this anxiety.
+
+## When are the alerts triggered?
+
+Budget alerts are sent under two conditions:
+
+1. When *forecasted monthly spend* at end of the month goes over our spending limit.
+   This is an *early warning* system, that helps us evaluate where spend is going
+   and make sure this is expected.
+2. When *current actual spend* goves over 80% of our spending limit.
+
+## What to do when we receive an alert?
+
+The current goal is to just make sure we don't end up spending *wildly* more money
+than budgeted. So if the forecasted spend busts through on day 5 of the month,
+we might need to do something different than if it does on day 30. If it is expected
+to overshoot by 500% vs by 10$, our actions might be different. One valid action is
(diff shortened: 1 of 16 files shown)


yuvipanda:
I haven't actually deployed it for anything *except* the pilot-hubs cluster, so at least I know it works.

yuvipanda:
I haven't done anything for the Callysto project because I don't think I have access to it.

From a documentation and policy perspective, I think two important things to add are:

1. What to do when we *get* an alert? Alerts *must* be actionable or they become problematic themselves.
2. How do we determine the alert amount?

yuvipanda:
The question really becomes, *what are we trying to solve*? At this point, the answer is to make sure there aren't *runaway costs* here. If we define runaway costs as say, 50% over 'expected costs'. We wanna make sure someone doesn't forget to turn something off, or use up a 5000 node dask cluster to run down the grant in a week. 

So to begin with, I think the alert should be something like ~25% over expected monthly cost, which we can determine by looking at current amount of money being paid for it. And I think the action to be taken is:

1. Check for goddamn cryptominers
2. *Let the community representative know*

And then we can go from there.

docs/topic/budget-alerts.md
```diff
@@ -0,0 +1,35 @@
...
+
+## What to do when we receive an alert?
+
+The current goal is to just make sure we don't end up spending *wildly* more money
+than budgeted. So if the forecasted spend busts through on day 5 of the month,
+we might need to do something different than if it does on day 30. If it is expected
+to overshoot by 500% vs by 10$, our actions might be different. One valid action is
+we just adjust the forecast. As an organization, we need more experience with costs
+to figure out what the right thing to do is. So our current primary goal would
+be to work with our stakeholders and gather that texperience.
```
sgibson91, line 26:
```suggestion
be to work with our stakeholders and gather that experience.
```

sgibson91 (approved):

yuvipanda:
@GeorgianaElena can you send me again? I didn't get it :(

GeorgianaElena:
@yuvipanda, I believe I've just sent one to your 2i2c email. The initial one looks to have been sent to your gmail account.

yuvipanda:
@GeorgianaElena got it! We don't have access there either to setup an alert so I've it turned off for now!

GeorgianaElena:
> @GeorgianaElena got it! We don't have access there either to setup an alert so I've it turned off for now!

Should we ask them for this permissions? I believe they might consider it beneficial

PR #1769, opened 2022-10-13
Use dask-gateway 2022.10.0 and rely on mounted api-token from jupyterhub chart, also revert the workaround removing the hub NetworkPolicy

consideRatio:
Relies on https://github.com/dask/dask-gateway/pull/612 released in 2022.10.0 so we don't have to set this explicitly.

While I've tested this in the JMTE hub. I've not tested the adjustments I've made to the binderhub chart that includes another dask-gateway installation. I made the changes I think is required for it though and think it should work fine there as well.

- Closes #1754

---

I'd be thankful if someone else can go for a merge of this and ensure that things work in all the 2i2c hubs. I opted to go for a PR as that felt like a clear way of documenting what I thought should be done to resolve #1754.

PR #1788, opened 2022-10-18
Bump pangeo image tags for LEAP & m2lines

yuvipanda:
Brings in https://github.com/pangeo-data/pangeo-docker-images/pull/395, which brings in https://github.com/conda-forge/pangeo-dask-feedstock/pull/96, which brings in newer versions of dask and dask-gateway.

Follow-up to https://github.com/2i2c-org/infrastructure/pull/1769


yuvipanda:
Am going to self-merge this as it's just an image tag bump, and our guidelines specifically state that's ok :D

PR #1789, opened 2022-10-18
Move uptime-checks terraform apply into its own workflow

yuvipanda:
This should be *temporary* as I couldn't figure out how to deploy it only if a staging hub, prod hub or support chart has been deployed


sgibson91 (approved):
Tbh, I don't think I'd be too mad if this wasn't temporary. If it's not obvious how to trigger this conditionally on what's changed, then that implies some complexity that needs some brain cycles to figure out

PR #1790, opened 2022-10-18
Bump to latest z2jh master

yuvipanda:
We're on a beta version of z2jh. 2.0 was released, and this brings us up to latest master.


consideRatio (approved):
I think the only change here was to relax the need to specify singleuser.cmd unless the dockerimage had a suitable CMD already.

yuvipanda:
Thanks @consideRatio

damianavila:
Why we are bringing the latest main/master if there is a 2.0 released version?
Unless there is some bugfix we are fetching, I think we should keep things on released (or at least tagged) versions.

PR #1791, opened 2022-10-18
Remove image_repo from our clusters

yuvipanda:
We don't actually use it - default image is now set in basehub/values.yaml. The python code was not called from anywhere either

### DIFF
--- a/config/clusters/2i2c/cluster.yaml
+++ b/config/clusters/2i2c/cluster.yaml
@@ -1,5 +1,4 @@
 name: 2i2c
-image_repo: "quay.io/2i2c/2i2c-hubs-image"
 provider: gcp
 gcp:
   key: enc-deployer-credentials.secret.json
--- a/deployer/cluster.py
+++ b/deployer/cluster.py
@@ -34,40 +34,6 @@ class Cluster:
         else:
             raise ValueError(f'Provider {self.spec["provider"]} not supported')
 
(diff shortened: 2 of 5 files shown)


yuvipanda:
image

deployer/cluster.py
```diff
@@ -34,40 +34,6 @@ def auth(self):
         else:
             raise ValueError(f'Provider {self.spec["provider"]} not supported')
 
-    def ensure_docker_credhelpers(self):
```
consideRatio, line 37: Aaaaaaahhhh, yepp. One block of code I've looked at with confusion in the past.

consideRatio (approved):

yuvipanda:
Thanks @consideRatio

config/clusters/2i2c/cluster.yaml
```diff
@@ -1,5 +1,4 @@
 name: 2i2c
-image_repo: "quay.io/2i2c/2i2c-hubs-image"
```
sgibson91, line 2: @yuvipanda Did this image used to be called `pilot-hub`? I am thinking we may need to i) clean the old repo out of quay.io https://quay.io/organization/2i2c, and ii) update the docs here as that still references `pilot-hub`: https://infrastructure.2i2c.org/en/latest/topic/hub-image.html

PR #1805, opened 2022-10-20
Use gs:// as protocol for GCP buckets

yuvipanda:
gcs:// was also valid for fsspec but not for gcloud, so let's switch to gs://

Ref https://2i2c.freshdesk.com/a/tickets/231


sgibson91 (approved):

yuvipanda:
@dhruvbalwada @tomnicholas just an FYI for the both of you, as you're the folks using this the most. It shouldn't affect any code or functionality as fsspec supports both.

Issue #1667, opened 2022-08-31
Create a source of truth for how to access cloud UI for all clusters


choldgraf:
### Context

We had an incident that involved a community that had their own AWS console information. When we decided to look at the AWS console for 2i2c, we realized that it wasn't managed by 2i2c directly. As a result, it was unclear where the CarbonPlan cloud infrastructure was running.

- https://github.com/2i2c-org/infrastructure/issues/1666

### Proposal

We need a single location that is the source of truth for how to access the cloud dashboards for each of the hubs we work with. We also need a set of instructions that a team member can reliably follow to gain access to the dashboards.

This could be something like an added column to our [list of running hubs](https://infrastructure.2i2c.org/en/latest/reference/hubs.html).

PR #1809, opened 2022-10-21
Generate provider console UI link

GeorgianaElena:
Generate provider console UI link inside the list of running hubs.
Ref: https://github.com/2i2c-org/infrastructure/issues/1667

- It adds a new cluster config option `account` where we should specify if a specific cluster uses  "non-default" account to login

Currently it covers the following:
- [x] standard (aka 2i2c) GCP clusters
- [x] pangeo (columbia) GCP clusters
- [x] aws through 2i2c SSO
- [x] aws through other shared accounts? (have no idea where we have those creds stored, ex for openscapes, carbonplan)
- [x] azure

### DIFF
--- a/config/clusters/carbonplan/cluster.yaml
+++ b/config/clusters/carbonplan/cluster.yaml
@@ -1,6 +1,6 @@
 name: carbonplan
 provider: aws
-account: 631969445205
+account: "631969445205"
 aws:
   key: enc-deployer-credentials.secret.json
   clusterType: eks
--- a/config/clusters/carbonplan/cluster.yaml
+++ b/config/clusters/carbonplan/cluster.yaml
@@ -1,5 +1,6 @@
 name: carbonplan
 provider: aws
+account: don't know
 aws:
   key: enc-deployer-credentials.secret.json
   clusterType: eks
--- a/config/clusters/carbonplan/cluster.yaml
+++ b/config/clusters/carbonplan/cluster.yaml
@@ -1,6 +1,6 @@
 name: carbonplan
 provider: aws
-account: don't know
+account: 631969445205
 aws:
   key: enc-deployer-credentials.secret.json
   clusterType: eks
--- a/config/clusters/openscapes/cluster.yaml
+++ b/config/clusters/openscapes/cluster.yaml
@@ -1,5 +1,6 @@
 name: openscapes
 provider: aws
+account: 2i2cAdministrator
 aws:
   key: enc-deployer-credentials.secret.json
   clusterType: eks
--- a/config/clusters/openscapes/cluster.yaml
+++ b/config/clusters/openscapes/cluster.yaml
@@ -1,6 +1,6 @@
 name: openscapes
 provider: aws
-account: 783616723547
+account: "783616723547"
 aws:
   key: enc-deployer-credentials.secret.json
   clusterType: eks
--- a/config/clusters/openscapes/cluster.yaml
+++ b/config/clusters/openscapes/cluster.yaml
@@ -1,6 +1,6 @@
 name: openscapes
 provider: aws
-account: 2i2cAdministrator
+account: 783616723547
 aws:
   key: enc-deployer-credentials.secret.json
   clusterType: eks
--- a/config/clusters/leap/cluster.yaml
+++ b/config/clusters/leap/cluster.yaml
@@ -1,5 +1,6 @@
 name: leap
 provider: gcp
+account: columbia
 gcp:
   key: enc-deployer-credentials.secret.json
   project: leap-pangeo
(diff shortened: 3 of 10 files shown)


config/clusters/openscapes/cluster.yaml
```diff
@@ -1,5 +1,6 @@
 name: openscapes
 provider: aws
+account: 2i2cAdministrator
```
GeorgianaElena, line 3: @damianavila, you once shared these credentials with me. Are these still stored encrypted somewhere?

sgibson91:
> ? for azure ??

We don't have any 2i2c-owned Azure infrastructure right now. But when we get there we will probably setup a 2i2c tenant linked to the 2i2c.org domain and enable SSO via Azure Active Directory. And so we would login in with our 2i2c.org emails.

I think we only manage one Azure cluster right now, which is UofT and requires a UTOR ID.

config/clusters/carbonplan/cluster.yaml
```diff
@@ -1,5 +1,6 @@
 name: carbonplan
 provider: aws
+account: don't know
```
GeorgianaElena, line 3: @2i2c-org/tech-team, can someone help with info about the account we currently use to access the carbonplan cluster on aws, please?

sgibson91 (approved):
The code LGTM. Unfortunately, I don't know how to access CarbonPlan's AWS account 😕

damianavila:
@GeorgianaElena, you will find the accounts ID on this sheet: https://docs.google.com/spreadsheets/d/1NSaAKLG2_njXxs6JlGUAhSWeHONz9QSGLVwEK790IZo/edit#gid=537065664 (which is documented in our docs: https://infrastructure.2i2c.org/en/latest/topic/access-creds/cloud-auth.html#access-individual-aws-accounts). Let me know if you need me to go into those accounts and try to find any further specific information. Not really sure which IAM users were created over there from the top of my mind.

GeorgianaElena:
@damianavila thanks a lot for the doc. 

> Let me know if you need me to go into those accounts and try to find any further specific information. Not really sure which IAM users were created over there from the top of my mind.

Yes, I believe we should have the IAM usernames in that spreadsheet also, since both the id and the username are needed to login, right? 
I added a new column over there and add the username for the openscapes one, since I knew it already.
 
- [ ] Do you want to add entries for CarbonPlan? (the faralon one is gone I believe so we don't need it, and the sandbox is accessible with our 2i2c accounts, so that's covered also)

And the other question is about the passwords? Where do we store those? I believe we used to have them encrypted in this repo, but not anymore?

damianavila:
Yes, I believe we should have the IAM usernames in that spreadsheet also, since both the id and the username are needed to login, right?

Yep.

> Do you want to add entries for CarbonPlan? 

Sure, I will try to do it ASAP.

> And the other question is about the passwords? Where do we store those? I believe we used to have them encrypted in this repo, but not anymore?

Not sure about it, although it seems the natural place given how we manage other secrets... @yuvipanda, WDYT?

yuvipanda:
So I think we should avoid a common shared password for AWS access wherever possible. To that end, I've created individual user accounts for @damianavila @consideRatio @sgibson91 @GeorgianaElena and emailed it to their 2i2c.org accounts. I did the same for carbonplan, but left @consideRatio out to not spam him - lmk if you want access too! 

Once folks can get into these accounts with their own passwords (which I assume is managed via a password manager), we can delete both the common 2i2c account for openscapes as well as the spreadsheet.

yuvipanda:
In general, unfortunately I think for AWS accounts we all need individual usernames and passwords. The good news is that we can create / delete them ourselves, so it isn't as involved as getting an external account via some other IT department!

damianavila:
> So I think we should avoid a common shared password for AWS access wherever possible.

Makes sense to me.

> Once folks can get into these accounts with their own passwords (which I assume is managed via a password manager), we can delete both the common 2i2c account for openscapes as well as the spreadsheet.

Is the common account still associated with the deployment somehow? We should make sure the deletion is not going to impact the existing deployment.

> In general, unfortunately I think for AWS accounts we all need individual usernames and passwords. The good news is that we can create / delete them ourselves, so it isn't as involved as getting an external account via some other IT department!

Yep, maybe we can automate that :wink:?

GeorgianaElena:
I've rebased this and updated the info with the account ids for the aws projects that aren't under the 2i2c sso.

Does this still LGTM?

And another quick check. Are the leap, m2lines and friends also accessible from within the columbia gcp account? I've logged in earlier today, but couldn't find them

GeorgianaElena:
Thanks! I was searching them under the columbia one 🤦🏼‍♀️

sgibson91 (approved):

Issue #1806, opened 2022-10-20
Add hub engineer to GitHub org as admin to facilitate hub setup process


yuvipanda:
When setting up a hub, we have to do an OAuth approval process that requires action from the admin of the GitHub org that is going to be used to log in. This requires a back-and-forth with the GitHub admin that can take sometime (see https://github.com/2i2c-org/infrastructure/issues/1702#issuecomment-1285882952) for example.

Instead, part of our hub setup process should include:

1. Add the hub engineer in question as an org admin to the GitHub org that'll be used for access control
2. They remove themselves once auth is set up

sgibson91:
I have documented this in commit [`53dfe3a` (#1814)](https://github.com/2i2c-org/infrastructure/pull/1814/commits/53dfe3a7cf8ec5a38a9409615dbb525120d427dd) of #1814, but I think to truly close this issue, we also need to update the "New Hub" issue(s) to ensure that this is a checklist item there.

In general, I don't think those issue templates are yet configured to capture whether a community wants github org/team style auth yet. We should upgrade that.

Issue #1570, opened 2022-07-28
Allow access to uwhackweeks grafana outside of 2i2c


GeorgianaElena:
### Context

We recently received a request for Grafana access https://2i2c.freshdesk.com/a/tickets/158.

Currently, the way to do this is to enable GitHub authentication for that particular cluster Grafana and allow access to the community rep either by adding them to our 2i2c org and adding this org to the allowed list, or allowing another organization entirely, where the community rep has access. 

### Proposal

We should enable GitHub auth for the uwhackweeks Grafana and:

- figure out if we should allow an entire org
- or if we should allow only a specific GitHub team

Haven't tested last one, but there are docs about it here -> https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/github/#team_ids.

We would need to grab the id of that team and add that to the allow list instead of the team name. Docs about getting extra info about a team here https://docs.github.com/en/rest/teams/teams#get-a-team-by-name (thanks @sgibson91 ✨ )

GeorgianaElena:
Signaling the discussion in https://github.com/2i2c-org/infrastructure/pull/1830 which I believe is relevant for this one too

yuvipanda:
I think we should just grant people access as described in https://infrastructure.2i2c.org/en/latest/sre-guide/support/grafana-account.html?highlight=grafana

consideRatio:
Closing as uwhackweeks cluster is terminated

PR #1830, opened 2022-10-28
Enable GitHub auth for m2lines grafana

### LINKED ISSUE #1803: Update m2lines grafana
### Context

https://2i2c.freshdesk.com/a/tickets/229

### Proposal

Enable GitHub Authentication for the m2lines grafana running at https://grafana.m2lines.2i2c.cloud and redeploy the https://github.com/jupyterhub/grafana-dashboards there to get the latest dashboards.

### Updates and actions

- [x] Enable GitHub Authentication (docs [here](https://infrastructure.2i2c.org/en/latest/howto/operate/grafana.html#optional-enable-github-authentication))
- [x] Redeploy the https://github.com/jupyterhub/grafana-dashboards (docs [here](https://infrastructure.2i2c.org/en/latest/howto/operate/grafana.html#deploying-the-grafana-dashboards-from-ci-cd))

sgibson91:
fixes #1803 

This is deployed and works!

### DIFF
--- a/config/clusters/m2lines/support.values.yaml
+++ b/config/clusters/m2lines/support.values.yaml
@@ -11,6 +11,17 @@ grafana:
       - secretName: grafana-tls
         hosts:
           - grafana.m2lines.2i2c.cloud
+  grafana.ini:
+    server:
+      root_url: https://grafana.m2lines.2i2c.cloud
+    auth.github:
+      enabled: true
+      allow_sign_up: true
+      scopes: "user:email,read:org"
+      auth_url: https://github.com/login/oauth/authorize
+      token_url: https://github.com/login/oauth/access_token
+      api_url: https://api.github.com/user
+      allowed_organizations: 2i2c-org
 
 prometheusIngressAuthSecret:
   enabled: true
--- a/config/clusters/m2lines/enc-support.secret.values.yaml
+++ b/config/clusters/m2lines/enc-support.secret.values.yaml
@@ -1,17 +1,22 @@
 prometheusIngressAuthSecret:
-    username: ENC[AES256_GCM,data:Gs9mrGaLFU6hsYF8/wwpODWAt8xpcEu/heuDkneWhqiL47KISnLGBVOL6BsuljKKREJxuXOSwU8YTI/DPzUQ2w==,iv:0MVlwiPlxxRW4FLobzwaclFASLIfhuAolyjWkT6ytcs=,tag:xhC0iH+Gf+3E/ISk1D88aA==,type:str]
-    password: ENC[AES256_GCM,data:xQPRntL+XM1w6s41zFu3QKvrPGUV+egeJfqdHiqV8Ps18ebvWEYevUyt8fuujtmlEt9EJOqKoy7UrbaptmF2jA==,iv:HJLjntlpEGUiHKXsbfPPd42GJ/FBOnaO7nI3Ac1PnDs=,tag:2dveWGpIs0u9cfGp/at+Iw==,type:str]
+    username: ENC[AES256_GCM,data:bisiUHR25ma+Q2MTh8UcjaWDvX/RUVGmAvdfl95ehQ6vmAa0PmMRgZc6pmNwRQSV6+l/D7POI39UaXNCS/mUgg==,iv:See29zO/q0q3xeCij8Gx8Dk1KObVuA2snECw24c8bQc=,tag:osdytedbJLqwYBFwD3GlPg==,type:str]
+    password: ENC[AES256_GCM,data:CKBA27Rq5acAl+BrMENZvLcrG0BBNC6VQCKMldyenS6tcwX2yM0ZH3mZLspDlWSuNhyh7UxSK/Ks0pds+olWpQ==,iv:/O8DoKRPWktMFKwJ6IxpVhSpnISnJGiguOkr+CmF+OE=,tag:J4fTMgoV7NtTG8/iIUCW9w==,type:str]
+grafana:
+    grafana.ini:
+        auth.github:
+            client_id: ENC[AES256_GCM,data:AwOgzy07q0befQjbfzk3HYVHW54=,iv:DUjQ1ceqlSMpIdlDVpmpm3ILfrNszWcaD7dvyqZZNdI=,tag:gwQpWYI8gg4uelWQx0Yb/g==,type:str]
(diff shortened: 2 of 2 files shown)


config/clusters/m2lines/support.values.yaml
```diff
@@ -11,6 +11,17 @@ grafana:
       - secretName: grafana-tls
         hosts:
           - grafana.m2lines.2i2c.cloud
+  grafana.ini:
+    server:
+      root_url: https://grafana.m2lines.2i2c.cloud
+    auth.github:
+      enabled: true
+      allow_sign_up: true
```
sgibson91, line 19: The docs said to set this to `false` but I noticed it was set to `true` for the `leap` hub. Also, if we use `false` we would probably need to ask the comm. rep. to set up something like a GitHub team so that not just _anyone_ can sign in and view the boards. This would add back-and-forth to the setup though, unless _we_ do this explicitly as a part of https://github.com/2i2c-org/infrastructure/issues/1806

sgibson91: Tbf, this would ultimately be fixed if we instead authenticate against JupyterHub and allow hub admins to see the dashboards... Which could be a part of https://github.com/2i2c-org/infrastructure/issues/535? I'm not sure...

yuvipanda:
@sgibson91 does this allow *anyone* to sign up?

sgibson91:
@yuvipanda ok, I'm not sure. I believe so. This is the same config as is on the leap hubs and Ryan confirmed access, _but_ he is also a member of the 2i2c org, right? We might need someone not a member test...

sgibson91:
@yuvipanda The alternative is to create a team in the m2lines org and add the people who want to access these boards and then add that team to this config.

yuvipanda:
@sgibson91 What we can do is:

1. Grant access to just the 2i2c-org github team members
2. Use that access to *invite* the community rep (Grafana does similar things as hub does, see https://grafana.com/docs/grafana/latest/administration/user-management/manage-org-users/)
3. Community rep can then invite other users this way.

How does that sound?

sgibson91:
@yuvipanda So I changed `allow_sign_up: false`, redeployed, then signed in as the admin user. But I can't find an option to _invite_ a new user. I can _create_ one and set a password for them, is that the same thing?

Screenshot 2022-10-28 at 19 32 55
Screenshot 2022-10-28 at 19 33 42

Clicking on "New User" above takes me to the below screen

Screenshot 2022-10-28 at 19 33 49

GeorgianaElena:
A few thoughts from me:

### 1. `allow_sign_up` option 
> @sgibson91 does this (``) allow anyone to sign up?

  [The docs](https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/github/#enable-github-in-grafana) say the following:
  ```
  You may allow users to sign-up via GitHub authentication by setting the allow_sign_up option to true.
  When this option is set to true, any user successfully authenticating via GitHub authentication will be
  automatically signed up.
  ```
  
  So, from what I experienced so far and what I underdstand, the allow_sign_up automatically creates Grafana users for the users that **successfully** authenticated via GitHub. Otherwise, I believe an admin would have to first manually create an user in Grafana for them, before them logging in. So, it's ok to be set to true, and I believe the docs should be udpated.
  
  ### 2. The invite option

   > But I can't find an option to invite a new user. I can create one and set a password for them, is that the same thing?

  I believe the invite is in the `users` config section at https://grafana.m2lines.2i2c.cloud/org/users
Screenshot 2022-10-31 at 12 07 15

3. ### Moving forward

I propose to move the discussion around best ways to provide access to cluster grafanas in another issue so these two aspects don't block on each other and to have more stucture in these discussions.

 There is also another issue about this same aspect https://github.com/2i2c-org/infrastructure/issues/1570 which we could help unblock in the process. Unfortunately I now realize that we didn't solve the support request part of it, and the `uwhackweeks` folks might still don't have their access :(

sgibson91:
> [The docs](https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/github/#enable-github-in-grafana) say the following:
> 
> ```
> You may allow users to sign-up via GitHub authentication by setting the allow_sign_up option to true.
> When this option is set to true, any user successfully authenticating via GitHub authentication will be
> automatically signed up.
> ```
> 
> So, from what I experienced so far and what I underdstand, the allow_sign_up automatically creates Grafana users for the users that **successfully** authenticated via GitHub. Otherwise, I believe an admin would have to first manually create an user in Grafana for them, before them logging in. So, it's ok to be set to true, and I believe the docs should be udpated.

But if "successfully authenticated via GitHub" means "anyone with a GitHub account", then I don't think it's ok to set this to true. I think we need to figure out if Ryan successfully got access to LEAP's grafana because he is a member of the 2i2c GitHub org, and therefore this config **wouldn't** work for other community reps that aren't also 2i2c members. Maybe the UW Hackweeks hub will be a good testbed for this?

sgibson91:
> I believe the invite is in the `users` config section at [grafana.m2lines.2i2c.cloud/org/users](https://grafana.m2lines.2i2c.cloud/org/users) Screenshot 2022-10-31 at 12 07 15

How do you get to that page from the homepage? I can see the "Invite" option if I click directly on the link you sent, but I cannot find how to get to that page. All pages I come across only have the "New user" option. https://scribehow.com/shared/Trying_to_find_Invite_option_on_Grafana__GzOzUdEoQCOMMwsqUrvlsg

GeorgianaElena:
> But if "successfully authenticated via GitHub" means "anyone with a GitHub account", then I don't think it's ok to set this to true. I think we need to figure out if Ryan successfully got access to LEAP's grafana because he is a member of the 2i2c GitHub org, and therefore this config wouldn't work for other community reps that aren't also 2i2c members (we would need them to setup a specific team). Maybe the UW Hackweeks hub will be a good testbed for this?

@sgibson91, I just tested this using a test GitHub user and also checked the grafana pod logs. We should be safe, and only have 2i2c org be allowed
Screenshot 2022-10-31 at 12 49 19

Screenshot 2022-10-31 at 12 52 47

GeorgianaElena:
> How do you get to that page from the homepage? I can see the "Invite" option if I click directly on the link you sent, but I cannot find how to get to that page. All pages I come across only have the "New user" option. https://scribehow.com/shared/Trying_to_find_Invite_option_on_Grafana__GzOzUdEoQCOMMwsqUrvlsg

From the settings icon, then `Users` option:
Screenshot 2022-10-31 at 13 17 27

sgibson91:
I'm going to deploy/merge this one as-is for now so we can at least respond on the support ticket, and have opened this issue for more discussion:

- https://github.com/2i2c-org/infrastructure/issues/1850

Issue #1832, opened 2022-10-28
[EVENT] gridSST Hack-A-Thon


damianavila:
### Summary

Alison Gray, an oceanographer at the University of Washington, requests a hub to support an upcoming OceanHackWeek.
This event is actually going to be called the "gridSST Hack-A-Thon".

### Event Info

- **Community Representative:** @alisonrgray 
- **Event begin:** Oct 31st
- **Event end:** Nov 4th
- **Active times:** US/Pacific?
- **Number of attendees:** up to ~40
- [**Hub Events Calendar**](https://calendar.google.com/calendar/u/2?cid=Y19rdDg0c2g3YW5tMHNsb2NqczJzdTNqdnNvY0Bncm91cC5jYWxlbmRhci5nb29nbGUuY29t)

### Hub info

- **Hub URL**: https://gridsst.2i2c.cloud/
- **Hub decommisioned after event?**: No, hub should be live until 12-31-2022 (depends a lot on funding) 

### Task List

**Before the event**

- [ ] Dates confirmed with the community representative and added to Hub Events Calendar.
- [ ] Quotas from the cloud provider are high-enough to handle expected usage.
- [ ] **One week before event** Hub is running.
- [ ] Confirm with Community Representative that their workflows function as expected.
  - <details>
    <summary>👉Template message to send to community representative</summary>
    
    ```
    Hey {{ COMMUNITY REPRESENTATIVE }}, the date of your event is getting close!
    
    Could you please confirm that your hub environment is ready-to-go, and matches your hub's infrastructure setup, by ensuring the following things:
    - [ ] Confirm that the "Event Info" above is correct
    - [ ] On your hub: log-in and authentication works as-expected
    - [ ] `nbgitpuller` links you intend to use resolve properly
    - [ ] Your notebooks and content run as-expected
    ```
    
    </details>  
- [ ] **1 day before event**, either a separate nodegroup is provisioned for the event or the cluster is scaled up.

**During and after event**

- [ ] Confirm event is finished.
- [ ] Nodegroup created for the hub is decommissioned / cluster is scaled down.
- [ ] Hub decommissioned (if needed).
- [ ] Debrief with community representative.
  - <details>
    <summary>👉Template debrief to send to community representative</summary>
      
     ```
     Hey {{ COMMUNITY REPRESENTATIVE }}, your event appears to be over 🎉
     
[...]

PR #1836, opened 2022-10-31
Scale up the notebook nodepool for the gridsst cluster

GeorgianaElena:
Reference https://github.com/2i2c-org/infrastructure/issues/1832

~I haven't run the `eksctl scale nodegroup` command for these 50 nodes yet.~
~But I have run it for 10 to test it out.~

Question for the @2i2c-org/tech-team: Should we also have a min number of dask nodes too? I don't think we have docs about this, so I'm not sure if this is something that we have done before or if we should.

## Update

I am now running the scale up for 50 m5.large nodes

P.S.

@yuvipanda, I know you are the one assigned to  https://github.com/2i2c-org/infrastructure/issues/1832, but I wanted to experiment with scaling up an aws cluster and see how it goes, because I never did that. Hope that's ok! 
The event is outside of my timezone however, so you'll probably still need to put off any fires that will come up, sorry about that :/

### DIFF
--- a/eksctl/gridsst.jsonnet
+++ b/eksctl/gridsst.jsonnet
@@ -11,10 +11,10 @@ local nodeAz = "us-west-2a";
 // A `node.kubernetes.io/instance-type label is added, so pods
 // can request a particular kind of node with a nodeSelector
 local notebookNodes = [
-    { instanceType: "m5.large" },
-    { instanceType: "m5.xlarge" },
-    { instanceType: "m5.2xlarge" },
-    { instanceType: "m5.8xlarge" },
+    { instanceType: "m5.large", minSize: 50 },
+    { instanceType: "m5.xlarge", minSize: 0 },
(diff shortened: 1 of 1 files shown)


damianavila:
> Question for the https://github.com/orgs/2i2c-org/teams/tech-team: Should we also have a min number of dask nodes too? I don't think we have docs about this, so I'm not sure if this is something that we have done before or if we should.

We have not pre-warmed dask nodes in AWS land, AFAIR.
I would bet we never pre-warmed dask nodes at any place... although I might be wrong.
And, in general, I do not think dask nodes need to be pre-warmed, I think...

GeorgianaElena:
## Update

I am now running the scale up for 50 `m5.large` nodes

GeorgianaElena:
## Update 2

Scale up finished successfully.

yuvipanda (approved):

PR #1837, opened 2022-10-31
Add the enc token for grafana to be able to use the deployer with it

GeorgianaElena:
@sgibson91, apparently somehow I managed to forget this also


sgibson91 (approved):
<3

PR #1844, opened 2022-11-01
Scale down gridsst cluster

yuvipanda:


GeorgianaElena:
I'm also merging this since I checked and has already been deployed.
I'm just wondering now if we should do this scaling up and down every day to reduce costs?

damianavila:
> I'm just wondering now if we should do this scaling up and down every day to reduce costs?

In previous events, I have performed the scaling up and down on a daily basis using `eksctl` commands.
So it is definitely possible if we want to save them money, but also time consuming.

PR #1846, opened 2022-11-01
Increase resources available for gridsst

yuvipanda:
- Earlier, it was inheriting the default 1G memory values. Increase it per https://github.com/2i2c-org/infrastructure/issues/1832#issuecomment-1299083691
- Put all users on bigger nodes, as this leads to faster autoscaling.
- Add another admin

Ref https://github.com/2i2c-org/infrastructure/issues/1832


yuvipanda:
Already deployed to unblock in-progress event

yuvipanda:
Going to self-merge this to unblock the running event, although I had also deployed it from my local machine.

PR #1847, opened 2022-11-02
Stop storing hub logs in the db disk dir for UToronto

yuvipanda:
Per https://github.com/2i2c-org/infrastructure/pull/886/commits/9c30396a46202de639f12f5b924cd9fe03a8f086, this matched prior config. This usually is used for analytics purposes, particularly to figure out nbigpuller usage. However, by putting it in the same disk as the hub database, we can accidentally cause outages when the log file fills up the disk.

This PR completely removes this logging, although we could perhaps figure out a way to keep logs on a more specific basis.

### DIFF
--- a/config/clusters/utoronto/common.values.yaml
+++ b/config/clusters/utoronto/common.values.yaml
@@ -85,8 +85,6 @@ jupyterhub:
       JupyterHub:
         authenticator_class: azuread
         concurrent_spawn_limit: 100
-        # We wanna keep logs long term, primarily for analytics
-        extra_log_file: /srv/jupyterhub/jupyterhub.log
       AzureAdOAuthenticator:
         username_claim: oid
         login_service: "University of Toronto ID"


sgibson91 (approved):

damianavila:
> Can we open an issue to shape/track this idea please?

Done: https://github.com/2i2c-org/infrastructure/issues/1860

damianavila:
@yuvipanda, do you want to add something else here? If that is not the case, can you move it out from the draft state? Thanks!

damianavila (approved):

yuvipanda:
@damianavila I'm trying to figure out if we should block this on actually finding a place to store the hub logs or not, as there is possibly useful info there, particularly for figuring out nbgitpuller usage patterns. However, perhaps we don't have the cycles for that right now?

damianavila:
> actually finding a place to store the hub logs 

That makes sense to me... 

> However, perhaps we don't have the cycles for that right now?

There are probably some other higher-priority things, so I would be OK with the temporary removal since we already captured a follow-up in #1860.

yuvipanda:
I think this is ok for now on Azure, so going to close this one out.

Issue #1825, opened 2022-10-27
Deploy kubespawner from specific commit into our infrastructure


damianavila:
Ref: https://github.com/2i2c-org/infrastructure/issues/1253

@yuvipanda suggested deploying from a specific commit because of breaking changes in kubespwner.

GeorgianaElena:
Before going down this path, just want to first double-check with @consideRatio what's his feeling around a new `kubespawner` release and whether or not I can be of help.

@consideRatio, do you think now is a good time to push for a new `kubespawner` release? I believe I can help out with that.
If not, maybe I can help with pushing forward the blocking pieces of a release.

Nonetheless, I will be deploying an unreleased version for 2i2c until we figure it out.

consideRatio:
> @consideRatio, do you think now is a good time to push for a new kubespawner release? I believe I can help out with that.
If not, maybe I can help with pushing forward the blocking pieces of a release?

@GeorgianaElena do go for it! I think a key point about making a release is to figure out if it is a good time or not. I'd say its pretty much always a good time though - I'm thinking "why not make a release" and don't have a clear answer.

I think if you get started by creating a changelog, you will automatically end up learning about the changes made and can reflect if something is seen as problematic.

Btw here are some added details to what I think of when making a release, not part of RELEASE.md for specific projects. Certainly worth adding to and updating etc if you think of something! https://github.com/jupyterhub/team-compass/issues/563

GeorgianaElena:
Thanks for the guidance @consideRatio!

After speaking with @damianavila, I believe the plan now is:
- get 2i2c on https://github.com/jupyterhub/kubespawner/commit/fa19acdbb056c8d105baab13bbaa2b8b59dfa962 version so we don't block the new profile list options
- get started on building the changelog for a new kubespawner release and see where or if it blocks in the review process. We can then see which are the sensitive or blocking points if any.

GeorgianaElena:
I'm reopening this because the new hub image deploy has broke the staging hubs that used profile options because of bug https://github.com/jupyterhub/kubespawner/pull/662 I've introduced :(

I hadn't time to check the entire changelog of kubespawner. @yuvipanda, do you know if there are any breaking changes for 2i2c hubs specifically there if I were to use latest main commit?

Otherwise, I was thinking to pin to a commit on your fork https://github.com/yuvipanda/kubespawner/commits/async until I figure out the kubespawner changelog and possibly release?

Issue #1589, opened 2022-08-03
Hub image - using unreleased versions of some jupyterhub packages 


GeorgianaElena:
### Context

The 2i2c hubs run a custom docker image as per the Dockerfile at https://github.com/2i2c-org/infrastructure/blob/master/helm-charts/images/hub/Dockerfile. More details about this image and how to update it, can be found in the docs at https://infrastructure.2i2c.org/en/latest/topic/hub-image.html.

In this custom hub image we're currently installing unreleased versions of [`oathenticator`](https://github.com/2i2c-org/infrastructure/blob/master/helm-charts/images/hub/Dockerfile#L28) and [`configurator`](https://github.com/2i2c-org/infrastructure/blob/master/helm-charts/images/hub/Dockerfile#L18). This is because we want to try out the latest changes of this packages and improve/fix anything that might not work as it should.

This however means that we're using unreleased versions of these packages and we should make sure we're upgrading to released versions as soon as those are available.

### Proposal

A few points about the current situation:

- Our current mechanism for automating the packages upgrades (`dependabot` and `bump-image-tags` action) won't work on the pip packages in this `Dockerfile`
- Not aware of other software we could use that does this already (please comment to this issue, if you know any)
- 2i2c members are maintainers/contributors to releases of this packages and are more or less up to speed to the changes that happen there

Starting from the current situation described above, how about instead of always opening new issues about going back to an official released version of a package whenever we pin to a specific commit (as in https://github.com/2i2c-org/infrastructure/pull/1583#issuecomment-1203305618), we document this as being a best practice to follow? 

### Updates and actions

My suggested plan is to:
- [ ] document going back to an official released version of the packages in https://github.com/2i2c-org/infrastructure/blob/master/helm-charts/images/hub/Dockerfile, whenever these are available. Maybe https://infrastructure.2i2c.org/en/latest/topic/hub-image.html is the appropriate docs place for it?
- [ ] add inline comments whenever we pin to specific commits, to go back to a released version in https://github.com/2i2c-org/infrastructure/blob/master/helm-charts/images/hub/Dockerfile

damianavila:
> Our current mechanism for automating the packages upgrades (dependabot and bump-image-tags action) won't work on the pip packages in this Dockerfile

Quick question, what about putting these packages into a `requirements` file and `pip install` the `requirements` file in the Dockerfile... in that way we may use the dependabot, right?

sgibson91:
[...]

PR #1849, opened 2022-11-02
Update hub image

GeorgianaElena:
This updates the hub image with the following:

- bumps `oauthenticator` from the commit it was pinned, to the latest `15.1.0` release
- bumps and pins `kubespawner` to this commit https://github.com/jupyterhub/kubespawner/commit/fa19acdbb056c8d105baab13bbaa2b8b59dfa962

It also puts the custom packages we're installing alogside their pinned versions inside a `requirement.txt` to be picked up by dependabot.

Closes https://github.com/2i2c-org/infrastructure/issues/1589 and closes https://github.com/2i2c-org/infrastructure/issues/1825

### DIFF
--- a/docs/topic/hub-image.md
+++ b/docs/topic/hub-image.md
@@ -41,6 +41,12 @@ When this hub image needs to be updated, the steps to take are:
 
      This package is also listed under `dev-requirements.txt`, so it should be present if you've installed the dev dependencies.
 
+   - Make sure you are in the `helm-charts` directory, where the `chartpress.yaml` is located:
+
+     ```bash
+     cd ./helm-charts
+     ```
+
    - Run chartpress to build the image, push it to the registry and update the basehub helm chart to use the updated image tag
 
      ```bash


sgibson91 (approved):

sgibson91:
I think for dependabot to pick up these files, we should remove this line from the config:

https://github.com/2i2c-org/infrastructure/blob/74b06d59a7bff32873995db46962568f3eebb501/.github/dependabot.yaml#L35

I _believe_ then dependabot will look for any requirements files anywhere in the repo

sgibson91:
Also remember that merging this will overwrite what was done manually in https://github.com/2i2c-org/infrastructure/pull/1848

We should probably not wait for Yuvi to merge that and just do it now

GeorgianaElena:
> I believe then dependabot will look for any requirements files anywhere in the repo, as opposed to just the root

Hmm, I actually thought that the `/` in the directory, meant *search recursively everywhere in `/`*, which should have meant everywhere.
Apparently, neither the `/`, nor leaving directory empty means *please go and search everywhere for such files* :/ as the `directory` is a **required** field. 

https://github.com/dependabot/dependabot-core/issues/3951 has more context about specifying multiple directories for dependabot. It looks like we need to list each individual location for the requirements file

GeorgianaElena:
> We should probably not wait for Yuvi to merge that and just do it now

Good point! I will merge it before this one

sgibson91:
RE dependabot: dang. Let's just merge as-is and I'll open a follow-up PR for dependabot

sgibson91:
@GeorgianaElena I just went straight for the PR because I had the commit ready

- https://github.com/2i2c-org/infrastructure/pull/1855

yuvipanda:
Thanks a lot for catching #1848 dependency, @sgibson91!

PR #1814, opened 2022-10-25
Restructure infrastructure docs

sgibson91:
fixes #1757 

**New structure:**

- Adds a Site Reliability Engineering guide
- Adds a Hub Deployment Guide
- Splits up the grafana documentation
  - In the Hub Deployment Guide
    - Deploying the support chart itself
    - Deploying the grafana dashboards and setting up CI/CD
    - Registering the cluster-specific prometheus with the central grafana
  - How-to
    - Optionally enable github auth for dedicated-cluster grafanas
  - Topic
    - What is the central grafana
    - How to access grafana
    - How to read the dashboards

**New documentation added:**

- Community-specific CNAMEs (ref: https://github.com/2i2c-org/infrastructure/issues/1757#issuecomment-1285659593)
- Troubleshooting --> Common problems and their solutions (in SRE guide)
  - How to resize the NFS disk when it is full (ref: https://github.com/2i2c-org/docs/issues/100#issuecomment-925962073)
  - How to reset GitHub OAuth apps when users get a 403 error (ref: https://github.com/2i2c-org/infrastructure/issues/1757#issuecomment-1287027429)

**Updated docs/other fixes:**

- fixes #1815 
- minor styling changes such as consistent casing for headings, adding new labels where appropriate, ensuring tabbed content syncs and uses the most modern dependencies
- Improvements to the new cluster docs
- restructure the topic sections
- group "enable specific hub features" subpages into categories for the howto section
- make it explicit that engineers should get org admin access when setting up GitHub auth
  - ref: https://github.com/2i2c-org/infrastructure/issues/1806

PR #1851, opened 2022-11-02
Remove repeated config in m2lines support file

sgibson91:
This is to stop pre-commit complaining in #1814 

https://results.pre-commit.ci/run/github/291960479/1667397065.8wB-nrlrSQysolH6BKRrgA


sgibson91:
I will self-merge this as it is a small change and will fix CI

PR #1857, opened 2022-11-02
Add profileOptions for gridsst hackathon

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/1832#issuecomment-1300980838


damianavila:
Quick question, @yuvipanda. You still have defined memory, CPU, and node selector outside of the profile option as per a previous change/PR. Should those be removed? Are you not generally limiting on the CPU side with the current configuration (since memory and node selector are overwritten by the profile option)?

damianavila:
Answered by https://github.com/2i2c-org/infrastructure/pull/1861.

PR #1863, opened 2022-11-03
Signal that dependabot PRs for the hub image require additional steps

GeorgianaElena:
We don't actually want to merge dependabot's PRs for updating the pip dependencies in the hub image, because updating it requires additional manual steps to take https://infrastructure.2i2c.org/en/latest/topic/infrastructure/hub-image.html#updating-the-hub-image

These PRs should be instead treated as a signal that new versions have been released and we need to check them out.


GeorgianaElena:
Context issue https://github.com/2i2c-org/infrastructure/issues/1589

sgibson91:
Could write a GitHub Actions workflow that checks opened PRs that come from a `dependabot/images/hub/...` branch (dependabot puts the location of the package file into the branch name) and comments on them with a link to the docs on how to fully update and merge them.

A "do not merge" label here feels a little bit ambiguous, because we don't want to merge them after the required steps have happened - maybe it should be "requires follow-up"?

GeorgianaElena:
Great ideas @sgibson91! ✨  I just added a couple of commits to implement your suggestions. LTMK what you think

PR #1864, opened 2022-11-03
[CI/CD] Refactor cluster validation workflow file

### LINKED ISSUE #1862: [CI/CD] Refactor helm validation workflow to dynamically generate matrix list of clusters to validate
We could also refactor the [validation workflow](https://github.com/2i2c-org/infrastructure/blob/master/.github/workflows/validate-clusters.yaml) so that the step that inspects changed files (and hence uses `GITHUB_TOKEN`) is extracted from the matrix and only executed once, we then dynamically generate a matrix of clusters to validate, similar to the deploy-hubs workflow.

_Originally posted by @sgibson91 in https://github.com/2i2c-org/infrastructure/issues/1686#issuecomment-1300604831_

sgibson91:
fixes #1862 

## Motivation

Over in #1686, it is reported that we are running into rate-limits on the use of `GITHUB_TOKEN` on a semi-regular basis and so the restructuring of this workflow is motivated by the desire to reduce the number of times we use `GITHUB_TOKEN` both in terms of times per workflow and workflow triggers.

### Structure

The big change here is that the original job in the workflow has been split into two jobs. The first job now inspects file changes caused by a Pull Request and dynamically generates a matrix strategy of which clusters to validate based on those changed files (similar to our deploy-hubs workflow). This is the part that uses `GITHUB_TOKEN` in order to inspect the changed files as part of the `dorny/paths-filter` action and **crucially this job is now only executed once per workflow run**. The second job runs cluster validation over the matrix generated by the first job and does not require use of the `GITHUB_TOKEN`.

The original workflow singular job had a matrix strategy that ran the `dorny/paths-filter` step in each matrix job which was very wasteful in terms of `GITHUB_TOKEN` since this would have generated the same results in each matrix job.

### The `dorny/paths-filter` action

This action uses `GITHUB_TOKEN` to inspect what files have changed, either in a Pull Request or from the `HEAD~1` commit in the case of push events.

Originally, we called this action twice: once to establish which 'common' files had changed (which would trigger all clusters to be validated), and again to establish which 'cluster-specific' files had changed (to decide whether or not the validate the cluster at all, providing common files hadn't changed).

Actually, we could combine these two filters into a single call, again saving the number of times `GITHUB_TOKEN` is used - so we do that now. Each filter is given its own output so we can still reference each filter individually despite the combination of the execution.

I also added some improvements to the invocation of this action:

1. **Set `list-files: csv` so we can access a csv formatted string of changed files for each defined filter.** We specifically use `outputs.cluster_specific_files` to dynamically generate the matrix of clusters to validate when validating _all_ clusters is not required.
2. **Use `added|modified` syntax to restrict what kind of changes we care about.** We generally only care about validating files when they are added or changed, not when they are deleted.

### Branch triggers

**Note:** If this section doesn't resonate, we can drop the following commits on this PR to revert: [`c8614ce` (#1864)](https://github.com/2i2c-org/infrastructure/pull/1864/commits/c8614ce27a53aa2d84646dadfe631a4fb3112e6d) and [`8d9e7a3` (#1864)](https://github.com/2i2c-org/infrastructure/pull/1864/commits/8d9e7a343aba02ccd6f35e9945dc425d6e089bbd)

Screenshot 2022-11-03 at 14 37 08

Above is a screenshot of the original trigger settings. The problems with these settings are:

1. We were running this workflow on every pull request opened, not only the ones that will be merged in to `master` so that is more unnecessary invocations of `GITHUB_TOKEN` (at least for PRs within this repo, forks will use their own token).
2. We were running this workflow on every push to any branch and only excluding branches that were created by dependabot or pre-commit. So the workflow would be triggered twice in a pull request: once from the PR trigger, once from the push trigger. Again, wasteful for `GITHUB_TOKEN` use within the main repo.

To address these issues I set the following so the workflow only runs in relation to the master branch:

```yaml
push:
  branches:
    - master
pull_request:
  branches:
    - master
```

Unfortunately when using the `pull_request` trigger, the branches listed refer to the **target** branch of the merge, and not the head branch of the merge. Therefore, it is not trivial to restrict workflows running on PRs _from_ certain branches (unless you use the `pull_request_target` trigger, which comes with its own difficulties). And so I wrote a condition for the first job such that it runs when:

- the workflow was manually triggered
- the event is a push to master
- the event is a PR to master that is **not** dependabot or pre-commit generated

### DIFF
--- a/.github/workflows/validate-clusters.yaml
+++ b/.github/workflows/validate-clusters.yaml
@@ -1,7 +1,7 @@
 # This is a GitHub workflow defining a set of jobs with a set of steps. ref:
 # https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions
 #
-# Runs the deployer script to validate clusters. This will both validate
+# Runs the deployer module to validate clusters. This will both validate
 # cluster.yaml files, any non-encrypted values files for the support chart (if they
 # exist), as well as each hubs passed non-encrypted values files against the Helm
 # charts' values schema.
@@ -99,7 +99,7 @@ jobs:
           with open(env_file, "a") as f:
               f.write(f"MATRIX={matrix}")
 
-      # Only run this step if there are *NO* changes under the common filepaths,
+      # Only run this step if there are *NO* changes under the common filter,
       # but *ARE* changes under the cluster_specific filter
       - name: Generate a matrix containing only clusters that have changes
         if: |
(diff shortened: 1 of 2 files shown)


yuvipanda:
Thanks a lot for working on this, @sgibson91!

Re: branches, I think we should always try to work in our own forks and then make PRs to this repo. I think reverts from the UI are the only things that don't follow that pattern. Is that right? If so, this makes total sense and we can also probably codify that somewhere (like conda-forge does)

sgibson91:
> Re: branches, I think we should always try to work in our own forks and then make PRs to this repo. I think reverts from the UI are the only things that don't follow that pattern. Is that right?

Yeah, PR reverts are the only thing that don't follow that pattern.

I'll also hold my hands up and say I have, on occasion, made some PRs from branches within the same repo mostly because they were easy changes to make _in the browser_, compared to pulling up VSCode and pushing something from my fork. It is rare though and probably not a change that would trigger any of these workflows.

> If so, this makes total sense

Just to check - is that an endorsement to keep this PR as it is?

> we can also probably codify that somewhere

Yes!

yuvipanda:
@sgibson91 yes total endorsement to keep PR as is. And yes I've done that too many, *many* times (re: edit from the web UI to make a PR). I wish it defaulted to doing that as a branch on my fork, because that's how it works when I try to use the web to make PRs on projects I don't have write access to. Oh well.

yuvipanda (approved):
This looks good to me based on my understanding of how GHA does things! Let's merge and find out :)

sgibson91:
Hmm, I merged and tried using the manual trigger to validate everything, and the second job skipped. Will debug that tomorrow.

sgibson91:
One benefit I forgot to mention about this change is that we no longer need to _manually_ add new clusters to this workflow, because they will get dynamically picked up by the file changes

sgibson91:
Below is the fix:

* https://github.com/2i2c-org/infrastructure/pull/1868

PR #1867, opened 2022-11-03
Add warning to make sure we ask for quota in right region

yuvipanda:
I almost asked for quota in wrong region just now :)


sgibson91 (approved):
This is the kind of real-world testing this documentation needed!

yuvipanda:
whee, thanks @sgibson91 :)

Issue #1472, opened 2022-06-27
Change deployer sub-scripts into deployer actions


GeorgianaElena:
### Context

Currently, the `deployer` module has lots of useful functions, available as deployer actions https://github.com/2i2c-org/infrastructure/blob/master/deployer/README.md|

But there are also deployer "sub-scripts" like [`update_central_grafana_datasources.py`](https://github.com/2i2c-org/infrastructure/blob/master/deployer/update_central_grafana_datasources.py) or [`cilogon_app.py`](https://github.com/2i2c-org/infrastructure/blob/master/deployer/cilogon_app.py).

Having these sub-scripts in the same place with various deployer utilities makes it hard to know which one is which. 

### Proposal

We should have the sub-scripts ported as deployer actions.

### Updates and actions

Sub-actions to be taken here:

- [x] Move deployer/cilogon_app.py
- [x] Move deployer/update_central_grafana_datasources.py
- [x] Move shared/keys/decrypt-age.py

_No response_

yuvipanda:
Sub-actions to be taken here:

- [ ] Move `deployer/cilogon_app.py`
- [ ] Move `deployer/update_central_grafana_datasources.py`
- [ ] Move `shared/keys/decrypt-age.py`

Issue #970, opened 2022-02-02
Let the deployer script be a standalone package


consideRatio:
### Description of problem and opportunity to address it

**Context to understand the problem**

"In the name of R2R" as @sgibson91 said, we should let the deployer script be a standalone module. I figure it should be a python package published to PyPI with changelog and semantic versioned releases etc.

**Problem or idea**

If we do this, we will force ourselves to think of the deployer script as the standalone tool we expect others to use if they decouple with 2i2c. By planning for this, we are way better prepared to help people fly on their own!

**Initial incremental step**
- [x] Make the deployer script a pip installable package within this repo
  See https://github.com/2i2c-org/infrastructure/issues/970#issuecomment-1028004884 by Sarah below
  See https://github.com/2i2c-org/infrastructure/pull/1869

**Proposed solution**

- [ ] We create a repo in 2i2c-org for the deployer script
- [ ] Add pre-commit CI stuff
- [ ] Add a few pytest tests for what is easy to test to start with
- [ ] Add MyST documentation for the project
  - [ ] Include a changelog
- [ ] Cut a release
- [ ] Let's plan for when we release 1.0.0 and make it not be far into the future
  By not getting stuck at 0.X.Y, we can communicate breaking changes a lot easier.

**What's the value and who would benefit**
This would allow users to decouple from 2i2c after a while if they wish and take on the responsibilities themselves, migrating to their standalone management while retaining use of the deployer script if they wish.

sgibson91:
I think a small change we could make even before all that is mentioned here is to add a `setup.py` file to `infrastructure` to make the deployer installable at least. Our invocations would then look something like:

```bash
python setup.py install
deployer deploy ...
```

The `setup.py` file can then be migrated out with the rest of the codebase as and when that action is taken

damianavila:
I like the idea of a standalone package!!

Regarding the setup.py comment @sgibson91 made above, maybe it is time to use new standards such as flit (or some other packaging tool) + pyproject.toml (and no setup.py)?

choldgraf:
Also agree this would be a really nice way to support R2R with our work. A question:

Why wouldn't this just be upstream improvements to `hubploy`? Do we imagine this replacing `hubploy` in the future? I think that this would be particularly impactful if we think that we can *push the boundary* and make something particularly useful for people who are deploying hubs, and not just turn this into a tool purely for 2i2c's R2R purposes. What do others think?

sgibson91:
[...]

PR #1869, opened 2022-11-04
Add debug helper scripts, make into a minimal package & use typer

yuvipanda:
This PR ended up being a bit longer than I had hoped, but I can't think of easy ways to disentangle this! Here's the three things that are happening here.

# 1. New scripts useful for debugging

While helping debug the last UToronto incident, I realized I've a bunch of useful aliases (https://github.com/yuvipanda/dotfiles/blob/master/aliases) focused on debugging JupyterHub issues that might be helpful for others too. Instead of fiddling with shell, I instead wrote a `debug.py` that can be invoked from the commandline similar to `exec-homes-shell`!

Want to look at the logs of the JupyterHub pod in the prod hub in the UToronto cluster? You can now with:

```
python3 -m deployer.debug component-logs utoronto prod hub
```

This command takes care of authenticating appropriately, and figuring out the correct kubectl commands to make this work. You can also look at the proxy logs with:

```
python3 -m deployer.debug component-logs utoronto prod proxy
```

or the dask-gateway controller logs with

```
python3 -m deployer.debug component-logs 2i2c dask-staging dask-gateway-controller
```

See `python3 -m deployer.debug component-logs --help` for more information.

In addition to looking at component-logs, I've added the following subcommands:

1. `user-logs` to look at logs from the user pod, given a particular username. It takes care of appropriate escaping for you as well, so you can give it the username the user sees without having to do any replacements.
2. `exec-hub-shell` gives you a shell in the hub pod itself, so you can poke around. Particularly helpful to poke around the sqlite db itself, or, uh, to delete logs if they're too big
3. `exec-homes-shell` is same as the old `exec-homes-shell`, which was under deployer. I think this is a better fit, so it's here now.

# 2. [typer](https://typer.tiangolo.com/) for CLI management

As I was trying to add these extra commands, the duplication when using `argparse` got a little frustrating to me. Adding a new command required we write a bunch of boilerplate, and duplicate help strings in multiple places. I looked around with ways to fix it, and found [typer](https://typer.tiangolo.com/) from the same nice people who made FastAPI.
It leverages type hints (present since python 3.6) to minimize amount of boilerplate needed for making new subcommands, and I *immensely* like it! It is what [click](https://click.palletsprojects.com/en/8.1.x/) could have been if click was built post Python 3.6. typer is built alongside FastAPI and uses an extremely similar API, so I'm confident it'll survive long term. 

I've also moved the existing commandline parsing code to using typer, and I think overall it is clearer because of this. There's less duplication, and adding new commands is easier as well. Should also help with https://github.com/2i2c-org/infrastructure/issues/1472

# 3. Making the deployer into a package https://github.com/2i2c-org/infrastructure/issues/970

While setting up the added script, I ran into a lot of difficulty with relative imports (or the lack thereof). The simplest way to fix this was to make it a 'real' module, so I did the *simplest* thing possible to do so:

1. Add a super simple `setup.py` file
2. Make all imports be relative
3. Add console scripts for `deployer`, so it can be simply called as `deployer` rather than as `python deployer`

Again, the primary goal here was to make sure that relative imports work fine, rather than anything to do with actually packaging this for use.


yuvipanda:
@sgibson91 I know you might also have to fiddle with these scripts and CI when debugging https://github.com/2i2c-org/infrastructure/issues/1862 and friends. Am happy to rebase this after that is done (if needed)

choldgraf:
fun fact, it seems like Type actually [uses Click under the hood](https://typer.tiangolo.com/alternatives/#click)

sgibson91:
> debugging #1862 and friends. Am happy to rebase this after that is done (if needed)

I actually think the conflict will be with #1870 

I never successfully got onboard with writing cli's with click or similar because I conceptually struggle with the decorator aspect and... yeah, it just never worked for my mental model. This was a few years ago though, so maybe typer is easier than I found click, or click is easier than it was then. Who knows! Either way, I'll give the job of providing an official review to someone on the team with the conceptual know-how.

A couple of general things...

1. I LOVE LOVE **_LOVE_** that we are moving towards the deployer being standalone! Thank you also for updating all of the documentation and GHA related stuff with the change to the invocation.
2. **Requested change.** I appreciate the minimal `setup.py`, but the files under the `shared` directory need to be ingested too. The deployer depends on those files and the whole folder would move if we were to extract the deployer to it's own repo, so we need to ensure they're included in the packaging too. I think JupyterHub does something like this:
   - https://github.com/jupyterhub/jupyterhub/blob/main/setup.py#L27
   - https://github.com/jupyterhub/jupyterhub/tree/main/share/jupyterhub
3. **For a follow-up PR.** Can we agree on a cli standard for commands/subcommands, _any_ standard, and document it somewhere? We will now have `deployer <command>`, `deployer.debug <command>` and then there's the by itself `deployer/update_central_grafana_datasources.py`, and it keeps me up at night 😅 In terms of documentation location, I think we could put this standard in `deployer/README.md` for now, but I could imagine it moving into a `CONTRIBUTING.md` if/when we split the deployer out into its own repo.

yuvipanda:
Thank you very much for your comment, @sgibson91! Here's a couple things I've done:

1. I've gotten rid of the separate `deployer.debug` command, and put everything just back under `deployer`. So we still have one command (`deployer`) with a bunch of subcommands, same as before. It might make sense to split that out into multiple commands in the future, but this simplifies the setup for now so we don't have to write any guidelines for what goes where :D 
2. Instead of the pattern from jupyterhub (and its `share/` directory), I've instead used the pattern from `repo2docker` instead, which puts the files the package needs inside the package itself. So I've moved the *one* file that is needed for the package to function (the json schema), and made sure to include it via `MANIFEST.in`. This turned out to be the pattern we were already using, as we are already including files there we need (the tests and the test notebooks!). This won't work outside this repo yet (as I think there are probably other things that need doing), but I don't personally want to work on that part quite yet - just wanted to do the least possible to get this working, and make sure I'm not doing any antipatterns.
3. I appreciate the concern around click, and for what it's worth I could never wrap my head around click either. I tried again this time, and came to the same conclusion as before - I just couldn't really figure out how to use click. I was going to go back to argparse before I found typer, and I definitely found it much more intuitive to use than I ever did click. I'm very happy to put in whatever time (or changes) required to make sure you feel comfortable working on the codebase! Please let me know (here, or in private) and I'll prioritize that. It also has default rich integration which I quite like (see the new command help output in the README, much nicer!)

I'm going to debug these test failures now to see what is up.

yuvipanda:
Tests fixed, ready to go!

sgibson91 (approved):
Let's get this merged before anything else underneath changes :)

yuvipanda:
Yay thanks @sgibson91!

PR #1874, opened 2022-11-04
Add GPUs for gridsst

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/1832


yuvipanda:
Going to self-merge as https://github.com/2i2c-org/infrastructure/issues/1832#issuecomment-1304003512 is around and their event ends in a few hours

yuvipanda:
Already deployed

PR #1876, opened 2022-11-07
Enable dependabot version bumps for the setup-deploy action

sgibson91:
In our original config, dependabot only checks actions to bump under the `.github/workflows` folder. This PR adds an extra rule to also check actions under the `.github/actions/setup-deploy` folder.

**Note:** We may have to rename `action.yaml` to `action.yml` for this to work, but let's try it first and see.
ref: https://github.com/dependabot/dependabot-core/issues/5970

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -23,7 +23,7 @@
 version: 2
 updates:
   # Maintain dependencies for GitHub Actions in workflow files
-  # By default, the directory will check .github/workflows for the github-actions
+  # By default, the directory to be checked is .github/workflows for the github-actions
   # package ecosystem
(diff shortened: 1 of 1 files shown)


.github/dependabot.yaml
```diff
@@ -22,7 +22,10 @@
 
 version: 2
 updates:
-  # Maintain dependencies for GitHub Actions
+  # Maintain dependencies for GitHub Actions in workflow files
+  # By default, the directory will check .github/workflows for the github-actions
```
sgibson91, line 26:
```suggestion
  # By default, the directory to be checked is .github/workflows for the github-actions
```

GeorgianaElena (approved):
🚀

sgibson91:
Woohoo! It worked without needing to rename the file!

PR #1881, opened 2022-11-07
Improve list of hubs script

choldgraf:
This adds a few improvements to our "list of hubs" script:

- Factors it out into a dedicated script instead of being embedded in `conf.py` -> makes it easier to debug locally, and de-clutters `conf.py`
- Adds a little "table of summary statistics" that filters out staging/demo hubs and breaks it down by cluster/provider
- Makes the tables more compact so they don't take up as much vertical space


docs/reference/hubs.md
```diff
@@ -1,16 +1,16 @@
+
 # List of running hubs
 
-Here's a list of running hubs.
+% NOTE: The CSV files used by this page are generated by `docs/scripts/render_hubs.py`
+
+Here's a list of running hubs and some statistics about them.
 It is automatically generated from the config stored in the [`config/clusters` folder of the `infrastructure` repository](https://github.com/2i2c-org/infrastructure/tree/HEAD/config/clusters) and can therefore be limited in the information it provides.
 
-```{admonition} Missing information for Azure hubs
```
sgibson91, line 6: I wouldn't remove this. There's a reason we say "most likely running on Azure". Testing for `kubeconfig` is not a robust test for azure based clusters, even though it's true right now. We could export a `kubeconfig` for any cluster because it is a _kubernetes specific thing_, not a cloud vendor specific thing.

docs/scripts/render_hubs.py
```diff
@@ -0,0 +1,110 @@
...
+                support_config.get("grafana", {}).get("ingress", {}).get("hosts", "")
+            )
+            # If we find a grafana url, set it here and break the loop, we are done
+            if isinstance(grafana_url, list):
+                grafana_url = grafana_url[0]
+                grafana_url = f"[{grafana_url}](http://{grafana_url})"
+                break
+
+    # Track the provider for this cluster for use later
+    provider = cluster["provider"] if cluster["provider"] != "kubeconfig" else "azure"
```
sgibson91, line 60: Be careful with this. While _right now_ it's true that the only `kubeconfig` clusters we have are on Azure, we could feasibly export a `kubeconfig` for all cloud vendors and use that. It's not a robust test for "running on Azure".

docs/reference/hubs.md
choldgraf: is there a more reliable way to infer the cloud provider from the information in there? Maybe I am using the wrong field?

sgibson91: The only reliable field is cluster provider, so I don't think we can reliably say "this is an Azure cluster" right now. This is because we only run one, and I doubt we have the correct permissions on that account to create the Azure version of service account and hence have the azure-specific config to authenticate. This is when we fall back onto `kubecobfig`

choldgraf: What do you recommend we do here? Just list "kubeconfig" in the table instead of azure?

sgibson91: Yeah, I think so. That was the decision I came to when I first introduced that column.

choldgraf: ok will do in next push

docs/scripts/render_hubs.py
```diff
@@ -0,0 +1,117 @@
...
+                "name": hub.get("display_name", "name"),
+                "domain": f"[{hub['domain']}](https://{hub['domain']})",
+                "id": hub["name"],
+                "hub_type": hub["helm_chart"],
+                "grafana": grafana_url,
+                "cluster": cluster["name"],
+                "provider": provider,
+                "data center location": datacentre_loc,  # Americanising for you ;)
+            }
+        )
```
sgibson91, line 87: I'm a little bit confused about how these two loops work together? The `for`'s seem to be at the same indentation so it looks like you run one loop first, then the other? From what I can work out, the first loop sets `provider` to `azure` and then gets used in the next loop, but I think this would set all providers to `azure` which isn't what we want? I will check the PR preview but yeah... I'm confused...

sgibson91: Oh ok, because the loop over the cluster's is waaaaay up the file, got it

docs/scripts/render_hubs.py
```diff
@@ -0,0 +1,117 @@
...
+
+    # Define the cloud provider for this cluster which we'll insert into the table below
+    # For clusters where we know the provider but can't guess it from the YAML file,
+    # we define a few custom mappings.
+    custom_providers = {"utoronto": "azure"}
+    for clustername, provider in custom_providers.items():
+        if cluster["name"] == clustername:
+            provider = provider
+        else:
+            provider = cluster["provider"]
```
sgibson91, line 67: I think you can more neatly write this as:

```python
if cluster["name"] in custom_providers.keys():
    provider = custom_providers[cluster["name"]]
else:
    provider = cluster["provider"]
```

Saves you looping over `custom_providers` every time you execute the loop

choldgraf, line 67:
```suggestion
    if cluster["name"] in custom_providers.keys():
        provider = custom_providers[cluster["name"]]
    else:
        provider = cluster["provider"]
```

sgibson91:
What are the files `docs/howto/hubs/new-hub/deploy.md` and `docs/howto/hubs/new-hub/information.md`? They are empty. I also don't think we have a `docs/howto/hubs` folder any more after the refactor - probably need to rebase onto master or delete these files.

docs/conf.py
```diff
@@ -79,84 +79,7 @@ def setup(app):
 import pandas as pd
```
choldgraf, line 79:
```suggestion
```

choldgraf:
ah good call - I guess they were accidentally added back in. I've deleted them in this PR

choldgraf:
gonna merge this one in so that we can iterate if needed

PR #1882, opened 2022-11-07
Disable IPython history manager across all hubs

yuvipanda:
By default, each cell being executed in a notebook adds an entry to this sqlite database on NFS. sqlite and NFS do not mix, causing hard to debug problems everywhere. This disables the history manager. Consoles aren't affected, just notebooks.

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -194,6 +194,15 @@ jupyterhub:
       # Most people do not expect this, so let's match expectation
       SHELL: /bin/bash
     extraFiles:
+      ipython_kernel_config.json:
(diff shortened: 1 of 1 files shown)


consideRatio (approved):

consideRatio:
> Consoles aren't affected, just notebooks

Affected by this issue of nfs + sqlite databases, or by this configuration making them no longer be used?

yuvipanda:
@consideRatio the latter

yuvipanda:
@consideRatio I also wrote a blog post about this https://words.yuvi.in/post/no-touch-nfs/

damianavila:
Should we link this blog post from our docs somehow so our users know about it?

PR #1884, opened 2022-11-08
Remove unused secret_key

yuvipanda:
Thanks to all the work by @consideRatio and @sgibson91, we don't need to generate any secrets at all!


consideRatio:
Wieee!

Could you update the commit to not be noisy begore merge?

Mentions in commits when people do things in forks, rebase, etc, they often end up creating noise - sometimes a long time after making the commit as well.

sgibson91 (approved):

yuvipanda:
Done, @consideRatio!

PR #1894, opened 2022-11-09
Use correct version of dask-gateway

yuvipanda:
Dask gateway versions don't have a `v` in front of them.

Follow-up to https://github.com/2i2c-org/infrastructure/pull/1869


yuvipanda:
Going to merge to unbreak CI

PR #1897, opened 2022-11-09
Use the latest kubespawner release in hub image

GeorgianaElena:
Follow-up to https://github.com/2i2c-org/infrastructure/issues/1825

Also,  because this PR wasn't opened by dependabot on Monday, I believe it means dependabot doesn't work for git pip packages :/?

So maybe it's worth reopening https://github.com/2i2c-org/infrastructure/issues/1589?

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -388,7 +388,7 @@ jupyterhub:
         admin: true
     image:
       name: quay.io/2i2c/pilot-hub
-      tag: "0.0.1-n4378.h27667b54"
+      tag: "0.0.1-n4497.haed80ebb"
     nodeSelector:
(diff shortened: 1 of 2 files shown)


sgibson91:
> dependabot doesn't work for git pip packages

I don't think it does and I believe that is out of scope for dependabot anyway. We should be using officially released versions anyway, so pinning to a specific commit (and the consequent revert back to a released version) should be a manual process, IMO. Because we are overriding the behaviour of how versioning works.

sgibson91 (approved):

GeorgianaElena:
> We should be using officially released versions anyway, so pinning to a specific commit (and the consequent revert back to a released version) should be a manual process, IMO. Because we are overriding the behaviour of how versioning works.

Agree!

PR #1903, opened 2022-11-11
Expand the `generate_cluster` command

GeorgianaElena:
This expands the `generate_cluster` command a bit.

#### 2. Adds support for a "`GCP`" generator that generates a `cluster_name`.tfvars file using:
- some input cluster characteristics
- two `.tfvars` "template" files:
        - a `basehub.tfvars` config file with the most common features of the educational basehub hubs we're currently deploying the most
        -  a `daskhub.tfvars` config file with the most common features of the research dask-enabled hubs we're currently deploying the most
        - these templates also contain some `Tips:` that I believe would be useful when clusters will need to have some other particularities than the defaults, as they would provide a starting search point.
 
Ref: #1917


deployer/README.md
```diff
@@ -42,6 +42,8 @@ This section descripts all the deployment related subcommands the `deployer` can
 │ exec-homes-shell                    Pop an interactive shell with the home directories of the given hub mounted on   │
 │                                     /home                                                                            │
 │ exec-hub-shell                      Pop an interactive shell in the hub pod                                          │
+│ generate-cluster                    Automatically generate the files required to setup a new cluster on a specific   │
```
yuvipanda, line 45: I usually set `COLUMNS=120` before exporting the help file so it all aligns nicely

eksctl/template.jsonnet
```diff
@@ -17,19 +17,19 @@ local notebookNodes = [
...
-// kind of node with a nodeSelector
-local daskNodes = [
-    { instancesDistribution+: { instanceTypes: ["m5.large"] }},
-    { instancesDistribution+: { instanceTypes: ["m5.xlarge"] }},
-    { instancesDistribution+: { instanceTypes: ["m5.2xlarge"] }},
-    { instancesDistribution+: { instanceTypes: ["m5.8xlarge"] }},
-];
-
+local daskNodes =
+    if "<< hub_type >>" == "daskhub" then [
```
yuvipanda, line 21: We can define this as a jsonnet variable earlier, and use that in the conditional here?

terraform/gcp/projects/basehub-template.tfvars
```diff
@@ -0,0 +1,31 @@
+prefix                 = "{{ cluster_name }}"
+project_id             = "{{ cluster_name }}"
+
+zone                   = "{{ cluster_region }}-b"
```
yuvipanda, line 4: We should probably just explicitly ask for the zone as well - in us-central1-b, we can't have GPUs :(

terraform/gcp/projects/basehub-template.tfvars
```diff
@@ -0,0 +1,31 @@
...
+
+core_node_machine_type = "n1-highmem-4"
+enable_network_policy  = true
+
+# Setup a filestore for in-cluster NFS
+enable_filestore       = true
+filestore_capacity_gb  = 1024
+
+# No plans to provide storage buckets to users on this hub, so no need to deploy config connector
+config_connector_enabled = false
```
yuvipanda, line 15: This defaults to false, so we can omit it

yuvipanda:
Thanks a lot @GeorgianaElena!

GeorgianaElena:
I believe this is be ready for review, and might be a decent enough first step towards https://github.com/2i2c-org/infrastructure/issues/1917?

There is still lots to do still, (including properly documenting this), but I believe its best to be done in following PRs to make it easier to review and provide feedback about the assumptions  made in each of them before going further.

yuvipanda (approved):
Doing this iteratively sounds great to me, thanks @GeorgianaElena

PR #1904, opened 2022-11-11
Add a --previous flag to user-logs

yuvipanda:
In case the user pod has restarted, this gives us the ability to look at the logs from the previous instance


sgibson91 (approved):
Good to go once the README has been updated too :)

yuvipanda:
Updated, @sgibson91!

Issue #1843, opened 2022-11-01
[Incident] Prometeus is down in pangeo-hubs


damianavila:
### Summary

* PageDuty incident: https://2i2c-org.pagerduty.com/incidents/Q0ZH5DBGBCSUNP (it says resolved but @yuvipanda confirmed this one has recurred).
* Dedicated Slack channel: https://app.slack.com/client/TKHFGBRC2/C046PBG348J

### Impact on users

This is not impacting the users, but we have zero visibility into usage on the pangeo-hubs.

### Important information

- Hub URL: us-central1-b.gcp.pangeo.io
- Support ticket ref: No support ticket, AFAIK this one was caught by the uptime check.

### Tasks and updates

- [ ] Discuss and address incident, leaving comments below with updates
- [ ] Incident has been dealt with or is over
- [ ] Copy/paste the after-action report below and fill in relevant sections
- [ ] Incident title is discoverable and accurate
- [ ] All actionable items in report have linked GitHub Issues

<details>
<summary>After-action report template</summary>

```
# After-action report

These sections should be filled out once we've resolved the incident and know what happened.
They should focus on the knowledge we've gained and any improvements we should take.

## Timeline

_A short list of dates / times and major updates, with links to relevant comments in the issue for more context._

All times in {{ most convenient timezone}}.

- {{ yyyy-mm-dd }} - [Summary of first update](link to comment)
- {{ yyyy-mm-dd }} - [Summary of another update](link to comment)
- {{ yyyy-mm-dd }} - [Summary of final update](link to comment)


## What went wrong

_Things that could have gone better. Ideally these should result in concrete
action items that have GitHub issues created for them and linked to under
Action items._

- Thing one
- Thing two

## Where we got lucky

_These are good things that happened to us but not because we had planned for them._

- Thing one
- Thing two

## Follow-up actions

_Every action item should have a GitHub issue (even a small skeleton of one) attached to it, so these do not get forgotten. These issues don't have to be in `infrastructure/`, they can be in other repositories._

### Process improvements

1. {{ summary }} [link to github issue]
2. {{ summary }} [link to github issue]

### Documentation improvements

1. {{ summary }} [link to github issue]
2. {{ summary }} [link to github issue]

### Technical improvements

1. {{ summary }} [link to github issue]
2. {{ summary }} [link to github issue]
```

</details>

damianavila:
Some notes from @yuvipanda's investigation in Slack:

Stage 1:

[...]

PR #1906, opened 2022-11-11
Increase memory & disk size limit for pangeo prometheus

yuvipanda:
While investigating why the pangeo hubs prometheus was dead, I discovered https://github.com/prometheus/prometheus/issues/6934 - where there can be a big temporary *spike* in memory usage when prometheus is recovering from a restart. It tries to read the WAL (the write-ahead log), to make sure it hasn't lost any data during the restart process itself. The details of the WAL are unimportant (in this specific case), but just the fact that prometheus spikes memory usage on restarts!

I manually hand edited the prometheus deployment with `k -n support edit deployment support-prometheus-server`, and gave it a higher limit (8G). Then I watched actual memory usage, with `watch kubectl -n support top pod`. I noticed that it momentarily spiked to almost 5G, before settling back to about 1.5G. The old memory limit was 4G, so during the spike the server gets killed! And then enters crashloopbackoff, as it can never survive.

This commit raises the memory limit, so it won't keep crashing :)

I also actually manually increased the size of the disk (with `kubectl -n support edit pvc`), but that wasn't the problem. However, we need to persist the change regardless, so here it is.

Hopefully this will fix https://github.com/2i2c-org/infrastructure/issues/1843

### DIFF
--- a/config/clusters/pangeo-hubs/support.values.yaml
+++ b/config/clusters/pangeo-hubs/support.values.yaml
@@ -18,6 +18,8 @@ prometheusIngressAuthSecret:
 
 prometheus:
   server:
+    persistentVolume:
+      size: 200Gi
     ingress:
       enabled: true
       hosts:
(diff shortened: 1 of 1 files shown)


sgibson91:
@yuvipanda can we document the manual debug steps here please? https://infrastructure.2i2c.org/en/latest/sre-guide/common-problems-solutions.html In case this happens on another cluster in the future.

sgibson91 (approved):

yuvipanda:
will do, @sgibson91!

PR #1914, opened 2022-11-14
Fix typo: Decryptor key is under secrets, not inputs

sgibson91:
follow-up to #1913


sgibson91:
Will self-merge this as it fixes a typo and will unbreak CI

Issue #1678, opened 2022-09-05
Lock up the Callysto hub


GeorgianaElena:
### Context

The Callysto hub is pretty open right now, so we should find ways to protect it against potential abuse.
We currently have [cryptnono deployed](https://docs.2i2c.org/en/latest/about/infrastructure/security.html#cryptocurrency-mining) on all of our clusters, but we should find ways to improve this protection, esp for hubs like the Callysto one.

Ref: https://github.com/2i2c-org/infrastructure/pull/1649#pullrequestreview-1094607859

colliand:
This is a good idea, especially to avoid unauthorized cloud usage explosions. That said, the Callysto use case involves their personnel introducing data science and coding ideas to teachers and students in grades 5-12 across Canada. These users tend to not be organized under CIlogon or InCommon authorization layers. Their usage scenario often involves launching students and teachers they've just met into the hub without forming an allow list in advance.

ianabc:
This is going to come up for us in a couple of weeks. We have a series of workshops where various groups of users will be logging and and trying out the service. Right now, it looks like users are being whitelisted, but that might not be practical for those workshops. As far as I know, we have

* Nov 18th ~ 20-30 users
*  Nov 24th ~ 50-60 users
* Nov 25th ~ 180-190 users
* December 6th ~ 120 users

In all cases, we want to make sure that the hub is ready (the right number of nodes warmed up by 9am) and that users can log in OK. Can you let me know what the best options are for this currently? For warming up hubs, I found [this issue](https://github.com/2i2c-org/infrastructure/issues/787), but I'm not sure what the current status is. If it is appropriate, I think we would be comfortable manually triggering the node scale up if needed.

byrcyb:
For the upcoming hackathons that @ianabc mentioned in November and December 2022, these will be with specific school divisions, is it possible to whitelist all emails associated with those school divisions instead of leaving the hub open for all to sign up? For example, whitelisting `xxxxx@schooldivisionA.ca, xxxxx@schooldivisionB.ca, xxxxx@schooldivisionC.ca, etc`

yuvipanda:
@byrcyb @ianabc if those schools are part of CILogon (you can check via https://technical.edugain.org/entities and https://incommon.org/community-organizations/) we can restrict them to that.

@ianabc re: scale up, what do you feel about us making sure you have rights to trigger the warmup yourself from the cloud console?

byrcyb:
[...]

PR #1919, opened 2022-11-15
Allow restricting Callysto logins by domain

yuvipanda:
With Callysto, we do not use email addresses to try preserve user privacy a little. However, this makes restricting the domains users can login be a bit difficult - all our validation code validates the username itself, rather than any other attribute of it.

Here, we customize the authenticator to validate the email of the user, but not use that as the username. This allows us to enforce a list of domains (including wildcards) that can be allowed to login without having to set their emails to be their usernames.

Ref https://github.com/2i2c-org/infrastructure/issues/1678

### DIFF
--- a/config/clusters/callysto/common.values.yaml
+++ b/config/clusters/callysto/common.values.yaml
@@ -55,19 +55,65 @@ jupyterhub:
             <img src="https://www.callysto.ca/wp-content/uploads/2022/08/Callysto-HUB_horizontal.png" alt="CallystoHub"/>
           {% endblock %}
   hub:
+    extraConfig:
+      001-cilogon-email-auth: |
+        # Custom override of CILogonOAuthenticator to allow access control via one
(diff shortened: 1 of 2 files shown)


yuvipanda:
Would love to hear how we can upstream this as well, @GeorgianaElena :)

GeorgianaElena (approved):
This looks like the best solution! Thank you! 🚀

Do you think we should upstream this availability to allow based on domain, even on the claim being used is not `email`? I believe it might be useful for others?

GeorgianaElena:
> Would love to hear how we can upstream this as well, @GeorgianaElena :)

Hhah, you were faster than me! Yes, I also believe this should be upstreamed 🚀

yuvipanda:
@GeorgianaElena yesss we should! And please do open an issue :)

GeorgianaElena:
Done https://github.com/jupyterhub/oauthenticator/issues/547 🚀

Issue #1925, opened 2022-11-16
Remove all config generation in the deployer


yuvipanda:
From a right to replicate perspective, our end users should be able to take the config out of this repo, and be able to use a `helm` command to continue to deploy the hubs, passing in just the config. The current impediment to this is the bits of config we autogenerate from the deployer: https://github.com/2i2c-org/infrastructure/blob/e351ddd68894a45f778caae7278fb1720056156d/deployer/hub.py#L27. That entire function should just go away. We'll just have to manually put in the config there as needed in all our hubs instead. 

# TODO:

- [ ] Move the hub homepage config out of deployer generated config
- [ ] Move ingress config out of deployer generated config
- [ ] Move binderhub registry config out of deployer generated config
- [ ] Move off auth0 completely, as CILogon already doesn't require we use deployer generate config

yuvipanda:
Just absolutely amazing work, @GeorgianaElena!!!!!!!

PR #1928, opened 2022-11-16
Remove entire binderhub generated config from deployer

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/1925


config/clusters/2i2c/binder-staging.values.yaml
```diff
@@ -1,10 +1,18 @@
 binderhub:
+  ingress:
+    hosts:
+      - binder-staging.2i2c.cloud
+    tls:
+      - secretName: https-auto-tls-binde
```
sgibson91, line 6:
```suggestion
      - secretName: https-auto-tls-binder
```

config/clusters/2i2c/binder-staging.values.yaml
```diff
@@ -35,3 +43,12 @@ binderhub:
   extraVolumeMounts:
     - name: custom-templates
       mountPath: /etc/binderhub/custom
+
+  jupyterhub:
+    ingress:
+      hosts:
+        - hub.binder-staging.2i2c.cloud
+      tls:
+        - secretName: https-auto-tls-binde
```
sgibson91, line 52:
```suggestion
        - secretName: https-auto-tls-binder
```

sgibson91 (approved):
It would be nice if there was some kind of templating we could so that we didn't have to repeat the URL as often _without_ obfuscating it in the deployer, specifically. Is that something we could achieve as we walk down the path of merging basehub and daskhub?

Not a blocker to merging, just a musing.

PR #1935, opened 2022-11-18
Bump version of dask-gateway

yuvipanda:
There are new changes here that will help a lot with memory usage!

See https://gateway.dask.org/changelog.html#id1

### DIFF
--- a/deployer/deployer.py
+++ b/deployer/deployer.py
@@ -199,7 +199,7 @@ def deploy(
         help="File to read secret deployment config from",
     ),
     dask_gateway_version: str = typer.Option(
-        "2022.11.0", help="Version of dask-gateway to install CRDs for"
+        "2023.1.0", help="Version of dask-gateway to install CRDs for"
     ),
 ):
     """
--- a/deployer/deployer.py
+++ b/deployer/deployer.py
@@ -199,7 +199,7 @@ def deploy(
         help="File to read secret deployment config from",
     ),
     dask_gateway_version: str = typer.Option(
-        "2022.10.0", help="Version of dask-gateway to install CRDs for"
+        "2022.11.0", help="Version of dask-gateway to install CRDs for"
     ),
 ):
     """
--- a/helm-charts/daskhub/Chart.yaml
+++ b/helm-charts/daskhub/Chart.yaml
@@ -9,7 +9,7 @@ dependencies:
     repository: file://../basehub
     # If bumping the version of dask-gateway, please also bump the default version set
     # in the deployer's CLI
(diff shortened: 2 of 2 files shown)


consideRatio (approved):
What part is influencing the memory usage?

deployer/deployer.py
```diff
@@ -199,7 +199,7 @@ def deploy(
         help="File to read secret deployment config from",
     ),
     dask_gateway_version: str = typer.Option(
-        "2022.10.0", help="Version of dask-gateway to install CRDs for"
+        "2022.11.0", help="Version of dask-gateway to install CRDs for"
```
consideRatio, line 202:
```suggestion
        "2023.1.0", help="Version of dask-gateway to install CRDs for"
```

helm-charts/daskhub/Chart.yaml
```diff
@@ -9,7 +9,7 @@ dependencies:
     repository: file://../basehub
     # If bumping the version of dask-gateway, please also bump the default version set
     # in the deployer's CLI
-    # https://github.com/2i2c-org/infrastructure/blob/HEAD/deployer/cli.py#L73-L80
+    # https://github.com/2i2c-org/infrastructure/blob/HEAD/deployer/deployer.py#L195
   - name: dask-gateway
-    version: "2022.10.0"
+    version: "2022.11.0"
```
consideRatio, line 14:
```suggestion
    version: "2023.1.0"
```

consideRatio (approved):
- dask-gateway the client in the user env images has support for Python 3.11
- dask-gateway-server run in the Helm chart now runs with Python 3.11
- I know dask/distributed also supports Python 3.11 at this point in time for anyone updating their user environment image

The performance benefits from `distributed`'s newer is coming from bumping the user environment image, as that will be used to start a scheduler pod and worker pods.

Bumping dask-gateway itself won't impact the performance though I think.

consideRatio (approved):

yuvipanda:
I wanna version match with the dask-gateway in the images, so opened https://github.com/conda-forge/pangeo-dask-feedstock/pull/104

consideRatio:
I've seen all PRs and no PR has touched the REST api between dask-gateway's client and the server, where the rest involving dask is just dask distributed's client working against a scheduler/worker started using the same image as in the user images by dask gateways server. Due to that, I'm quite confident its safe to use mixed versions in general - there shouldn't be breaking changes for dask-gateway client/server interactions.

yuvipanda:
ok that's pretty awesome, thanks @consideRatio!!!

Issue #1917, opened 2022-11-15
Enhance the `deployer` to make us more efficient at deploying hubs


GeorgianaElena:
## I. Context

Now that the `deployer` has been refactored https://github.com/2i2c-org/infrastructure/pull/1869, building on top of it has become more scalable.

We are currently using the `deployer` to automatize various manual tasks related to hub deployment and management and with each option added to it, the engineer team have become more efficient 👉🏼  https://github.com/2i2c-org/infrastructure/blob/master/deployer/README.md

The deployer is a tool to help us, but is too over-powered for someone not in our engineering team to use. The manual steps are important because that is how a hub admin would setup the hub without 2i2c involvement, from a Right to Replicate perspective. (_credit goes to Sarah for signal boosting this aspect_)

### Personal experience

After deploying two similar clusters and hubs on aws, I manage to build some muscle memory and the time it took to deploy the second one was cut in half. But this still meant a day, which in my opinion is a lot.

I believe a significant amount of time was spent on:

1. context switching between following the docs and tasks done in the terminal, editor and various UI's (aws, namecheap, Grafana, GitHub)
2. copying config over from other similar hubs, making sure you've enabled all of the features requested and also, not introducing unneeded customizations

## II. Proposal

The `deployer` is ready for a new round of enhancements, maintenance and  documentation in order to help us become more efficient at deploying new clusters and hubs, but also form the r2r perspective to remove any config generation out of it, while creating and maintaining a clear documentation of the deploy workflow.

- [ ] **ENHANCEMENTS**:  improve the time it takes to deploy a new cluster and hub using the `deployer`
- [ ] **MAINTENANCE:** be thoughtful of the Right to Replicate and remove any config generation from the deployer
- [ ] **DOCUMENTATION:** make sure our expose in documentation, the "magic" happening in the deployer  so that the steps can be performed outside of 2i2c and outside the deployer.

### 1. ENHANCEMENTS

- [ ] https://github.com/2i2c-org/infrastructure/issues/2198

**`deployer.generate_cluster`**
```[tasklist]
- [x] Create template files for the **terraform** infrastructure files and add support to `deployer.generate_cluster` command to generate them https://github.com/2i2c-org/infrastructure/pull/1903
- [x] https://github.com/2i2c-org/infrastructure/pull/1937
- [ ] https://github.com/2i2c-org/infrastructure/issues/2187
```

**`deployer.generate_hub`**
```[tasklist]
[...]

PR #1937, opened 2022-11-21
Add support to generate both infrastructure and config files for new gcp clusters

GeorgianaElena:
- Separates aws and gcp cluster generators into their own files. This should make it easier to remove the aws one once eksctl is gone. We can then generalize the gcp one to support both

- The new cluster generator now looks like this (currently only gcp has been expanded to generate config files also)
    - generate infrastructure files (eg. the terraform file)
    - generate the config files i.e. the ones in the `config/clusters/new-cluster` dir
       - a `cluster.yaml` file with a sample config for an initial hub (usually staging)
       - the `support.values.yaml` file
       - the `enc-support.values.yaml` file with the prometheus credentials
       
  There's also a PR coming up about generating the grafana token programmatically

https://github.com/2i2c-org/infrastructure/issues/1917


deployer/generate_aws_cluster.py
```diff
@@ -76,27 +76,6 @@ def aws(cluster_name, hub_type, cluster_region):
...
-        "cluster_zone": cluster_zone,
-    }
-
-    with open(
-        REPO_ROOT / "terraform/gcp/projects" / f"{cluster_name}.tfvars", "w"
-    ) as f:
-        f.write(tfvars_template.render(**vars))
-
-
 @app.command()
```
yuvipanda, line 79: can we move this to a `generate/` subdirectory?

deployer/generate/generate_aws_cluster.py
```diff
@@ -5,7 +5,7 @@
 import jinja2
 import typer
 
-from .cli_app import app
+from deployer.cli_app import app
```
consideRatio, line 8: I see source code using relative imports and absolute imports. I think the source code should stay consistently with relative imports.

In this case you can do this instead I think.

```suggestion
from ..cli_app import app
```

The one exception I see to use absolute imports is for anything outside the root module, for example if there is a `jupyterhub` folder for the jupyterhub source code, and a `tests` folder for the tests. Then the tests in the tests folder should reference `from jupyterhub import <stuff>` while if code is writting from within the jupyterhub folder, it would relate to code like `from .generate import stuff` or `from ..cli_app import app` etc.

GeorgianaElena: Thanks @consideRatio ✨ ! I usually believe that being explicit tends to be better, though I don't always respect this rule. I actually used relative imports here and changed it before pushing it upstream for this reason.

My **very personal** preference is to use relative imports when in the same directory (or when prototyping something and want to be quick), and use absolute ones when going up the hierarchy for readability and clarity purposes. 

But I agree this yells inconsistency and I understand why using relatives everywhere has its advantages. I added a commit to switch to relative imports https://github.com/2i2c-org/infrastructure/pull/1937/commits/b3ac262096958e7489f2b2a2d8e698264153d74b (and also spotted an issue with the repo root path thanks to this)

consideRatio: Oooooh!!!

I've seen so much relative imports in the jupyterhub ecosystem, which I've used as a guide on whats right, that I assumed it was the recommended practice. Its the other way around according to PEP8 - absolute imports are recommended, but not strictly enforced.

- PEP8: https://peps.python.org/pep-0008/#imports
  image
- A text about absolute vs relative imports: https://realpython.com/absolute-vs-relative-python-imports/#pros-and-cons-of-absolute-imports

Thanks for iterating with me on this @GeorgianaElena! Do feel free to adjust as you think is best and disregard my previous comments where I thought relative imports were recommended.

GeorgianaElena:
I believe this should be ready for review again.
I know there aren't any proper docs that integrate this feature into the deploying a new cluster process :(
But those docs will follow given that there is agreement on:
- the templates introduced here
- the overall look and feel of the "generator"

.github/workflows/validate-clusters.yaml
```diff
@@ -13,20 +13,22 @@ on:
     branches:
       - master
     paths:
-      - config/clusters/**
-      - helm-charts/**
-      - deployer/**
-      - requirements.txt
-      - .github/workflows/validate-clusters.yaml
+      - "config/clusters/**"
+      - "!config/clusters/templates"
```
yuvipanda, line 17: Add a comment here on why the `!` is here?

.github/workflows/validate-clusters.yaml
```diff
@@ -13,20 +13,22 @@ on:
...
```
yuvipanda, line 27: Same as above?

.github/workflows/validate-clusters.yaml
```diff
@@ -110,7 +112,8 @@ jobs:
           # Construct a matrix of all clusters
           matrix = []
           for cluster in cluster_folders:
-              matrix.append({"cluster_name": cluster})
+              if cluster != "templates":
```
yuvipanda, line 115: Add a comment on why?

config/clusters/templates/gcp/cluster.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - support.values.yaml
+    - enc-support.secret.values.yaml
+hubs:
+  - name: {{ hub_name }}
+    # Tip: consider changing this to something more human friendly
+    display_name: "{{ cluster_name }} - {{ hub_name }}"
+    domain: {{ hub_name }}.{{ cluster_name }}.2i2c.cloud
+    helm_chart: {{ hub_type }}
+    auth0:
+      # Tip: Uncomment the line below and enable auth0
```
yuvipanda, line 20: Let's remove this by comment by default, as we don't really wanna use auth0 for anything new.

deployer/generate/generate_gcp_cluster.py
```diff
@@ -0,0 +1,153 @@
...
+    with open(REPO_ROOT / "config/clusters/templates/gcp/support.values.yaml") as f:
+        support_values_yaml_template = jinja2.Template(f.read())
+
+    with open(cluster_config_directory / "support.values.yaml", "w") as f:
+        f.write(support_values_yaml_template.render(**vars))
+    print_colour(f"{cluster_config_directory}/support.values.yaml created")
+
+    # Generate and encrypt prometheus credentials into `enc-support.secret.values.yaml`
+    print_colour("Generating the prometheus credentials encrypted file...", "yellow")
+    alphabet = string.ascii_letters + string.digits
```
yuvipanda, line 70:
```suggestion
    alphabet = string.ascii_letters + string.digits + string.punctuation
```

To add special chars too for the password

yuvipanda (approved):
Some suggested changes, feel free to merge and move with it after those are done :)

sgibson91 (approved):
+1 on merging after Yuvi's comments around commenting have been addressed

GeorgianaElena:
Thanks @yuvipanda and @sgibson91 🚀

Issue #1838, opened 2022-10-31
Renew Grafana tokens for a few clusters


GeorgianaElena:
### Context

When I first setup the grafana tokens for some of the clusters to use them in the CI, I made available for only 1y.

Some of those have expired and we need to re-generate them and add them to the repo encrypted.

The list of expired tokens is probably the same with the list of failing workflows here:
<img width="427" alt="Screenshot 2022-10-31 at 16 36 23" src="https://user-images.githubusercontent.com/7579677/199034140-d157475f-d4e1-409e-8c2d-6b14615a7a79.png">

PR #1938, opened 2022-11-21
Add a new deployer command to generate grafana tokens

GeorgianaElena:
This PR creates a new `deployer` command that generates (or re-generates) a Grafana API token and stores it into `enc-grafana-token.secret.yaml` file.

It also generates new tokens  (using this new cmd option), for the clusters in #1838 that have expired tokens.

Notable things that changed:
   - deployer grafana-related commands were moved into their own dir and module.
   - a few grafana-related functions from the `central_grafana.py` file were moved into an `utils` file
   -  the new command also allows to overwrite an existing token and shows whether it has expired or not
   - we are not generating API keys anymore, but tokens for service accounts as per Grafana's latest update 
      Screenshot 2022-11-24 at 14 52 27

More about Grafana service accounts here https://grafana.com/docs/grafana/latest/administration/service-accounts/ 

Closes https://github.com/2i2c-org/infrastructure/issues/1838
Reference https://github.com/2i2c-org/infrastructure/issues/1917


deployer/grafana/central_grafana.py
```diff
@@ -0,0 +1,161 @@
+"""
+Ensures that the central grafana at https://grafana.pilot.2i2c.cloud
+is configured to use as datasource the authenticated prometheus instances
+of all the clusters that we run.
+"""
```
GeorgianaElena, line 5: This is not new content! It's the same code, with only some of the comments updates.
I believe this is why GitHub got fooled to believe it's a new file and not one that got moved 🙄

deployer/grafana/grafana_utils.py
```diff
@@ -0,0 +1,149 @@
...
+
+    # Read the secret grafana token file
+    with get_decrypted_file(grafana_token_file) as decrypted_file_path:
+        with open(decrypted_file_path) as f:
+            config = yaml.load(f)
+
+    return config["grafana_token"]
+
+
+def update_central_grafana_token(cluster_name, token):
```
GeorgianaElena, line 121: Only this function is new. All the others used to exist in `central_grafana.py`.
There are only minor tweaks to the existing ones, like updating the function decription and fixing a small bug https://github.com/2i2c-org/infrastructure/pull/1938/commits/fe182d700f9533cf1974531f81b9e31944cb70c3

deployer/README.md
```diff
@@ -192,6 +196,21 @@ This allows us to optimise and parallelise the automatic deployment of our hubs.
...
+and wait for cli input about whether to generate a new one or not.
+
+The API token is stored encrypted inside the `enc-grafana-token.secret.yaml` file of the cluster.
+If such a file doesn't already exist, it will be created by this function.
+
+  Generates:
+    - an `enc-grafana-token.secret.yaml` file if not already present
+
+  Updates:
+    - the content of `enc-grafana-token.secret.yaml` with the new token if one already existed
```
consideRatio, line 213: I love these kinds of summaries!!

Could you add some clarity what kinds of API tokens (a Grafana API token?), and service accounts (a grafana service account? a k8s service account?) are meant, and also what kind of credential is used to set things up?

GeorgianaElena: Added! Thanks @consideRatio! I spent this time with my head inside Grafana-stuff and forgot there exist other things  😅

deployer/grafana/grafana_tokens.py
```diff
@@ -0,0 +1,210 @@
...
+    grafana_password_string = subprocess.check_output(
+        [
+            "sops",
+            "--decrypt",
+            REPO_ROOT / "helm-charts/support/enc-support.secret.values.yaml",
+        ]
+    ).decode("utf-8")
+
+    # Parse the sops output to find the admin's username password
+    keyword = "adminPassword: "
```
yuvipanda, line 34: You can use https://github.com/2i2c-org/infrastructure/blob/348a88b436edcb14d104577a8479d9edd9a5ca84/deployer/file_acquisition.py#L72 to read the encrypted file, and it'll handle decryption correctly. We should also definitely use yaml to read the decrypted content, rather than using string parsing I think.

deployer/grafana/grafana_tokens.py
```diff
@@ -0,0 +1,210 @@
...
+        # Check if there's a "deployer" token for the "deployer service account" already
+        token = get_deployer_token(sa_endpoint, sa_id, headers)
+
+    if token:
+        print_colour(
+            "Token already exists!\n"
+            f'Checking if has expired: {token["hasExpired"]}\n',
+            "yellow",
+        )
+        print_colour(
```
yuvipanda, line 189: Is there a reason we would want to keep an expired token? Could we automatically replace them if it exists and is expired?

yuvipanda (changes requested):

deployer/grafana/grafana_tokens.py
GeorgianaElena: Haha, yes! I have used this before. I have no idea why this time my impulse was to reinvent the wheel do everything the ugly manual way 😅 Thanks for raising this

deployer/grafana/grafana_tokens.py
GeorgianaElena: Yes, no reason to keep them.
This script creates tokens with no expiration date however. I couldn't find in their docs how to set one https://grafana.com/docs/grafana/latest/developers/http_api/serviceaccount

GeorgianaElena:
@2i2c-org/tech-team, signal boosting that this has addressed prior feedback and should now be ready for review again. Thank you!

deployer/README.md
```diff
@@ -43,15 +43,23 @@ This section descripts all the deployment related subcommands the `deployer` can
 │                                     /home                                                                            │
 │ exec-hub-shell                      Pop an interactive shell in the hub pod                                          │
 │ generate-aws-cluster                Automatically generate the files required to setup a new cluster on AWS          │
+<<<<<<< HEAD
```
sgibson91, line 46: Left over from a merge conflict?

deployer/README.md
```diff
@@ -43,15 +43,23 @@ This section descripts all the deployment related subcommands the `deployer` can
...
 │ generate-gcp-cluster                Automatically generates the initial files, required to setup a new cluster on    │
 │                                     GCP                                                                              │
+=======
+│ generate-gcp-cluster                Automatically generate the terraform config file required to setup a new cluster │
+│                                     on GCP                                                                           │
+>>>>>>> e5f0bce8 (Update the README)
```
sgibson91, line 52: merge conflict?

deployer/README.md
```diff
@@ -194,6 +202,22 @@ This allows us to optimise and parallelise the automatic deployment of our hubs.
 ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
 ```
 
+### `new-grafana-token`
+This function uses the admin credentials located in `helm-charts/support/enc-support.secret.values.yaml` to check if a [Grafana Service Account](https://grafana.com/docs/grafana/latest/administration/service-accounts/) named `deployer` exists for a cluster's Grafana, and creates it if it doesn't.
+For this service account, it then generates a Grafana token named `deployer`.
+This token will be used by the [`deploy-grafana-dashboards` workflow](https://github.com/2i2c-org/infrastructure/tree/HEAD/.github/workflows/deploy-grafana-dashboards.yaml) to authenticate with Grafana’s HTTP API
```
sgibson91, line 208:
```suggestion
This token will be used by the [`deploy-grafana-dashboards` workflow](https://github.com/2i2c-org/infrastructure/tree/HEAD/.github/workflows/deploy-grafana-dashboards.yaml) to authenticate with Grafana’s REST API
```

I mean, probably :D

deployer/grafana/central_grafana.py
sgibson91: Yeah the comment updates will always make git thinks it's a new file in diffs, unfortunately. From the object database point of view, this is a completely new object, and that's why it behaves in this way.

deployer/grafana/grafana_tokens.py
```diff
@@ -0,0 +1,208 @@
...
+    Assumes:
+        the maximum number of service accounts created for a grafana is no more than 10.
+
+    Returns:
+        id: the id of the service account with the name `deployer`
+            `None` if no service account with this name was found
+    """
+
+    # Assumes we don't have more than 10 service accounts
+    # FIXME if this changes
```
sgibson91, line 81: We already have 17 clusters - 18 once #2022 is merged. Assuming one grafana instance per cluster and they will all be using a service account, maybe best to bump this number now?

docs/hub-deployment-guide/deploy-support/setup-grafana-dashboards.md
```diff
@@ -23,22 +23,24 @@ Here are the credentials for logging in:
...
-Create the file `config/clusters/<cluster>/grafana-token.secret.yaml` with the following content.
-
-```yaml
-grafana_token: PASTE_YOUR_API_KEY_HERE
+```bash
+deployer new-grafana-token <cluster_name>
 ```
 
-Then encrypt this file using `sops` like so:
+If the command succeded, it should have created:
```
sgibson91, line 35:
```suggestion
If the command succeeded, it should have created:
```

small typo

sgibson91 (approved):
LGTM!

I have a small wondering on file naming. For example, now that we have a `grafana` subfolder, I would say that `grafana/utils.py` and `grafana/manage_tokens.py` would reduce redundancy compared to `grafana/grafana_utils.py` and `grafana/grafana_tokens.py` etc, particularly when importing, i.e. `from grafana import utils` versus `from grafana import grafana_utils`. In the `utils`-specific case we'd need to avoid naming clashes with the top-level `utils` though. Which indicates to me either (i) we are overusing `utils` and should be more creative in our filenaming, or (ii) I am overthinking package design :D

I don't think this should block merging this PR, but something for us to think about going forward.

deployer/grafana/grafana_tokens.py
sgibson91: Hmmm, I guess there are some long-lived tokens that will mean we might not get to this number?

deployer/README.md
GeorgianaElena: Grafana calls it HTTP API in their docs https://grafana.com/docs/grafana/latest/developers/http_api/ and I think this is why I called it http api also?

But I went and searched if there are any differences and found out that http api it's is own thing apparently https://www.programsbuzz.com/article/what-difference-between-rest-api-and-http-api.

I will link these in the docs to have them at hand. Thanks for sparking a dive into this topic. I had no idea they are different

sgibson91: Me either - that's super interesting!

deployer/grafana/grafana_tokens.py
GeorgianaElena: *` # Assumes we don't have more than 10 service accounts`* should have been *` # Assumes we don't have more than 10 service accounts per Grafana`*.

Sorry for the confusion. I added a commit to improve the comments around the service accounts per Grafana https://github.com/2i2c-org/infrastructure/pull/1938/commits/7645df1e66a1ba32a0bca6e82d3f8e379407700b.
 LMK is the comments are more clear now.

Basically, with this PR, we are creating the infrastructure to create our first service accounts **per Grafana**, called `deployer`. In case we will create other service accounts in the future for rbac purposes for example, and that number is >10 we should update this. But my guess is that this might take a while, if ever.

sgibson91: Ahhh, per grafana makes way more sense - thanks!

GeorgianaElena:
> I have a small wondering on file naming. For example, now that we have a grafana subfolder, I would say that grafana/utils.py and grafana/manage_tokens.py would reduce redundancy compared to grafana/grafana_utils.py and grafana/grafana_tokens.py etc, particularly when importing, i.e. from grafana import utils versus from grafana import grafana_utils. In the utils-specific case we'd need to avoid naming clashes with the top-level utils though. Which indicates to me either (i) we are overusing utils and should be more creative in our filenaming, or (ii) I am overthinking package design :D

Very good point!! I am not proud at all about the naming of this files and to be honest, I have always been bad at naming! Do you mind if I open an issue about improving the design after merging this PR? I just notice the same naming pattern for the `deployer/generate` directory too (also my doing 🤦🏼‍♀️).

sgibson91:
@GeorgianaElena a follow up issue is perfect, thank you! And I agree, naming things is hard :)

Issue #1918, opened 2022-11-15
[EVENT] Callysto


damianavila:
### Summary

We have a series of workshops where various groups of users will be logging and and trying out the service.
Ref: https://github.com/2i2c-org/infrastructure/issues/1678#issuecomment-1302631346

- **Community Representative:** @ianabc @byrcyb
- **Event begin:** (dates already updated)
  - Nov 24th ~ 50-60 users
  - Nov 25th ~ 180-190 users
  - December 5th ~200 students
  - December 6th ~ 120 users
  - December 9th ~ 30 users (new)
  - December 16th (to be confirmed) ~30 students
- **Event end:**
- **Active times:** Starting at 9 am (timezone?)
- **Number of attendees:** See above on the dates section
- [**Hub Events Calendar**](https://calendar.google.com/calendar/u/2?cid=Y19rdDg0c2g3YW5tMHNsb2NqczJzdTNqdnNvY0Bncm91cC5jYWxlbmRhci5nb29nbGUuY29t) (already added after dates confirmation)

### Hub info

- **Hub URL**: 2i2c.callysto.ca
- **Hub decommisioned after event?**: NO

### Task List

**Before the event**

- [x] Dates confirmed with the community representative and added to Hub Events Calendar.
- [x] Quotas from the cloud provider are high-enough to handle expected usage.
- [x] **One week before event** Hub is running.
- [x] Confirm with Community Representative that their workflows function as expected.
  - <details>
    <summary>👉Template message to send to community representative</summary>
    
    ```
    Hey {{ COMMUNITY REPRESENTATIVE }}, the date of your event is getting close!
    
    Could you please confirm that your hub environment is ready-to-go, and matches your hub's infrastructure setup, by ensuring the following things:
    - [ ] Confirm that the "Event Info" above is correct
    - [ ] On your hub: log-in and authentication works as-expected
    - [ ] `nbgitpuller` links you intend to use resolve properly
    - [ ] Your notebooks and content run as-expected
    ```
    
    </details>  
- [ ] **1 day before event**, either a separate nodegroup is provisioned for the event or the cluster is scaled up.

**During and after event**

- [x] Confirm event is finished.
- [ ] Nodegroup created for the hub is decommissioned / cluster is scaled down.
- [ ] Hub decommissioned (if needed).
- [x] Debrief with community representative.
  - <details>
    <summary>👉Template debrief to send to community representative</summary>
      
     ```
[...]

PR #1943, opened 2022-11-22
Enable prepuller on callysto hub

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/1918#issuecomment-1317855777

### DIFF
--- a/config/clusters/callysto/common.values.yaml
+++ b/config/clusters/callysto/common.values.yaml
@@ -9,6 +9,14 @@ nfs:
     # Name of Google Filestore share
     baseShareName: /homes/
 jupyterhub:
+  # pre-puller is necessary as the image is pretty big, and
+  # pulling during first user spawn might cause timeouts and poor user
+  # experience. Also helps with node pre-warming.
+  prePuller:
+    continuous:
+      enabled: true
+    hook:
+      enabled: true
   custom:
     2i2c:
       add_staff_user_ids_to_admin_users: false
--- a/config/clusters/utoronto/common.values.yaml
+++ b/config/clusters/utoronto/common.values.yaml
@@ -14,9 +14,8 @@ jupyterhub:
     userScheduler:
       enabled: true
   # pre-puller is necessary as the image is pretty big, and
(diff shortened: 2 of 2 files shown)


GeorgianaElena (approved):

config/clusters/callysto/common.values.yaml
```diff
@@ -9,6 +9,14 @@ nfs:
...
     baseShareName: /homes/
 jupyterhub:
+  # pre-puller is necessary as the image is pretty big, and
+  # pulling during first user spawn might cause timeouts and poor user
+  # experience. Also helps with node pre-warming.
+  prePuller:
+    continuous:
+      enabled: true
+    hook:
+      enabled: true
```
consideRatio, line 19: Pre-puller is a helm chart thing that is decoupled from the configurator, due to that, only whats configured in the helm chaft actually helps here.

If there is a mismatch, this is slowing things down instead, because nodes will pull both this helm chart managed image, and then on demand the actual image.

I think a note to clarify this is a judgement call concluded to be positive in this case, because some motivation.

ianabc: Thank you, I think I understand. To update the image, I update `singleuser.image.name` and `singleuser.image.tag` here, _**and**_ the corresponding values in the configurator. That way, when a new node is added, the image is `prePulled` before any users arrive. When they do, they'll be given whatever image is defined in the configurator, so we want to make sure this is the same image which was prePulled, correct?

Right now,  when I go to the configurator I get a 403, but I can sort that out with support as long as I have the right idea above. If that is the case, please go ahead and merge this.

sgibson91: No need to change the configurator, just update the image in this config _and leave the configurator blank_. It will default to pulling the image here.

yuvipanda:
Given the event is so near, I've just merged this for now. I'll address your comments later, @consideRatio! Will not be forgotten! I did already leave a comment descrigint that here: https://github.com/2i2c-org/infrastructure/issues/1918#issuecomment-1323165627

PR #1950, opened 2022-11-23
Update image and adjust admin list

ianabc:
Some of our notebooks use the mobilechelonian module which seems to have broken with the release of ipywidgets 8. I've pinned ipywidgets at 7.x in our image while we investigate and this change uses that image for our hub. At the same time I've adjusted the list of admin users to match the Callysto admin team from Cybera and PIMS.

Signed-off-by: Ian Allison <iana@pims.math.ca>

### DIFF
--- a/config/clusters/callysto/common.values.yaml
+++ b/config/clusters/callysto/common.values.yaml
@@ -39,7 +39,7 @@ jupyterhub:
   singleuser:
     image:
       name: callysto/2i2c
-      tag: 0.1.0
+      tag: 0.1.1
     extraFiles:
       tree.html:
         mountPath: /usr/local/share/jupyter/custom_template/tree.html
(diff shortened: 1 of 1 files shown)


yuvipanda (approved):

ianabc:
Sure! But what do I need to do? I don't see a merge action here

yuvipanda:
@ianabc try now?

yuvipanda:
@ianabc yay! generally gotta watch the github action through to make sure it succeeds...

ianabc:
Ah...I think I missed something, it's failing when it tries to run the health check on the staging hub, it complains about [not finding gke-gcloud-auth-plugin](https://github.com/2i2c-org/infrastructure/actions/runs/3535379720/jobs/5933595162#step:6:43)

yuvipanda:
@ianabc can you try logging into the staging hub to see if it works? The gke-gcloud-auth-plugin is a warninng unrelated I think

ianabc:
Yep, will do. I think it might be just a tag issue with my image. I'll re-run the task then take a look on the staging hub

Issue #1905, opened 2022-11-11
[EVENT] UToronto Exam on JupyterHub


yuvipanda:
### Summary

https://2i2c.freshdesk.com/a/tickets/220 has more information.

UToronto wants to try running a pilot, where they use the JupyterHub to run an exam for students. This needs a highly available JupyterHub, with a support engineer available during the specific time of the exams (I think they're about 3h).

### Event Info

- **Community Representative:** 
- **Event time:** Wednesday December 14, 9-12pm noon ET
- **Number of attendees:** 90
- [**Hub Events Calendar**](https://calendar.google.com/calendar/u/2?cid=Y19rdDg0c2g3YW5tMHNsb2NqczJzdTNqdnNvY0Bncm91cC5jYWxlbmRhci5nb29nbGUuY29t)

### Hub info

- **Hub URL**:  jupyter.utoronto.ca
- **Hub decommisioned after event?**:  no

### Task List

**Before the event**

- [x] Dates confirmed with the community representative and added to Hub Events Calendar.
- [x] Quotas from the cloud provider are high-enough to handle expected usage.
- [x] NFS disk has enough free space for this event to go fine
- [x] Exam setters have tried out the exam on the JupyterHub, and confirm that all the code for solutions executes as required with the environment in the hub. The environment will be frozen temporarily after this.
- [x] Exam setters have figure out how much RAM (and CPU) the solution will take, and we set these as *guarantees* for the student user pods during the course of the exam. The config changes will be deployed 1 working day before the exam.
- [x] Pre-warm the cluster to make sure there is enough nodes for all the students coming in, so no student will have to wait for a node scale up. This should be done a few hours before the exam starts.
- [x] Make sure an engineer is assigned to watch support tickets during this time period.

**During and after event**

- [x] Confirm event is finished.
- [x] Nodegroup created for the hub is decommissioned / cluster is scaled down.
- [x] Debrief with community representative. ->  https://2i2c.freshdesk.com/a/tickets/220

yuvipanda:
Progress being tracked in https://2i2c.freshdesk.com/a/tickets/220

yuvipanda:
Copying relevant message from freshdesk ticket:

We currently guarantee users 1G of RAM, and limit then to 2G of RAM. We do not set any CPU limits or guarantees, but to ensure that during an exam everyone has access to equal resources, I suggest we set these for the duration of the exam.

[...]

PR #1962, opened 2022-11-29
[Merge on Dec 14, 2022] Give all students equal resources during exam

yuvipanda:
During an exam, it is important that all students have access to the same computational resources. 2G should be enough, given it was the limit for everyone in general. We are *adding* a CPU limit here, where we did not have one before - so this must be validated beforehand to make sure it doesn't have adverse effects.


yuvipanda:
Do not merge until Dec 14, 2022

sgibson91 (approved):

PR #1964, opened 2022-11-29
Enable named servers in cryocloud

yuvipanda:
Ref https://github.com/CryoInTheCloud/hub-image/issues/7


sgibson91 (approved):
LGTM to me with one question, do we want to/should we limit the number of named servers a user can create to avoid one user taking up all the resources?

yuvipanda:
@sgibson91 i think it is ok to let it be until it becomes a problem and we can limit it if it does!

PR #1976, opened 2022-12-05
Update new hub issue template to ask specifics when setting up GitHub auth

### LINKED ISSUE #1806: Add hub engineer to GitHub org as admin to facilitate hub setup process

sgibson91:
Just selecting "I want to login with GitHub" is not sufficient information for the engineer now since we support multiple features that rely on GitHub auth, such as:

- Org-based auth,
- Teams-based auth
- Profile options restricted by Teams membership

This PR attempts to update the template to also collect this information.

It also adds an extra step that the engineer should be given Owner rights on the target Org in order to setup authentication properly.

fixes #1806 

See rendered template [here](https://github.com/sgibson91/infrastructure/blob/update-auth-info-template/.github/ISSUE_TEMPLATE/2_new-hub-provide-info.yml)

### DIFF
--- a/.github/ISSUE_TEMPLATE/2_new-hub-provide-info.yml
+++ b/.github/ISSUE_TEMPLATE/2_new-hub-provide-info.yml
@@ -62,7 +62,7 @@ body:
       required: true
 
   - type: dropdown
-    attributes: 
+    attributes:
       label: "[GitHub Auth only] How would you like to manage your users?"
       description: |
         Please describe how you would prefer to manage your users accessing the hub via GitHub Auth.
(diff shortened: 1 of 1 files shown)


damianavila (approved):
This is great addition, @sgibson91.

Issue #1975, opened 2022-12-05
What happened to the static site demo on our staging.2i2c.cloud hub?


sgibson91:
### Context

We have a support question about distributing content that I think would be an ideal case for the static site hosting setup we have. I have the link to our docs, but I wanted to provide a demo of it working.

1. I believed we have this feature set up on the https://staging.2i2c.cloud hub by config inspection
   https://github.com/2i2c-org/infrastructure/blob/master/config/clusters/2i2c/staging.values.yaml
2. From that config inspection, the repo should be https://github.com/yuvipanda/test-repo-push and it should be hosted on the path `textbook` relative to the hub
3. The `/textbook` page only shows the message "A private repo hosted at https://github.com/yuvipanda/test-repo-push" and it seems the actual repo is a 404

Are these problems because it is a private repo? I notice from the config that `staticWebsite.githubAuth.enabled = true` but the auth for the actual hub is Google, so maybe that's not compatible?

I generally think it would be nice to have a public demo of this feature, I feel like we used to have one that linked to the nbgitpuller docs?

GeorgianaElena:
> The /textbook page only shows the message "A private repo hosted at https://github.com/yuvipanda/test-repo-push" and it seems the actual repo is a 404

@sgibson91, I believe that is the actual content of the repo and accessing it directly returns a 404 because it's private.

Before putting the static content under hub auth, it used to host the 2i2c docs https://github.com/2i2c-org/docs. So maybe you can change it back to that for demo purposes?

GeorgianaElena:
Reopening because of https://github.com/2i2c-org/infrastructure/pull/1983

PR #1977, opened 2022-12-05
Use public repo for private auth demo

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/1975


yuvipanda:
Thanks to @GeorgianaElena for diagnosing this!

damianavila (approved):

Issue #1972, opened 2022-12-02
Our "Table of all running hubs" is missing


jmunroe:
### Context

Our "Table of all running hubs" at https://infrastructure.2i2c.org/en/latest/reference/hubs.html appears to be missing today.

### Proposal

Suggestion from @choldgraf 

> 
> I wonder if we should deploy that in a more persistent location, similar to our kpis page (and cross-link it from the infrastructure docs)
> e.g. [2i2c.org/list-of-hubs](http://2i2c.org/list-of-hubs)

sgibson91:
I don't think the location affects it's missing-ness.

1. This python script is what collects the info about the hubs: https://github.com/2i2c-org/infrastructure/blob/master/docs/scripts/render_hubs.py
2. It is executed during the sphinx build https://github.com/2i2c-org/infrastructure/blob/c4d6ad2a180999e0e141885df4469b58fa527d69/docs/conf.py#L80
3. The resultant CSV files are then embedded into the docs https://github.com/2i2c-org/infrastructure/blob/master/docs/reference/hubs.md#table-of-all-running-hubs

If it is missing then something went wrong either in the Python execution (generating the csv file) or the sphinx build (embedding the csv into the rendered html). I would check the ReadTheDocs build logs to see if there's anything useful there.

damianavila:
> I don't think the location affects it's missing-ness.

Well... actually the location might be triggering unnecessary updates (and be exposed to transient issues like this one), maybe?

sgibson91:
> Well... actually the location might be triggering unnecessary updates (and be exposed to transient issues like this one), maybe?

We're not that fine-grained. ReadTheDocs builds on every push/PR, whether it's to the docs folder or not (though they recently provided support to override this behaviour). So long as this is a page _somewhere_ in the docs, it'll get built frequently. We also don't want it too far away from our infrastructure since it relies on reading cluster.yaml files etc to compile the table, so we do want it to update as we create new hubs and such.

We also don't know what caused this yet. So let's not speculate too much.

choldgraf:
Sorry that I'm confusing things - I wasn't suggesting that the location of this page is *causing* this problem, but I was thinking that "a list of our running hubs" might benefit from an easier-to-remember location that was also more prominent in our documentation, since we (at least I) refer to it often and it feels weird to dig into the infrastructure for the page. But we can have that discussion elsewhere and focus this issue on fixing the problem.

[...]

PR #1981, opened 2022-12-06
Remove the templates directory from the list of hubs

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/1972

Running `make clean` and then `make html` revealed the following error:

Screenshot 2022-12-06 at 11 03 10

This probably originated when the `templates` directory was added.


GeorgianaElena:
This should be straightforward enough to merge without approval,so I'm going to go ahead and do it to fix the docs

PR #1989, opened 2022-12-12
Set utoronto staging hub limits to match exam limits

yuvipanda:
Helps the instructor test that resource limits are ok


yuvipanda:
I've deployed this so it can be tested

GeorgianaElena (approved):
Thank you @yuvipanda :sparkles:

PR #1992, opened 2022-12-12
Add deployment to measure free space in hub home dir

yuvipanda:
This adds a deployment with a single pod of
[node_exporter](https://github.com/prometheus/node_exporter) running, configured only to report filesystem statistics. We mount the same home directory configuration that is used in the hub to this pod, so it will provide us with information on total disk size and used disk. This would allow us to have a Grafana panel with 'disk used %' for each of these hubs! We can even alert on those later.

### DIFF
--- a/helm-charts/basehub/templates/home-space-reporter.yaml
+++ b/helm-charts/basehub/templates/home-space-reporter.yaml
@@ -1,0 +1,58 @@
+# Deploy a prometheus node_exporter with the same home directory
+# we have for our hub mounted so we can monitor free space usage.
+{{- if or .Values.nfs.enabled .Values.azureFile.enabled }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: jupyterhub
+    component: home-metrics
+  name: home-metrics
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: jupyterhub
+      component: home-metrics
+  template:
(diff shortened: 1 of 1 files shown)


yuvipanda:
https://github.com/jupyterhub/grafana-dashboards/pull/51 adds the graph.

helm-charts/basehub/templates/home-space-reporter.yaml
```diff
@@ -0,0 +1,58 @@
...
+        volumeMounts:
+          - name: home
+            mountPath: /home
+            # Mount it readonly to prevent accidental writes
+            readOnly: true
+      securityContext:
+        fsGroup: 65534
+        runAsGroup: 65534
+        runAsNonRoot: true
+        runAsUser: 65534
```
GeorgianaElena, line 49: Maybe add a comment about where the 65534 uid comes from?

```suggestion
      # More about the `65534` user (also called `nobody`) at https://wiki.ubuntu.com/nobody
      securityContext:
        fsGroup: 65534
        runAsGroup: 65534
        runAsNonRoot: true
        runAsUser: 65534
```

GeorgianaElena (approved):

helm-charts/basehub/templates/home-space-reporter.yaml
yuvipanda: Done!

Issue #1961, opened 2022-11-29
[Request deployment] New Hub: UToronto R Hub


yuvipanda:
We will be deploying a second hub in the UToronto cluster, focused on R. This will mostly be a copy of the current hub, with a new image.

### Important dates

- **Target start date**: Dec 15, 2022
- **Required start date**: Jan 5, 2023

### Hub Authentication

This will reuse the same AzureAD authentication we use in the main hub. However, it was decided we will no longer use opaque oids for usernames, but switch to readable usernames. Jeremy from UToronto indicated that this is not a privacy concern any longer, so we're good.

- [ ] Request another set of credentials from UToronto IT for AzureAD
- [ ] Switch username_claim to the  email of the user

### Front Page

Replicate same settings as current UToronto hub.

- [ ] Default should switch to RStudio

### Hub user image GitHub repository

New R focused image will need to be created.

- [ ] Create a new image template based on the Rocker image, to be used for R based hubs

### Domain this should live on

Currently, the JupyterHub lives on jupyter.utoronto.ca, which is a CNAME to utoronto.2i2c.cloud. I have proposed this new hub live in r.utoronto.ca - let's see how they respond.

yuvipanda:
I've emailed UToronto IT + decision makers in https://2i2c.freshdesk.com/a/tickets/308

yuvipanda:
Image set up at https://github.com/2i2c-org/utoronto-r-image

sgibson91:
We should make sure there is an up-to-date version of R, ref: https://2i2c.freshdesk.com/a/tickets/317

yuvipanda:
I've responded there, and yep, it's using latest version of R (4.2.2). Updates will also be far easier here.

yuvipanda:
With #1995, a staging and prod R hub are now setup!

We need to figure out how to change the default interface selector on the R hub to point to RStudio, rather than to classic notebook though.

yuvipanda:
They have internal discussions going on about the domain name to use. Let's wait on that, but should be just a CNAME change when done.

yuvipanda:
This is now complete! Available at r.datatools.utoronto.ca! \o/

PR #1995, opened 2022-12-13
Setup new R hub for UToronto

yuvipanda:
Adds a README too with some more info on how this specific cluster of hubs is configured.

Ref https://github.com/2i2c-org/infrastructure/issues/1961


GeorgianaElena (approved):
yuhuuu 🎉 🎉 🎉

PR #1997, opened 2022-12-13
Add instructions about how to scale up an Azure cluster

GeorgianaElena:


docs/sre-guide/node-scale-up/azure.md
```diff
@@ -0,0 +1,46 @@
...
+    ```{figure}
+    ../../images/azure-scale-node-pool-window.png
+    ```
+1. If the `Autoscale` option is selected like in the screenshot above (the default, recommended option),
+   then in order to scale up the nodepool to an exact number of nodes, temporarily deactivate the autoscaler,
+   by selecting the `Manual` option, introduce the desired number of nodes then click `Apply`.
+
+1. After the Apply succeded, you should see the new nodes coming up.
+   You can then click on `Scale node pool` option again, enable the `Autoscale`,
+   and set the min number of nodes to the desired one the you set in the step before.
```
GeorgianaElena, line 31: @yuvipanda, is this the correct thing to do? I figured we want the autoscaler to be running in case cluster needs scale up based on usage.

The workflow I image will happen is: we have `x` number of nodes manually created. We enable the autoscaler, then after 10 mins, a scale down event is triggered (because let's say not all users accessed the hub), the autoscaler will check the limits and because the min number of nodes is `x`, it won't scale it down.

And this way, we benefit from automatic scale ups, but always ensure a min number of nodes.

Does this make any sense?

docs/sre-guide/node-scale-up/azure.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Azure Console UI instructions
+
+1. Login to https://portal.azure.com using the appropriate cluster credentials
+
+1. Search the `Kubernetes service` section for the appropriate cluster
+
+1. Click on the cluster name, then go to `Settings` and select the `Node pools` option
+
+1. Click on the appropriate nodepool that you would like to scale up
```
sgibson91, line 16:
```suggestion
1. Click on the appropriate node pool that you would like to scale up
```

For consistency with the rest of the document :)

docs/sre-guide/node-scale-up/azure.md
```diff
@@ -0,0 +1,47 @@
...
+1. Then select the `Scale node pool` option from the top of the page
+
+1. A new window should pop up that looks like this
+
+    ```{figure} ../../images/azure-scale-node-pool-window.png
+    Scale nodepool
+    ```
+
+1. If the `Autoscale` option is selected like in the screenshot above (the default, recommended option),
+   then in order to scale up the nodepool to an exact number of nodes, temporarily deactivate the autoscaler,
```
sgibson91, line 27:
```suggestion
   then in order to scale up the node pool to an exact number of nodes, temporarily deactivate the autoscaler,
```

Again, maintaining consistency

docs/sre-guide/node-scale-up/azure.md
```diff
@@ -0,0 +1,47 @@
...
+   by selecting the `Manual` option, introduce the desired number of nodes then click `Apply`.
+
+1. After the Apply succeded, you should see the new nodes coming up.
+   You can then click on `Scale node pool` option again, enable the `Autoscale`,
+   and set the min number of nodes to the desired one the you set in the step before.
+
+```{warning}
+The cluster autoscaler doesn't enforce the nodepool size after updating the `Min` or `Max` counts.
```
sgibson91, line 35:
```suggestion
The cluster autoscaler doesn't enforce the node pool size after updating the `Min` or `Max` counts.
```

docs/sre-guide/node-scale-up/azure.md
sgibson91: If it helps, this feels consistent with what I used to do when managing node pools on Azure and your reasoning makes sense, especially since the exam won't be the _only_ set of users active on the hub during the event time window.

sgibson91 (approved):
Thank you for documenting this!

docs/sre-guide/node-scale-up/azure.md
yuvipanda: @GeorgianaElena This is *exactly* the right thing to do, for exactly the reasons you specify :) 

Can you add a warning here to make sure we do **not** forget to turn autoscale back on? That would be only noticed when we would need to scale up from the max manual limit, and we'll get stuck there.

For scaling down, I think we should just reduce the autoscaler minimum, and not touch manual controls.

Thank you so much for working on this! I think this can be merged once the warning is added, and we document how to scale down

PR #1998, opened 2022-12-13
Document policy for self-merging as a partner

yuvipanda:
Starts working on the specifics of our shared responsibility model, as we currently have two orgs (Cloudbank & Callysto) who have engnineers with merge rights in our repo.

### DIFF
--- a/docs/contributing/code-review.md
+++ b/docs/contributing/code-review.md
@@ -50,7 +50,18 @@ merge rights to partner engineers. Ideally, they would be only allowed to
 merge PRs related to their own cluster - we currently do not have an automatic
 way of enforcing this. 
 
-They *may* self-merge if they want to, provided the following conditions are met:
+```{note}
+While our current two partners who have such access (Cloudbank & Callysto)
+are grandparented in, we must have such technical enforcement mechanism
+in place before adding more such partners with merge rights
+```
+
+The goal here is that 2i2c is ultimately responsible for the configuration, so
+we do not want it to get too funky without 2i2c engineers knowing
+about it. As a community partner, if you think what you are doing
+*might* be funky, err on the side of asking for review!
+
+Community partners *may* self-merge if they want to, provided the following conditions are met:
 
 1. They have cloud access to the cluster, and are confident of being able to either
    debug the issues that may arise from the *particular* change they are making,
@@ -67,13 +78,11 @@ They *may* self-merge if they want to, provided the following conditions are met
    Here are some examples of *novel* configuration that requires approval from a 2i2c engineer before merging:
    - Adding python code to `hub.extraConfig` to enable new functionality, such as
      adding a postgres database to each user pod.
-   - Significant alterations to the shape of the user pod, such as setting
+   - Significant alterations to the configuration of the user pod, such as setting
      `singleuser.extraContainers`.
    - Modifications to how NFS home directory storage is managed.
    
-   The goal here is that 2i2c is ultimately responsible for the configuration, so
-   we do not want it to get too funky without us knowing about it. If you think
-   what you are doing *might* be funky, err on the side of asking for review!
+As a general rule, when in doubt, ask for review :)
 
 ## Changing authentication related configuration
 
--- a/docs/contributing/code-review.md
+++ b/docs/contributing/code-review.md
@@ -60,7 +60,7 @@ However, merge rights are a big responsibility, so please be careful in your act
 
 Community partners *may* self-merge if they want to, provided the following conditions are met:
 
-1. **They are confident debugging any issues that arise from the issue**.
+1. **They are confident debugging any issues that arise from the self-merged PR**.
    If any issues arise from your self-merge, **you are responsible for resolving them, or reverting the change**.
    You should understand the potential repurcussions of your change, and be ready to fix things if they break.
 2. **They have access to their cloud cluster to debug changes**.
--- a/docs/contributing/code-review.md
+++ b/docs/contributing/code-review.md
@@ -42,6 +42,38 @@ Here are some guidelines for merging and reviewing in this case:
   make sure you push a PR with the change *and merge it quickly* to make sure that the changes
   are persisted across future deploys. Self merging is acceptable here, although this general
   class of changes should be limited as much as possible.
+  
+## Self-merging as a partner
+
+In some cases, as part of our shared responsibility model, we have granted
+merge rights to partner engineers. Ideally, they would be only allowed to
+merge PRs related to their own cluster - we currently do not have an automatic
+way of enforcing this. 
+
+They *may* self-merge if they want to, provided the following conditions are met:
+
+1. They have cloud access to the cluster, and are confident of being able to either
+   debug the issues that may arise from the *particular* change they are making,
+   or be able to revert it.
+2. The change *only* touches files inside the `config/clusters/<cluster-name>`
+   directory for their cluster.
+3. The change is fairly routine, and not a super novel configuration. This is hard
+   to quantify, but here are some examples of *routine* changes:
+   - Adding a new hub that looks exactly like other hubs in the cluster
+   - Changing resources provided to the hub
+   - Adding / removing admin users
+   - Changing profile options available to the hub
+   
+   Here are some examples of *novel* configuration:
+   - Adding python code to `hub.extraConfig` to enable new functionality, such as
+     adding a postgres database to each user pod.
+   - Significant alterations to the shape of the user pod, such as setting
+     `singleuser.extraContainers`.
+   - Modifications to how NFS home directory storage is managed.
+   
+   The goal here is that 2i2c is ultimately responsible for the configuration, so
+   we do not want it to get too funky without us knowing about it. If you think
+   what you are doing *might* be funky, err on the side of asking for review!
 
 ## Changing authentication related configuration
 
--- a/docs/contributing/code-review.md
+++ b/docs/contributing/code-review.md
@@ -56,7 +56,7 @@ Merge rights are only given for partners with which we have built a relationship
 ### Guidelines for community partners
 
 Our goal is to provide trusted communities the ability to more quickly make changes to their infrastructure in order to lead to a better, more collaborative service for their community.
-However, merge rights it a big responsibility, so please be careful in your actions.
+However, merge rights are a big responsibility, so please be careful in your actions.
 
 Community partners *may* self-merge if they want to, provided the following conditions are met:
 
--- a/docs/contributing/code-review.md
+++ b/docs/contributing/code-review.md
@@ -43,9 +43,9 @@ Here are some guidelines for merging and reviewing in this case:
   are persisted across future deploys. Self merging is acceptable here, although this general
   class of changes should be limited as much as possible.
   
-## Self-merging as a partner
+## Self-merging as a community partner
 
-In some cases, as part of our shared responsibility model, we have granted
+In some cases, as part of our [shared responsibility model](https://docs.2i2c.org/en/latest/about/service/shared-responsibility.html), we grant
 merge rights to partner engineers. Ideally, they would be only allowed to
 merge PRs related to their own cluster - we currently do not have an automatic
 way of enforcing this. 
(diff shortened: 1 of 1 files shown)


docs/contributing/code-review.md
```diff
@@ -42,6 +42,38 @@ Here are some guidelines for merging and reviewing in this case:
   make sure you push a PR with the change *and merge it quickly* to make sure that the changes
   are persisted across future deploys. Self merging is acceptable here, although this general
   class of changes should be limited as much as possible.
+  
+## Self-merging as a partner
```
choldgraf, line 46:
```suggestion
## Self-merging as a community partner
```

docs/contributing/code-review.md
```diff
@@ -42,6 +42,38 @@ Here are some guidelines for merging and reviewing in this case:
...
+
+In some cases, as part of our shared responsibility model, we have granted
```
choldgraf, line 48:
```suggestion
In some cases, as part of our [shared responsibility model](https://docs.2i2c.org/en/latest/about/service/shared-responsibility.html), we grant
```

docs/contributing/code-review.md
```diff
@@ -42,6 +42,38 @@ Here are some guidelines for merging and reviewing in this case:
...
+2. The change *only* touches files inside the `config/clusters/<cluster-name>`
+   directory for their cluster.
+3. The change is fairly routine, and not a super novel configuration. This is hard
+   to quantify, but here are some examples of *routine* changes:
+   - Adding a new hub that looks exactly like other hubs in the cluster
+   - Changing resources provided to the hub
+   - Adding / removing admin users
+   - Changing profile options available to the hub
+   
+   Here are some examples of *novel* configuration:
```
choldgraf, line 67:
```suggestion
   Here are some examples of *novel* configuration that requires approval from a 2i2c engineer before merging:
```

docs/contributing/code-review.md
```diff
@@ -42,6 +42,38 @@ Here are some guidelines for merging and reviewing in this case:
...
+   - Adding python code to `hub.extraConfig` to enable new functionality, such as
+     adding a postgres database to each user pod.
+   - Significant alterations to the shape of the user pod, such as setting
```
choldgraf, line 70: Is "shape" a well-known or standard term? I am not sure what it means but if it's common then no big deal

docs/contributing/code-review.md
```diff
@@ -42,6 +42,38 @@ Here are some guidelines for merging and reviewing in this case:
...
+     `singleuser.extraContainers`.
+   - Modifications to how NFS home directory storage is managed.
+   
+   The goal here is that 2i2c is ultimately responsible for the configuration, so
```
choldgraf, line 74: What do you think about beginning with the goal / principle, and then describe the policy just after (so move this to the top). That way the rationale comes first, and we also force ourselves to keep it relatively short so it doesn't crowd out the policy :-)

choldgraf (commented):
This generally seems good to me - a few quick thoughts in the comments.

I think there are two major questions I have:

1. Is our team OK with this? We are potentially introducing mechanisms to generate unexpected work for our team.
2. Are our communities OK with this? I am a little bit worried that some of the community representatives will not accept the idea that a person they don't know has the ability to self-merge a PR that will auto-deploy a change to their hub. How can we minimize this worry?

yuvipanda:
The goal is for (2) to never happen - is folks can only self-merge for cases when it is affecting *just* their own cluster, and nobody else.

choldgraf:
> The goal is for (2) to never happen - is folks can only self-merge for cases when it is affecting just their own cluster, and nobody else.

Right but there's nothing technically preventing this from happening, right? So if that's not the case, we need to be able to make a case that the people with this power have a **lot** of trust and vetting from us.

yuvipanda:
I definitely want broad approval from the rest of the team for this to be merged - (1) is super important! I agree (2) is also important, but I hope it does not need to block this PR.

choldgraf:
I'd really like input from @colliand on number 2 - particularly whether he thinks this might violate some of the expectations that we have set with communities in contracts etc.

If not, then I think we can move forward but that we need to be really careful. We should also make sure there's nothing catastrophic that a person could do (e.g., a bad-faith act to abuse the cloud infrastructure of each community in some really bad way like a ddos or exposing user data?).

yuvipanda:
So, we should / must have long term ways for people to self-merge changes to their own clusters without our intervention. Unfortunately, right now it also gives them the ability to merge changes to *other* people's communities. We rely on the honor system right now. The question is not *if* but *when* we want to also have technical means of preventing issues here.

GitHub's [CODEOWNERS](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) doesn't help us as it only automatically adds reviewers.

Taking a step back, conceptually this repository is a *monorepo*, and what we are looking for is granular per-directory permissions for subsets of people. This is implementable as a GitHub action / bot. We can maintain a mapping of teams to directories, and have the bot auto-merge when a command is given by someone in a team when *only* files in that directory have changed. I think the kubernetes repo uses something like this, for example.

So the question really is - do we want to stop this practice now until we get this automated bot solution working? Or rely on the existing honor system, and block any new *additions* until we have this working?

yuvipanda:
You can see this in action here: https://github.com/kubernetes/kubernetes/pull/114489#pullrequestreview-1219294445

A bot recognizes that the person who typed `/lgtm` has the rights to do so, and merges it.

sgibson91:
> My current suggestion is to say the current set of people (which is cloudbank and callysto) can exist on the honor system, but we must implement this before we add anyone else.

Agree

choldgraf:
IMO this is reasonable - can you sync w @colliand and / or @jmunroe on this just to make sure it wouldn't violate the trust of any expectations we have already set?

yuvipanda:
@choldgraf I've incorporated your suggestions wrt wording here now, and explicitly listed that we can't add more community partners with merge rights until we have a technical enforcement mechanism in place

choldgraf:
I took a pass and pushed a commit to the policy wording - how does that look?

In addition I've made one change:

- I've made the `@collaborators` team have write access to this repository. I suggest we remove the `callysto` and `dsep` teams from this repository, and provide access solely via `@collaborators`. That way we can have an easy-to-reference source of truth for "who outside of 2i2c has write access to this repository". If we want a dedicated team just for the `infrastructure` repository access, we could create a specific sub-team of `@collaborators`, but it didn't seem necessary to me for now.

WDYT?

docs/contributing/code-review.md
```diff
@@ -45,37 +45,42 @@ Here are some guidelines for merging and reviewing in this case:
...
+[See the `2i2c-org/collaborators` team](https://github.com/orgs/2i2c-org/teams/collaborators) for a list of sub-teams and individuals with write access to this repository.
 
-The goal here is that 2i2c is ultimately responsible for the configuration, so
-we do not want it to get too funky without 2i2c engineers knowing
-about it. As a community partner, if you think what you are doing
-*might* be funky, err on the side of asking for review!
+### Guidelines for community partners
+
+Our goal is to provide trusted communities the ability to more quickly make changes to their infrastructure in order to lead to a better, more collaborative service for their community.
+However, merge rights it a big responsibility, so please be careful in your actions.
```
choldgraf, line 59: `is` a big responsibility 🤦 - but for some reason github isn't letting me suggest an edit 😕

yuvipanda: @choldgraf fixeed

yuvipanda:
@choldgraf i think when we have the technical implementation in place, we definitely need to have per cluster teams. So I'd suggest not removing those, and reverting the catch-call collaborators team. Think you'd be able to do that?

Otherwise, as usual, your re-wording makes everything clearear and cooler so ty :)

choldgraf:
My plan was to keep the teams for individual communities, but just nest them under the "collaborators" team so that they inherit their write privileges via that parent team. But I definitely don't have strong opinions about this, it just seemed a bit cleaner.

sean-morris:
I am jumping in way late folks as I navigated my way through the end of the semester and lost track of this thread. I appreciate all the thought and want to make sure I am in line with the team's needs and wishes. I definitely have been self-merging -- under the impression that this was expected when I was making changes on the cloudbank cluster. 

It sounds like this is still OK to do and, of course, I will continue to be vigilant -- very aware that I could inadvertently break something else if I am not careful with what files are changed. 

This said everything is going quite well on the cluster. We are bringing in three more schools in the Spring with probably at least another three lining up for the Fall.

I hope your Holidays are going well and you are all healthy.

-----
I had written this to @yuvipanda and @damianavila on the Slack thread when I first wondered a few weeks ago if I should actually be self-merging.

Yes, Yuvi, the pieces you line out in regards to self-merge are exactly why I just did it. I somehow thought you had told me that as long as a change only hit the cloudbank cluster it was fine. I would say I am generally nervous to mess with anything I don't fully understand.

What mostly happens is we get the hub for the CC up using personal emails while we figure out if CiLogon is going to work. Once I figure that out, I then set up the hub correctly. Normally two PRs, which is annoying.
There are two pieces going on for me: 1) The changes are very simple and cookie cutter so I am not worried about screwing up something I can't fix; 2) Sometimes, I need it up quickly so I am skipping the review which is suspect really but happens when I have the client on zoom or a meeting is about to happen and somehow I lost track of me needing to update the hub. Moments like this.

yuvipanda:
@choldgraf I forgot that subteams exist! So am happy to leave the collaborators team in as is. I've re-added their access to this repo, and invited @sean-morris to the appropriate team.

I'll leave this up for a week for @colliand and @jmunroe to respond, and merge after.

docs/contributing/code-review.md
```diff
@@ -42,6 +42,58 @@ Here are some guidelines for merging and reviewing in this case:
...
+[See the `2i2c-org/collaborators` team](https://github.com/orgs/2i2c-org/teams/collaborators) for a list of sub-teams and individuals with write access to this repository.
+
+### Guidelines for community partners
+
+Our goal is to provide trusted communities the ability to more quickly make changes to their infrastructure in order to lead to a better, more collaborative service for their community.
+However, merge rights are a big responsibility, so please be careful in your actions.
+
+Community partners *may* self-merge if they want to, provided the following conditions are met:
+
+1. **They are confident debugging any issues that arise from the issue**.
```
damianavila, line 63:
```suggestion
1. **They are confident debugging any issues that arise from the self-merge PR**.
```

docs/contributing/code-review.md
```diff
@@ -42,6 +42,58 @@ Here are some guidelines for merging and reviewing in this case:
...
+   If any issues arise from your self-merge, **you are responsible for resolving them, or reverting the change**.
+   You should understand the potential repurcussions of your change, and be ready to fix things if they break.
+2. **They have access to their cloud cluster to debug changes**.
+   Sometimes an issue requires intervention directly in the cloud infrastructure.
+   Before self-merging, ensure that you are authorized and have login-access to this infrastructure.
+3. **The change *only* touches files inside the `config/clusters/<cluster-name>` directory for their cluster.**
```
damianavila, line 69: We should be able to have fine-grained access to their own files with [codeowners](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) in the future, right? For now, this advice is OK-ish.

damianavila (approved):
A minor request of change, otherwise, LGTM.

docs/contributing/code-review.md
sgibson91: No, all codeowners does is automatically requests reviews from those listed in the codeowners file https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#about-code-owners

damianavila: Yeah, my previous sentence was a wish to have fine-grained permission through codeowners in the future...

sgibson91: I don't know if that is on GitHub's roadmap, but I don't believe it is achievable right now. It is only about assigning reviewers. So you _can_ assign specific reviewers to PRs that affect a specific filepath. But you _cannot_ grant _write access_ to only a certain filepath within the repo.

yuvipanda:
ok, I'm going to merge this for now as codification of what exists :) We can amend later if there are objections from @jmunroe or @colliand! Thanks for all the comments, everyone.

PR #2003, opened 2022-12-14
Revert "[Merge on Dec 14, 2022] Give all students equal resources during exam"

sgibson91:
Reverts 2i2c-org/infrastructure#1962

The exam is now over


sgibson91:
Will self-merge since this is a reversion

PR #2004, opened 2022-12-14
Bumping image version to include new packages

ianabc:
Byron wants haversine installed. It has been added to this environment and relocked as https://github.com/callysto/2i2c-image/commit/1c8498b404f2e8d9863a53d89bb89e0c3d5b6d89

Signed-off-by: Ian Allison <iana@pims.math.ca>

### DIFF
--- a/config/clusters/callysto/common.values.yaml
+++ b/config/clusters/callysto/common.values.yaml
@@ -39,7 +39,7 @@ jupyterhub:
   singleuser:
     image:
       name: callysto/2i2c
-      tag: 0.1.1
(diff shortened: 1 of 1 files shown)


ianabc:
The staging image hasn't picked up the change. This doesn't need to be merged until I figure out why.

ianabc:
I think this might be related to the image prePuller, I guess it is pulling the production value for our cluster at the moment. If that is the case I was able to test the image manually on the staging cluster so I think this PR is ready to merge.

yuvipanda (approved):

Issue #1986, opened 2022-12-09
Openscapes hub image update


betolink:
### Context

At Openscapes we're using different Docker images to support R, Python and Matlab users, I thought there was a github action somewhere at 2i2c to detect when a new image has been pushed and opens a PR to update the [hub config](https://github.com/2i2c-org/infrastructure/blob/master/config/clusters/openscapes/common.values.yaml)

We have a new image for Python `openscapes/python:11de2c5` and we want to test it on staging (staging.openscapes.2i2c.cloud/) is it possible without deploying it to prod? 

Is there a way to streamline these updates to the hub config? is the bot/github action only configured for the Python image? Thanks!

sgibson91:
Hi @betolink You can use the configurator to test the image on staging without affecting prod. Docs are here: https://docs.2i2c.org/en/latest/admin/howto/configurator.html This is probably the quickest way for you to manage your image (it work on both staging and prod)

Using the configurator in this way does not update hub config, though it will always override what is listed in the hub config. Some folk find this confusing and we need to find some dev cycles to put into improving it I think 😕 

https://github.com/2i2c-org/infrastructure/blob/b6ad29f7ee3662cf821dd4b5be2877b4b9c6faf0/.github/workflows/bump-image-tags.yaml#L35-L39

This is how the action is configured to auto-bump your images, and they are looking at all three images. This action only runs _weekly on Monday_ though, it is not triggered by pushing a tag to a Docker repository.

I think eventually, we would like to auto-populate a list of tags from a image repo, but that work is still ongoing https://github.com/2i2c-org/infrastructure/issues/1253

sgibson91:
Also, these kinds of questions are really good to send to support@2i2c.org rather than opening an issue :)

betolink:
Thanks for the info @sgibson9! Before opening the issue I used the configurator and had no effect on the spawned instance =( 
![Screenshot from 2022-12-09 09-21-25](https://user-images.githubusercontent.com/717735/206734978-329c96de-f927-4fc0-b48a-a6b2f88b83c7.png) I can follow up on the email.

sgibson91:
Oh yes - sorry my bad. The configurator doesn't work for profile list options like you've got. Sorry for the false information on my part 🤦🏻‍♀️

PR #2011, opened 2022-12-20
Update images in openscapes staging hub

betolink:
A little context: https://github.com/2i2c-org/infrastructure/issues/1986 

We released new [R](https://hub.docker.com/r/openscapes/rocker/tags) and [Python](https://hub.docker.com/r/openscapes/python/tags) images for Openscapes and we would like to test them on staging, I separated the chart values to hopefully only affect staging and if everything goes well, we'd like to use them in prod.


yuvipanda:
@betolink I have updated the PR so staging only has appropriate overrides, and most stuff is still in common. If this looks good to you, am happy to merge :)

betolink:
Thanks Yuvi!! I think this is ready to be merged.

PR #2013, opened 2022-12-23
Add ML notebook choice to medium server at leap hub

jbusecke:
We would like to be able to select the ML specific images not just for the GPU specific server, but also for the medium server. This option is available for more members and will in particular be used for an upcoming workshop. I am not sure if this simple change does the trick or if anything elsewhere needs to be specified.

### DIFF
--- a/config/clusters/leap/common.values.yaml
+++ b/config/clusters/leap/common.values.yaml
@@ -83,6 +83,26 @@ basehub:
           allowed_teams:
             - leap-stc:leap-pangeo-users
             - 2i2c-org:tech-team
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thanks, @jbusecke! This lgtm! Note that these users on medium will still only have CPU access

jbusecke:
Thanks a lot for the quick turnaround! 

> Note that these users on medium will still only have CPU access

Yup that is intended for these rather simple educational usecases.

yuvipanda:
@jbusecke i appreciate the PR!

PR #2018, opened 2023-01-04
Bump to recent pangeo image version

jbusecke:
For an upcoming LEAP workshop we require a version bump of the images. I hope this is the right way to request that.

### DIFF
--- a/config/clusters/leap/common.values.yaml
+++ b/config/clusters/leap/common.values.yaml
@@ -55,7 +55,7 @@ basehub:
     singleuser:
       image:
         name: pangeo/pangeo-notebook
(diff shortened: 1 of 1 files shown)


sgibson91:
I think the problem is with the sops hook pre-commit is running. The code of the hook itself is using a YAML parser to try and read a JSON file that contains a hard tab character `\t` (which interestingly is the only case where valid JSON _is not_ also valid YAML). So the fix needs to happen upstream, not in this repo.

I opened https://github.com/yuvipanda/pre-commit-hook-ensure-sops/issues/9 to track.

sgibson91 (approved):

PR #2034, opened 2023-01-10
Setup domain redirects for UToronto R hub

yuvipanda:
We want to make sure that users who were given the old R hub's URL have a smooth experience.

Ref https://github.com/2i2c-org/infrastructure/issues/1961


GeorgianaElena (approved):

yuvipanda:
Thanks @GeorgianaElena

Issue #1807, opened 2022-10-21
Update/change certain action steps in our workflows that are running old versions of Node


sgibson91:
### Context

On a recent CI/CD deployment run, I got the following warning(s):

```
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: jitterbit/get-changed-files
```

I don't think `jitterbit/get-changed-files` is well maintained, so we may need to swap to the `dorny/paths-filter` action we use in our validation workflow which has similar functionality and _didn't_ trigger this warning.

sgibson91:
PR is here:

* https://github.com/2i2c-org/infrastructure/pull/1870

sgibson91:
Reopening because there are other actions that are triggering this warning. They will be more difficult to swap out though, and we may need to wait for upstream releases.

- [azure/setup-helm](https://github.com/Azure/setup-helm)
- [azure/setup-kubectl](https://github.com/Azure/setup-kubectl)
- [mdgreenwald/mozilla-sops-action](https://github.com/mdgreenwald/mozilla-sops-action)
- [nick-fields/retry](https://github.com/nick-fields/retry)

damianavila:
@sgibson91, I just assigned you because you already started to work on this one, we can change the assignment if/when someone else picks this up later down the road.

sgibson91:
Wouldn't it be better to leave it unassigned so that we know someone needs to think about it? I'm worried if we leave my assignment it will just default to being my responsibility to solve.

damianavila:
That's a good point... let me just adjust the priority so we can quickly rediscover it.

sgibson91:
<img width="653" alt="Screenshot 2023-01-11 at 10 03 12" src="https://user-images.githubusercontent.com/44771837/211778219-65b62977-4d39-4484-a732-84ca06c51d7a.png">

Two actions are still giving us this node warning. I checked the releases for both and they have both switched over to Node 16. I'm going to bump our versions of the actions to see if that removes the annotation.

damianavila:
We think it is resolved.

consideRatio:
I checked manually and found one missing piece, resolved fully by merging https://github.com/2i2c-org/.github/pull/13

PR #2037, opened 2023-01-11
Bump third party actions to latest versions

sgibson91:
Both of these actions are now running node 16, but our GitHub Actions log is still reporting the node 12 annotation. I'm hoping updating these tags will reset the cache and remove that annotation.

Actions that are bumped are: mdgreenwald/mozilla-sops-action, nick-fields/retry

ref: #1807

### DIFF
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -198,7 +198,7 @@ jobs:
       # Retry action: https://github.com/marketplace/actions/retry-step
       - name: Run health check for staging hub on cluster ${{ matrix.jobs.cluster_name }}
         if: matrix.jobs.upgrade_staging
-        uses: nick-fields/retry@v2.8.3
+        uses: nick-fields/retry@v2
         with:
           timeout_minutes: 10
           max_attempts: 2
@@ -218,7 +218,7 @@ jobs:
       # Retry action: https://github.com/marketplace/actions/retry-step
       - name: Run health check for dask-staging hub on cluster ${{ matrix.jobs.cluster_name }} if it exists
         if: matrix.jobs.upgrade_staging && matrix.jobs.cluster_name == '2i2c'
-        uses: nick-fields/retry@v2.8.3
+        uses: nick-fields/retry@v2
         with:
           timeout_minutes: 10
           max_attempts: 2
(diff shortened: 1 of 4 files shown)


consideRatio (approved):
- I saw that v1 of the sops action was outdated and opened https://github.com/mdgreenwald/mozilla-sops-action/issues/167.
- The nick-fields/retry action is providing an updated v2 that matches the latest v2.x.y tag, do you wish to stay pinned to v2.x.y rather than going for v2 for that action? I'm :+1: for pinning only to the major version for an action that seem to maintain things well.

.github/workflows/deploy-hubs.yaml
```diff
@@ -198,7 +198,7 @@ jobs:
       # Retry action: https://github.com/marketplace/actions/retry-step
       - name: Run health check for staging hub on cluster ${{ matrix.jobs.cluster_name }}
         if: matrix.jobs.upgrade_staging
-        uses: nick-fields/retry@v2.8.1
+        uses: nick-fields/retry@v2.8.3
```
consideRatio, line 201:
```suggestion
        uses: nick-fields/retry@v2
```

.github/workflows/deploy-hubs.yaml
```diff
@@ -218,7 +218,7 @@ jobs:
       # Retry action: https://github.com/marketplace/actions/retry-step
       - name: Run health check for dask-staging hub on cluster ${{ matrix.jobs.cluster_name }} if it exists
         if: matrix.jobs.upgrade_staging && matrix.jobs.cluster_name == '2i2c'
-        uses: nick-fields/retry@v2.8.1
+        uses: nick-fields/retry@v2.8.3
```
consideRatio, line 221:
```suggestion
        uses: nick-fields/retry@v2
```

.github/workflows/deploy-hubs.yaml
```diff
@@ -328,7 +328,7 @@ jobs:
 
       # Retry action: https://github.com/marketplace/actions/retry-step
       - name: Run health check against ${{ matrix.jobs.hub_name }} hub on cluster ${{ matrix.jobs.cluster_name}}
-        uses: nick-fields/retry@v2.8.1
+        uses: nick-fields/retry@v2.8.3
```
consideRatio, line 331:
```suggestion
        uses: nick-fields/retry@v2
```

consideRatio (commented):
Provided a few v2.8.3 -> v2 suggestions for your convenience if you wanted to go for that.

sgibson91:
> * do you wish to stay pinned to v2.x.y rather than going for v2 for that action? I'm 👍 for pinning only to the major version for an action that seem to maintain things well.

I'm not much opinionated on pinning to be honest, because dependabot watches for major and minor versions usually :)

consideRatio:
> I'm not much opinionated on pinning to be honest, because dependabot watches for major and minor versions usually :)

Ah, is it okay if we go with the `v2` then? I'm somewhat opinionated here looking to avoid the notifacation noise and git history noise related to PRs like these.

PR #2038, opened 2023-01-11
Move redirect docs to Manage Domains section

sgibson91:
@yuvipanda I moved the redirect docs from the "hub features" section to the "manage domains" section where I think it fits better, if that's ok!


yuvipanda:
Sure, ty @sgibson91!

Issue #1890, opened 2022-11-09
Document our logging setup


yuvipanda:
(child of https://github.com/2i2c-org/infrastructure/issues/1167)

Logging is *the most* important part of debugging outages, and we need a lot more documentation for that. Here's an outline

- [ ] Looking at 'live' logs (currently running components) #1869 helps with this a little, but needs more docs
- [ ] Where logs are stored *longer term* if at all
   - [ ] GCP (Stackdriver)
   - [ ] AWS (Cloudwatch - is this even set up?)
   - [ ] Azure ???
- [ ] Common things you might find in various logs
   - [ ] 5xx errors during login or server start? hub logs!
   - [ ] Errors about scaling? Autoscaler logs!
   - [ ] Dask errors? dask controller, api, traefik logs

damianavila:
Some sort of cloud-agnostic logging aggregator would be lovely to see in our infra, IMHO.

PR #2046, opened 2023-01-11
Document how to look at logs with GCP

yuvipanda:
unlike k8s logs, these provide logs for upto the last 30 days.

Ref https://github.com/2i2c-org/infrastructure/issues/1890

### DIFF
--- a/docs/howto/troubleshoot/logs.md
+++ b/docs/howto/troubleshoot/logs.md
@@ -96,3 +96,21 @@ show logs from specific schedulers or worker pods a user might have started.
 labels.k8s-pod/app_kubernetes_io/name="dask-gateway"
 resource.labels.namespace_name="<namespace>"
 ```
+
+#### Full-text search across logs
+
+If you are looking for a specific string across *all* logs, you can
+use `textPayload` as a field to search for.
+
+```
+textPayload=~"some-string"
+```
+
+This is most useful when *combined* with any of the other queries here. For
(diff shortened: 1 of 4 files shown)


sgibson91 (approved):
We should add a cross-ref to the ["Common problems and their solutions" page](https://infrastructure.2i2c.org/en/latest/sre-guide/common-problems-solutions.html) that says something like "For more general tips, see our Troubleshooting guide". Other than that, this is good to go!

docs/howto/troubleshoot/logs.md
```diff
@@ -0,0 +1,116 @@
...
+```{tip}
+Another super-quick shortcut is to replace any `-` in the username with `-2d`,
+any `.` with `-2e` and any `@` with `-40`. If your username contains more
+special characters, highly recommend using the script instead - escaping
+errors can be frustrating!
+```
+
+#### Look at dask-gateway logs
+
+The following query will show logs from all the components of dask-gateay
```
consideRatio, line 91:
```suggestion
The following query will show logs from all the components of dask-gateway
```

consideRatio (approved):
Super helpful to have this level of detail on the examples!

I would have hesitated try to get logs from google's web UI before reading this, but not any more when this resources is available!

jbusecke:
This seems super helpful for somebody like me. Thanks for putting in the effort.

PR #2056, opened 2023-01-18
AWS EKS infra: transition to k8s 1.24 and add required storage addon

### LINKED ISSUE #2054: Update AWS EKS terraform jinja templates to so we can use k8s 1.24
I wrote this issue quickly as I think it should be blocking #2017. I can get it done myself I think.

### Background

In https://github.com/2i2c-org/infrastructure/pull/2044 deploying a AWS EKS hub I did some changes related to "EBS" so that we could deploy k8s 1.24 instead of be stuck at k8s 1.22.

I did not do those changes to [a eksctl related jinja template](https://github.com/2i2c-org/infrastructure/blob/master/eksctl/template.jsonnet) though, so it needs to be re-done for future hubs we setup. So, I'm creating this issue to track the need to update the terraform template now that we can conclude that the changes I made did the trick.

#### k8s version should be 1.24 (latest supported by eksctl atm I think)
https://github.com/2i2c-org/infrastructure/blob/ad9179cd27a4c324979e3b23d70ac2b006d5e40d/eksctl/template.jsonnet#L35-L47

#### We need the EBS addon + iam permissions for nodes

Without this, what happens is that PVC (request for storage) is not given storage, and pods that wants to mount this storages doesn't get scheduled.

https://github.com/2i2c-org/infrastructure/blob/ad9179cd27a4c324979e3b23d70ac2b006d5e40d/eksctl/ubc-eoas.jsonnet#L45-L52

https://github.com/2i2c-org/infrastructure/blob/ad9179cd27a4c324979e3b23d70ac2b006d5e40d/eksctl/ubc-eoas.jsonnet#L67-L71

consideRatio:
This adds the required changes to the eksctl cluster configuration files, as generated from a template.jsonnet file (depending on a libsonnet/nodegroup.jsonnet file).

## Questions

1. Should we add iam.withAddonPolicies.ebs=true to the default node group object so all node groups get support for mounting PVCs using EBS storage, or should we just add it to the core node-group that is known to need it?

Closes #2054, but work to migrate existing hubs remain, this is covered by #2057.

### DIFF
--- a/eksctl/template.jsonnet
+++ b/eksctl/template.jsonnet
@@ -1,4 +1,14 @@
-// Exports an eksctl config file for carbonplan cluster
+// This file is a jinja2 template of a jsonnet template of a eksctl's cluster
+// configuration file, which is in turn can be used with the `eksctl` CLI to both
+// update and initialize a AWS EKS based cluster.
+//
+// This jinja2 template is only used by the deployer script as part of creating
+// new clusters. If a relevant change is made here or the dependent file
+// libsonnet/nodegroup.jsonnet, one may consider if we should manually update
+// already generated jsonnet files in this folder.
+//
+// Configuration reference: https://eksctl.io/usage/schema/
+//
 local ng = import "./libsonnet/nodegroup.jsonnet";
 
 // place all cluster nodes here
@@ -36,15 +46,31 @@ local daskNodes =
     metadata+: {
         name: "<< cluster_name >>",
         region: clusterRegion,
-        // Warning: version 1.23 introduces some breaking changes
-        // Checkout the docs before upgrading
-        // ref: https://docs.aws.amazon.com/eks/latest/userguide/ebs-csi-migration-faq.html
-        version: '1.22'
+        version: '1.24'
     },
     availabilityZones: masterAzs,
     iam: {
         withOIDC: true,
     },
+    addons: [
+        {
+            // aws-ebs-csi-driver ensures that our PVCs are bound to PVs that
+            // couple to AWS EBS based storage, without it expect to see pods
+            // mounting a PVC failing to schedule and PVC resources that are
+            // unbound.
+            //
+            // For this to work well, the nodeGroups scheduling nodes referencing
+            // PVCs with persistent storage from EBS (prometheus, etc.), need to
+            // have iam.withAddonPolicies.ebs=true as well.
+            //
+            // Related docs: https://docs.aws.amazon.com/eks/latest/userguide/managing-ebs-csi.html
+            //
+            name: 'aws-ebs-csi-driver',
+            wellKnownPolicies: {
+                ebsCSIController: true,
+            },
+        },
+    ],
     nodeGroups: [
         ng {
             name: 'core-a',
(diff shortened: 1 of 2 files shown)


yuvipanda (approved):
NAICE!

consideRatio:
Thanks for reviewing @yuvipanda! I just updated inline comments and force-pushed, and updated the PR description to include a question:

> Should we add iam.withAddonPolicies.ebs=true to the default node group object so all node groups get support for mounting PVCs using EBS storage, or should we just add it to the core node-group that is known to need it?

eksctl/template.jsonnet
```diff
@@ -1,4 +1,14 @@
-// Exports an eksctl config file for carbonplan cluster
+// This file is a jinja2 template of a jsonnet template of a eksctl's cluster
+// configuration file, which is in turn can be used with the `eksctl` CLI to both
+// update and initialize a AWS EKS based cluster.
+//
+// This jinja2 template is only used by the deployer script as part of creating
+// new clusters. If a relevant change is made here or the dependent file
+// libsonnet/nodegroup.jsonnet, one may consider if we should manually update
+// already generated jsonnet files in this folder.
```
consideRatio, line 8: For reference, this is the comment that I updated to clarify the situation about this being a jinja2 template of a jsonnet template.

yuvipanda:
@consideRatio the question to me is - where does the actual controller run? does it run just on the core nodes (as does the autoscaler, for example?) - in which case just letting it be there for the core nodes is the way to go. Or, is it a daemonset that needs to run on every node? In which case we should still probably just set it up for the core nodes (least principle and all that) but *document* that it needs to be explicitly enabled.

My intuition is that the actual controller is a deployment / statefulset that just runs on the core nodes and so having the permissions be just on the core nodes is fine.

yuvipanda:
However this intuition has no evidence :D Should be easy enough to figure out though.

yuvipanda:
@consideRatio I think the controller creates PVs, but the ds is what 'binds' them. I would suggest trying to launch a pod that has a PVC that will run on a non-core-nodepool to test this. I think the controller will always run in the core nodepool, the question is, what permissions does the daemonset need to do the attach? So we want to test 'will EBS volumes *mount* on arbitrary nodes even if they don't have this permission explicitly defined?'

consideRatio:
Thanks for your help thinking about this @yuvipanda!!!

Test: to add a new PVC, and a new Pod to mount it, and force the pod to run on the user node without considered iam ebs stuff.
Test situation 1: Run test as defined above
Test situation 2: Run test as defined above, but let the ebs controller run on the user node

Both worked out. Due to this, I figure we go with removing the nodeGroup's addition of `iam` for ebs stuff entirely and see if it works, if not, we add it. This would be a win for us, as its easier to just add the addon to the EKS based k8s cluster than to node pools with delete/create operations.

In 8fe4009 I remove the nodeGroup's iam ebs stuff and this is ready to merge. In case things won't work, lets add it back to the core node pool.

yuvipanda:
@consideRatio sounds good to me.

GeorgianaElena:
Will do! Thanks @consideRatio!

PR #2061, opened 2023-01-19
Deploy node_exporter everywhere

yuvipanda:
While exploring some questions about kernel restarts (courtesy a request by @briannalind) I realized that node_exporters were not actually running on all the nodes - only on core nodes! This prevents some helpful node info (such as oom kill count, etc) from being captured.

This PR will deploy node_exporter on all nodes as intended, and I've tested it on the openscapes cluster.

### DIFF
--- a/helm-charts/support/values.yaml
+++ b/helm-charts/support/values.yaml
@@ -31,7 +31,13 @@ prometheus:
         key: hub.jupyter.org_dedicated
         value: user
       - effect: NoSchedule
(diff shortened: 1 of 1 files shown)


consideRatio (approved):
I pushed a commit fixing a typo declaring `k8s.dask.org_dedicated` twice instead of `k8s.dask.org_dedicated` + `k8s.dask.org/dedicated`.

I also relocated `key` to be at the top as that helped me read this configuration.

PR #2062, opened 2023-01-20
basehub: add custom.2i2c.temp_folder config for ~/temp ephemeral storage

consideRatio:
@jbusecke requested a temporary storage folder under the home folder in https://2i2c.freshdesk.com/a/tickets/375, so I added this feature to meet that need. It is currently trialed out in https://staging.leap.2i2c.cloud and it seems to work out well there.

Going onwards, any hub that wants this feature can toggle `jupyterhub.custom.2i2c.temp_folder.enabled=true` and they will get it.

image

### DIFF
--- a/helm-charts/basehub/values.schema.yaml
+++ b/helm-charts/basehub/values.schema.yaml
@@ -414,3 +414,19 @@ properties:
                 type: array
                 items:
                   type: string
+              temp_folder:
+                type: object
+                additionalProperties: false
+                required:
+                  - enabled
+                  - mountPath
+                  - emptyDir
+                properties:
+                  enabled:
+                    type: boolean
+                  mountPath:
+                    type: string
+                  emptyDir:
+                    type: object
+                    # k8s schema: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#emptydirvolumesource-v1-core
+                    additionalProperties: true
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -99,6 +99,10 @@ jupyterhub:
         - jmunroe@2i2c.org
         - sgibson@2i2c.org
         - yuvipanda@2i2c.org
+      temp_folder:
+        enabled: false
+        mountPath: /home/jovyan/temp
+        emptyDir: {}
   ingress:
     enabled: true
     annotations:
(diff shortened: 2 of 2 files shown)


yuvipanda:
My general understanding of how ‘scratch’ folders work is that they are exactly the same as our ‘shared’ but get wiped out automatically every few weeks or so. This allows multiple users to see each others’ files, like a /tmp in a traditional, non containerized hpc system. I would be curious to see what the use case for replicating /tmp under ~.

if we do do this, i would very much prefer that we document explicitly adding extraVolume and similar config to the hub config directly instead of introducing a custom property that is 2i2c specific - primarily for right to replicate reasons.!

consideRatio:
> if we do do this, i would very much prefer that we document explicitly adding extraVolume and similar config to the hub config directly instead of introducing a custom property that is 2i2c specific - primarily for right to replicate reasons.!

@yuvipanda this is directly tied to z2jh + basehub. I've not assumed a user should switch away from basehub to a plain jupyterhub if they migrate, so I think the r2r aspect is unrelated to putting this feature in basehub.

Note that we can't use singleuser.extraVolumes for a clusters staging values without overriding basehub's singleuser.extraVolumes and singleuser.extraVolumeMounts due to lists being replaced rather than merged, unless we manually merge such config.

This is why I made it into Python logic under `hub.extraConfig`. If you want, we can avoid making this a general feature for basehub by putting this just into the leap cluster, but I think there is a good point to this feature.

## Why I think this feature is generally useful

User storage (NFS storage typically) can be costly, so if users start saving temporary files in their home folders, it can absolutely incur notable cost. I've myself prefered of putting things in my home folder anyhow, as I wanted to benefit from the UI provided by jupyterlab which favors the home folder over for example the `/tmp` folder.

The outcome of users preferring ~/home without a temp folder in home:
- Higher costs of storage
- More load on the NFS server

I have also reviewed it from a complexity perspective, where things that are too complicated may not be worth promoting because they may be fragile and break etc.
- I worried it could cause issues with mounting a folder into another mount - but this is fine for emptyDir volumes
- I worried about permissions on files/folders, but its fine because emptyDir respects the pod's securityContext's fsGroup, which is set by z2jh already.

But in my view, this wasn't too complicated to be worth bundling with basehub, and it did serve a value for the UX.

yuvipanda:
> Note that we can't use singleuser.extraVolumes for a clusters staging values without overriding basehub's singleuser.extraVolumes and singleuser.extraVolumeMounts due to lists being replaced rather than merged, unless we manually merge such config.

Ah damn this is good enough reason for this to be in extraConfig. And that sucks :(

yuvipanda:
I don't have any technical objections to this now, just curious if this is what Julius intended now vs a ~/temp or ~/scratch that is visible to all users without isolation

consideRatio:
I'll go for a close here as this wasn't what was wanted, and I don't think this will be adjusted to be what is wanted in the end either. No action point is on us atm with regards to this.

Julius looks to open a new support ticket when a technical request from users Julius represents is clear.

PR #2063, opened 2023-01-20
Set dry_run to default to false so PRs are actually opened

sgibson91:
An error in logic meant these workflows were not opening PRs automatically when they should have


consideRatio (approved):

sgibson91:
Relevant repos if, like me, you need to double check the logic:

- https://github.com/sgibson91/bump-helm-deps-action
- https://github.com/sgibson91/bump-jhub-image-action

PR #2077, opened 2023-01-23
docs: update notes about aws service quotas

### LINKED ISSUE #1793: Enable Service Quota Request templates on our AWS accounts
### Context

For pretty much every cloud account we open on any cloud vendor, we want to bump certain quotas so that usage is not restricted. E.g., total number of CPUs in the region/zone, total GPUs in the region/zone, etc.

AWS provide a [Service Quota Request templates service](https://docs.aws.amazon.com/servicequotas/latest/userguide/organization-templates.html) where a maximum of 10 quota increases can automatically be requested whenever a new account is created within the organisation.

### Proposal

We should enable this feature for our AWS org and remove a manual step from our "new AWS account" setup process.

### Updates and actions

_No response_

consideRatio:
- Closes #1793 that just required documentation to be added.

---

I failed to understand this fully, but I don't think this service is mature enough to be something that can be fully understood yet.

- I failed to realize how to understand what quota increases will be requested when a new AWS account is created and coupled to our AWS management account. There seem to be a bit of UI for this, but it wasn't mentioning the correct quotas.
- I failed to understand what they intent to call this, where they used at least three different ways to speak about this feature.


GeorgianaElena (approved):
Thanks @consideRatio!

Issue #2041, opened 2023-01-11
Misc documentation updates I'd like to make


consideRatio:
# What remains to be resolved

### 9

In [this aws step](https://infrastructure.2i2c.org/en/latest/hub-deployment-guide/new-cluster/aws.html#grant-eksctl-access-to-other-users) about granting eksctl access to other users.

I wonder if we should even do that if we can use the deployer script to get credentials though, hmm. Also, I have now not created additional users for that account.

UPDATE: Yes, because we need such credentials when we do operations like adding/removing node pools as well. Why one may ask, but the answer is because for example `kubectl drain` is used which is a k8s api interaction. The action point is to make it clear why we add this permission.

### 14

Its unclear to me in the [Enable authentication](https://infrastructure.2i2c.org/en/latest/hub-deployment-guide/configure-auth/index.html) section if we are supposed to add 2i2c members to be able to authenticate, and if so, with what identity provider (GitHub team? Google email accounts from 2i2c.org?).

### 23 

I saw no mention of cleaning up scratch buckets, but I think we should consider that as well in the decomissioning process.

### 24

This section didn't link to how to create an incident response issue. I asked myself, where? In 2i2c-org/infrastructure?

https://team-compass.2i2c.org/en/latest/projects/managed-hubs/incidents.html#key-terms

### 25

I'm not sure if `/pd trigger` works, or in what channel, or similar. I never managed to see a popup like described in https://team-compass.2i2c.org/en/latest/projects/managed-hubs/incidents.html#incident-response-process.

### 32

In this comment, in step 1, I ask the community reps to help authorize the github oauth application to receive organizational membership info from users instead of asking to become an owner and do it for them. With it, I provided a screenshot example.

https://github.com/2i2c-org/infrastructure/issues/2323#issuecomment-1505721800

### 33

Setting up a new GCP Project with the existing billing account should make it clear that only _new billing accounts_ as compared to new gcp projects already linked to the 2i2c billing account need to configure cost exports.

https://infrastructure.2i2c.org/hub-deployment-guide/cloud-accounts/new-gcp-project/#create-a-new-gcp-project

### 34

There should be a final step linking to setting up quotas in https://infrastructure.2i2c.org/hub-deployment-guide/cloud-accounts/new-gcp-project

### 35

The docs for creating a new gcp project doesn't mention the ability to generate that and cluster config etc via `deployer generate-gcp-cluster`

[...]

PR #2082, opened 2023-01-23
Set daskNodes = [] if dask isn't specified

yuvipanda:
Use jinja's templating abilities here, as otherwise we get a `if "daskhub" == "daskhub"` in the jsonnet itself that is a little hard to read

### DIFF
--- a/eksctl/template.jsonnet
+++ b/eksctl/template.jsonnet
@@ -26,20 +26,24 @@ local notebookNodes = [
     { instanceType: "m5.2xlarge" },
     { instanceType: "m5.8xlarge" },
 ];
-
-local daskNodes =
-    if "<< hub_type >>" == "daskhub" then [
+<% if hub_type == "daskhub" %>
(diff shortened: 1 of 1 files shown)


consideRatio:
> Use jinja's templating abilities here, as otherwise we get a if "daskhub" == "daskhub" in the jsonnet itself that is a little hard to read

This resolves my point 3 in https://github.com/2i2c-org/infrastructure/issues/2041#issue-1529041983 that I was going to ask you about =D Nice!!

If this works, this LGTM!

GeorgianaElena (approved):
This is much better indeed! Thanks @yuvipanda!

damianavila (approved):

consideRatio:
I'll go for a merge, this doesn't influence existing hubs and seems right to all of us having reviewed this.

yuvipanda:
thank you, @consideRatio!

Issue #2080, opened 2023-01-23
[Incident] Upgrading ingress-nginx caused brief outage in carbonplan


consideRatio:
### Summary

The CI/CD pipeline when merging a PR (#2067) upgrading the support chart's dependencies on `ingress-nginx`, `prometheus`, and `grafana` [failed like this](https://github.com/2i2c-org/infrastructure/actions/runs/3989923635). Among the failures I observed most of them related to a mundane failure to auth with GKE that by chance showed up now, but at least one was related to AWS - the carbonplan cluster.

In the carbonplan cluster, the staging and prod wasn't accessible so I reverted the version bump in #2079 which restored carbonplan but overall failed to deploy fully to other hubs as there was need for a patch commit still that was reverted.

#### GKE Auth issue

```
Unable to connect to the server: getting credentials: exec: executable gke-gcloud-auth-plugin not found

It looks like you are trying to use a client-go credential plugin that is not installed.

To learn more about this feature, consult the documentation available at:
      https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins

Install gke-gcloud-auth-plugin for use with kubectl by following https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke
```

#### Carbonplan issue

I'm very confident it relates to carbonplan being the one eksctl based AWS cluster that has k8s 1.19, while others have k8s 1.21 or higher. I assume the ingress-nginx chart has dropped support for so ancient versions, just like z2jh has already.

### Impact on users

I think only users of carbonplan was disrupted, and it seems nobody used it.

### Important information

- Hub URL: carbonplan.2i2c.cloud
- Support ticket ref: I saw this myself

### Tasks and updates

- [x] Discuss and address incident, leaving comments below with updates
- [x] Incident has been dealt with or is over
- [x] Copy/paste the after-action report below and fill in relevant sections
- [x] Incident title is discoverable and accurate
- [ ] All actionable items in report have linked GitHub Issues

<details>
<summary>After-action report template</summary>

```
# After-action report

These sections should be filled out once we've resolved the incident and know what happened.
They should focus on the knowledge we've gained and any improvements we should take.

## Timeline

_A short list of dates / times and major updates, with links to relevant comments in the issue for more context._

All times in UTC+1.

- 2023-01-23 - 20:38, #2067 was merged
[...]

PR #2074, opened 2023-01-23
meom-ige: add dakkar-demo, a temporary basehub

consideRatio:
- Closes #2049

PR #2086, opened 2023-01-24
Bumping helm chart dependency versions: support (again)

consideRatio:
- In short, this PR is a new attempt of #2067 which is reasonable to make now that carbonplan's k8s version is upgraded from 1.19 to 1.24, which makes all of our k8s clusters be versioned 1.21 or higher - where the only cluster stuck at 1.21 is openscapes.
- "In long", this reverts commit d680896f1ba32a80643861a72650fffc6c1dc400, that was the merge commit for #2079, that in turn reverted #2067 after the incident in carbonplan #2080.

### Related action points

- [x] Review #2074 which I'd like to merge before this
- [ ] Review this PR
- [ ] Review a subsequent PR I'll open removing the patch to the deployer script in this PR

### DIFF
--- a/deployer/cluster.py
+++ b/deployer/cluster.py
@@ -61,6 +61,19 @@ class Cluster:
         )
         print_colour("Done!")
 
+        # FIXME: Revert this addition after one upgrade is made
+        print_colour("Preparing for prometheus v16 v19 migration...")
+        subprocess.check_call(
+            [
+                "kubectl",
+                "delete",
+                "daemonset",
+                "--namespace=support",
+                "support-prometheus-node-exporter",
+            ]
+        )
+        print_colour("Done!")
+
         print_colour("Provisioning support charts...")
 
         support_dir = (Path(__file__).parent.parent).joinpath("helm-charts", "support")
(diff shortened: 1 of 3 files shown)


consideRatio:
@GeorgianaElena @sgibson91 does this look okay to you? Deploying this yesterday night caused the outage in carbonplan, but I've now checked and seen that all k8s clusters use v1.21 or higher, so I think this is safe to go for.

Yet, I'd still like to do it now before people in USA wake up and I have time to babysit it.

deployer/cluster.py
```diff
@@ -61,6 +61,19 @@ def deploy_support(self, cert_manager_version):
...
+        subprocess.check_call(
+            [
+                "kubectl",
+                "delete",
+                "daemonset",
+                "--namespace=support",
+                "support-prometheus-node-exporter",
+            ]
+        )
+        print_colour("Done!")
```
GeorgianaElena, line 75: This is meant to be deleted (by opening a new PR) after the clusters has been successfully upgraded, right?

GeorgianaElena (approved):
Yes, this looks good to me! Let's merge and see how it goes!

deployer/cluster.py
consideRatio: Yepp?

consideRatio:
Two failures ocurred, and I think they relate to temporary broken connections by upgrading ingress-nginx. They did not cause an outage in the hubs. I've re-deployed the support charts manually and things work well!

consideRatio:
Scratch that, there was one failure with pangeo-hubs restarting prometheus-server - I presume it relates to a lack of memory to startup. I'll resolve this but this is not related to the upgrade here.

PR #2022, opened 2023-01-05
[RECONSTRUCTION] Setup shared cluster on AWS and deploy 'researchdelight' hub

sgibson91:
Reconstruction of #1967 since something went horribly wrong git-wise over there.

Adds 2i2c-aws-us cluster with staging, dask-staging and researchdelight hubs.

fixes #1949

PR #1967, opened 2022-11-30
Setup shared cluster on AWS and deploy 'researchdelight' hub

sgibson91:
ref 2i2c-org/infrastructure#1949 

Since this is a new shared cluster, I will deploy a staging and dask-staging hub alongside the research delight hub requested, mimicking the GCP shared cluster setup.

Hubs added:

- https://staging.aws.2i2c.cloud
- https://dask-staging.aws.2i2c.cloud
- https://researchdelight.2i2c.cloud

PR #2087, opened 2023-01-24
Add jsonnet file for 2i2c-aws-us cluster

### LINKED ISSUE #2084: Add missing .jsonnet file representing a eksctl config for 2i2c-aws-us
Looking through all the k8s clusters we manage, 2i2c-aws-us stood out among the AWS EKS based clusters for not having a etsctl/2i2c-aws-us.jsonnet file generated from eksctl/template.jsonnet like the others.

Yuvi suspected it was missing from https://github.com/2i2c-org/infrastructure/pull/2022 by accident.

### Action point

- [ ] Let's create such file to match what we have in the 2i2c-aws-us EKS based cluster
  In practice, we need to know what k8s version, region, zones, and maybe node pools setup etc, then we generate it from the template. Perhaps as the template file was at the time of #2022.

sgibson91:
This file was missed in #2022, and recovered from #1967

fixes #2084 

How this PR was created:

```bash
gh pr checkout 1967  # This reinstated the researchdelight branch from PR 1967
git switch master
git switch -c add-2i2c-aws-us-jsonnet
git checkout researchdelight eksctl/2i2c-aws-us.jsonnet
```

Then add, commit, push as normal

### DIFF
--- a/eksctl/2i2c-aws-us.jsonnet
+++ b/eksctl/2i2c-aws-us.jsonnet
@@ -1,0 +1,118 @@
+// Exports an eksctl config file for carbonplan cluster
+local ng = import "./libsonnet/nodegroup.jsonnet";
+
+// place all cluster nodes here
+local clusterRegion = "us-west-2";
+local masterAzs = ["us-west-2a", "us-west-2b", "us-west-2c"];
+local nodeAz = "us-west-2a";
+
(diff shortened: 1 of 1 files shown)


consideRatio (approved):
Thanks @sgibson91, feels good that this was found rather than re-created!

PR #2089, opened 2023-01-24
deployer: cleanup no longer needed patch for prometheus upgrade

consideRatio:
- This cleans up the patch to the deployer script that was added in #2086 to handle upgrading prometheus from v16 to v19


GeorgianaElena (approved):
🎉

PR #2091, opened 2023-01-24
Rewrite the "documentation structure" page to better reflect how our docs are currently organised

### LINKED ISSUE #2055: Rewrite our documentation structure page
### Context

This page that aims to describe how our documentation is structured feels a little bit outdated https://infrastructure.2i2c.org/en/latest/structure.html

For example, it does not reference our new Hub Deployment Guide or SRE guide, and it references audiences that likely are not visiting these particular docs (perhaps it was written before we had docs.2i2c.org?)

### Proposal

Give the page a refresh to reflect the current docs structure so folk can more easily orient themselves in the docs

### Updates and actions

_No response_

sgibson91:
This mostly reduces the complexity of this page by only focussing on 2i2c engineers as the audience, and the structure of the infrastructure.2i2c.org site only. It documents the SRE and Hub Deployment guides as well as the breakdown of our other docs that don't fall into those guides.

**Note:** I have kept a local copy of the original file contents because I do really like the idea of thinking about audiences and breaking down docs accordingly, and how we split that between infrastructure repo and the docs repo. But I think that discussion would be better placed into the team compass than in the infrastructure repo itself. So I will follow up with a PR over there bringing back some of the nuance and different audiences that this PR removes.

Link to rendered page: https://2i2c-pilot-hubs--2091.org.readthedocs.build/en/2091/structure.html

fixes #2055

### DIFF
--- a/docs/hub-deployment-guide/cloud-accounts/new-aws-account.md
+++ b/docs/hub-deployment-guide/cloud-accounts/new-aws-account.md
@@ -48,7 +48,7 @@ More information on these terms can be found in [](cloud-access:aws).
    * On the "Review and submit assignments" page, click "Submit".
 
 You have successfully created a new AWS account and connected it to our SSO Management Account!
-Now, [setup a new cluster](new-cluster:new-cluster-aws) inside it via Terraform.
+Now, [setup a new cluster](new-cluster:aws) inside it via Terraform.
 
 ## Checking quotas and requesting increases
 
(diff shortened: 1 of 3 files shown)


sgibson91:
@yuvipanda really keen to hear what you think and if this rewrite helps at all?

GeorgianaElena (approved):
I agree with keeping the docs structure description reflect what is actually happening in the docs rather that what we'd like to achieve. And then move the conversation around how we want to improve it somewhere else, so we can plan around it.
LGTM 👍🏼

damianavila (approved):

yuvipanda:
Thanks @sgibson91!

PR #2093, opened 2023-01-25
update m2lines image

rabernat:
This is my attempt to bring the m2lines hub up to date with leap (see #2013).


rabernat:
Thanks Yuvi!

Issue #1253, opened 2022-04-27
Support choosing images in profile list for pangeo hubs


yuvipanda:
### Context

Implementing https://github.com/jupyterhub/kubespawner/issues/307 will give users more choice in which images they want for various profiles.

<img width="965" alt="image" src="https://user-images.githubusercontent.com/30430/165631724-21c7b2e5-2e83-45f6-9004-c00b82be8f0e.png">

### Proposal

- [ ] Implement image selection on a per-profile basis
- [ ] Get PR to kubespawner
- [ ] Deploy that on to the pangeo hubs

yuvipanda:
Upstream PR in https://github.com/jupyterhub/kubespawner/pull/607

yuvipanda:
blocked on https://github.com/2i2c-org/infrastructure/issues/1055

damianavila:
* Kubewspawner release (or pull a particular hash): https://github.com/jupyterhub/kubespawner/pull/640 (something can tackle on this sprint (@georgiana)
* Dynamically provide the list is the goal.

PR #2094, opened 2023-01-25
Add new image option for LEAP classes

jbusecke:
I would like to test a custom image (details overe [here](https://github.com/jbusecke/leap_education_docker_image), made according to the instructions [here](https://docs.2i2c.org/en/latest/admin/howto/environment/hub-user-image-template-guide.html?highlight=image#create-a-custom-user-image-for-your-hub)). If I understand correctly using the ['configurator'](https://pilot.2i2c.org/en/latest/admin/howto/configurator.html) this will change the image for every user starting a new server (which I want to avoid at all cost rn, since this is very early testing 😬). 

Ultimately I would like to have an option to pull arbitrary images using tags (but I will send a separate email for that, also cc @jmunroe, who might have already raised this more general issue).

### DIFF
--- a/config/clusters/leap/common.values.yaml
+++ b/config/clusters/leap/common.values.yaml
@@ -112,7 +112,7 @@ basehub:
                   display_name: LEAP Education Notebook (Testing Prototype)
                   slug: "leap_edu"
                   kubespawner_override:
-                    image: "quay.io/jbusecke/leap-edu-image:e8cf3f0ccb57"
+                    image: "quay.io/jbusecke/leap-edu-image:9896814442a9"
           kubespawner_override:
             mem_limit: 15G
             mem_guarantee: 11G
--- a/config/clusters/leap/common.values.yaml
+++ b/config/clusters/leap/common.values.yaml
@@ -83,6 +83,27 @@ basehub:
             mem_guarantee: 4.5G
             node_selector:
               node.kubernetes.io/instance-type: n1-standard-2
+         - display_name: Medium Admin Test
+          description: 11GB RAM, 4 CPUs
+          allowed_teams:
+            - leap-stc:webmasters # basically just julius and ryan for testing things. I could make a separate admin group?
+            - 2i2c-org:hub-access-for-2i2c-staff
+          profile_options:
+            image:
+              display_name: Image
+              choices:
+                pangeo:
+                  display_name: Base Pangeo Notebook
+                  default: true
+                  slug: "pangeo"
+                  kubespawner_override:
+                    image: "pangeo/pangeo-notebook:2023.01.03"
+                custom:
+                  display_name: Prototype Education Notebook
+                  slug: "leap_edu"
+                  kubespawner_override:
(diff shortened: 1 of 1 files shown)


sgibson91:
Is testing the image on the staging hub not possible? https://staging.leap.2i2c.cloud You can use the configurator on that hub without affecting users on the prod hub because it's a whole separate JupyterHub.

jbusecke:
Ah, that makes a lot of sense! Let me test and report back!

sgibson91:
Actually, the reason the configurator probably doesn't work because the hubs use `profileList` and the configurator isn't compatible with that yet.

jbusecke:
I got an error with this: 
image

not sure if that is consistent. 

How about an alternate proposal: 
I can modify this PR to offer an additional image selection (which will be needed by the LEAP instructors and students soon). This might however require some version bumps if things are not behaving as expected? 

To make this abundantly clear I can add `(Testing Prototype)` to the description?

sgibson91:
The 401 Unauthorised error makes me think that your quay.io repo is private, not public.

jbusecke:
You were right, I just changed that. Sorry for the baby mistakes here. Pretty new to all of this.

So now I was able to start a server, but the image still seems to be the default pangeo-notebook (not quite sure how to confirm this, but the dependencies I need are definitely not there, and a bunch of the pangeo packages are installed instead). Is that the behavior you expected to see?

jbusecke:
I was actually able to test this image by selecting one of the server instances without a dropdown menu (here `Small`), and everything looks like it works to me. 

I also added some words to the docs about making the image repo public [here](https://github.com/2i2c-org/docs/pull/181) (in the hope it might help future folks...or maybe future me hehe). 

With this working I would like to get this image option exposed so that the educators can develop their material directly on the hub. 

Thanks for the detailed help here @sgibson91! I really appreciate it.

sgibson91:
> So now I was able to start a server, but the image still seems to be the default pangeo-notebook (not quite sure how to confirm this, but the dependencies I need are definitely not there, and a bunch of the pangeo packages are installed instead). Is that the behavior you expected to see?

> I was actually able to test this image by selecting one of the server instances without a dropdown menu (here `Small`), and everything looks like it works to me.

The reason you see this behaviour is because the configurator can only override images that are defined in this section of the JupyterHub config:

https://github.com/2i2c-org/infrastructure/blob/a591bc4f486cac521ec35018b76438d32242d341/config/clusters/leap/common.values.yaml#L61-L63

Any of the images that have dropdown menus, and are nested under `profileList`, **cannot** be overridden by the configurator. So yes this is expected behaviour. Not ideal behaviour, but expected. We do have a longstanding issue open to allow arbitrary selection of any tags in an image repo here: https://github.com/2i2c-org/infrastructure/issues/1253

> I also added some words to the docs about making the image repo public [here](https://github.com/2i2c-org/docs/pull/181)

Thank you so much for the PR!

sgibson91 (approved):

PR #2100, opened 2023-01-26
reverting to Python 3.9

betolink:
Datashader and friends are not totally compatible with Python 3.11, until they are we'll keep using 3.9 :(


GeorgianaElena (approved):
Thanks @betolink

Issue #2042, opened 2023-01-11
Rethink the new hub deployment request/provide information


GeorgianaElena:
### Context

We currently have two issue templates that should facilitate
- gathering information about a new hub deploying
- presenting the information to the engineer team

But in reality, they are not serving their purpose. For example, the `request information` template has never been used event when not enough information was available about a new hub to request a deployment.
I believe this is because:

- we don't want to be tasking the community representatives with opening an issue themselves
- opening two issues about the same new hub is generating work that's not necessary
- other?

### Proposal

I propose we remove the `request information` issue template and instead replace it with a bot that comments to an incomplete request deployment issue with a message that asks for the missing info and tags both the community rep and the person opening the issue.
(Maybe we can also make it work as a reminder too and ask every week until the top line comment is filled in or something like this?)

### Updates and actions

- remove the `request information` issue template
- create the bot to comment to the `request deployment` issue for any missing info

damianavila:
I agree with the proposal, the two forms are confusing for everyone, I think.
Simplification with just one form and assistance with a bot to remind us to get the info seems like a plausible and interesting next step.

PR #2115, opened 2023-01-30
Replace the two template workflow for new hubs with a commenter workflow

GeorgianaElena:
Fixes: https://github.com/2i2c-org/infrastructure/issues/2042

## Example

Example of the workflow in action https://github.com/GeorgianaElena/pilot-hubs/issues/6 and below

Screenshot 2023-01-31 at 18 09 14

## How does it work
Any time an issue with a `new hub` label is opened, the top comment is checked for any fields that were left empty or filled in with "pending". A comment is then posted by the action with all of of these form fields and the community rep and the person opening the issue is tagged.

## Question

Do you think it would be better to have the workflow run and the comment be posted when an eng is assigned with such an issue, instead the first time the issue is opened?

## Todo:
- [x] fix the comment template
- [x] add more comments
- [ ] update new hub workflow docs


choldgraf:
This is pretty cool! Thanks for the effort @GeorgianaElena .

Regarding your question: I think the question comes down to who generally opens these issues and what the workflow looks like. Sometimes an issue might be created that isn't quite yet ready for others to look at, and if that's the case we might want to require one extra action (e.g. assigning a team member to it) to trigger this bot. But if we expect that we already have all the information needed _before the issue is opened_, then I think it's OK to just use the "issue open" event as a trigger.

.github/comment-templates/comment-pending-deployment-info.md
```diff
@@ -0,0 +1,9 @@
+Hey @{{ .community_rep }} and @{{ .issue_author }}! 👋
+I noticed there is still pending information about the new hub deployment.
+Can you please help us fill it in?
+
+The information pieces still missing, are: **{{ .pending_info }}**
```
choldgraf, line 5: could this be formatted as a list or a little table? It might be easier to parse that way.

.github/ISSUE_TEMPLATE/2_new-hub-provide-info.yml
```diff
@@ -12,13 +12,25 @@ body:
...
-    id: important-dates
     attributes:
-      label: Important dates
+      label: Hub important dates
       description: |
-        Any important dates that we should consider for this hub. For example, if it will be used for a class, the starting day of class and any days with tests.
-      value: |
+        Any important dates that we should consider for this hub.
+        For example, if it will be used for a class, the starting day of class and any days with tests.
+        **If the dates are not yet known, only write "pending" in the text area below.**
```
sgibson91, line 32:
```suggestion
        **If the dates are not yet known, please write "pending" in the text area below.**
```

.github/comment-templates/comment-pending-deployment-info.md
```diff
@@ -0,0 +1,9 @@
...
+
+Details about each of them can be found in the top comment. But if you have questions about any of them, please ping the "2i2c/engineering" and they will help you.
```
sgibson91, line 8:
```suggestion
Details about each of them can be found in the top comment. But if you have questions about any of them, please ping the `2i2c/engineering` team and they will help you.
```

.github/workflows/comment-pending-deployment-info.yaml
```diff
@@ -0,0 +1,59 @@
...
+                  and "optional" not in form_entry_id
+                  and (not form_entry_value or form_entry_value == "pending")
+              ):
+                  # Form ids have `_` in them, so replace their occurrences with a space
+                  # to improve readability
+                  pending_info += f'{form_entry_id.replace("_", " ")}, '
+          # Export the `pending_info` string to GITHUB_OUTPUT to be used in following steps
+          with open(os.environ["GITHUB_OUTPUT"], "a") as f :
+              print(f"pending_info={pending_info[:-2]}", file=f)
+              print(f"community_rep={community_rep}", file=f)
```
sgibson91: I believe you will need to explicitly append a newline character `\n` here, otherwise both variables may end up on the same line and will not be parsable. I got bitten by this in another workflow.

sgibson91 (commented):
Thanks, this is great work @GeorgianaElena! I think some clarity on what the posted comment would actually look like would help a bit. E.g., will the whole template get re-rendered again?

sgibson91:
Ignore me, I properly read the top comment and saw your link now

.github/workflows/comment-pending-deployment-info.yaml
GeorgianaElena: I echoed the two vars in a step below at some point, and fortunately they worked, so I think for some reason they are also ok  like this? 🤷🏼‍♀️

sgibson91: Yeah, it might be because you used `print(..., file=...)` which I think would append a newline. But I was using `f.write()` which I don't think does.

.github/comment-templates/comment-pending-deployment-info.md
GeorgianaElena: Is something like this good enough @choldgraf?

I had to format the `pending_info` string because I don't think there's a way to parse a list or a string from within this kind of markdown template and https://github.com/chuhlomin/render-template won't know what to do with it.

Screenshot 2023-01-31 at 18 09 14

choldgraf:
@jmunroe and @colliand please provide feedback for Georgiana, thanks.

@GeorgianaElena I think it might be easiest if you requested their review in the PR, because it is easy to miss pings that are embedded in GitHub comments

GeorgianaElena:
> @GeorgianaElena I think it might be easiest if you requested their review in the PR, because it is easy to miss pings that are embedded in GitHub comments

Good point. For some reason I am unable to request @jmunroe's review 🤔
 
Screenshot 2023-02-01 at 17 27 49

jmunroe:
I agree we should combine the two templates into a single commenter workflow.

My main concern is who the main audience is for this workflow.  In the the text, it refers to "you" as if the intended respondee is the community rep.  My opinion is that would only be the case for a very small number of cases, namely, community reps that we already have a established working relationship with.

In most cases, I think these questions should not be answered directly from the community rep.  I like that the community rep has visibility into the deployment of their hub but I think the primary audience should be from 2i2c partnerships who is responsible for translating our community facing service offering agreement into an engineering request.

I need to discuss this further with @colliand before making specific recommendations.

jmunroe:
I think I understand the commenter workflow as a way of automatically checking when all the required fields have been filled in properly and, if not, requesting that additional information. Neat.

Could we add information on the 2i2c partnership representative who was involved in this particular hub request? Is there a way to have a 2i2c partnerships rep "sign-off" on the hub request indicating that they believe it is ready to proceed to engineering? (If we were a larger organization, I'd expect this "sign-off" to be the repsonsiblity a partnership manager but we don't have those organizational roles yet)

GeorgianaElena:
Thank you very much for the feedback @jmunroe!

> Could we add information on the 2i2c partnership representative who was involved in this particular hub request? 

Right now the bot will ping the person listed as the community representative in the form, and also the person that opened the hub request issue, which usually is 2i2c partnership representative who was involved in the particular hub request.

But we could also add another field to the form where we would explicitly list the 2i2c partnership representative if we expect that to be different?

Also, when you say _"add info on the 2i2c partnership rep"_ you mean mention this information somewhere in the bot's message?

> Is there a way to have a 2i2c partnerships rep "sign-off" on the hub request indicating that they believe it is ready to proceed to engineering?

The way I understand the workflow at the moment is that ideally, when the request is being assigned to an eng, then it's considered to be ready to be worked on.

With this in mind, then maybe instead of triggering the bot message, with the missing info, right after an issue was opened, maybe we can make it comment when someone was assigned to the issue, but there is still pending info.

If the GitHub handle of the community rep is provided, then the bot will ping them and the person that opened the request initially (a partnerships rep), if not, just the parnerships rep, letting them know about the missing info they need to be gathering from the community.

Do you think something like this would make sense @jmunroe ?

damianavila:
> With this in mind, then maybe instead of triggering the bot message, with the missing info, right after an issue was opened, maybe we can make it comment when someone was assigned to the issue, but there is still pending info.

I like that idea.

GeorgianaElena:
This is great news @colliand! 🎉 Thank you for working on improving the process.

I was thinking the other day what to do with this PR, and given your feedback, what do you all think about the following:
- trigger a comment with the missing info only when someone is assigned to a new hub issue and there is missing info
- in the comment, only ping the 2i2c representative that opened the new hub deployment issue and not the community representative, to allow for the discussion around the missing info to happen on other channels, outside of GitHub

Given the new process on Community and Partnerships side, I expect the workflow to not be needed anymore, or at least not that often. But I would like to have it still, for those rare cases when info is not fully available.
The important end goal for me is avoid the situation where an engineer is assigned to deploy a hub one week ahead the target/required date, and timezone sensitive back and forth to get some missing info is still required.
Does this make sense?

damianavila:
> to allow for the discussion around the missing info to happen on other channels, outside of GitHub

Or in the same GH issue, I am not sure why you are proposing to port the discussion to another place, @GeorgianaElena.
Can you clarify this further?

> Does this make sense?

In general, it makes sense to me, @colliand and @jmunroe WDYT?

colliand:
I think I understand what @GeorgianaElena is suggesting in the comment above so I'll answer the question posed by @damianavila as best as I can.

The new process for launching a hub involves a two step communications channel: 
```mermaid
  graph LR;
      A(Champion)--semi-structured--> B(Community + Partnership);
      B--structured-->C(Engineering);
```

The approach here is for Community + Partnerships (C+P) to serve as a liaison between the hub champion and 2i2c's engineering team. The goal is for C+P to gather the information required from the hub champion and convert that into a structured payload that engineering can act on without requiring exchanges with the hub champion. 

I interpret the comment from @GeorgianaElena above as a suggestion that engineering ping C+P  for information requests in the comments under a new hub request issue rather than making direct requests from the hub champion. The new hub request issue can remain publicly visible under 2i2c's infrastructure repo.

damianavila:
I think that is a good general description but I will ask to @GeorgianaElena to add anything else we are not yet interpreting from her writings (if there is anything left 😉 ).

GeorgianaElena:
@damianavila, @colliand, your summary is spot on. Also, love the graph ✨ 

> Or in the same GH issue, I am not sure why you are proposing to port the discussion to another place, @GeorgianaElena.
Can you clarify this further?

@damianavila, this is to match the current workflow that Jim described above, where the "semi-structured" discussion between C+P and the hub representative happens outside of GitHub and the GitHub issue is for C+P and engineering with (optional) additional input from the hub representative when the relationship is well established and there's familiarity with GitHub (more about this in https://github.com/2i2c-org/infrastructure/pull/2115#issuecomment-1419306668).


### Next steps

If there are no objections, I am planning to take the following actions:

- keep only one new hub issue template (the `request deployment` that we're currently using exclusively)
- when people are assigned to these new hub issues (usually engineers), post a comment about the `pending` info, pinging the person that opened the issue in the first place (usually someone from C+P)

These should be pretty straightforward and not a big change from our current workflow.

colliand (approved):

PR #2116, opened 2023-01-30
Bump LEAP edu image version

jbusecke:
Adding updated image version for LEAP edu image to support @RobertPincus teaching efforts.

### DIFF
--- a/config/clusters/leap/common.values.yaml
+++ b/config/clusters/leap/common.values.yaml
@@ -112,7 +112,7 @@ basehub:
                   display_name: LEAP Education Notebook (Testing Prototype)
                   slug: "leap_edu"
                   kubespawner_override:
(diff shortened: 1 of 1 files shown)


jbusecke:
Running into some unexpected pre-commit errors here. We just had the same issue over at [xgcm]() and we got it fixed by setting the `default-language-version` to [`python3`](https://github.com/TomNicholas/xgcm/blob/1a38c364036c0e56d40edac251516d8c05cdc3ba/.pre-commit-config.yaml#L2-L3) from `python3.9`. Not sure if that is helpful here...

sgibson91:
@jbusecke It was fixed by https://github.com/2i2c-org/infrastructure/pull/2117 Rebasing your PR should resolve the tests

consideRatio (approved):
The failure is unrelated so we can merge as it is!

PR #2132, opened 2023-02-02
docs: use env vars in docs bash cmd

GeorgianaElena:
This should make copying commands from the docs more straightforward, by removing the need to edit them before.


consideRatio (approved):
Wieee great @GeorgianaElena :heart: :tada: :sunflower: !!

It has bugged me a bit to copy strings that I then edit parts from, making sure to setup a few env vars and then sticking to pure copy/paste going onwards is a far better experience!

I also value that the `export SOMETHING=<asdf>` sections are _in dedicated code snippets_, because then you can use the copy button for entire code snippets free of things that needs updating.

Issue #1888, opened 2022-11-08
Add Monthly Active Users to our Global Usage Grafana dashboard


choldgraf:
### Context

We track "global usage" in [this grafana dashboard](https://grafana.pilot.2i2c.cloud/d/global-usage-dashboard/global-usage-dashboard). Currently it consists of a single plot showing the total active users of the last 7 days for each cluster.

In a recent conversation we agreed that **Monthly Active Users** was one of the metrics we'd like to track to observe the growth of 2i2c's overall service operations.

### Proposal

I propose that we add a plot that shows **Unique users with a 30-day rolling window for each cluster, and across all clusters**. This would be a timeseries plot over the last 6 months. We can use this as the source of truth for our "Monthly Active Users" KPI for now.

Alternatively we could roll our own MAU plot via the Prometheus data we can access in the instructions here:

- https://github.com/2i2c-org/infrastructure/issues/1785

### Updates and actions

* We also received a request from UofT around these numbers as well https://2i2c.freshdesk.com/a/tickets/269

yuvipanda:
Note that we removed the closest we had to this upstream recently as nobody found time to validate those metrics: https://github.com/jupyterhub/grafana-dashboards/pull/45

choldgraf:
@yuvipanda are you saying that in order to implement this we will need to do our own validation?

yuvipanda:
"Upstream" is a bit blurred as usual, given I'm the person who made that PR and merged it - primarily because we couldn't find time / resources on the berkeley side to actually validate that that dashboard isn't lying.

I think steps here are to:

1. Look at the queries in that dashboard
2. Validate them against a *source of truth* - which is *hub logs*
3. If these look valid, revert that PR
4. Deploy that PR here.

The upshot basically is that the query for this already exists (in that reverted PR), but definitely needs work validating it before we can actually use it.

choldgraf:
To me the question comes down to whether we should we roll a "good enough for 2i2c" implementation and not consider the considerations of the broader JupyterHub team for now. And then plan to make an upstream improvement once we (and Berkeley as well?) have had the time to validate the numbers. Maybe we can be a test-case that validates the right implementation? IMO this is worth devoting 2i2c resources towards since it solve both our problems and would be useful for the broader community

yuvipanda:
Cross-referencing https://github.com/2i2c-org/infrastructure/issues/1785#issuecomment-1308494647 which might be helpful.

[...]

PR #2136, opened 2023-02-03
Bump z2jh to get new JupyterHub 3.1

yuvipanda:
Brings in https://github.com/jupyterhub/jupyterhub/pull/4214,
which lets us graph active user metrics! It's a fairly small release
(see changelog: https://jupyterhub.readthedocs.io/en/stable/changelog.html#id1),
and I don't expect many changes.

Ref https://github.com/2i2c-org/infrastructure/issues/1888


consideRatio (approved):
It should fail on staging at worst if there was a breaking change i forgot about in z2jh!

yuvipanda:
did a quick test, seems ok!

Issue #2017, opened 2023-01-03
[Request deployment] New Hub: NASA VEDA Hub


colliand:
### Important dates

- **Target start date**: 2023-01-24
- **Required start date**: 2023-02-01

### Hub Authentication Type

GitHub (e.g., @mygithubhandle)

### First Hub Administrators

@abarciauskas-bgse
@freitagb
@j08lue
@rezuma

### [GitHub Auth only] How would you like to manage your users?

Allowing members of specific GitHub team(s): `@veda-analytics-access/all-users`

### [GitHub Teams Auth only] Profile restriction based on team membership

> No restriction for now, thanks - @veda-analytics-access/all-users should get the one machine size we have. We might want to add more in the future, though, if that is okay.

### Hub logo image URL

pending

### Hub logo website URL

https://www.earthdata.nasa.gov/esds/veda

### Hub user image GitHub repository

> 2i2c default / pangeo-notebook https://github.com/pangeo-data/pangeo-docker-images/tree/master/pangeo-notebook

### Hub user image tag and name

> 2i2c default / latest tagged release - at the time of writing pangeo/pangeo-notebook:2023.01.04

### Extra features you'd like to enable

- [x] Specific cloud provider or datacenter (otherwise GCP)
- [ ] Dedicated Kubernetes cluster
- [x] Scalable Dask Cluster

### (Optional) Preferred cloud provider

> [Science Managed Cloud Environment (SMCE)](https://www.nccs.nasa.gov/systems/SMCE) - a managed Amazon Web Service (AWS) based infrastructure for NASA funded projects

> We had a coordinating discussion between 2i2c and NASA earlier this week and found a working arrangement. SMCE creates the AWS organization and 2i2c gets accounts with full access, as I recall.

### (Optional) Billing and Cloud account

To be provided by SMCE @rezuma

### Tasks to deploy the hub

- [x] 1. Deploy information filled in above
- [x] 2. Engineer who will deploy the hub is assigned
- [ ] 3. If using GitHub Orgs/Teams Auth, Engineer is given Owner rights to the org to set this up.
- [ ] 4. Initial Hub deployment PR: https://github.com/2i2c-org/infrastructure/pull/2083
- [ ] 5. Administrators able to log on -> Hub now in steady-state

damianavila:
@colliand, I see a lot of unknowns (pending) on this request 😉.
[...]

PR #2137, opened 2023-02-03
Specify funding source and add admin users

j08lue:
Providing a missing piece of information from https://github.com/2i2c-org/infrastructure/issues/2017#issuecomment-1405190865

And using the chance to add two admin users - I realize I can probably also do that through the JupyterHub admin interface.


yuvipanda:
Thanks, @j08lue!

PR #2140, opened 2023-02-05
Document setting 'region' when sshing into EKS node

yuvipanda:


consideRatio (approved):

consideRatio:
@GeorgianaElena made a lot of our docs shell commands be written in a way that directly references env vars in #2132, which makes it easier to copy paste snippets of code overall. This is especially relevant if you run many commands all referencing some cluster name, cluster region, or similar variable.

Since this is not a lengthy instruction with repeated use of the region variable, let's go for a merge!

PR #2143, opened 2023-02-06
docs: unpin python dependencies, use ubuntu 22.04 build context on RTD

consideRatio:


choldgraf (approved):
+1 from me - I am not sure why these are pinned to specific versions and agree it's probably fine.

I mean, a much worse practice is probably pinning to `main` for our theme repository 🙃

PR #2147, opened 2023-02-06
Add a regex so main tag for r2d image is ignored

sgibson91:
Fixes #2141


consideRatio (approved):
A clear improvement on the current situation, let's go for it.

PR #2148, opened 2023-02-06
daskhub: update comments in values, removing two fixme notes

consideRatio:
@yuvipanda I've requested your review since you added the two fixme notes that I'm removing now. I think maybe we are in agreement based on insights gained over the last two years that this isn't something we want to go for.

If you think the fixme ideas are still relevant to track, I think they could need to be outlined in more detail as part of a github issue - what do you think?


yuvipanda (approved):
I think this is good!

PR #2150, opened 2023-02-06
Updating the Openscapes hub

betolink:
Updating our Python and R images, we already tested them on staging. The full Conda environment is attached for future reference.

[Export Full Conda Environment.txt](https://github.com/2i2c-org/infrastructure/files/10667976/9_Export.Full.Conda.Environment.txt)


yuvipanda:
@betolink do you want to keep the automatic updater in place, or would you prefer the manual setup so you have better control?

PR #2135, opened 2023-02-03
Mount /home/jovyan/allusers for LEAP hub

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/2105

PR #2151, opened 2023-02-06
leap: make allusers mount read only

consideRatio:
As requested in https://2i2c.freshdesk.com/a/tickets/387 by @jbusecke, following #2135.


jbusecke:
thanks a lot

Issue #2123, opened 2023-02-01
Setup a publicly accessible bucket that hub users can write to


consideRatio:
We have agreed to setup a new public bucket for m2lines prod hub.

### Action point

- [ ] Setup a public (persistent) bucket for m2lines prod hub.
- [ ] Communicate back to Pavel about it in https://2i2c.freshdesk.com/a/tickets/405, and cc @jbusecke as well who was interested in this work.

### Related support ticket

- https://2i2c.freshdesk.com/a/tickets/405

### Outdated deliberations

From reading at [this blog post's second paragraph](https://gcpfreaks.medium.com/restricting-write-permissions-on-folders-in-google-cloud-storage-with-iam-conditions-e82f36e47639) I conclude it isn't straight forward to provide public access to parts of a bucket - for example under a "folder". The technical workaround strategy described to use GCP "IAM conditions" couldn't be applied either, because those conditions could not be used with the `allUsers` identity representing unauthenticated access.

PR #2159, opened 2023-02-07
Add public scratch buckets support and use it for m2lines

GeorgianaElena:
Ref https://github.com/2i2c-org/infrastructure/issues/2123 and https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_default_object_access_control.html#role

Todo:
- [x] add docs
- [ ]

### DIFF
--- a/config/clusters/m2lines/prod.values.yaml
+++ b/config/clusters/m2lines/prod.values.yaml
@@ -8,7 +8,7 @@ basehub:
         SCRATCH_BUCKET: gs://m2lines-scratch/$(JUPYTERHUB_USER)
         PANGEO_SCRATCH: gs://m2lines-scratch/$(JUPYTERHUB_USER)
         PERSISTENT_BUCKET: gs://m2lines-persistent/$(JUPYTERHUB_USER)
-        PUBLIC_BUCKET: gs://m2lines-public/$(JUPYTERHUB_USER)
+        PERSISTENT_PUBLIC_BUCKET: gs://m2lines-persistent-public/$(JUPYTERHUB_USER)
     hub:
       config:
         GitHubOAuthenticator:
--- a/config/clusters/m2lines/prod.values.yaml
+++ b/config/clusters/m2lines/prod.values.yaml
@@ -8,7 +8,7 @@ basehub:
         SCRATCH_BUCKET: gs://m2lines-scratch/$(JUPYTERHUB_USER)
         PANGEO_SCRATCH: gs://m2lines-scratch/$(JUPYTERHUB_USER)
         PERSISTENT_BUCKET: gs://m2lines-persistent/$(JUPYTERHUB_USER)
-        PERSISTENT_PUBLIC_BUCKET: gs://m2lines-persistent-public/$(JUPYTERHUB_USER)
+        PUBLIC_PERSISTENT_BUCKET: gs://m2lines-public-persistent/$(JUPYTERHUB_USER)
     hub:
       config:
         GitHubOAuthenticator:
--- a/config/clusters/m2lines/prod.values.yaml
+++ b/config/clusters/m2lines/prod.values.yaml
@@ -8,6 +8,7 @@ basehub:
         SCRATCH_BUCKET: gs://m2lines-scratch/$(JUPYTERHUB_USER)
         PANGEO_SCRATCH: gs://m2lines-scratch/$(JUPYTERHUB_USER)
         PERSISTENT_BUCKET: gs://m2lines-persistent/$(JUPYTERHUB_USER)
+        PUBLIC_BUCKET: gs://m2lines-public/$(JUPYTERHUB_USER)
     hub:
       config:
         GitHubOAuthenticator:
--- a/docs/howto/features/buckets.md
+++ b/docs/howto/features/buckets.md
@@ -56,7 +56,7 @@ on why users want this!
             # If we have a bucket that does not have a `delete_after`
             PERSISTENT_BUCKET: <s3 or gs>://<bucket-full-name>/$(JUPYTERHUB_USER)
             # If we have a bucket defined in user_buckets that should be granted public read access.
-            PUBLIC_BUCKET: <s3 or gs>://<bucket-full-name>/$(JUPYTERHUB_USER)
+            PERSISTENT_PUBLIC_BUCKET: <s3 or gs>://<bucket-full-name>/$(JUPYTERHUB_USER)
 
    ```
 
--- a/docs/howto/features/buckets.md
+++ b/docs/howto/features/buckets.md
@@ -32,7 +32,16 @@ on why users want this!
    in the same `.tfvars` file. Follow all the steps listed there - this
    should create the storage buckets and provide all users access to them!
 
-3. You can set the `SCRATCH_BUCKET` (and the deprecated `PANGEO_SCRATCH`)
+3. (If requested) Enable public read access to these buckets by editing the
+   `bucket_public_access` list in the same `.tfvars`:
+
+   ```terraform
+   bucket_public_access = [
+      "persistent-public"
+   ]
+   ```
+
+4. You can set the `SCRATCH_BUCKET` (and the deprecated `PANGEO_SCRATCH`)
    env vars on all user pods so users can use the created bucket without
    having to hard-code the bucket name in their code. In the hub-specific
    `.values.yaml` file in `config/clusters/<cluster-name>`,
@@ -46,6 +55,9 @@ on why users want this!
             PANGEO_SCRATCH: <s3 or gs>://<bucket-full-name>/$(JUPYTERHUB_USER)
             # If we have a bucket that does not have a `delete_after`
             PERSISTENT_BUCKET: <s3 or gs>://<bucket-full-name>/$(JUPYTERHUB_USER)
+            # If we have a bucket defined in user_buckets that should be granted public read access.
+            PUBLIC_BUCKET: <s3 or gs>://<bucket-full-name>/$(JUPYTERHUB_USER)
+
    ```
 
    ```{note}
@@ -67,5 +79,5 @@ on why users want this!
 
    You can also add other env vars pointing to other buckets users requested.
 
-4. Get this change deployed, and users should now be able to use the buckets!
+5. Get this change deployed, and users should now be able to use the buckets!
    Currently running users might have to restart their pods for the change to take effect.
--- a/docs/howto/features/buckets.md
+++ b/docs/howto/features/buckets.md
@@ -37,7 +37,7 @@ on why users want this!
 
    ```terraform
    bucket_public_access = [
-      "persistent-public"
+      "public-persistent"
    ]
    ```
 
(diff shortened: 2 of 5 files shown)


terraform/gcp/buckets.tf
```diff
@@ -38,13 +38,32 @@ locals {
...
+
 resource "google_storage_bucket_iam_member" "member" {
   for_each = { for bp in local.bucket_permissions : "${bp.hub_name}.${bp.bucket_name}" => bp }
   bucket   = google_storage_bucket.user_buckets[each.value.bucket_name].name
   role     = "roles/storage.admin"
   member   = "serviceAccount:${google_service_account.workload_sa[each.value.hub_name].email}"
 }
 
+resource "google_storage_default_object_access_control" "public_rule" {
+  for_each = { for bp in local.public_buckets : "${bp.hub_name}.${bp.bucket_name}" => bp }
```
GeorgianaElena, line 61: I used the same mapping mechanism to get pairs of [hub, public buckets], but I don't think it makes sense to have this in the context of a public bucket.
Or is there a way to enforce a namespace on a scratch bucket?

GeorgianaElena:
terraform plan output:

```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

  # google_storage_bucket.user_buckets["persistent-public"] will be created
  + resource "google_storage_bucket" "user_buckets" {
      + force_destroy               = false
      + id                          = (known after apply)
      + location                    = "US-CENTRAL1"
      + name                        = "m2lines-persistent-public"
      + project                     = "m2lines-hub"
      + public_access_prevention    = (known after apply)
      + self_link                   = (known after apply)
      + storage_class               = "STANDARD"
      + uniform_bucket_level_access = (known after apply)
      + url                         = (known after apply)

      + versioning {
          + enabled = (known after apply)
        }

      + website {
          + main_page_suffix = (known after apply)
          + not_found_page   = (known after apply)
        }
    }

  # google_storage_default_object_access_control.public_rule["prod.persistent-public"] will be created
  + resource "google_storage_default_object_access_control" "public_rule" {
      + bucket       = "m2lines-persistent-public"
      + domain       = (known after apply)
      + email        = (known after apply)
      + entity       = "allUsers"
      + entity_id    = (known after apply)
      + generation   = (known after apply)
      + id           = (known after apply)
      + project_team = (known after apply)
      + role         = "READER"
    }

Plan: 2 to add, 0 to change, 0 to destroy.

Changes to Outputs:
  ~ buckets = {
      + persistent-public  = "m2lines-persistent-public"
        # (4 unchanged elements hidden)
    }

```

terraform/gcp/variables.tf
```diff
@@ -282,7 +282,7 @@ variable "max_cpu" {
 }
 
 variable "hub_cloud_permissions" {
-  type        = map(object({ requestor_pays : bool, bucket_admin_access : set(string), hub_namespace : string }))
+  type        = map(object({ requestor_pays : bool, bucket_admin_access : set(string), public_buckets : set(string), hub_namespace : string }))
```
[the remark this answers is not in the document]
consideRatio: I'm not sure, but I think let's go with this!

terraform/gcp/buckets.tf
consideRatio: I'm new to terraform code, so I'm for actively learning as part of this review.

### References

- Read that [`for_each`](https://developer.hashicorp.com/terraform/language/meta-arguments/for_each) has a special meaning, available for all resources
- I understand that [`bucket` / `role` / `member`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_bucket_iam#google_storage_bucket_iam_member) has resource type (`google_storage_default_object_access_control`) specific meaning.

### Refactoring to avoid complexity

Looking at this resource definition and the one above, I see that `for_each` is used to provide a key and value (key => value), but we never make use of the key (each.key). So, we could switch this to a statement like this instead. Both here and above. That would have made me understand this section faster, and probably myself in the future as well.

```diff
-  for_each = { for bp in local.public_buckets : "${bp.hub_name}.${bp.bucket_name}" => bp }
+  for_each = [ for b in local.public_buckets : b ]
   bucket = google_storage_bucket.user_buckets[each.value.bucket_name].name
   role   = "READER"
   entity = "allUsers"
 }
```

I think actually this may work also.

```diff
-  for_each = { for bp in local.public_buckets : "${bp.hub_name}.${bp.bucket_name}" => bp }
+  for_each = local.public_buckets
   bucket = google_storage_bucket.user_buckets[each.value.bucket_name].name
   role   = "READER"
   entity = "allUsers"
 }
```

- [ ] Refactor for readability like above, do it also for the other resource for consistency.

### Namespacing

It doesn't make sense for the `google_storage_default_object_access_control` resource to consider namespace. The purpose is to declare public readability.

On the other hand, we still need to provide hub users access to write to the bucket which is declared in the `google_storage_bucket_iam_member` above. Looking at that, I conclude that permission is only granted to a namespaced account for buckets listed in `bucket_admin_access` - where the public bucket isn't listed!

So, the public bucket still needs to be provisioned write access via namespaced IAM accounts (coupled to a k8s ServiceAccount via google workload-identity).

- [ ] Add the public bucket name to the `bucket_admin_access` list for m2lines

terraform/gcp/variables.tf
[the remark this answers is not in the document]
consideRatio: What do you think about renaming this to align with previous naming?

```diff
   "prod" : {
     requestor_pays : true,
     bucket_admin_access: ["scratch", "persistent"],
-    public_buckets: ["persistent-public"],
+    bucket_public_access: ["persistent-public"],
     hub_namespace: "prod"
   },
```

consideRatio (commented):
Thank you for your work on this @GeorgianaElena! This was clearly a more work than I thought initially!

GeorgianaElena:
@consideRatio, thanks a lot for the feedback. I've just updated this PR with:

- made `bucket_public_access` be its own terraform variable (a list holding the bucket names), instead of having it be a field of `hub_cloud_permissions`
   - this simplified the config
   - and since it isn't linked to a namespace, I think having it separate from the hub cloud permissions is less confusing.
-  added the public bucket name to the bucket_admin_access list for m2lines
- updated the terraform plan output above with the new changes

terraform/gcp/variables.tf
```diff
@@ -298,6 +298,15 @@ variable "hub_cloud_permissions" {
   EOT
 }
 
+variable "bucket_public_access" {
+  type        = list
+  default     = []
+  description = <<-EOT
+  A list of GCS storage buckets with public read access.
```
consideRatio, line 305:
```suggestion
  A list of GCS storage buckets defined in user_buckets that should be granted public read access.
```

consideRatio (approved):
Thank you @GeorgianaElena!!

Will you follow up at https://2i2c.freshdesk.com/a/tickets/405 with Pavel as well?

---

Oh I forgot, I think we should update this also!

https://github.com/2i2c-org/infrastructure/blob/45db56a031469da7982c11fd288009539d77a3c9/config/clusters/m2lines/prod.values.yaml#L6-L10

### Naming

What do you think about skipping `persistent` in the naming, and just go with `m2lines-public` as a name and `PUBLIC_BUCKET` env var?

GeorgianaElena:
> What do you think about skipping persistent in the naming, and just go with m2lines-public as a name and PUBLIC_BUCKET env var?

Hmm, I think it makes sense to be explicit that the bucket was given a persistent nature, esp since we can also have (if requested) buckets that are not persistent, but public, right>?

docs/howto/features/buckets.md
```diff
@@ -32,7 +32,16 @@ on why users want this!
    in the same `.tfvars` file. Follow all the steps listed there - this
    should create the storage buckets and provide all users access to them!
 
-3. You can set the `SCRATCH_BUCKET` (and the deprecated `PANGEO_SCRATCH`)
+3. (If requested) Enable public read access to these buckets by editing the
+   `bucket_public_access` list in the same `.tfvars`:
+
+   ```terraform
+   bucket_public_access = [
+      "persistent-public"
```
consideRatio, line 40: To match `PUBLIC_BUCKET: gs://m2lines-public/$(JUPYTERHUB_USER)`, this would need to be like this I think:

```suggestion
      "public"
```

I think I'd favor the more brief `public` overall.

GeorgianaElena: @consideRatio, if we were to have public buckets that aren't persistent, woudn't be better to be explicit about this?
My motivation is that there can be buckets that are:
- scratch
- persistent 
- public + scratch
- public + persistent

I'm now thinking if it shouldn't in fact be the other way around, i.e. have the env var match this composed pattern, i.e. a PUBLIC_PERSISTENT_BUCKET and a PUBLIC_SCRATCH_BUCKET? 😅

sgibson91: I agree with Georgiana. Better to be explicit now than have someone complain that their data was wiped, or made public when it shouldn't, because of unclear naming conventions leading to using the inappropriate bucket. Even if we later find out it's not needed, at least we won't find out we made the wrong choice because someone's work was ruined.

(I also agree with the switching the word order, it syntactically reads better :) )

consideRatio: @GeorgianaElena I've updated my comment after you replied to it!

I like the consistency you suggest, but i guess just like how we omit `private` or `local` etc for the non-public buckets, I was thinking we may want to not declare `persistent` because its kind of what you would assume - and could be misleading for someone in the public that may think that data URLs are immutable and that is the meaning of persistent.

consideRatio: Okay seeing @sgibson91's input I think we should not think further and go with what you outlined above @GeorgianaElena!

Thanks for helping out discussing thing @sgibson91!!!

I think we should use an env var name that is directly associated with the name of the bucket, so lets make the env var `PUBLIC_PERSISTENT_BUCKET` then as well - right?

GeorgianaElena:
Thanks a lot for the feedback @consideRatio and @sgibson91. I'm currently applying the changes to the cluster, then I'll merge this, then report back in the ticket.

config/clusters/m2lines/prod.values.yaml
```diff
@@ -8,7 +8,7 @@ basehub:
         SCRATCH_BUCKET: gs://m2lines-scratch/$(JUPYTERHUB_USER)
         PANGEO_SCRATCH: gs://m2lines-scratch/$(JUPYTERHUB_USER)
         PERSISTENT_BUCKET: gs://m2lines-persistent/$(JUPYTERHUB_USER)
-        PUBLIC_BUCKET: gs://m2lines-public/$(JUPYTERHUB_USER)
+        PERSISTENT_PUBLIC_BUCKET: gs://m2lines-persistent-public/$(JUPYTERHUB_USER)
```
consideRatio: @GeorgianaElena this is the wrong way around, right? Wasnt it public-persistent?

consideRatio: Oh, i was linked to an old commit. Ignore this!

Issue #2156, opened 2023-02-07
Now we use `notebook` - we need to allow for `jupyter_server`


consideRatio:
In [a slack thread](https://2i2c.slack.com/archives/CKJS000F4/p1675062390719029) @yuvipanda clarified that we are starting all user servers using the classical `notebook` server by configuring this ([see JupyterHub docs](https://jupyterhub.readthedocs.io/en/stable/reference/config-user-env.html#switching-back-to-the-classic-notebook)):

https://github.com/2i2c-org/infrastructure/blob/a9d8816bb423cb3117ef94935f303e4afd1cc9ee/helm-charts/basehub/values.yaml#L181-L186

In [this support ticket from the `temple` hub](https://2i2c.freshdesk.com/a/tickets/423) we conclude that a modern `jupyter/datascience-notebook` image won't work with JupyterLab any more, where the notebook server presents 404 Not found when accessing `/lab`, because of this:

```
[I 2023-02-07 06:17:06.501 SingleUserNotebookApp mixins:609] Starting jupyterhub single-user server version 3.1.1
[I 2023-02-07 06:17:06.502 SingleUserNotebookApp mixins:623] Extending notebook.notebookapp.NotebookApp from notebook 6.5.2

# ...

[W 2023-02-07 06:17:07.775 SingleUserNotebookApp serverextension:13] Loading JupyterLab as a classic notebook (v6) extension.
[C 2023-02-07 06:17:07.775 SingleUserNotebookApp serverextension:18] You must use Jupyter Server v1 to load JupyterLab as notebook extension. You have v2.2.1 installed.
    You can fix this by executing:
        pip install -U "jupyter-server<2.0.0"
```

## What to do

__What we should do no matter what__

- We should overview all things we do that is `notebook` or `jupyter_server` specific.
- We should plan what we to communicate to users managing user environment images

__Four strategies to consider__

1. We can ask users to pin `jupyter_server<2`, but that will come back and haunt 
us when we need them to have `jupyter_server>=2` now or because they request features that only v2 supports.
2. We can start overriding to `JUPYTERHUB_SINGLEUSER_APP=jupyter_server.serverapp.ServerApp` for individual hubs, and along with that also updating configuration of `notebook` classic to configuration of `jupyter_server`
3. We can start overriding to `JUPYTERHUB_SINGLEUSER_APP=` for individual hubs, and along with that make sure to provide configuration to both `notebook` classic and `jupyter_server`
[...]

PR #2160, opened 2023-02-07
Stop forcing use of NotebookApp

yuvipanda:
This was a temporary fix as a PR landed in JupyterHub & we tried to upgrade JupyterHub in-image everywhere. As this just needs to be JupyterHub 2.3.1 to fix the bug we were guarding against, I think we're safe to get rid of this now.

The specific issue we were guarding against was that selecting 'rstudio' as the default URL will not work - or in general, setting the default URL won't work, and everyone will just land in JupyterLab. If we can keep an eye out for that, we're good. If we do run into that, we can both upgrade JupyterHub package in their image, or bring back this env var temporarily.

Fixes https://github.com/2i2c-org/infrastructure/issues/2156

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -179,11 +179,6 @@ jupyterhub:
       # https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2449
       - jupyterhub-singleuser
     extraEnv:
-      # until https://github.com/jupyterhub/jupyterhub/pull/3918 or equivalent lands,
-      # and we upgrade to jupyterhub >= 2.3.1 on all images.
-      # Note: please checkout all occurences of `NotebookApp` configurations
-      # when moving away from it.
-      JUPYTERHUB_SINGLEUSER_APP: "notebook.notebookapp.NotebookApp"
       # notebook writes secure files that don't need to survive a
       # restart here. Writing 'secure' files on some file systems (like
       # Azure Files with SMB) seems buggy, so we just put runtime dir on
@@ -220,9 +215,7 @@ jupyterhub:
             # this is what allows us to shutdown servers
             # when people leave a notebook open and wander off
             cull_connected: true
-          # If we switch from setting `JUPYTERHUB_SINGLEUSER_APP` to be the NotebookApp
-          # this config kere might need to go under ServerApp instead
-          NotebookApp:
+          ServerApp:
             extra_template_paths:
               - /usr/local/share/jupyter/custom_template
     startTimeout: 600 # 10 mins, because sometimes we have too many new nodes coming up together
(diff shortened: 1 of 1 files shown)


consideRatio:
Thanks @yuvipanda for working this, I think the changes looks good. I think the change may very likeley cause things to break, so I'm reviewing this quite carefully.

Below I've considered:
- what we can relatively quickly test manually
- additional configuration for jupyter_server
- possible exceptions for hubs to stick with NotebookApp
- preparing support stewards to understand if reported issues likely is or isn't related to this change and what to do

### Practical checks

- [ ] Does default_url work for notebook/lab/rstudio in up to date image?

### Understanding check

- If users report default url issues, we should expect issues with old versions of jupyterhub in the user image and ask them to upgrade to a new version - right?
- Without `JUPYTERHUB_SINGLEUSER_APP` specified, what server software will be started? If we can understand this, we can better understand if this change could be the cause of something malfunctioning or not.
  If we don't understand this, we may consider this change to have caused something unrelated etc.
  I think `JUPYTERHUB_SINGLEUSER_APP` influences `jupyterhub-singleuser`, which in turn is coupled to the `jupyterhub` version's logic.

### Considering used server extensions

Will we break use of `notebook` or `jupyter_server` extensions configured by either us or users?

- We captured references to `NotebookApp` and `ServerApp`, but I see hardcoded examples we provide related to `jupyter_notebook_config.[py|json]` and not `jupyter_server_config.[py|json]`.
  - I think we may need to provide an equivalent `jupyter_server_config.[py.json]` file, but I don't know. Will jupyter_server v1 and/or v2 read config from `jupyter_notebook_config` files ?

- What extensions do we know are used in various communities, are they compatible with either and configured for either?
  - What about the communities with various graders? Are those server extensions?
  - Do we expect nbgitpuller to keep working well?
  - I think scottyhq/jupyter-cloud-scoped-creds only works with `notebook` currently, but its an easy fix upstream and this breaking is probably acceptable.

yuvipanda:
@consideRatio without this set, jupyter_server is used by default. Crucially, I think jupyter_server was the default running for a month or two before we realized that the default_url wasn't working, and I explicitly reverted us back to notebookapp. So I *think* things should be ok to switch back, as https://github.com/jupyterhub/jupyterhub/pull/3918  was fixed.

yuvipanda:
I think the newest JupyterLab no longer works with this set, so we are pretty much forced to unset this now.

yuvipanda:
This can be overriden on a per-hub basis, as is done for https://github.com/2i2c-org/infrastructure/pull/2183

consideRatio:
What about the culling configuration etc? It seems to me that if we provide a file that is named `jupyter_notebook_config.json` it won't influence the kernel manager's culler - right?

consideRatio (approved):
I'm approving because I'm afraid of the consequences of either choice and you lean towards this is good to go.

If I find time, I'll try dig into this because otherwise I'll always have it as a open questionmark going onwards. Did this break the kernel culling? Is this the reason some server extension no longer works used for ottergrader or some other software I'm not read up on?

consideRatio (approved):
I think I understand things quite a bit better now, and I feel quite comfortable with a merge now.

yuvipanda:
Thanks @consideRatio

PR #2161, opened 2023-02-07
openscapes: use of r5 machines with a profile list allowing for node sharing

consideRatio:
I've already deployed things to openscapes hub manually, so if this isn't merged and a dummy change happens - those changes are reset.

image

### Notes communicated in [a freshdesk ticket](https://2i2c.freshdesk.com/a/tickets/382)

**Small / Medium / Large**
Decides the kind of node (machine type) the user's server container will be started on. Even if multiple users start on the same node, they are isolated as they they run code within their containers.

**Node share**
Decides the amount of capacity is guaranteed  for the user's server container, but its not the upper bound. A user running on a machine alone can make use of almost all of the CPU / RAM capacity if other users aren't using it at that very moment. The typical user is often idling at 0 CPU, so anyone not idling can often make use of far more CPU.

The default choice is 1 GB for Small, 4 GB for medium, and 16 GB for large. This makes 32 users fit on each node type if users only use the default node share option for respective node size.


eksctl/openscapes.jsonnet
```diff
@@ -92,12 +90,12 @@ local daskNodes = [
...
     nodeGroups: [n + {clusterName:: $.metadata.name} for n in [
         ng {
-            name: 'core-a',
+            name: 'core-b',
             availabilityZones: [nodeAz],
             ssh: {
                 publicKeyPath: 'ssh-keys/openscapes.key.pub'
             },
-            instanceType: "m5.xlarge",
+            instanceType: "r5.xlarge",
```
yuvipanda, line 98: May I ask why the core node size change too? Aren't r5 nodes more expensive than m5? Does it result in a reduction in total number of nodes? If not, I'd love for this to still be m5 to be consistent with other clusters.

yuvipanda (commented):
Everything else looks good, but one question

eksctl/openscapes.jsonnet
consideRatio: We have clusters with multiple core nodes, and I think its related to a constraint memory more than anything else, where prometheus-server is a big complexity biting us with its extreme memory footprint during startup. I've seen a cluster with 4 core nodes for example.

Maybe getting many core nodes relates to user pods scheduling on core nodes (#2163), I'm not sure.

I'm thinking quite confidently at the moment this is a cost saver due to prometheus-server making us memory constrained overall. The cost of a m5 vs r5 is $0.192 vs $0.252, a ratio of 1.3125 or invesively 0.7619. This is ~4:3 in CPU per $ and 4:8 in GB Ram per $ when comparing m5:r5.

I'll go for a merge here as this reflects the current state @yuvipanda! Do you think its okay to stick with r5 for the core node pool or should i change it back?

yuvipanda: @consideRatio I think it's ok for now! What I'd love is for us to evaluate what nodes actually make sense for core nodes and make that uniform :) The current setup again is just a guess.

yuvipanda (approved):

eksctl/openscapes.jsonnet
yuvipanda: @consideRatio I've approved!

PR #2183, opened 2023-02-08
Update LEAP hub images to latest pangeo image

yuvipanda:
This brings in a new dask-gateway version to match what we have in the chart, which might help with some of the issues the LEAP users have been facing.

This also brings in a newer version of JupyterLab, which requires we stop using NotebookApp. This overrides it for this specific hub until
https://github.com/2i2c-org/infrastructure/pull/2160 is merged.

Ref https://2i2c.freshdesk.com/a/tickets/433

### DIFF
--- a/config/clusters/leap/common.values.yaml
+++ b/config/clusters/leap/common.values.yaml
@@ -17,7 +17,7 @@ basehub:
       extraImages:
         tensorflow-image:
           name: pangeo/ml-notebook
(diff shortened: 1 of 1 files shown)


yuvipanda:
This has unblocked @jbusecke, so I'm going to just merge this now so it doesn't accidentally get undone. Hopefully the need to merge without review will be reduced going forward as our engineering capacity grows!

PR #2194, opened 2023-02-10
Force user pods to schedule on user nodes, enable user-scheduler where needed

### LINKED ISSUE #2163: Force user pods to schedule on user nodes (not next to hub pod etc)
I think [`scheduling.userPods.nodeAffinity.matchNodePurpose=require`](https://z2jh.jupyter.org/en/stable/resources/reference.html#scheduling-userpods-nodeaffinity-matchnodepurpose) is a jupyterhub configuration we should go for by default. It makes user pods get a required nodeAffinity (like `nodeSelector` but separate) forcing them to schedule on user nodes so they can't end up at core nodes - as they can now.

```yaml
  jupyterhub:
    scheduling:
      userPods:
        nodeAffinity:
          matchNodePurpose: require
```

That they tolerate the tainted user nodes makes them prefer scheduling there, but they can schedule on the core node pool as well without this if there is no space on an already available user node.

If this is enabled, and we don't have the proper taints on the user nodes - we will break things. But this should be caught by our health check that tests start a user pod!

### LINKED ISSUE #2190: Ensure user-scheduler is active on AWS clusters
GCP GKE based clusters have a scheduler setting allowing for `OPTIMIZE_UTILIZATION` to make pods schedule on the most used nodes. This allows other nodes to become empty and scaled down.

On AWS, this is not an option and the JupyterHub chart's user-scheduler is meant to handle that. The user-scheduler is disabled in the basehub helm chart though, so unless AWS based charts opts into it, we don't get to make use of it.

I observed this when manually ensuring there was many k8s nodes started for an openscapes event. In practice, its not so problematic but we will waste a bit of cloud resources and drive cost more than needed when using shared nodes like openscapes does.

image
_NOTE: I scaled up nodes after the red node had started filling up. But since it was the only node for a while it started filling up. When all the other nodes became ready, new users arriving started spreading out between them because we didn't have user-scheduler activated._

### Related

- https://github.com/2i2c-org/infrastructure/pull/1026

consideRatio:
Since this PR touches basehub and the jupyterhub helm chart, it means a lot of installations will have a hub pod that restarts I think. Due to that I've bundled a few changes to minimize disruption of our hubs.

- Closes #2163
- Closes #2190

## Changes by commit

1. **basehub: fix node selection using z2jh config option**
   This is a bugfix and complexity reduction by relying on the jupyterhub chart's functionality.
   
   Instead of specifying `nodeSelector` on various jupyterhub chart pods and user pods to schedule on "core" and "user" nodes, we let the jupyterhub chart apply a `nodeAffinity` that works the same way via [`scheduling.corePods.matchNodePurpose=require`](https://z2jh.jupyter.org/en/stable/resources/reference.html#scheduling-corepods-nodeaffinity-matchnodepurpose).
   
   This way, we avoid a bug where we override the user pods' `nodeSelector` via KubeSpawner's profile_list choices, which can make user nodes schedule on core nodes for example.
2. **basehub: avoid re-specification of z2jh default values**
   Just a cleanup to reduce complexity, and generally avoid issues by following z2jh defaults in case they change.
3. **Add link in cluster.yaml to reach cloud console**
   Makes it easier to find the clusters associated cloud console, which is a generally useful shortcut.
4. **Enable user-scheduler on all non-gcp hubs**
   This is a bugfix as I see it. We intend for this behavior overall, but it was disabled for all clusters when enabled out of the box on GKE clusters. Without this, we fail to scale down efficiently on AWS/Azure clusters and using shared nodes.

   The user-scheduler is a jupyterhub chart provided kubernetes pod scheduler to help get user pods scheduled on the most busy node. Like this, we improve scale down significantly when using shared nodes. This is enabled only for AWS/Azure hubs, and not on GCP hubs, as there we activate the same feature from GKE out of the box.
5. **Let user-placeholder's statefulset be around with 0 replicas**
   The user-placeholder resource with 0 replicas isn't harmful to have around, but ensures it becomes an option if we want to use it.

### DIFF
--- a/config/clusters/2i2c-aws-us/cluster.yaml
+++ b/config/clusters/2i2c-aws-us/cluster.yaml
@@ -1,5 +1,5 @@
 name: 2i2c-aws-us
-provider: aws
+provider: aws # https://2i2c.awsapps.com/start#/
 aws:
   key: enc-deployer-credentials.secret.json
   clusterType: eks
--- a/config/clusters/2i2c-aws-us/cluster.yaml
+++ b/config/clusters/2i2c-aws-us/cluster.yaml
@@ -26,6 +26,7 @@ hubs:
     auth0:
       enabled: false
     helm_chart_values_files:
+      - common.values.yaml
       - dask-staging.values.yaml
       - enc-dask-staging.secret.values.yaml
   - name: researchdelight
@@ -35,4 +36,5 @@ hubs:
     auth0:
       connection: github
     helm_chart_values_files:
+      - common.values.yaml
       - researchdelight.values.yaml
--- a/config/clusters/2i2c-uk/cluster.yaml
+++ b/config/clusters/2i2c-uk/cluster.yaml
@@ -1,5 +1,5 @@
 name: 2i2c-uk
-provider: gcp
+provider: gcp # https://console.cloud.google.com/kubernetes/clusters/details/europe-west2/two-eye-two-see-uk-cluster/nodes?authuser=3&project=two-eye-two-see-uk
 gcp:
   key: enc-deployer-credentials.secret.json
   project: two-eye-two-see-uk
--- a/config/clusters/2i2c-uk/cluster.yaml
+++ b/config/clusters/2i2c-uk/cluster.yaml
@@ -1,5 +1,5 @@
 name: 2i2c-uk
-provider: gcp # https://console.cloud.google.com/kubernetes/clusters/details/europe-west2/two-eye-two-see-uk-cluster/nodes?authuser=3&project=two-eye-two-see-uk
+provider: gcp # https://console.cloud.google.com/kubernetes/clusters/details/europe-west2/two-eye-two-see-uk-cluster/nodes?project=two-eye-two-see-uk
 gcp:
   key: enc-deployer-credentials.secret.json
   project: two-eye-two-see-uk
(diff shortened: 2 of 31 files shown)


config/clusters/2i2c-aws-us/cluster.yaml
```diff
@@ -1,5 +1,5 @@
 name: 2i2c-aws-us
-provider: aws
+provider: aws # https://2i2c.awsapps.com/start#/
```
pnasrat, line 2: Perhaps not necessary for this commit but might be worth in the future it might be useful to move the console link from a commento be a defined variable in the cluster.yaml, meaning that it can be extracted programatically (maybe exposed through something like `deployer cluster-console $CLUSTER`) to help oncall flow as the number of consoles and providers grows.

consideRatio: @pnasrat good idea, opened #2198 about it!

config/clusters/2i2c-uk/cluster.yaml
```diff
@@ -1,5 +1,5 @@
 name: 2i2c-uk
-provider: gcp
+provider: gcp # https://console.cloud.google.com/kubernetes/clusters/details/europe-west2/two-eye-two-see-uk-cluster/nodes?authuser=3&project=two-eye-two-see-uk
```
yuvipanda, line 2:
```suggestion
provider: gcp # https://console.cloud.google.com/kubernetes/clusters/details/europe-west2/two-eye-two-see-uk-cluster/nodes?project=two-eye-two-see-uk
```

yuvipanda (approved):
Thanks Erik! I left one suggestion about removing an extra query param

consideRatio:
Thank you for the review @pnasrat and @yuvipanda!!

PR #2197, opened 2023-02-13
Make cilogon client manager a deployer subcommand

GeorgianaElena:
Part of https://github.com/2i2c-org/infrastructure/issues/1472

Todo:
- [x] update the documentation where this is referenced with new syntax

### DIFF
--- a/deployer/cilogon_app.py
+++ b/deployer/cilogon_app.py
@@ -109,7 +109,9 @@ class CILogonAdmin:
             print_colour(f"{response.text}", "yellow")
             return
 
-        print_colour(f"Successfully got the details for {id} client!")
+        if id:
+            print_colour(f"Successfully got the details for {id} client!")
+
         return response.json()
 
     def update(self, id, body):
--- a/deployer/cilogon_app.py
+++ b/deployer/cilogon_app.py
@@ -81,9 +81,9 @@ class CILogonAdmin:
 
         client_name = body["client_name"]
 
-        if response.status_code != 200:
+        if not response.ok:
             print(
-                f"An error occured when creating the {client_name} client. \n Error was {response.text}."
+                f"An error occured when creating the {client_name} client. \n Error was {response.text}"
             )
             response.raise_for_status()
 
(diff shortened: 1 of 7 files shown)


deployer/cilogon_app.py
```diff
@@ -133,13 +134,16 @@ def update(self, id, body):
...
-            )
-            response.raise_for_status()
+        if not response.ok:
+            if not response.ok:
+                print_colour(
+                    f"Updating the details of {id} client returned a {response.status_code} status code.",
+                    "red",
+                )
+                print_colour(f"{response.text}", "yellow")
+                return
```
consideRatio, line 144: Typo with double if statements

deployer/cilogon_app.py
```diff
@@ -148,25 +152,27 @@ def delete(self, id):
         Args:
            id (str): Id of the client to delete
 
-        Returns 200 OK or raises an error if the delete request returned anything other than 200.
+        Returns status code if response.ok or raises an error if the delete request returned anything not in the range 200-299.
```
consideRatio, line 155: It doesn't raise an error in the python sense, but returns None and prints information about it.

My expectation is that we would somewhere after printing an error message do `sys.exit(1)` to make the exit with a non-0 status code.

This it not critical, and if would require systematic updates, then let's not burden this PR with that.

consideRatio (approved):
Nice @GeorgianaElena!!!

I think we could benefit from having subgroups of command, for example to handle cilogin-clients (get/update/delete etc). These seemed to belong together, so having a sub-command parser of some kind could make sense here.

I think this is an improvement and that its its better we get things merged and iterate in future PRs overall, so :+1: for merge and only adjusting to input if you think it makes sense and feel like it!

deployer/cilogon_app.py
GeorgianaElena: Good point! I've updated the comment to match what's actually happening. Before this PR, it used to do `raise_for_status()`, but I removed it, because it just cluttered the screen with an unuseful error trace, when what we want really is the reason + status code. I believe it's nicer this way, i.e. letting the program to exist gracefully and just print a custom message with details about the error.

PR #2200, opened 2023-02-13
linked-earth: switch to e2-highmem instances and shared nodes

consideRatio:
Already deployed, works!

### DIFF
--- a/config/clusters/linked-earth/common.values.yaml
+++ b/config/clusters/linked-earth/common.values.yaml
@@ -53,40 +53,127 @@ basehub:
         # User image repo: https://quay.io/repository/linkedearth/pyleoclim
         name: quay.io/linkedearth/pyleoclim
         tag: "latest"
+
+      # FIXME: extraFiles and extraEnv has been updated since the combination of
+      #        jupyter_server 2 and jupyterlab doesn't work with a classical
+      #        notebook server.
+      #
+      #        This can probably be removed with this PR that is meant to transition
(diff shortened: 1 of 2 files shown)


yuvipanda:
@consideRatio at least last time I checked, I think the e2- nodes are actually slightly more expensive because they don't have committed use discounts (https://github.com/berkeley-dsep-infra/datahub/issues/1354). Do you know if that has changed?

yuvipanda (approved):
I don't think the e2- question should block this PR!

consideRatio:
> @consideRatio at least last time I checked, I think the e2- nodes are actually slightly more expensive because they don't have committed use discounts (https://github.com/berkeley-dsep-infra/datahub/issues/1354). Do you know if that has changed?

It seems [sustained use discounts are not provided for E2](https://cloud.google.com/compute/docs/sustained-use-discounts), which is what we may benefit from, as compared to comitted use. Sustained use is especially relevant for the core nodes I guess.

Doh!

yuvipanda:
@consideRatio yep! sustained use discounts is what I meant. So without that, e2- nodes actually end up slightly more expensive for us, not just for core nodes but for *all* kinds of nodes - as sustained use discount is not on a per-VM basis but a per-type-of-VM basis, based on my understanding of https://cloud.google.com/compute/docs/sustained-use-discounts. So it is benificial to both notebook and core nodes pretty heavily.

PR #2202, opened 2023-02-14
Disable dex / staticwebsite in staging hub

yuvipanda:
This seems broken, and is blocking deployment to anything on the 2i2c hub :( Disabling it to unblocko deployments.

### DIFF
--- a/config/clusters/2i2c/staging.values.yaml
+++ b/config/clusters/2i2c/staging.values.yaml
@@ -1,9 +1,9 @@
 dex:
-  enabled: true
+  enabled: false
   hubHostName: staging.2i2c.cloud
 
 staticWebsite:
-  enabled: true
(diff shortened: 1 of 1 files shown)


yuvipanda:
See deployment failures in https://github.com/2i2c-org/infrastructure/actions/runs/4168960897 and https://github.com/2i2c-org/infrastructure/actions/runs/4168594911.

The primary culprit seems to be that the hub can not connect to the services to add them to the proxy. My suspicion is something about networkpolicies and how enforcement of those changed with the recent k8s upgrade.

GeorgianaElena (approved):
Let's merge this to unblock deploy. Opened https://github.com/2i2c-org/infrastructure/issues/2206 to track the issue the behing

PR #2224, opened 2023-02-17
Fix broken sops link in grafana docs

pnasrat:


consideRatio (approved):
LGTM!

I'm merging directly since this is a documentation pr that won't trigger CD, while if it was a PR influencing a deployment the practice we have is to delegated it to the PR author to do typically!

PR #2225, opened 2023-02-17
Fix typo causing openscapes users to be under resourced

yuvipanda:
Without the G, we were only guaranteeing them these values in bytes!
And all the CPU guarantees were off by one order of magnitude.

### DIFF
--- a/config/clusters/openscapes/common.values.yaml
+++ b/config/clusters/openscapes/common.values.yaml
@@ -118,43 +118,43 @@ basehub:
                 mem_1:
                   display_name: ~1 GB, ~0.125 CPU
                   kubespawner_override:
-                    mem_guarantee: 0.969
+                    mem_guarantee: 0.969G
                     cpu_guarantee: 0.013
                 mem_2:
                   display_name: ~2 GB, ~0.25 CPU
                   kubespawner_override:
-                    mem_guarantee: 1.938
+                    mem_guarantee: 1.938G
                     cpu_guarantee: 0.025
                 mem_4:
                   default: true
                   display_name: ~4 GB, ~0.5 CPU
                   kubespawner_override:
-                    mem_guarantee: 3.875
+                    mem_guarantee: 3.875G
                     cpu_guarantee: 0.05
                 mem_8:
                   display_name: ~8 GB, ~1.0 CPU
                   kubespawner_override:
-                    mem_guarantee: 7.75
+                    mem_guarantee: 7.75G
                     cpu_guarantee: 0.1
                 mem_16:
                   display_name: ~16 GB, ~2.0 CPU
                   kubespawner_override:
-                    mem_guarantee: 15.5
+                    mem_guarantee: 15.5G
                     cpu_guarantee: 0.2
                 mem_32:
                   display_name: ~32 GB, ~4.0 CPU
                   kubespawner_override:
-                    mem_guarantee: 31.0
+                    mem_guarantee: 31.0G
                     cpu_guarantee: 0.4
                 mem_64:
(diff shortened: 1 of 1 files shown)


pnasrat (approved):
LGTM

Also who needs more than 0.969 bytes anyway

consideRatio:
Ah :) I've realized this and cordoned nodes to ensure they are not running out of capacity.

There is also a need for this fix on linked-earth's hub. Let's not merge it before the event they have right now is over!

yuvipanda:
@consideRatio i've already deployed this - they reached out to me on the openscapes slack (i am involved there in a few different capacities), as their servers were all slow right now! Deploying this fixed it. I'm going to merge this now so it doesn't get undone by another deploy.

Can I leave the linked-earth fix to you? :D

consideRatio (changes requested):
@yuvipanda the thing I want fixed is addition of a G, but not a change in cpu-guarantee. I've not yet documented this, but its intentional!

consideRatio:
I cancelled this workflow to avoid a hub restart during their event!

I'll fix things and can clarify why i want a systematic 10th of the actual CPU designated for the user and such as part of piloting node sharing later.

yuvipanda:
@consideRatio I'd love to read that! Nobody was getting 16 CPUs anywhere and everyone (including me) was pretty confused, and stuff was running way slower than they expected. Sorry if that happened in another PR and I missed that.

I also don't think the hub would have restarted in this case, as this would've been a no-op deploy.

consideRatio:
> Nobody was getting 16 CPUs anywhere and everyone (including me) was pretty confused, and stuff was running way slower than they expected. 

Yikes is things running slower than expected? My understanding is that they are ~30 users on 64 CPU nodes, and should be able to go to 64 cores if other users idle, but if they all go full throttle should get 2 each by having the same cpu-share configured via requests.

Are they speaking to you from another channel besides support tickets?

yuvipanda:
@consideRatio unfortunately yes, this was on the openscapes slack. The issue seems to be that the description says '16 CPUs' but then the guarantee was for 1.6 CPUs only. I didn't want to bounce them to the support channel at that time as this was happening while a presentation was going on and the presenter was struggling with speed. I'll chat about that once their event is over. I'll also share those messages with you on slack.

consideRatio:
> The issue seems to be that the description says '16 CPUs' but then the guarantee was for 1.6 CPUs only

If they pick something with 16 CPU guarantee, they should not arrive at a node being outcompeted to use less than 16 CPU by anything else in the current setup.

yuvipanda:
ok i'm super confused, especially as I see no limits. I'll just wait for a longer explanation later! :)

consideRatio:
You get little CPU if you are limited, or if you fail to compete against other containers with a cpu share far higher than yourself. That is my understanding. Without limit, the only way they would be slow CPU wise would be for being outcompeted - which they won't be if all users have the same reduction of CPU capacity.

I'll try to verify my understanding later, but for now I'll head off work!

consideRatio:
https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource/#if-you-do-not-specify-a-cpu-limit

Maybe there was a automatic limit assigned? Argh I can't pull away so easily from a mystery but I have to for now.

Thanks for helping openscapes people out Yuvi!

consideRatio:
I did an investigation in #2228, and I don't think for example `distributed` would be negatively impacted by a low cpu request, as it doesn't seem to consider it - but only the nodes capacity and possible cpu limits.

I don't know what software was used, but its technically possible that software could self-hamper itself based on cpu requests rather than the node's capacity or cpu limits. I suspect most software won't do that though.

PR #2227, opened 2023-02-18
Bump images for LEAP & m2lines hubs

yuvipanda:
Also consistently provide same image options for all node options.

Brings in https://github.com/pangeo-data/pangeo-docker-images/pull/443

Ref https://2i2c.freshdesk.com/a/tickets/458

### DIFF
--- a/config/clusters/leap/common.values.yaml
+++ b/config/clusters/leap/common.values.yaml
@@ -17,7 +17,7 @@ basehub:
       extraImages:
         tensorflow-image:
           name: pangeo/ml-notebook
-          tag: "2023.02.08"
+          tag: "ebeb9dd"
     custom:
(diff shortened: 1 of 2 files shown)


config/clusters/leap/common.values.yaml
```diff
@@ -150,7 +152,7 @@ basehub:
             mem_guarantee: 52G
             node_selector:
               node.kubernetes.io/instance-type: n1-standard-16
-
+          profile_options: *profile_options
```
consideRatio, line 155: On mobile, but this looks like a yaml error putting options outside a profile list element

consideRatio: On computer now, this seems correct! I got tricked by an inconsistency on placement in relation to other configuration.

consideRatio (approved):

PR #2233, opened 2023-02-20
Update OHW Python image and some metadata

emiliom:
Update Python image, translate profile display names to Spanish, and change funder info to AECID


GeorgianaElena (approved):
Thanks @emiliom

Issue #2173, opened 2023-02-08
Decomission cluster: uwhackweeks


consideRatio:
We no longer have a hub comissioned for the uwhackweeks cluster, so the entire cluster should be cleaned up based on @scottyhq's input in https://github.com/2i2c-org/infrastructure/issues/2031#issuecomment-1398637957 and [a discussion](https://2i2c.slack.com/archives/C01GLCC1VCN/p1675293787696439) with @jmunroe and @colliand I opened to ensure that we not only stop the cloud costs, but also the billing to uwhackweeks.

### Action points

Here are a few of the action points I see for 2i2c's engineering team to take.

- Cleanup all resources in our github repo
- Cleanup all template adjustments possibly made in brances to https://github.com/2i2c-org/default-hub-homepage
- Cleanup misc resources mentioned in the guide on decomissioning a hub
- Confirmation and cleanup of cloud accounts or at least disabling of billing in some way to cut all running costs to zero.

### Related

- #1336
- #2031 

#### Phase III - Cluster Removal

_This phase is only necessary for single hub clusters._

- [x] Run `terraform plan -destroy` and `terraform apply -destroy` from the [appropriate workspace](https://infrastructure.2i2c.org/en/latest/topic/terraform.html#workspaces), to destroy the cluster
- [x] Delete the terraform workspace: `terraform workspace delete <NAME>`
- [x] Remove the associated `config/clusters/<cluster_name>` directory and all its contents
- Remove the cluster from CI:
  - [x] [`deploy-hubs.yaml`](https://github.com/2i2c-org/infrastructure/blob/HEAD/.github/workflows/deploy-hubs.yaml)
  - [x] [`validate-clusters.yaml`](https://github.com/2i2c-org/infrastructure/blob/HEAD/.github/workflows/validate-clusters.yaml)
- [x] Remove the cluster from the list of grafana datasources at https://grafana.pilot.2i2c.cloud/datasources
- [x] Remove A record from Namecheap account

pnasrat:
Additional req: Document the process for cluster decomissioning

consideRatio:
@pnasrat note that this is a AWS based hub and there are be difference between how to tear down AWS / GCP based cloud infra. I suggest narrowing the scope to not consider GCP cloud infra decomissioning.

pnasrat:
```
terraform plan -destroy -var-file=projects/uwhackweeks.tfvars -no-color

No changes. No objects need to be destroyed.
```

[...]

PR #2235, opened 2023-02-20
Decomission uwhackweeks

pnasrat:
See #2173


pnasrat:
I'm going to close this and break into separate commits so I can better debug this failure

PR #2236, opened 2023-02-20
Remove uwhackweeks cluster config

pnasrat:
See #2173


pnasrat:
```
Error: HttpError: API rate limit exceeded for installation ID 10159873.
Error: API rate limit exceeded for installation ID 10159[8](https://github.com/2i2c-org/infrastructure/actions/runs/4225353974/jobs/7337521958#step:4:9)73.
```

pnasrat:
Might need to setup a GITHUB_TOKEN to get higher API rate limit

https://github.com/terraform-linters/setup-tflint#github_token

consideRatio (commented):
EDIT: I think we may need the config still to cleanup the EKS cluster and verify the cleanup, so due to that I write about these steps! Thank you for your work on this @pnasrat!!!!

---

I think this is the first time this is done, and its undocumented on how to do.

Right now, the EKS cluster still lives because its not managed by terraform* still and I think that should be removed first as well using something like `eksctl delete cluster --config-file=...`.

But, how to do this?

1. Setup credentials to the AWS project setup via env vars by visiting the AWS management console for this project (docs are available as part of creating a new AWS based cluster i think in docs.2i2c.org in the infrastructure part)
2. You need to generate a actual eksctl config from the generated eksctl/uwhackweeks.jsonnet template like `jsonnet uwhackweeks.jsonnet > uwhackweeks.eksctl.yaml`
3. You trial doing `eksctl delete cluster --config-file=uwhackweeks.eksctl.yaml`

If this seems to work out, a followup step of relevance I think is to verify that all node pools are cleaned up. There should be "ASG" (auto scaling groups) that you can search for from the AWS management console for example, are those removed after the `eksctl delete cluster` operation?

---

*Terraform wasn't used for this as in the past at least, but maybe no longer as of a month or so ago I think, terraform could only work with "managed node pools" or similar, where EKS couldn't have a managed node pool scale to zero, which was essential for us. So, we've used whats called unmanaged node pools.

pnasrat:
@consideRatio thanks I had already found the `eksctl` docs. My thought here was that removing the cluster config causes the deployer to remove `uptime-monitoring`, which should be done prior to tearing down the actual cluster to prevent accidental pages when removing a cluster.

I'll be putting together a step by step process to improve the docs as part of the overall issue.

consideRatio (approved):
I'm not sure about the uptime checks situation, but do feel free to delete it all. Its part of the git history if we need it for retroactive tear-down anyhow.

Thank you for working this @pnasrat!!!

Issue #2237, opened 2023-02-21
LEAP prod hub unable to start Large profile notebooks


pnasrat:
### Context

Filed through support. 

Get  a ValueError when spawning `Expected option gpu-image for profile large, not found in posted form`

pnasrat:
```pytb
[W 2023-02-21 17:07:50.443 JupyterHub base:1039] 2 consecutive spawns failed.  Hub will exit if failure count reaches 5 before succeeding
[E 2023-02-21 17:07:50.443 JupyterHub gen:630] Exception in Future <Task finished name='Task-439647' coro=<BaseHandler.spawn_single_user.<locals>.finish_user_spawn() done, defined at /usr/local/lib/python3.11/site-packages/jupyterhub/handlers/base.py:963> exception=ValueError('Expected option gpu-image for profile large, not found in posted form')> after timeout
    Traceback (most recent call last):
      File "/usr/local/lib/python3.11/site-packages/tornado/gen.py", line 625, in error_callback
        future.result()
      File "/usr/local/lib/python3.11/site-packages/jupyterhub/handlers/base.py", line 970, in finish_user_spawn
        await spawn_future
      File "/usr/local/lib/python3.11/site-packages/jupyterhub/user.py", line 851, in spawn
        raise e
      File "/usr/local/lib/python3.11/site-packages/jupyterhub/user.py", line 748, in spawn
        url = await gen.with_timeout(timedelta(seconds=spawner.start_timeout), f)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/jovyan/.local/lib/python3.11/site-packages/jupyterhub_configurator/mixins.py", line 45, in start
        return await super().start(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/local/lib/python3.11/site-packages/kubespawner/spawner.py", line 2645, in _start
        await self.load_user_options()
      File "/usr/local/lib/python3.11/site-packages/kubespawner/spawner.py", line 3103, in load_user_options
        await self._load_profile(selected_profile, selected_profile_user_options)
      File "/usr/local/lib/python3.11/site-packages/kubespawner/spawner.py", line 3034, in _load_profile
        raise ValueError(
    ValueError: Expected option gpu-image for profile large, not found in posted form
    
[...]

PR #2238, opened 2023-02-21
LEAP Work around for slug handling and user options in kubespawner

pnasrat:
See #2237


consideRatio (changes requested):
Add `slug: gpu` to the `profile_list` element for `Large + GPU` to workaround the bug https://github.com/jupyterhub/kubespawner/issues/702 for now to avoid making a user facing change.

pnasrat:
Please take another look

consideRatio (approved):
Nice LGTM!

Issue #2245, opened 2023-02-22
Get all sqlite DBs off NFS


yuvipanda:
### Context

Jupyter uses a couple of sqlite DBs:

1. [nbsignatures.db](https://jupyter-notebook.readthedocs.io/en/stable/security.html#the-details-of-trust) for managing trusted notebooks
2. (NEW!) [file_id_manager.db](https://github.com/jupyter-server/jupyter_server_fileid), for handling file ids for real time collaboration.

By default, these end up on the jupyter *data* path (not runtime path), as they (maybe? someone should verify) need to persist between sessions.

However, that menas they end up in `~/.local/share/jupyter`, which is on NFS.

NFS and sqlite kinda hate each other, and generally using them together causes problems that are hard to debug yet inevitable. So we should find a way to get these off NFS.

This was triggered by https://github.com/2i2c-org/infrastructure/issues/2244.

### Proposal

1. Figure out if we need nbsignatures.db to persist across sessions at all in our context (hosted jupyterhub). If we don't, let's put it in /tmp or somewhere else. If we do, let's document this.
2. Figure out if we can turn off the fileid sqlite db unless it is absolutely required for RTC. And even in that case, it can *not* exist on NFS - it uses WAL, which is not supported. So I guess we should just turn this off.

yuvipanda:
https://github.com/jupyter-server/jupyter_server_fileid/issues/60 reported upstream

consideRatio:
I'm unsure if there is an action point on us regarding this to track here, I'm pre-emptively closing this rather than keeping it around pending that.

- [x] We have avoided the issue
- [x] We have provided sufficient comments in our basehub config about how its being avoided

The remaining action point if any as I see it, is to follow https://github.com/jupyter-server/jupyter_server_fileid/issues/60 for updates that makes us think if we want to do something else.

yuvipanda:
I was looking at the logs from the UToronto incident, and see similar messages posted for nbsignatures.db too:

```
[I 2023-02-22 14:27:39.397 SingleUserLabApp contentsmanager:138] Saving Untitled1.ipynb
[E 2023-02-22 14:27:39.407 SingleUserLabApp filemanager:449] Error while saving file: Untitled1.ipynb unable to open database file
    Traceback (most recent call last):
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/services/contents/filemanager.py", line 434, in save
        self.check_and_sign(nb, path)
[...]

Issue #2244, opened 2023-02-22
Failure to save notebooks due to issues with jupyter_server_fileid on jupyterlab >= 3.6


yuvipanda:
Based on report in https://2i2c.freshdesk.com/a/tickets/476, me and @pnasrat had a debugging session trying to figure out why File saving is broken for some users (turned out to be very few!) on the UToronto hub as a result of #2240.

Our conclusions are:

1. JupyterLab 3.6 brings in something called [ydoc](https://github.com/jupyter-server/jupyter_ydoc), related to real time collab
2. This brings in a fileid server extension (that wasn't present in 3.5) - https://github.com/jupyter-server/jupyter_server_fileid
3. This puts file ids on a sqlite db
4. This is on the critical save path - if there are any failures here, no notebooks can be saved
5. And there was failures
```
[W 2023-02-22 14:05:53.096 SingleUserLabApp manager:359] jupyter_server_fileid | extension failed loading with message: unable to open database file
[E 2023-02-22 14:05:53.096 SingleUserLabApp manager:360] jupyter_server_fileid | stack trace
    Traceback (most recent call last):
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/manager.py", line 355, in load_extension
        extension.load_all_points(self.serverapp)
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/manager.py", line 229, in load_all_points
        return [self.load_point(point_name, serverapp) for point_name in self.extension_points]
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/manager.py", line 229, in <listcomp>
        return [self.load_point(point_name, serverapp) for point_name in self.extension_points]
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/manager.py", line 222, in load_point
        return point.load(serverapp)
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/manager.py", line 148, in load
        return loader(serverapp)
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/application.py", line 453, in _load_jupyter_server_extension
        extension.initialize()
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/application.py", line 420, in initialize
        self._prepare_settings()
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/application.py", line 302, in _prepare_settings
        self.initialize_settings()
[...]

PR #2246, opened 2023-02-22
Move file_id_manager.db sqlite db off NFS

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/2245 Cause of https://github.com/2i2c-org/infrastructure/issues/2244 Reported upstream as https://github.com/jupyter-server/jupyter_server_fileid/issues/60

I think without this, images that upgrade to JupyterLab 3.6 might cause intermittent failures, even when not using RTC.

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -236,11 +236,18 @@ jupyterhub:
           ServerApp: &server_config_server_app
             extra_template_paths:
               - /usr/local/share/jupyter/custom_template
+          # Move the sqlite file used by https://github.com/jupyter-server/jupyter_server_fileid
+          # off the default path, which is under ~/.local/share/jupyter.
+          # That is NFS, and sqlite + NFS don't go well together. In addition,
+          # it uses WAL mode of sqlite, and that is completely unsupported on NFS
+          BaseFileIdManager: &server_config_base_file_id_manager
+            db_path: /tmp/file_id_manager.db
       jupyter_notebook_config.json:
         mountPath: /usr/local/etc/jupyter/jupyter_notebook_config.json
         data:
           MappingKernelManager: *server_config_mapping_kernel_manager
           NotebookApp: *server_config_server_app
+          BaseFileIdManager: *server_config_base_file_id_manager
     startTimeout: 600 # 10 mins, because sometimes we have too many new nodes coming up together
     defaultUrl: /tree
     image:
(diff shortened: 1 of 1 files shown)


yuvipanda:
I tested this, and can confirm that it puts the db under /tmp

pnasrat (commented):
LGTM

pnasrat (approved):

helm-charts/basehub/values.yaml
```diff
@@ -236,11 +236,18 @@ jupyterhub:
           ServerApp: &server_config_server_app
             extra_template_paths:
               - /usr/local/share/jupyter/custom_template
+          # Move the sqlite file used by https://github.com/jupyter-server/jupyter_server_fileid
+          # off the default path, which is under ~/.local/share/jupyter.
+          # That is NFS, and sqlite + NFS don't go well together. In addition,
+          # it uses WAL mode of sqlite, and that is completely unsupported on NFS
```
consideRatio, line 242:
```suggestion
          # it uses WAL mode of sqlite, and that is completely unsupported on NFS
          # Upstream discussion in https://github.com/jupyter-server/jupyter_server_fileid/issues/60.
```

consideRatio (approved):
Wiee!!! Nice work tracking this down!

I'd love to learn the debugging path to realize this was the issue!

---

@yuvipanda @pnasrat do you think this could have caused the slowness reported by the openscapes folks in the workshop they had in https://2i2c.freshdesk.com/a/tickets/439?

yuvipanda:
@consideRatio it *could be*, but I somehow doubt it, as the failure mode observed here was total failure to save rather than a service degradation. However, consistently making NFS ops on sqlite does cause service degradation for everyone, but only at scale - the nbsignatures db is still on NFS, and mostly fine. So I would suspect it is not.

As for debugging, @pnasrat was poking around with cool new functionality of kubectl / ephemeral debug pods that was pretty fun! After trying to figure out if it was NFS or jupyter, the fact that RStudio worked fine but Jupyter (classic nor lab) did not helped us narrow this down. After that, looking at the stacktrace (and using strace) to try figure out *which* db was the problem led us to the issue at hand. I didn't know fileid was using an sqlite db...

We put this information in the #managed_jupyte_inc_81 channel on slack if you are interested, @consideRatio

yuvipanda:
Given I'll be riding out tomorrow, I'd appreciate if someone else can take this one through?

consideRatio:
> Given I'll be riding out tomorrow, I'd appreciate if someone else can take this one through?

@pnasrat could you follow this through?

pnasrat:
@consideRatio sorry I didn't see the mention yesterday. Just for clarification what does "take this one through" mean - did it mean merge and deploy.

consideRatio:
No worries @pnasrat!!

Typically for changes to hub configuration and cloud infrastructure, we self-merge after approval. So I think what Yuvi meant, was that someone helps merge this PR and ensures it gets deployed successfully without creating an incident.

PR #2251, opened 2023-02-24
daskhub: remove redundant nodeSelector

### LINKED ISSUE #1574: Remove default nodeSelector configuration for dask-gateway core pods (`controller` and `api`)
The dask-gateway helm charts comes with three core pods: `api`, `controller`, `traefik`. Two of these have been explicitly required to schedule on k8s nodes with the label `k8s.dask.org/node-purpose: core`. But not all three pods, only `api` and `controller`.

https://github.com/2i2c-org/infrastructure/blob/840b3bdf6d643ce0f27ed286cbfe78a00492f28d/helm-charts/daskhub/values.yaml#L105-L110

I think everything would be fine if this configuration just isn't done at all for `api` and `controller`. It adds complexity forcing users to create k8s nodes with this label but doesn't add clear value in my mind - just like it doesn't add value to declare a node label specific for grafana, prometheus, cert-manager, and other such pods.

A dedicated nodeSelector for on demand user created `worker` and `scheduler` pods (together makes up a Dask Cluster created on demand) could still make perfect sense. So I'm just suggesting that `controller.nodeSelector` and `gateway.nodeSelector` (api pod) isn't set.

## Related

- https://github.com/2i2c-org/infrastructure/pull/1573#issuecomment-1199181157

consideRatio:
- Closes #1574

The nodes we have that isn't core nodes are dedicated nodes to dask workers and jupyterhub user servers. By removing this, we can stop needing to provide another set of labels to our node pools and use more similar config independent on basehub/daskhub setup etc.

### DIFF
--- a/helm-charts/daskhub/values.yaml
+++ b/helm-charts/daskhub/values.yaml
@@ -101,12 +101,7 @@ dask-gateway:
   enabled: true # Enabling dask-gateway will install Dask Gateway as a dependency.
   # Futher Dask Gateway configuration goes here
   # See https://github.com/dask/dask-gateway/blob/master/resources/helm/dask-gateway/values.yaml
-  controller:
-    nodeSelector:
-      k8s.dask.org/node-purpose: core
   gateway:
-    nodeSelector:
-      k8s.dask.org/node-purpose: core
     backend:
       scheduler:
         extraPodConfig:


pnasrat (approved):
LGTM

consideRatio:
Thank you @pnasrat for all your reviews!!! :heart: :tada: :sunflower:

PR #2258, opened 2023-02-27
Add a separate nodepool for notebook and dask pods before ohw event

GeorgianaElena:
Same resources as for last year's event


consideRatio (approved):
@GeorgianaElena seems reasonable.

I think there is room for tweaks overall in cpu/memory requests/limits etc and that we should use `n2-highmem` overall as well as compared to `n1-highmem`, but this isn't the time to do tweaks. As it is now will provide 15 pods per node I think which is fine and functional.

PR #2270, opened 2023-02-27
Update openscapes hub images

betolink:
Updating the hub images on Staging to use earthaccess v0.5.0 and Rocker to include the same Python environment as the Python image. After we test them on staging we'll promote them to prod.


betolink:
I noticed that the latest tags were already merged and deployed.

Issue #2273, opened 2023-02-27
[2i2c: temple] Investigate user report of kernel crashes 


pnasrat:
### Context

Reported via: https://2i2c.freshdesk.com/a/tickets/494

On our current week's project we are getting an inordinate number of kernel crashes. The data set is larger than others we use at 6.6 mb.

pnasrat:
Initial log exploration link https://cloudlogging.app.goo.gl/uvmQ9KNvDpnyCzSq7

yuvipanda:
I think you can also add `resource.labels.container_name="notebook"` to the query, as kernels are in the user's container, which is identified by the name 'notebook'

yuvipanda:
The most common cause of kernel restarts is notebooks hitting the memory limit.  This is memory usage graph for last week (https://grafana.pilot.2i2c.cloud/d/MMHgC_Qnz/cluster-information?orgId=1&viewPanel=10&from=now-7d&to=now)

<img width="1671" alt="image" src="https://user-images.githubusercontent.com/30430/221693356-8076534d-a5ac-45ab-a0a0-209557c66e1d.png">

And here is memory usage per-user as a heatmap:

<img width="1674" alt="image" src="https://user-images.githubusercontent.com/30430/221693710-209b22fd-e831-4787-aada-b3249b85a820.png">

It's not entirely clear there's a *lot* of people hitting the limit, but some people definitely are.

We also have an OOM killer graph, that reports some killings - but we can't be sure if those are from temple or elsewhere (the recent ones are definitely from oceanhackweeks):

<img width="1668" alt="image" src="https://user-images.githubusercontent.com/30430/221693882-59689e5d-07ca-4da7-926b-d49f1229a68c.png">

yuvipanda:
The users have a 256M Guarantee and a 1G limit (which is the default for all of basehub, and i see no overrides in temple's config). Perhaps because the nodes are getting fuller, they are getting killed before they reach 1G as well.

So two options here are:

1. Increase the memory *guarantee*, to maybe 512MB.
2. Increase the memory *limit*, perhaps even temporaryily, to 2G.

I think in the longer term, this should be written into contracts (cc @colliand and @jmunroe) as this will increase our cloud costs.

pnasrat:
- [ ]  Roll back #2274 after March 4

pnasrat:
No OOM kills since guarantee bump. Putting in waiting as pending until after March 4

PR #2274, opened 2023-02-27
Temporarily increase amount of RAM for temple

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/2273


yuvipanda:
This will increase cloud costs, so we should revert this once their project is over.

pnasrat (approved):
LGTM

Thanks fo rlinking to the issues in the comments

PR #2279, opened 2023-03-01
terraform: relax pins of providers by pinning to major versions only

consideRatio:
By pinning to the patch or minor versions, we may miss out on bugfixes and additions of relevance.

In this commit, I've made use of the `~>` version specifier, which means that the last provided version specifier is a lower bound. So, `~> 3.0` means `>= 3.0, < 4.0` and `~> 0.7.2` means `>= 0.7.2, < 0.8`.

I see this as a good trade off between pinning tightly and not pinning at all. Like this we can avoid bugs, rely on up-to-date documentation, without checking when something is introduced, and avoid breaking changes quite well.

In this commit, I've also updated the lower bound to the most recent version to help us narrow the range of actively used versions.

### DIFF
--- a/terraform/aws/main.tf
+++ b/terraform/aws/main.tf
@@ -1,13 +1,15 @@
 terraform {
   required_providers {
     aws = {
+      # ref: https://registry.terraform.io/providers/hashicorp/aws/latest
       source  = "hashicorp/aws"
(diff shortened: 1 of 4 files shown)


pnasrat (approved):
LGTM

Issue #2228, opened 2023-02-18
Need to understand how cpu requests/limits impacts software in containers


consideRatio:
## Updated action points

- [x] Understand this situation
- [ ] Make this understanding be conveyed at least partially as part of documentation in a section related to 2i2c's node sharing configurations.

## Original issue

Kubernetes allows us to specify requests and limits for cpu and memory, but without a very clear understanding on how these influence containers, we can't make good decisions on what we declare.

### Key question

Can and is software like dask's `distributed` running in a k8s pod be aware of:
- contaier's cpu requests
- container's cpu limits
- node's cpu capacity

### Thought example 1

A user pod scheduled by itself on a 4 CPU node with a 0.1 CPU request and 2 CPU limit. What happens when the user runs a Python program using multiprocessing that starts four processes running Python code, each able to load one CPU?

Erik: I think 2 CPU is fully loaded. The Python program wasn't aware of the 0.1 request and 2 limit, and got throttled by the cpu limit.

### Thought example 2

Again a user pod scheduled by itself on a 4 CPU node with a 0.1 CPU request and 2 CPU limit. "Smart" software (dask?) decides somehow how to make use of the available CPU capacity to do work, is it possible that this software self constrains to 0.1 CPU (cpu request), 2 CPU (cpu limit), or 4 CPU (node capacity)?

Erik: I'm really not sure.

### Reading up

- [Kubernetes docs about requests and limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits)
- [Specifying only a limit is a shorthand for specifying request/limit to be the same](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits)
- [A k8s node running `kubelet` tasked to start a pod, starts containers using the container runtime, which on linux typically makes use of cgroups to enforce requests/limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#how-pods-with-resource-limits-are-run)
  - [CPU requests are often associated with weighting](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#how-pods-with-resource-limits-are-run), so if a 1 CPU request container and a 3 CPU request container competes to consume CPU on a 16 CPU node, the CPU request container will get three times more.

### Key question reformulated

Can and is software like dask's `distributed` be aware of properties about the `cgroup` it runs within?

### Reading up

[...]

PR #2281, opened 2023-03-01
openscapes: revert "Fix misplaced decimal"

consideRatio:
This reverts commit a7cbcc2f77fbce908a9079b6c4dbd8e645647f24 which tried to fix an issue that I quite strongly believe was unrelated to the change following an investigation in #2228.

I'd like to let our stratgy on CPU requests/limits and Memory requests/limits stay cohesive between hubs and not have this be an exception.

If we have this commit not reverted, we must mitgate the bug it introduced:
- 4 times 1 CPU request on a 4 CPU node won't fit, 3 pods only
- 2 times 2 CPU request on a 4 CPU node won't fit, 1 pod only
- 1 times 4 CPU request on a 4 CPU node won't fit, it will fail to schedule

### Verification

- [x] Try out that I can indeed make use of all CPU even if I make a small request on staging.
  See https://github.com/2i2c-org/infrastructure/pull/2281#issuecomment-1450054008 for a practical verification of the theory in #2228, which in my mind makes it safe to revert this precautionary fix.

### DIFF
--- a/config/clusters/linked-earth/common.values.yaml
+++ b/config/clusters/linked-earth/common.values.yaml
@@ -55,11 +55,24 @@ basehub:
         tag: "latest"
 
       profileList:
+        # NOTE: About node sharing
+        #
+        #       CPU/Memory requests/limits are actively considered still. This
+        #       profile list is setup to involve node sharing as considered in
+        #       https://github.com/2i2c-org/infrastructure/issues/2121.
+        #
+        #       - Memory requests are lower than the description, with a factor
+        #         of (node_max_mem - 4GB) / node_max_mem.
+        #       - CPU requests are lower than the description, with a factor of
+        #         10%.
+        #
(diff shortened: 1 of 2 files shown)


yuvipanda:
The reason i thought it was a mistake was because the numbers don’t match the description. While i still don’t understand this, I would request at least a minimum comment stating this is intentional inline in the docs to prevent confusion. Am looking forward to killer docs you will eventually write i am sure.

yuvipanda:
I am happy with this revert once a comment is added to prevent future confusion.

consideRatio:
Thank you @yuvipanda!

I've added comments in 2361db4 and I saw that the profile list entry for small didn't configure memory limits correctly still and patched that in ad141b4 also.

consideRatio:
> I am happy with this revert once a comment is added to prevent future confusion.

I consider this as an approval given that I've added a comment, and will go for a merge now!

Issue #2129, opened 2023-02-01
Ensure our hub pod's templates-sync containers shut down quickly on SIGTERM


consideRatio:
When `kubectl delete pod` is run for example, a pod's container's entrypoint's are sent a SIGTERM. If they ignore that, they will not understand they are asked to terminate, so they will end up shutting down by force after 30 seconds instead.

I think z2jh's pods, including the hub pod, are optimized to shut down fast. But I think basehub's hub pod is slow to shut down because of the side-car containers we inject via the deployer CLI.

https://github.com/2i2c-org/infrastructure/blob/a9d8816bb423cb3117ef94935f303e4afd1cc9ee/deployer/hub.py#L87-L121

I think we should ensure that our hub pod's extra container can shut down without needing to wait for 30 seconds. If we for example ensure that the hub pod is faster to shut down, it can lead to less downtime during upgrades, faster CI/CD workflow executions, and such.

### Action point

- [x] Start a docker container locally and validate my belief it blocks a fast shut down, as can be seen using `docker stop` [similar to how it was tested in this issue](https://github.com/jupyterhub/configurable-http-proxy/issues/215).
- [x] Make changes until its respecting SIGTERM and shuts down properly.
- [x] Make the equivalent changes in our k8s context

### Related

- https://github.com/search?q=org%3Ajupyterhub+sigterm+author%3Aconsideratio&type=pullrequests

pnasrat:
Some notes from local testing

```sh
# Create volume
docker run -ti --rm  -v custom-templates:/srv/repo -l templates-clone alpine/git clone -- https://github.com/2i2c-org/default-hub-homepage /srv/repo
# Replicate templates-sync container
docker run -ti --rm -v custom-templates:/srv/repo -w /srv/repo --entrypoint /bin/sh alpine/git -c "ls -lhd /srv/repo; while true; do git fetch origin; git reset --hard origin/master; sleep 5m; done"
```

pnasrat:
My signal handling in bash is a little rusty so I am going to refresh how `trap` and `wait` work with `sleep` under `/bin/sh -c`

consideRatio:
Excellent! Note I recall alpine doesnt come with bash anyhow, even though switching to another image is an opion also.

pnasrat:
I've started looking at https://github.com/kubernetes/git-sync as one mechanism to do this. 

[...]

PR #2283, opened 2023-03-01
Add templates-sync image

pnasrat:
This creates a new image built of alpine/git and adding bash that can be
used to allow extra containers to shutdown quickly.

See #2129


pnasrat:
Not actively working on this right now so closing for the moment

Issue #2009, opened 2022-12-19
A way to merge and not trigger CD pipeline


consideRatio:
- I figured I'd hit merge on https://github.com/2i2c-org/infrastructure/pull/2008, but its not motivating a full redeploy so I'd like to opt out. For example by setting a label on the PR before merge.

1. Is there agreement its worth adding complexity to the CI system to support that kind of feature?

yuvipanda:
IMO, for that PR, we *should* actually do a full redeploy because it might change the version of helm installed and cause issues.

consideRatio:
I think we would only test if they have or havn't introduced a bug in the action itself by running the entire CD pipeline for this PR.

```diff
-    - uses: azure/setup-helm@v3.4
+    - uses: azure/setup-helm@v3.5
       with:
         # version is pinned for helm to avoid an automatic update of its version
         # which would cause something unexpected without an action on our
         # behalf.
         version: v3.8.2
```

But still then, there are situations where it can make sense to not run the CD pipeline by manual intent in my mind. So, should a feature like this be reasonable to support?

Example: you have 4 dependabot PRs updating actions/setup-python, actions/checkout, etc. You could merge them and trigger redeploys x4, or you can merge three with a label to not trigger a redeploy, and then the fourth with a label to trigger a redeploy.

yuvipanda:
> Example: you have 4 dependabot PRs updating actions/setup-python, actions/checkout, etc. You could merge them and trigger redeploys x4, or you can merge three with a label to not trigger a redeploy, and then the fourth with a label to trigger a redeploy.

I would say in this case it makes sense, but only if we make sure that CI is not skipped completely - some human makes a choice on which one to not skip CI on, and if things break, figures out which one to rollback.

The basic thing I want to avoid is if we skip CI, then it breaks the next time someone does a deploy. Generally `noop` helm deploys are fairly quick and do not bounce anything, so my personal preference is to always just let it deploy :)

consideRatio:
A reason to not trigger re-deploy from the docs of decomissioning a hub.

>## 2. Remove the hub values file
>
>Without removing the values file from the repository first, the hub could be redeployed by any merged PR that triggers our CI/CD pipeline.

consideRatio:
Here is perhaps the prime candidate situation for wanting a CI Skip. If this is merged, we will make a full re-deploy of everything. With the health checks we make, it likeley also even triggers server startups in users hub, incurring costs for them.

[...]

PR #2285, opened 2023-03-01
Fix typo in deployer start-docker-proxy

pnasrat:


consideRatio (approved):
LGTM.

I would suggest merging and aborting the deploy and test hubs workflow just after is created to not run another healthcheck and start servers etc for this.

a practice discussed in https://github.com/2i2c-org/infrastructure/issues/2009

pnasrat:
@consideRatio I don't see consensus the issue linked that is an adopted policy.

consideRatio:
@pnasrat the cancelled workflow was the workflow responsible for commenting on this PR that a CD has started I think, while its actually ended up making a full re-deploy in https://github.com/2i2c-org/infrastructure/actions/runs/4305541369.

Feel free to ignore the failure in utoronto staging, I'm working to figure that out already!

pnasrat:
Ugh that was unobvious to me, I guess I thought it was just the early stage of the deploy and cancelling would not trigger any further actions

Issue #2288, opened 2023-03-02
Failure to run health check on utoronto


consideRatio:
In [jobs to deploy to uturonto's staging, which involves to run a health check](https://github.com/2i2c-org/infrastructure/actions/runs/4305541369/jobs/7508146091) we are failing when the [`jhub_client`](https://github.com/Quansight/jhub-client/tree/master/jhub_client) software tries to ensure that a user exists.

```python-traceback
Running hub health check for staging...
Testing locally, do not redirect output
F                                                                                                                              [100%]
============================================================== FAILURES ==============================================================
__________________________________________________________ test_hub_healthy __________________________________________________________

hub_url = 'https://staging.utoronto.2i2c.cloud', api_token = '41515c727ecea362e053f00778095f4a71abf1dbc5e9aa10501c51788914c2d3'
notebook_dir = PosixPath('/home/erik/dev/2i2c-org/infrastructure/deployer/tests/test-notebooks/basehub'), check_dask_scaling = False

    @pytest.mark.asyncio
    async def test_hub_healthy(hub_url, api_token, notebook_dir, check_dask_scaling):
        try:
            print(f"Starting hub {hub_url} health validation...")
            for root, directories, files in os.walk(notebook_dir, topdown=False):
                for i, name in enumerate(files):
                    # We only want to run the "scale_dask_workers.ipynb" file if the
                    # check_dask_scaling variable is true. We continue in the loop if
                    # check_dask_scaling == False when we iterate over this file.
                    if (not check_dask_scaling) and (name == "scale_dask_workers.ipynb"):
                        continue
    
                    print(f"Running {name} test notebook...")
                    test_notebook_path = os.path.join(root, name)
                    await check_hub_health(hub_url, test_notebook_path, api_token)
    
            print(f"Hub {hub_url} is healthy!")
        except Exception as e:
            print(
                f"Hub {hub_url} not healthy! Stopping further deployments. Exception was {e}."
            )
>           raise (e)

deployer/tests/test_hub_health.py:84: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
deployer/tests/test_hub_health.py:77: in test_hub_healthy
[...]

PR #2290, opened 2023-03-02
deployer: unfreeze requirements.txt so we pin to major version only

consideRatio:
While a frozen environment can help with reproducibility, I think we now have a very bad in-between in our deployer's requirements.txt file. I think its acceptable that we would have either:

- a requirements.in outlining the actual dependencies, and a requirements.txt representing a frozen snapshot that together with automation to update that regularly
- just a requirements.txt representing the actual dependencies, perhaps pinned to the major version also

Right now, we have pinned the direct dependencies, but not the full environment - only the direct dependencies, not the indirect ones. That I think is bound to cause issues down the line.

I open this PR now as when I was debugging #2288, I saw an error from `jhub_client`. That led me to looking at version we used, which wasn't the latest. We used 0.1.6 and there was 0.1.7 out with a bugfix - was that relevant? In this case, it wasn't, but I was side-tracked to investiate that as the cause of the issue. Of course, things can go both ways. We can end up investigating if there is a regression in some dependency by not pinning.

Overall, I think this approach is what is best for us right now, and the main option considered would be to use a `requirements.in` file and [automation as found in z2jh / binderhub](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/29b041ced53678357d062c188f52e8983bfb1f03/.github/workflows/watch-dependencies.yaml#L199-L230) to update a frozen `requirements.txt` file.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -4,10 +4,10 @@
 
 # chartpress is relevant to build and push helm-charts/images/hub/Dockerfile and
 # update basehub's default values to reference the new image.
-chartpress
+chartpress==2.*
 
 # requests is used by extra_scripts/rsync-active-users.py
-requests
+requests==2.*
 
 # rich is used by extra_scripts/count-auth0-apps.py
 rich==13.*
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -4,10 +4,10 @@
 
 # chartpress is relevant to build and push helm-charts/images/hub/Dockerfile and
 # update basehub's default values to reference the new image.
-chartpress==2.1.0
+chartpress
 
 # requests is used by extra_scripts/rsync-active-users.py
-requests==2.28.1
+requests
 
 # rich is used by extra_scripts/count-auth0-apps.py
-rich==13.3.1
+rich==13.*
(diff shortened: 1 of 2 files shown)


dev-requirements.txt
```diff
@@ -4,10 +4,10 @@
 
 # chartpress is relevant to build and push helm-charts/images/hub/Dockerfile and
 # update basehub's default values to reference the new image.
-chartpress==2.1.0
+chartpress
```
pnasrat, line 7: The PR states pinning to major version but that's not happening here and below for requests. 

Can you add major pinning for those two

consideRatio: Makes sense, done!

consideRatio:
Thank you @pnasrat for reviewing!!

damianavila:
> Overall, I think this approach is what is best for us right now, and the main option considered would be to use a requirements.in file and [automation as found in z2jh / binderhub](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/29b041ced53678357d062c188f52e8983bfb1f03/.github/workflows/watch-dependencies.yaml#L199-L230) to update a frozen requirements.txt file.

Can you capture this idea in a new issue @consideRatio? Otherwise, we are going to lose it in this conversation. Thanks!

Issue #2287, opened 2023-03-01
[QCL] Tracking: New Hub Deployment 


pnasrat:
This is a tracking issue to track the technical steps of the new hub deployment in #2118

pnasrat:
## TODO 

- [ ] New GCP project https://infrastructure.2i2c.org/en/latest/hub-deployment-guide/cloud-accounts/new-gcp-project.html
- [ ] New Kubernetes cluster https://infrastructure.2i2c.org/en/latest/hub-deployment-guide/new-cluster/new-cluster.html
- [ ] Deploy support https://infrastructure.2i2c.org/en/latest/hub-deployment-guide/deploy-support/index.html
- [ ] Hub deployment docs https://infrastructure.2i2c.org/en/latest/hub-deployment-guide/hubs/new-hub.html
- [ ] Authentication https://infrastructure.2i2c.org/en/latest/hub-deployment-guide/configure-auth/index.html#

pnasrat:
I followed the intial steps to create `qcl-hub` GCP project (note GCP project naming requires 6-30 letters so not just `qcl`). Next step would be terraform.

pnasrat:
Terraform run and follow up plan gives

```
No changes. Your infrastructure matches the configuration.
```

pnasrat:
Notes to self

<details><summary>terraform IAM error</summary>
<p>

```
╷
│ Error: Error applying IAM policy for service account 'projects/qcl-hub/serviceAccounts/qcl-prod@qcl-hub.iam.gserviceaccount.com': Error setting IAM policy for service account 'projects/qcl-hub/serviceAccounts/qcl-prod@qcl-hub.iam.gserviceaccount.com': googleapi: Error 400: Identity Pool does not exist (qcl-hub.svc.id.goog). Please check that you specified a valid resource name as returned in the `name` attribute in the configuration API., badRequest
│ 
│   with google_service_account_iam_binding.workload_identity_binding["prod"],
│   on workload-identity.tf line 26, in resource "google_service_account_iam_binding" "workload_identity_binding":
│   26: resource "google_service_account_iam_binding" "workload_identity_binding" {
│ 
╵
╷
[...]

Issue #2121, opened 2023-01-31
New default machine types and profile list options - sharing nodes is great!


consideRatio:
Our clusters ships with a default set of machine types, and we provide a KubeSpawner profile_list to start pods on each type where there is a 1:1 relation between nodes:pods. For each node, there is only one user pod.

I thinnk a `1:n` nodes:pods relationship should be the default, not `1:1`! With `1:n`, I think we will provide a far better experience for users and for ourselves since it will reduce the amount of support tickets we will get I think. I think I can motivate this thoroughly, but to not write a long post, let me propose new defaults instead.

### Proposed new defaults

1. **High memory machine types - Intel or AMD**
   UPDATE: Now tracked via #2511, #3210
   Let's provide high memory machine types instead of normal machine types as memory is the limitation when sharing nodes.
   - I suggest the either the `n2-highmem-4` / `e2-highmem-4` on GCP and `r5.xlarge` / `r5a.xlarge` as the smallest node types for GCP / AWS.
   - Note that `n1-highmem` machines doesn't have 1:8 ratio between CPU:Memory and shouldn't be considered, the `n2` machines are also having more performant CPUs. Only when GPU's are needed, we must use `n1` as compared to `n2`.
2. **Intel or AMD**
   UPDATE: we stick with intel 
   The difference between n2/e2 and r5/r5a on GCP and AWS is between Intel and AMD processors, where the AMD processors are ~10% and ~30% cheaper on GCP and AWS respectively. I suggest we default to AMD unless we foresee common issues. EDIT: It was GCP that was 30% cheaper, and AWS 10% cheaper.
3. **4 / 16 / 64 CPU machines - no in-between choices**
   UPDATE: Tracked via #3256
   Provide machine types that increment 4x in size to simplify a lot, where we start at 4 CPU, then 16 CPU, then 64 CPU.
4. **1:n request of a machine, where n is 1 / 4 / 16**
   UPDATE: tracked via #3030
   We default to provide a profile list that include three choices representing the three machine types, but for each, we allow the user to specify the share they want. Do they want a dedicated server, a 4th of a server, or a 16th of a server?
5. **About how requests/limits are set**
   UPDATE: tracked via #3030
   The CPU requests/limits and Memory requests/limits should become as below, where `n` is 1 for dedicated, 4 for a 4th of a server etc.
   - CPU requests 1/n, CPU limits 1
   - Memory requests 1/n, Memory limits 1/n
[...]

PR #2291, opened 2023-03-02
2i2c-org/infrastructure#2287 QCL new cluster

pnasrat:
Add terraform config for new QCL cluster


terraform/gcp/projects/qcl.tfvars
```diff
@@ -0,0 +1,84 @@
...
+    min : 0,
+    max : 100,
+    machine_type : "n2-highmem-16",
+    labels: {},
+    gpu: {
+      enabled: false,
+      type: "",
+      count: 0
+    }
+  },
```
consideRatio, line 45: I've called these `small` and `medium` rather than `medium` and `large` in the profile list I'm trying advocating we standardize a bit around. That includes highmem 4 / 16 / 64 nodes, that the profile list refers to as small / medium / large.

To avoid confusion it would be good if these two are either named technically specific like `n2-highmem-4` like the machine or aligns with the user facing practice where `highmem-4` maps to `small` and `highmem-16` maps to `medium`.

This is all based on me trying to [nudge towards standarizing abit around specifically highmem 4 / 16 / 64 nodes](https://github.com/2i2c-org/infrastructure/issues/2121), but the legacy configuration has called small -> huge a range between 2 and 24 CPU i believe.

consideRatio (commented):
This LGTM overall!

- I made a suggestion about updating the highmem node-pools names.
- I'm not sure what k8s version this will result in as it seems its not declared in the terraform config, but I hope its quite modern and I believe k8s 1.25 would be fine for example because our charts are well updated and don't include deprecated resources removed in k8s 1.25 any more.

terraform/gcp/projects/qcl.tfvars
pnasrat: Will using n2-highmem* and n2-highcpu* names endup in the profile list automatically and thus potentially suffer from the prefix matching issue seen in kubespawner?

consideRatio: I think this name only influences the node pool name, which influences what you see when you do `kubectl get node`, as well as what users may see in grafana dashboards --- but the profile list is manually configured against labels, which are provided based on machine types.

consideRatio (approved):

pnasrat:
Re kube version I created (and will rerun with fixed node naming having deleted the nb pools)

```
kubectl version --short
Flag --short has been deprecated, and will be removed in the future. The --short output will become the default.
Client Version: v1.26.1
Kustomize Version: v4.5.7
Server Version: v1.24.9-gke.3200
```

Issue #2205, opened 2023-02-14
deployer UX notes - pnasrat


pnasrat:
### Context

Creating this following on from @GeorgianaElena's  https://github.com/2i2c-org/infrastructure/issues/1917#issuecomment-1429360248 to note my initial experiences/expectations re deployer as a new engineer.

pnasrat:
No mechanism to silence alerts or add/remove uptime checks - for the latter you use the python API for uptimeChec do against the `two-eye-two-see` project


https://cloud.google.com/monitoring/uptime-checks/manage

https://github.com/GoogleCloudPlatform/python-docs-samples/blob/ebf05af5900b66d614d0108d106b3ae2db420d8f/monitoring/snippets/v3/uptime-check-client/snippets.py#L105

pnasrat:
Ok for above it looks like terraform is being used for the uptime checks, as per

https://infrastructure.2i2c.org/en/latest/topic/monitoring-alerting/uptime-checks.html#how-are-the-checks-set-up

I guess in my brain I was assuming terraform for everything up to getting the k8s clusters running (ie cloud provisioning, storage, etc) and deployer for hubs, support, monitoring.

pnasrat:
Curious why cloud project setup for new hub isn't either documented via gcloud commands or automated in  terraform

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_project

pnasrat:
`deployer use-cluster-credentials` fails on empty cluster with no hub section.

pnasrat:
`deployer deploy` should take a `--debug` to pass to helm, etc

pnasrat:
Deployer should also support passing `--timeout` to helm or the github action should at least

See #2315 comment https://github.com/2i2c-org/infrastructure/pull/2315#issuecomment-1458603780

We get helm error is Error: UPGRADE FAILED: pre-upgrade hooks failed: timed out waiting for the condition

The defaulkt helm timeout is 5min so I think we are hitting this occasionally maybe we should set timeout in the github deployer run

https://helm.sh/docs/intro/using_helm/#helpful-options-for-installupgraderollback

PR #2294, opened 2023-03-02
Enable deployer use-cluster-credentials for empty cluster

pnasrat:
Enable use of kubectl on clusters with no hubs

See #2205


sgibson91 (approved):

consideRatio (approved):

pnasrat:
Merged with cancelling the deploy actions as this is a small bug fix in deployer that shouldn't trigger a global deploy

Issue #1671, opened 2022-08-31
Issue with adding staff as hub admins if `admin_users` is not explicitly listed in the config (even as an empty list)


GeorgianaElena:
### Context

When trying out the hub in https://github.com/2i2c-org/infrastructure/pull/1664 I realized that I'm not of admin, even though the config to add 2i2c staff as admins seems to be correct.

### Proposal

Confirm this is a bug and try and fix it somehow

### Updates and actions

- I did some checks on other hubs too (earthlab, staging, demo), each of them using a different authenticator by deleting @damianavila's username (sorry about this) from the admin panel and then I did a re-deploy of that hub, expecting the username to be added back and be shown in the admin panel, but it didn't happen.
- Explicitly adding the admins through `admin_users` from the hub yaml config works
- 2023-01-05: Sarah confirms that problem arises when `admin_users` is not defined in the config, but as soon as it is added, even just an empty list, 2i2c staff members are granted admin access

GeorgianaElena:
I will close this at it appears to have resolved itself for now? ref: https://github.com/2i2c-org/infrastructure/pull/1664#issuecomment-1240356519 and https://github.com/2i2c-org/infrastructure/pull/1664#issuecomment-1240413937

sgibson91:
I am reopening this issue since I have been struggling with this in #1967 I have scoped the problem to just affecting **Google IDs**.

damianavila:
@sgibson91 is Sherlock mode 😉 

![sherlock](https://media.giphy.com/media/3o7TKIgQuu1EfOBcly/giphy.gif)

PR #2299, opened 2023-03-03
Take into account all possible states of c.Authenticator.admin_users …

GeorgianaElena:
…list in a hub config

Fixes https://github.com/2i2c-org/infrastructure/issues/1671 based on @sgibson91 's awesome detective work ✨ 

With this, we can remove redundant config that specifies `"admin_users = []"` in the config, but won't do it in this PR yet, because it might cause conflicts for https://github.com/2i2c-org/infrastructure/pull/2277
Will do all necersary simplification of the auth config in that big PR instead.

This came up when playing with CILogon on the staging hubs, when I realized that the `2i2c dask-staging` hub had no admin users.


sgibson91 (approved):
Awesome! 🚀

PR #2311, opened 2023-03-06
add qcl auth

pnasrat:
- Added doc note to grant org access after first login
- Add quantifiedcarbon GitHub admins and org


consideRatio (commented):
I'm not sure about the read:org vs read:user configuration here, but I know I've not fully understood this.

The info of relevance is if a user is part of certain org's team, and such info may not be public, but could be public. If its not public memberhsip for that user, which users can configure individually, i recall read:org has been required.

Do you think this is correct?

pnasrat:
I think you are right, I've changed it back to scope:org

consideRatio (approved):

Issue #2293, opened 2023-03-02
Goal: k8s maintenance


consideRatio:
I would greatly value catching up with k8s operation maintenance. This would involve upgrading a lot of clusters and establishing documentation on the technical steps, as well as documentation on the community interactions leading up to an upgrade. This can involve figuring out planned maintenance windows or similar.

### Initial catch up upgrades

helm and kubectl tooling:
- `helm` and `kubectl` version pinning policies, #2314

k8s clusters:
- AWS EKS, #2057 
- GCP GKE, #2157 
- Azure AKS, #2158

### Documentation of technical steps during upgrades

- [x] AWS EKS, #2142
- [ ] GCP GKE, #3250
- [ ] Azure AKS, #3251

### Documentation of related community interactions

During ugprades of k8s clusters, we will cause a disruption that take a while. We should have some routines for how to communicate and such when planning a larger maintenance operation. This can involve having a clear guidance on how to get in contact with the community representatives among other things.

### Notes

- After agreement with a community represenative on when I could perform disruptive maintenance, I added an announcement: https://github.com/2i2c-org/infrastructure/pull/2318. The community representative also communicated that: a) running servers would be forcefully stopped if running, b) service shouldn't be expected to be available during this time.

consideRatio:
Closing as the umbrella issue is no longer relevant, but retaining #3250 and #3251 open

PR #2314, opened 2023-03-07
cd: upgrade helm 3.8 -> 3.11 and kubectl 1.23 -> 1.24

consideRatio:
K8s 1.25 is out, and we may very well upgrade to it or create a new cluster with it very soon. That makes it relevant to ahead of time upgrade `kubectl` to version 1.24 so that we can assume based on a k8s official version skew poolicy that kubectl can work against the not yet deployed/upgaded k8s 1.25 cluster.

I think historically we have been held back by k8s clusters versioned 1.21 or older, but now we are at k8s 1.22 and above so we can probably make this upgrade.

I didn't find the reason why we were still using helm 3.8.2, but I recall issues of using something newer at some point. But with all Helm charts upgraded to modern versions and k8s clusters upgraded to 1.22+ I think we can also upgrade helm at this point.

### Related

- #2293

### DIFF
--- a/.github/actions/setup-deploy/action.yaml
+++ b/.github/actions/setup-deploy/action.yaml
@@ -71,16 +71,41 @@ runs:
     # This action use the github official cache mechanism internally
     - uses: azure/setup-helm@v3
       with:
-        # version is pinned for helm to avoid an automatic update of its version
-        # which would cause something unexpected without an action on our
-        # behalf.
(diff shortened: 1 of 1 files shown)


damianavila (approved):
LGTM as well.

consideRatio:
Thanks for the review @pnasrat and @damianavila!!

PR #2317, opened 2023-03-07
carbonpan add pangeo tensorflow ML and edge images

### LINKED ISSUE #2310: carbonplan: update image selection profileList
### Context

Requested via support ticket: https://2i2c.freshdesk.com/a/tickets/512

Community support request to update the images to the below

[main single user]: same as before
[pangeo ml notebook]: https://github.com/pangeo-data/pangeo-docker-images/tree/master/ml-notebook
[edge single user]: https://github.com/carbonplan/envs/tree/main/edge/single-user

pnasrat:
Fixes #2310


damianavila:
> Yes but see my comment in slack I'm unsure on the policy and impact of removal, but here adding is going to be a safe operation

Yep, your proposal makes sense to me now. We should eventually come up with a policy for these changes. A two-step policy with addition (+ somehow deprecation message) and then removal seems less risky.

damianavila:
So, maybe the next step here is to "mark" the old options as deprecated in the display name?

consideRatio:
In this case I understand [the request in the freshdesk ticket](https://2i2c.freshdesk.com/a/tickets/512) to be a switch from `[a, b, c]` to `[a, d, e]` rather than to `[a, b, c, d, e]`. I think its reasonable to:

- ask for a clarification if `[a, d, e]` or `[a, b, c, d, e]` is wanted
  - optionally also to suggest that option `[b, c]` can be labelled as "deprecated"
- merge a `[a, d, e]` update directly

I think it will be impossible for us in 2i2c to have a solid policy about this, and that we at best could provide some ideas. But, that we should otherwise just go with what the community representatives sais.

pnasrat:
@consideRatio I'm fully aware I can ask for clarification, via the support ticket. However I don't think there are reasons why we shouldn't consider policies and impact of changes.

pnasrat:
PTAL adding deprecated per support ticket discussion

consideRatio (approved):
I appreciate how thorough you are @pnasrat!!!

damianavila (approved):

pnasrat:
image

PR #2321, opened 2023-03-08
Decrease uptime check frequency and reduce check locations

pnasrat:
See #2320

PR #2324, opened 2023-03-09
resource requests/limits updates, mainly to ingress-nginx and prometheus

consideRatio:
- support, ingress-nginx: increase memory request from 90Mi to 250Mi
- support, prometheus: update comment about resource requests/limit
- support, prometheus: let default memory request match limit (2Gi)
- support, prometheus: lower CPU request to 0.1 and increase CPU limit to 4
- support, prometheus: increase memory request/limit to 4Gi from 2Gi
- 2i2c, prometheus: use default cpu/memory requsts/limits
- meom-ige, hub/proxy: use default cpu/memory requsts/limits
- meom-ige, prometheus: use default cpu/memory requsts/limits
- carbonplan, prometheus: use default cpu request, increase memory request to limit
- leap, prometheus: use default cpu request/limits, increase memory request to limit
- openscapes, prometheus: use default cpu request/limits, increase memory request to limit
- pangeo-hubs, prometheus: use default cpu request/limits
- pangeo-hubs, prometheus: increase memory req/lim from 16 to 20Gi
- utoronto, prometheus: use default cpu request/limits, increase memory request to limit
- basehub/binderhub: update fixme comments about resource req/lim
- meom-ige: remove no longer needed workaround for dask-gateway 0.9.0


pnasrat (approved):
Overall this looks good - and correctly provisioning our needed support infrastructure is definitely a good thing as can cause spurious issues like the one that prompted your investigation. 

LGTM

Would this trigger core pool scale up in some places? There might be a slight risk of  a flurry of pages if prometheus reschedules and then takes a while (11 mins) to come up if it has to move to a newly created node. Probably worth keeping an eye on the pagerduty-notifications and post to maintenance-notices before deploying.

consideRatio:
Thank you for the review @pnasrat!!!

> Would this trigger core pool scale up in some places?

I think possibly! But even if we don't wait for a scalup the prometheus-server pod's will typically restart which can make us trigger pagerduty notifications. When they restart they can take several minutes, I think maybe about 1-10, before reading through a write-ahead log (WAL). I think during that time, they may be unresponsive as well!

Also, when we now change the resource requests on the ingress-nginx pod, we will disrupt networking briefly when it restarts.

pnasrat:
Nod - my change in #2321 will increase the page window to 31 min of being down for prometheus which will stop noisy pages for restarts. 

The nginx reschedule worries me more, how might you release this in a way that doesn't impact hub users?

consideRatio:
> The nginx reschedule worries me more, how might you release this in a way that doesn't impact hub users?

When I think of it, I think this change may not be as disruptive as I thought initially:

1. The upgrade strategy is a rolling update, so another replica is added before the old one is removed.
2. I think maybe if ingress-nginx is being asked to stop after the new replica is running, that it may become unready and no longer receive new connections from the k8s Service redirecting to only ready pods.
3. I don't understand the networking well enough to think confidently about this, but I suspect many web interactions, besides tens of second long file downloads or similar, may not experience any disruptions.

I'm not sure about the consequences, but I think we should accept the risk. I could delay merging this and do work on the Sunday evening / monday Swedish morning as I've planned.

pnasrat:
Note I don't think this merits out of our deployment, merely thinking through the impact and understanding it.

I think you are right on the deployment upgrade strategy https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#updating-a-deployment

LGTM - go ahead and merge and monitor the deployment.

PR #2325, opened 2023-03-09
Prevent validate clusters workflow failing if there's nothing to validate

sgibson91:
I noticed that on occasion the validate clusters workflow would fail with an error message that valid JSON was not provided to the `strategy` field of the `validate-helm-charts-values-files` job. Example here: https://github.com/2i2c-org/infrastructure/actions/runs/4372364264

It turned out, these failures happen in cases when the paths filter return false for BOTH common paths, and cluster specific paths. In this case, we would skip the step(s) that would generate a matrix and set the `MATRIX` environment variable. While we should also then skip the `validate-helm-charts-values-files` job, it fails to properly initialise because the `MATRIX` env var, which is passed to its `strategy` field, has not been set at all.

By simply forcing the `MATRIX` env var to be set to an empty list, we achieve the desired effect of the second job in the workflow being skipped, instead of registering as failed.

### DIFF
--- a/.github/workflows/validate-clusters.yaml
+++ b/.github/workflows/validate-clusters.yaml
@@ -100,6 +100,18 @@ jobs:
             cluster_specific:
               - added|modified: config/clusters/**
 
+      - name: Test outputs
+        run: |
+          echo "event name"
+          echo ${{ github.event_name == 'workflow_dispatch' }}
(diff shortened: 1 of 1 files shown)


pnasrat (approved):
Great job figuring this out!

LGTM

PR #2327, opened 2023-03-09
leap, prometheus: reduce memory req/lim from 22Gi to 20Gi

consideRatio:
Otherwise, we may fail to schedule on a node at all.


pnasrat (approved):
LGTM

Issue #2320, opened 2023-03-08
Revise uptime check configuration as exceed free quota


pnasrat:
### Context

We have 92 checks every 5 minutes from 6 regions (US = 3 by self)

```
So (24*60/5) * 92 *  6 * 30 = 4769280
$0.30/1,000 executions (free monthly allotment is in theory 1 million executions per Google Cloud project, however these are all in one project)
4769280-1000000 = 3769280
```

Pricing from October 1 2022 monitoring uptime checks are charged
https://cloud.google.com/stackdriver/pricing

### Proposal

Back of the envelope calculation for a 100 hubs to give num checks per hour to fit in free tier

```math
\begin{equation}
x = \cfrac{1e6}{(100*24*31*3)}
\end{equation}
=4.48
```

pnasrat:
See https://cloud.google.com/monitoring/uptime-checks#monitoring_uptime_check_create-api for region list 

We currently check from all 6:

-  usa-virginia
-  usa-oregon
-  usa-iowa
-  eur-belgium
-  apac-singapore
-  sa-brazil-sao_paulo

pnasrat:
Created draft PR to consider - need to confirm the selected_regions use in terraform is correct

pnasrat:
Breaking down my other changes

`google_monitoring_uptime_check_config.period: ="900s"`

[period](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/monitoring_uptime_check_config#period)- (Optional) How often, in seconds, the uptime check is performed. Currently, the only supported values are 60s (1 minute), 300s (5 minutes), 600s (10 minutes), and 900s (15 minutes). Optional, defaults to 300s.

`google_monitoring_alert_policy.duration = "1860s"`

[duration](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/monitoring_alert_policy#duration) - (Required) The amount of time that a time series must fail to report new data to be considered failing. Currently, only values that are a multiple of a minute--e.g. 60s, 120s, or 300s --are supported.

`google_monitoring_alert_policy.aggregations.alignment_period = "900s"` https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/monitoring_alert_policy#alignment_period

pnasrat:
Old apparently fixed issue re this https://github.com/hashicorp/terraform-provider-google/issues/3132

pnasrat:
API docs and web ui state 15m period is valid

[...]

PR #2330, opened 2023-03-09
Reduce number of regions uptime-checks run in

pnasrat:
See #2320

### DIFF
--- a/terraform/uptime-checks/checks.tf
+++ b/terraform/uptime-checks/checks.tf
@@ -24,6 +24,7 @@ resource "google_monitoring_uptime_check_config" "hub_simple_uptime_check" {
   for_each = { for h in local.hubs : h.domain => h }
 
   display_name = "${each.value.domain} on ${each.value.cluster}"
+  selected_regions = ["USA",]
   timeout      = "30s"
 
   # Check every 5 minutes


pnasrat:
Tested on 2i2c staging

image

consideRatio (approved):
This LGTM!

As part of reviewing this, I tried catch up with my understanding about these uptime checks. Do you think this understanding is correct @pnasrat?

- We have "uptime checks" defined in the GCP project `two-eye-two-see` making HTTP(s) requests to check if something is up and running.
- Each uptime check has so far been a set of uptime checks, where the service checked is accessed from multiple locations of the world
- This change makes us only check service avaibility from USA
- This change is a change only to the `two-eye-two-see` GCP project, and not all individual cloud accounts with clusters.

pnasrat:
@consideRatio see the docs here about the central project https://infrastructure.2i2c.org/en/latest/topic/monitoring-alerting/uptime-checks.html but yes. The main issue #2320 contains the detailed analysis of why (our free quota limit is being blown through) and this is an attempt to reduce costs until we can revisit monitoring and alerting design in general.

This is split off from the changing period due to the apparent need to destroy then apply to change period. I'm updating the docs when I change the period, and will add about region selection there when I get that out for review after testing if my assumption around the INVALID ARGUMENT error and period is correct. 

Longer term I think these probes will be supplemented with detailed monitoring and alerting, that runs in cluster.

PR #2331, opened 2023-03-09
Add uptime_check boolean switch to hub config

pnasrat:
Updates schema and filters disabled hubs in uptime_check terraform creation

### DIFF
--- a/deployer/cluster.schema.yaml
+++ b/deployer/cluster.schema.yaml
@@ -268,3 +268,9 @@ properties:
               helm chart values for a single hub deployment. These filepaths are defined
               RELATIVE to the location of the cluster.yaml file and have the naming
               convention '{{ hub name }}.values.yaml'.
+          uptime_check:
+            type: boolean
+            default: true
+            description: |
+              Whether or not to enable uptime monitoring for this hub.
+              This should generally be true.
--- a/terraform/uptime-checks/checks.tf
+++ b/terraform/uptime-checks/checks.tf
@@ -7,7 +7,7 @@ locals {
     helm_chart = h["helm_chart"]
     cluster    = cy["name"]
     provider   = cy["provider"]
-  }]]))
+  } if lookup(h, "uptime_check", true)]]))
   # A list of all prometheus servers
   prometheuses = flatten([
     for cy in local.cluster_yamls : [


deployer/cluster.schema.yaml
```diff
@@ -268,3 +268,9 @@ properties:
               helm chart values for a single hub deployment. These filepaths are defined
               RELATIVE to the location of the cluster.yaml file and have the naming
               convention '{{ hub name }}.values.yaml'.
+          uptime_check:
+            type: boolean
+            default: true
+            description: |
+              Whether or not to enable uptime monitoring for this hub.
+              This should generally be true.
```
consideRatio, line 276: What do you think about putting this in a terraform variable instead?

If its put here, for me it currently implies that `deployer deploy <cluster>` or `deployer deploy-support <cluster>` should be used after changing this value as compared to `terraform apply`. It also influences logic tied to if there should be a redeployment by the deployer - any change to cluster.yaml implies a full re-deploy of all hubs in the cluster.

What do you think about adding a list of hubs that shouldn't get uptime checks under the terraform/uptime-checks folder somewhere?

pnasrat: To explain my reasoning here - I view config/clusters/*/cluster.yaml as the source of truth for clusters. For the moment putting it there as the terraform is already parsing the configs for the hub list seemed to be the simplest way to add an operational switch. The current structure means that the [github action](https://github.com/2i2c-org/infrastructure/blob/master/.github/workflows/ensure-uptime-checks.yaml) that does the deploy runs when new hubs are added and deleted, whether that should have been a deployer command or we should have seperate monitoring configs are bigger changes as part of #1917 

The fact our monitoring is relying on terrafom and not a deployer commandis an implementation detail that may change. I think there is a broader discussion about improving monitoring and alerting for Q2, and also broader thoughts re the design of the deployer.

I'd argue that currently as SRE's are not using the terraform/uptime-checks regulary but it is being done via a github action (not even a deployer action) putting it under terrraform would be less discoverable.

Most hubs would not have this set as it defaults to true but it means that binder-staging which is noisy paging us for days can be turned off - pages and alerts should be actionable. This just fills a gap using the as-is design. 

If you strongly believe that this should be outside of cluster.yaml, my recommendation would be to have a config/monitoring or a top level monitoring/ directory that eg added prometheus rules (such as disk fulll) can be added to and other alerts configured. However I'd want a change like that to be part of the Q2 monitoring story and not as an operational bandaid around the fact one of our pages is spamming us.

consideRatio: Thank you @pnasrat for digging into this and helping me catch up as well!

I didn't know we had automation to do `terraform apply` for the uptime checks, and that it triggered on changes in clusters.yaml files and on changes to the terraform folder. The cluster.yaml and terraform/uptime-checks are already more coupled than I thought, so this seems like a good decision.

Since this new configuration will stand out as an exception to the other existing configuration, I'd like to clarify that as part of this PR in the schemas description to help guide us in the future.

With this description update, I think this is perfectly fine to merge! I recognize this situation now as when program / helm chart / etc provides a configuration schema for itself, but some external program wants to append unrelated configuration. For k8s resources managed by some controller this is typically done via annotations, where unrecognized annoations are simply ignored by the controller. In the JupyterHub helm chart we provide the [`custom` configuration](https://z2jh.jupyter.org/en/stable/resources/reference.html#custom) where users may declare misc values to be read by custom logic later.

```suggestion
          uptime_check:
            type: boolean
            default: true
            description: |
              Whether or not to enable uptime monitoring for this cluster's hubs
              and prometheus server. This should generally be true.

              FIXME: This configuration is not applied by the deployer, but by
                     the GitHub workflow ensure-uptime-checks.yaml using
                     terraform based on scripts in terraform/uptime-checks. In
                     the future we may want to for example refactor out this
                     configuration to a section under the cluster.yaml dedicated
                     for deployer unrelated configuration.

              For more information, see
              https://infrastructure.2i2c.org/en/latest/topic/monitoring-alerting/uptime-checks.html.
```

consideRatio (approved):
I didn't initially understand that this PR was essential to disable the uptime check thanks to the automation in .github/workflows/ensure-uptime-checks.yaml that re-adds it.

I added a fixme note suggestion that I'd like to see part of this PR, but it otherwise seems great to me!

consideRatio (commented):
Hmmm, this will change core infra and will redeploy all hubs. Currently, there are a few hubs already deployed to match new machine types.

Feel free to merge, but please cancel the in-this-case unnecessary automated deployment of all hubs if you merge it while a some of my infra PRs remain open still.

- #2337
- #2340
- #2343

---

UPDATE: Currently I know of no issues with redeploying all hubs - we'll be fine if the CD pipeline runs!

pnasrat:
I think I've figured how to do this in alert rules in Pager Duty. There is also a way to snooze these in GCP based on alert policy that might be easier than the pagerduty alert filters

https://cloud.google.com/monitoring/alerts/manage-snooze#gcloud-cli

pnasrat:
Removing this pull request in favor of using snoozes in #2345

Issue #2118, opened 2023-01-31
[Request deployment] New Hub: QCL aka QuantifiedCarbon


ASamarkRoth:
### Important dates

- **Target start date**: As soon as possible.
- **Required start date**: As soon as possible.
- **Any important dates for usage**:

### Hub Authentication Type

GitHub

### First Hub Administrators

Daniel Cox, dmcox@qcl.sh, @gizmo404
Joseph McKenna, jmckenna@qcl.sh, @jtkmckenna

### [GitHub Auth only] How would you like to manage your users?

Allowing members of a specific GitHub organization team

### [GitHub Teams Auth only] Profile restriction based on team membership

@QuantifiedCarbon/jupyterhub

### Hub logo image URL

https://avatars.githubusercontent.com/u/124042132?s=400&u=b84f1c7dfd1f9699b2adec7c8eb9ca7b9b2b0a6e&v=4

### Hub logo website URL

https://quantifiedcarbon.com/

### Hub user image tag and name

to-be-decided

### Extra features you'd like to enable

- [X] Specific cloud provider or datacenter (otherwise GCP)
- [X] Dedicated Kubernetes cluster
- [ ] Scalable Dask Cluster

### (Optional) Preferred cloud provider

GCP, `europe-west1`, where either zone b,c,d would work but let's pick `europe-west1-d`.

### (Optional) Billing and Cloud account

2i2c creates the GCP project.

### Other relevant information to the features above

Based on information requests tracked in https://2i2c.freshdesk.com/a/tickets/485, Erik concluded the following should be configured.

#### Custom domain name

QCL control's `quantifiedcarbon.com` and has provided the following CNAME configurations for the staging/production hubs and grafana server.

jupyter.quantifiedcarbon.com -> qcl.2i2c.cloud
staging.quantifiedcarbon.com -> staging.qcl.2i2c.cloud
grafana.quantifiedcarbon.com -> grafana.qcl.2i2c.cloud

#### Machine types and profile lists

Core nodes:
- `n2-highmem-2`

User nodes:
- `n2-highmem-4`
- `n2-highmem-16`
- `n2-highcpu-32`
- `n2-highcpu-96`

`singleuser.profileList` configuration:
[...]

PR #2332, opened 2023-03-10
QCL turnup add profile options based of linked earth

pnasrat:
See #2287

### DIFF
--- a/config/clusters/qcl/common.values.yaml
+++ b/config/clusters/qcl/common.values.yaml
@@ -182,7 +182,8 @@ jupyterhub:
         description: "Start a container on a dedicated node"
         slug: "n2_highcpu_32"
         kubespawner_override:
-          node.kubernetes.io/instance-type: n2-highcpu-32
+          node_selector:
+            node.kubernetes.io/instance-type: n2-highcpu-32
           mem_guarantee: 28.887G
           cpu_guarantee: 3.2
       # TODO(pnasrat): check on value once node running
--- a/config/clusters/qcl/common.values.yaml
+++ b/config/clusters/qcl/common.values.yaml
@@ -135,43 +135,43 @@ jupyterhub:
               mem_1:
                 display_name: ~1 GB, ~0.125 CPU
                 kubespawner_override:
-                  mem_guarantee: 0.969G
+                  mem_guarantee: 0.903G
                   cpu_guarantee: 0.013
               mem_2:
                 display_name: ~2 GB, ~0.25 CPU
                 kubespawner_override:
-                  mem_guarantee: 1.938G
+                  mem_guarantee: 1.805G
                   cpu_guarantee: 0.025
               mem_4:
                 default: true
                 display_name: ~4 GB, ~0.5 CPU
                 kubespawner_override:
-                  mem_guarantee: 3.875G
+                  mem_guarantee: 3.611G
                   cpu_guarantee: 0.05
               mem_8:
                 display_name: ~8 GB, ~1.0 CPU
                 kubespawner_override:
-                  mem_guarantee: 7.75G
+                  mem_guarantee: 7.222G
                   cpu_guarantee: 0.1
               mem_16:
                 display_name: ~16 GB, ~2.0 CPU
                 kubespawner_override:
-                  mem_guarantee: 15.5G
+                  mem_guarantee: 14.444G
                   cpu_guarantee: 0.2
               mem_32:
                 display_name: ~32 GB, ~4.0 CPU
                 kubespawner_override:
-                  mem_guarantee: 31.0G
+                  mem_guarantee: 28.887G
                   cpu_guarantee: 0.4
               mem_64:
                 display_name: ~64 GB, ~8.0 CPU
                 kubespawner_override:
-                  mem_guarantee: 62.0G
+                  mem_guarantee: 57.775G
                   cpu_guarantee: 0.8
               mem_128:
                 display_name: ~128 GB, ~16.0 CPU
                 kubespawner_override:
-                  mem_guarantee: 124.0G
+                  mem_guarantee: 115.549G
                   cpu_guarantee: 1.6
         kubespawner_override:
           cpu_limit: null
--- a/config/clusters/qcl/common.values.yaml
+++ b/config/clusters/qcl/common.values.yaml
@@ -54,3 +54,123 @@ jupyterhub:
           - quantifiedcarbon:JupyterHub
         scope:
           - read:org
+  singleuser:
+      # TODO(pnasrat): Add image customizations initially get community to try using configurator with default to 2i2c-hubs-image
+      profileList:
+        # NOTE: About node sharing
+        #
+        #       CPU/Memory requests/limits are actively considered still. This
+        #       profile list is setup to involve node sharing as considered in
+        #       https://github.com/2i2c-org/infrastructure/issues/2121.
+        #
+        #       - Memory requests are lower than the description, with a factor
+        #         of (node_max_mem - 4GB) / node_max_mem.
+        #       - CPU requests are lower than the description, with a factor of
+        #         10%.
+        #
+        - display_name: "Small: up to 4 CPU / 32 GB RAM"
+          description: &profile_list_description "Start a container with at least a chosen share of capacity on a node of this type"
+          slug: small
+          profile_options:
+            requests:
+              # NOTE: Node share choices are in active development, see comment
+              #       next to profileList: above.
+              display_name: Node share
+              choices:
+                mem_1:
+                  default: true
+                  display_name: ~1 GB, ~0.125 CPU
+                  kubespawner_override:
+                    mem_guarantee: 0.875G
+                    cpu_guarantee: 0.013
+                mem_2:
+                  display_name: ~2 GB, ~0.25 CPU
+                  kubespawner_override:
+                    mem_guarantee: 1.75G
+                    cpu_guarantee: 0.025
+                mem_4:
+                  display_name: ~4 GB, ~0.5 CPU
+                  kubespawner_override:
+                    mem_guarantee: 3.5G
+                    cpu_guarantee: 0.05
+                mem_8:
+                  display_name: ~8 GB, ~1.0 CPU
+                  kubespawner_override:
+                    mem_guarantee: 7.0G
+                    cpu_guarantee: 0.1
+                mem_16:
+                  display_name: ~16 GB, ~2.0 CPU
+                  kubespawner_override:
+                    mem_guarantee: 14.0G
+                    cpu_guarantee: 0.2
+                mem_32:
+                  display_name: ~32 GB, ~4.0 CPU
+                  kubespawner_override:
+                    mem_guarantee: 28.0G
+                    cpu_guarantee: 0.4
+          kubespawner_override:
+            cpu_limit: null
+            mem_limit: null
+            node_selector:
+              node.kubernetes.io/instance-type: n2-highmem-4
+        - display_name: "Medium: up to 16 CPU / 128 GB RAM"
+          description: *profile_list_description
+          slug: medium
+          default: true
+          profile_options:
+            requests:
+              # NOTE: Node share choices are in active development, see comment
+              #       next to profileList: above.
+              display_name: Node share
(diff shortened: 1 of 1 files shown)


consideRatio (commented):
Nice work Pris!

I hope its okay that I reviewed this now before a review is explicitly requested, I'll be quite busy on monday with other maintenance efforts and wanted to offload myself from doing it then.

I list what I overview as the remaining work for this PR below:

- [ ] Small as the default profile list entry
- [ ] Both highcpu-32 and highcpu-96 have profile list entries
- [ ] display_name and description of the highcpu-32 and highcpu-96 entries
  I think it would be good if we manage to reserve size adjective for the standardized choices of small/medium/large corresponding to highmem-4/highmem-16/highmem-64. Here is an idea of what we could have in display_name and description without a size adjective.
  ```yaml
        - display_name: "n2-highcpu-32: 32 CPU / 32 GB RAM"
          description: "Start a container on a dedicated node"
  ```
- [ ] If you look to include image choices in this PR, I think the two choices mentioned in https://github.com/2i2c-org/infrastructure/issues/2118#issuecomment-1460090187 should be included, and that we provide `pangeo/pangeo-notebook:latest` to represent the wish about using a prebuild "2i2c docker image". Me suggesting that image instead of `quay.io/2i2c/2i2c-hubs-image`. I opened #2336 to motivate why a new community should.
- [ ] Because of the default value in kubespawner to set `imagePullPolicy` to `IfNotPresent`, I think now that we reference `latest`, we should make this `Always` via `singleuser.image.pullPolicy`. (Related comment about this upstream: https://github.com/jupyterhub/kubespawner/issues/679#issuecomment-1319759943)

pnasrat:
@consideRatio could you share the script as I can't quite follow the rounding and math you are using  in #2338

consideRatio:
@pnasrat I put it here https://gist.github.com/consideRatio/071110916cb58220657398c61c14af7c.

Its based on having `kubectl get node` report information under `allocatable` for each node type as summarized in https://github.com/2i2c-org/infrastructure/issues/2121#issuecomment-1465277039.

I've done some manual transformations from KiB units to GiB and GB units as well, so the source for the logic is k8s reported allocatable memory for a node of each type in X KiB.

pnasrat:
@consideRatio re your point about changing pull policy that sounds like something that should be done outside of the single cluster and in basehub, is that correct?

consideRatio:
About `singleuser.image.pullPolicy`. Its not needed for any image that uses a immutable tag, but most hubs doesn't do that but instead has a specific image pinned. What would the downside be of using Always? More requests to the container registry, but if the container registry lists hashes for all the image's layers and those layer hashes are already around on the node - its not being downloaded again.

I have seen a binderhubf or example run into rate limitations by hub.docker.com for example though, so it can be relevant to not have `Always` as a imagePullPolicy by default unless its known to be needed for a mutable tag.

Here is an existing exception for carbonplan:

https://github.com/2i2c-org/infrastructure/blob/de1b067c83edadc99faebdfaf9c6e4b74f06dc82/config/clusters/carbonplan/common.values.yaml#L36-L41

I think it makes sense to opt-in for this rather than rely on a default value change in the basehub helm chart. Ideally in my mind, this should not be set at all in kubespawner, and then the k8s default behavior of acting as if its "Always" if the tag is `latest` specifically, and acting as if its "IfNotPresent" if the image tag is something else. Many helm charts, for example the boilerplate generated when using `helm create <new chart name>` comes with explicit configuration of the `imagePullPolicy` for some reason though.

pnasrat:
Not a downside I am merely ensuring that I would be addressing in the correct place!

pnasrat:
Ah great reading https://cloud.google.com/kubernetes-engine/docs/concepts/plan-node-sizes makes this clearer.

consideRatio:
I read that too!! Then i calculated based on that and couldt observe that the allocatable memory from the described 128GB memory for the n2-highmem-16 node for example. At that point i gave up and looked at the actually allocatable memory as reported by k8s for pods on the node.

pnasrat:
@consideRatio please take another look. I've done most things but will double check the allocatable once the profile list is up to both validate my understanding and understand the process you used to get the numbers

config/clusters/qcl/common.values.yaml
```diff
@@ -54,3 +54,142 @@ jupyterhub:
...
+          mem_guarantee: 28.887G
+          cpu_guarantee: 3.2
+      # TODO(pnasrat): check on value once node running
+      - display_name: "n2-highcpu-96: 96 CPU / 96 GB RAM"
+        description: "Start a container on a dedicated node"
+        slug: "n2_highcpu_96"
+        kubespawner_override:
+          node.kubernetes.io/instance-type: n2-highcpu-96
+          mem_guarantee: 86.662G
+          cpu_guarantee: 9.6
```
consideRatio, line 195: Ah you are consistent with it the requests here on nodes that users isn't sharing also, sure!

Since we specify a nodeSelector to spawn on a certain node, and we wish to provide users with dedicated nodes (1:1 pod:node), we can make any request as long as one user, but not two, fits on the node.

config/clusters/qcl/common.values.yaml
```diff
@@ -54,3 +54,142 @@ jupyterhub:
...
+                  cpu_guarantee: 0.4
+              mem_64:
+                display_name: ~64 GB, ~8.0 CPU
+                kubespawner_override:
+                  mem_guarantee: 62.0G
+                  cpu_guarantee: 0.8
+              mem_128:
+                display_name: ~128 GB, ~16.0 CPU
+                kubespawner_override:
+                  mem_guarantee: 124.0G
```
consideRatio, line 174:
```suggestion
                  mem_guarantee: 0.903G
                  cpu_guarantee: 0.013
              mem_2:
                display_name: ~2 GB, ~0.25 CPU
                kubespawner_override:
                  mem_guarantee: 1.805G
                  cpu_guarantee: 0.025
              mem_4:
                default: true
                display_name: ~4 GB, ~0.5 CPU
                kubespawner_override:
                  mem_guarantee: 3.611G
                  cpu_guarantee: 0.05
              mem_8:
                display_name: ~8 GB, ~1.0 CPU
                kubespawner_override:
                  mem_guarantee: 7.222G
                  cpu_guarantee: 0.1
              mem_16:
                display_name: ~16 GB, ~2.0 CPU
                kubespawner_override:
                  mem_guarantee: 14.444G
                  cpu_guarantee: 0.2
              mem_32:
                display_name: ~32 GB, ~4.0 CPU
                kubespawner_override:
                  mem_guarantee: 28.887G
                  cpu_guarantee: 0.4
              mem_64:
                display_name: ~64 GB, ~8.0 CPU
                kubespawner_override:
                  mem_guarantee: 57.775G
                  cpu_guarantee: 0.8
              mem_128:
                display_name: ~128 GB, ~16.0 CPU
                kubespawner_override:
                  mem_guarantee: 115.549G
```

config/clusters/qcl/common.values.yaml
```diff
@@ -54,3 +54,142 @@ jupyterhub:
...
+                  cpu_guarantee: 0.1
+              mem_16:
+                display_name: ~16 GB, ~2.0 CPU
+                kubespawner_override:
+                  mem_guarantee: 14.0G
+                  cpu_guarantee: 0.2
+              mem_32:
+                display_name: ~32 GB, ~4.0 CPU
+                kubespawner_override:
+                  mem_guarantee: 28.0G
```
consideRatio, line 119:
```suggestion
                  mem_guarantee: 0.836G
                  cpu_guarantee: 0.013
              mem_2:
                display_name: ~2 GB, ~0.25 CPU
                kubespawner_override:
                  mem_guarantee: 1.671G
                  cpu_guarantee: 0.025
              mem_4:
                display_name: ~4 GB, ~0.5 CPU
                kubespawner_override:
                  mem_guarantee: 3.342G
                  cpu_guarantee: 0.05
              mem_8:
                display_name: ~8 GB, ~1.0 CPU
                kubespawner_override:
                  mem_guarantee: 6.684G
                  cpu_guarantee: 0.1
              mem_16:
                display_name: ~16 GB, ~2.0 CPU
                kubespawner_override:
                  mem_guarantee: 13.369G
                  cpu_guarantee: 0.2
              mem_32:
                display_name: ~32 GB, ~4.0 CPU
                kubespawner_override:
                  mem_guarantee: 26.738G
```

config/clusters/qcl/common.values.yaml
```diff
@@ -54,3 +54,142 @@ jupyterhub:
...
+      #
+      #       CPU/Memory requests/limits are actively considered still. This
+      #       profile list is setup to involve node sharing as considered in
+      #       https://github.com/2i2c-org/infrastructure/issues/2121.
+      #
+      #       - Memory requests are different from the description, based on:
+      #         whats found to remain allocate in k8s, subtracting 1GiB
+      #         overhead for misc system pods, and transitioning from GB in
+      #         description to GiB in mem_guarantee
+      #         https://cloud.google.com/kubernetes-engine/docs/concepts/plan-node-sizes.
```
consideRatio, line 76: Yikes I struggle to understand my own comment above.

I disregarded the steps described in this linked documentation though when I was trying to understand how much memory was available for pods in k8s. I don't think there is an algorithm documented to take us from the reported specification of 128 GB to the allocatable memory as the linked documentation provides among other things.

So in a way, this was not used for the small/medium memory calculations - only the directly reported value for allocatable memory of a node of given types. I got these values using `kubectl get node | grep -A20 allocatable`

I think it could make sense to relocate this link to its own paragraph for readability, but I'd also want to rework this entire comment for all hubs with node sharing in a standalone PR.

No change requested, feel free to mark as resolved!

pnasrat: Yes. I think a clearer writeup in docs of shared node profile configuration would be helpful but yes lets not block QCL turnup on that.

consideRatio:
Btw if you get multiple code suggestions you want to commit, you have the choice to combine them into one via the GitHub UI, but only from the "Files changed" tab!

consideRatio (approved):
There is a todo note still in the PR, but I figure you can resolve it how you like either before or after merge!

My procedure here would be to deploy this PR manually to qcl staging, test it startup of highcpu-32 and 96, then if needed adjust the mem requests, remove the todo note from the PR, merge.

This approval applies even if highcpu-32 / 96 mem requests are adjusted if needed!

pnasrat:
Thanks for the review/approval. I will likely do this with the  manual staging deploy first thing my time tomorrow!

pnasrat:
Manual deploy to QC: staging

```
Running helm upgrade --install --create-namespace --wait --namespace=staging staging /home/pnasrat/workspace/src/github.com/2i2c-org/infrastructure/helm-charts/basehub --values=/tmp/tmp8yyyv0vu --values=/home/pnasrat/workspace/src/github.com/2i2c-org/infrastructure/config/clusters/qcl/common.values.yaml --values=/home/pnasrat/workspace/src/github.com/2i2c-org/infrastructure/config/clusters/qcl/staging.values.yaml --values=/tmp/tmptogaic_g
NAME: staging
LAST DEPLOYED: Tue Mar 14 09:54:56 2023
NAMESPACE: staging
STATUS: deployed
REVISION: 15
TEST SUITE: None
```

pnasrat:
```
kubectl get node gke-qcl-cluster-nb-huge-highcpu-56ceaf12-5fnb   -o json | jq '.["status"]["allocatable"]'
{
  "attachable-volumes-gce-pd": "127",
  "cpu": "95690m",
  "ephemeral-storage": "47060071478",
  "hugepages-1Gi": "0",
  "hugepages-2Mi": "0",
  "memory": "91143984Ki",
  "pods": "110"
}
```

allocatable memory 93.33GB

pnasrat:
```
kubectl get node gke-qcl-cluster-nb-large-highcpu-9a199cd3-75db    -o json | jq '.["status"]["allocatable"]'
{
  "attachable-volumes-gce-pd": "127",
  "cpu": "31850m",
  "ephemeral-storage": "47060071478",
  "hugepages-1Gi": "0",
  "hugepages-2Mi": "0",
  "memory": "29130868Ki",
  "pods": "110"
}
```

PR #2333, opened 2023-03-10
Update carbonplan default image

pnasrat:
See https://2i2c.freshdesk.com/a/tickets/512


pnasrat:
Merging without review as this changes the default on a single cluster per customer request.

PR #2335, opened 2023-03-10
carbonplan update default image to main

pnasrat:


pnasrat:
Ok this worked all along just local environment issue trying to work around CI/CD flakiness

Issue #2209, opened 2023-02-15
Transition LEAP cloud infra to use shared nodes


consideRatio:
I've together with @jbusecke distilled a support ticket to [improve user startup times] and [this this request for information](https://2i2c.freshdesk.com/a/tickets/446) among other things to this technical work items.

### Node pool changes

- There should only be one dask node pool, and it should use `n2-highmem-16`.
- The user node pools small/medium/large/huge should be removed
- A new user node pool (medium) should be added with a `n2-highmem-16` machine (16 CPU, 128 GB mem)

### New profile list

- The choices for small-huge is to be replaced with _two similar entries_ with slight differences to start on the 16 core machine.
  - This entry should present the image choices already available in the existing profile list
  - One version of the entry should be provided for the `leap-pangeo-base-access` github group, and one for `leap-pangeo-full-access`. They differ in the sense that one provides the node share options (1,2,4 CPU) and the other high (1,2,4,8,16 CPU)
  - The node share options should also set CPU/Memory limits to match the requests. This should also come with a comment that this is to be an exception rather than a default.
  
### Followup of actions

- #2364
- #2051

jbusecke:
Thanks for working on this. Please let me know before any changes are made, so I can make sure that:
- The new user groups `leap-pangeo-base-access` and `leap-pangeo-full-access` are populated (to avoid interruption of service for users
- I give people some heads up warning to not perform any work during the transition period

jbusecke:
Quick update: I have created and populated the new teams in our org: https://github.com/orgs/leap-stc/teams/leap-pangeo-users/teams

These are non-overlapping at the moment.

consideRatio:
@jbusecke I suggest that I work this on Monday morning until mid day, Swedish time UTC+1. Is that okay?

consideRatio:
There is currently three running users so I don't cancel their sessions and start an upgrade. @jbusecke should we aim for next week on monday morning Swedish timezone UTC+1?

jbusecke:
Sorry for the delay. I was really sick all last week and not able to stare at a screen. 

I have notified the LEAP community to log out and stop any long running computation before Sun Mar 12, and said that service should be back to normal on Monday after 8AM. Is that correct @consideRatio? 

I will make another announcement closer to the weekend. Thank you for working on this!

consideRatio:
@jbusecke absolutely correct!!

[...]

PR #2337, opened 2023-03-12
leap: k8s 1.25, node sharing setup, no more pre-pulling, no maintenance notice

consideRatio:
- Addresses all but the "dask-gateway options" section of #2209
- Since I was about to recreate all node pools to get highmem machines, I wanted to upgrade k8s to 1.25 while I was at it. But, this required support from our terraform scripts that I added in 64da264.
- In a93926a I disable prePuller with a motivation in the commit message

### Already deployed

This is already deployed and the terraform changes are applied together with the profileList changes that are tightly coupled with them. Due to that, it is important that this is merged before another PR makes a redeploy of the leap hub by accident with an old profileList.

### DIFF
--- a/config/clusters/leap/common.values.yaml
+++ b/config/clusters/leap/common.values.yaml
@@ -10,9 +10,6 @@ basehub:
       # Name of Google Filestore share
       baseShareName: /homes/
   jupyterhub:
-    prePuller:
-      continuous:
-        enabled: true
     custom:
       # Extra mount point for admins to access to all users' home dirs
       # Ref https://github.com/2i2c-org/infrastructure/issues/2105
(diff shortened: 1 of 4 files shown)


pnasrat (commented):
I don't see an explanation in here or #2209 about removing the pre-puller can you just explain why that's part of this change (or link me to the appropriate bug). Is there any impact to community startup time with that removed?

Everything else looks fine and I'll approve once you've responded

pnasrat (approved):
Thanks - yeah if you go from the PR straight to Files Changed for review you don't necessarily see commit messages.

LGTM

consideRatio:
Thank you for the quick reviews @pnasrat!!

I opened https://github.com/2i2c-org/infrastructure/issues/2344 about this, we lack a policy about this configuration overall still.

PR #2338, opened 2023-03-12
openscapes/linked-earth: fix node sharing to fit on nodes perfectly

consideRatio:
This fixes a bug that hasn't been reported yet I think. Its a bug related to requesting 101% or similar when we should request <100% of the allocatable memory for a node.

Like this, we account for the allocatable memory as reported by k8s for specific node types and provide 1 GiB of headroom for system pods on the node like kube-proxy etc.

### Already deployed

I've tested this and it solves the issue of failing to start on the largest node share variant.

### Related
- https://github.com/2i2c-org/infrastructure/issues/2121#issuecomment-1465277039

### DIFF
--- a/config/clusters/linked-earth/common.values.yaml
+++ b/config/clusters/linked-earth/common.values.yaml
@@ -61,8 +61,10 @@ basehub:
         #       profile list is setup to involve node sharing as considered in
         #       https://github.com/2i2c-org/infrastructure/issues/2121.
         #
-        #       - Memory requests are lower than the description, with a factor
-        #         of (node_max_mem - 4GB) / node_max_mem.
+        #       - Memory requests are different from the description, based on:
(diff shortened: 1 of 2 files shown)


pnasrat (approved):
LGTM

Longer term if this is a common pattern for profile sizing it might be nice to have code that generates the guarantees overrides with the appropriate deductions to encode the logic. However that's beyond the scope of this change.

consideRatio:
> Longer term if this is a common pattern for profile sizing it might be nice to have code that generates the guarantees overrides with the appropriate deductions to encode the logic. However that's beyond the scope of this change.

:100: I have a Python script locally on my computer for now, but its such an actively deliberated topic still so I've not focused on getting it into this repo yet.

consideRatio:
Thank you for review efforts @pnasrat!!!

PR #2340, opened 2023-03-12
nasa-veda: upgrade to k8s 1.25, highmem nodes, profile list with node sharing

consideRatio:
The nasa-veda cluster didn't have a profile list and they had run into 1GB memory limitations as noted in https://2i2c.slack.com/archives/C04R9B69T5W/p1678479120570969. I took initiative and did some disruptive maintenance now that people wasn't using it, setting up small / medium / large node options with an ability to pick a node share as well.

### Already deployed

This is already deployed, and if someone deploys the default branchs configuration to nasa-veda before this is merged, it will make it break.

### DIFF
--- a/config/clusters/nasa-veda/common.values.yaml
+++ b/config/clusters/nasa-veda/common.values.yaml
@@ -60,6 +60,176 @@ basehub:
       image:
         name: pangeo/pangeo-notebook
         tag: "2023.01.13"
+      profileList:
+        # NOTE: About node sharing
+        #
+        #       CPU/Memory requests/limits are actively considered still. This
+        #       profile list is setup to involve node sharing as considered in
+        #       https://github.com/2i2c-org/infrastructure/issues/2121.
+        #
+        #       - Memory requests are different from the description, based on:
+        #         whats found to remain allocate in k8s, subtracting 1GiB
+        #         overhead for misc system pods, and transitioning from GB in
+        #         description to GiB in mem_guarantee.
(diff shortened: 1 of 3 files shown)


GeorgianaElena (approved):

config/clusters/nasa-veda/common.values.yaml
```diff
@@ -55,11 +50,192 @@ basehub:
...
+        enabled: true
     singleuser:
       defaultUrl: /lab
       image:
+        # NOTE: To have prePuller enabled makes sense as long as we stick with a
+        #       single image option configured from here. If the configurator is
+        #       used, or a list of images is chosen from, then prePuller should
+        #       be disabled to avoid pulling images not to be used, which could
+        #       potentially delay startup of a user pod wanting to pull another
+        #       image.
```
GeorgianaElena, line 66: I think it's best to have this comment in the `prePuller` section.

consideRatio: I think I maybe you are right and i'll relocate it! 

It is a tricky trade off. I made the comment here under singleuser.image and put prePuller section closer. I figure it relates to: prePuller, singleuser.image, profileList where a dropdown of choices may be configured.

consideRatio: Resolved by [96399b8](https://github.com/2i2c-org/infrastructure/pull/2340/commits/96399b8db9e8405cec18ba6fa5fefab0b59c54f4)

Issue #2057, opened 2023-01-18
Migrate existing AWS EKS based hubs from k8s 1.21+ to 1.24+


consideRatio:
> Amazon EKS has deprecated Kubernetes version 1.22 and this version will no longer be supported on June 4, 2023. Starting that day, you will no longer be able to create new 1.22 clusters and all EKS clusters running Kubernetes version 1.22 will be updated to the latest available platform version of Kubernetes version 1.23.

This is the current status of AWS EKS clusters.
- [x] 2i2c-aws-us, k8s v1.25 (2i2c AWS SSO, two-eye-two-see), upgraded to 1.25
  #2343
- [x] carbonplan, k8s 1.19, upgraded to 1.24
  #2085
- [x] gridsst, k8s 1.22 (2i2c AWS SSO), upgraded to 1.25,
  #2373
- [x] nasa-cryo, k8s 1.22 (2i2c AWS SSO), upgraded to 1.25
  discussed for upgrade in https://2i2c.freshdesk.com/a/tickets/543
  #2374
- [x] nasa-veda, k8s 1.25, upgraded to 1.25
  #2340
- [x] openscapes, k8s 1.21, upgraded to 1.24
  #2139
- [x] ubc-eoas, k8s 1.24 (2i2c AWS SSO)
  Created at 1.24
- [x] victor, k8s 1.22 (2i2c AWS SSO), upgraded to 1.25
  #2375

### Action points

- [x] Upgrade all EKS clusters to k8s 1.24+
- [x] Document a upgrade procedure for EKS clusters
  #2142

### Related

- #2157
- #2158

### Original outdated issue

<details>
    <summary>Click to expand</summary>

Such migration will require some additional steps related to #2054 and #2056.

In practice, I think it involves updating the .jsonnet templates of eksctl cluster configuration files that we have in the [eksctl](https://github.com/2i2c-org/infrastructure/tree/HEAD/eksctl) folder to match how they would look if we re-generated them from the [jinja2 template (template.jsonnet)](https://github.com/2i2c-org/infrastructure/blob/master/eksctl/template.jsonnet) of these .jsonnet template files.

It also involves manually adding the EKSCTL addon with a `eksctl` command, and recreating the node pools, etc. This will cause disruption.

[...]

PR #2343, opened 2023-03-13
2i2c-aws-us: k8s 1.25, highmem nodes, node sharing profile list, ssh-keys

consideRatio:
I took the chance to upgrade this EKS cluster while it wasn't used. As part of that, I also transitioned into using node sharing.

### Related

- #2057
- #2121


pnasrat (approved):
LGTM

PR #2346, opened 2023-03-13
Simplify display name

jbusecke:
Thanks so much for the work on the server options refactor @consideRatio et al.!  I just checked the server start up times and it was massively improved!

I got some feedback from users that the `display_name` was slightly confusing. Particularly for members of `leap-pangeo-base-access` the description does say up to 16cores/128GB, which they are not able to select. I would suggest to rename to a simpler description in line with the GPU option.


consideRatio (approved):

consideRatio:
Sorry for the slow turnaround @jbusecke, thanks for your initiative!

PR #2347, opened 2023-03-13
Decrease uptime check frequency 

pnasrat:
See #2320

Deployment needs to be manual to destroy then create


consideRatio (approved):
This looks good to me, and I think an already existing comment motivates this change:

> This shoulod prevent alerts if the hub is momentarily down during a deployment. All alerts *must* be actionable, so we trade-off some latency here for resiliency.

pnasrat:
Have manually deployed and update checks created (checked in console)

PR #2349, opened 2023-03-14
Don't generate config in deployer anymore

GeorgianaElena:
❗ The only relevant change here can be checked out at https://github.com/GeorgianaElena/pilot-hubs/commit/80cca07f6b3b109aca9d7932d91de59b7aef6362.

I've only added the config to the 2i2c-staging hub to get an early approval about the workflow, but it should be added eventually to all of the hubs. What do you think? (Checkout https://github.com/2i2c-org/infrastructure/issues/1925 for more context)

- This is based off the work in #2277, so it should be merged after it.
- Towards closing https://github.com/2i2c-org/infrastructure/issues/1925 🚀


damianavila:
> ❗ The only relevant change here can be checked out at https://github.com/GeorgianaElena/pilot-hubs/commit/80cca07f6b3b109aca9d7932d91de59b7aef6362.

I would probably suggest merging #2277 first and then rebasing this one on top of main.

> I've only added the config to the 2i2c-staging hub to get an early approval about the workflow, but it should be added eventually to all of the hubs. What do you think? 

I concur, although should not be added to the basehub config instead? Well, you may be thinking about that when you say "it should be added eventually to all of the hubs", I presume.

consideRatio (changes requested):
This work is now unblocked by #2277 being merged and I'm excited about this follow up work!!

---

The PRs current approach is to duplicate the declaration of `jupyterhub.hub.initContainers` into each separate hub, at least that has provided customized login page configuration, because there is a reference to a branch name that should be configured per hub.

I'd like to see that instead of doing this, we let the basehub chart define a ConfigMap template that renders values we wish to be different per hub's initContainer. Then, we let the hub's added initContainer instead reference an environment variable, which we have mounted from the basehub's rendered ConfigMap.

Looking at the initContainers and extraContainers, I see only a branch name of relevance that differs. So I figure the following:

1. basehub adds a ConfigMap template that renders based on some variable we configure in values.yaml for basehub
2. basehub defines [`hub.initContainers`](https://z2jh.jupyter.org/en/stable/resources/reference.html#hub-initcontainers) and [`hub.extraContainers`](https://z2jh.jupyter.org/en/stable/resources/reference.html#hub-extracontainers), but also lets the extra container defined that references a hub specific variable do that via a [ConfigMap mounted environment variable](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#use-configmap-defined-environment-variables-in-pod-commands).

Considering the right to replicate, I also suggest that we allow the github repo we git clone to be a basehub values configuration that is rendered by the configmap template and read as an environment variable as well.

helm-charts/basehub/values.yaml
```diff
@@ -400,7 +404,77 @@ jupyterhub:
...
+        image: alpine/git
+        workingDir: /srv/repo
+        command:
+          - /bin/sh
+        args:
+          - -c
+          - |
+            ls -lhd /srv/repo;
+            while true; do git fetch origin;
+            git reset --hard origin/$(HUB_TEMPLATES_REPO_BRANCH);
```
GeorgianaElena, line 452: To simplify containers setup, we will be trading off automatically checking if a branch with `<cluster-name>-<hub-name>` exists and pull from there for a given hub. Instead, we will need to specify the branch in each hub's config. See the 2i2c-ohw config in this PR.

But I believe this is super ok, as it's always better to be explicit and the template containers configs gets to stay in one central place (the basehub)

GeorgianaElena:
@consideRatio, thanks for the great suggestion of simplifying this and for the instructions! Take a look at the latest changes. I've left a comment with the tiny tradeoff that we will need to do (will also add a todo in the top comment to update the template repo docs) as part of this, but I'm super ok with it. LMK what you think.

helm-charts/basehub/values.yaml
```diff
@@ -449,11 +449,7 @@ jupyterhub:
           - |
             ls -lhd /srv/repo;
             while true; do git fetch origin;
-            if [[ $(git ls-remote --heads origin $(HUB_TEMPLATES_REPO_BRANCH) | wc -c) -ne 0 ]]; then
-                git reset --hard origin/$(HUB_TEMPLATES_REPO_BRANCH);
-            else
-                git reset --hard origin/master;
-            fi
+            git reset --hard origin/$(HUB_TEMPLATES_REPO_BRANCH);
```
consideRatio, line 452: I think the syntax for referencing an env variable to use here, is valid if you were in the k8s `env` section defining another env var based on another env var.

This is the container command, inside a `sh` based shell though, so I think the syntax should be `$SOMETHING` or `${SOMETHING}` instead.

```suggestion
            git reset --hard origin/${HUB_TEMPLATES_REPO_BRANCH};
```

helm-charts/basehub/values.yaml
```diff
@@ -400,7 +404,77 @@ jupyterhub:
...
                     title: RStudio
                     description: An IDE For R, created by the RStudio company
-
+    initContainers:
+      - name: templates-clone
+        image: alpine/git
+        args:
+          - clone
+          - --
+          - $(HUB_TEMPLATES_REPO)
```
consideRatio, line 413:
```suggestion
          - ${HUB_TEMPLATES_REPO}
```

I'm not confident about this situation, thinking maybe env var expansion is trouble. It depends on if the container has `ENTRYPOINT` defined directly to `git` or `sh -c git` or similar maybe.

helm-charts/basehub/values.yaml
consideRatio, line 452: I think the syntax for referencing an env variable to use here, is valid if you were in the k8s `env` section defining another env var based on another env var.

This is the container command, inside a `sh` based shell though, so I think the syntax should be `$SOMETHING` or `${SOMETHING}` instead.

```suggestion
            git reset --hard origin/${HUB_TEMPLATES_REPO_BRANCH};
```

helm-charts/basehub/values.yaml
consideRatio: Ah nevermind, this is correct - see https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config.

helm-charts/basehub/values.yaml
consideRatio: Ah ignore me, this will work using k8s env var expansion and doesn't have to rely on the shell env var expansion.

GeorgianaElena: Btw @consideRatio, I've tested this on staging hub and it appeared to do the correct thing, so at least there shouldn't be syntax issues. (I'm doing it again now to double check 😅 )

config/clusters/2i2c/ohw.values.yaml
```diff
@@ -73,6 +73,8 @@ basehub:
       2i2c:
         add_staff_user_ids_to_admin_users: true
         add_staff_user_ids_of_type: "github"
+      hubTemplateRepoDetails:
+        repo_branch: "2i2c-ohw"
       homepage:
         templateVars:
           org:
```
consideRatio, line 80: There is already `custom.homepage`, and this is directly related to that right?

If so, should we just let this be nested under `custom.homepage` instead?

consideRatio (commented):
This looks reasonable to me!

helm-charts/basehub/templates/configmap-hub-templates.yaml
```diff
@@ -0,0 +1,9 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: hub-custom-templates-config
+  labels:
+    app: jupyterhub
+data:
+  HUB_TEMPLATES_REPO: {{ .Values.jupyterhub.custom.hubTemplateRepoDetails.repo_url }}
+  HUB_TEMPLATES_REPO_BRANCH: {{ .Values.jupyterhub.custom.hubTemplateRepoDetails.repo_branch }}
```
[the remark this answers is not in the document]
GeorgianaElena: Thnaks @consideRatio

config/clusters/2i2c/ohw.values.yaml
```diff
@@ -74,6 +74,8 @@ basehub:
         add_staff_user_ids_to_admin_users: true
         add_staff_user_ids_of_type: "github"
       homepage:
+        hubTemplateRepoDetails:
+          repo_branch: "2i2c-ohw"
         templateVars:
           org:
             name: OceanHackWeek
```
consideRatio, line 81: I figure `custom.homepage` is already implying `custom hub template` since there is `custom.homepage.templateVars` already.

What do you think about a structure like this?

```yaml
custom:
  homepage:
    gitRepoUrl: ...
    gitRepoBranch: ...
    templateVars: ...
```

I switched to camelCase for consistency and the practice of using such syntax for what's pure chart config, added `git` prefix for what I think could be helpful verbosity.

consideRatio: Note that to de-nest this to be flat - directly under custom.homepage is aligning with advice in helm chart docs: https://helm.sh/docs/chart_best_practices/values/#flat-or-nested-values

> In most cases, flat should be favored over nested. The reason for this is that it is simpler for template developers and users.

GeorgianaElena: > I figure custom.homepage is already implying custom hub template since there is custom.homepage.templateVars already.

I like the proposed removal of the `hubTemplateRepoDetails` details. Thank you for thinking and explaining the reasoning behind it <3! In my head I couldn't reason why they are supposed to be on the same tree level, but adding the `git` prefix in the key name does a great job grouping those together and explaing it.

helm-charts/basehub/values.schema.yaml
```diff
@@ -283,25 +283,21 @@ properties:
             required:
               - templateVars
```
consideRatio, line 284: gitRepoUrl and gitRepoBranch is also required for this to work and can be added here.

config/clusters/test/cluster.yaml
```diff
@@ -0,0 +1,35 @@
+name: openscapes
+provider: aws # https://783616723547.signin.aws.amazon.com/console
```
consideRatio, line 2: `config/clusters/test/cluster.yaml` also debugging remnant

config/clusters/2i2c-aws-us/researchdelight.values.yaml
```diff
@@ -15,6 +15,14 @@ basehub:
...
       baseShareName: /
   jupyterhub:
+    ingress:
+      hosts: [researchdelight.2i2c.cloud]
+      tls:
+        - hosts: [researchdelight.2i2c.cloud]
+          secretName: https-auto-tls
+    proxy:
+      https:
+    hosts: [researchdelight.2i2c.cloud]
```
consideRatio, line 25:
```suggestion
        hosts: [researchdelight.2i2c.cloud]
```

helm-charts/basehub/values.schema.yaml
GeorgianaElena: I agree. But isn't this obsolete since we're setting defaults in the basehub chart? I noticed that the config set by basehub is not listed in the schema and I guess it's to avoid the toil of checking every single bit of config.
But I'll add these two since they're new.

Actually, I will also add `jupyterhub.ingress` and `jupyterhub.proxy` to the schema, as we might miss to add it for future hubs since they've always been in basehub.

GeorgianaElena: Done in https://github.com/2i2c-org/infrastructure/pull/2349/commits/3f5a15ed46a255091f3270da2021bed34ee2e504

consideRatio: We depend on them to not error, so they are just as required as `templateVars`. Practically, since we have templateVars as required, we are already catching a common where one unset all values in `custom.homepage` for example - then its good that at least one variable is required because otherwise the schema validation would pass.

I'd like us to work towards being consistently thorough in the schema, so that we and anyone else can use the schema to understand what is and what is and isn't required by our templates etc. I've made z2jh and binderhub consistently thorough with things like this and it has helped for sure overall, for example catching a few "oops we unset all default values by setting a dictionary to null" situations.

consideRatio: > Actually, I will also add jupyterhub.ingress and jupyterhub.proxy to the schema, as we might miss to add it for future hubs since they've always been in basehub.

They are already handled by the jupyterhub schema and they are marked as required by the jupyterhub schema already.

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/6d3c13984d78791f6c6980a70ddf8cf0648cf723/jupyterhub/schema.yaml#L17-L22

You could make the schema tighter though, for example requiring a non-empty value for some fields etc. For example currently `proxy.hosts` is [allowed to be a blank array](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/6d3c13984d78791f6c6980a70ddf8cf0648cf723/jupyterhub/schema.yaml#L1742-L1751), while in basehub we require it to be set to something for a functional setup.

GeorgianaElena: @consideRatio, does this look ok https://github.com/2i2c-org/infrastructure/pull/2349/commits/89f445ee704747a5a0a1cedecf04c1e71ce64719?

helm-charts/basehub/templates/configmap-hub-templates.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: v1
+metadata:
+  name: hub-custom-templates-config
+  labels:
+    app: jupyterhub
+data:
+  {{- with .Values.jupyterhub.custom.homepage }}
+  HUB_TEMPLATES_REPO_URL: {{ .gitRepoUrl | quote }}
+  HUB_TEMPLATES_REPO_BRANCH: {{ .gitRepoBranch | quote }}
+  {{- end }}
```
consideRatio, line 11: Maybe rename to GIT_REPO_X now for consistency between naming? A bit nitty comment but I figure its a bit helpful to stay consistent for quick code comprehension

helm-charts/basehub/values.schema.yaml
```diff
@@ -216,20 +216,53 @@ properties:
...
     properties:
+      proxy:
+        type: object
+        additionalProperties: true
+        required:
+          - https
+        properties:
+          hosts:
+            type: array
+            minItems: 1
```
consideRatio, line 246: Excellent!!!

helm-charts/basehub/values.schema.yaml
```diff
@@ -282,7 +315,24 @@ properties:
             additionalProperties: false
```
consideRatio, line 315: Nice, thanks to this we can know that there was not a single typo of gitRepoUrl or similar!

helm-charts/basehub/values.schema.yaml
GeorgianaElena: Already caught an issue in utoronto: https://github.com/2i2c-org/infrastructure/actions/runs/4511216409/jobs/7943109492 yay 🎉

consideRatio (approved):
Nice!

I think thanks to the schema of `minValues` we can be certain that you have updated all hubs to have `jupyterhub.proxy.hosts` and `jupyterhub.ingress.tls` configuration.

I nitpicked about systematic naming for env vars, but this LGTM!

helm-charts/basehub/values.schema.yaml
consideRatio: Wiee nice yepp this looks great!!!

PR #2354, opened 2023-03-14
Playing with Roles and Groups on AVC

sean-morris:


config/clusters/cloudbank/avc.values.yaml
```diff
@@ -1,4 +1,38 @@
 jupyterhub:
+    load_groups:
```
consideRatio, line 2: This is JupyterHub Helm chart configuration, and it will error if you put `load_groups` here under `jupyterhub`.

You can see the [chart configuration reference](https://z2jh.jupyter.org/en/stable/resources/reference.html) for details, there is `hub.loadRoles` but not `hub.loadGroups`.

You can configure jupyterhub configuration directly via `hub.config.JupyterHub` though!

sean-morris: tracking this down! Thanks. You were on that one! I am wondering if there is any documentation on testing the hub configuration files locally?  It is embarrassing to commit all over the place trying to figure out how it all comes together.

config/clusters/cloudbank/avc.values.yaml
```diff
@@ -19,7 +19,28 @@ jupyterhub:
...
+        scopes: [read:groups!group=section1]
+        groups: [section1]
+      section2-student:
+        scopes: [read:groups!group=section2]
+        groups: [section2]
+      section1-teacher:
+        scopes: [read:users!group=section1, servers!group=section2]
+        users: [sean.smorris@gmail.com]
     config:
+      loadGroups:
```
consideRatio, line 35: This is jupyterhub (the software) native config, so it should be load_groups. The helm chart use camelCase for things while the jupyterhub python native config is camel_case

PR #2355, opened 2023-03-14
QCL staging Fix case for allowed_teams

pnasrat:
See #2287

### DIFF
--- a/config/clusters/qcl/common.values.yaml
+++ b/config/clusters/qcl/common.values.yaml
@@ -52,7 +52,7 @@ jupyterhub:
         populate_teams_in_auth_state: true
         allowed_organizations:
           - 2i2c-org:hub-access-for-2i2c-staff
-          - quantifiedcarbon:JupyterHub
+          - QuantifiedCarbon:jupyterhub
         scope:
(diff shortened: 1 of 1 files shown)


pnasrat:
Manually deployed and tested spawn page works for users in group via admin page

consideRatio (commented):
I expected us to not need to declare `allowed_teams` at all as I assumed it is an opt-in feature only to be used when we need to make a smaller set of users be allowed to see a profile list entry. Do you believe we currently are forced to declare `allowed_teams` to avoid a bug?

That feature is not part of KubeSpawner itself, but added by 2i2c. In my mind, if we needed to add it - we have a bug in that logic.

This is an incremental improvement on the current state, but ideally I'd love to see us avoid having `allowed_teams` not declared for a purpose of working around a bug or a comment to clarify its a workaround rather than something having an actual purpose for us.

consideRatio (approved):

pnasrat:
Thanks. Tomorrow I’ll test if it can be removed but wanted to unblock user testing. Particularly as I’m less familiar with the auth internals here.

PR #2376, opened 2023-03-19
docs: update aws upgrade docs based on k8s 1.24+ is used now

consideRatio:
Updates to docs based on some additional experience making EKS based cluster upgrades.


pnasrat (approved):
lgtm

PR #2377, opened 2023-03-19
gridsst: default to start on smallest node pool with min_nodes set to 1

consideRatio:
I happened to see that the smallest node pool was configured with `min_nodes` set to 1 - so one is always running. Due to that, it didn't make much sense to let the profile list options default to the medium node pool that needed to be started first.

I've not checked in with the gridsst community representative about this, but I consider it a bugfix that should be fine to just go for.


pnasrat (approved):
lgtm

consideRatio:
Thank you @pnasrat for your reviews!!!

PR #2385, opened 2023-03-20
Add admin user to cryocloud hub

yuvipanda:
Based on request by @tsnow03


yuvipanda:
Self merging, as it's just an admin list update: https://infrastructure.2i2c.org/en/latest/contributing/code-review.html#changing-active-infrastructure-in-general

PR #2389, opened 2023-03-21
Add image selector for CryoCloud

yuvipanda:
Ref https://2i2c.freshdesk.com/a/tickets/572


consideRatio (changes requested):
Currently we have prePuller enabled, singleuser.image specified, and these profile list options.

That makes a new node likeley to end up with a puller pod, and the user pod waiting for the image by the puller pod to be pulled before the kubelet/CRI or whatevers involved starts pulling the relevant image chosen.

I think we should:
- [ ] remove at least prePuller configuration, making it disabled by default via basehub
- [ ] remove singleuser.image specification because there is the profileList anyhow always picking an image

yuvipanda:
@consideRatio done!

consideRatio (approved):

PR #2390, opened 2023-03-21
Enable showing hidden files in CryoCloud

yuvipanda:
Fixes https://github.com/CryoInTheCloud/hub-image/issues/3

### DIFF
--- a/config/clusters/nasa-cryo/common.values.yaml
+++ b/config/clusters/nasa-cryo/common.values.yaml
@@ -63,6 +63,13 @@ basehub:
           scope:
             - read:org
     singleuser:
+      extraFiles:
+        jupyter_server_config.json:
+          data:
+            ContentsManager:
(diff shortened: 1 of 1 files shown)


config/clusters/nasa-cryo/common.values.yaml
```diff
@@ -63,6 +63,13 @@ basehub:
           scope:
             - read:org
     singleuser:
+      extraFiles:
+        jupyter_server_config.json:
```
consideRatio, line 67: A comment like this demystifies how this works quite a bit I think, and may help us avoid broken config due to changes in either basehub or here.

```suggestion
        # jupyter_server_config.json is defined by basehub, this entry adds to it
        jupyter_server_config.json:
```

consideRatio (approved):
LGTM - I added a code suggestion to add a comment also!

Issue #2026, opened 2023-01-06
[Decommission Hub] binder-staging


sgibson91:
### Summary

We had a crypto-mining incident on this hub ages ago. I attempted to lock it up with CILogon [in this PR](https://github.com/2i2c-org/infrastructure/pull/1579/) but was unsuccessful. And I've just spent the last bit of time trying with GitHub OAuth instead, but I just can't get it to work. My suspicion is that the docs [here](https://binderhub.readthedocs.io/en/latest/authentication.html) are outdated because I don't think the `jupyterhub.auth` key exists anymore...

Anyway, in the spirit of cleaning up, I propose we just get rid of this hub now.

#### Phase II - Hub Removal
(These steps are described in more detail in the docs at https://infrastructure.2i2c.org/en/latest/howto/hubs/delete-hub.html.)

- [ ] Manage existing data (migrate data from the hub or delete it)
- [ ] Remove the appropriate `config/clusters/<cluster_name>/<hub_name>.values.yaml` files. A complete list of relevant files can be found under the appropriate entry in the associated `cluster.yaml` file.
- [ ] Remove the associated hub entry from the `config/clusters/<cluster_name>/cluster.yaml` file.
- [ ] Remove the hub deployment
  - `helm --namespace HUB_NAME delete HUB_NAME`
  - `kubectl delete namespace HUB_NAME`
- [ ] Delete the hub's authentication application on auth0, GitHub or CILogon
- [ ] Also cleanup the deploy-hubs workflow file as I think there's some commented out code related to special-casing binder-staging in there too

yuvipanda:
I'd love for us to instead fixup the instructions - even if we don't run the pangeo binder from here, I think we will run other ones.

GeorgianaElena:
> I'd love for us to instead fixup the instructions - even if we don't run the pangeo binder from here

Opened https://github.com/jupyterhub/binderhub/pull/1646 to update binderhub auth instructions to match the working config in https://github.com/2i2c-org/infrastructure/pull/2393.

> even if we don't run the pangeo binder from here, I think we will run other ones.

Wondering if this is still true or if we should decommission it instead?

yuvipanda:
I still think it is worth it. And thank you for fixing that, @GeorgianaElena!

damianavila:
Given the authentication piece is fixed by @GeorgianaElena, I think there is value in keeping it alive since we are most likely deploying binder instances in the future. So, I am gonna close this one for now.

PR #2391, opened 2023-03-21
Enable auth for binder-staging

GeorgianaElena:
While manually deploying the changes in https://github.com/2i2c-org/infrastructure/pull/2277, I didn't realize that we shouldn't touch `2i2c-binder-staging` and that this hub doesn't get updates anymore, so I've deployed it 🤦🏼‍♀️ Then, I got sucked into debugging why we didn't manage to get authentication to work in the past and spent too much time on it, to not turn it into a PR, although I realize we want to decommission it? https://github.com/2i2c-org/infrastructure/issues/2026 

Developed based on @sgibson91's awesome work in https://github.com/2i2c-org/infrastructure/pull/1579 ❤️ 

CILogon authentication now works, but I'm not sure about the binder part, as build appears to hang. @2i2c-org/engineering try it out at https://binder-staging.2i2c.cloud/

Todos left:
- [ ] Update Binderhub's docs about running an authenticated binderhub https://binderhub.readthedocs.io/en/latest/authentication.html#enabling-authentication
- [ ] update `cilogon_app.py` to support a `binderhub` key and not just `basehub` and `daskhub`
- [ ] note somewhere that when creating the cilogon client, you need to provide the jupyterhub url and not the binderhub one
- [ ] docs?


GeorgianaElena:
@sgibson91, I only checked your draft PR after I had some progress with this one, because I had totally forgot about it. Sorry for not building on top of that and opening a separate one :/

Also, looking at it again, it looks like the only thing missing there was a misplaced `jupyterhub.hub.config.services` instead of `jupyterhub.hub.services` pff

GeorgianaElena:
Closing in favour of https://github.com/2i2c-org/infrastructure/pull/2393

PR #2395, opened 2023-03-21
[Bugfix] Fix pasadena config filenmae

GeorgianaElena:
Follow-up to #2277


config/clusters/cloudbank/cluster.yaml
```diff
@@ -169,7 +169,7 @@ hubs:
       # to the helm upgrade command in, and that has meaning. Please check
       # that you intend for these files to be applied in this order.
       - enc-pasadena.secret.values.yaml
-      - pasadena.secret.values.yaml
+      - pasadena.values.yaml
```
consideRatio, line 172: This file ought to be applied before the secret file. It may very well not break, but its very likeley the wrong order which make this configuration sensitive to things.

Issue #1923, opened 2022-11-15
[Billing] Automatically figure out how much a shared GCP cluster spends in a month (and map those costs to inner hubs)


damianavila:
## Context

Shared hubs have both a shared infrastructure component (kubernetes, prometheus, etc) and utilization of the hub, notebooks, dask-workers, etc.

Automate the manual process done up until March 2023. 

## Formula

Given a shared cluster in GCP (currently two-eye-two-see project) for each hub calculate monthly utilization (RAM model).
Calculate total RAM and Utilization proportion of each naamespace (shared utiliztaion) 
Calculate support cost from  kube-system + support cost 
For each hub Cost = Utilization proportion times total cost plus (Utilization proportion times total support cost)

pnasrat:
Notes

```prometheus
avg(
		label_replace(
			label_replace(
				avg(
					count_over_time(kube_pod_container_resource_requests{resource="memory", unit="byte", container!="",container!="POD", node!=""}[$__interval])
					*
					avg_over_time(kube_pod_container_resource_requests{resource="memory", unit="byte", container!="",container!="POD", node!=""}[$__interval])
				) by (namespace,container,pod,node) , "container_name","$1","container","(.+)"
			), "pod_name","$1","pod","(.+)"
		)
	) by (namespace)
```

https://grafana.pilot.2i2c.cloud/goto/SrIMEnBVk?orgId=1

pnasrat:
Timeselector 90d

```
sum by (namespace) (max_over_time(kube_pod_container_resource_requests{resource="memory",unit="byte",container!="",container!="POD", node!=""}[$__interval]))
```

Step 1d

https://grafana.pilot.2i2c.cloud/goto/rtyVU7fVz?orgId=1

pnasrat:
Opencost worked and was able to generate shared cluster breakdown however it's not calculating all the costs. It might not handle all our initial needs and there might be some bugs in it to fix.  

I've redeployed support from master to clean up for the weekend

```
kubectl get service -n support
NAME                                         TYPE           CLUSTER-IP     EXTERNAL-IP     PORT(S)                      AGE
support-grafana                              ClusterIP      10.3.250.43    <none>          80/TCP                       654d
support-ingress-nginx-controller             LoadBalancer   10.3.246.74    35.223.43.171   80:30226/TCP,443:30605/TCP   654d
support-ingress-nginx-controller-admission   ClusterIP      10.3.240.103   <none>          443/TCP                      654d
[...]

PR #2405, opened 2023-03-22
Add get_cluster_prometheus util method

pnasrat:
This method returns a requests.Session to allow querying the prometheus HTTP API

API Docs: https://prometheus.io/docs/prometheus/latest/querying/api/

See #1923


pnasrat:
Tested via ipython similar to the example in docstring.

yuvipanda (approved):

PR #2406, opened 2023-03-22
Add ability to give users / groups write access to buckets + multi-zone GPUs

yuvipanda:
GCS allows individual Google Users as well as Google Groups to have permissions to read / write to GCS buckets (unlike AWS). We can use this to allow community leaders to manage who can read and write to GCS buckets from outside the cloud by managing membership in a Google Group!

In this commit, we set up the persistent buckets of the LEAP hubs to have this functionality. Access is managed via a Google Group - I have temporarily created this under the 2i2c org and invited Julius (the community champion) as an administrator. But perhaps it should be just created as a regular google group. Using groups here allows management of this access to not require any 2i2c engineering work.

Future work would probably fold the separate variable we have for determining if a bucket is accessible publicly as an attribute as well.

Ref https://github.com/2i2c-org/infrastructure/issues/2096

TODO:
- [x] Test this actually works
- [x] Write documentation on when to use this and why, for 2i2c engineers
- [x] Write documentation on how end users can use this
- [x] Write documentation on how to create & manage google groups
- [x] Decide where this google group should live

Additionally, a request to make GPU nodes available in whatever zone possible in
the region we are in came in before this PR could be merged, and so was added on
here. This includes documentation on how to set up GPUs on GCP too!


jbusecke:
I just successfully tested this on one of the Lamont servers! Thanks a lot for making this work @yuvipanda.

yuvipanda:
@jbusecke has written useful documentation for this too!

pnasrat:
@yuvipanda what's needed to move this from draft to review - asking due to link with support tickets.

yuvipanda:
@pnasrat documentation! I'll try get this done over the next week.

GeorgianaElena (commented):
While trying to fix https://github.com/2i2c-org/infrastructure/issues/2696, I realized that the terraform config has lagged behind and then I discovered this PR.

@yuvipanda, is there anything I can do to help push this forward? Maybe we can track the documentation bit in another issue + PR?

yuvipanda:
@GeorgianaElena ok, I've pushed docs and this is ready to go! Sorry for the delay :( This has been apply'd already.

yuvipanda:
Thanks @GeorgianaElena!

PR #2407, opened 2023-03-22
deployer: Add support for binderhubs in the cilogon_app script

GeorgianaElena:
Follow-up to https://github.com/2i2c-org/infrastructure/pull/2393


pnasrat (approved):
lgtm

Issue #2336, opened 2023-03-11
Intervention to have new hubs not couple to the 2i2c-hubs-image, but something more up to date


consideRatio:
I want this issue to be focused around a swift intervention, not a final solution. The intervention I want to see is to not have new hubs be recommended to use the [2i2c-hubs-image](https://github.com/2i2c-org/2i2c-hubs-image), and instead use [pangeo/pangeo-notebook](https://github.com/pangeo-data/pangeo-docker-images).

I'm not sure if the 2i2c-hubs-image delivers something beyond pangeo/pangeo-notebook, but it comes a cost as we don't keep the 2i2c-hubs-image updated. Even if we tried to keep it updated, its almost impossible to have multiple communities use a single image. Some will get stuck relying on old versions, other will end up needing new versions.

Right now, this image has a [a requirements.txt](https://github.com/2i2c-org/2i2c-hubs-image/blame/main/requirements.txt) that includes for example pinning to the patch version of a ~two year old pinning of jupyter-resource-usage, even though there are new versions of that package. The same is true for several other packages, even `dask-gateway`.

In the spirit of helping open source software develop, we should strive towards using modern software and not get stuck with old versions. This way, we avoid having projects get reports about already resolved issues for example, and get reports on new issues that needs help being detected and resolved.

### Hubs about to couple to a 2i2c official image

- #2323 
- #2332 

### Related

- #1145

jmunroe:
I recently created
- https://github.com/2i2c-org/2i2c-hubs-image/issues/6

that is on a similar theme. I am not sure that the pangeo-notebook meets the needs of a good default (being Python only).  Also pangeo-notebook is a large image tailored to earth sciences. I wonder if using base-notebook could be a better default. I also liked that 2i2c-hubs-image includes Rstudio by default.

The idea of using base-notebook to as reference does have the great benefit of being kept up to date by a larger community than just 2i2c. Maybe it's possible to track pangeo/base-notebook as a dependency for 2i2c-hubs-image

Communities need to have a pathway to modify the default image we provide.  So 2i2c-hubs-image needs to follow the same instructions that we give in our template hub image repo.

Let's discuss this further at our next Product/Engineer meeting.

pnasrat:
Thinking about #2360 I also believe part of providing whatever base image we seettle on it would be beneficial to involve have some tests to catch reqressions for the common workflows/featuresets that our communties use so more frequently update images are possible to deploy with safety.

[...]

PR #2411, opened 2023-03-22
Undo Jupyter ServerApp usage to fix jupyter-nbcontrib-extensions

yuvipanda:
jupyter-nbcontrib-extensions required the following fixes to work:

1. Pin nbclassic<0.5 (done in https://github.com/2i2c-org/utoronto-image/pull/49, until https://github.com/Jupyter-contrib/jupyter_nbextensions_configurator/pull/152 is merged and released)
2. Upgrade jupyter-nbcontrib-extensions to latest released version (done in https://github.com/2i2c-org/utoronto-image/pull/48)
3. Stop using ServerApp, and use NotebookApp (https://github.com/Jupyter-contrib/jupyter_nbextensions_configurator/issues/153)

Ref https://github.com/2i2c-org/utoronto-image/pull/48

### DIFF
--- a/config/clusters/utoronto/default-common.values.yaml
+++ b/config/clusters/utoronto/default-common.values.yaml
@@ -3,6 +3,7 @@ jupyterhub:
     extraEnv:
       # Required to get jupyter-contrib-nbextensions to work
       # See https://github.com/2i2c-org/infrastructure/issues/2380
+      # Upstream issue at https://github.com/Jupyter-contrib/jupyter_nbextensions_configurator/issues/153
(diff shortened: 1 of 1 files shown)


yuvipanda:
Tested on utoronto staging. Will need to raise an upstream issue about jupyter_server migration causing more problems here.

pnasrat (approved):
lgtm

Thanks for the great work debugging this

pnasrat:
Should the same be done for the basehub image? I'm not sure how widely used extensions are

yuvipanda:
@pnasrat i left a comment in #2336 about what I think we should do with the current image there.

PR #2413, opened 2023-03-23
Bump image tag for cryocloud

yuvipanda:
Ref https://2i2c.freshdesk.com/a/tickets/578


yuvipanda:
Self-merging, as it is just an image bump.

PR #2416, opened 2023-03-23
deployer: add the option to pass the debug flag to helm when deploying hubs

GeorgianaElena:
Ref: https://github.com/2i2c-org/infrastructure/issues/2205#issuecomment-1453716299 and https://github.com/2i2c-org/infrastructure/issues/1917

Wished I had this when debugging #2349 😅

### DIFF
--- a/deployer/README.md
+++ b/deployer/README.md
@@ -91,9 +91,9 @@ It takes a name of a cluster and a name of a hub (or list of names) as arguments
 │                                    [default: None]                                                                   │
 ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
 ╭─ Options ────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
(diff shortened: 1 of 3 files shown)


pnasrat (changes requested):
This is great. If it's not too work I've  a suggestion for another flag that maps to `--dry-run --debug` or just allows `--dry-run` to also be set per

https://helm.sh/docs/chart_template_guide/debugging/

GeorgianaElena:
> This is great. If it's not too work I've a suggestion for another flag that maps to --dry-run --debug or just allows --dry-run to also be set per

Yep, this is a very good idea. I will do that.

I also realized that I also missed updating the deployer readme.

pnasrat (approved):
lgtm

consideRatio (approved):
LGTM and should be fine to iterate on later without trouble if we decide to change something in this implementation.

Issue #2396, opened 2023-03-21
LEAP Investigation to intermittent failing workloads for dask


pnasrat:
### Context

via https://2i2c.freshdesk.com/a/tickets/561

After some initial investigation focused on a specific event (that turned out to be node shutdown), the intermittent failures 

I have asked for the following to auid systematic debugging:

1. ​How this manifests symptomatically (log text, dashboard messages, etc)
2. Sample code that mirrors the setup for the failing workloads  - see below

I also asked for 

> Do you have an example workload dask code/configuration that we could use to attempt to run to attempt to replicate and debug various hypothesis as to why jobs are not completing. Such an example should include the same pattern of resilience handling https://examples.dask.org/resilience.html https://distributed.dask.org/en/latest/resilience.html#hardware-failures that the workloads that are failing use.

### Proposal

Identify hypothesis - memory spikes, spot instances, network capacity for the failure. 
Create automated test that hits the issue as best we can. This may require multiple runs

Change one thing, test,

In addition further monitoring, logging, or tracing might be added for more visibility

consideRatio:
There are multiple paths to addressing this as I see it, but I think the most sustainable path is to help ensure our dask-gateway setup starting a dask scheduler pod and many dask worker pods (using the user environment image by default) terminate gracefully no matter if they are terminated by memory pressure on the node or a spot instance is shut down by the cloud provider.

### Is the SIGTERM propegated to the dask worker?

I know that depending on how a container is started via k8s `command` and `args` and/or the equivalent Dockerfile's `ENTRYPOINT` and `CMD`, there can be a failure to propegate a SIGTERM signal onwards to the dask worker software running in the container, which can make the dask scheduler software running in another pod see it as a loss of contact as compared to a unexpected but known failure.

### Related

- [dask/dask-gateway](https://github.com/dask/dask-gateway), startups a dask cluster (dask scheduler pod + dask worker pods, running the user image by default), also likeley involves the k8s `command` and `args` for started scheduler and worker pods
- [dask/distributed](https://github.com/dask/distributed), defines the dask scheduler and dask worker stuff - I don't know details about this overall

pnasrat:
@consideRatio thanks for the additional context, and questions for my investigation. 

[...]

PR #2419, opened 2023-03-24
Make LEAP dask workers be not pre-emptible

yuvipanda:
Possible way to reduce confounding factors for
https://github.com/2i2c-org/infrastructure/issues/2396

### DIFF
--- a/terraform/gcp/projects/leap.tfvars
+++ b/terraform/gcp/projects/leap.tfvars
@@ -84,6 +84,7 @@ dask_nodes = {
   "medium" : {
     min : 0,
     max : 200,
+    preemptible: false,
     machine_type : "n2-highmem-16",
     labels : {},
     gpu : {
--- a/terraform/gcp/projects/leap.tfvars
+++ b/terraform/gcp/projects/leap.tfvars
@@ -84,6 +84,9 @@ dask_nodes = {
   "medium" : {
     min : 0,
     max : 200,
+    # Disable preemptive nodes for dask so we can remove possible complications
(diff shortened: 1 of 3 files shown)


consideRatio (approved):
LGTM!

Can you add a comment to the leap.tfvars about deviating from the default of `preemptible: true` to `false` for leap? I think its good if this intentional workaround stands out

terraform/gcp/projects/leap.tfvars
```diff
@@ -84,6 +84,7 @@ dask_nodes = {
   "medium" : {
     min : 0,
     max : 200,
+    preemptible: false,
```
pnasrat, line 87: Can you add a comment linking back to the logic for this change

pnasrat (changes requested):
minor documentation nit will approve once fixed

terraform/gcp/projects/leap.tfvars
pnasrat: oh I see @consideRatio and I collided on this review with same feedback!

consideRatio:
Btw they may very well need a cloud quota increase for on-demand CPU now. pre-emptible/spot CPU quotas are separate from on-demand I believe, so they may run out on on-demand now that its the only kind of CPU used.

yuvipanda:
Am also going to wait for the LEAP credits issue to be resolved before deploying this.

yuvipanda:
Leap credits are done now

pnasrat (approved):
lgtm

PR #2429, opened 2023-03-28
Add grafana bigquery datasource plugin

pnasrat:
Plugin https://grafana.com/grafana/plugins/grafana-bigquery-datasource/

Helm installation docs https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml#L520

### DIFF
--- a/config/clusters/2i2c/support.values.yaml
+++ b/config/clusters/2i2c/support.values.yaml
@@ -26,3 +26,5 @@ grafana:
       - secretName: grafana-tls
         hosts:
           - grafana.pilot.2i2c.cloud
+  plugins:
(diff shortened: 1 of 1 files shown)


yuvipanda (approved):
This is pretty awesome, especially if we can also figure out how to get auth done.

yuvipanda:
There is also the question of permissions - there are people who have Grafana access who do not have GCP access. Something to consider, as we don't want to accidentally leak how much one project is spending to another project.

pnasrat:
Nod yes this just pulls the plugin and will allow me to try do a tranformation across prometheus and bigquery. If we go this route for billing visualization permissions are definitely part of that story

Issue #2339, opened 2023-03-12
nasa-veda: requested password change after 60 days :scream:


consideRatio:
I just signed into the nasa-veda associated AWS account and was requested to update my password after signing in with 2FA. I expect this to cause a few of us to get locked out or disrupted 6+ times a year due to this strict policy.

![image](https://user-images.githubusercontent.com/3837114/224574825-0dc8efb4-ffb2-4923-9c21-eaf95ec1ad3f.png)

![image](https://user-images.githubusercontent.com/3837114/224574762-bfbc2b88-bd07-472f-8bda-01e851941535.png)

![image](https://user-images.githubusercontent.com/3837114/224575071-35eebed7-841f-4e8f-b4c9-f4f38c9bcb31.png)

Is this something we should embrace or push back against? If we embrace it, I'd like advice on how to handle these resets practically.

damianavila:
60 days feel a little bit annoying... I would probably advocate for once a year if they really want to have it.

yuvipanda:
My view is that:

1. This is super annoying, and insecure
2. Based on my experience working with NASA, the only hammer we have is that [NIST Guidelines no longer recommend password expiry](https://www.bleepingcomputer.com/news/security/aligning-your-password-policy-enforcement-with-nist-guidelines/)
3. Based on my personal experience, I expect this to trickle down into actual NASA policy in a few years
4. We all have limited energy in pushing for 'right' changes, that cause death by a thousand papercuts.
5. My suggestion is to feel horrified, but change the password. I suspect pushing back will not actually lead to change, and will only cause deep frustration on our part.

yuvipanda:
I opened https://github.com/2i2c-org/meta/issues/525 to make sure we can contractually get compensated for this too

yuvipanda:
https://github.com/2i2c-org/infrastructure/issues/2434 is documentation of how i rotated these

consideRatio:
I've got a response back from Aaron who confirms that this is indeed a NASA security policy that the SMCE team (Science Managed Cloud Environment) can't deviate from.

PR #2432, opened 2023-03-29
Fix nasa-veda SSO link in hubs page

pnasrat:
Adds correct account field to cluster config

Tested by local docs build and inspecting docs/tmp/hub-table.csv


pnasrat:
Deploy will fail due to #2339

PR #2433, opened 2023-03-29
Regenerate nasa veda deployer credentials

yuvipanda:
Looks like these also need to be regenerated after every 60 days or so?

Ref https://github.com/2i2c-org/infrastructure/issues/2339


pnasrat:
can you add a calendar event to remind us or a cron job to poke us to do this too

pnasrat (approved):
lgtm 

We should also sort the method of reminding ourselves but it's more important to get this in

Issue #2445, opened 2023-03-30
[utoronto] Investigate report of slow/unresponsive hub


pnasrat:
### Context

via https://2i2c.freshdesk.com/a/tickets/608

The U of T JupyterHub seems to be slow and/or unresponsive right now. I
got a few reports from some TAs and students just now, and confirmed it
myself.

pnasrat:
Some server start failures
https://grafana.utoronto.2i2c.cloud/d/hub-dashboard/jupyterhub-dashboard?orgId=1&from=1680112235570&to=1680198635570&viewPanel=13

pnasrat:
One node was slammed on CPU

https://grafana.utoronto.2i2c.cloud/d/Qx58n92nk/cluster-information?from=1680113838028&to=1680200238028&viewPanel=9&orgId=1

![image](https://user-images.githubusercontent.com/13798/228928064-e200b4d5-394e-49c0-9a2a-a95c465b3ea0.png)

pnasrat:
Note debug shell can be obtained via

```
kubectl debug -it  node/aks-nbdefault-34239724-vmss0000nw  -n prod --image=ubuntu
```

yuvipanda:
I notice that the JupyterHub pod itself has had 16 restarts - this is probably what triggered the pagerduty alert.

It restarted because *the hub automatically restarts upon 5 spawn failures*.

```
[C 2023-03-30 17:05:15.030 JupyterHub base:1046] Aborting due to 5 consecutive spawn failures
[E 2023-03-30 17:05:15.030 JupyterHub gen:630] Exception in Future <Task finished name='Task-4212' coro=<BaseHandler.spawn_single_user.<locals>.finish_user_spawn() done, defined at /usr/local/lib/python3.11/site-packages/jupyterhub/handlers/base.py:963> exception=TimeoutError("Server at http://10.244.10.227:8888/user/570308c0-141b-4dd6-b816-18cf4d4b1716/ didn't respond in 120 seconds")> after timeout
    Traceback (most recent call last):
      File "/usr/local/lib/python3.11/site-packages/tornado/gen.py", line 625, in error_callback
        future.result()
      File "/usr/local/lib/python3.11/site-packages/jupyterhub/handlers/base.py", line 970, in finish_user_spawn
        await spawn_future
      File "/usr/local/lib/python3.11/site-packages/jupyterhub/user.py", line 863, in spawn
        await self._wait_up(spawner)
[...]

PR #2446, opened 2023-03-30
Set a CPU limit for user pods in UToronto

yuvipanda:
https://github.com/2i2c-org/infrastructure/issues/2445 had an outage because a node was taken out by having its 8 CPU be fully utilized. This restricts a single user from being able to do that. Multiple users running heavy stuff still can, but this is an improvement!


pnasrat (approved):
lgtm

A CPU limit not being set suprised me!

PR #2450, opened 2023-03-31
prometheus-node-exporter: declare cpu/mem requests and limits

consideRatio:
We've seen the node-exporter pod be evicted because it has no resource requests/limits when what needs to be evicted are other pods to resolve the situation. Let's not allow that to happen!

- This relates to https://github.com/2i2c-org/infrastructure/issues/90, about setting requests and limits for support chart's pods.

### CPU (leap, 30 last days)

Note that the use dropped at some point. That was me upgrading the GKE based k8s cluster in #2337.

image

### Memory (leap, 30 last days)

image

### Why I choose the requests and limits

1. We don't want this pod to starve other pods of CPU if it bursts, but it seems it can only burst to ~3m CPU. Setting a limit on 5m CPU seemed fine.
2. We want to ensure the typical operation is guaranteed CPU wise, which seem to be ~0-3m CPU. Setting a request of 5m CPU seemed fine.
3. We want requests/limits of memory to be the same to avoid getting evicted by memory, and crashing reliably if this is too low instead. I saw that it has peaked at ~21Mi, so I went for 30Mi to be safe, knowing there is a few hundred Mi of memory available still to not cause issues with planned requests/limits in #2121.

### DIFF
--- a/helm-charts/support/values.yaml
+++ b/helm-charts/support/values.yaml
@@ -100,6 +100,22 @@ prometheus:
       - key: k8s.dask.org/dedicated
         value: worker
         effect: NoSchedule
+    # resources for the node-exporter was set after inspecting cpu and memory
+    # use via prometheus and grafana.
+    #
+    # node-exporter is typically found using between 0-3m CPU and 2-22Mi memory.
+    #
+    # PromQL queries for CPU and memory use:
+    # - CPU:    sum(rate(container_cpu_usage_seconds_total{container="node-exporter", namespace="support"}[5m])) by (pod)
+    # - Memory: sum(container_memory_usage_bytes{container="node-exporter", namespace="support"}) by (pod)
+    #
+    resources:
+      limits:
+        cpu: 5m
+        memory: 30Mi
+      requests:
+        cpu: 5m
+        memory: 30Mi
 
   networkPolicy:
     enabled: true


helm-charts/support/values.yaml
```diff
@@ -100,6 +100,26 @@ prometheus:
       - key: k8s.dask.org/dedicated
         value: worker
         effect: NoSchedule
+    # resources for the node-exporter was set after inspecting cpu and memory
+    # use in a few clusters using prometheus queries from
```
pnasrat, line 104: Could you maybe add a note that upstream don't set deliberately leave setting this to user choice as per

https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-node-exporter/values.yaml#L53

helm-charts/support/values.yaml
```diff
@@ -100,6 +100,26 @@ prometheus:
...
+    # https://gist.github.com/payam-int/edf977c6af603fee0ce1b05da7792fe7.
```
pnasrat, line 105: remove the trailing . so the link can be copy pasted easily

helm-charts/support/values.yaml
```diff
@@ -100,6 +100,26 @@ prometheus:
...
+    #
+    # - Memory query:
+    #   100 * (1 - ((avg_over_time(node_memory_MemFree_bytes[10m]) + avg_over_time(node_memory_Cached_bytes[10m]) + avg_over_time(node_memory_Buffers_bytes[10m])) / avg_over_time(node_memory_MemTotal_bytes[10m])))
```
pnasrat, line 108: Should we have things like  this graphed in grafana?

pnasrat (changes requested):
Minor changes for comments/docs

This generally makes sense though.

helm-charts/support/values.yaml
consideRatio: If you do `helm create test` and and `cat test/values.yaml` you will find this comment - its a boilerplate comment from `helm create` rather than the prometheus community maintaining that chart.

Our support chart isn't packaged and distributed for others to install, its used as our bundled configuration of various charts we want to install with misc common configuration.

Do you have a concrete suggestion of a comment that you think should be added?

---

This is from `helm create test`'s values.yaml file:

```yaml
resources: {}
  # We usually recommend not to specify default resources and to leave this as a conscious
  # choice for the user. This also increases chances charts run on environments with little
  # resources, such as Minikube. If you do want to specify resources, uncomment the following
  # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
  # limits:
  #   cpu: 100m
  #   memory: 128Mi
  # requests:
  #   cpu: 100m
  #   memory: 128Mi
```

helm-charts/support/values.yaml
consideRatio: I realized that what I've copied from there was cpu/memory usage of node exported by node exporter, rather than the usage of the node-exporter pod itself. Due to this, I've reworked this PR and no longer link out.

helm-charts/support/values.yaml
consideRatio: Yes maybe, but also maybe not develop the dashboards for this ourselves. I figure for now, its helpful if there is a quick and easy trace for how to come up with data to decide on other resource requests/limits set in the support chart.

consideRatio:
Thank you for reviewing @pnasrat! I realized that I had copy pasted incorrect prometheus queries so I drew the wrong conclusions. It turns out this pod is peaking at very low CPU (~3m) and memory (~21Mi), so I've made updates on queries, screenshots, and requests/limits etc.

helm-charts/support/values.yaml
pnasrat: Ok. Just ignore this request

pnasrat (approved):
lgtm

consideRatio:
Thank you @pnasrat for reviewing!!

PR #2452, opened 2023-03-31
grafana: declare cpu/mem requests and limits

consideRatio:
We've seen misc pods be evicted because they has no resource requests/limits and are considered as prime candidates for eviction during node pressure. This is not a big deal, but we can set some resource requests and limits to avoid the situation.

- This relates to https://github.com/2i2c-org/infrastructure/issues/90, about setting requests and limits for support chart's pods.

### CPU (leap, last 6 hours)

I've used this grafana instance a lot recently. This load is the peak load over the year so far.

image

### Memory (leap, last year)

image

### DIFF
--- a/helm-charts/support/values.yaml
+++ b/helm-charts/support/values.yaml
@@ -188,6 +188,31 @@ grafana:
     namespaced: true
     pspEnabled: false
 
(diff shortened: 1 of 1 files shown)


yuvipanda (approved):
We can always revise if needed

PR #2455, opened 2023-04-03
Remove pydata-sphinx-theme fix

pnasrat:
Needs 2i2c-org/sphinx-2i2c-theme#17 for sphinx-book-theme fix

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -6,6 +6,4 @@ sphinx-autobuild
 sphinx-copybutton
 sphinx-design
 # Pin pydate-sphinx-theme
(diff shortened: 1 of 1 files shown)


choldgraf:
Closing and re-opening to retrigger the RTD build...let's see if it's fixed now

choldgraf:
The docs preview correctly looks like it is 1.0.1 and the build worked, so I'll merge this!

### Future reference: edit the sphinx-2i2c-theme

In case somebody stumbles on this PR in the future, when we need to pin theme versions, we should do so in https://github.com/2i2c-org/sphinx-2i2c-theme so that all of our docs inherit the fix when they are re-built

PR #2458, opened 2023-04-03
Add ephemeral hub for HackAnExoplanet

yuvipanda:
- Uses tmpauthenticator to create binder like temporary users. This was created for the older tmpnb.
- Provide a *shared* username & password for the entire hub at the nginx level, using the same mechanism we use for protecting our prometheus. This wards off the cryptobros.
- We turn off all persistent storage, so nothing is stored
- Add documentation on ephemeral hub

Ref https://github.com/2i2c-org/infrastructure/issues/2461


pnasrat:
Could you add some background/context on this beyond implementation choices - what is this for, why we're creating, etc

yuvipanda:
@pnasrat I've tied this to a particular hub being deployed, and added more specific documentation as well now!

pnasrat:
Maybe a longer term issue but if this is something that frequently gets deployed I'd like to see a more robust user system than shared http basic auth. Foir example off the top of my head might be providing an invite link mechanism maybe extending/building on https://github.com/jupyterhub/firstuseauthenticator with a signup flow of workshop admin signup flow, invite generation (guards against replay), to create account on ephemeral hub.

There is also a risk of a malicious signup to do crypto generation anyway as http basic auth is easy to automate, so not sure how much this design mitigates.

helm-charts/basehub/values.schema.yaml
```diff
@@ -19,7 +19,27 @@ required:
   - userServiceAccount
   - dex
   - staticWebsite
+  - ingressBasicAuth
 properties:
+  ingressBasicAuth:
+    type: object
+    additionalProperties: true
```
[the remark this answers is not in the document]
yuvipanda: @consideRatio good catch, was left over from debugging!

yuvipanda:
@pnasrat yeah, a better auth system would definitely be more valuble. I don't think any of the existing ones fit well though :( Some sort of ephemeral and pseudonymous invite-link based system would be great. Don't think we should block on that for now, though.

pnasrat (approved):
lgtm

helm-charts/basehub/values.yaml
```diff
@@ -5,6 +5,14 @@ userServiceAccount:
   enabled: true
   annotations: {}
 
+ingressBasicAuth:
+  enabled: false
+  # Primarily here for validation to 'work',
+  # as these are set in secret config otherwise. I don't like this,
+  # as we won't catch these values missing if they aren't set.
+  username: ""
+  password: ""
```
consideRatio, line 14: This can be handled in the helm templates rendering conditionally on `.Values.ingressBasicAuth.enabled` by having a pipe after the reference to `.Values.ingressBasicAuth.username` like this: `| required "ingressBasicAuth.username is required"`

So for example you can have something like this in the template.

```yaml
  auth: {{ htpasswd (.Values.ingressBasicAuth.username | required "ingressBasicAuth is enabled, a username is required" ) (.Values.ingressBasicAuth.password | required "ingressBasicAuth is enabled, a username is required") | b64enc | quote}}
{{- end }}
```

Another option is to let this be validated in the jsonschema, but no matter what this will end up being caught during helm template rendering.

consideRatio: An example like this can be found in z2jh here: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/034c3a4ec71710dbba2d605361187003156d6b92/jupyterhub/values.schema.yaml#L81-L173

helm-charts/basehub/templates/ingress-auth/secret.yaml
```diff
@@ -0,0 +1,9 @@
+{{- if .Values.ingressBasicAuth.enabled -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: ingress-basic-auth
+type: Opaque
+data:
+  auth: {{ htpasswd .Values.ingressBasicAuth.username .Values.ingressBasicAuth.password | b64enc | quote}}
```
consideRatio, line 8: nit detail

```suggestion
  auth: {{ htpasswd .Values.ingressBasicAuth.username .Values.ingressBasicAuth.password | b64enc | quote }}
```

helm-charts/basehub/values.yaml
yuvipanda: @consideRatio I'd like to do this in a follow-up PR, given that the ESA hub needs to be out shortly. We also already use the empty string trick elsewhere, so should probably be fixed alltogether.

yuvipanda: I opened https://github.com/2i2c-org/infrastructure/issues/2494

PR #2462, opened 2023-04-05
Update helm charts every month, not every week

yuvipanda:
Doing https://github.com/2i2c-org/infrastructure/pull/2453 every week takes too much work I think.

### DIFF
--- a/.github/workflows/bump-helm-versions.yaml
+++ b/.github/workflows/bump-helm-versions.yaml
@@ -11,7 +11,7 @@ on:
         default: false
         required: false
   schedule:
-    - cron: "0 0 1 * * # Run first of every month at 00:00 UTC
+    - cron: "0 0 1 * *" # Run first of every month at 00:00 UTC
 
 env:
   team_reviewers: engineering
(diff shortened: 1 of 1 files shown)


.github/workflows/bump-helm-versions.yaml
```diff
@@ -11,7 +11,7 @@ on:
         default: false
         required: false
   schedule:
-    - cron: "0 0 * * 1" # Run every Monday at 00:00 UTC
+    - cron: "0 0 1 * * # Run first of every month at 00:00 UTC
```
sgibson91, line 14:
```suggestion
    - cron: "0 0 1 * *" # Run first of every month at 00:00 UTC
```

consideRatio (approved):
With Sarah's syntax bugfix, this LGTM!

Its also worth noting that whenever we know we want to bump something, we can always trigger this workflow thanks to `on.workflow_dispatch` is defined via https://github.com/2i2c-org/infrastructure/actions/workflows/bump-helm-versions.yaml.

image

With this in mind, a big :+1: for not doing this more often than once a month.

pnasrat:
Adding to the general lgtm - looks like the docs just say schedule or on dispatch so don't think a docs change is needed.

https://infrastructure.2i2c.org/en/latest/reference/ci-cd/auto-bumping.html

PR #2464, opened 2023-04-05
Adding dummy docs to test PR merging

ianabc:
Throwaway PR to test review workflow

### DIFF
--- a/config/clusters/callysto/common.values.yaml
+++ b/config/clusters/callysto/common.values.yaml
@@ -109,6 +109,8 @@ jupyterhub:
     config:
       EmailAuthenticatingCILogonOAuthenticator:
         allowed_domains:
+          # Email addresses must originate in one of these domains
+          # to log in to the callysto hub
           - 2i2c.org
           - btps.ca
           - callysto.ca


yuvipanda:
Note for others: we are testing @ianabc's access to this repo. https://2i2c.freshdesk.com/a/tickets/621 has more context.

ianabc (commented):
It looks like I can comment, but I can't approve or request changes.

yuvipanda:
Hmm, I see you're a member of the callysto team: https://github.com/orgs/2i2c-org/teams/callysto

And it does have 'Write' rights:

image

ianabc:
I'm not sure, when I try to leave a review it says that pull request authors can't approve their own pull requests

Screen Shot 2023-04-06 at 09 01 39

yuvipanda:
@ianabc I've given you 'maintain' permissions. Try now?

ianabc:
No, it still says the same thing

damianavila:
>  or to bump the users permissions to the level that it can be bypassed.

I think this option might help...

Screen Shot 2023-04-11 at 18 22 02

on the branch protection rule, but I am a little bit worried about the implications of that permission granting.
If we are ok with this, then I would suggest going back to the "write" level (instead of admin) and then allowing the exception on the branch protection rule.

ianabc:
With the current settings it looks like I _could_ merge this PR
Screen Shot 2023-04-12 at 08 15 46

but

damianavila:
@ianabc, to confirm, the option to "merge without waiting for requirement..." was only available after @yuvipanda raised your permission to the admin level, correct? Or did you see that option also with the "write" and "maintain" levels?

ianabc:
Yes, I think that was the case. At the time I closed it I was able to merge (I just chose not to). I think at that time I was in the admin role.

Issue #2430, opened 2023-03-28
Investigate many users' servers stopping simultaneously during cryocloud tutorial


yuvipanda:
From @tsnow03 on slack:

> The CryoCloud hub has just shut down for all of our users during our tutorial.

> We did have some commands run by a tutorial lead that may have messed up our server:

![image](https://user-images.githubusercontent.com/30430/228288244-f2013739-84a4-4d05-9737-ad783542c217.png)

> Then we shut down our server and it took a while for that server to clear and for us to be able to start a new one. In the mean time, most people in the tutorial were kicked off. But we were able to get back up and running fine after that. Thinking through how it happened, it brings up a few questions on our new shared server system.
> 1. We currently tell our users to go to hub control and stop server before logging off. With the shared servers, does this shut down a server for everyone sharing it too?
> 2. Also, if my server gets stuck and needs to be shut down, I'm guessing everyone else I'm sharing with also experiences the loss of their server? (edited) 

> On my end, it was a cell where the pip install/import cell never stopped running - kernel said it was busy but wouldn't interrupt out of it. After a couple kernel restarts, the cells weren't responding and kernel continued to say Connecting so we stopped server and logged off. We had a few tries at starting the server again from the login page and it gave the "Your server is stopping. You will be able to start it again once it has finished..." error. That is when many people in the room started saying they were being kicked off.

> I need to check with people to find out what their errors were because I was a little too distracted to check other people's computers in that moment. I'm sure my personal server issue was the bad cell commands and maybe an overload, and was mostly the server being slow to shutdown, but I was thrown off by a bunch of people losing their servers. (edited)

tsnow03:
People were able to log back in but it took a few minutes. I just heard back that our users experienced a window popping up about the server. I think it was the Server Unavailable or Unreachable window that pops up after a person stops their server.

yuvipanda:
In local time (to me, UTC+0530), the issue was reported on slack at 9:11 PM.

Looking at Grafana (https://grafana.cryointhecloud.2i2c.cloud/d/iqt8mUSVk/cluster-information?orgId=1&from=1679657257498&to=1679694542140) I see:

<img width="1666" alt="image" src="https://user-images.githubusercontent.com/30430/229051632-5a6f7913-acd5-4a77-89db-e67cee5009fe.png">

[...]

PR #2470, opened 2023-04-07
nasa-cryo: default node share requests to 4GB/16GB for small/medium

consideRatio:
Opened based on discussion in https://github.com/2i2c-org/infrastructure/issues/2430#issuecomment-1499775619


yuvipanda (approved):
I like this better!

Issue #2201, opened 2023-02-14
Integrate JMTE hub into our current infrastructure


yuvipanda:
The Jupyter Meets the Earth grant is what is currently funding my time, and I'm going to work with @fperez to bring https://github.com/2i2c-org/infrastructure/pull/436 into the 2i2c infrastructure. The current hub + that PR was the amazing work of @consideRatio, blazing several trails that have led our current infrastructure to be where it is at. This issue tracks the various bits needed to integrate that hub into our infrastructure as a regular hub so it can be supported via our regular means.

## TODO

- [x] Make sure all 2i2c engineers have access to the JMTE AWS Account
  - [x] yuvipanda
  - [x] Erik
  - [x] Pris
  - [x] Georgiana
  - [x] Sarah
  - [x] Damian
  - [x] James
- [x] Create a generic daskhub terraform infrastructure on AWS on the JMTE AWS Account
- [x] Use [terraform import](https://developer.hashicorp.com/terraform/cli/import) to import the EFS filesystem and the S3 buckets in use.
- [x] Setup a new eksctl based cluster
- [x] Deploy JupyterHubs on this cluster *identical* to the config from https://github.com/2i2c-org/infrastructure/pull/436
- [x] Consider copying the prometheus data from existing cluster to new cluster
- [x] Co-ordinate with current users to move them over to new cluster
- [ ] 2i2c-org/features#20
- [ ] 2i2c-org/infrastructure#2512
- [x] Decomission existing cluster
- [ ] Quick audit of existing resources to make sure they're clean and there isn't anything unwanted
   - [x] EC2 instances
   - [x] EBS Volumes
   - [x] EFS file systems
   - [ ] S3 buckets

damianavila:
@yuvipanda, do we have a deadline for this one to happen? Just trying to figure it out a timeline...

consideRatio:
> Add SFTP service to basehub, as that service is used to bring data in and out of home directories. JupyterHub SSH itself doesn't seem to be used much, so we can probably ignore that one.

If there is agreement that we should support this, then this is what I suggest as a solution to this ticket for the LIS basehub in the shared 2i2c cluster.

Does that seem reasonable? If so, I suggest we extract that out as a dedicated issue as that work item is large enough on its own and should be easy to self-contain I think.

consideRatio:
I've opened https://github.com/2i2c-org/infrastructure/issues/2208 with this

yuvipanda:
[...]

PR #2474, opened 2023-04-10
Integrate JMTE hub to our existing infrastructure

yuvipanda:
- Use 'jupyter-meets-the-earth' rather than jmte as name, because the existing cluster is already called 'jmte'.
- Replicates config from https://github.com/2i2c-org/infrastructure/pull/436/files
  to the extent possible
- Uses our IRSA config for AWS permissions, rather than the eksctl created service account in use earlier.
- Uses CILogon+GitHub for authentication, rather than auth0+github
- Re-use the same EFS filesystem from before, avoiding the need to copy a few terabytes of data around
- Hub is now at jmte.2i2c.cloud, and the old URL (hub.jupyterearth.org) redirects here. Same for staging.
- Existing sqlite db has been copied over to the new cluster, so all existing users have access. The JupyterHub
  admin panel was used to manage access, so the hub was the source of user lists
- No prometheus existed on old cluster, so no prometheus data needs to be copied over.
- The staging hub config matches the prod hub config to the extent possible, matching rest of our infrastructure.
- SFTP service has been *removed* for now.
- Fix typo in our domain redirects code that prevented more than one redirect from being set

Ref https://github.com/2i2c-org/infrastructure/issues/2201


yuvipanda:
I've deployed and set all this up already, as the existing cluster had gone down and I wanted to prevent disruptions.

consideRatio (approved):
Great work on this @yuvipanda!! LGTM!!!

consideRatio:
Like for new hub setups, we want to provide permissions to work on this aws account for 2i2c engineers.

So, sending out invites is relevnt as well, but not directly related to this PR

config/clusters/jupyter-meets-the-earth/cluster.yaml
```diff
@@ -0,0 +1,27 @@
+name: jupyter-meets-the-earth
```
consideRatio, line 1: Can you add a cloud console link like other cluster.yaml files has here?

config/clusters/jupyter-meets-the-earth/cluster.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - name: staging
+    domain: staging.hub.jupytearth.org
+    helm_chart: daskhub
+    helm_chart_values_files:
+      - common.values.yaml
+      - staging.values.yaml
+      - enc-staging.secret.values.yaml
+  - name: prod
+    display_name: "Jupyter Meets the Earth"
+    domain: hub.jupytearth.org
```
consideRatio, line 22: Hmmm should this be jmte.2i2c.cloud?

I got cert issues visiting hub.jupytearth.org

yuvipanda: Ah, changed, but not sure why the redirects aren't working. Checking that now.

yuvipanda: @consideRatio good catch, and fixed!

Issue #2005, opened 2022-12-14
Callysto core nodepool rightsizing


ianabc:
### Context

The callysto hub infrastructure is currently running on a pair of `n1-highmem-4` (GCE) nodes which are relatively expensive, and I was asked to check that these are the right size for the job. Poking around in our grafana stats, it looks like they might be a bit underutilized, would it make sense to resize these to `n1-highmem-2` nodes? I think the [default cluster config uses n1-highmen-2](https://github.com/2i2c-org/infrastructure/blob/333d70d17142ca06c1110228bd31b7ffbcaf5099/terraform/gcp/variables.tf#L121) with the caveat 

> For single-tenant clusters, a single n1-highmem-2 node seems enough - if network policy and config connector are not on.  For others, please experiment to see what fits.

but we override that in our [current terraform config](https://github.com/2i2c-org/infrastructure/blob/333d70d17142ca06c1110228bd31b7ffbcaf5099/terraform/gcp/projects/callysto.tfvars#L21). Are there any reasons for us _not_ to resize to `n1-highmem-2` nodes?

If it is possible for us to do this, do we just propose the change to the terraform config? I guess applying the change would tear down our current cluster and redeploy it; that would be OK, but I would the user data (FileStore volume) be preserved and re-attached?

### Proposal

Explore the possibility of resizing the core node pool of the Callysto cluster. If it is feasible, redeploy.

yuvipanda:
@ianabc we can try it out! However, I am curious if that is enough for the prometheus nodes.

ianabc:
Thanks, I think we might give it a go and see what happens. My understanding is we would make the update in terraform and then issue a PR, but I'm a little fuzzy on what will happen when we merge the PR. I think it has to destroy the existing code nodepool and create a new one, but can you confirm that it would leave the user storage intact? i.e. when it adds the new core nodepool and spins things up would existing users see their existing files? Or should we arrange to have them backed up somewhere then reimported?

PR #2488, opened 2023-04-13
Use of highmem nodes for new clusters node pools and require a few tf variables to be set

consideRatio:
I've deliberated quite a bit on node pool's instance types, and I think these are good defaults for node pools in new k8s clusters to provide.

I've updated documentation a bit about the choices made for the core node pool specifically. The documentation is certainly not exhausting the discussion or is perfectly written, but I hope it can be seen as an incremental improvement.

### Summary of changes to node pools

GKE:
- `n2` is used systematically over `n1`. They provide a bit better performance per CPU and a 1:8 ratio of CPU:Memory instead of 1:6.5
- `n2-highmem-2` as core node pool for basehubs
- `n2-highmem-4` as core node pool for daskhubs
- `n2-highmem-4`, `-16`, and `-64` for user node pool for basehub and daskhub

EKS:
- `r5` is used systematically over `m5`. They provide a bit better performance per CPU and a 1:8 ratio of CPU:Memory instead of 1:4.
- `r5.xlarge` as core node pool for basehubs and daskhubs (equivalent to `n2-highmem-4`), where the smaller option `r5.large` allows for too few pods per node
- `r5.xlarge`, `r5.4xlarge`, and `r5.16xlarge` for user node pool for basehub and daskhub (equivalent to `n2-highmem-4`, `-16`, and `-64`)

### Related

- #2121
- #2005

### DIFF
--- a/docs/topic/infrastructure/cluster-design.md
+++ b/docs/topic/infrastructure/cluster-design.md
@@ -1,43 +1,93 @@
 # Cluster design considerations
 
-## GKE
+## Core nodes' instance type
 
-## Core node size
+Each cluster is setup with a *core node pool*. Its expected to be static in size
+and always running. The core node's instance types should have: allocatable CPU,
+allocatable Memory, and allowed Pods per node, to efficiently schedule and
+reliably run:
 
-In each cluster, we have a *core node pool* that is fairly static in size
-and always running. It needs enough capacity to run:
+1. Kubernetes system pods - network policy enforcement, cluster
+   autoscaler, kube-dns, etc.
 
-1. Kubernetes system components - network policy enforcement, config connector
-   components, cluster autoscaler, kube-dns, etc.
+2. Per-cluster support pods - like prometheus, grafana, cert-manager, etc.
 
-2. Per-cluster support components - like prometheus, grafana, cert-manager,
-   etc.
+3. JupyterHub core pods - like hub, proxy, user-scheduler, etc
 
-3. Hub core components - the hub, proxy, userscheduler, etc
-
-4. (Optional) Dask gatway core components - the API gateway, controller, etc.
+4. (Optional) DaskGatway core pods - like gateway, controller, proxy.
 
 Since the core nodes are *always running*, they form a big chunk of the
-cluster's *base cost* - the amount of money it costs each day, regardless
-of current number of running users. Picking an apporpriate node size and
-count here has a big effect.
+cluster's *base cost* - the amount of money it costs each day, regardless of
+current number of running users. Picking an appropriate instance type for the
+core nodes has a big effect.
 
-### On GKE
+### CPU and Memory
 
-GKE makes sizing this nodepool difficult, since `kube-system` components can take up quite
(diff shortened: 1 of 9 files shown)


docs/topic/infrastructure/cluster-design.md
```diff
@@ -1,43 +1,93 @@
...
+        [250, 4],
+        [500, 5],
+        [1000, 6],
+        [1500, 7],
+        [2000, 8]
+      ]
+    }
+```
+
+On GKE clusters with network policy enforcement, we look to edit the
```
pnasrat, line 76: Can you clarify this - is this something that just happens through terraform, deployer, does it need manual intervention (if so document steps),

pnasrat (changes requested):
Minor doc question - however my main concern is around deploying this with minimal disruption

What's the deployment/roll out impact of this - will it cause all existing core nodes to need to be upgraded (ie does terraform plan run clean for existing clusters or want to upgrade existing pools?  How do user nodes get upgraded to the right machine type with this - will it cause rescheduling?

If so is there a plan to incrementally roll out?

consideRatio:
Thanks for reviewing @pnasrat!!

> What's the deployment/roll out impact of this - will it cause all existing core nodes to need to be upgraded (ie does terraform plan run clean for existing clusters or want to upgrade existing pools?

Super relevant question, I should have made it clear in the PR that this is not meant to influence `terraform plan` for any clusters existing node pools. If it does, its a non-intentend mistake to catch. In practice this just updates template files for new clusters, and the default value of `core_node_machine_type` - so, is this default used?

Apparently no, the defautl value for the `core_node_machine_type` variable isn't used. Its good I think, but it would been even better if it wasn't provided at all and instead was something required to be explicitly set by individual projects. I'll look into updating that for unused default values to reduce the complexity.

docs/topic/infrastructure/cluster-design.md
consideRatio: I opened https://github.com/2i2c-org/infrastructure/issues/2490 to offload further clarifications into another PR, right now we don't have a procedure or agreement on what to do.

An explicit action point in the linked issue is to update this text once its done!

consideRatio:
It seems by removing the default value specification, it becomes required! I'll do so.

> If present, the variable is considered to be optional and the default value will be used if no value is set when calling the module or running Terraform.

pnasrat (approved):
lgtm

consideRatio:
Thank you @pnasrat!!

damianavila:
> Super relevant question, I should have made it clear in the PR that this is not meant to influence terraform plan for any clusters existing node pools. If it does, its a non-intentend mistake to catch.

Is there any next intention/step to move the existing clusters to this new default, @consideRatio? If the answer is yes (I presume that is the case given #2121 context), is that captured in a new issue?

consideRatio:
@damianavila I don't have a clear next step that I think we should actively work by itself, but I think its relevant to opportunistically consider transitioning to highmem nodes when perform disruptive maintenance.

I opened https://github.com/2i2c-org/infrastructure/issues/2511 about this!

Issue #2316, opened 2023-03-07
[EVENT] UToronto final exams


damianavila:
### Summary

Ref: https://2i2c.freshdesk.com/a/tickets/484

### Event Info

- **Community Representative:** Nathan Taback

```
The GGR and EEB numbers will likely go down bit more over coming weeks.
GGR274 (~96)
EEB125 (~88)
RSM312 (~40)

Faculty	Department	Course	Section	Sesson	Course Name	Examdate	        Begin	Ends	Duration
ARTSC	EEB	EEB125H1S	        ALL	        20231	Comp & Data Sci - Sciences	        4/19/2023	9:00	12:00	180
ARTSC	GGR	GGR274H1S	ALL	        20231	Comp & Data Sci - Soc Sciences	4/25/2023	9:00	12:00	180
ARTSC	COMPG	RSM312H1S	ALL	        20231	Special Topics Mgmt	                 4/13/2023	9:00	12:00	180
```

- [**Hub Events Calendar**](https://calendar.google.com/calendar/u/2?cid=Y19rdDg0c2g3YW5tMHNsb2NqczJzdTNqdnNvY0Bncm91cC5jYWxlbmRhci5nb29nbGUuY29t)

### Hub info

- **Hub URL**: jupyter.utoronto.ca
- **Hub decommisioned after event?**: NO

### Task List

**Before the event**

- [ ] Dates confirmed with the community representative and added to Hub Events Calendar.
- [ ] Quotas from the cloud provider are high-enough to handle expected usage.
- [ ] **One week before event** Hub is running.
- [ ] Confirm with Community Representative that their workflows function as expected.
  - <details>
    <summary>👉Template message to send to community representative</summary>
    
    ```
    Hey {{ COMMUNITY REPRESENTATIVE }}, the date of your event is getting close!
    
    Could you please confirm that your hub environment is ready-to-go, and matches your hub's infrastructure setup, by ensuring the following things:
    - [ ] Confirm that the "Event Info" above is correct
    - [ ] On your hub: log-in and authentication works as-expected
    - [ ] `nbgitpuller` links you intend to use resolve properly
    - [ ] Your notebooks and content run as-expected
    ```
    
    </details>  
- [ ] **1 day before event**, either a separate nodegroup is provisioned for the event or the cluster is scaled up.

**During and after event**

- [ ] Confirm event is finished.
- [ ] Nodegroup created for the hub is decommissioned / cluster is scaled down.
[...]

PR #2489, opened 2023-04-13
utoronto: Provide exact same resources for all exam users

yuvipanda:
For equity reasons, everyone shall have access to the exact same computing resources.

I checked grafana, nobody has used more than 2G of RAM or 1 CPU for a while.

Ref https://github.com/2i2c-org/infrastructure/issues/2316


yuvipanda:
The 1 CPU guarantee will mean only about 7 users will fit on each node. I am going to use the azure CLI to scale up number of nodes before the event so users will also *start* quickly and *consistently*

yuvipanda:
I do think 1 CPU might be a bit excessive, and next time we should work with utoronto to see if 0.5 might be enough? But at least for yesterday's unplanned exam, there were users who used a full CPU.

consideRatio (approved):

PR #2484, opened 2023-04-12
Bump utoronto memory for exam

pnasrat:
See #2316

PR #2491, opened 2023-04-13
Reset UToronto Resource Requests

yuvipanda:
- Revert https://github.com/2i2c-org/infrastructure/pull/2489 because the exam is over
- Revert https://github.com/2i2c-org/infrastructure/pull/2484 because we got a misreport

Ref #2316


pnasrat (approved):
lgtm

PR #2492, opened 2023-04-13
Bump memory for temple final project period

pnasrat:
As per https://2i2c.freshdesk.com/a/tickets/643

This change puts back the temporary limit bump when Temple were experiencing   user kernel kills at request for final projects running until May 2

To be updated if instructor/community lead gets back with explicit RAM needs.


consideRatio (approved):

damianavila:
> This change puts back the temporary limit bump when Temple were experiencing user kernel kills at request for final projects running until May 2

Is this captured in a new issue? Otherwise, we are most likely going to miss it.

pnasrat:
I added a slack reminder for myself to do  this already

Issue #2461, opened 2023-04-04
[Request deployment] New Ephemeral Hub: ESA HackAnExoplanet


colliand:
### Important dates

- **Target start date**: as soon as possible and certainly before April 11, 2023
- **Required start date**: April 11, 2023
- **Any important dates for usage**: Usage peak is forecast for April 18, 2023

### Hub Authentication Type

Other (may not be possible, please specify in comments)

### First Hub Administrators

-  Maximilian N. Günther @MNGuenther

### [GitHub Auth only] How would you like to manage your users?

None

### Hub logo image URL

pending...

### Hub logo website URL

https://hackanexoplanet.esa.int/

### Hub user image GitHub repository

pending....

### Hub user image tag and name

pending...

### Extra features you'd like to enable

- [ ] Specific cloud provider or datacenter (otherwise GCP)
- [ ] Dedicated Kubernetes cluster
- [ ] Scalable Dask Cluster

### (Optional) Preferred cloud provider

None

### (Optional) Billing and Cloud account

None

### Other relevant information to the features above

This is a request for an _ephemeral hub_ following up on discussions with @yuvipanda in 2i2c's Slack.

### Tasks to deploy the hub

- [ ] 1. Deploy information filled in above
- [ ] 2. Engineer who will deploy the hub is assigned
- [ ] 3. If using GitHub Orgs/Teams Auth, Engineer is given Owner rights to the org to set this up.
- [ ] 4. Initial Hub deployment PR <link to PR>
- [ ] 5. Administrators able to log on -> Hub now in steady-state

colliand:
Hello @MNGuenther! With this issue, we are gathering the required information from you so that 2i2c's engineering team can launch the ephemeral hub for HackAnExoplanet. You will likely exchange messages with @yuvipanda. Yuvi and I are excited to support you and your colleagues at ESA and help you to inspire students with this project.

We request some information from you.

Please share a **URL pointing at the "splash image"** that users will see when they land on the login page. For example, notice the image at the top of the Openscapes login page here: https://openscapes.2i2c.cloud/

Do you have a Docker image that specifies the software environment you'd like for the hub? [Documentation is available here](https://docs.2i2c.org/en/latest/admin/howto/environment/hub-user-image-template-guide.html). 2i2c can provide a standard image or you can work with us to customize the environment for your usage scenario.

yuvipanda:
Ah unfortunately no splash screen is possible here, you'll just get directly sent to your notebook.

[...]

PR #2496, opened 2023-04-17
Specify image used for exoplanet hackathon hub

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/2461


yuvipanda:
Self-merging, as it is an image update

Issue #2497, opened 2023-04-17
Handle health checks for hubs with HTTP Basic Auth


yuvipanda:
We now support HTTP Basic Auth for protecting a hub completely with a shared username/pw - *purely* for use with ephemeral hubs with Tmpauthenticator.

However, neither our health checks nor our uptime checks can handle this correctly. They should.

- [ ] Figure out how [health checks](https://github.com/2i2c-org/infrastructure/tree/master/deployer/tests) can talk to tmpauthenticator based hubs with admin user accounts - without this, we can not make appropriate API calls
- [ ] Figure out how our GCP Uptime Checks can support being ok with getting a 401 HTTP response, rather than a 2xx.

agoose77:
I think this is likely not needed now? @yuvipanda

PR #2498, opened 2023-04-18
Allow customizing the expected status code for uptime checks

yuvipanda:
Required when we support ephemeral password protected hubs, as the password protection is accomplished via HTTP Basic Auth. All endpoints are thus protected, including `/hub/health`.

Ref https://github.com/2i2c-org/infrastructure/issues/2497


pnasrat (approved):
LGTM

PR #2499, opened 2023-04-19
Remove separate nodepool for OHW

yuvipanda:
It barely has any users, and due to lack of taints, is being used by other hubs as well when needed. This lets us consolidate into the existing pool.

We can bring this back when they are having an event or similar.


consideRatio (approved):

yuvipanda:
Applied

Issue #2456, opened 2023-04-03
[billing] Single source of truth for contracts and method to link hubs to contracts


pnasrat:
Part of billing story. Placeholder - define single source of truth for contracts and integration with hub metadata

## Context

2i2c need a stable and central source of truth for contracts that can be queried for current and historical state.

Questions - what is the life of a contract?

## TODO 

- [ ] Write up requirements for single source of truth
- [ ] Document current options (including airtable, grafana, google sheets)
- [ ] Define schema for single source of truth
- [ ] Populate existing data for current (and historical? contracts)
- [ ] Design and document manual process for updating source of truth
- [ ] integrate manual process for invoicing with this source of truth

choldgraf:
Just a note that I think we'll also need to define a source of truth for **invoices**, so that we can link invoices to contracts, as well as to outstanding cloud costs. That way we can answer "which invoice is the one that paid for cloud charge X for contract Y"

choldgraf:
I've got a PR to document our Contracts and Invoices airtable tables here:

- https://github.com/2i2c-org/team-compass/pull/698

If we somehow got our "currently active hubs" data into AirTable form[^1], we could link hubs to contracts (and thus to invoices) via [Linked Records](https://www.airtable.com/guides/build/connect-data-with-linked-records).

[^1]: How would we do that? We could either programmatically update this somehow, or we could potentially manually maintain a record of all running hubs in AirTable that we cross-reference with our deployment information. For example, we might have a daily job that checks our `infrastructure/` repository for all currently deployed hubs, checks a `Hubs` AirTable to make sure that all hubs in our deployment are represented in the AirTable, and if not then adds new records to reflect any new hubs that have been added. We could then add extra metadata to the AirTable and link them to other contracts / invoice records in AirTable.

choldgraf:
Also just for team visibility - I'm also iterating a bit with Joe at CS&S in a notion page he set up for us to provide feedback. I've requested that @colliand and @jmunroe have access to this page as well so they can see the conversation there too.

colliand:
I have access to the Notion page. Thank you.

choldgraf:
### Quick idea: keep an auto-updated "hubs" airtable?

The goal of this comment is to satisfy the "method to link hubs to contracts" part of this issue[^1].

[^1]: apologies for the weekend update, but the thought just popped into my head and I wanted to write it down while it was fresh.

[...]

PR #2509, opened 2023-04-24
DOCS: Package hub list as a JSON object in our documentation

choldgraf:
This PR modifies our "generate a list of our active hubs" script to output a **JSON file that is served with our docs** in addition to the CSV file that is used to create the HTML tables.

This is part of an experiment at automating an AirTable script that would automatically read in the JSON file and update a table of records that list all of our currently-running hubs. I don't think publishing the JSON has negative implications, given that all the same data is already present at https://infrastructure.2i2c.org/reference/hubs.html

I don't know if this JSON approach will work or not, but I wanted to quickly expose this data to see if I can get something working with minimal effort to decide if it's worth exploring further.

Here's an example of what the JSON would look like: https://2i2c-pilot-hubs--2509.org.readthedocs.build/_static/hub-table.json

- related to: #2456

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -19,8 +19,8 @@ extensions = [
 
 intersphinx_mapping = {
     "z2jh": ("https://zero-to-jupyterhub.readthedocs.io/en/latest/", None),
-    "tc": ("https://team-compass.2i2c.org/en/latest/", None),
-    "dc": ("https://docs.2i2c.org/en/latest/", None),
+    "tc": ("https://team-compass.2i2c.org", None),
+    "dc": ("https://docs.2i2c.org", None),
 }
 
 # -- MyST configuration ---------------------------------------------------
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -19,8 +19,8 @@
 
 intersphinx_mapping = {
     "z2jh": ("https://zero-to-jupyterhub.readthedocs.io/en/latest/", None),
-    "tc": ("https://team-compass.2i2c.org/en/latest/", None),
-    "dc": ("https://docs.2i2c.org/en/latest/", None),
+    "tc": ("https://team-compass.2i2c.org", None),
```
choldgraf, line 22: this fixes our intersphinx links while I was at it

pnasrat (approved):
LGTM to experiment with and get you unblocked

PR #2500, opened 2023-04-19
utoronto Provide consistent resources for exams

pnasrat:
See: #2316

PR #2513, opened 2023-04-26
Revert "utoronto Provide consistent resources for exams"

yuvipanda:
Reverts 2i2c-org/infrastructure#2500

Exams are over


yuvipanda:
ok to self-merge

Issue #2447, opened 2023-03-31
Set CPU limits for all educational hubs


yuvipanda:
Follow-up from https://github.com/2i2c-org/infrastructure/issues/2445.

A node was pegged back to 100% CPU by a single user using all the CPU. We want to be generous in allowing users to spike up their CPU usage, but not take down an entire node. Setting CPU limit to be like 1/2 of a single node's capacity works decently, as it requires *two* people on the *same* node to do things at the same time. 

So for all educational hubs, we should set a CPU limit for users.

consideRatio:
I'd like to investigate why "take down an entire node" is possible like this at all, and if it may be cloud provider specific. This was seen on utoronto's AKS. I think it may be related to cloud provider specific overhead of resources for system pods on the node and resource allocation to kubelet for example.

I look to look into these questions as part of getting node sharing setup robustly.

yuvipanda:
@consideRatio It mostly just starved everyone else of CPU, so jupyter-server / jupyter-notebook process couldn't get enough CPU to run I think. If you look at the logs in https://github.com/2i2c-org/infrastructure/issues/2445#issuecomment-1490820734, that error (and pretty much everything else there) is because while the pod started, the server process couldn't actually reach out to the hub in time.

yuvipanda:
@consideRatio note that I've specifically scoped this to just educational hubs, where we do not want to show a profileList at all.

PR #2514, opened 2023-04-27
Set CPU limit for cloudbank hubs

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/2447

### DIFF
--- a/config/clusters/cloudbank/common.values.yaml
+++ b/config/clusters/cloudbank/common.values.yaml
@@ -4,9 +4,9 @@ jupyterhub:
       # Each node has about 8 CPUs total, and if we limit users to no more than
       # 4, no single user can take down a full node by themselves. We have to
       # set the guarantee to *something*, otherwise it is set to be equal
-      # to the limit!
+      # to the limit! We don't explicitly set a guarantee, because there is
+      # a guarantee of 0.05 set in basehub/values.yaml
       limit: 4
-      guarantee: 0.01
     image:
       name: quay.io/2i2c/2i2c-hubs-image
       tag: "14107b8a85fb"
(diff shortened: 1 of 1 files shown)


config/clusters/cloudbank/common.values.yaml
```diff
@@ -1,5 +1,12 @@
 jupyterhub:
   singleuser:
+    cpu:
+      # Each node has about 8 CPUs total, and if we limit users to no more than
+      # 4, no single user can take down a full node by themselves. We have to
+      # set the guarantee to *something*, otherwise it is set to be equal
+      # to the limit!
+      limit: 4
+      guarantee: 0.01
```
consideRatio, line 9: I'm not fully confident, but I think we ought to make the guarantee high enough to startup, and I think ~5% of a CPU may be required. I've seen user pods starting up use about that much CPU.

```suggestion
      guarantee: 0.05
```

consideRatio: 5% of a CPU in gaurantee would allow for up to about 80 pods on a 4 CPU node.

If memory requests are 0.5 GB per user, they would still run short by memory before CPU with 5% on a 4 CPU / 32 GB ram node.

On a 8 CPU or larger node, they would run out of the total number of pods allowed for a node before the 5% of a CPU guaranteed became a problem.

consideRatio: Oh, it is default to 0.05 in basehub: https://github.com/2i2c-org/infrastructure/blob/e1d2bd3ad79a9fe276d13a93c0e174710c1b5fb6/helm-charts/basehub/values.yaml#L291-L309

yuvipanda: @consideRatio I've removed the guarantee and added a comment on why.

consideRatio (approved):

PR #2518, opened 2023-05-02
Allow executing a single command line with use-cluster-credentials

yuvipanda:
Very helpful when trying to execute a single command targeted at a particular cluster and use its result. For example, to run "kubectl version" on all clusters, you can run:

```
ls config/clusters | grep -v template | xargs -L1 -I{} deployer use-cluster-credentials {} "kubectl version"
```


pnasrat (approved):
lgtm

GeorgianaElena (approved):

PR #2534, opened 2023-05-09
Update CarbonPlan images

maxrjones:


config/clusters/carbonplan/common.values.yaml
```diff
@@ -183,7 +183,6 @@ basehub:
       config:
         Authenticator:
           allowed_users: &users
-            - jhamman
```
consideRatio, line 186: @maxrjones I'm not right away confident that having an empty list defined here will be okay from a security perspective - that any user may gain access.

I don't have capacity to review that part more in depth right now. Can we add jhamman back for now to avoid needing to look into the details until next week and get this merged now?

config/clusters/carbonplan/common.values.yaml
```diff
@@ -183,7 +183,6 @@ basehub:
       config:
         Authenticator:
           allowed_users: &users
```
yuvipanda, line 185:
```suggestion
          allowed_users: &users
            - maxrjones
```

yuvipanda: Or replace it with your own username

config/clusters/carbonplan/common.values.yaml
maxrjones: I've committed Yuvi's suggestion to put in my username. We definitely don't want any user to have access :) I opened a separate ticket to discuss how this works yesterday - so we could continue the discussion there when you have more time. thanks!

yuvipanda:
@maxrjones test this out and let us know if this works?

maxrjones:
> @maxrjones test this out and let us know if this works?

Works great! I'll open a separate PR to make Pangeo notebook the default while we work on the custom images.

Thanks for both of your reviews! Hope the conference is going well!

PR #2539, opened 2023-05-16
awi-ciroh: update min nodes during event 16-18 May

consideRatio:


consideRatio:
@pnasrat good morning!

Yes I agree, I outlined a plan in https://2i2c.slack.com/archives/C055A1J1DRP/p1684233173413399 just now - making the point about not making a UI change as well!

config/clusters/awi-ciroh/common.values.yaml
```diff
@@ -61,35 +61,53 @@ basehub:
         # on these nodes. They need to be just under total allocatable
         # RAM on a node, not total node capacity. Values calculated using
         # https://learnk8s.io/kubernetes-instance-calculator
+        #
```
pnasrat, line 64: Add a link to the event issue here for reference

terraform/gcp/projects/awi-ciroh.tfvars
```diff
@@ -68,6 +68,23 @@ notebook_nodes = {
       count: 0
     }
   },
+  # added stressfully before an event where we ran out of ssd quota, see
+  # https://github.com/2i2c-org/infrastructure/pull/2539.
```
pnasrat, line 72: Ditto link to event issue not PR.

pnasrat (changes requested):
minor doc tweaks but consider this lgtm'd pending those fixes

terraform/gcp/projects/awi-ciroh.tfvars
consideRatio: I want to link to the PR where I posted screenshots about the failures related to ssd quota though. I added a commit to link to the event as well directly.

config/clusters/awi-ciroh/common.values.yaml
```diff
@@ -51,16 +51,13 @@ basehub:
...
           scope:
             - read:org
-    # FIXME: The configurator is used, so we should not be using prePuller at
-    #        the same time. This is just used to help pre-warm nodes during an
-    #        event.
-    #
-    prePuller:
-      continuous:
-        enabled: true
     singleuser:
```
pnasrat, line 54: Maybe we should minimize the changes in this PR to the node updates -  is there any harm in just match the configurator image `quay.io/2i2c/awi-ciroh-image:584293e50d4c` and continuing to prepull for the event to  decrease startup time.

pnasrat (changes requested):

config/clusters/awi-ciroh/common.values.yaml
pnasrat: Fine ignore this comment. However this sounds like our system the owner of the image (configurator/kubespawner) should control the prepulling not the helm chart longer term.

pnasrat (approved):
lgtm

consideRatio:
Thanks for working on this with me @pnasrat!!!

Issue #2542, opened 2023-05-16
Change machine types for QCL


jmunroe:
### Context

Original context: https://2i2c.freshdesk.com/a/tickets/706

> At QCL, we are getting a little too much memory pressure on our CPU count instances. We would like to change the two highcpu (n2-highcpu-32 and n2-highcpu-96) instances for:
n2-standard-48
n2-standard-96

### Proposal

The current profiles for "n2-highcpu-32: 32 CPU / 32 GB RAM" and "n2-highcpu-96: 96 CPU / 96 GB RAM" are not for shared use so I assume the 'highmem' variant is not required. 

Please update the machine types for these to n2-standard-48 and n2-standard-96 for the QCL hub.

PR #2548, opened 2023-05-17
Qcl machine type change

pnasrat:
See #2542


yuvipanda (approved):
Should we track removing these deprecated values somewhere?

pnasrat:
I'm going to follow up once I inform the community they've been updated and get a timeline for them. I just didn't want to delete existing in use nodepools or cause unschedulable running user servesr

yuvipanda:
@pnasrat sounds great! ty!

PR #2553, opened 2023-05-19
Add image option for CarbonPlan Hub

maxrjones:
Using the latest tag because we anticipate needing to make changes to the image

### DIFF
--- a/config/clusters/carbonplan/common.values.yaml
+++ b/config/clusters/carbonplan/common.values.yaml
@@ -59,9 +59,9 @@ basehub:
                   slug: "main"
                   kubespawner_override:
                     image: carbonplan/main-single-user:latest
-                main:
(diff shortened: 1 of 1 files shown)


config/clusters/carbonplan/common.values.yaml
```diff
@@ -59,6 +59,11 @@ basehub:
                   slug: "main"
                   kubespawner_override:
                     image: carbonplan/main-single-user:latest
+                main:
+                  display_name: Forest Offset Fires
+                  slug: "forest-offset-fires"
```
consideRatio, line 64:
```suggestion
                forest-offset-fires:
                  display_name: Forest Offset Fires
                  slug: forest-offset-fires
```

yuvipanda (approved):

Issue #2157, opened 2023-02-07
Document procedure and migrate existing GCP GKE based hubs from k8s 1.22+ to 1.24+


consideRatio:
This is the current status of GCP GKE clusters. We should get them all to 1.24+. This is where the "regular" release channel of GKE is currently at.

- [x] 2i2c, k8s v1.26.4-gke.1400 - [GCP link](https://console.cloud.google.com/kubernetes/clusters/details/us-central1-b/pilot-hubs-cluster/details?project=two-eye-two-see)
  - Node groups at: v1.26
  - Upgrade done by Yuvi 2023-06-09, from v1.22 to v1.26
- [x] pangeo-hubs, k8s v1.26.4-gke.1400 - [GCP link - passport submission for access](https://console.cloud.google.com/kubernetes/clusters/details/us-central1-b/pangeo-hubs-cluster/nodes?project=columbia)
  - Node groups at: v1.26
  - Upgrade done by Yuvi 2023-06-09, from v1.22 to v1.26
- [x] meom-ige, k8s  v1.26.4-gke.1400 - [GCP link](https://console.cloud.google.com/kubernetes/clusters/details/us-central1-b/meom-ige-cluster/nodes?project=meom-ige-cnrs)
  - Node groups at: v1.26
  - Upgrade done by Yuvi 2023-06-09, from v1.22 to v1.26
- [x] cloudbank, k8s v1.26.4-gke.1400 - [GCP link](https://console.cloud.google.com/kubernetes/clusters/details/us-central1-b/cb-cluster/nodes?project=cb-1003-1696)
  - Node groups at: v1.26
  - Upgrade done by Yuvi 2023-06-09, from v1.22 to v1.26
- [x] callysto, k8s v1.25.6-gke.1000 - [GCP link](https://console.cloud.google.com/kubernetes/clusters/details/northamerica-northeast1/callysto-cluster/details?project=callysto-202316)
  - Node groups at: v1.25
  - Maintenance upgrade suggested at https://2i2c.freshdesk.com/a/tickets/562, tracked in https://github.com/2i2c-org/infrastructure/issues/2388, k8s upgraded from 1.23 -> 1.25, node pools from 1.22 -> 1.25
- [x] m2lines, k8s v1.24.5-gke.600 - [GCP link](https://console.cloud.google.com/kubernetes/clusters/details/us-central1/m2lines-cluster/details?project=m2lines-hub)
  - Node groups at: v1.24
[...]

PR #2554, opened 2023-05-19
awi-ciroh: upgrade k8s to 1.25, and adjust node pools following event

consideRatio:
### Prepared for node sharing

The awi-ciroh cluster had a workshop where we ran into limitations from using 1:1 user:node. Due to that we did a behind the scenes quick migration to a node sharing setup.

This node sharing setup is now complemented, allowing us to reconfigure the awi-ciroh hubs without further terraform changes based on their wishes.

### K8s cluster upgrade

I made an opportunistic upgrade of the k8s cluster while adjusting node pools. I also gained some experience which helps create GKE k8s cluster upgrade docs. Those notes are captured and linked from #2157.

### Future work

They now have a UI with small / medium / large / huge, but they will just start on the n2-highmem-16 machine no matter what with a pod requesting memory based on the legacy 1:1 user:node profile list options.

I'll reach out in order to align their setup to a typical node sharing setup, which from this point onwards is just a hub config change.

### DIFF
--- a/config/clusters/awi-ciroh/common.values.yaml
+++ b/config/clusters/awi-ciroh/common.values.yaml
@@ -78,6 +78,9 @@ basehub:
         #        the previous 1:1 user:node setup with a node sharing setup. It
         #        is not meant to be retained long term!
         #
+        #        -[ ] Make this cluster have a node sharing setup like in the
+        #             basehub/daskhub template.
+        #
         - display_name: "Small"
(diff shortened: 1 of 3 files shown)


yuvipanda (approved):

consideRatio:
Thanks @yuvipanda! I didn't finish the PR description in time for your review, but there is now one :)

Issue #2523, opened 2023-05-03
[Request deployment] New Hub: NCAR-CISL for UCAR


colliand:
### The GitHub handle of the community representative

nwehrheim

### Hub important dates

- Target start date: June 1, 2023
- Target end date: perhaps Oct 1, 2023 or May 30, 2023...working out details in procurement process. 

### Hub Authentication Type

GitHub (e.g., @mygithubhandle)

### First Hub Administrators

- @kcote-ncar
- @NicholasCote
- @nwehrheim

### [GitHub Auth only] How would you like to manage your users?

Allowing members of specific GitHub team(s)

### [GitHub Teams Auth only] Profile restriction based on team membership

NCAR/2i2c-cloud-users (learned this in the Running Notes doc comment exchange with @nicholascote).

### Hub logo image URL

https://www.vmcdn.ca/f/files/longmontleader/import/2017_06_ncar_highres_transparent.png

### Hub logo website URL

https://www2.cisl.ucar.edu/

### Hub user image GitHub repository

pending

### Hub user image tag and name

pending

### Extra features you would like to enable

- [ ] Dedicated Kubernetes cluster
- [X] Scalable Dask Cluster

### (Optional) Preferred cloud provider

AWS

### (Optional) Billing and Cloud account

None

### Other relevant information to the features above

- Cluster:
    - Dedicated or shared? Shared; operated by 2i2c, cloud costs passed through
    - Billing account? 2i2c operates shared cluster; cloud costs will be passed through to NCAR-CISL/UCAR
- Cloud compute details
    - Hardware profile(s) and Machine types
         - Small: t3.large
         - Medium: t3.xlarge
         - Large: T3.2xlarge
         - Large w/GPU: T2.2xlarge + GPU
- GPU? Yes
- specification:
    - Cloud vendor: AWS
          - Data center: US West (Oregon) us-west-2
          - Zone in data center: us-west-2a, us-west-2b

Proposed hub URL: ncar.2i2c.cloud with associated staging.ncar.2i2c.cloud hub. 

### Tasks to deploy the hub

- [x] 1. Deploy information filled in above
- [x] 2. Engineer who will deploy the hub is assigned
- [x] 3. If using GitHub Orgs/Teams Auth, Engineer is given Owner rights to the org to set this up.
- [x] 4. Initial Hub deployment PR <link to PR>
- [x] 5. Administrators able to log on -> Hub now in steady-state

colliand:
[...]

Issue #2564, opened 2023-05-19
Allow hub admins to modify profileLists


yuvipanda:
Currently, the hub configurator can not be used to modify *profileLists*. This means changes to image tags *must* take up 2i2c engineer time. This is a form of toil we should avoid.

We should work on augmenting the configurator so it supports profileLists properly, so we admins can self-administer that.

yuvipanda:
Note that while I prototyped this up, that doesn't necessarily mean I lead the actual implementation of this.

jmunroe:
I was *just* trying to compare some different images for use on researchdelight. What I wanted was to be able to different images at the same time. If this goal gets implemented it is far better than trying make PRs on our infrastructure hub to update the profile.

consideRatio:
> We develop this on its own repo (not as part of 2i2c-org/infrastructure), with the aim of being tied only to z2jh and no 2i2c dependencies. 

:100:

I suggest the aim is made stricter to reduce coupling to 2i2c further as an open source project. I'm thinking of for example committing to not writing docs coupling to 2i2c's basehub helm chart, and overall writing content to the git repo it as if it would have lived independently from 2i2c, in for example the jupyterhub github organization.

PR #2584, opened 2023-05-29
2i2c-aws-us: ncar-cisl added, researchdelight GPU image option removed

consideRatio:
- for #2523 

Includes some non-hub specific commits in docs (a129484, 1c1a2f2), and also impacts researchdelights GPU options by not providing a image choice for the GPU node (7bd16d2). I was thinking that either we have the profileList in common.values.yaml, or we make it hub specific. If we have it in common.values.yaml I was thinking we can't have it include an image specific choice as that sets us up for trouble of managing images for the communities, and making them unable to do it themselves via the configurator.


config/clusters/2i2c-aws-us/ncar-cisl.values.yaml
```diff
@@ -0,0 +1,68 @@
...
+        - noresvport
+      serverIP: fs-0b70db2b65209a77d.efs.us-west-2.amazonaws.com
+      baseShareName: /
+  jupyterhub:
+    ingress:
+      hosts: [ncar-cisl.2i2c.cloud]
+      tls:
+        - hosts: [ncar-cisl.2i2c.cloud]
+          secretName: https-auto-tls
+    proxy:
```
yuvipanda, line 23: This is actually not required, as we do not use `proxy` anymore for HTTPS. The `proxy` stanza can be removed

consideRatio: Hmmmm, I recall it was required for some reason still, but I recall thinking the same thing.

If it isn't, it should be addressed for all our hubs, because its consistently setup for all hubs after https://github.com/2i2c-org/infrastructure/pull/2349.

yuvipanda (approved):
One extreneous config stanza but otherwise lgtm

config/clusters/2i2c-aws-us/ncar-cisl.values.yaml
consideRatio: Ah, I figure it may be because of this, which is something we probably can relax: https://github.com/2i2c-org/infrastructure/blob/3a6b47f68443345faff31ca805459cd52c1fdd78/helm-charts/basehub/values.schema.yaml#L258-L268

yuvipanda: Aaah yes we should! Not sure why that's listed as required. Am ok if you wanna do that as a separate PR, or open an issue and deal with it later as well. Not required to block this PR

consideRatio:
@yuvipanda I updated the PR title and PR description to reflect the PR better as it was a bit more than just add the hub for ncar-cisl.

Do you think the decision to remove the image choice for researchdelight in the GPU section was acceptable as well? I didn't think it was important enough to put the profileList config in individual hubs, but at the same time didn't want image choices to be shared among hubs in a shared cluster for the gpu option.

yuvipanda:
@consideRatio I do think whenever we offer GPUs, we should offer those two choices. The default image doesn't have any GPU stuff, and these two images are different enough that I think we should do this each time.

yuvipanda:
It probably makes sense to put these in the hub specific config though.

consideRatio:
> I do think whenever we offer GPUs, we should offer those two choices.

Okay lets do it, but I feel uneasy about this because I see how we provide blurry lines about image management responsibility by doing this.

To quickly clarity why without entering a discussion about it here, consider the note I wrote in the config for singleuser.image below.

```yaml
      image:
        # image choice preliminary and is expected to be setup via
        # https://ncar-cisl.2i2c.cloud/services/configurator/ by the community
        #
        # pangeo/pangeo-notebook is maintained at: https://github.com/pangeo-data/pangeo-docker-images
        name: pangeo/pangeo-notebook
        tag: "2023.05.18"
```

The gist is that I don't think we have a clear and simple enough policy for this image management if we specify a kubespawner_override for the GPU profile's image. By specifying it, we discard the community's ability to self-manage the GPU profile's image via the default image in the configurator, and perhaps also make the community adopt expectations that it needs to be a dedicated image.

yuvipanda:
@consideRatio yes i agree with your concerns re: management, hence opened https://github.com/2i2c-org/infrastructure/issues/2564. However, if we don't provide them the option to use the GPU image, they can not use any of the GPU features. I appreciate you sticking to the status quo here so users can benefit despite your uneasiness.

consideRatio:
I read your excellent writeup in #2564 and I figure we'll skip a discussion about handling this interim situation where this is a problem to save our time and focus on the long term things instead.

PR #2588, opened 2023-05-30
2i2c-uk, lis: stop bumping image tags, configurator is used

consideRatio:
- Closes #2424

> I think this is currently managed by the configurator, so not sure if this should be autobumped

I've observed the same thing in https://ds.lis.2i2c.cloud/services/configurator/ and concluded that its a hub running in a shared cluster 2i2c-uk, so having prePulling wouldn't scale as whenever there is a scaleup a user from hub A may end up waiting for a node pulling an image from hub B C and D. This can happen because the k8s node's software kubelet is pulling image to start pods in sequence, so there is no guarantee that the pod that triggered scaleup gets its image pulled before the prePuller pod that are functioning as dummy user pods.


yuvipanda (approved):

GeorgianaElena:
Just a note that the auto bumping of the lis image was requested by the lis folks some time ago. https://github.com/2i2c-org/infrastructure/issues/1742 has more details. Not sure how much they are relying on this or for what, but maybe it's worth letting them know that we're deactivating this and explain the prePuller scaling situation that Erik described above.

consideRatio:
Thanks for linking back to that @GeorgianaElena! I've sent a message to Matthew https://2i2c.freshdesk.com/a/tickets/749 that the configurator changes the default image for all users, and we have disabled the automation to keep this bypassed configuration updated.

yuvipanda:
Confirmation received that this is ok! https://2i2c.freshdesk.com/a/tickets/749

PR #2592, opened 2023-05-31
[Attept 2] Update workflow to trigger on commit and remove redundant step

GeorgianaElena:
https://github.com/2i2c-org/infrastructure/pull/2590 didn't solve the issue. I now believe I've misunderstood it and the issue is that the workflow doesn't trigger for commits if not explicitly listed as a trigger and not the job is skipped (it might be both though)

Previously, I thought that the triggers listed in this workflow are ignored when enabling the required workflows and the commit one is auto-added.

Note that I plan to self-merge this to see if it works, before abandoning this required workflows idea.

### DIFF
--- a/.github/workflows/auto-author-assign-pull-request.yaml
+++ b/.github/workflows/auto-author-assign-pull-request.yaml
@@ -6,6 +6,12 @@ on:
     types:
       - opened
       - reopened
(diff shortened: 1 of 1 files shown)


GeorgianaElena:
Yay. This solved it https://github.com/2i2c-org/team-compass/pull/746

PR #2600, opened 2023-06-01
Decom earthlab hub

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/2598


consideRatio (approved):
I updated the title to reflect that this was a hub in a shared cluster - and what cluster, which I find a bit helpful overall.

yuvipanda:
Cleaned out the homedir too!

PR #2606, opened 2023-06-01
Add autoformatting to terraform code

yuvipanda:
Uses terraform_fmt with defualt options.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,7 +42,7 @@ repos:
       - id: flake8
 
   - repo: https://github.com/yuvipanda/terraform-bin
-    rev: 367f73c8086f9e559b7f52c0be73e46d1a5f1cfc
+    rev: c3c7973e77cdf9bbb44730e55a2de0475d683540
     hooks:
       - id: terraform-fmt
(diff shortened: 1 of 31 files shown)


pnasrat (approved):
LGTM

yuvipanda:
Unfortunately pre-commit.ci is failing because it does not have `terraform` installed so i'm shaving a yak somewhere in nepal...

yuvipanda:
@pnasrat ok, so yak has been shaved and there's now a new pre-commit, using https://github.com/yuvipanda/terraform-bin. Does this still look good to you? I could also move that pre-commit hook into the 2i2c org instead of my personal repo.

pnasrat:
LGTM

I think for now it's ok in a personal repo

At some point as an org we probably want to figure what should live under the org repo/best practices on that but that can wait for another day!

yuvipanda:
Thanks, @pnasrat!

sgibson91:
> pre-commit.ci is failing because it does not have `terraform` installed

This is exactly the problem I ran into when I tried to implement this ages ago. Thank you for figuring out a solution!

PR #2612, opened 2023-06-05
Fixing dependabot comment workflow

sgibson91:
### Context

Certain PRs opened by dependabot require extra actions to be taken when merging, such as updating our hub image. We have a GitHub Actions workflow that adds comments to these PRs explaining the next steps to take, but it was not working. This PR aims to fix that issue.

---

Problems encountered causing this workflow to fail and how they were resolved:

- The issue/pr number was not located under `github.context.issue.number`, but instead `github.context.number`
- `GITHUB_TOKEN` did not have enough permissions to work in PRs from forks, and the peter-evens/create-or-update-comment action does not support this. So had to update the trigger to `pull_request_target`
  - **Note:** _For workflows that are triggered by the pull_request_target event, the GITHUB_TOKEN is granted read/write repository permission **unless the permissions key is specified**_ (which it is, emphasis added) from https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target (warning box)

### DIFF
--- a/.github/workflows/comment-dependabot-prs.yaml
+++ b/.github/workflows/comment-dependabot-prs.yaml
@@ -1,7 +1,7 @@
 # From https://docs.github.com/en/actions/managing-issues-and-pull-requests/commenting-on-an-issue-when-a-label-is-added
 name: Comment dependabot PRs with specific label
 on:
-  pull_request:
+  pull_request_target:
     types:
       - labeled
 jobs:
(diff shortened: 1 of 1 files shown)


sgibson91:
Ah, this is annoying:

> Note: In public repositories this action does not work in pull_request workflows when triggered by forks. Any attempt will be met with the error, Resource not accessible by integration. This is due to token restrictions put in place by GitHub Actions.

from https://github.com/peter-evans/create-or-update-comment#action-inputs

sgibson91:
So switching to `pull_request_target` as a trigger _should_ work, but I don't think it is testable until the PR is merged. Right now, removing and readding the label no longer triggers the workflow in this PR.

consideRatio (approved):

Issue #2562, opened 2023-05-19
Setup a EBS volume per user for cryo cloud hub


yuvipanda:
Based on https://github.com/CryoInTheCloud/hub-image/issues/60, it looks like they are bottlenecking on reading 3G of data from NFS into QGis, taking about 5 minutes when tested with 1 user. Turns out they also have a workshop this coming week (we should write down and make more explicit our event notification policy), and that might have multiple users trying to do this at the same time. 

We can support this with a per-user PVC that users can copy the data into once and then load from there. We already have the 'per-user PVC' feature setup with https://github.com/2i2c-org/infrastructure/blob/master/docs/howto/features/per-user-db.md - we only need to enable the PVC, not the postgres database type situation.

yuvipanda:
We should cleanup these EBS volumes after the event.

yuvipanda:
Per https://github.com/CryoInTheCloud/hub-image/issues/60#issuecomment-1557449813, this event is till 'end of the week'. I think early next week we can undo this change and cleanup the EBS volumes.

@MattF-NSIDC  are you happy with us getting rid of the data in the EBS volumes by monday?

MattF-NSIDC:
That works for us! The data can be deleted any time after Friday 5/26.

yuvipanda:
@MattF-NSIDC I'm going to delete these tomorrow, and get rid of the per-user EBS volume. Does that sound ok to you?

How did the workshop go?

MattF-NSIDC:
That sounds great, @yuvipanda, thank you! Sorry for the late response, was out of office recovering after the workshop :)

The workshop went amazingly well, couldn't have been happier. Great insights from participants -- particularly excited about [thoughts on FAIR/CARE principles](https://github.com/orgs/qgreenland-workshop-2023-researcher/discussions/categories/fair-care) shared by participants.

No hitches on CryoCloud whatsoever. We created a script (`public/QGreenland/sync.sh` IIRC) that we asked users to run on their first day to set up their data in the correct locations, and this worked _very_ well.

@trey-stafford and I will be prioritizing getting that blog post started this week!

yuvipanda:
Absolutely glad to hear, @MattF-NSIDC!!! This is also the first workshop using qgis at scale like this, so am very happy to hear that went well too :)

Would love to do the blog post! Let me know as soon as you have anything at all and I'll take a look :)

I'll be tearing down the EBS volumes in a day or two.

damianavila:
[...]

PR #2613, opened 2023-06-05
cryocloud: Stop providing an EBS volume per user

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/2562


yuvipanda:
When done, I'll delete all the existing PVCs which will delete the underlying EBS volumes.

PR #2618, opened 2023-06-05
nasa-veda: Provide access to a few more buckets

ranchodeluxe:
Provide access to a few more buckets

Ref: https://2i2c.freshdesk.com/support/tickets/759

### DIFF
--- a/terraform/aws/projects/nasa-veda.tfvars
+++ b/terraform/aws/projects/nasa-veda.tfvars
@@ -36,7 +36,14 @@ hub_cloud_permissions = {
                     "s3:GetBucketLocation",
                     "s3:ListMultipartUploadParts"
                 ],
-                "Resource": "*"
+                "Resource": [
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thanks @ranchodeluxe - this looks ok to me. However, my concern usually is that sometimes there's a tf state bucket in the same AWS account that this might give access to if it is not secured specifically. Can I just get an explicit check to make sure this isn't the case? Happy to merge and deploy after that.

ranchodeluxe:
> Thanks @ranchodeluxe - this looks ok to me. However, my concern usually is that sometimes there's a tf state bucket in the same AWS account that this might give access to if it is not secured specifically. Can I just get an explicit check to make sure this isn't the case? Happy to merge and deploy after that.

Yep, not the case so we are good

yuvipanda:
Thanks @ranchodeluxe!

yuvipanda:
@ranchodeluxe applied! LMK if it works ok

PR #2621, opened 2023-06-06
Add an efficient, user homedirectory size prometheus reporter

yuvipanda:
This deploys https://github.com/yuvipanda/prometheus-dirsize-exporter, an *efficient* per-user homedirectory stats (size, no. of files, last modified date, etc) collector. It is capped at performing no more than 250 IO operations per second, to not overwhelm NFS servers. Metrics are refreshed every 2h after completion, although on large servers (like LEAP), they can take many many hours to complete with just 250 IO operations per second. This is perfectly fine though, as we do not need 'up to date' information. Trading off metric latency for minimal resource usage is pretty good here.

Ref https://github.com/2i2c-org/infrastructure/issues/764

### DIFF
--- a/helm-charts/basehub/templates/home-dirsize-reporter.yaml
+++ b/helm-charts/basehub/templates/home-dirsize-reporter.yaml
@@ -1,0 +1,80 @@
+{{- if or .Values.nfs.enabled .Values.azureFile.enabled }}
+# To provide data for the jupyterhub/grafana-dashboards dashboard about free
+# space in the shared volume, which contains users home folders etc, we deploy
+# prometheus node-exporter to collect this data for prometheus server to scrape.
+#
+# This is based on the Deployment manifest in jupyterhub/grafana-dashboards'
+# readme: https://github.com/jupyterhub/grafana-dashboards#additional-collectors
+#
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: shared-dirsize-metrics
+  labels:
+    app: jupyterhub
+    component: shared-dirsize-metrics
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: jupyterhub
+      component: shared-dirsize-metrics
+  template:
+    metadata:
+      annotations:
+        # This enables prometheus to actually scrape metrics from here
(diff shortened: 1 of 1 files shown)


yuvipanda:
Experimental grafana dashboard on LEAP cluster:

image

Lookin good.

As it churns through this, it is being CPU limited at the 0.05 we have given it, which seems alright with me.

pnasrat (changes requested):
Could you add an entry say in the support section of the infra docs on where to look for these dashboards

Maybe expanding the below 

https://infrastructure.2i2c.org/sre-guide/support/home-dir/

yuvipanda:
@pnasrat the graph panel doesn't exist yet, I just made a PR to add it: https://github.com/jupyterhub/grafana-dashboards/pull/80.

I'll add some docs too.

yuvipanda:
@pnasrat given the situation with utoronto and q planning, I won't be able to write docs this week. Are you ok with me merging this now so it can start collecting data, and I'll work on docs before end of the month?

pnasrat:
As an async team documentation for new services including operational procedures (eg how to turn off)  is necessary to ensure the team can both use and troubleshoot new services

If you want to merge without docs, please add a specific feature flag enabled and test in a single staging hub cluster to prove it out rather than turn on everywhere

yuvipanda:
Alright, I'll just mark this as draft and let this be until I have time to come back to it.

yuvipanda:
I think the right place for documenting this is upstream, so I've finally opened https://github.com/jupyterhub/grafana-dashboards/pull/84 to setup an actual docs site for the grafana dashboard. Once that's merged, I'll add a page documenting this panel.

yuvipanda:
I've added documentation about this graph to https://github.com/jupyterhub/grafana-dashboards/pull/84, so I think this is ready for review again.

pnasrat (approved):
LGTM

Can you make sure you announce this on the engineering channel so folks know there is a new component and graph they might find useful

yuvipanda:
Thanks @pnasrat, and will do once I confirm data is coming in!

PR #2623, opened 2023-06-07
utoronto, ingress-nginx: add azure required annotation to Service

consideRatio:
When Yuvi upgraded utoronto's k8s cluster, the k8s Service for ingress-nginx that is of type LoadBalancer stopped receiving traffic.

It seems that this was caused by a failing health check by a AKS k8s service-controller, which then concluded the ingress-nginx-controller pods supposed to receive traffic via the Service should't get traffic. The issue was that that health check had changed and had to be configured now via an annotation.

This PR adds such annotation to the ingress-nginx Service resource, not just for utoronto which is an azure/AKS based cluster, but for all helm charts. The annotation is azure specific, so my assumption is that this should be fine in non-azure based clusters.

We could put this in utoronto config as well, but then we need to document and keep manage a specific step for Azure based cluster's helm chart config. I lean towards thinking setting it systematically is preferred but am open to setting it just for utoronto.

---

EDIT: Config put in utoronto specifically, but no docs are added on this being required for all AKS based clusters yet.
EDIT: docs added

### DIFF
--- a/config/clusters/utoronto/support.values.yaml
+++ b/config/clusters/utoronto/support.values.yaml
@@ -42,3 +42,17 @@ grafana:
       - secretName: grafana-tls
         hosts:
           - grafana.utoronto.2i2c.cloud
+
+ingress-nginx:
+  controller:
+    service:
+      annotations:
+        # This annotation is a requirement for use in Azure provided
+        # LoadBalancer.
+        #
+        # ref: https://learn.microsoft.com/en-us/azure/aks/ingress-basic?tabs=azure-cli#basic-configuration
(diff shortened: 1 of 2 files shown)


yuvipanda:
I would definitely like us to keep this contained just for azure specific things rather than uniformly apply it across all cloud providers.

consideRatio:
Updated to reflect Yuvi's comment - @pnasrat what do you think?

pnasrat (approved):
lgtm

pnasrat:
I agree with @yuvipanda as  under an active incident the focus of changes should be constrained to recovery to operable state, so keeping it to utoronto here makes sense and we can generalize to automatically add all azure at a later point

consideRatio:
Thank you @pnasrat!!

Issue #2594, opened 2023-06-01
Ensure UofT machines are using an up-to-date version of Ubuntu


sgibson91:
### Context

Ticket ref: https://2i2c.freshdesk.com/a/tickets/750

The following email from Azure was forwarded to support:

> Canonical Ubuntu 18.04 LTS will reach end of standard support on 31 May 2023.
> Ubuntu 18.04 LTS end of standard support is 31 May 2023—ensure your VMs are on a supported operating system You're receiving this email because you've deployed Ubuntu 18.04 on Azure.
> On 31 May 2023, Ubuntu 18.04 LTS will reach the end of standard support. After this date, you'll still be able to use your existing virtual machines, however, security, maintenance, and feature updates will no longer be provided by Canonical and your VMs may become vulnerable.
> To avoid potential vulnerabilities, review the following options and choose the one that best fits your needs:
>  
> * Upgrade to Ubuntu 20.04 LTS <body-link> or Ubuntu 22.04 LTS <body-link>, the latest operating systems which offer improved performance and hardware enablement.
> * Purchase Ubuntu Pro <body-link>, which includes security patching for all Ubuntu packages. Ubuntu Pro 18.04 LTS will remain fully supported until April 2028.
> Read more details <body-link> about these options.
> Recommended action To ensure your VMs are running on a supported OS and can still receive security updates, upgrade <body-link> Ubuntu or purchase <body-link> Ubuntu Pro by 31 May 2023.
> More information If you have questions, get answers from community experts in Microsoft Q&A <body-link>. If you have a support plan and you need technical help, please create a support request <body-link>.
> Links provided herein may take you to a third-party website and are provided for convenience only. Third-party websites are subject to the third-party's terms and privacy statements.
>  
> Account information Subscription ID: ead3521a-d994-4a44-a68d-b16e35642d5b
> Subscription name: P574_ACT_JUPYTER

### Action point

- [ ] Ensure the machines on the UofT subscription are using an up-to-date version of Ubuntu

consideRatio:
This relates to https://github.com/2i2c-org/meta/issues/539

yuvipanda:
Given I don't fully trust tf with Azure, I'm going to do the following:

- [ ] Use the Azure portal to upgrade the cluster to 1.26 (currently at 1.22)
- [ ] Use the Azure portal to upgrade the nodepools to 1.26 (currently at 1.22)
- [ ] Use the Azure portal to upgrade the nodepools base image
[...]

Issue #890, opened 2021-12-10
AzureFile NFS network settings may block terraform access once applied


sgibson91:
### Description

https://github.com/sgibson91/pilot-hubs/pull/94 partnered with https://github.com/2i2c-org/infrastructure/pull/887 represents an effort to get NFS working on AzureFile storage and involved making some network changes in terraform so that the NFS share could be accessed and mounted by the k8s nodes.

While working on the Carbon Plan Azure cluster, I applied this new terraform config and then ran another `terraform plan` command, mostly to confirm to myself that the infrastructure was up-to-date, however I ran into this error message:

```
│ Error: shares.Client#GetProperties: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code="AuthorizationFailure" Message="This request is not authorized to perform this operation.\nRequestId:8150832e-d01a-0012-63ad-edeedb000000\nTime:2021-12-10T10:02:58.0047296Z"
│ 
│   with azurerm_storage_share.homes,
│   on storage.tf line 21, in resource "azurerm_storage_share" "homes":
│   21: resource "azurerm_storage_share" "homes" {
```

I am now worried that by making the NFS accessible to k8s, we have locked ourselves out from managing the infrastructure via terraform.

### Value / benefit

We need to retain access via terraform to sustainably manage infrastructure.

yuvipanda:
This is probably https://github.com/hashicorp/terraform-provider-azurerm/issues/2977. Looks like https://github.com/hashicorp/terraform-provider-azurerm/pull/14220 is supposed to fix it.

In the meantime, can we 'ignore' that particular change somehow in terraform so we can move forward with other changes?

sgibson91:
> In the meantime, can we 'ignore' that particular change somehow in terraform so we can move forward with other changes?

This is probably my fault for excessively trimming the error message. This error crops up during the "refreshing state" phase of terraform plan, it hasn't even got the the point of calculating the change yet, but it can't check the current state of the file share. So there's nothing to 'ignore'.

sgibson91:
Full error message:

> $ tf plan -var-file=projects/carbonplan.tfvars -out=carbonplan -refresh-only
> azurerm_resource_group.jupyterhub: Refreshing state... [id=/subscriptions/c5e7a734-3dbf-4285-80e5-4c0afb1f65dc/resourceGroups/2i2c-carbonplan-cluster]
[...]

PR #2624, opened 2023-06-07
Cleanup Azure terraform

yuvipanda:
- Mark optional parts of node / dask node definition as optional, so utoronto.tfvars will actually apply
- Parameterize core node size, and specify it explicitly.
- Remove default for k8s version, specify it explicitly. This matches the current k8s version
- Parameterize storage size, and match it to current reality. Note that this can't be applied via tf quite yet, due to https://github.com/2i2c-org/infrastructure/issues/890.

Ref https://github.com/2i2c-org/infrastructure/issues/2594


pnasrat (approved):
LGTM

Issue #2436, opened 2023-03-29
Post to a specific slack channel whenever GitHub Actions for hub deployment fails


yuvipanda:
Currently, we get no notifications if a deployment fails. It would be nice to have GitHub action *deployment* failures post to the #gh-failures channel on Slack.

pnasrat:
Definitely. I think it will also make us feel the pain of the API rate limiting failures that I frequently hit #1686 etc

PR #2625, opened 2023-06-08
Add step that notifies Slack when jobs in the hub deployement workflo…

GeorgianaElena:
…w fail

Fixes https://github.com/2i2c-org/infrastructure/issues/2436

When or if https://github.com/integrations/slack/issues/1563 is implemented, we should switch to that to avoid adding an extra step on every workflow's job.


pnasrat (approved):
What channel will these notifications go to?


Otherwise LGTM

GeorgianaElena:
It should go to `#gha-failures` 🤞🏼

PR #2626, opened 2023-06-08
Document auto-assignement workflow and consider moving the required o…

GeorgianaElena:
I added a section in our CI/CD docs about the auto-assignment workflow.

The docs also reflect another step I wish to take, which is to:
- [x] remove the sync trigger from this workflow (in this PR)
- [x] add another workflow similar to this one that triggers on PR sync events to the `.github` repo (PR to come)
- [x] make the workflow in `.github` required rather than this one because I don't want it to add up to the GitHub API rate limit.

This is because I feel like I've been seeing rate limit error more frequently since adding that.

Reference:  https://github.com/2i2c-org/infrastructure/issues/1686

### DIFF
--- a/.github/workflows/auto-author-assign-pull-request.yaml
+++ b/.github/workflows/auto-author-assign-pull-request.yaml
@@ -6,12 +6,6 @@ on:
     types:
       - opened
       - reopened
-      # Normally, it's enough for this workflow to be triggered only once, on PR opening/reopening.
-      # But because we use this workflow as a required workflow for other repositories
-      # in the 2i2c organization, it is **expected** to run successfully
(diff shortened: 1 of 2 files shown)


sgibson91 (commented):
Hey, I think this is great! Can we update some of the language in the docs though please? You use the phrase "this repository" a lot, but when we're viewing the docs on the readthedocs site it may not be obvious that you are referring to `infrastructure`! Could you make that a bit more explicit please?

Otherwise, LGTM!

GeorgianaElena:
Oh, you're right! Updated the doc section with more clear details about the repositories involved.

Issue #2628, opened 2023-06-08
Investigate AzureAD login issues with utoronto


pnasrat:
### Context

Reported via freshdesk https://2i2c.freshdesk.com/a/tickets/764

### Updates and actions

<details><summary>log grep for 500</summary>
<p>
```
ovyan@hub-749bb69658-fpk6n:/srv/jupyterhub$ tail -10000 jupyterhub.log | grep \]\ 500
[E 2023-06-08 20:12:29.357 JupyterHub log:186] 500 GET /hub/oauth_callback?code=[secret]&state=[secret]&session_state=[secret] (@10.1.0.4) 160.73ms
[E 2023-06-08 20:41:24.749 JupyterHub log:186] 500 GET /hub/oauth_callback?code=[secret]&state=[secret]&session_state=[secret] (@10.1.0.7) 220.95ms
[E 2023-06-08 20:41:44.609 JupyterHub log:186] 500 GET /hub/oauth_callback?code=[secret]&state=[secret]&session_state=[secret] (@10.1.0.7) 140.10ms
[E 2023-06-08 20:49:14.602 JupyterHub log:186] 500 GET /hub/oauth_callback?code=[secret]&state=[secret]&session_state=[secret] (@10.1.0.15) 331.61ms
jovyan@hub-749bb69658-fpk6n:/srv/jupyterhub$ tail -20000 jupyterhub.log | grep \]\ 500
[W 2023-06-08 19:33:47.616 JupyterHub web:1796] 500 GET /hub/user/2c94f0af-dd06-4e88-8563-a6bfa646d6aa/rstudio/?redirects=2 (10.1.0.5): Redirect loop detected. Notebook has jupyterhub version 3.1.0, but the Hub expects 3.1.1. Try installing jupyterhub==3.1.1 in the user environment if you continue to have problems.
[E 2023-06-08 19:33:47.617 JupyterHub log:186] 500 GET /hub/user/2c94f0af-dd06-4e88-8563-a6bfa646d6aa/rstudio/?redirects=2 (2c94f0af-dd06-4e88-8563-a6bfa646d6aa@10.1.0.5) 2.87ms
[E 2023-06-08 19:46:00.632 JupyterHub log:186] 500 GET /hub/oauth_callback?code=[secret]&state=[secret]&session_state=[secret] (@10.1.0.10) 258.97ms
[...]

PR #2636, opened 2023-06-09
Switch utoronto R hub to use CILogon

yuvipanda:
AzureAD has given us trouble in the past too. This makes that CILogon's problem, and the current issues with AzureAD don't seem to happen here. This can't currently work for the main hub because we use opaque oid as user identifier, rather than email. If we can make a mapping, that should also be possible.

Ref https://github.com/2i2c-org/infrastructure/issues/2628

### DIFF
--- a/config/clusters/utoronto/enc-r-prod.secret.values.yaml
+++ b/config/clusters/utoronto/enc-r-prod.secret.values.yaml
@@ -1,20 +1,24 @@
 jupyterhub:
+    singleuser:
+        extraFiles:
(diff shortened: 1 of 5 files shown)


yuvipanda:
Deployed and seems to work.

pnasrat (approved):
LGTM

yuvipanda:
This does remove the private gh puller support, we can add that back in later.

yuvipanda:
Added.

PR #2641, opened 2023-06-09
Move utoronto hub to using UToronto login via CILogon

yuvipanda:
Follow-up to https://github.com/2i2c-org/infrastructure/pull/2636

Ref https://github.com/2i2c-org/infrastructure/issues/2628

MUST BE MERGED ONLY AFTER WE RUN A SCRIPT THAT MATCHES EXISTING OID BASED USERNAMES TO EMAIL BASED USERNAMES!


pnasrat:
Can you test the migration process works with symlinks against your account on staging as if I remember correctly if the notebook container is using a subMount it won't resolve/follow the link

yuvipanda:
@pnasrat oooh, so that's what I expected to happen as well. BUT, it's not what actually happened.

If I just deploy this to staging, and login, I get `yuvi.panda@utoronto.ca` as my username, and here's the `mount` output from inside the user server:

```
2i2cutorontohubstorage.file.core.windows.net:/2i2cutorontohubstorage/homes/prod/yuvi-2epanda-40utoronto-2eca on /home/jovyan type nfs4 (rw,relatime,vers=4.1,rsize=1048576,wsize=1048576,namlen=255,hard,proto=tcp,timeo=600,retrans=2,sec=sys,clientaddr=10.1.0.7,local_lock=none,addr=20.150.16.40)
```

You can see it's mounting `/2i2cutorontohubstorage/homes/prod/yuvi-2epanda-40utoronto-2eca`, as expected.

Now, if I delete the directory and make it a symlink to my existing dir, with ` ln -s 7c76d04b-2d2a80-2d4db1-2db985-2da2d2fa2f708c/ yuvi-2epanda-40utoronto-2eca`, and then restart my user server up, I see this in `mount`:

```
2i2cutorontohubstorage.file.core.windows.net:/2i2cutorontohubstorage/homes/prod/7c76d04b-2d2a80-2d4db1-2db985-2da2d2fa2f708c on /home/jovyan type nfs4 (rw,relatime,vers=4.1,rsize=1048576,wsize=1048576,namlen=255,hard,proto=tcp,timeo=600,retrans=2,sec=sys,clientaddr=10.1.0.7,local_lock=none,addr=20.150.16.40)
```

You'll see that the mount is at `/2i2cutorontohubstorage/homes/prod/7c76d04b-2d2a80-2d4db1-2db985-2da2d2fa2f708c`, which is actually the resolved path! So maybe because it's Azure file, but looks like just symlinks *will* work here for us, as it seems to be resolved prior to mounting.

yuvipanda:
I was able to verify that end user could write contents as well, and this is reflected across too.

yuvipanda:
I've made a list of oids we would need mapping for, with the following script run on the output of `ls` on the homedirectories:

```python
from escapism import unescape
import sys


with open(sys.argv[1]) as f:
    for escaped_username in f:
        # All oids have 5 - (encoded as -2d) separated parts in there
        if len(escaped_username.split('-2d')) == 5:
            print(unescape(escaped_username.strip(), '-'))
```

yuvipanda:
First, note that on Utoronto, **staging and prod share home directories!!!**. So while you can deploy to staging to test, the home directories are same. So when performing operations manually, *be careful*. I attempted to create a snapshot, but I couldn't find a way to create a snapshot for an NFS file share (vs SMB share).

I've  created a script that'll take as input a *csv file* with first column being oid, and second being email. It'll do the escaping properly, matching whatever JupyterHub does, and create symlinks as appropriate.

```python
import os
from escapism import unescape, escape
import sys
import csv
import string
import subprocess

safe_chars = set(string.ascii_lowercase + string.digits)

with open(sys.argv[1]) as f:
    reader = csv.reader(f)
    for row in reader:
        # Expects a csv file with first column being oid, second being email
        oid, email = row

        escaped_oid = escape(oid, safe=safe_chars, escape_char='-').lower()
        escaped_email = escape(email, safe=safe_chars, escape_char='-').lower()

        if os.path.exists(escaped_email):
            print(f'{escaped_email} exists, not making a symlink')
        else:
            cmd = [
                'ln',
                '-s',
                escaped_oid,
                escaped_email
            ]

            print(f'Running {" ".join(cmd)}')
            subprocess.check_call(cmd)
```

You can run this by doing the following:

```bash
# Get a shell with home directories mounted
deployer exec-homes-shell utoronto staging    
# Look at the script, a copy of which I've put under `/home`
cd /home
cat symlink.py
# In another terminal, copy the csv file to `/home`
deployer use-cluster-credentials utoronto      
k -n staging cp mapping.csv utoronto-staging-shell:/home/mapping.csv
# back in first terminal, run the script
apt update
apt install python3 python3-escapism
python3 symlink.py mapping.csv
```

yuvipanda:
Also, the new IP for jupyter.utoronto.ca is  20.220.157.28. The Old IP used to be 20.200.99.250

Issue #2435, opened 2023-03-29
Q2 2023 Goal: Move CloudBank hubs to their own data8 style image


yuvipanda:
# Rationale
The default image we use for 2i2c hubs has grown well beyond i[ts initial purpose](https://github.com/2i2c-org/infrastructure/issues/2336#issuecomment-1480842625), and needs a lot of catchup maintenance. The process around it also needs fixing. Many universities using us have a break in this time period, making it possible to change images.

# Definition of done
1. We have an image built for all the cloudbank hubs that teach data8 style classes
2. Stretch goal: We have clear guidance on when this particular image would be updated, and how that can be communicated to users

# TODO

- [x] Specify a `common.yaml` file for cloudbank, specifying the image + tag https://github.com/2i2c-org/infrastructure/pull/2437
- [x] Build a data8-style image that is small and specific
- [ ] Figure out which hubs were just teaching data8, and which are not
- [ ] Decide on a date to do the swap for data8 teaching classes

yuvipanda:
We'd probably also want to make it easy to extend, so there can be a temporary extension for different versions of otter.

sean-morris:
Re: Who is using R on the cloudbank hubs

Who: Edward Pineda-Castro at Los Angeles City College
Hub: https://lacc.cloudbank.2i2c.cloud/

He is using R Studio on the hub with a couple of different classes.

ericvd-ucb:
@sean-morris  what about CSU hub with CSU Sacramento ?
Aren't they also R users?

https://csu.cloudbank.2i2c.cloud/
Clark Fitzgerald at CSU Sacramento

sean-morris:
> @sean-morris what about CSU hub with CSU Sacramento ? Aren't they also R users?
> 
> https://csu.cloudbank.2i2c.cloud/ Clark Fitzgerald at CSU Sacramento

CSUS moved to a different solution; we don't have any R needs on the csu hub

yuvipanda:
Well, no more cloudbank hubs are using the old image so I think we can call this one done!

PR #2645, opened 2023-06-10
Update cloudbank image to use new data8 style image

yuvipanda:
Before we merge this, we must use the configurator to specify the old image for any hubs that want to use otter 3 or RStudio.

Ref https://github.com/2i2c-org/infrastructure/issues/2435
Ref https://github.com/2i2c-org/infrastructure/issues/2435

### DIFF
--- a/config/clusters/cloudbank/common.values.yaml
+++ b/config/clusters/cloudbank/common.values.yaml
@@ -1,5 +1,5 @@
 jupyterhub:
   singleuser:
     image:
-      name: quay.io/2i2c/2i2c-hubs-image
-      tag: "14107b8a85fb"
+      name: quay.io/2i2c/cloudbank-data8-image
+      tag: aa7e0df7adf4


yuvipanda:
@sean-morris can you use the configurator to keep the current image for hubs that want to keep using otter 3 or RStudio? We can work on the RStudio stuff later.

sean-morris:
@yuvipanda I want to touch base more directly with pilots this week. I sent an email out but have not heard back from the contacts. 

I am wondering if a redeploy occurs how does the hub know that we used the configurator to change the image? 

Give me a couple of days to move this merge.

sean-morris (approved):
This looks great. I tested it on about five hubs with no issues. 

I have not had a chance to check in with about eight pilots. I should get to them over the next two weeks. I will use the configurator to put them on the old image and we can start integrating the new image into the workflow. 

Also the R image has become not necessary on our end.

PR #2649, opened 2023-06-12
support, prometheus: fix issue from v21 regression / v22 upgrade

consideRatio:
v21.0.0 of prometheus chart introduced a regression, and v22.0.0 fixed it, but it was breaking changes involved as v21 introduced a Deployment's `spec.selector.matchLabels` label that would be changing between versions (not good!), and v22 removed it.

The v22 bump was made in https://github.com/2i2c-org/infrastructure/pull/2593 with a workflow that also failed.

### DIFF
--- a/deployer/cluster.py
+++ b/deployer/cluster.py
@@ -61,6 +61,18 @@ class Cluster:
         )
         print_colour("Done!")
 
+        # FIXME: Revert this addition after one upgrade is made
+        print_colour("Fixing prometheus v22 migration...")
+        subprocess.check_call(
+            [
+                "kubectl",
+                "delete",
+                "--namespace=support",
+                "deploy/support-prometheus-server",
+            ]
+        )
+        print_colour("Done!")
+
         print_colour("Provisioning support charts...")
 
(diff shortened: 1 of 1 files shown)


deployer/cluster.py
```diff
@@ -61,6 +61,18 @@ def deploy_support(self, cert_manager_version):
...
+        subprocess.check_call(
+            [
+                "kubectl",
+                "delete",
+                "--namespace=support",
+                "deploy/support-prometheus-server",
+            ]
+        )
+        print_colour("Done!")
+
```
consideRatio, line 75: The fix idea is to delete the deployment first, and then let the helm chart install add it back. Like that, no change request will need to be made as part of removing the matchSelector label which isn't allowed because that section is immutable.

yuvipanda (approved):
I generally suggest doing this manually instead of going through CI/CD (https://github.com/2i2c-org/infrastructure/pull/2516#issuecomment-1530900520), but not going to block this on that.

Feel free to self-merge the revert as well.

PR #2650, opened 2023-06-12
carbonplan, openscapes, 2i2c-ohw: cleanup of misc config

consideRatio:
- 2i2c, ohw: stop continuous prePuller, supposed to be disabled earlier
- carbonplan: remove redundant config
- carbonplan: remove no longer necessary opt-out of readinessProbe
- carbonplan + openscapes: stop disabling hub network policy

### DIFF
--- a/config/clusters/openscapes/common.values.yaml
+++ b/config/clusters/openscapes/common.values.yaml
@@ -243,9 +243,6 @@ basehub:
           matchNodePurpose: require
     hub:
       allowNamedServers: true
-      networkPolicy:
-        # FIXME: For dask gateway
-        enabled: false
       readinessProbe:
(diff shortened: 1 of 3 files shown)


config/clusters/openscapes/common.values.yaml
```diff
@@ -243,9 +243,6 @@ basehub:
           matchNodePurpose: require
     hub:
       allowNamedServers: true
-      networkPolicy:
-        # FIXME: For dask gateway
-        enabled: false
```
yuvipanda, line 248: Is this set someplace else now?

yuvipanda (approved):
Thanks, @consideRatio

PR #2653, opened 2023-06-12
cilogin delete_client: Remove duplicated params

pnasrat:
Fix duplicated admin_id and admin_secret call params in delete_client

Ref:

`TypeError: stored_client_id_same_with_cilogon_records() takes 5 positional arguments but 7 were given`

### DIFF
--- a/deployer/cilogon_app.py
+++ b/deployer/cilogon_app.py
@@ -314,8 +314,6 @@ def delete_client(admin_id, admin_secret, cluster_name, hub_name, client_id=None
         if not stored_client_id_same_with_cilogon_records(
             admin_id,
             admin_secret,
-            admin_id,
-            admin_secret,
             cluster_name,
             hub_name,
             client_id,


GeorgianaElena (approved):
Thank you @pnasrat!

pnasrat:
@GeorgianaElena I just added a commit updating the docs as the code requires the cluster/hub yaml else exits - with

`CILogon records are different than the client app stored in the configuration file. Consider updating the file.`

PTAL (please take another look)

GeorgianaElena:
> CILogon records are different than the client app stored in the configuration file. Consider updating the file.

Hmm, yes! Thank you for spotting this issue! I remember wanting to have one `delete` command requiring cluster and hub name and the config file to exist and one requiring just the client id for when the file doesn't exist.
But I guess I messed this up in later implementations.

@pnasrat, does it work without the `--client-id`? That flag should be optional, but I wonder if there's a weird logic there as well.

consideRatio (approved):

PR #2656, opened 2023-06-13
Add some explicitly requested packages to UToronto R Hub

yuvipanda:
Brings in https://github.com/2i2c-org/utoronto-r-image/pull/6.

Ref https://2i2c.freshdesk.com/a/tickets/741


consideRatio:
@yuvipanda as a support steward I'm looking through tickets and evaluating if this is resolved.

Note that I saw `tidymodels` was requested by not explicitly added in https://github.com/2i2c-org/utoronto-r-image/pull/6 (and `learnr` was added but not explicitly requested).

Do you think this is resolved?

damianavila:
For future readers, they will send package requests as they arise, so the referenced ticket was closed.

PR #2657, opened 2023-06-13
support, redirects: fix failure to acquire HTTPS certs for redirected domains

consideRatio:


helm-charts/support/values.schema.yaml
```diff
@@ -70,12 +72,13 @@ properties:
...
-              type: integer
+            type:
+              type: string
+              enum: ["redirect", "permanent"]
               description: |
-                HTTP Status code to use for the redirect.
+                Relates to the HTTP Status code to use for the redirect.
 
-                302 (Temporary Redirect) is used if none is specified
+                Specify "redirect" for 302 (default), or "permanent" for 301.
```
consideRatio, line 81: The `code` configuration wasn't used, so I switched it to allowing users to specify `type` as that was easier to implement using the `rewrite` nginx config.

helm-charts/support/templates/redirects.yaml
```diff
@@ -1,21 +1,29 @@
...
+*/}}
+{{- range .Values.redirects.rules }}
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
+  name: ingress-redirect-{{ .from }}
   annotations:
     cert-manager.io/cluster-issuer: letsencrypt-prod
-    nginx.ingress.kubernetes.io/server-snippet: |
-      return {{ .code | default 302 }} $scheme://{{ .to }}$request_uri;
```
consideRatio, line 8: The core issue was that this took precedence of all routing of traffic by Ingress resources, not giving priority to the Ingress resource saying that `.well-known/acme-challenge` should be redirected to a certain cert-manager managed pod.

helm-charts/support/templates/redirects.yaml
```diff
@@ -1,21 +1,29 @@
...
-  name: ingress-redirect-{{ .from }}
+    nginx.ingress.kubernetes.io/configuration-snippet: |
+      rewrite "^\/(?!\.well-known\/acme-challenge)(.*)$" "$scheme://{{ .to }}/$1" {{ .type | default "redirect" }};
```
consideRatio, line 15: The strategy remains similar, but we use the `rewrite` config option that allows us to omit the `.well-known/acme-challenge` path.

yuvipanda (approved):

consideRatio:
Thank you @yuvipanda!!!

PR #2664, opened 2023-06-15
tf: Allow omitting optional fields in node definitions

yuvipanda:
`optional` is a newish feature of Terraform that lets us omit certain fields from node definitions. This makes node definitions much easier to read, and stops requiring us to change *all* .tfvar files each time we add another way a nodepool can be customized.


yuvipanda:
@consideRatio looks like the terraform autoformatter was not autoformatting tfvars files. I've fixed that now, so these should all work now I think.

Issue #2524, opened 2023-05-03
[Request deployment] New Hub: Climatematch Academy


colliand:
### The GitHub handle of the community representative

@abodner

### Hub important dates

**Target start date:** 2023-06-01
**Target end date:** 2023-08-31 

Heavy usage will take place during the course. The course will run July 17-28 2023.

### Hub Authentication Type

GitHub (e.g., @mygithubhandle)

### First Hub Administrators

- @abodner
- @WesleyTheGeolien
- Who else? **Abigail, please let us know whether other colleagues should be added in a comment below?**

### [GitHub Auth only] How would you like to manage your users?

Allowing members of specific GitHub team(s)

### [GitHub Teams Auth only] Profile restriction based on team membership

pending

**Abigail, can you please point to the GitHub team that Climatematch will use to manage user access to the hub?**

### Hub logo image URL

https://lh6.googleusercontent.com/pK1Zrf_NmWJ5KqhFB___4p8HPTf4D6u2om5UQkJbVQcwGjDSwlELPibkFfqW809chxybGrQwgiln8v0fRC00fYGzrsb6vIfFtsbh6PetpJKrk_UPoUb-4-RAH6ibtpXyxQ=w1280

### Hub logo website URL

https://academy.climatematch.io/

### Hub user image GitHub repository

pending, likely best to use latest pangeo image

### Hub user image tag and name

pending; likely latest pangeo image

### Extra features you would like to enable

- [ ] Dedicated Kubernetes cluster
- [ ] Scalable Dask Cluster

### (Optional) Preferred cloud provider

AWS

### (Optional) Billing and Cloud account

None

### Other relevant information to the features above

Climatematch Academy will train a cohort of ~1000 students in computational methods for climate science. The academy is partly inspired by Pangeo and builds on a similar virtual school in Neuroscience created and operated by Neuromatch. 

- shared cluster; cloud costs passed through by 2i2c/CS&S to be paid by Climatematch/Neuromatch
- for Pangeo style work? 2i2c can suggest best cloud vendor and data center; perhaps AWS us-west-2; or GCP?
- likely with a single hardware profile suitable for Pangeo-style student work; @abodner reports happiness using the `small` machine type deployed for M2Lines hub. 
- suggested hub url: climatematch.2i2c.cloud with staging.climatematch.2i2c.cloud for companion staging hub. @abodner and team may redirect using DNS to point at hub.climatematch.io.

### Tasks to deploy the hub

[...]

PR #2665, opened 2023-06-15
Bump climatematch hub memory request

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/2524#issuecomment-1592437489


yuvipanda:
(self merging as a resource bump)

PR #2667, opened 2023-06-16
Refactor workflows that use ghapi to use requests instead

sgibson91:
related to https://github.com/2i2c-org/infrastructure/issues/1686

- Extracts Python that was written inline in workflow files in their own scripts. Allows for linting, syntax highlighting, formatting, etc. This was mostly a copy-paste action.
- Replace use of `ghapi` with `requests` since we are suspicious of `ghapi`'s `paged` function: https://github.com/2i2c-org/infrastructure/issues/1686#issuecomment-1594434751
- Refactor workflow files to call the new scripts


extra-scripts/comment-deployment-plan-pr.py
```diff
@@ -0,0 +1,112 @@
...
+response = requests.get(url, headers=headers, params=params)
+all_artifacts = response["artifacts"]
+
+# If "Link" is present in the response headers, that means that the results are
+# paginated and we need to loop through them to collect all the results.
+# It is unlikely that we will have more than 100 artifact results for a single
+# worflow ID however.
+while "Link" in response.headers.keys():
+    response = requests.get(response.headers["Link"], headers=headers)
+    all_artifacts.extend(response["artifacts"])
```
sgibson91, line 50: This pattern won't work - working on a bugfix

sgibson91: I believe this is fixed by [`d5d0afa` (#2667)](https://github.com/2i2c-org/infrastructure/pull/2667/commits/d5d0afacb086bc5f8a70c96461bebbd83066917e)

consideRatio (approved):
Nice! I figure this needs to be tried and if it doesn't work out we can still manage daily operations so lets try it out!

Before merging it would be good if you do a check for `extra_scripts` references as well, I found for example this reference in dev-requirements.txt now that could benefit from an update.

```
# requests is used by extra_scripts/rsync-active-users.py
requests==2.*
```

Btw it can be nice to have a dedicated commit for rename/relocate as it can help with review efforts and tracking changes over time via git history.

sgibson91:
> Btw it can be nice to have a dedicated commit for rename/relocate as it can help with review efforts and tracking changes over time via git history.

There is one: [`b4f8ccd` (#2667)](https://github.com/2i2c-org/infrastructure/pull/2667/commits/b4f8ccd296c8c18e08f28960fbcdbf33ec943ba6)

consideRatio:
Ah true! I realize I was thinking about the extraction + change in https://github.com/2i2c-org/infrastructure/pull/2667/commits/62665a382bbba260ff1b5016127da3df55381af5 that didn't enable me to see the change from ghapi to requests in isolation.

Thanks for working on this @sgibson91!!

Issue #2582, opened 2023-05-24
Identify where 2i2c engineers spend time on image management


yuvipanda:
There's a general feeling that we spend a lot of time on image management that counts as toil. We should investigate what exactly we are spending time on, and use that to *inform* toil reduction work we do. It shouldn't be the *only* thing we use, but should help us validate wether something we work on will actually help or not. Doing things purely off this list would be too *reactive* and might not be exactly what we need.

Taking the last 6 months as a study period, I'm going through the following and collecting image related work we have done in this google spread sheet: https://docs.google.com/spreadsheets/d/1sWUsWvgkKsQKs1grvvL1BS3Ee4ufyAYq7blNK4zd0S4/edit#gid=0

- [x] Freshdesk tickets
- [x] GitHub PRs to this repository
- [x] New Hub issues that result in work in other repositories

Here is a summary of my coding from the Sheet

# Support tasks, GH PRs related to image management

Task Type | Count 
-- | --
UI Install | 3 
Package Install | 7
Image Consult | 3
Deploy profileList | 31
Deploy image | 8
Debug notebook infra | 2
Change default UI | 2
Build New Image form Scratch | 2
Add repo2docker-action to an existing repo | 2

Note that a limitation here is that we don't know *how much time* was spent on each of these, just that we got these many of them. For example, image consult probably took more time than deploying a profile list!

# How images are built

We have a total of 36 unique images, several of which I couldn't actually find the source for.

Type | Number of Hubs using them 
-- | --
Default Image (on Cloudbank) | 20
Default image (not on Cloudbank) | 5
Full Dockerfile | 1
Upstream Pangeo Image | 17
repo2docker | 10
NA | 7
Modified rocker, pangeo, matlab | 8 (3 + 3 + 2)
Modified conda/mambaforge | 4
Full fork of jupyter/docker-stacks | 1

# Profile Lists
Type | No. of Hubs
-- | --
Present | 18 (34% of all hubs)
Image selected as a profile, no dropdown | 2 (11% of hubs with profileLists)
Image selected as a dropdown on a profile | 7 (38% of hubs with profileLists)

Code used for a lot of this data is at https://gist.github.com/yuvipanda/7bca69d42d7b849fa17d9e8c3da4404f

pnasrat:
@yuvipanda this is great and really informative. One of my initial takeaways here is that if we maintained the current spread of things scaling for 10 x that would mean a large number of images!

[...]

PR #2671, opened 2023-06-19
Switch to jupyter/scipy-notebook as 'default'

### LINKED ISSUE #2336: Intervention to have new hubs not couple to the 2i2c-hubs-image, but something more up to date

yuvipanda:
https://github.com/2i2c-org/infrastructure/issues/2336 lists issues with the current 'default' image, and the history of how it came to be. The primary reason for its existence was addressed in https://github.com/2i2c-org/infrastructure/issues/2435, so it fundamentally does not really need to exist anymore.

However, there are still 4 hubs using that image, inherited from the default. This PR does the following:

- Switches to jupyter/scipy-notebook as the 'default', when nothing else is specified. Note that because all our hubs have an image specified *somewhere* except for these 4 (see the spreadsheet linked to from https://github.com/2i2c-org/infrastructure/issues/2582), this will actually have *no effect* for existing hubs at all! Just means that *future* hubs will get this image as the default.
- Explicitly sets the image for the 4 hubs still using the old 'default' image. These communities will need to be reached out to, and the image changed.

After this, I believe we can archive the old 'default' image!

Ref https://github.com/2i2c-org/infrastructure/issues/2336 (EDIT by Erik: fixes #2336)

### DIFF
--- a/config/clusters/2i2c/aup.values.yaml
+++ b/config/clusters/2i2c/aup.values.yaml
@@ -26,6 +26,9 @@ jupyterhub:
           name: AUP
           url: https://www.aup.edu/
   singleuser:
+    image:
+      name: quay.io/2i2c/2i2c-hubs-image
+      tag: "14107b8a85fb"
     memory:
       limit: 8G
       guarantee: 6G
--- a/config/clusters/2i2c/pfw.values.yaml
+++ b/config/clusters/2i2c/pfw.values.yaml
@@ -25,6 +25,10 @@ jupyterhub:
         funded_by:
           name: JROST & IOI
           url: https://investinopen.org/blog/jrost-rapid-response-fund-awardees
+  singleuser:
+    image:
+      name: quay.io/2i2c/2i2c-hubs-image
+      tag: "14107b8a85fb"
   hub:
     config:
       JupyterHub:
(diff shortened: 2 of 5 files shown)


consideRatio (approved):
Wieee looks great!!

I opened #2674 with the remaining work items you wanted to track, allowing #2336 be closed by this PR.

PR #2672, opened 2023-06-19
Move *most* hubs on shared cluster to filestore (2i2c and cloudbank cluster)

yuvipanda:
Looking at Grafana, temple is the only hub that seems to currently consistently have users. The process for moving a hub over is:

1. Wait for the hub to have no users
2. Do an rsync just for that hub's home directories. For most of these hubs I expect this will be pretty quick.
3. If no users had logged on while this rsync was happening, proceed. If not, go to 1.
4. Delete the PV, PVC and pod providing shared homedirectory metrics. This will need to be done manually once, as we are changing the PV and that is immutable. This is a rare enough event that not automating this is fine, plus I don't actually want to automate deleting PVs (just in case it deletes data we actually want!). Note that if any users are actually on the hub, it won't actually delete (thanks to https://kubernetes.io/docs/concepts/storage/persistent-volumes/#storage-object-in-use-protection)
5. Deploy the change, move things to the new homedirectory.
6. Repeat!

Ref https://github.com/2i2c-org/infrastructure/issues/1105

### DIFF
--- a/config/clusters/2i2c/basehub-common.values.yaml
+++ b/config/clusters/2i2c/basehub-common.values.yaml
@@ -1,0 +1,9 @@
+nfs:
+  enabled: true
+  pv:
+    mountOptions:
+      - soft
+      - noatime
+    serverIP: 10.234.45.250
+    # MUST HAVE TRAILING SLASH
+    baseShareName: /homes/
--- a/config/clusters/2i2c/basehub-common.values.yaml
+++ b/config/clusters/2i2c/basehub-common.values.yaml
@@ -6,4 +6,4 @@ nfs:
       - noatime
     serverIP: 10.234.45.250
     # MUST HAVE TRAILING SLASH
-    baseShareName: /homes/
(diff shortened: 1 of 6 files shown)


yuvipanda:
The deletion can be done with:

```bash
NS=<namespace>
k -n ${NS} delete pvc/home-nfs pv/staging-home-nfs $(k -n $NS get pod -l component=shared-volume-metrics -o name)
```

yuvipanda:
Tested on staging (and put back to the old NFS server)

consideRatio (approved):
Looks great @yuvipanda!!!

Super happy about all efforts to align hubs with the practices as they have developed over time, it is a significant complexity reduction!

yuvipanda:
I see just temple is active now, so going to go for this :)

yuvipanda:
@consideRatio I've reverted your edit to the title as this is *just* for the 2i2c cluster, not the cloudbank one.

PR #2676, opened 2023-06-20
Remove support for SMB based AzureFiles

yuvipanda:
We *used* to not use NFS when running on Azure, with AzureFiles. However, we use NFS there too now, and it works fine. We only have one Azure instance, and it uses NFS.

This just removes all the code required for Azure File to work with SMB.


consideRatio (approved):
Wieeeeeeeee nice!!!

Issue #2583, opened 2023-05-26
Document when to use what kind of image building strategy


yuvipanda:
At JupyterCon, I gave a talk on what I think of as the 4 different primary ways to build images common in our ecosystem.

<img width="1383" alt="image" src="https://github.com/2i2c-org/infrastructure/assets/30430/633b0638-8c40-43dd-bbbc-6431a27c8bca">

(slides at https://docs.google.com/presentation/d/16V6ylmirxxBTlVq-tZ9dsFYCf1Abbl83Rv350Rc-DL4/edit#slide=id.g23f74c3a6a0_0_106)

We currently have a bunch of different images using different building mechanisms. In particular, when we expect the community to self-manage the image, we suggest they use the template built in https://github.com/2i2c-org/hub-user-image-template/, which uses *repo2docker files*. While if we manage the image fully ourselves, we often use a `Dockerfile` explicitly (https://github.com/2i2c-org/utoronto-image/) or inherit from another (https://github.com/2i2c-org/utoronto-r-image).

I propose we make explicit and adopt the criteria for choosing which image building technique to use from the talk (summarized in the slide above), and then do an audit of all the images we manage to move them to an appropriate method. Some images are also using repo2docker files *purely* because that is the only template we have available - we should also make at least a template for inheriting from a community maintained image available.

PR #2683, opened 2023-06-20
Offer image choices for the JMTE hub

yuvipanda:
I think overall, we want to reduce the number of images we maintain for our end users. A big part of this to use upstream images directly wherever possible, and allow users to choose. This helps us benefit from upstream fixes as quickly as possible, and reduces the total amount of work done. For example, instead of specifically bumping the version of Julia just for this one
image (https://github.com/pangeo-data/jupyter-earth/pull/166), we could instead do that upstream and benefit
everyone (https://github.com/jupyter/docker-stacks/pull/1917).

Faster startup times is another benefit, as the more specific images are smaller than a big 'all-in-one' image.

There are some features of the all-in-one image that currently don't easily exist upstream:

- Linux desktop
- Nix
- Specific extra packages that maybe installed

We can figure these out over time, but not maintaining the all-in-one 
image is a nice goal to shoot for. To this end, the JMTE image is
still the default, but marked as 'deprecated' as I don't want to
continue doing a lot of maintenance on it.

Ref https://github.com/2i2c-org/infrastructure/issues/2201

This is how it looks like.

image

### DIFF
--- a/config/clusters/jupyter-meets-the-earth/common.values.yaml
+++ b/config/clusters/jupyter-meets-the-earth/common.values.yaml
@@ -126,8 +126,34 @@ basehub:
             mem_limit: null
             node_selector:
               node.kubernetes.io/instance-type: m5.xlarge
+          profile_options: &profile_options
+            image: 
+              display_name: Image
+              choices:
+                jmte:
+                  display_name: JMTE all-in-one image (deprecated)
+                  default: true
+                  slug: "jmte"
+                  kubespawner_override:
+                    image: "286354552638.dkr.ecr.us-west-2.amazonaws.com/jmte/user-env:9baee2d"
+                tensorflow:
+                  display_name: Pangeo Tensorflow ML Notebook
+                  slug: "tensorflow"
+                  kubespawner_override:
+                    image: "quay.io/pangeo/ml-notebook:2023.05.18"
+                pytorch:
+                  display_name: Pangeo PyTorch ML Notebook
+                  slug: "pytorch"
+                  kubespawner_override:
+                    image: "quay.io/pangeo/pytorch-notebook:2023.05.18"
+                datascience:
+                  display_name: Jupyter DockerStacks DataScience Image (Python, Julia, R)
+                  slug: "datascience"
+                  kubespawner_override:
+                    image: "jupyter/datascience-notebook:2023-06-19"
         - display_name: "4th of Medium: 1-4 CPU, 4-16 GB"
           description: "A shared machine."
+          profile_options: *profile_options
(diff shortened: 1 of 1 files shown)


yuvipanda:
I will communicate this to the users of the hub.

yuvipanda:
The following message was communicated to hub users:

Hello hub users! More exciting changes (and features) afoot, thanks to integration with the 2i2c infrastructure! Once https://github.com/2i2c-org/infrastructure/pull/2683 lands, you'll be able to specifically choose from multiple images to launch into your session! To begin with, there are 4 images:

The existing, 'all-in-one' big image. No changes here.
Upstream Pangeo image for pytorch. (1) is based off an older version of this, so this newer image will bring in newer versions of libraries that may help!
Upstream pangeo image for tensorflow. New!
Upstream jupyter/docker-stacks' 'datascience notebook' image, which brings in an up to date version of Julia as well as R.

In the long run, I'd like us to stop maintaining (1) and instead allow users to choose from various special purpose images for the following reasons:

Faster startup time! The all in one image is about 9G big, while the special purpose images are much smaller. For example, the datascience-notebook image is only 1.8G, compared to 9G for our all-in-one image! This means pod startups should be much faster.
Image changes are shared across the whole ecosystem, and hence faster. There's a larger community maintaining these images than just us, so this benefits everyone. The example I have is of bumping version of Julia I did for our image (https://github.com/pangeo-data/jupyter-earth/pull/166), vs me doing that upstream (https://github.com/jupyter/docker-stacks/pull/1917).
Allows us to introduce more features without paying a huge complexity penalty. The more combination of things you want in a single image, the more complex it becomes. It's far easier to have a special purpose image that just does one thing. For example, an image focused on showing the Linux Desktop would be far more suited to that than putting that here.

Nothing changes right now, as the old image is still available. However, I ask that y'all try out the other images and let me know what you think! Particularly,

Are you using GPUs? Try out the pytorch & tensorflow images and let me know if they are enough, or if further modifications are needed
Are you using Julia? Try out the jupyter/datascience-notebook image, and let me know if that is enough!

@here

yuvipanda:
https://github.com/2i2c-org/infrastructure/issues/2583 is related

yuvipanda:
A good example of this is the fact that the `julia` package is installed in the JMTE image for interfacing between python and julia, but it doesn't actually work because that package isn't compatible by default with python installed from conda! I'd rather we solve that upstream than downstream here :)

GeorgianaElena (approved):
This makes sense to me and agree it's a good pattern to follow! Thank you @yuvipanda

PR #2684, opened 2023-06-21
Udpate openscapes matlab staging image

GeorgianaElena:
Per https://2i2c.freshdesk.com/a/tickets/778


consideRatio (approved):
I guess we lack a better option to allow a user to help test something on staging than to do this for now.

PR #2685, opened 2023-06-21
support chart, prometheus: allow long startup times and adjust mem limit for pangeo-hubs

consideRatio:
- support chart, prometheus: ensure prometheus-server has time to startup
- pangeo-hubs: increase prometheus memory from 20Gi to 30Gi


consideRatio:
Thanks @GeorgianaElena!!

PR #2686, opened 2023-06-21
daskhub: stop scraping worker pods for dask specific metrics

### LINKED ISSUE #2366: daskhub: prometheus scraping of worker pods?
I see from our dask-gateway client configuration, we make annotate our worker pods to be scraped. Is that used by us? If not, we should not do this because prometheus can easily be overloaded with scraping work I think. Especially if someone starts 100-1000 workers with 100 or so nodes that also have node metrics to be scraped.

https://github.com/2i2c-org/infrastructure/blob/cf7bfc3778001791bd83472b943fc695ea5e314d/helm-charts/daskhub/values.yaml#L189-L190

consideRatio:
Our prometheus-server instances blow up in memory use and gets OOMKilled when dask-gateway is used extensively. Why isn't obvious, but I suspect scraping individual dask-worker pods metrics adds to the problem, and since we don't use the dask-worker specific metrics, we shouldn't scrape them.

- Closes #2366

### DIFF
--- a/config/clusters/jupyter-meets-the-earth/common.values.yaml
+++ b/config/clusters/jupyter-meets-the-earth/common.values.yaml
@@ -308,7 +308,7 @@ dask-gateway:
                 if ":" not in options.image:
                     raise ValueError("When specifying an image you must also provide a tag")
                 extra_labels = {}
-                extra_annotations = {
+                scheduler_extra_pod_annotations = {
                     "prometheus.io/scrape": "true",
                     "prometheus.io/port": "8787",
                 }
@@ -321,9 +321,9 @@ dask-gateway:
                     # A default image is suggested via DASK_GATEWAY__CLUSTER__OPTIONS__IMAGE env variable
                     "image": options.image,
                     "scheduler_extra_pod_labels": extra_labels,
-                    "scheduler_extra_pod_annotations": extra_annotations,
+                    "scheduler_extra_pod_annotations": scheduler_extra_pod_annotations,
                     "worker_extra_pod_labels": extra_labels,
-                    "worker_extra_pod_annotations": extra_annotations,
+                    "worker_extra_pod_annotations": scheduler_extra_pod_annotations,
                     "worker_cores": 0.85 * chosen_worker_cpu,
                     "worker_cores_limit": chosen_worker_cpu,
                     "worker_memory": "%fG" % (0.85 * chosen_worker_memory),
--- a/helm-charts/daskhub/values.yaml
+++ b/helm-charts/daskhub/values.yaml
@@ -182,7 +182,7 @@ dask-gateway:
             def option_handler(options):
                 if ":" not in options.image:
                     raise ValueError("When specifying an image you must also provide a tag")
-                extra_annotations = {
+                scheduler_extra_pod_annotations = {
                     "hub.jupyter.org/username": safe_username,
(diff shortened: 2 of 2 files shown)


yuvipanda:
Let's turn this off just for the pangeo cluster? I think it's helpful to collect this data for usage, and even if we currently don't have it we will in the future. I also think there's one dashboard already that displays this data?

consideRatio:
> Let's turn this off just for the pangeo cluster? I think it's helpful to collect this data for usage, and even if we currently don't have it we will in the future.

We have seen failures like this in LEAP as well when dask-gateway is heavily used. We have a stability issue unresolved for all active large-scale users of dask-gateway I'd say, and this mitigates that.

I'd like to have this turned off until we do know we need metrics from individual dask-worker pods. Note that we still get metrics like memory usage etc, because that is provided by node-exporter pod - we just don't get dask-worker provided metrics which involves dask specific metrics we don't consume via a grafana daskboard or similar. I've not turned off dask-scheduler specific metrics as that seems more plausible that we could want to inspect.

The dask-worker metrics we wouldn't scrape are these: https://distributed.dask.org/en/latest/prometheus.html#worker-metrics. I really think we shouldn't scrape those before we have a use case for it. It has a cost to do this - we need bigger requests for prometheus-server driving costs and our time to handle when prometheus-server has crashed.

consideRatio:
> I also think there's one dashboard already that displays this data?

Which?

consideRatio:
These are metrics that still will be collected because they come from node-exporter scraping metrics a lot of pods. We will miss out on metrics like the number of threads in the specific dask-worker etc.

image

yuvipanda:
This is a very important conversation to have, but I'd like to decouple that from resolving the incident ongoing right now. Can we do the smallest thing possible right now, and move forward from there?

This dashboard uses them: https://grafana.leap.2i2c.cloud/d/usage-report/usage-report?orgId=1

image

yuvipanda:
Ah, alright. I have opinions but not time, so I'll just disengage. If you feel this is the right thing to do, please proceed.

yuvipanda (approved):

consideRatio:
Thanks @yuvipanda!! I'm confident about this decision is whats right at the moment, and I don't expect any dashboard we have to stop working

config/clusters/jupyter-meets-the-earth/common.values.yaml
```diff
@@ -321,9 +321,9 @@ dask-gateway:
                     # A default image is suggested via DASK_GATEWAY__CLUSTER__OPTIONS__IMAGE env variable
                     "image": options.image,
                     "scheduler_extra_pod_labels": extra_labels,
-                    "scheduler_extra_pod_annotations": extra_annotations,
+                    "scheduler_extra_pod_annotations": scheduler_extra_pod_annotations,
                     "worker_extra_pod_labels": extra_labels,
-                    "worker_extra_pod_annotations": extra_annotations,
+                    "worker_extra_pod_annotations": scheduler_extra_pod_annotations,
```
damianavila, line 326: Out of curiosity, any reason to still keep the worker pods annotations in the JMTE case?

consideRatio: Nope, this is a mistake! Nice catch!

Issue #2695, opened 2023-06-22
pangeo-hubs: help prometheus-server startup again (node pool changes, I lack permissions)


consideRatio:
We have a prometheus-server that is OOMKilled and can't live on the core node pool because it runs out of memory, where the largest request we can make is 21.7Gi is already to little on the current n1-highmem-4 nodes that are specified to provide 26 GB of memory.

I suggest an upgrade to the n2-highmem-4 nodes like used by default for new daskhub clusters and in many others already, they provide 32GB of memory, and may help us get prometheus-server functional again.

Another option is to provide a new node pool for prometheus server by tainting it and letting prometheus server tolerate it, or both.

I can't work this because I lack access to pangeo-hubs cloud console.

GeorgianaElena:
@consideRatio, we should all have access to pange-hubs, but not through our @2i2c google accounts, but though ones provided by columbia with `@columbia.edu` I think. Can you check if you have such an account?

GeorgianaElena:
Btw, I added this issue and https://github.com/2i2c-org/infrastructure/issues/2688 to the sprint board with a `critical` label, so we take advantage that we're still planning the next sprint and we can plan them accordingly. cc @damianavila

PR #2701, opened 2023-06-22
Increase size of pangeo-hubs core nodes & prometheus RAM

yuvipanda:
Fix https://github.com/2i2c-org/infrastructure/issues/2695

### DIFF
--- a/terraform/gcp/projects/pangeo-hubs.tfvars
+++ b/terraform/gcp/projects/pangeo-hubs.tfvars
@@ -2,7 +2,7 @@ prefix                 = "pangeo-hubs"
 project_id             = "pangeo-integration-te-3eea"
 zone                   = "us-central1-b"
 region                 = "us-central1"
-core_node_machine_type = "n1-highmem-4"
+core_node_machine_type = "n1-highmem-8"
 enable_private_cluster = true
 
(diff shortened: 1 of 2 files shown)


terraform/gcp/projects/pangeo-hubs.tfvars
```diff
@@ -2,7 +2,7 @@ prefix                 = "pangeo-hubs"
 project_id             = "pangeo-integration-te-3eea"
 zone                   = "us-central1-b"
 region                 = "us-central1"
-core_node_machine_type = "n1-highmem-4"
+core_node_machine_type = "n1-highmem-8"
```
consideRatio, line 5: I suggest sticking with n2 series at all times as they provide a robust 1:8 ratio of CPU:RAM while the n1 series is an exception to the ratio of 1:8, providing only 1:6.5. The cost increase is 10% but the RAM increase is 23%, and CPU performance appears to be ~20-30% faster as well.

```suggestion
core_node_machine_type = "n2-highmem-8"
```

There is a case to be made that if we never need to bump this again that its better to stay with the slightly cheaper n1 machine so you can merge without going n2, but I think we should stick with n2 by default when choosing.

consideRatio (approved):

consideRatio:
Thank you @yuvipanda!!!

PR #2702, opened 2023-06-23
jmte, daskhub: fix failure to stop scraping dask-worker metrics

consideRatio:
A mistake was made in https://github.com/2i2c-org/infrastructure/pull/2686 that was spotted by @damianavila in https://github.com/2i2c-org/infrastructure/pull/2686/files#r1239775340, where we still scraped JMTE's dask-worker pod's metrics.


damianavila (approved):
LGTM

PR #2711, opened 2023-06-24
Run reports automatically & publish them to a repo

yuvipanda:
# Background

Regularly scheduled 'reports' can help the team keep track of the status of various
parts of our infrastructure. Examples of code written in the recent past that would be
useful as a report include:

1. Producing a list of all our clusters, the kubernetes version they are in (both master
    and nodes) for https://github.com/2i2c-org/infrastructure/issues/2157#issuecomment-1584292226
2. Produce a list of all images and the hubs they are used in, taking into account the
    different ways images can be specified for the hub (and a bunch of others) for
    https://github.com/2i2c-org/infrastructure/issues/2582

The code used for generating these is scattered in gists and slack, and also require
someone with appropriate credentials to run the code, and upload the result somewhere.
There is also no easy way to trace a particular output back to the code that produced it,
nor a way to discover these.

# Automated reports

This PR adds a `reports/` directory, which can contain any number of Jupyter Notebooks.
Each notebook comprises a 'report', and a GitHub Action will run every week that will:

1. Execute the notebook with [papermill](https://papermill.readthedocs.io/), generating a copy with the filename being the current date. This has access to the same context as our hub deployment script, so can access all the kubernetes clusters, prometheus API, etc. 
2. Push that output to a separate repository (currently https://github.com/yuvipanda/2i2c-reports/, should be moved into the 2i2c org). A sops encrypted [deploy-key](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/managing-deploy-keys) is used to do this pushing.
3. The target repository uses sphinx + nb-myst to build a static site, currently published [here](https://yuvipanda.github.io/2i2c-reports/). You can see the output of an individual notebook like https://yuvipanda.github.io/2i2c-reports/kubernetes-versions/2023/06/24-07h33m32sUTC/index.html. Using myst-nb allows us to use [tags like `hide-cell` and `hide-input`](https://myst-nb.readthedocs.io/en/v0.9.0/use/hiding.html) to control what gets shown in the output.

# TODO

- [ ] Generate a ToC appropriately
- [ ] Add docs on the reporting system to this repo
- [ ] Move the repo containing the reports into the 2i2c org


damianavila:
This is a really nice idea, @yuvipanda!

yuvipanda:
I'm not going to be working on this anytime soon, we can pick this back up later if it gets prioritized via our sprint process.

PR #2724, opened 2023-06-27
adding multiple images to choose from removing node sharing options

jmunroe:
Changes based on a conversation with @yuvipanda that I am attempting to make on my own.

### DIFF
--- a/config/clusters/2i2c-aws-us/researchdelight.values.yaml
+++ b/config/clusters/2i2c-aws-us/researchdelight.values.yaml
@@ -56,172 +56,66 @@ basehub:
         name: quay.io/2i2c/researchdelight-image
         tag: "872f0c4578af"
       profileList:
-        # NOTE: About node sharing
-        #
-        #       CPU/Memory requests/limits are actively considered still. This
-        #       profile list is setup to involve node sharing as considered in
-        #       https://github.com/2i2c-org/infrastructure/issues/2121.
-        #
-        #       - Memory requests are different from the description, based on:
-        #         whats found to remain allocate in k8s, subtracting 1GiB
-        #         overhead for misc system pods, and transitioning from GB in
-        #         description to GiB in mem_guarantee.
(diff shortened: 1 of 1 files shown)


yuvipanda:
@jmunroe take a look at https://github.com/2i2c-org/infrastructure/pull/2726 for docs on how to use the rocker images, hope that helps?

jmunroe:
Thanks @yuvipanda -- those Rocker instructions are very clear!

config/clusters/2i2c-aws-us/researchdelight.values.yaml
```diff
@@ -56,172 +56,82 @@ basehub:
...
+            mem_limit: null
+            node_selector:
+              node.kubernetes.io/instance-type: r5.xlarge
+
+        - display_name: "Medium: 16 CPU, 128 GB"
+          description: "A dedicated machine for you."
+          profile_options: *profile_options
           kubespawner_override:
-            cpu_limit: null
+            mem_guarantee: 120.513G
```
yuvipanda, line 123: Can you tell me where these numbers came from?

yuvipanda:
The image stuff looks good to me, although we should try it out and see. I had a question about where the guarantee numbers have come from...

config/clusters/2i2c-aws-us/researchdelight.values.yaml
consideRatio: Its the node share option request size when requesting an entire machine, where all other options were slimmed away.

My calculations were using allocatable memory and removing 4GB for headroom, and then dividing that into 1, 2, 4, 8, 16, 32, 64, 128 to plan the requested share sizes to choose from.

In this case, it works fine to use this number, but could be anything being 51% of the nodes memory to end up having a dedicated node as no smaller request options are available.

yuvipanda: Ah cool cool

yuvipanda (approved):

Issue #2725, opened 2023-06-27
Home folder issues with use of rocker-org/rocker-versioned2 images with rstudio


consideRatio:
When using a `rocker/binder` based image, jupyterlab and rstudio are installed, together with https://github.com/jupyterhub/jupyter-rsession-proxy to proxy to rstudio via jupyter-server-proxy.

We are not using the `jovyan` user in those images though but instead the `rstudio` user, and we mount storage to `/home/jovyan` and end up in `/home/jovyan` when we start into jupyterlab etc. I wonder how we should go about this, but we want to:

- be some linux user in the container
- mount our storage to that user folder
- preferably we are the jovyan user consistently rather than sometimes being the `jovyan` and sometimes being `rstudio`
- have both jupyterlab and rstudio work from that directory

Ideas on addressing this...

- We could mount user storage to to `/home/rstudio`
- We could maybe trick rstudio to work from `/home/jovyan` like jupyterlab is tricked, perhaps like setting `HOME` env variable or something
- We could maybe start successfully also with the jovyan user as configured via kubespawner on the container
- We could maybe do something when RStudio starts via https://github.com/jupyterhub/jupyter-rsession-proxy
- We could maybe configure RStudio somehow

In [@yuvipanda's guide](https://jupyterhub-image.guide/rocker.html#step-7-setup-zero-to-jupyterhub-configuration-for-home-directory), the home folder is mounted to /home/rstudio instead, but it would be nice to do a less invasive config instead, such as declaring that we want to start as the jovyan user instead of the rstudio user which is the default in the rocker docker image.

I'm not sure what to do, but it would be great for us if we can avoid mounting to a new location depending on the image we use. In the [case of smithsonian](https://2i2c.freshdesk.com/a/tickets/787), they want to choose from both a rocker based image and a jupyter-docker stacks based image, so they would then even need that config to be dependent on the image choice.

consideRatio:
> We could maybe start successfully also with the jovyan user as configured via kubespawner on the container

There is no "jovyan" user, the key property is the user id of `1000` rather than the username. If I for exaple start `docker run -it --user jovyan` I can get that the user doesn't exist, it won't be created on the fly or similar if it doesn't have a reference to a user id already.

It is possible to do something like `--user 1000` though, but that is the `rstudio` user anyhow.

PR #2726, opened 2023-06-27
Add documentation on how to use rocker images

yuvipanda:
Fixes https://github.com/2i2c-org/infrastructure/issues/2725

### DIFF
--- a/docs/howto/features/index.md
+++ b/docs/howto/features/index.md
@@ -32,6 +32,7 @@ shared-db.md
 :maxdepth: 2
 :caption: User image
 image.md
+rocker.md
 ```
 
 ```{toctree}
--- a/docs/howto/features/rocker.md
+++ b/docs/howto/features/rocker.md
@@ -1,0 +1,131 @@
+# Use Rocker based R images
+
+The [Rocker](https://rocker-project.org/) project is a community led set of
+containers for use in the R ecosystem, similar to the [jupyter/docker-stacks](https://jupyter-docker-stacks.readthedocs.io/en/latest/)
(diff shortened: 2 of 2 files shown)


GeorgianaElena (approved):

consideRatio (approved):
Thank you @yuvipanda!!!

I hoped we could find a better way than directly configuring the volume mounts from the profile overrides, but thinking about that I've come up blank after thinking about some ideas that I had to reject.

config/clusters/smithsonian/common.values.yaml
```diff
@@ -97,7 +97,17 @@ basehub:
                   default: true
                   slug: geospatial
                   kubespawner_override:
-                    image: rocker/geospatial:4.3
+                    image: rocker/binder:4.3
```
consideRatio, line 100:
```suggestion
                    # rocker/binder builds on top of rocker/geospatial and will
                    # for releases after 4.3.1 include jupyterhub as required to
                    # be compatible with our jupyterhub.
                    #
                    # FIXME: Replace quay.io/2i2c/rocker-binder with
                    #        rocker/binder as soon as a version beyond 4.3.1 is
                    #        released or https://github.com/rocker-org/rocker-versioned2/pull/671 resolves it.
                    #
                    image: quay.io/2i2c/rocker-binder:4.3.1
```

yuvipanda: @consideRatio this is not necessary, as the image was rebuilt and pushed to contain jupyterhub now https://github.com/rocker-org/rocker-versioned2/pull/668#issuecomment-1610066198

PR #2727, opened 2023-06-27
VEDA: Update pangeo-notebook image

jsignell:
:wave: Hi!

I wanted to update the pangeo-notebook image on the VEDA hub. In case this is helpful (@ranchodeluxe mentioned that it might be) it looks like the jupyterhub version is 4.0.1 and the jupyterlab version is 4.0.2. 

As a side note: Is there a standard cadence at which these images get updated or is this the right way to do it?

### DIFF
--- a/config/clusters/nasa-veda/common.values.yaml
+++ b/config/clusters/nasa-veda/common.values.yaml
@@ -66,7 +66,7 @@ basehub:
       defaultUrl: /lab
       image:
         name: pangeo/pangeo-notebook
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thanks @jsignell! We're working on finding an automated cadence for this, but until then bumps are manual.

yuvipanda:
Also, welcome @jsignell!

jsignell:
Woohoo!! Thanks Yuvi!

PR #2731, opened 2023-06-28
prometheus: avoid loosing more than 10m data during restarts

consideRatio:
When we clear the prometheus write ahead log (WAL) on startup, it seems we loose data even when its termenating regularly during for example prometheus chart upgrades. The amount of data lost seem to be up to 2 hours or so.

From summarizing an issue about memory peaking in https://github.com/prometheus/prometheus/issues/6934#issuecomment-1610921555, I think it makes sense for us to pass two flags to help us only loose up to ~30 minutes of metrics. This change has also been observed to reduce memory spikes.

```
# ...
# the end of startup in the line below
ts=2023-06-28T09:25:53.263Z caller=head.go:1293 level=info component=tsdb msg="Head GC completed" caller=truncateMemory duration=2.920359ms

# an event triggered 30m after startup below
ts=2023-06-28T09:55:56.731Z caller=compact.go:523 level=info component=tsdb msg="write block" mint=1687943452954 maxt=1687944600000 ulid=01H40PWMX1QF0JJ7FD4SEG9FG6 duration=1.81850325s
ts=2023-06-28T09:55:56.779Z caller=head.go:1293 level=info component=tsdb msg="Head GC completed" caller=truncateMemory duration=43.444241ms

# `kubectl delete pod` test
ts=2023-06-28T10:01:52.045Z caller=main.go:854 level=warn msg="Received SIGTERM, exiting gracefully..."
ts=2023-06-28T10:01:52.045Z caller=main.go:878 level=info msg="Stopping scrape discovery manager..."
ts=2023-06-28T10:01:52.045Z caller=main.go:892 level=info msg="Stopping notify discovery manager..."
ts=2023-06-28T10:01:52.045Z caller=manager.go:1009 level=info component="rule manager" msg="Stopping rule manager..."
ts=2023-06-28T10:01:52.045Z caller=manager.go:1019 level=info component="rule manager" msg="Rule manager stopped"
ts=2023-06-28T10:01:52.045Z caller=main.go:929 level=info msg="Stopping scrape manager..."
ts=2023-06-28T10:01:52.045Z caller=main.go:874 level=info msg="Scrape discovery manager stopped"
ts=2023-06-28T10:01:52.046Z caller=main.go:888 level=info msg="Notify discovery manager stopped"
ts=2023-06-28T10:01:52.047Z caller=main.go:921 level=info msg="Scrape manager stopped"
ts=2023-06-28T10:01:52.143Z caller=notifier.go:603 level=info component=notifier msg="Stopping notification manager..."
ts=2023-06-28T10:01:52.143Z caller=main.go:1150 level=info msg="Notifier manager stopped"
ts=2023-06-28T10:01:52.143Z caller=main.go:1162 level=info msg="See you next time!"
```

When I forcefully restarted after this PR, the observed data loss was about exactly 30 minutes, from 10:34 (11:34 in gif animation because of timezone stuff). I note that this was after the log message "Head GC completed" as seen in logs above. My guess is that we will have a consistent data loss of 30 min of data loss going onwards when prometheus-server is restarted.

restart-caused-data-loss

---

I also updated two comments in basehub values in a7d704f

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -351,7 +351,7 @@ jupyterhub:
         # Otherwise our prometheus server can't get hub metrics
         authenticate_prometheus: false
       KubeSpawner:
-        # Make sure working directory is ${HOME}
+        # Make sure working directory is where we mount the home folder
         working_dir: /home/jovyan
         # Increase timeout for Jupyter server to become 'ready', until
         # https://github.com/2i2c-org/infrastructure/issues/2047 is fixed
--- a/helm-charts/support/values.yaml
+++ b/helm-charts/support/values.yaml
@@ -126,12 +126,17 @@ prometheus:
   server:
     # Prometheus sometimes seems to run into bugs where the 'write ahead log'
     # used for recovery from crashes and network failures gets massive, and
-    # requires an order of magnitude more RAM to start up than usual (see
-    # https://github.com/prometheus/prometheus/issues/6934). Deleting the WAL
-    # at each startup means we *might* lose some data during restarts, but
-    # that is far preferable to having the entire prometheus server be down.
-    # This may be fixed by prometheus 2.45 which is not out yet
-    # https://github.com/prometheus/prometheus/issues/6934#issuecomment-1586060255
+    # requires an order of magnitude more RAM to start up than usual. This could
+    # have been greatly improved in prometheus 2.45 which is used as of
+    # 2023-06-28, but we stick with a workaround for now until other users have
+    # concluded things have greatly improved for them in prometheus 2.45+.
+    #
+    # ref: https://github.com/prometheus/prometheus/issues/6934#issuecomment-1610921555
+    #
+    # We are deleting the WAL at each startup now, and it has made us lose some
+    # data during restarts, but that is preferable to having the entire
+    # prometheus server be down.
+    #
     # We can't use initContainers here, as they are not run when a container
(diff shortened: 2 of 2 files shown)


yuvipanda:
@consideRatio great! Can we make this number even smaller? 10min?

consideRatio:
> Can we make this number even smaller? 10min?

I mean yes maybe, but I'm not able to say if its okay with 1m, 10m, 30m, or 2h (default for the max value I think). I have no real idea about the implications of changing this, but its a property I've understood prometheus maintainers prefer users don't fiddle much with through indications by the community when reading issues.

So I figure the answer is maybe. Currently it is 30m because I just took values suggested by others configuring it to 30m with positive results - but they didn't clear the WAL on restart either which causes regular negative results for us when we restart.

I'm okay with configuring this more agressively, but I'd rather play it safer and trial stopping deleting the WAL in a month or two now that 2.45 is out and this is configured to 30m instead of the 2h default. I'm happy to go with any judgement call of yours on this though, I have no real reason for opting for 30m besides being cautious of changing something I don't understand well.

yuvipanda (approved):
Let's try 10 and see what happens! Thanks for digging through this, @consideRatio!

yuvipanda:
And yes, I agree, let's try taking this out in a few weeks and see.

yuvipanda:
@consideRatio ok I bumped it back up to 1h in https://github.com/2i2c-org/infrastructure/pull/3093, let's see if that helps.

PR #2733, opened 2023-06-28
leap: Bump pangeo notebook version

jbusecke:
Unsure if I bumped the version in all the right places. Particularly I am not sure what the tag [here](https://github.com/2i2c-org/infrastructure/blob/13accf612fefb5a718c51849152f33b288fe8611/config/clusters/leap/common.values.yaml#L79) does.

### DIFF
--- a/config/clusters/leap/common.values.yaml
+++ b/config/clusters/leap/common.values.yaml
@@ -76,7 +76,7 @@ basehub:
     singleuser:
       image:
         name: pangeo/pangeo-notebook
-        tag: "2023.02.08"
+        tag: "2023.06.20"
       extraEnv:
         GH_SCOPED_CREDS_CLIENT_ID: "Iv1.0c7df3d4b3191b2f"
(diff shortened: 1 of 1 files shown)


jbusecke:
Saw some [errors](https://github.com/2i2c-org/infrastructure/actions/runs/5403107197/jobs/9815432776) in the deployment. Is that a concern?

jbusecke:
FWIW our hub shows some display errors, e.g. in the tabs of the dask-extension
image

Not sure this is related, but wanted to document.

consideRatio:
The staging hub deployed but the test-startup of a user server failed. It can happen if a new node had to start and image was slow to pull or similar. I tested myself and it works, so I'm deploying to the production hub now manually.

What you observe in the jupyterlab extension for dask is perhaps related to upgrading the images and the software in them. I'd guess either jupyterlab or the labextension was upgraded and there was a regression in the UI that probably relates to both codebases.

PR #2746, opened 2023-06-30
Bump openscapes matlab image

yuvipanda:
Ref https://2i2c.freshdesk.com/a/tickets/801


yuvipanda:
Self-merging as an image bump.

PR #2754, opened 2023-07-03
pangeo-hubs: add memory request/limit for prometheus-server

consideRatio:
This community has had 310 nodes started at the same time. When that
happens, prometheus server collapses, probably because scraping from
310+ node exporters at the same time is too much.

### DIFF
--- a/config/clusters/pangeo-hubs/support.values.yaml
+++ b/config/clusters/pangeo-hubs/support.values.yaml
@@ -35,3 +35,8 @@ prometheus:
         - secretName: prometheus-tls
           hosts:
             - prometheus.gcp.pangeo.2i2c.cloud
+    resources:
+      requests:
+        memory: 16Gi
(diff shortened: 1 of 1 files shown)


config/clusters/pangeo-hubs/support.values.yaml
```diff
@@ -35,3 +35,8 @@ prometheus:
         - secretName: prometheus-tls
           hosts:
             - prometheus.gcp.pangeo.2i2c.cloud
+    resources:
+      requests:
+        memory: 16Gi
+      limits:
+        memory: 16Gi
```
consideRatio, line 42: 16 Gi is quite arbitrary, but its 4 times larger than 4 Gi which it had by default.

sgibson91 (approved):
Let's give it a go and see if the issue reccurs

PR #2755, opened 2023-07-03
2i2c, climatematch: Logo update

jmunroe:
It appears that URL to the logo accessed https://academy.climatematch.io/ is not persistent -- I've copied the contents into a gist as a alternative place to host the image file for this logo. https://2i2c.freshdesk.com/a/tickets/804

### DIFF
--- a/config/clusters/2i2c/climatematch.values.yaml
+++ b/config/clusters/2i2c/climatematch.values.yaml
@@ -29,8 +29,8 @@ jupyterhub:
       templateVars:
         org:
           name: ClimateMatch Academy
-          # Logo picked up from https://academy.climatematch.io/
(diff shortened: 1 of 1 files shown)


jmunroe:
@2i2c-org/engineering Is a storing the image within a gist a reasonable solution? Or there is a better place to add an image that doesn't have an existing URL?

consideRatio:
> https://github.com/orgs/2i2c-org/teams/engineering Is a storing the image within a gist a reasonable solution? Or there is a better place to add an image that doesn't have an existing URL?

Great question, and I think overall github doesn't appreciate when they are used to host content like that, but I think for the time being this is a good interim decision. If the image was sent as a file, I don't think we have a better practice in place, and I don't have a suggestion better than doing this either.

consideRatio (approved):

consideRatio:
A trick I use when I quickly want to do this, is to simply post the image it to a github issue, like this one:
image

Then, inside it I get a URL, in this case https://github.com/2i2c-org/infrastructure/assets/3837114/ac3ec52f-1e62-440a-b2bd-8052e5f38897

jmunroe:
That's a brilliant trick! I'll have to remember that one.

jmunroe:
I changed it to use the image embedded within this PR comment section.  That seems like a better solution than a gist "owned" by my personal account.

consideRatio (approved):
Thanks @jmunroe!!!

PR #2763, opened 2023-07-05
nasa-veda: Pin to using JupyterLab 3

yuvipanda:
This brings in https://github.com/pangeo-data/pangeo-docker-images/pull/469, which
pins JupyterLab < 4 so we don't break jupyterlab-git or dask-labextension

Ref https://2i2c.freshdesk.com/a/tickets/798


consideRatio (approved):

ranchodeluxe:
Looks good @yuvipanda, thanks!

yuvipanda:
I've done one better, and worked with our upstream image to pin to JupyterLab < 4 so jupyterlab-git and dask-labextension don't break! Brings in https://github.com/pangeo-data/pangeo-docker-images/pull/469 now.

PR #2764, opened 2023-07-05
Bump julia image in JMTE hub

yuvipanda:
Brings in https://github.com/jupyter/docker-stacks/pull/1929


yuvipanda:
Self merging as an image bump

Issue #2753, opened 2023-07-03
[INCREASED HUB ACTIVITY] 2i2c climatematch


GeorgianaElena:
### Summary

The Climatematch community is expecting an increased hub activity starting with Wednesday **this week** until August 1st.

### Event Info

There are multiple triggers for the increased hub activity as specified in the https://2i2c.freshdesk.com/a/tickets/804 ticket:

- [additional nodes requested]: July 5-6 (Wednesday-Thursday this week) they are holding an instructor training event where they expect about 50 logins intermittently throughout the day.
- [no additional nodes requested] Next week they will be adding over 500 students to the course GitHub team and ask them to test access into the hub.
- [no additional nodes requested] The course officially begins on July 17. Between students, instructors, and content creators, thet expect about 700 active users during the two weeks of the live course. 
 - Activity will likely drop after August 1 but we will monitor usage after that, with potential renewal until the end of December to allow students to continue to work on projects.

### Hub info

- **Hub URL**:  https://climatematch.2i2c.cloud
- **Hub decommisioned after event?**:  No

pnasrat:
I'll handle the increase in nodes for Wednesday.

GeorgianaElena:
Thank you @pnasrat!

pnasrat:
Looking at the climate match configuration. kubespawner allocates notebook servers with a limit of 7G and guarantee of 5G onto a pool of n1-highmem-2	2vcPUs	13G, so we can only fit 2 users per node at guarantee but if both burst memory it'll cause a reschedule. 

That seems low. I'm wondering why this isn't larger nodes more densely packed with some overhead capacity

700 active users at peak would need 700 nodes which isn't ideal in my mind. 

. A user profile isn't suitable for this setup as it's a class. CC @consideRatio for thoughts on sizing

pnasrat:
Created initial pull request in #2757 to handle the instructor case but we should resolve the plan for actual node sizing for  next week   and the event itself.

consideRatio:
> That seems low. I'm wondering why this isn't larger nodes more densely packed with some overhead capacity

Legacy reasons I presume. When there are 100+ users, having at minimum 10 users per node is essential to not run out of quotas of node disks or public IPs and avoid startup times I'd say.

> we should resolve the plan for actual node sizing for next week and the event itself.

[...]

PR #2771, opened 2023-07-07
terraform, gcp: set location to cluster location for user and dask-worker node pools

consideRatio:
- terraform, gcp: add location to the user node specification
- terraform, gcp: collect info about k8s versions 1.24-1.27

PR #2775, opened 2023-07-07
climatematch larger node size

pnasrat:
See #2753


sgibson91 (approved):

pnasrat:
Manually applied with #2771 patched in

Issue #2769, opened 2023-07-07
Cluster architecture - core node pools in any zone, or in the user zones?


consideRatio:
In the terraform/gcp configuration, we provide `node_locations` for user nodes but not for the core nodes. That means that they will start in any zone in the regional cluster I think. In practice, this makes us able to end up with core nodes in another zone than the user nodes.

I suspect this is a bit inefficient, but perhaps not a big deal either. Are there costs of zone-to-zone networking etc that we want to avoid?

Changing this in the common terraform config may require re-creating the core nodes or similar, so I figure the only path towards configuring the core nodes to the zone(s) of the user nodes would have to be done cluster by cluster with a cluster-specific override until we can do it systematically in the common config.

yuvipanda:
> That means that they will start in any zone in the regional cluster I think.

This should not be true, as they should instead inherit the default from the cluster's node_locations: https://github.com/2i2c-org/infrastructure/blob/1aac5c8773698d7750505f92c9297768821cb73e/terraform/gcp/cluster.tf#L55. I'll investigate the recent issues again to make sure this is the case.

consideRatio:
@yuvipanda ah, looking at a newly created cluster without explicitly configuring node_locations for the core node pool I conclude you are right.

![node-zones](https://github.com/2i2c-org/infrastructure/assets/3837114/4c3a47cd-a1bb-4d38-a0cd-10459d438db4)

![image](https://github.com/2i2c-org/infrastructure/assets/3837114/46bab7df-2c7d-43d1-931a-d75537208524)

---

I draw the wrong conclusion from seeing how they were explicitly configured for the user nodes, but they are explicit there because they were changed over time perhaps?

yuvipanda:
In the latam cluster just created, I see:

```
$  terraform state show google_container_node_pool.core
# google_container_node_pool.core:
resource "google_container_node_pool" "core" {
    cluster                     = "latam-cluster"
    id                          = "projects/catalystproject-392106/locations/southamerica-east1/clusters/latam-cluster/nodePools/core-pool"
    initial_node_count          = 1
    instance_group_urls         = [
        "https://www.googleapis.com/compute/v1/projects/catalystproject-392106/zones/southamerica-east1-c/instanceGroupManagers/gke-latam-cluster-core-pool-01cfc23e-grp",
    ]
    location                    = "southamerica-east1"
    managed_instance_group_urls = [
[...]

PR #2777, opened 2023-07-07
Clarify when node_locations is overriden

yuvipanda:
If zones is not explicitly set for nodepools, it will inherit whatever is set for the cluster itself. This makes the code clearer so that is more obvious.

Fixes https://github.com/2i2c-org/infrastructure/issues/2769


consideRatio (approved):

yuvipanda:
I've applied this to leap and pilot-hubs and they were no-ops as expected.

consideRatio:
Beautiful thanks @yuvipanda!!

PR #2782, opened 2023-07-10
Bump prometheus volume size for 2i2c shared cluster

yuvipanda:
100Gi filled up on 30/6

Discovered by looking at https://grafana.pilot.2i2c.cloud/d/bUnRC_Qnz/nfs-and-support-information?orgId=1

### DIFF
--- a/config/clusters/2i2c/support.values.yaml
+++ b/config/clusters/2i2c/support.values.yaml
@@ -3,6 +3,9 @@ prometheusIngressAuthSecret:
 
 prometheus:
   server:
+    persistentVolume:
+      # 100Gi filled up!
+      size: 256Gi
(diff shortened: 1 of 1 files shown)


pnasrat:
Any reason not to give it 512 or 1T - basically a laptop disk prior to actually capacity planning

yuvipanda:
None! bumped to 1Ti

yuvipanda:
I've actually bumped it up to 512Gi instead, as I don't know if we'll cross that threshold with only 1y of retention. And we can resize up, but not down.

pnasrat (approved):
LGTM

yuvipanda:
Thanks @pnasrat

Issue #2128, opened 2023-02-01
[Request deployment] New Hub: CosmicDS/Harvard


colliand:
### Important dates

- **Target start date**: 2023-03-01
- **Required start date**: 2023-02-08
- **Any important dates for usage**: No

### Hub Authentication Type

Other (may not be possible, please specify in comments)

### First Hub Administrators

@nmearl, [Nicholas Earl](http://nicholasearl.me/)
@patudom, [Patricia Udomprasert](https://pweb.cfa.harvard.edu/people/patricia-udomprasert)

### [GitHub Auth only] How would you like to manage your users?

None

### Hub logo image URL

pending...

### Hub logo website URL

pending...

### Hub user image GitHub repository

pending...

### Hub user image tag and name

pending...

### Extra features you'd like to enable

- [ ] Specific cloud provider or datacenter (otherwise GCP)
- [ ] Dedicated Kubernetes cluster
- [ ] Scalable Dask Cluster

### (Optional) Preferred cloud provider

None

### (Optional) Billing and Cloud account

None

### Other relevant information to the features above

1. **access** CosmicDS is deployed into classes at various institutions. As a result, the access/auth layer is more complicated than it would be for a single university. CosmicDS has built and is improving upon a web app that users log into and this web app will be used as the SSO source of truth. Jim advised that CosmicDS may want to investigate CIlogon since this integration may then allow learners at various colleges and universities to access CosmicDS using their institutions credentials instead of with a separate login. I invite @nmearl to provide additional guidance in comments below. 
2. **will it work?** CosmicDS is deployed using a docker image that specifies the software environment inside JupyterHub. Nick built this and has experimented with it atop k8s in AWS. A delineation that is not clear at this point is whether CosmicDS may require deeper access to the JupyterHub deployment than is typically in the “software environment” scope. Jim and Nick discussed the idea to pilot a deployment using an education hub on a (2i2c owned) shared cluster with the understanding that we may need to change course and deploy instead on a dedicated cluster depending on the result of the pilot. I invite @nmearl to provide additional guidance in comments below. 

Navigating the complexities associated with auth and software environment are the responsibility of the CosmicDS team (with some input and suggestions from 2i2c's Engineering Team). If significant work and software development is required from 2i2c, we will need to revisit the business terms of this pilot.

### Tasks to deploy the hub

- [ ] 1. Deploy information filled in above
- [ ] 2. Engineer who will deploy the hub is assigned
[...]

PR #2809, opened 2023-07-12
Enable username anonymization feature for cosmicds

yuvipanda:
- Implement the feature
- Write documentation on how to enable it, and why not to
- Enable it for cosmicds

TODO:

- [ ] Describe *how* this works
- [ ] Validate that authorization can work with this at all

Ref https://github.com/2i2c-org/infrastructure/issues/2128#issuecomment-1630745032

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -757,3 +757,68 @@ jupyterhub:
             # Customize list of profiles dynamically, rather than override options form.
             # This is more secure, as users can't override the options available to them via the hub API
             c.KubeSpawner.profile_list = custom_profile_list
+
+      06-salted-username: |
+        # Allow anonymizing username to not store *any* PII
+        import json
+        import os
+        import base64
+        import hashlib
+        from z2jh import get_config
+
+        USERNAME_DERIVATION_PEPPER = bytes.fromhex(os.environ['USERNAME_DERIVATION_PEPPER'])
+
+        def salt_username(authenticator, handler, authentication):
+          # Combine parts of user info with different provenances to eliminate
+          # possible deanonym attacks when things get leaked.
+
+          # FIXME: Provide useful error message when using an auth provider that
+          # doesn't give us 'oidc'
+          # FIXME: Raise error if this is attempted to be used with anything other than CILogon
+
+          cilogon_user = authentication['auth_state']['cilogon_user']
+          user_key_parts = {
+            # Opaque ID from CILogon
+            "sub": cilogon_user['sub'],
+            # Combined together, opaque ID from upstream IDP (GitHub, Google, etc)
+            "idp": cilogon_user['idp'],
+            "oidc": cilogon_user['oidc']
+          }
+
+          # Use JSON here, so we don't have to deal with picking a string
+          # delimiter that will not appear in any of the parts.
+          # keys are sorted to ensure stable output over time
+          user_key = json.dumps(user_key_parts, sort_keys=True).encode('utf-8')
+
+          # The cryptographic choices made here are:
+          # - Use blake2, because it's fairly modern
+          # - Set blake2 to output 32 bytes as output, which is good enough for our use cas
+          # - Use base32 encoding, as it will produce maximum of 56 characters
+          #   for 32 bytes output by blake2. We have 63 charcter username
+          #   limits in many parts of our code (particularly, in usernames
+          #   being part of labels in kubernetes pods), so this helps
+          # - Convert everything to lowercase, as base64.b32encode produces
+          #   all uppercase charcters by default. Our usernames are preferably
+          #   lowercase, as uppercase characters must be encoded for kubernetes'
+          #   sake
+          # - strip the = padding provided by base64.b32encode. This is present
+          #   primarily to be able to determine length of the original byte
+          #   sequence accurately. We don't care about that here. Also = is
+          #   encoded in kubernetes and puts us over the 63 char limit.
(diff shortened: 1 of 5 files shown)


GeorgianaElena (commented):
@yuvipanda, I know we did something similar for another hub (callysto), where we had the hub usernames be the `oidc` claim directly and then authorize based on email.

Any reason why we're using something different here? 

https://github.com/2i2c-org/infrastructure/blob/044485ef605259e71fa50694c6ea57e33e46a536/config/clusters/callysto/common.values.yaml#L68-L108

yuvipanda:
@GeorgianaElena great question! I've written some more docs here on why I think this is needed, and what it offers over just using oidc: https://github.com/2i2c-org/infrastructure/pull/2809/files#diff-ac02bcb032968f861b77821bb371824dfa6bfad4da743117e853e463cb902682. Take a look and let me know what you think?

docs/howto/features/anonymized-usernames.md
```diff
@@ -0,0 +1,134 @@
...
+   breaches in those websites can also be used to de-anonymize our
+   users.
+   
+2. We live in a world where user data leaks are a fact of life, and you can buy
+   tons of user identifiers for pretty cheap. This may also happen to *us*, and
+   we may unintentionally leak data too! So users should still be hard to 
+   de-anonymize when the attacker has in their posession the following:
+
+   1. List of user identifiers (emails, usernames, numeric user ids,
+      etc) from *other data breaces*.
```
sgibson91, line 46:
```suggestion
      etc) from *other data breaches*.
```

helm-charts/basehub/values.yaml
```diff
@@ -757,3 +759,71 @@ jupyterhub:
...
+        # Allow anonymizing username to not store *any* PII
+        import json
+        import os
+        import base64
+        import hashlib
+        from z2jh import get_config
+
+        USERNAME_DERIVATION_PEPPER = bytes.fromhex(os.environ['USERNAME_DERIVATION_PEPPER'])
+
+        def salt_username(authenticator, handler, authentication):
```
GeorgianaElena, line 773: Just a small personal preference to use the same name for the dict info returned by the `authenticate` call [as it's in `oauthenticator`](https://github.com/jupyterhub/oauthenticator/blob/8ff022c0479893aede07714e522f5f48f97141d9/oauthenticator/oauth2.py#L973-L978) if we choose to not document the parameters.

```suggestion
        def salt_username(authenticator, handler, auth_model):
```

helm-charts/basehub/values.yaml
```diff
@@ -757,3 +759,71 @@ jupyterhub:
...
+          ).digest()).decode('utf-8').lower().rstrip("=")
+
+          # Replace the default name with our digested name, thus
+          # discarding the default name
+          authentication["name"] = digested_user_key
+
+          return authentication
+
+        if get_config('custom.auth.anonymizeUsername', False):
+          c.Authenticator.post_auth_hook = salt_username
```
GeorgianaElena, line 829:
```suggestion
        if get_config('custom.auth.anonymizeUsername', False):
          # https://jupyterhub.readthedocs.io/en/stable/reference/api/auth.html#jupyterhub.auth.Authenticator.post_auth_hook
          c.Authenticator.post_auth_hook = salt_username
```

helm-charts/basehub/values.yaml
GeorgianaElena: I now realized jupyterhub calls it `the authentication dict` while oauthenticator `auth_model`. Please feel free to ignore my suggestion.

helm-charts/basehub/values.yaml
```diff
@@ -757,3 +759,71 @@ jupyterhub:
...
+          }
+
+          # Use JSON here, so we don't have to deal with picking a string
+          # delimiter that will not appear in any of the parts.
+          # keys are sorted to ensure stable output over time
+          user_key = json.dumps(user_key_parts, sort_keys=True).encode('utf-8')
+
+          # The cryptographic choices made here are:
+          # - Use blake2, because it's fairly modern
+          # - Set blake2 to output 32 bytes as output, which is good enough for our use cas
```
GeorgianaElena, line 797:
```suggestion
          # - Set blake2 to output 32 bytes as output, which is good enough for our use case
```

helm-charts/basehub/values.yaml
```diff
@@ -757,3 +759,71 @@ jupyterhub:
...
+          # - Use base32 encoding, as it will produce maximum of 56 characters
+          #   for 32 bytes output by blake2. We have 63 charcter username
```
GeorgianaElena, line 799:
```suggestion
          #   for 32 bytes output by blake2. We have 63 character username
```

helm-charts/basehub/values.yaml
```diff
@@ -757,3 +759,71 @@ jupyterhub:
...
+          #   limits in many parts of our code (particularly, in usernames
+          #   being part of labels in kubernetes pods), so this helps
+          # - Convert everything to lowercase, as base64.b32encode produces
+          #   all uppercase charcters by default. Our usernames are preferably
```
GeorgianaElena, line 803:
```suggestion
          #   all uppercase characters by default. Our usernames are preferably
```

GeorgianaElena (approved):
Thanks for documenting the reasoning behind the updated anonymization practice @yuvipanda!

helm-charts/basehub/values.yaml
yuvipanda: `auth_model` is a better name, we should change upstream to that too! I'll change it here.

PR #2816, opened 2023-07-14
Fix a broken link in docs

sgibson91:


sgibson91:
Self-merging as it is not controversial and I've confirmed that this link is now correct in the readthedocs preview

PR #2815, opened 2023-07-14
New AWS cluster: catalystproject-africa with staging hub

sgibson91:
ref: https://github.com/2i2c-org/infrastructure/issues/2704

PR #2817, opened 2023-07-14
Tweak logic for when artifacts should be uploaded

sgibson91:
I noticed I wasn't getting a deployment plan comment on https://github.com/2i2c-org/infrastructure/pull/2815, even though the PR would trigger a deploy of the support chart and a staging hub - but there isn't a prod hub being added.

Due to the `and` logic in the statement controlling when artifacts get uploaded, this case was being missed and a deployment plan not provided. This PR changes that logic to include an `or`  which I believe which allow either one of the support-staging or prod matrices to be empty, but not both, for artifacts to be uploaded.


pnasrat (approved):
LGTM

PR #2819, opened 2023-07-14
Don't look for USERNAME_DERIVATION_PEPPER unless necessary

yuvipanda:
Follow-up to https://github.com/2i2c-org/infrastructure/pull/2809


yuvipanda:
Going to merge as my earlier PR broke hubs that didn't have this setting enabled. Only staging hubs though.

PR #2823, opened 2023-07-15
Bump image tag and set an environment variable for KubeSpawmer

ianabc:
This PR bumps the singleuser image for the callysto cluster and sets an environment variable for KubeSpawner. The environment variable seems to fix a minor issue some of our users have reported with plotly output occasionally vanishing. I'm not 100% sure where I should set the environment variable, so please let me know if it should be set somewhere else.


config/clusters/callysto/common.values.yaml
```diff
@@ -106,6 +106,7 @@ jupyterhub:
             return None
 
         c.JupyterHub.authenticator_class = EmailAuthenticatingCILogonOAuthenticator
+        c.KubeSpawner.environment = {'PLOTLY_RENDERER': 'notebook_connected'}
```
yuvipanda, line 109: I would suggest using `singleuser.extraEnv` instead.

yuvipanda (changes requested):

config/clusters/callysto/common.values.yaml
yuvipanda: Alternatively, you can also set this in your image! That *may* be a better fit, given this is probably tied to versions of plotly?

ianabc: Completely forgot that was an option! I've updated the branch. There is a group on there today so I don't want to change anything, I'll probably wait till Friday to merge if that is OK.

yuvipanda (approved):

Issue #2833, opened 2023-07-18
[cloudbank] Replace `username_pattern` with a per idp `allowed_domains` list


GeorgianaElena:
### Context

Using `Authenticator.username_pattern` to restrict access into a hub to specific domains, is an old configuration, remaining from when we were not using CILogonOAuthenticator or we were using an older version of `oauthenticator` that didn't supported `CILogonOAuthenticator.allowed_idps.<idp>.allowed_domains`.

### Proposal

We should replace all the occurrences of `Authenticator.username_pattern` with `CILogonOAuthenticator.allowed_idps.<idp>.allowed_domains`, like in this config for example: https://github.com/2i2c-org/infrastructure/blob/5d100582809967470fe92581950ed5af595a4076/config/clusters/catalystproject-latam/unitefa-conicet.values.yaml#L40-L46

This will also be beneficial in the context of https://github.com/2i2c-org/infrastructure/issues/2774, because starting with `oauthenticator 16.0.0`, the `allowed_idps` disct has become a mandatory field.

PR #2835, opened 2023-07-18
[cloudbank] Replace username_pattern for allowed_domains

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/2833

cc @sean-morris 

### Notes

1. For the following hubs it was not possible to remove the `username_pattern`, because these hubs need to also allow `gmail` accounts:

```
csu
pasadena
wlac
demo
skyline
norco
miracosta
srjc
```

2. All the hubs login option has been restricted to be allowed only login through Google,  Berkeley, or their specific institutional provider, where that was available and listed in https://cilogon.org/idplist
The ones with their own institutional logins were:
```
CSU Long Beach
Cal Poly Humboldt
MiraCosta College
Santa Barbara City College sbcc
```

3. The `csu` hub is a bit particular, because it wants to allow access to all unis under `California State University` (about 20). I didn't touch that one, but when we'll be upgrading to `oauthenticator 16.0.0`, where we must specify `allowed_idps` and `username_claim` for each of these idps, it would be fun to do it for 20 possible idps 😬 cc @consideRatio and the discusscussion in https://github.com/jupyterhub/oauthenticator/issues/615

### DIFF
--- a/config/clusters/cloudbank/csm.values.yaml
+++ b/config/clusters/cloudbank/csm.values.yaml
@@ -31,6 +31,7 @@ jupyterhub:
         oauth_callback_url: https://csm.cloudbank.2i2c.cloud/hub/oauth_callback
         shown_idps:
           - http://google.com/accounts/o8/id
+          - urn:mace:incommon:berkeley.edu
         allowed_idps:
           http://google.com/accounts/o8/id:
             username_derivation:
@@ -40,6 +41,9 @@ jupyterhub:
               - "berkeley.edu"
               - "my.smccd.edu"
               - "smccd.edu"
+          urn:mace:incommon:berkeley.edu:
+            username_derivation:
+              username_claim: "email"
       Authenticator:
         admin_users:
           - ericvd@berkeley.edu
--- a/config/clusters/cloudbank/csm.values.yaml
+++ b/config/clusters/cloudbank/csm.values.yaml
@@ -29,7 +29,17 @@ jupyterhub:
         authenticator_class: cilogon
       CILogonOAuthenticator:
         oauth_callback_url: https://csm.cloudbank.2i2c.cloud/hub/oauth_callback
-        username_claim: email
+        shown_idps:
+          - http://google.com/accounts/o8/id
+        allowed_idps:
+          http://google.com/accounts/o8/id:
+            username_derivation:
+              username_claim: "email"
+            allowed_domains:
+              - "2i2c.org"
+              - "berkeley.edu"
+              - "my.smccd.edu"
+              - "smccd.edu"
       Authenticator:
         admin_users:
           - ericvd@berkeley.edu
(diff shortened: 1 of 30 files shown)


config/clusters/cloudbank/csm.values.yaml
```diff
@@ -29,15 +29,28 @@ jupyterhub:
...
         oauth_callback_url: https://csm.cloudbank.2i2c.cloud/hub/oauth_callback
-        username_claim: email
+        shown_idps:
+          - http://google.com/accounts/o8/id
+          - urn:mace:incommon:berkeley.edu
+        allowed_idps:
+          http://google.com/accounts/o8/id:
+            username_derivation:
+              username_claim: "email"
+            allowed_domains:
```
sean-morris, line 39: Is the preferred method for listing allowed domains a yaml list or an array-like listing["....", "..."]?

sean-morris (approved):

sean-morris:
> ### Notes
> 1. For the following hubs it was not possible to remove the `username_pattern`, because these hubs need to also allow `gmail` accounts:
> 
> ```
> csu
> pasadena
> wlac
> demo
> skyline
> norco
> miracosta
> srjc
> ```
>
For wlac, norco, and lacc, I am working out their login pieces. They all use Microsoft as an identity provider but cilogon does not have permission to pass through. My thought after talking to @yuvipanda was to use Github Orgs to handle these cases rather than a manual upload of users if we can't get it worked out.

The csu hub could be converted and renamed to sjsu so we are not providing a general-use hub  -- it is used by one instructor at San Jose State right now.

pasadena, miracosta, and skyline can be switched over. Their login works from the institution's domain with Google or Microsoft. After this PR I can put in the change?

This leaves the demo hub and srjc. I am just beginning to work with srjc so I should have a better idea of what they need soon.


> 2. All the hubs login option has been restricted to be allowed only login through Google,  Berkeley, or their specific institutional provider, where that was available and listed in https://cilogon.org/idplist
>    The ones with their own institutional logins were:
> 
> ```
> CSU Long Beach
> Cal Poly Humboldt
> MiraCosta College
> Santa Barbara City College sbcc
> ```
> 
> 3. The `csu` hub is a bit particular, because it wants to allow access to all unis under `California State University` (about 20). I didn't touch that one, but when we'll be upgrading to `oauthenticator 16.0.0`, where we must specify `allowed_idps` and `username_claim` for each of these idps, it would be fun to do it for 20 possible idps 😬 cc @consideRatio and the discusscussion in [[CILogon] Require specification of `allowed_idps`? jupyterhub/oauthenticator#615](https://github.com/jupyterhub/oauthenticator/issues/615)

yuvipanda:
Yep, GitHub auth + orgs should be the fallback here.

Thanks a lot for cleaning this up, @GeorgianaElena!

config/clusters/cloudbank/csm.values.yaml
GeorgianaElena: Either works. I started with with the square brackets when there were around 3 domains and switched to having them one a row when they seemed to many or to long to fit on the same row.

GeorgianaElena:
Thanks for the feedback and additional context @sean-morris and @yuvipanda! I will go for a merge on this one then and leave you @sean-morris to update the remaining ones to GitHub authentication once there's info about the relevant orgs.

I think I will also open an issue to track the intention of swithching the remaining to GitHub orgs auth.

PR #2836, opened 2023-07-19
docs: Add an admin link to the hubs table

pnasrat:
More quickly access a hub without spawning a server


pnasrat:
Tested via make live in docs folder and opening a link. Added as separate column/entry in json as don't want to break airtable injestion and use of the domain field

sgibson91 (approved):
LGTM!

PR #2837, opened 2023-07-19
Adding teams for OHW23

abkfenris:
We're coming up on OceanHackWeek '23, so here are the github teams that should now get access to our hub.


yuvipanda:
Thanks, @abkfenris!

PR #2839, opened 2023-07-19
[doc]: add new table for listing each hub's options; start with authenticator info

GeorgianaElena:
First step towards https://github.com/2i2c-org/meta/issues/628

This PR:
1. refactors the `render_hubs.py` script into functions as it became quite big (copy-paste, nothing existing was changed)
2. creates a new table called 'Hub options table' (because the first one was already big)

The plan for this table is to start by adding the options identified in [this doc](https://docs.google.com/spreadsheets/d/1ZRCAMyIBOSfPlfHywEKva7WG4-xNVKm18eAWGo_4KJQ/edit?usp=sharing) and for each hub either put an `yes/no` or some value depending on each option.

For now, in this PR, only the authenticator class is listed. Note that it's not that smart to find an authenticator class inside an `extraConfig` setup, like it's the case for the callysto hub.

It looks like this:
Screenshot 2023-07-19 at 19 07 08

This PR is meant to serve as a starting point for https://github.com/2i2c-org/meta/issues/628 and can definitely be improved with docstrings, and better function names, but I think it should be good enough for a first iteration in making it scalable.

### DIFF
--- a/docs/scripts/render_hubs.py
+++ b/docs/scripts/render_hubs.py
@@ -15,70 +15,33 @@ path_tmp = path_root / "tmp"
 path_static = path_root / "_static"
 path_clusters = path_root / "../config/clusters"
 
-# Grab the latest list of clusters defined in infrastructure/ explicitly ignoring
-# the test clusters in the ./tests directory
-clusters = [
-    filepath
-    for filepath in path_clusters.glob("**/*cluster.yaml")
-    if "tests/" not in str(filepath) and "templates" not in str(filepath)
-]
 
-hub_list = []
-for cluster_info in clusters:
-    cluster_path = cluster_info.parent
-    if "schema" in cluster_info.name or "staff" in cluster_info.name:
-        continue
-    # For each cluster, grab it's YAML w/ the config for each hub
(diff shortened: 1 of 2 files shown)


docs/scripts/render_hubs.py
```diff
@@ -98,59 +61,195 @@
...
-                if cluster_console_url
-                else None,
-            }
+    return cluster_console_url
+
+
+def get_hub_authentication(cluster_path, hub_values_files):
+    # Loop through hub values files, look for authenticator_class config, and grab the info
+    authenticator_info = ""
+    for config_file in hub_values_files:
```
sgibson91, line 70: I wonder if there's a way to write this loop such that if we populate `authenticator_info` on the pass of the first file, we can skip reading in the following files?

GeorgianaElena: Yes, will do that in an upcoming commit. But I believe we will end up restructuring this function anyway really soon. This is because most of the other features will require searching these config values files and we really want to open all of them once and get all the info we want about the hub options.

sgibson91 (approved):

PR #2840, opened 2023-07-19
Add new admin users to CryoCloud hubs

sgibson91:
ref: https://2i2c.freshdesk.com/a/tickets/838


sgibson91:
Self-merging as this is only adding admin users

sgibson91:
Haha, thanks @GeorgianaElena !

PR #2843, opened 2023-07-20
Document what happens to user installed packages in rocker images

yuvipanda:
Based on question in https://2i2c.freshdesk.com/a/tickets/787

### DIFF
--- a/docs/howto/features/rocker.md
+++ b/docs/howto/features/rocker.md
@@ -138,3 +138,5 @@ which provides fast binary package installs.
 
 They are installed under `/usr/local`, and hence *cleared on server stop*. So installed
 packages will *not* persist across user sessions. 
+
+If users need to have persistent packages installed, we would need a custom image.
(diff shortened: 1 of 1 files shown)


GeorgianaElena (approved):

docs/howto/features/rocker.md
```diff
@@ -128,4 +128,13 @@ jupyterhub:
...
+```
+
+## User installed packages
+
+In R, users can install packages with `install.packages` or similar R commands. Rocker
+is configured to automatically install packages from the [Posit Package Manager](https://packagemanager.posit.co/client/#/),
+which provides fast binary package installs.
+
+They are installed under `/usr/local`, and hence *cleared on server stop*. So installed
+packages will *not* persist across user sessions.
```
GeorgianaElena, line 140: Maybe it's worth adding a line about how to make them persistent, i.e. adding them to the image.

damianavila:
> I feel like this is important info for community champions to know too - should we start thinking about where in docs.2i2c.org we can surface this as well?

cc @jmunroe

Issue #2848, opened 2023-07-21
Enable pulling from private nbgitpuller repos for Neurohackademy


yuvipanda:
I couldn't quite figure out how to get our authenticated static website working again. So instead, for neurohackademy, we should:

1. Enable private github repo pulling for nbgitpuller access
2. Make an nbgitpuller link pointing to the private content (including the GitHub link)
3. We can distribute this link to end users, and it will be only available to authenticated users.

Ref https://github.com/2i2c-org/infrastructure/issues/2681#issuecomment-1629374195

damianavila:
I think this one was completed by #2849 and #2762 and confirmed in #2681, so closing now (re-open if you disagree).

PR #2849, opened 2023-07-21
Add docs on enabling pulling from private GitHub repos

yuvipanda:
We already have this enabled for UToronto, just not documented.

Ref https://github.com/2i2c-org/infrastructure/issues/2848

### DIFF
--- a/docs/howto/features/private-nbgitpuller.md
+++ b/docs/howto/features/private-nbgitpuller.md
@@ -1,0 +1,105 @@
+# Allow nbgitpuller to pull from private GitHub repos
+
+[nbgitpuller](https://github.com/jupyterhub/nbgitpuller) is very popularly
+used to pull repos from public GitHub repositories. We can also allow users
+to pull from *private* GitHub repositories, with [git-credential-helpers](https://github.com/yuvipanda/git-credential-helpers).
+
+## Configure the image
+
+The image used in the JupyterHub must have the [git-credential-helpers](https://pypi.org/project/git-credential-helpers/)
+package installed from PyPI. Unfortunately this is currently not installed in
+any common upstream images, so this will only work with custom images.
+
+## Setup GitHub app
+
+`git-credentials-helper` uses a [GitHub App](https://docs.github.com/en/developers/apps)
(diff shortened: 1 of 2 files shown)


docs/howto/features/private-nbgitpuller.md
```diff
@@ -0,0 +1,105 @@
...
+         gitconfig:
+           mountPath: /etc/gitconfig
+           stringData: |
+             [credential "https://github.com"]
+             helper = !git-credential-github-app --app-key-file /etc/github/github-app-private-key.pem --app-id <app-id>
+          useHttpPath = true
+   ```
+   
+   Use the app-id of the GitHub app you just created. This goes in the hub's
+   `values.yaml` file. If used for a `daskhub`, next the whole thing under a
```
sgibson91, line 61:
```suggestion
   `values.yaml` file. If used for a `daskhub`, nest the whole thing under a
```

docs/howto/features/private-nbgitpuller.md
```diff
@@ -0,0 +1,105 @@
...
+   ```yaml
+   jupyterhub:
+     singleuser:
+       extraFiles:
+         gitconfig:
+           mountPath: /etc/gitconfig
+           stringData: |
+             [credential "https://github.com"]
+             helper = !git-credential-github-app --app-key-file /etc/github/github-app-private-key.pem --app-id <app-id>
+          useHttpPath = true
```
sgibson91, line 57: I'm slightly confused about the `useHttpPath` parameter. I'm guessing it is either part of `stringData` and needs further indentation like so:

```suggestion
            useHttpPath = true
```

Or it is a new key and should be like so:

```suggestion
          useHttpPath: true
```

Since you typed `=`, I'm going to assume the former first.

sgibson91 (approved):
I think this is pretty good! A couple of clarifying comments. I also only really did the helm chart values part because I had already set up a GitHub App for the static site.

docs/howto/features/private-nbgitpuller.md
yuvipanda: Yep, it had wrong indent! Fixed, thanks for catching this.

PR #2853, opened 2023-07-21
[bugfix] Fix func arg in render_hubs.py

GeorgianaElena:
follow-up to https://github.com/2i2c-org/infrastructure/pull/2839


pnasrat (approved):
LGTM

Issue #2707, opened 2023-06-23
[Request deployment] New Hub: UNITEFA (Catalyst Project LatAm)


jmunroe:
### The GitHub handle of the community representative

@nwolovick

### Hub important dates

**Target Date**: July 12, 2023 (flexible)

### Hub Authentication Type

Institutional Provider via CILogon (e.g., username@institution.org)

### First Hub Administrators

@nwolovick
@alfredoq

### [GitHub Auth only] How would you like to manage your users?

None

### [GitHub Teams Auth only] Profile restriction based on team membership

N/A

### Hub logo image URL

https://unitefa.conicet.unc.edu.ar/wp-content/blogs.dir/50/files/sites/50/2019/12/cropped-Captura-de-pantalla-de-2019-12-17-10-29-47.png

### Hub logo website URL

https://unitefa.conicet.unc.edu.ar/

### Hub user image tag and name

jupyter/scipy-notebook

### Extra features you would like to enable

- [ ] Dedicated Kubernetes cluster
- [ ] Scalable Dask Cluster

### (Optional) Preferred cloud provider

None

### (Optional) Billing and Cloud account

None

### Other relevant information to the features above

This hub is for UNITEFA-CONICET as a Latin American community as part of the Catalyst Project so should be on the same cluster as the Catalyst Project LatAm hub.

This community interaction is being managed through MetaDocencia  (Melissa Black, Nicolas Wolovick)

The hub image will eventually be a jupyter/scipy-notebook with the additional modules RDKit and pymol added.

Authentication is to provided through institutional login at [Universidad Nacional de Córdoba](https://www.unc.edu.ar/) (which is a Google workspace organization) 

### Tasks to deploy the hub

- [ ] 1. Deploy information filled in above
- [ ] 2. Engineer who will deploy the hub is assigned
- [ ] 3. If using GitHub Orgs/Teams Auth, Engineer is given Owner rights to the org to set this up.
- [ ] 4. Initial Hub deployment PR <link to PR>
- [ ] 5. Administrators able to log on -> Hub now in steady-state

nwolovick:
Pass this info to Romina Racca. Querying what are the UNC requirements for OAuth.

yuvipanda:
If the institutions are listed in https://cilogon.org/idplist/, no further action is needed.

nwolovick:
We are not there :cry: 
Try to find out if can make into this list.

nwolovick:
Romina from UNC says:

[...]

PR #2855, opened 2023-07-24
[unitefa] add additional allowed domain

GeorgianaElena:
Per https://github.com/2i2c-org/infrastructure/issues/2707#issuecomment-1646268227


GeorgianaElena:
Self-merging as it's just an extra domain

Issue #2854, opened 2023-07-24
[EVENT] Smithsonian


jmunroe:
### Summary

Teaching a course for about 20 users 7/24 - 8/5.  Having sufficient EC2 instances warmed up (8Gb and 1 cpu per user for around 20 users) would make the instructor's life a lot easier.

[Request received over email]

### Event Info

- **Community Representative:** @aewhite100
- **Event begin:** 2023-07-24
- **Event end:** 2023-08-05
- **Active times:** all day
- **Number of attendees:** 20
- [**Hub Events Calendar**](https://calendar.google.com/calendar/u/2?cid=Y19rdDg0c2g3YW5tMHNsb2NqczJzdTNqdnNvY0Bncm91cC5jYWxlbmRhci5nb29nbGUuY29t)

### Hub info

- **Hub URL**: https://smithsonian.2i2c.cloud/
- **Hub decommissioned after event?**: no

### Task List

**Before the event**

- [ ] Dates confirmed with the community representative and added to Hub Events Calendar.
- [ ] Quotas from the cloud provider are high-enough to handle expected usage.
- [ ] **One week before event** Hub is running.
- [ ] Confirm with Community Representative that their workflows function as expected.
  - <details>
    <summary>👉Template message to send to community representative</summary>
    
    ```
    Hey {{ COMMUNITY REPRESENTATIVE }}, the date of your event is getting close!
    
    Could you please confirm that your hub environment is ready-to-go, and matches your hub's infrastructure setup, by ensuring the following things:
    - [ ] Confirm that the "Event Info" above is correct
    - [ ] On your hub: log-in and authentication works as-expected
    - [ ] `nbgitpuller` links you intend to use resolve properly
    - [ ] Your notebooks and content run as-expected
    ```
    
    </details>  
- [ ] **1 day before event**, either a separate nodegroup is provisioned for the event or the cluster is scaled up.

**During and after event**

- [ ] Confirm event is finished.
- [ ] Nodegroup created for the hub is decommissioned / cluster is scaled down.
- [ ] Hub decommissioned (if needed).
- [ ] Debrief with community representative.
  - <details>
    <summary>👉Template debrief to send to community representative</summary>
      
     ```
     Hey {{ COMMUNITY REPRESENTATIVE }}, your event appears to be over 🎉
     
[...]

PR #2856, opened 2023-07-24
[smithsonian] Set minimum nodepool size to 5

sgibson91:
ref: https://github.com/2i2c-org/infrastructure/issues/2854

I have performed the scale up.

However, because the jsonnet file is written in such a way that nodepool definitions are defined in a loop, I had to scale _all_ notebook nodes to a minimum of 5. Whereas I think scaling _just_ the r5.xlarge nodepool would have sufficed. I'm not sure what the simplest way to achieve this in jsonnet would be though.

### DIFF
--- a/eksctl/smithsonian.jsonnet
+++ b/eksctl/smithsonian.jsonnet
@@ -92,7 +92,7 @@ local daskNodes = [
             // instanceTypes always have a .
             name: "nb-%s" % std.strReplace(n.instanceType, ".", "-"),
             availabilityZones: [nodeAz],
-            minSize: 0,
+            minSize: 5,
             maxSize: 500,
             instanceType: n.instanceType,
             ssh: {


yuvipanda:
The jsonnet config for each node (https://github.com/2i2c-org/infrastructure/blob/47042f827b255455dd390a005e4e36fbdbb99974/eksctl/smithsonian.jsonnet#L28) is deep merged with the default config, so you can set `minSize` here: https://github.com/2i2c-org/infrastructure/blob/47042f827b255455dd390a005e4e36fbdbb99974/eksctl/smithsonian.jsonnet#L28.

If possible, can you also find a place to document this? Sorry it isn't already documented!

yuvipanda:
https://github.com/2i2c-org/infrastructure/blob/0dfc2045ba0c7d740886519e57d8ff6281735182/eksctl/carbonplan.jsonnet#L39 is an existing example

sgibson91:
Cool, thanks @yuvipanda! Yep, I can document that too

sgibson91:
I will open a new PR for docs, so we can just revert this PR at the end of the event. This is now ready to be merged.

sgibson91:
docs PR: https://github.com/2i2c-org/infrastructure/pull/2864

PR #2857, opened 2023-07-24
Clarify NFS docs

sgibson91:
We have this hanging sentence/section in the docs on storage which I think has been left after a rewrite.

Screenshot 2023-07-24 at 15 07 50

I guess there's two scenarios here, either:

1. It needs to go completely,
2. There's missing info that needs filling in

@yuvipanda I think you worked on these docs last, any insight?


sgibson91:
Thanks @yuvipanda, now I've got a better idea I can clean that up :)

sgibson91:
I reworded the sentence a little bit to make it clearer

sgibson91:
self-merging as it's a small change

PR #2858, opened 2023-07-24
Revert "Decommission coessing"

sgibson91:
Reverts 2i2c-org/infrastructure#1802


sgibson91:
Note that in #1802 the scratch buckets weren't deleted, so they should still be present

yuvipanda (approved):
LGTM thanks!

Issue #2146, opened 2023-02-06
Setup the profile list to allow a user specified image for leap.2i2c.cloud - for all server options


consideRatio:
As requested in [this support ticket](https://2i2c.freshdesk.com/a/tickets/386), setup the ability for users of leap.2i2c.cloud to specify an image - independent of their server choice.

### Action points
- [x] Implement this
  This was not possible, at least not yet. I've opened https://github.com/jupyterhub/kubespawner/issues/699 for now.
  - [x] This is being worked in https://github.com/jupyterhub/kubespawner/pull/735
    We can help with review efforts
- [x] Followup with Julius in [the support ticket](https://2i2c.freshdesk.com/a/tickets/386)
  I've communicated that we couldn't accomplish a free form text image choice.

GeorgianaElena:
@consideRatio, this sounds like this could be achieved with the configurator? 
If the configurator plays nicely with the kubespawner profiles, then setting the image in the configurator, then choosing a  server option should be enough right?

Otherwise, I'm not sure specifying the image in an input text of some sort is possible with the current profiles implementation

consideRatio:
@GeorgianaElena I understand it as they are aware of the configurator, but still wanted to allow individual users to specify their own image for a specific server startup - also indepentent of being an admin or not.

consideRatio:
Oh no, I thought this was part of profile_options in https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html#kubespawner.KubeSpawner.profile_list.

consideRatio:
I'll close this for now.

yuvipanda:
@consideRatio can we keep this open until we can deploy the change?

yuvipanda:
Currently, manually testing works but our healthcheck fails with:


```
[E 2023-07-25 07:58:24.152 JupyterHub user:884] Unhandled error starting deployment-service-check's server: Expected option image for profile medium-full or image--unlisted-choice, not found in posted form
    Traceback (most recent call last):
      File "/usr/local/lib/python3.11/site-packages/jupyterhub/user.py", line 798, in spawn
        url = await gen.with_timeout(timedelta(seconds=spawner.start_timeout), f)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2865, opened 2023-07-25
Temporarly set user_options for test users and leap hubs

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/2146

A persistent fix requires some cleanup related to profile handling in `kubespawner` so that it's easier to see where default settings are failing or if they are inconsistent (tracked in https://github.com/jupyterhub/kubespawner/issues/758)

### DIFF
--- a/deployer/tests/test_hub_health.py
+++ b/deployer/tests/test_hub_health.py
@@ -41,6 +41,16 @@ async def check_hub_health(hub_url, test_notebook_path, service_api_token):
                 # This is because we would have lost its api token from the previous run.
                 await hub.delete_user(username)
 
+        # Temporary fix for https://github.com/2i2c-org/infrastructure/issues/2146#issuecomment-1649445203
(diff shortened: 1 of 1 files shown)


GeorgianaElena:
hmm ⬆️ modifying the test, shouldn't actually trigger the hubs to deploy. Maybe we can exclude this test file from the triggers

sgibson91:
I think we should actually just put the deployer in it's own repo

GeorgianaElena:
@sgibson91, until then, I've opened https://github.com/2i2c-org/infrastructure/pull/2866

yuvipanda (approved):
Let’s open an issue to track reverting this? Otherwise lgtm

GeorgianaElena:
Opened https://github.com/2i2c-org/infrastructure/issues/2869 to track reverting this. Thank you!

PR #2872, opened 2023-07-26
Enable pre-puller for climatematch image

yuvipanda:
Ref https://climatematch.2i2c.cloud/services/configurator/

Should make startups faster

### DIFF
--- a/config/clusters/2i2c/climatematch.values.yaml
+++ b/config/clusters/2i2c/climatematch.values.yaml
@@ -12,8 +12,8 @@ jupyterhub:
         effect: "NoSchedule"
     defaultUrl: /lab
     image:
-      name: pangeo/pangeo-notebook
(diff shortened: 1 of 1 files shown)


yuvipanda:
Gotta wait to hear back from them.

consideRatio (commented):
If they dont have a dedicated node pool, pre-pulling is likely going to slow down startup of other users starting up on a node.

Kubelet pulls images in sequence and two separate pods needing images can end up block pulling for another pods.

Due to this, i think prepulling in a shared cluster must come with a major disclaimer of needing a dedicated node pool to not slow things down for all other hubs when a user pod is scheduled ona new node.

yuvipanda:
I've added a note about this being a dedicated nodepool as well.

yuvipanda:
(self merging as an image setup!)

Issue #2884, opened 2023-07-27
Update the `imagebuilding-demo` hub use a `hub-experimental` image with the prototype kubespawner UI


GeorgianaElena:
- Deploy https://github.com/yuvipanda/prototype-kubespawner-dynamic-building-ui onto a new experimental hub image
- Enabling the prototype UI for https://imagebuilding-demo.2i2c.cloud/

Part of https://github.com/2i2c-org/infrastructure/issues/2863

GeorgianaElena:
Last sprint, I've experimented a bit with this, but couldn't get it working on our cluster. Thanks to @yuvipanda's great suggestion to first try if everything works ok locally, I've surfaced a little bug https://github.com/yuvipanda/prototype-kubespawner-dynamic-building-ui/pull/1 which I think it's why the new prototype templates were not found?

Anyway, after https://github.com/yuvipanda/prototype-kubespawner-dynamic-building-ui/pull/1 is merged, I can have a second try enabling this UI for the imagebuilding-demo hub.

Issue #2883, opened 2023-07-27
Expand the `hub-experimental` image to accomodate multiple experiments


GeorgianaElena:
### Proposal

Expand our `hub-experimental` image to support multiple different hub images so we can run multiple experiments simultaneously.

Part of https://github.com/2i2c-org/infrastructure/issues/2863

PR #2885, opened 2023-07-27
Allow multiple hub image experiments

GeorgianaElena:
Related to https://github.com/2i2c-org/infrastructure/issues/2883 and https://github.com/2i2c-org/infrastructure/issues/2884

@yuvipanda, I didn't document this before I didn't get your approval about the implementation. Does using a build argument to account for the multiple requiremets file make sense or did you have something elese in mind?

Also, I added a second image in chartpress for the second experimental hub, but we could just use the initial one and jus manually change the requirements file that it's being used.

What do you think?


yuvipanda:
OMG @GeorgianaElena this is pretty cool! So we can have just *one* `hub` image, and use bulid-args for different packages to be installed. This also avoids duplication and we can keep just one `Dockerfile` right?

So I would say we can:

1. Remove the `hub-experimental` directory completely
2. Have a `unlisted-choice-experiment` image, that is built off an `unlisted-choice-requirements.txt` file
3. Have a `dynamic-image-building` image, that is built off a `dynamic-image-building.txt` file.
4. Document this process.

How does that sound? Thanks for coming up with this, @GeorgianaElena

docs/howto/custom-jupyterhub-image.md
```diff
@@ -39,8 +39,23 @@ flowchart TD
   -- More testing for hubs that are configured differently ---
   deploy_everywhere --> end_condition[fa:fa-exclamation `hub-experimental` image is the same as `hub`]
 ```
+## How to create a new hub experimental image
 
-## How to install an unreleased version of a package in the `hub-experimental` image
+1. You will first need to create a new `.txt` file with a name relevant for the experiment, let's say `new-experiment-requirements.txt`.
```
yuvipanda, line 44: Should we also add an item here about creating the repository on quay.io?

yuvipanda (approved):
One documentation suggestion but otherwise LGTM!

Thanks for this, @GeorgianaElena!

docs/howto/custom-jupyterhub-image.md
GeorgianaElena: I noticed that the quay repos actually get created when running `chartpress`. But they get created private and you still need to go to the website to change its privacy because the org is not allowed to have any private repos. 

I think it makes sense to add the step to create a public repo on quay before building and pushing. Thank you!!!

GeorgianaElena:
Thanks for suggesting to move this setup under the main `hub` image, @yuvipanda. I think this is a great idea and I'm really happy how it turned out! 🚀

yuvipanda:
@GeorgianaElena \o/ Thanks for the nice implementation.

Issue #2879, opened 2023-07-27
[EVENT] OceanHackWeek 2023


abkfenris:
### Summary

[OceanHackWeek 2023](https://oceanhackweek.org/) is coming up August 7-11th.

### Event Info

- **Community Representative:** Alex Kerney @abkfenris 
- **Event begin:** 8/7 8 AM Pacific
  - **In your timezone:** https://arewemeetingyet.com/Los%20Angeles/2023-08-07/08:00/OHW%2023
- **Event end:** 8/12 8 PM Australia
  - **In your timezone:** https://arewemeetingyet.com/Perth/2023-08-12/20:00/OHW%2023%20-%20Australia%20edition
- **Active times:** I'm expecting the most activity 8 am - 8 pm Pacific, with another group in an Australian timezone
  - **In your timezone:** https://arewemeetingyet.com/Los%20Angeles/2023-08-07/08:00/OHW%2023
- **Number of attendees:** ~100
- [**Hub Events Calendar**](https://calendar.google.com/calendar/u/2?cid=Y19rdDg0c2g3YW5tMHNsb2NqczJzdTNqdnNvY0Bncm91cC5jYWxlbmRhci5nb29nbGUuY29t)

### Hub info

- **Hub URL**: https://oceanhackweek.2i2c.cloud
- **Hub decommisioned after event?**: Eventually, @colliand has the exact date, I've lost track of the email where we set it.
### Task List

**Before the event**

- [ ] Dates confirmed with the community representative and added to Hub Events Calendar.
- [ ] Quotas from the cloud provider are high-enough to handle expected usage.
- [ ] **One week before event** Hub is running.
- [ ] Confirm with Community Representative that their workflows function as expected.
  - <details>
    <summary>👉Template message to send to community representative</summary>
    
    ```
    Hey {{ COMMUNITY REPRESENTATIVE }}, the date of your event is getting close!
    
    Could you please confirm that your hub environment is ready-to-go, and matches your hub's infrastructure setup, by ensuring the following things:
    - [ ] Confirm that the "Event Info" above is correct
    - [ ] On your hub: log-in and authentication works as-expected
    - [ ] `nbgitpuller` links you intend to use resolve properly
    - [ ] Your notebooks and content run as-expected
    ```
    
    </details>  
- [ ] **1 day before event**, either a separate nodegroup is provisioned for the event or the cluster is scaled up.

**During and after event**

- [ ] Confirm event is finished.
[...]

PR #2907, opened 2023-08-01
Update OceanHackWeek image and access

abkfenris:
Additionally remove old user groups, update the funding and logo, and the display names for the profiles.

xref #2879


abkfenris:
In the updated Python image, I also added the github-scoped-creds package. Does that require any additional setup from the 2i2c side?

yuvipanda:
Thanks @abkfenris

PR #2908, opened 2023-08-02
Bump openscapes python image

yuvipanda:
Fixes https://2i2c.freshdesk.com/a/tickets/858


yuvipanda:
Self merging as an image bump only

PR #2910, opened 2023-08-02
Fix a few thing about the hubs tables

GeorgianaElena:
This PR:
- fixes the pagination of the tables. For some reason, without explicitly listing jquery as an extension it stopped working and I suspect it wasn't working for a while
- fix the stats table. It started showing once the write to the csv file had `index` set to true
- turns True/False into checkboxes in the options table


sgibson91 (approved):

GeorgianaElena:
Thank you @sgibson91! 🚀

PR #2924, opened 2023-08-04
Enable pre-puller in ohw hub for event

sgibson91:
ref: https://github.com/2i2c-org/infrastructure/issues/2879


consideRatio (approved):
I think this will cause the default image, and the two kubespawner profiles images to be pulled. If pre-warming nodes will be done this is a speedup for startup times, but otherwise a possible cause for a slowdown.

Issue #2911, opened 2023-08-02
OpenScapes Prometheus is unstable


jules32:
### Context

Hi! 

We're interested in using grafana to provide more granular information per user, and I am unsure of its current scope vs our use/setup. This is for our [Openscapes 2i2c JupyterHub](https://openscapes.2i2c.cloud/hub/), cc: @betolink @erinmr

I’m looking at our [openscapes grafana dashboard](https://grafana.openscapes.2i2c.cloud/d/hub-dashboard/jupyterhub-dashboard?orgId=1&from=now-60d&to=now) (screenshot below) - and the per-user charts would be really helpful for us to have. Are these blank because we use AWS credits instead of paying directly with a credit card? And if so, is there a way to get the billing per service per day even though we're using AWS credits instead of direct billing/payment?

<img width="1433" alt="image" src="https://github.com/2i2c-org/infrastructure/assets/5891909/bcc11a65-b4a8-42be-b63f-b620b20e1378">

Thanks for any insight and next steps you can provide.

yuvipanda:
Thanks for opening this, @jules32. I'm not actually sure why the graphs are broken - they should be displaying alright. we need to investigate that. What is the timeline on when this would be most useful to you?

jules32:
Thanks @yuvipanda ! We really appreciate it. This would be super helpful by early/mid September if that's possible for you?

yuvipanda:
https://github.com/2i2c-org/infrastructure/issues/2930 looks like a systemic issue!

yuvipanda:
@jules32 ok, we have done a bunch of work making this more stable. Not complete yet! But want to take another shot?

jules32:
Y'all are so awesome @yuvipanda ! I've just had a quick peek: 

The user CPU distribution is showing up! I realize I don't know what the colors mean, if each is 100% CPU the colors seem to be count of users (when I hover over the colors I see 11, 8, 2). Is that correct?

<img width="1421" alt="image" src="https://github.com/2i2c-org/infrastructure/assets/5891909/84aa3c31-8f3f-4bba-8df7-7f1535ca98e3">

It's cool on the Container images to see the legend with Python v R v Matlab :

<img width="712" alt="image" src="https://github.com/2i2c-org/infrastructure/assets/5891909/42fc1c1c-f063-4820-b147-928328d2bab9">

yuvipanda:
[...]

PR #2929, opened 2023-08-04
Increase resources for openscapes prometheus

yuvipanda:
It's not been collecting any data for a few weeks as it was in a CrashLoop. Looking
at disk space graph, it's full

image

Fixes https://github.com/2i2c-org/infrastructure/issues/2911


yuvipanda:
The problem was of course actually disk space:

image

yuvipanda:
I can't quite get the dashboard to work properly, but you can see here almost all our prometheuses have only a few mb left:

image

yuvipanda:
Superseeded by https://github.com/2i2c-org/infrastructure/pull/2932

consideRatio (approved):

Issue #2821, opened 2023-07-14
[Request deployment] New Hub -- Indian Ocean Hackweek 2023


colliand:
### The GitHub handle of the community representative

eeholmes (EDIT: fixed typo)

### Hub important dates

- **Target Start Date:** 2023-08 -01 (likely not possible until a bit later given current work load)
- **Required Start Date:** 2023--08-08
- **Important dates for usage:** 2023-09-10 through 2023-09-22. This is when the Hackathon will take place. 

### Hub Authentication Type

GitHub (e.g., @mygithubhandle)

### First Hub Administrators

@eeholmes

### [GitHub Auth only] How would you like to manage your users?

Allowing members of a specific GitHub organization (EDIT: specific team, see below)

### [GitHub Teams Auth only] Profile restriction based on team membership

EDIT: https://github.com/orgs/Hackweek-ITCOocean/teams/itcoocean-hackweek-2023

### Hub logo image URL

EDIT: https://user-images.githubusercontent.com/2545978/253672085-ec5ca6fb-147b-4fcd-87f1-431108b62558.png

Looks like:
![](https://user-images.githubusercontent.com/2545978/253672085-ec5ca6fb-147b-4fcd-87f1-431108b62558.png)

### Hub logo website URL

https://incois.gov.in/ITCOocean/index.jsp

### Hub user image GitHub repository

pending

### Hub user image tag and name

pending

### Extra features you would like to enable

- [ ] Dedicated Kubernetes cluster
- [ ] Scalable Dask Cluster

### (Optional) Preferred cloud provider

AWS

### (Optional) Preferred cloud region

us-west-2 (shared 2i2c cluster: 2i2c-aws-us)

### (Optional) Billing and Cloud account

None

### Other relevant information to the features above

Research hub on a shared cluster (likely hosted in AWS us-west-2). Additional details are being gathered through async g-doc running notes with @eeholmes. Thanks in advance to @jmunroe for assistance with this while @colliand is away on vacation. 

This hub will run from early August through mid-October.

When filling in operated by, funded by, etc fields (https://github.com/2i2c-org/infrastructure/issues/2821#issuecomment-1636388679):

>"Openscapes" should be "ITCOocean" with link to https://incois.gov.in/ITCOocean/index.jsp
>
>Funded by: Should be "ITCOocean Hackweek Team" with no url (EDIT: Should be `ESIP` instead)

### Tasks to deploy the hub

- [x] 1. Deploy information filled in above
[...]

PR #2940, opened 2023-08-07
2i2c-aws-us, itcoocean: allow users to start on 128 GB machines

consideRatio:


consideRatio:
My workday is ending and this is for a deployment of a hub with a target start date that has passed and required start date tomorrow, due to that, I'll go for a merge here to ensure the manual deployment I've done isn't reverted by our automation and causing a broken expectation that this is addressed as I reported in https://github.com/2i2c-org/infrastructure/issues/2821#issuecomment-1668180545

Issue #2906, opened 2023-08-01
AWS clusters: get `allowVolumeExpansion: true` by default in StorageClass, or document procedure to update it once


consideRatio:
We want our StorageClass resource to have `allowVolumeExpansion: true`, as that makes us able to increase the PVC's requested storage. This is a requirement for the troubleshooting docs [here for example](https://infrastructure.2i2c.org/howto/troubleshoot/prometheus/#running-out-of-disk-space).

I believe its not the default at least for AWS based clusters default StorageClass resource, but I think its the default for GCP based clusters.

yuvipanda:
A lot of our prometheuses were down, so I did this now.

1. Make a script:
   ```bash
   #!/bin/bash
   echo "kubectl patch storageclass gp2 --patch '{\"allowVolumeExpansion\": true}'" | deployer use-cluster-credentials $1
   ```
2. Run this against all the clusters. This will just error out in GKE and Azure clusters, and that's fine:
   ```bash
   ls config/clusters | xargs -L1  ./wat.bash
   ```

yuvipanda:
After that, I re-ran https://github.com/2i2c-org/infrastructure/actions/runs/5781619537 again, which completely passed this time. However, dynamic resizing only takes effect when the pod itself is recreated, so am doing that as well now.

yuvipanda:
Given we've already manually done this for existing clusters, I think for new clusters, we should use https://github.com/2i2c-org/infrastructure/pull/2941 and manage storageclass explicitly. This alleviates r2r concerns that @sgibson91 raised in https://github.com/2i2c-org/infrastructure/pull/2718 as well as allows us to use the cheaper & better gp3 storageclass.

Issue #2717, opened 2023-06-26
Make sure prometheus PVs of all our clusters have deletionPolicy set to retain


yuvipanda:
We don't really want to lose prometheus data, so we should mark all the PVs created for prometheus to have deletionPolicy set to Retain. Follow-up to https://github.com/2i2c-org/infrastructure/issues/2688

PR #2941, opened 2023-08-07
Add gp3 storageclass

yuvipanda:
We add this unconditionally to all clusters for simplification, so we can set storageClass: gp3 for new clusters that come up on AWS without issue. This doesn't change the default, and does not change the storageclass in existing clusters. In addition to using gp3, it also sets reclaimPolicy to Retain, so if the PVC is deleted, it does not delete the PV or the underlying EBS volume.

Ref https://github.com/2i2c-org/infrastructure/issues/2906 
Ref https://github.com/2i2c-org/infrastructure/issues/2717


helm-charts/support/templates/storageclass/gp3.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: storage.k8s.io/v1
```
consideRatio, line 1:
```suggestion
# By default EKS clusters doesn't provide a StorageClass for gp3 based storage, that is
# reported to be more cost efficient than gp2 based storage. Due to that we install a gp3
# StorageClass that can be explicitly referenced in AWS clusters.
apiVersion: storage.k8s.io/v1
```

helm-charts/support/templates/storageclass/gp3.yaml
```diff
@@ -0,0 +1,18 @@
...
+kind: StorageClass
+metadata:
+  name: gp3
+# Use the AWS EBS CSI provisioner
```
consideRatio, line 5:
```suggestion
```

yuvipanda:
Going to close this in favor of opening something along the lines of https://github.com/2i2c-org/infrastructure/pull/3112 later for AWS, once I get the GCP stuff done.

Issue #2953, opened 2023-08-09
Figure out what that we mount in user pods that we must `chown` in an initContainer and why


consideRatio:
We are currently using `chown` (change owner in a linux filesystem) on the mounted `/home/jovyan` and `/home/shared` folder (not on the files within, but the folder itself). I've always been a confused about the details to this, and hope that we could get to the bottom of this. If we do, we could perhaps use the knowledge to avoid redundant config and upstream improved docs in z2jh etc.

## Understanding goals

- To understand in detail when the `initContainer` doing a `chown` on the mounted directory is needed
  - Is it dependent on the NFS Server?
  - Is it dependent on the NFS mount options?
  - Is it dependent on the k8s version?
  - Is it dependent on mounting inside another NFS mount, like mounting /home/jovyan and then /home/jovyan/shared inside it?
    - If mounting for example shared inside /home/jovyan, how does an existing directory's permission influence the outcome?

## Related

- In #2946 I concluded that mounting the `shared` volume inside a mounted user's home directory, we end up creating a new directory in the mounted home directory.

## Practical goals

- Understand and document what goes on
- Cleanup our own config to not include uneccessary things
- Upstream docs improvements to z2jh

consideRatio:
I digged into this and learned quite a bit.

## Conclusions

**The concludes are based on:**

- using the test pod below in GKE v1.26.x against a GCP Filestore, 2023-08-09
- using the NFS mount options "soft" and "noatime"
- only using volumeMounts with `subPath` specified as non-root directory
- creating folders and files at mount target paths with various owners
- changing owners after a container has started from various locations

**The conclusions are:**

- When mounting a subPath that doesn't exist, its created by root
- When nesting mounts, such as mounting B in A at path a/b, then:
  - If path b didn't already exist in A, then an empty folder b owned by root is created in A.
  - If path b existeded in A, then b is hidden and untouched in the container by B.
- The owner of a directory in NFS determines the owner of the directory in the container - during container startup.
- `chown` on a subPath mounted NFS directory impacts the ownership in NFS of the directory.

So this means that new directories for a new user for example to be chowned once, or alternatively pre-created with the right owner.

## A test pod

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-nfs
  namespace: staging
spec:
  volumes:
    - name: home
      persistentVolumeClaim:
        claimName: home-nfs

  securityContext:
    fsGroup: 100

  initContainers:
    - image: busybox
[...]

Issue #2946, opened 2023-08-09
Non-Admins can see shared-readwrite directory when they shouldn't


sgibson91:
### Context

ref: https://2i2c.freshdesk.com/a/tickets/870

There seems to be a bug on the OHW hub where non-admin users can see the shared-readwrite folder which is supposed to be admin only. We know at least one non-admin who can see it cannot write to it, so if that is true across the board, damage is limited and we are just dealing with confused users. We establish "who is an admin" via some [custom Kubespawner code](https://github.com/2i2c-org/infrastructure/blob/123f691f18d334d180632c6e7841a47b3f3bf3c9/helm-charts/basehub/values.yaml#L556-L573). I think a lot of work has been going on over there recently, and I'm wondering if something could have broken/modified this functionality a bit? But I'm not sure how to go about digging into this one.

consideRatio:
I've started looking into this:

I tried starting a non-admin user's server and saw this:

![image](https://github.com/2i2c-org/infrastructure/assets/3837114/d9b06997-7bdc-44c1-887e-5dd391997b4d)

The pod created by KubeSpawner revealed:

```yaml
  # pod spec snippet
  volumes:
  - name: home
    persistentVolumeClaim:
      claimName: home-nfs

  # main container spec snippet
    volumeMounts:
    - mountPath: /home/jovyan
      name: home
      subPath: redacted-username
    - mountPath: /home/jovyan/shared
      name: home
      readOnly: true
      subPath: _shared

  # init container spec snippet
    volumeMounts:
    - mountPath: /home/jovyan
      name: home
      subPath: redacted-username
    - mountPath: /home/jovyan/shared-readwrite
      name: home
      subPath: _shared

  # init container's command
  - command:
    - sh
    - -c
    - 'id && chown 1000:1000 /home/jovyan && chown 1000:1000 /home/jovyan/shared-readwrite
      && ls -lhd /home/jovyan '
```

consideRatio:
This UX bug is seen in other clusters as well, not just the 2i2c cluster's ohw hub with GKE v1.26, such as openscapes on EKS k8s 1.24.

Running the `mount` command from a terminal as a non-admin user showed this.

## On 2i2c / ohw

```
[...]

PR #2954, opened 2023-08-09
basehub: avoid creating empty shared-readwrite directory for non-admins

consideRatio:
This is one step towards resolving #2946. It was caused by mounting the `_shared` directory to a users directory in the path `shared-readwrite` in an initContainer, as that causes the creation of an empty folder in the home directory that isn't later hidden by mounting `shared-readwrite` again on top of it.

We can avoid creating these empty `shared-readwrite` folders in the users home directory for non-admin users by not mounting that folder in the initContainer. This is what this PR does.

---

In the future we can cleanup the empty `shared-readwrite` folder for all users (non-admin is users is what matters), but if an admin becomes a non-admin at some point, they may still have an empty `shared-readwrite` folder owned by root lurking in their home directory - not because it was mounted, but once was mounted and that led to the creation of an empty root owned dir in their home directory.

For notes from a deep dive I did into this, see #2953.

### DIFF
--- a/docs/topic/infrastructure/storage-layer.md
+++ b/docs/topic/infrastructure/storage-layer.md
@@ -117,7 +117,7 @@ jupyterhub:
           [
             "sh",
             "-c",
-            "id && chown 1000:1000 /home/jovyan && chown 1000:1000 /home/jovyan/shared-readwrite && chown 1000:1000 /home/jovyan/shared-public && ls -lhd /home/jovyan ",
+            "id && chown 1000:1000 /home/jovyan && chown 1000:1000 /home/jovyan/shared && chown 1000:1000 /home/jovyan/shared-public && ls -lhd /home/jovyan ",
           ]
         securityContext:
           runAsUser: 0
@@ -128,7 +128,7 @@ jupyterhub:
           # Mounted without readonly attribute here,
           # so we can chown it appropriately
           - name: home
-            mountPath: /home/jovyan/shared-readwrite
+            mountPath: /home/jovyan/shared
             subPath: _shared
           - name: home
             mountPath: /home/jovyan/shared-public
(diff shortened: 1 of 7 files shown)


docs/topic/infrastructure/storage-layer.md
```diff
@@ -128,7 +128,7 @@ jupyterhub:
           # Mounted without readonly attribute here,
           # so we can chown it appropriately
           - name: home
-            mountPath: /home/jovyan/shared-readwrite
+            mountPath: /home/jovyan/shared
             subPath: _shared
```
consideRatio, line 132: The `mountPath` is irrelevant for the `chown` operation as the `_subpath` is influenced the same no matter what. By adjusting the `mountPath` to `shared` we avoid creating an empty directory `shared-readwrite` owned by root in the user's home directory and instead get a `shared` empty directory. But, this isn't a problem as it will be hidden by a mount later anyhow no matter if its an admin user or not.

sgibson91 (approved):
Thanks for working through this!

I think we may also need a docs update over here: https://docs.2i2c.org/data/sharing/#the-shared-directory

consideRatio:
Thanks for reviewing @sgibson91!!

> I think we may also need a docs update over here: https://docs.2i2c.org/data/sharing/#the-shared-directory

Looking at it, I think the documentation is updated overall for the user perspective. If i cleanup the empty `shared-readwrite` directories created inside users home folders, I don't think we need a caveat mentioning anything about that in the docs either.

sgibson91:
Oh, so will the admins still see a folder called `shared-readwrite`? From the removal of that part of the name, I thought it wouldn't be called `shared-readwrite` anymore when it showed up in an admin's directory, and would just be called `shared` instead.

consideRatio:
> Oh, so will the admins still see a folder called `shared-readwrite`? From the removal of that part of the name, I thought it wouldn't be called `shared-readwrite` anymore when it showed up in an admin's directory, and would just be called `shared` instead.

Yes! The change is only in the initContainer. What matters there us that the _shared NFS path no matter where its mounted to in the init container gets chowned. But, since mounting in the initContainer will cause an empty dir to be formed in the user's home directory we dont want to mount it as shared-readwrite as only the admins will hide that empty directory in the normal container by another mount.

PR #2967, opened 2023-08-10
Increase LEAP prometheus persistent storage to 500Gi

yuvipanda:
200Gi was full.

Ref https://2i2c.freshdesk.com/a/tickets/875

### DIFF
--- a/config/clusters/leap/support.values.yaml
+++ b/config/clusters/leap/support.values.yaml
@@ -16,6 +16,8 @@ prometheus:
         - secretName: prometheus-tls
           hosts:
             - prometheus.leap.2i2c.cloud
+    persistentVolume:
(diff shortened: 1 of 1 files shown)


consideRatio (approved):

consideRatio:
Nice catch!

yuvipanda:
@consideRatio I took a look at all other prometheus servers as well, and they all now have space. We still need to figure out why this storage situation exploded though.

consideRatio:
> We still need to figure out why this storage situation exploded though.

I looked into adjusting a scraping interval for certain metrics - specifically for the dirsize scraping that updates every 2 hours or anyhow. If we scraped that on an hourly basis instead it would make it fine to have finegrained details about individual users etc.

Issue #2687, opened 2023-06-21
Transition to using only a single node pool for dask-gateway workers (16 CPU, highmem)


consideRatio:
Currently, users of dask-gateway don't specify the machine type, they just create worker pods that get scheduled on some node that is being started up.

When dask-gateway is used, many many nodes can get started. Seeing for example 200 nodes isn't unusual in pangeo-hubs or leap.

I'd like us to transition from having varying amounts of node pools for dask workers in different clusters (n1-standard-4, n1-standard-8, n1-standard-16 for example), to having a single node pool of n2-highmem-16 on GCP and r5.4xlarge on AWS.

Like this:
- we can reduce the amount of nodes needed for X workers by providing on average larger nodes
  - this can reduce prometheus-server memory use blowing up because it has to scrape fewer nodes separate node-exporter pods for metrics
- we can make assumptions on the kind of cpu and memory requests that can make sense to make for users, and we help ensure that pods will fit well (its otherwise easy to request 51% of a nodes capacity etc, making only one pod fit).
  - Without helping users like this, they end up making requests like request/limit 4 CPU or similar, which would fit 3.9 pods on a 16 CPU node, which means in practice only 3. I have seen various DaskCluster's created where users end up doing these things which makes them use the resources ineffectively.
  
---

### Relevant

- This would enable dask-gateway config from https://github.com/2i2c-org/infrastructure/issues/2364 to be adopted by all daskhub clusters.

### Action points

- [ ] Transition clusters infra
- [ ] Document this policy for users

Issue #2364, opened 2023-03-16
daskhub / leap: optimize dask-gateway options for a 16 CPU / 128 GB mem worker


consideRatio:
This is a followup to #2209.

### dask-gateway client's provided options

We configure dask-gateway server to expose various options on how to create clusters to dask-gateway clients.

LEAP wants to trial us providing options like we provide for node share, match 1 / 2 / 4 / 8 / 16 CPU requests, where limits are matched as well. For dask-gateway workers, cpu and memory requests should equal limits.

1. Ensure https://github.com/2i2c-org/infrastructure/issues/2145 is resolved
2. Configure 1 / 2 / 4 / 8 / 16 CPU + associated memory requests as options, and remove the other specification of resource requests.

The JMTE has config for something like this already, it looks like this:

![dask-gateway-worker-selection](https://github.com/2i2c-org/infrastructure/assets/3837114/f5ae3892-efc1-4977-a964-5fa70ad80f26)

Issue #2923, opened 2023-08-04
Phasing out use of `n1` machines in favor of `n2` machines on GKE


consideRatio:
I've researched to conclude that `n2` machines seems like an almost pure win with:

- n2-highmem is more cost effective per RAM
- n2-highmem is more cost effective per CPU performance
- n2-highmem has 1:8 CPU:RAM ratio, while the n1 stands out with the unique ratio of 1:6.5

### Proposal

- To change anything we touch anyhow to `n2` from `n1`
  _Exception_: we need to retain `n1` nodes for GPU machines as they may only come as `n1` variants
- [x] To update templates so new clusters are bootstrapped with `n2` machines by default

### Current use of `n1` machines

- [x] 2i2c-uk core nodes
- [x] 2i2c-uk user nodes
- [ ] cloudbank core nodes
- [ ] cloudbank user nodes
- [x] cloudbank dask nodes (*use only one node pool with `n2-highmem-16`)
- [x] m2lines core nodes
- [ ] m2lines user nodes
- [x] m2lines dask nodes (*use only one node pool with `n2-highmem-16`)
- [x] meom-ige core nodes
- [ ] meom-ige user nodes
- [x] meom-ige dask nodes (*use only one node pool with `n2-highmem-16`)
- [x] pangeo-hubs core nodes
- [ ] pangeo-hubs user nodes
- [x] pangeo-hubs dask nodes (*use only one node pool with `n2-highmem-16`)
- [x] two-eye-two-see core nodes
- [x] two-eye-two-see user nodes
- [x] two-eye-two-see dask nodes (*use only one node pool with `n2-highmem-16`)
- [x] two-eye-two-see neurohackademy nodes

*In https://github.com/2i2c-org/infrastructure/issues/2687 I'm proposing we standardize on n2-highmem-16 nodes for dask workers.

### Related

The following issues relate to this smaller and more tightly scoped issue.

- #2687
- #2212
- #2121

consideRatio:
Not much changes remains, closing in favor of an issue that is cloud provider agnostic as I want to convey the same strategy for all cloud providers.

PR #2921, opened 2023-08-04
Create a dedicated nodepool for OHW event

sgibson91:
ref: https://github.com/2i2c-org/infrastructure/issues/2879

I've set the minimum to 0 for now. I can scale it up Monday AM before the start of the event.

PR #2974, opened 2023-08-11
gcp/aws, dask worker nodes: towards single r5.4xlarge/n2-highmem-16 dask worker node pool

consideRatio:
This is addressing #2687 for all clusters (AWS, GCP) except the pangeo-hubs GCP cluster. With this PR all clusters' dask-worker node pools are reduced to a single node pool of type `r5.4xlarge` (AWS) or `n2-highmem-16` (GCP). The motivation is found in #2687.

I've considered the change individuall for each community, and I think this is fine. Overall, I expect some cost reduction to follow.

This work is motivated to provide a default dask-gateway configuration that works for all our daskhub clusters, without providing individual dask-gateway config for individual clusters like we do for singleuser.profileList now. LEAP is interested in testing out such dask-gateway configuration, and this is tracked in https://github.com/2i2c-org/infrastructure/issues/2364.

### DIFF
--- a/eksctl/2i2c-aws-us.jsonnet
+++ b/eksctl/2i2c-aws-us.jsonnet
@@ -43,6 +43,10 @@ local daskNodes = [
     // *first* item in instanceDistribution.instanceTypes, to match
     // what we do with notebook nodes. Pods can request a particular
     // kind of node with a nodeSelector
+    //
+    // A not yet fully established policy is being developed about using a single
+    // node pool, see https://github.com/2i2c-org/infrastructure/issues/2687.
+    //
     { instancesDistribution+: { instanceTypes: ["r5.4xlarge"] }},
 ];
 
--- a/eksctl/carbonplan.jsonnet
+++ b/eksctl/carbonplan.jsonnet
@@ -44,13 +44,17 @@ local notebookNodes = [
     },
 ];
 
-// Node definitions for dask worker nodes. Config here is merged
-// with our dask worker node definition, which uses spot instances.
-// A `node.kubernetes.io/instance-type label is set to the name of the
-// *first* item in instanceDistribution.instanceTypes, to match
-// what we do with notebook nodes. Pods can request a particular
-// kind of node with a nodeSelector
 local daskNodes = [
+    // Node definitions for dask worker nodes. Config here is merged
+    // with our dask worker node definition, which uses spot instances.
+    // A `node.kubernetes.io/instance-type label is set to the name of the
+    // *first* item in instanceDistribution.instanceTypes, to match
+    // what we do with notebook nodes. Pods can request a particular
+    // kind of node with a nodeSelector
+    //
+    // A not yet fully established policy is being developed about using a single
+    // node pool, see https://github.com/2i2c-org/infrastructure/issues/2687.
+    //
     { instancesDistribution+: { instanceTypes: ["r5.4xlarge"] }},
 ];
 
--- a/eksctl/carbonplan.jsonnet
+++ b/eksctl/carbonplan.jsonnet
@@ -51,10 +51,7 @@ local notebookNodes = [
 // what we do with notebook nodes. Pods can request a particular
 // kind of node with a nodeSelector
 local daskNodes = [
-    { instancesDistribution+: { instanceTypes: ["r5.large"] }},
-    { instancesDistribution+: { instanceTypes: ["r5.xlarge"] }},
-    { instancesDistribution+: { instanceTypes: ["r5.2xlarge"] }},
-    { instancesDistribution+: { instanceTypes: ["r5.8xlarge"] }},
+    { instancesDistribution+: { instanceTypes: ["r5.4xlarge"] }},
 ];
 
 {
--- a/eksctl/gridsst.jsonnet
+++ b/eksctl/gridsst.jsonnet
@@ -44,6 +44,10 @@ local daskNodes = [
     // *first* item in instanceDistribution.instanceTypes, to match
     // what we do with notebook nodes. Pods can request a particular
     // kind of node with a nodeSelector
+    //
+    // A not yet fully established policy is being developed about using a single
+    // node pool, see https://github.com/2i2c-org/infrastructure/issues/2687.
+    //
     { instancesDistribution+: { instanceTypes: ["r5.4xlarge"] }},
 ];
 
--- a/eksctl/jupyter-meets-the-earth.jsonnet
+++ b/eksctl/jupyter-meets-the-earth.jsonnet
@@ -64,6 +64,10 @@ local daskNodes = [
     // *first* item in instanceDistribution.instanceTypes, to match
     // what we do with notebook nodes. Pods can request a particular
     // kind of node with a nodeSelector
+    //
+    // A not yet fully established policy is being developed about using a single
+    // node pool, see https://github.com/2i2c-org/infrastructure/issues/2687.
+    //
     { instancesDistribution+: { instanceTypes: ["r5.4xlarge"] }},
 ];
 
--- a/eksctl/jupyter-meets-the-earth.jsonnet
+++ b/eksctl/jupyter-meets-the-earth.jsonnet
@@ -64,9 +64,7 @@ local daskNodes = [
     // *first* item in instanceDistribution.instanceTypes, to match
     // what we do with notebook nodes. Pods can request a particular
     // kind of node with a nodeSelector
-    { instancesDistribution+: { instanceTypes: ["m5.large"] }},
-    { instancesDistribution+: { instanceTypes: ["m5.4xlarge"] }},
-    { instancesDistribution+: { instanceTypes: ["m5.16xlarge"] }},
+    { instancesDistribution+: { instanceTypes: ["r5.4xlarge"] }},
 ];
 
 
--- a/eksctl/nasa-cryo.jsonnet
+++ b/eksctl/nasa-cryo.jsonnet
@@ -37,6 +37,10 @@ local daskNodes = [
     // *first* item in instanceDistribution.instanceTypes, to match
     // what we do with notebook nodes. Pods can request a particular
     // kind of node with a nodeSelector
+    //
+    // A not yet fully established policy is being developed about using a single
+    // node pool, see https://github.com/2i2c-org/infrastructure/issues/2687.
+    //
     { instancesDistribution+: { instanceTypes: ["r5.4xlarge"] }},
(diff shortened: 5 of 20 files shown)


GeorgianaElena (changes requested):
Thanks for pushing forward towards standardizing this @consideRatio! 🚀 

### Questions
I have a few questions about some things that I don't fully understand before I can give a meaningful feedback.

1. In #2687 it's the following statement
   > Currently, users of dask-gateway don't specify the machine type, they just create worker pods that get scheduled on some node that is being started up.
   
   Does this mean that they don't specify the machine type because they _cannot_  or they don't it because of lack of relevant documentation or lack of information about the type of node/s that are being spawned?
   
2. Do we know for sure that the node pools that we created for dask workers were 2i2c's choice/recommendation or were these explicitly requested by the communities?

### Opinion so far, given the missing pieces

My feeling right now, is that I believe it is a good idea to have a single node pool for dask workers that we default to for _new_ hubs, unless otherwise requested by the community.

For existing hubs, I believe we should double-check that they didn't explicitly requested the nodepools they have or if they've communicated us some special use-case that might have motivated the current choices.

### Request for docs

Nonetheless, I believe we should document this new default policy that we want to set and enforce on the existing clusters.

Ideally we would have an engineer facing documentation and one community facing that we would use to communicate to the communities' we're transitioning, but also new communities that wish to have dask.

GeorgianaElena:
⬆️ this was meant as a rewiew comment not a request for changes. Sorry about getting it wrong

consideRatio:
Thanks for reviewing this @GeorgianaElena!!

---

> 1. In [Transition to using only a single node pool for dask-gateway workers (16 CPU, highmem) #2687](https://github.com/2i2c-org/infrastructure/issues/2687) it's the following statement
>    > Currently, users of dask-gateway don't specify the machine type, they just create worker pods that get scheduled on some node that is being started up.
>
>    Does this mean that they don't specify the machine type because they _cannot_  or they don't it because of lack of relevant documentation or lack of information about the type of node/s that are being spawned?

They cannot, they typically lack information about available machine types, and we can't say what they end up using unless its specified or only a single node pool is used.

Currently the dask-gateway chart configuration enables dask-gateway clients to configure this:

https://github.com/2i2c-org/infrastructure/blob/19c612f46b4352d9bffd41efb46d273bc49871b1/helm-charts/daskhub/values.yaml#L193C1-L210

This is the config that influences how it looks in this situation - here from JMTE with another config than we provide by default:

image

Note that there is no `nodeSelector` or similar related to machine type. Like this, a user that wanted to control what machine type would be used among wouldn't be able to. I'm quite confident we provided machine types without input from communities in general, and even if we did, they ended up unable to control it. What machine is started by the k8s cluster-autoscaler isn't in any way obvious if a one or more pods can fit on a node in multiple node pools.

---

> 2. Do we know for sure that the node pools that we created for dask workers were 2i2c's choice/recommendation or were these explicitly requested by the communities?

No, at least not me. But even if they had requested something, they didn't control what they got unless the requested a single machine type given that they didn't influence the `nodeSelector` of the pod.

---

> ### Opinion so far, given the missing pieces
> My feeling right now, is that I believe it is a good idea to have a single node pool for dask workers that we default to for _new_ hubs, unless otherwise requested by the community.
> 
> For existing hubs, I believe we should double-check that they didn't explicitly requested the nodepools they have or if they've communicated us some special use-case that might have motivated the current choices.

Among the cluster changed in this PR, I think only `meom-ige` should be double checked because its the only cluster where a user creating a very large worker could in theory notice something different - they can't get their worker to start if they request for example 32 CPU.

- [x] Check git history of meom-ige about the initial setup of dask-worker node pools.

  The reason for this change related to meom-ige [mirroring the existing user node pools to be dask worker node pools](https://github.com/2i2c-org/infrastructure/pull/850/files#diff-ee87c6f05b36503acf0a933f2034fd906d65e5bb0bec27abad510e719180df05
) it seems. I consider this check to be enough, another followup could be to ask a community representative and wait for their response.

---

> ### Request for docs
> Nonetheless, I believe we should document this new default policy that we want to set and enforce on the existing clusters.
> 
> Ideally we would have an engineer facing documentation and one community facing that we would use to communicate to the communities' we're transitioning, but also new communities that wish to have dask.

I'm in full agreement that we should document a policy that is established, but I'd like to wait before its established and let it be represented as #2687 only at this point in time.

I distinguish three kinds of docs:
- **intent on a policy**
  I think this is captured by #2687 for engineers well enough as it is
- **user facing docs about policy**
  I think we're not ready to declare something, because we have common no policy yet.
- **engineering facing docs about policy**
  I'd like to do it once we've ended up in a stable state, and lean on work in #2687 to complete and possibly #2364 to be piloted to some extent as its very connected.

GeorgianaElena:
> Note that there is no nodeSelector or similar related to machine type. Like this, a user that wanted to control what machine type would be used among wouldn't be able to. I'm quite confident we provided machine types without input from communities in general, and even if we did, they ended up unable to control it. What machine is started by the k8s cluster-autoscaler isn't in any way obvious if a one or more pods can fit on a node in multiple node pools.

Thanks for the extra context @consideRatio. This makes more sense now and it helps me understand the motivation better.

> The reason for this change related to meom-ige [mirroring the existing user node pools to be dask worker node pools](https://github.com/2i2c-org/infrastructure/pull/850/files#diff-ee87c6f05b36503acf0a933f2034fd906d65e5bb0bec27abad510e719180df05) it seems. I consider this check to be enough, another followup could be to ask a community representative and wait for their response.

Agree this is enough! Thanks for going extra and checked this.

### Docs
On the docs side, I agree with the intent of the policy and the user facing policy docs points. 

For the engineering facing docs, since we're already migrating existing clusters towards the single, one type of nodepool and we wish to have future hubs following the same pattern, this already sounds like a policy, even if it's just experimental and subject to change.

I believe we should document this, even if it has a warning message about it being still experimental and not matured completely. In my opinion, the engineer docs should have all the info needed for an engineer to deploy a new hub at any given time. When deploying a new hub, I always follow the docs, and rarely search the issues for additional guidelines.

See for example https://github.com/2i2c-org/infrastructure/pull/2921 in relationship with https://github.com/2i2c-org/infrastructure/issues/2923. While the source of truth, i.e. the docs, is not updated, we will end up in situations where engineers are not aware of current (experimental) policy.

Does this make sense @consideRatio ?

consideRatio:
New hubs are already getting a single node pool of the right type by default via the templates, already updated.

I'm not confident on what I'd document at this point in time related to new hub setup docs.

Do you have something specific in mind you'd like to see in engineering docs at this time about machine types?

sgibson91:
Just a reminder that our practices are to run `tf plan` for PRs, and run `tf apply` on approval https://infrastructure.2i2c.org/contributing/code-review/#changes-to-running-infrastructure I just had to contend with the following unexpected changes when applying #2983 I'm going to leave the changes made in this PR in place for now

```
  # google_container_node_pool.dask_worker["worker"] must be replaced
-/+ resource "google_container_node_pool" "dask_worker" {
      ~ id                          = "projects/two-eye-two-see/locations/us-central1-b/clusters/pilot-hubs-cluster/nodePools/dask-worker" -> (known after apply)
      ~ instance_group_urls         = [
          - "https://www.googleapis.com/compute/v1/projects/two-eye-two-see/zones/us-central1-b/instanceGroupManagers/gke-pilot-hubs-cluster-dask-worker-ff0d09f6-grp",
        ] -> (known after apply)
      ~ managed_instance_group_urls = [
          - "https://www.googleapis.com/compute/v1/projects/two-eye-two-see/zones/us-central1-b/instanceGroups/gke-pilot-hubs-cluster-dask-worker-ff0d09f6-grp",
        ] -> (known after apply)
      + max_pods_per_node           = (known after apply)
        name                        = "dask-worker"
      + name_prefix                 = (known after apply)
      ~ node_count                  = 0 -> (known after apply)
      ~ node_locations              = [
          - "us-central1-b",
        ] -> (known after apply)
      + operation                   = (known after apply)
      ~ version                     = "1.26.5-gke.1400" -> (known after apply)
        # (4 unchanged attributes hidden)

      ~ autoscaling {
          ~ location_policy      = "ANY" -> (known after apply)
          - total_max_node_count = 0 -> null
          - total_min_node_count = 0 -> null
            # (2 unchanged attributes hidden)
        }

      ~ node_config {
          ~ disk_size_gb      = 100 -> (known after apply)
          ~ guest_accelerator = [] -> (known after apply)
          ~ image_type        = "COS_CONTAINERD" -> (known after apply)
          ~ local_ssd_count   = 0 -> (known after apply)
          ~ machine_type      = "n2-highmem-16" -> "n1-highmem-4" # forces replacement
          ~ metadata          = {
              - "disable-legacy-endpoints" = "true"
            } -> (known after apply)
          + min_cpu_platform  = (known after apply)
          - resource_labels   = {} -> null
            tags              = []
            # (8 unchanged attributes hidden)

          - shielded_instance_config {
              - enable_integrity_monitoring = true -> null
              - enable_secure_boot          = false -> null
            }

            # (1 unchanged block hidden)
        }

      - upgrade_settings {
          - max_surge       = 1 -> null
          - max_unavailable = 0 -> null
          - strategy        = "SURGE" -> null
        }

        # (1 unchanged block hidden)
```

consideRatio:
I know @sgibson91 I'm sorry about this =/ I opportunisticly applied this during a low-activity moment.

> Then maybe just add the link to https://github.com/2i2c-org/infrastructure/issues/2687 in the template files as a comment?

Done!

sgibson91:
@consideRatio No worries, I sorted it out :)

yuvipanda:
I've overwritten this change as I needed to bump up home directory capacity for pilot hubs in https://github.com/2i2c-org/infrastructure/pull/2993

consideRatio:
@GeorgianaElena @yuvipanda if you think this change is acceptable, I'd like to go for it again for any cluster I can without disrupting users.

GeorgianaElena:
@consideRatio, yes, I'm 👍🏼 on continuing this work! Thank you for working on it <3

consideRatio:
I now also transitioned three AWS clusters dask worker node pools (carbonplan, openscapes, jupyter-meets-the-earth) via `eksctl` changes.

I won't be able to transition the last cluster pangeo-hubs on GCP as I lack access to do that, making me opt to not try get that done as part of this PR.

At this point, I'd like to merge this - @GeorgianaElena can you do a final review?

consideRatio:
Thank you @GeorgianaElena!!

PR #2979, opened 2023-08-14
Deploy hub for Michigcan Technological University

sgibson91:
ref: https://github.com/2i2c-org/infrastructure/issues/2878

### DIFF
--- a/config/clusters/2i2c/mtu.values.yaml
+++ b/config/clusters/2i2c/mtu.values.yaml
@@ -49,14 +49,9 @@ jupyterhub:
               username_claim: "email"
             allowed_domains:
               - "2i2c.org"
+          # Allow MTU to login via Shibboleth
+          https://sso.mtu.edu/idp/shibboleth:
+            username_derivation:
+              username_claim: "email"
+            allowed_domains:
               - "mtu.edu"
-          # Currently experiencing issues with CAS authorisation
-          # ref: https://2i2c.freshdesk.com/a/tickets/886
-          # # Allow MTU to login via Shibboleth
-          # https://sso.mtu.edu/idp/shibboleth:
-          #   username_derivation:
-          #     username_claim: "email"
-          #     action: "strip_idp_domain"
-          #     domain: "mtu.edu"
-          #   allowed_domains:
-          #     - "mtu.edu"
--- a/config/clusters/2i2c/mtu.values.yaml
+++ b/config/clusters/2i2c/mtu.values.yaml
@@ -33,8 +33,8 @@ jupyterhub:
     config:
       Authenticator:
         admin_users:
-          - dbkc
-          - lebrown
+          - "dbkc@mtu.edu"
+          - "lebrown@mtu.edu"
       JupyterHub:
         authenticator_class: cilogon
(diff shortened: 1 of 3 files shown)


sgibson91:
@GeorgianaElena I'd appreciate your eye on the CILogon Authenticator config please. I feel it always catches me out 😅

config/clusters/2i2c/mtu.values.yaml
```diff
@@ -0,0 +1,59 @@
...
+            username_derivation:
+              username_claim: "email"
+            allowed_domains:
+              - "2i2c.org"
+          # Allow MTU to login via Shibboleth
+          https://sso.mtu.edu/idp/shibboleth:
+            username_derivation:
+              username_claim: "email"
+              action: "strip_idp_domain"
+              domain: "mtu.edu"
```
GeorgianaElena, line 57: Was the intention here for the hub usernames to be stripped of the domains?
With this, instead of  `user@mtu.ed`u, the hub users will be just be `user`.

Also, since the rule for the google login doesn't have the stripping options, those ones will be left as emails.

GeorgianaElena (commented):
> @GeorgianaElena I'd appreciate your eye on the CILogon Authenticator config please. I feel it always catches me out 😅

The config looks good to me @sgibson91. I left a small question to make sure a certain  setting was intended, but otherwise is looking good.

I'm sorry that the CILogon config feels surprizing sometimes. Do you have any ideas on how we could improve the docs to mitigate this?

config/clusters/2i2c/mtu.values.yaml
sgibson91: > Also, since the rule for the google login doesn't have the stripping options, those ones will be left as emails.

I think this has to be true for the `add_staff_user_ids_of_type` to work, otherwise it's comparing `sgibson` to `sgibson@2i2c.org` and not giving me admin rights. (I figured this out because I originally followed the example exactly with the 2i2c prefix.)

sgibson91: I guess I find our CILogon docs confusing and more explanatory of what we _could_ do, not what we _should_ do.

sgibson91: It also seems to not be working, so maybe I shouldn't be stripping the domains? https://github.com/2i2c-org/infrastructure/issues/2878#issuecomment-1677644549

sgibson91:
> I'm sorry that the CILogon config feels surprizing sometimes. Do you have any ideas on how we could improve the docs to mitigate this?

Sorry, I missed this before. I expect docs under the Hub Deployment Guide to be our standard practices and should be very copy-pasteable in order to get a hub up and running in as short a time as possible. I feel like I'm currently being presented with too many options and I often don't know which the most appropriate one to pick is. Perhaps we could standardise the options here and move some of the more advanced config, such as IdP stripping and prefixing, to the HowTo section with a cross-ref to the pages in the hub deployment guide?

GeorgianaElena (approved):
LGMT! I've also opened https://github.com/2i2c-org/infrastructure/issues/2988 about the cilogon docs improvement.

PR #2987, opened 2023-08-17
Fix typo in nasa-ghg sign-in link

sgibson91:
Related to confusion in #2951


sgibson91:
I will self-merge this since it is not actually a config change

Issue #2992, opened 2023-08-17
Shared directories using up space?


sgibson91:
### Context

In https://2i2c.freshdesk.com/a/tickets/890, it is reported that Neurohackademy are running out of space in their shared directories space. However, I cannot find anywhere in config or docs that describes how much space we set for these?

yuvipanda:
This is same as space for all home directories, and it looks pretty low:

```
$  deployer exec-homes-shell 2i2c neurohackademy    
$ df -h
10.234.45.250:/homes/homes/neurohackademy  3.5T  3.3T   43G  99% /home
```

yuvipanda:
I bumped up the capacity to 4T (https://github.com/2i2c-org/infrastructure/pull/2993) as otherwise I think this disk would've gotten full very soon.

PR #2993, opened 2023-08-17
Bump 2i2c shared cluster home directory storage

yuvipanda:
We were at 99% full.

Ref https://github.com/2i2c-org/infrastructure/issues/2992


yuvipanda:
I applied this already because if the disk fills up all our hubs on shared cluster stop working.

GeorgianaElena (approved):

sgibson91:
I'm going to merge this so it doesn't get overwritten

PR #2994, opened 2023-08-17
Document that shared dirs are on same space as user homedirs

yuvipanda:
fixes https://github.com/2i2c-org/infrastructure/issues/2992

[edited by sgibson91 to close the linked issue]

### DIFF
--- a/docs/topic/infrastructure/storage-layer.md
+++ b/docs/topic/infrastructure/storage-layer.md
@@ -53,6 +53,10 @@ Note that `<username>` is specified as a `subPath` - the *subdirectory* **in the
 
 #### Shared directories
 
+Shared directories use the same storage as users' home directories, and are
+not size limited in any other way. If more space is required, resize the
+home directory storage up.
+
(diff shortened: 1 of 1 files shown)


docs/topic/infrastructure/storage-layer.md
```diff
@@ -53,6 +53,10 @@ Note that `<username>` is specified as a `subPath` - the *subdirectory* **in the
 
 #### Shared directories
 
+Shared directories use the same storage as users' home directories, and are
+not size limited in any other way. If more space is required, resize the
+home directory storage up.
```
GeorgianaElena, line 58: Maybe also link to the docs describing how to increase it?

```suggestion
not size limited in any other way. If more space is required, [resize the
home directory storage up](https://infrastructure.2i2c.org/sre-guide/common-problems-solutions/#no-space-left-on-device-error).
```

GeorgianaElena (approved):

sgibson91:
[Non-blocking comment] Are there other ways to diagnose this? We hadn't actually reached the "no space left on device error" message state. I also tried looking at (central) grafana but everything seems to be no data over there.

yuvipanda:
@sgibson91 I'll work on figuring out how we can keep track of this!

consideRatio (approved):
The hardcoded https link could be made to a sphinx reference to not break as easily, but this is an incremental improvement as it is

PR #3008, opened 2023-08-21
basehub: faster hub restarts by letting template-sync react to SIGTERM

### LINKED ISSUE #2129: Ensure our hub pod's templates-sync containers shut down quickly on SIGTERM

consideRatio:
This change makes the hub pod restart ~29s faster consistently, and that means JupyterHub is unresponsive 29 seconds less during upgrades.

- Closes #2129

## Before
hub-restart-old

## After
hub-restart-new

```
Starting template sync...

Info about local git repo to be synced:
+ git remote -v
origin	https://github.com/2i2c-org/default-hub-homepage (fetch)
origin	https://github.com/2i2c-org/default-hub-homepage (push)
+ ls -lhd /srv/repo
drwxrwsrwx    5 1000     1000         124 Aug 22 07:44 /srv/repo

Syncing local git repo /srv/repo against origin's branch master every 5 minutes...
HEAD is now at 9472884 Merge pull request #22 from GeorgianaElena/update-instructions
HEAD is now at 9472884 Merge pull request #22 from GeorgianaElena/update-instructions
```

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -474,7 +474,7 @@ jupyterhub:
 
                 # signal handling can only be done between sleep calls, so this
                 # shouldn't be reduced to the otherwise equivalent "sleep 5m"
-                for i in {1..300}; do
+                for i in $(seq 300); do
                     sleep 1s;
                 done
             done
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -445,10 +445,33 @@ jupyterhub:
         args:
           - -c
           - |
-            ls -lhd /srv/repo;
-            while true; do git fetch origin;
-            git reset --hard origin/$(GIT_REPO_BRANCH);
-            sleep 5m; done
+            handle_sigterm() {
+                echo "SIGTERM received, terminating...";
+                exit;
+            }
+            trap handle_sigterm SIGTERM
+
+            echo "Starting template sync...";
+
+            # print debugging info about commands run and their output using a
+            # temporary subshell with "set -x"
+            (
+                set -x;
+                git remote -v;
+                ls -lhd /srv/repo;
+            )
+
+            echo "Syncing local git repo /srv/repo's against remote origin's branch $GIT_REPO_BRANCH";
+            while true; do
+                git fetch origin;
+                git reset --hard origin/$(GIT_REPO_BRANCH);
+
+                # signal handling can only be done between sleep calls, so this
+                # shouldn't be reduced to the otherwise equivalent "sleep 5m"
+                for i in {1..300}; do
+                    sleep 1s;
+                done
+            done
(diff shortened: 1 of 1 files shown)


helm-charts/basehub/values.yaml
```diff
@@ -445,10 +445,33 @@ jupyterhub:
...
           - |
-            ls -lhd /srv/repo;
-            while true; do git fetch origin;
-            git reset --hard origin/$(GIT_REPO_BRANCH);
-            sleep 5m; done
+            handle_sigterm() {
+                echo "SIGTERM received, terminating...";
+                exit;
+            }
+            trap handle_sigterm SIGTERM
```
consideRatio, line 452: Adding this is the key thing for this PR, but it also had to be combined with breaking apart sleep 5m into 500 sleep 1s. The rest of the changes is just about providing some misc info of relevance.

helm-charts/basehub/values.yaml
```diff
@@ -445,10 +445,33 @@ jupyterhub:
...
+
+            echo "Starting template sync...";
+
+            # print debugging info about commands run and their output using a
+            # temporary subshell with "set -x"
```
yuvipanda, line 457: Can you add a comment on *why* this is done rather than just using `set -x` globally for the whole script?

yuvipanda (approved):
This is pretty awesome, thanks for taking care of this, @consideRatio! I left a comment about adding a comment on why the `set -x` is in a subshell (probably to avoid printing all the sleeps), but once you add that comment feel free to merge.

consideRatio:
Thanks for reviewing @yuvipanda!!!!!

PR #3014, opened 2023-08-22
Decom hackanexoplanet

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/3013


yuvipanda:
Confirmed in https://github.com/2i2c-org/infrastructure/issues/3013#issuecomment-1699239527

PR #3031, opened 2023-08-25
Simplify OpenScapes profile options

yuvipanda:
Based on discussing how profiles were actually being used with the openscapes folks
(https://github.com/2i2c-org/infrastructure/issues/2882)

image

Generated via the scripts in
https://github.com/2i2c-org/infrastructure/pull/3030. See the description there for
various options available here, the principles behind it and the user focused nature of
this change. Reading that PR description is highly recommended before proceeding with
this PR.

This is a separate PR than #3030 as this can be deployed without having to wait
for that to go through.

## Why was r5.xlarge and r5.4xlarge chosen?

r5.xlarge has 32GB of RAM, and I think it's the appropriate node size for *most* of the openscapes
workload.

Their current utilization on larger nodes is pretty small. Unused utilization in this graph is stuff we are paying for, but not using.

image

We also see mostly there are under 8-10 users, and even with just an r5.xlarge, they should be able to fit most users.

image

There was a couple uses of 100+ GB memory uses a few weeks ago. I think in general it's acceptable
for users who want a *lot* of resources to have to wait for a bit for nodes to spin up. This is why r5.4xlarge was added, with two RAM options.

image

The community is generally ok with us deploying this next week!

image

In addition, because memory_limit was used, it actually shows up clearly in the JupyterLab status bar!

image

This was the most requested feature during the conversation.

## What to measure after this?

The primary thing to measure here quantitatively is to see if utilization % goes up, which brings cost down. Secondarily we should look at server startup time.

Qualitatively, we can check back in with the openscapes folks again to see how their instructors and users feel about this.

We can trade-off cloud cost and startup speed by switching the base from r5.xlarge to perhaps r5.2xlarge. But we can check back in in 2 weeks to see.


yuvipanda:
This is being tweaked with feedback from the openscapes community via slack https://openscapes.slack.com/archives/C02NC3Y62J1/p1692913592863859. Please let me know if you need an invite!

yuvipanda:
The following things were checked with the community:

1. The somewhat fractional memory allocations. These are all *accurate*, and will reflect the amount of RAM they'll have access to. This is what is shown in JupyterLab too. The trade-off is that these numbers look a little odd, but it's the same number throughout, and easily understandable for the user what it means. The community *approves*
2. These memory allocations are enforced at all times, regardless of what other users are doing. If someone wants more RAM, they are expected to ask for it. This is a feature of this specific resource allocation strategy. The community *approves*.
3. The word "Resource Allocation" is more familiar and understandable than the previous "node share". Some of this is due to familiarity with this from older HPC systems. "Node Share" is also an implementation detail, and one of the reasons we are doing this is that end users should be able to stop thinking about 'size of nodes'. The community *approves* this wording too
4. It's unclear if the large 100+ GB allocation will be used at all, and perhaps the person who used it in the past did so by mistake. We're going to *revaluate* this in a few weeks. We need to make sure we can *track what profiles people are selecting*.

I'll await more feedback from the community, but the hope is we get this deployed early next week.

consideRatio (approved):
:+1: for piloting the use of this without much further thinking!

I'm a bit worried that there isn't ahead margins now, right? What is bound to happen I think is that we will need to tweak resource requests of system pods, and suddenly you can't get a guaranteed request of memory or cpu to fit one user to get 100% of a node or two users to get 50% of a node etc.

That concern is mostly a blocker for wider adoption, not adoption in a hub piloting this.

config/clusters/openscapes/common.values.yaml
```diff
@@ -34,207 +34,98 @@ basehub:
...
-                mem_4:
+                mem_1_9:
+                  display_name: 1.9 GB RAM, upto 3.75 CPUs
+                  kubespawner_override:
+                    mem_guarantee: 1992701952
+                    mem_limit: 1992701952
+                    cpu_guarantee: 0.234375
+                    cpu_limit: 3.75
+                    node_selector:
+                      node.kubernetes.io/instance-type: r5.xlarge
```
consideRatio, line 63: This makes sense to have in a kubespawner_override on a higher level in a way, but at the same time, this is generated, so its acceptable to have it explicit here.

yuvipanda:
@consideRatio thanks for the review!

Yes, we would need to adjust this based on system pods resource requests. It's handled as part of https://github.com/2i2c-org/infrastructure/pull/3030. I do agree that adding an additional 'buffer' there might help as well! Let's figure that out while reviewing that one.

config/clusters/openscapes/common.values.yaml
yuvipanda: Yes, the primary change here is to have the engineers and community champions make the choice of node size. I agree that during workshops it makes sense to pick a larger node size. Let's address that as part of https://github.com/2i2c-org/infrastructure/pull/3030 as well. 

This `node_selector` is in fact superfluous, and I've removed it.

yuvipanda: > I see now that the machine choice switches when a request requires 100% of a machine. So for example when ~32GB is required, should the user really run on a 32GB machine or should the user arrive on a 16 CPU 128GB machine instead perhaps?

For this particular resource allocation strategy (more about that in #3030), one assumption is that if you are requesting more resources, it's ok for your startup to take longer. Where should that cut-off be? That depends on the community! This should feed into selecting the kinds of nodes made available to users and be part of that guidance. Needs automation and documentation for sure.

yuvipanda:
Ok, I heard back from the community and they're ok with me deploying it today. Thanks for the prompt review making this possible, @consideRatio!

image

colliand:
> This is being tweaked with feedback from the openscapes community via slack https://openscapes.slack.com/archives/C02NC3Y62J1/p1692913592863859. Please let me know if you need an invite!

Yes please. @colliand would like an invitation to the OpenScapes Slack.

Issue #3029, opened 2023-08-24
Do not use the word 'NASA' in our Domains


yuvipanda:
Per @freitagb, any domains that have the word 'nasa' in them but do not end in .gov are blocked on most NASA networks and machines as an anti phishing feature. This means they aren't able to access some of our clusters.

Let's just avoid using the word 'nasa' in domains.

yuvipanda:
@slesaad heya! Can you change the CNAME for `hub.center.ghg` from `nasa-ghg.2i2c.cloud` to just `ghg.2i2c.cloud`? It shouldn't affect end users but lets us get rid of `nasa-ghg` from any domains.

slesaad:
@yuvipanda done!

yuvipanda:
@slesaad hmm, `dig hub.ghg.center` still points me to `nasa-ghg.2i2c.cloud`.

slesaad:
Sorry, my bad, should be updated now!
![image](https://github.com/2i2c-org/infrastructure/assets/7830949/e8431ab5-c223-4143-a167-75594babf7fa)

yuvipanda:
@slesaad perfect! I see it now. I'll get rid of the `nasa-ghg` target in a few days, to allow any lagging DNS servers to catch up.

yuvipanda:
@freitagb https://github.com/2i2c-org/infrastructure/pull/3037 changes nasa-veda URL from `nasa-veda.2i2c.cloud` to `veda.2i2c.cloud`, and keeps a redirect in place, to prevent existing links from breaking. That sound good?

PR #3033, opened 2023-08-25
Fix domain of GHG staging hub

yuvipanda:
Also a trivial change in common to retrigger deployment

Ref https://github.com/2i2c-org/infrastructure/issues/3029


yuvipanda:
Follow-up to unblock https://github.com/2i2c-org/infrastructure/actions/runs/5977100479

Issue #2356, opened 2023-03-15
Systematically setup redirects for x.2i2c.cloud to community specified domain names


consideRatio:
I believe we don't yet have practices regarding custom domain names as provided by the communities. Among handling custom domain names, I appreciate what @pnasrat did for the new QCL hub.

@pnasrat made use of the support chart's templates configured via `redirects` values to render additional Ingress resources that our ingress controller would use to redirect for example `staging.qcl.2i2c.cloud` to `staging.quantifiedcarbon.com`. Like this, we as 2i2c engineers doesn't have to remember the hubs custom domain names!

I think by doing this, we can avoid some complexity of remembering domain names for us in 2i2c mostly, but we also add a bit of extra chore work. I've found myself very often looking in cluster.yaml to know what URL to visit, but that is perhaps acceptable. I think I'd rather do some extra chore work for each hub to never have to remember custom domain names as the amount of clusters we handle grow.

### Action point

- [ ] Go through all clusters and their defined hubs and look for custom domain names
- [ ] If needed, add A or CNAME records via namecheap.com for `<something>.2i2c.cloud`, and then configure clusters' support charts via `redirects` to redirect from a `<something>.2i2c.cloud` domain to the community preferred domain name.

pnasrat:
I based my config off UToronto and the docs here

https://infrastructure.2i2c.org/en/latest/howto/manage-domains/index.html

yuvipanda:
This is now codified as part of https://github.com/2i2c-org/infrastructure/pull/3683 for any new hubs we set up.

consideRatio:
Closing as stale for now

PR #3036, opened 2023-08-25
Get rid of 'NASA' from nasa-veda staging hub

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/3029

### DIFF
--- a/config/clusters/nasa-veda/cluster.yaml
+++ b/config/clusters/nasa-veda/cluster.yaml
@@ -13,7 +13,7 @@ support:
 hubs:
   - name: staging
     display_name: "NASA VEDA (staging)"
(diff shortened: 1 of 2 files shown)


consideRatio (approved):

consideRatio (commented):
I would appreciate the staging url redirect, but its part of my wish in https://github.com/2i2c-org/infrastructure/issues/2356 which would for example enable us to visit `nasa-cryo.2i2c.cloud` without needing to think what the community chosen domain name was.

yuvipanda:
@consideRatio I'll do the redirect as part of https://github.com/2i2c-org/infrastructure/pull/3037, to avoid conflicts.

Issue #2988, opened 2023-08-17
[maint] Improve our docs about CILogonOAuthenticator setup


GeorgianaElena:
### Context

In https://github.com/2i2c-org/infrastructure/pull/2979 @sgibson91 pointed out that is not clear from our cilogon docs which setup is considered "standard" and what is considered additional tweaking:

_I expect docs under the Hub Deployment Guide to be our standard practices and should be very copy-pasteable in order to get a hub up and running in as short a time as possible. I feel like I'm currently being presented with too many options and I often don't know which the most appropriate one to pick is. Perhaps we could standardise the options here and move some of the more advanced config, such as IdP stripping and prefixing, to the HowTo section with a cross-ref to the pages in the hub deployment guide?_

### Proposal

As Sarah also suggested, the most common CILogon setup should be as straight-forward and as copy-pastable as possible.

### Updates and actions

I believe we could split the docs at https://2i2c-pilot-hubs--2980.org.readthedocs.build/hub-deployment-guide/configure-auth/cilogon/#cilogon in the following way:

- Keep the section `3.2. Most common configurations for 2i2c clusters` in the deployment guide. This section describes how to achieve various configs, but doesn't contain a "complete" one, that is the "standard", most frequently needed one. We should add one there.
- Keep more complex cilogon setups into their own section, or add them somewhere in the how-to guides
- Move the `Switch Identity Provi
ders or user accounts` section or parts of it to the troubleshoot section. This is because parts of this section explains what to do in various scenarios when you're not able to login using a certain idp.

PR #3040, opened 2023-08-28
Simplify CILogon authentication docs

GeorgianaElena:
Together with https://github.com/2i2c-org/docs/pull/188, fixes https://github.com/2i2c-org/infrastructure/issues/2988


docs/hub-deployment-guide/configure-auth/cilogon.md
```diff
@@ -1,247 +1,135 @@
...
-```{figure} ../../images/403-forbidden.png
-```
+3. All the identity providers must define a `username_derivation` scheme, with their own `username_claim`, that the user *cannot change*. If they can, it can be easily used to impersonate others! For example, if we allow both GitHub and `utoronto.ca` as allowed authentication providers, and only use `email` as `username_claim`, for both providers, any GitHub user can set their email field in their GitHub profile to a `utoronto.ca` email and thus gain access to any `utoronto.ca` user's server! So a very careful choice needs to
+be made here.
+
+    ```{note}
+    You can check the [CILogon scopes section](https://www.cilogon.org/oidc#h.p_PEQXL8QUjsQm) to checkout available values for `username_claim`. This *cannot* be changed afterwards without manual migration of user names, so choose this carefully.
+    ```
+
+## Other most common CILogon configurations across 2i2c hubs
```
sgibson91, line 109:
```suggestion
## Other common CILogon configurations across 2i2c hubs
```

I think saying "most common" is a bit superfluous here

sgibson91 (approved):
This looks great, thank you Georgiana! <3

Issue #2986, opened 2023-08-15
Performance issues in CosmicDS hub


damianavila:
### Context

Ref: https://github.com/2i2c-org/infrastructure/issues/2128

Particularly, https://github.com/2i2c-org/infrastructure/issues/2128#issuecomment-1661076249

> Hey @yuvipanda. Yes, you can go ahead and remove the generic auth now. The test did not go as well as hoped, and we were forced to revert back to our own deployment. There are a few facets to this that I had hoped we could talk about over our Zoom call, but many revolved around subpar performance when a handful of active users were on the hub, as well as a database communication issue which only seems to occur on the 2i2c hub and that we're hoping we can get access to the hub server logs to diagnose.

### Proposal

It seems a Zoom call would need to be scheduled in the future for getting details of the feedback provided above.

yuvipanda:
@nmearl since https://github.com/2i2c-org/infrastructure/issues/2128#issuecomment-1690098583 was closed, let's move that conversation here.

You can look at the logs in the home directory - I think it's called `.jupyter-logs.txt` or similar.

yuvipanda:
@nmearl great! I also sent you an email about your other authentication question!

yuvipanda:
@nmearl ok, at least we no longer seem to require another refresh for the dashboard to load! Can you try that out?

Can you show me the repo used to build this image? Somehow the logs don't seem to be present still.

nmearl:
@yuvipanda I confirm that the dashboard no longer needs another reload! Awesome.

The repo for the build image is [here](https://github.com/cosmicds/hub-user-image).

yuvipanda:
yay! I'm going to close this one out now :)

Issue #2947, opened 2023-08-09
Very slow network on shared 2i2c cluster


sgibson91:
### Context

ref: https://2i2c.freshdesk.com/a/tickets/872

> I am trying upload a 18G zip file to the shared-readonly drive. Upload speeds are really slow at 1.5M/sec.
> I can upload to other sites at 30M/sec.

yuvipanda:
Looking through the various parts that could be bottlenecks, the flow of data is:

User -> User's Internet connection -> GCP connection -> nginx-ingress -> CHP -> jupyter notebook ContentsManager -> Filestore.

User has validated that their connection is fast enough.

I just now validated that Filestore is ok too, by using the `fio` io benchmarking tooll.

```bash
$  deployer exec-homes-shell 2i2c ohw 
$ # Filestore now mounted in /home
$ cd /home/
$ mkdir testing-fio
$ apt update && apt install -y fio
# command copied from https://cloud.google.com/compute/docs/disks/benchmarking-pd-performance,
# but adapted to numjobs=1, as we're on a single container mounting this over NFS!
# But this should mimic exactly what our user homedirectory storage gets
$  fio --name=write_throughput --directory=testing-fio --numjobs=1 \
--size=10G --time_based --runtime=60s --ramp_time=2s --ioengine=libaio \
--direct=1 --verify=0 --bs=1M --iodepth=64 --rw=write \
--group_reporting=1 --iodepth_batch_submit=64 \
--iodepth_batch_complete_max=64
write_throughput: (g=0): rw=write, bs=(R) 1024KiB-1024KiB, (W) 1024KiB-1024KiB, (T) 1024KiB-1024KiB, ioengine=libaio, iodepth=64
fio-3.28
Starting 1 process
write_throughput: Laying out IO file (1 file / 10240MiB)
Jobs: 1 (f=1): [W(1)][64.3%][w=80.0MiB/s][w=80 IOPS][eta 00m:35s] 
write_throughput: (groupid=0, jobs=1): err= 0: pid=692: Mon Aug 21 04:55:54 2023
  write: IOPS=106, BW=108MiB/s (113MB/s)(6538MiB/60564msec); 0 zone resets
    slat (usec): min=206, max=38728, avg=512.62, stdev=1142.26
    clat (msec): min=376, max=1128, avg=594.98, stdev=39.53
     lat (msec): min=377, max=1128, avg=595.49, stdev=39.44
    clat percentiles (msec):
[...]

PR #3041, opened 2023-08-28
Move CosmicDS to AWS hub

yuvipanda:
The dashboards are a affected at first load by the network slowdown being debugged in
https://github.com/2i2c-org/infrastructure/issues/2947#issuecomment-1685985572. Originally we thought it only affects uploads, but definitely seems to affect downloads too.

That seems to be actually the *only* cluster affected, so we can temporarily move them to the AWS cluster to see how that works. There is no persistent storage nor user management, so this is a simple move.

Ref https://github.com/2i2c-org/infrastructure/issues/2986


consideRatio (approved):
Not confident on anything related to billing or community expectations of being on GCP/region vs AWS/region, but if you think thats not a blocker go for it as the other parts looks fine.

Issue #3015, opened 2023-08-23
Considerations of max user session lifetime


consideRatio:
I just saw some straggling user pods that has been around for a month or so hogging up a total of 160 CPU/month. That could have been averted if we for example declared an upper bound on user session length.

I suggest we evaluate if we want to develop a way to help community champions decide if they want to use a safeguard like this, and if we want to introduce a default in existing clusters, and if we want to set a default for new clusters.

The jupyterhub-idle-culler, as configurable via the jupyterhub helm chart, provides this simple option. See notes about [the `--max-age` flag](https://github.com/jupyterhub/jupyterhub-idle-culler#command-line-flags).

colliand:
Thanks @consideRatio. I like your suggestion. 

Cloud costs are a concern among hub champions.  Some users, anecdotally this is rare, launch long-running jobs that persist during intervals of little UX interaction. 

I suggest that 2i2c implements as default an aggressive culling strategy designed to minimize cloud costs. We should write associated documentation that describes the default and provides guidance on how hub champions can override the default with additional risk of cloud cost explosions. Partnerships should inform hub champions about the default and override options during the service launch process. With this approach, 2i2c makes opinionated choices we think will lower cloud costs without troubling most usage scenarios while offering flexibility with more risk explicitly described as the champion/community responsibility.

consideRatio:
Assuming there is consensus that we want to introduce a new behavior to existing clusters, the biggest blocker in my mind is communication.

In software projects one can cut a new major release and document a breaking change in the changelog, and the users of the software can upgrade to the new version but is kind of responsible for reading the changelog before they do.

I think we lack an equivalent to this for the service we provide. I opened #3017 about this.

yuvipanda:
Conservatively, let's just start with a week long max pod duration. That should cut the worst offenders. I agree communication needs work, but currently there's nobody assigned to it and I don't want us to block having *some* limit on that.

colliand:
I love the way @consideRatio considers the impact of changes on our users. Thanks Erik! I want 2i2c to have optimized best practices for communicating changes to our partner communities. That said, we are a small team and have to do what we can with the resources available so the guidance from @yuvipanda seems wise here. I'll write more on the policy issue.

jmunroe:
PR #3042 is now unblocked.

We next modify the documentation:

[...]

Issue #3017, opened 2023-08-23
A procedure for communicating breaking changes is needed


consideRatio:
In 2i2c-org/infrastructure#3015, the idea is surfaced to introduce a "breaking change" for the jupyterhub deployments. In this case, its the idea to configure a max session duration for a user server to act as a safeguard from failing to scale down.

In this issue I want to highlight that we have a need for handling a "breaking change" like this, because it would be breaking. Maybe some communities want a change like this, but I know that not all wants it - and introducing it could do significant harm.

I think we need a procedure to handle a situation like this.

## Situations to consider

- Maintenance of dedicated / shared clusters causing downtime for communities
- Making a change like 2i2c-org/infrastructure#3015, setting an upper limit on the duration of a user session

## Related

- 2i2c-org/infrastructure#3048 - a smaller scoped issue about emailing specifically

damianavila:
cc @colliand and @jmunroe who should be involved in this sort of communication with the communities.

colliand:
Thanks @consideRatio for creating this issue. Within current constraints, I'll share some guidance here since we don't have bandwidth to develop a carefully considered policy. 

2i2c (and the upstream communities we rely upon) delivers services and software with a variety of parameters. Some of these are promised as part of our service agreement. Some are explicit in our documentation and service descriptions. Others are more implicit, visible in the code base but not highlighted in business documents. 

2i2c must not introduce "breaking changes" that break promises in service agreements. 

2i2c should not break expected services in our product/services description. 

2i2c can introduce "breaking changes" on implicit parameters but should consider impacts on our partner communities. 

2i2c must track these changes to diagnose and resolve issues and bugs that may be caused by the changes.

consideRatio:
Thank you @colliand!!

I think you've answered a related topic but not exactly what i was looking for.

> Changes that may generate "significant harm" require care and communication

What I'd like is to understand how to go about making a change like this. I think #3042 could be seen as such change, but how do we then communicate about it?

I proposed that we would write relevant docs and send an email to all community champ

colliand:
I like the proposed plan. That said, I miss enough email to wonder if this strategy is enough. Can we also use "in Jupyter lab" banners? I imagine logging into the hub, seeing a banner with info on the breaking change, and a hyperlink pointing to the documentation.

consideRatio:
@colliand thanks for thinking with me about this!!

[...]

PR #3042, opened 2023-08-28
Cull pods that run for longer than 7 days

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/3015

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -59,7 +59,7 @@ global: {}
 
 jupyterhub:
   cull:
-    # Don't allow any nodes to run for longer than 7 days
+    # Don't allow any user pods to run for longer than 7 days by default
     maxAge: 604800 # 7 days in seconds
   custom:
     auth:
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -58,6 +58,9 @@ inClusterNFS:
 global: {}
 
 jupyterhub:
+  cull:
+    # Don't allow any nodes to run for longer than 7 days
+    maxAge: 604800 # 7 days in seconds
   custom:
     auth:
       anonymizeUsername: false


damianavila (approved):

consideRatio:
I agree with this change, but i dont think we should introduce it silently. I think this is something the QCL community should be informed about at least as they have in the past had longer time running calculations on a single node.

I think it could make sense to notify all community champions about this, and i figure this config is us influencing the user env - so it should be documented somewhere for them.

I opened #3017 about communication in situations like this among other things as i've felt that I lack agency on how to communicate a change influencing users if we've decided its the right thing to do.

For this change, i suggest:
- we communicate that we planning this change at a certain date, referencing docs we've written explaining the default of having a max age.
- invite users to opt-out or get another max-age
- do the change

yuvipanda:
Thanks @consideRatio.

I don't think I've the capacity to work on communicating at this point, and am not sure who does. So I'm going to unassign this to myself as I don't think I can take this to completion.

jmunroe:
We just received feedback from QCL (https://2i2c.freshdesk.com/a/tickets/972) on this:

> 7 days would be excellent for us. Will this be the same limit as the timeout for a disconnected session (that still has processes running)?

I think the answer to the question is "yes" but I welcome corrections if there is subtly on this.

In any case I think this response unblocks this PR!

consideRatio (approved):
I think a lot of the long running jobs observed was caused by the KubeSpawner bug in 6.0.0 that has been fixed and cleaned up. With that in mind, I no longer expect this to be as breaking as I thought when seeing several long lived pods.

I'll push a commit to update the docs currently saying there isn't a max lifetime limit to say that there is one of 7 days: https://docs.2i2c.org/admin/howto/control-user-server/#stop-user-servers-after-inactivity

consideRatio:
@jmunroe i pinged you for review by mistake. I meant to assign to for attribution for work done!

PR #3043, opened 2023-08-28
2i2c-aws-us: Move NFS config to common files

yuvipanda:


damianavila (commented):
I like the idea but it seems there is a real failure.
Maybe there is a missing piece here at the time to specify the `basehub-common.values.yaml` file for each if the hubs in the `clusters/2i2c-aws-us/cluster.yaml` file?

yuvipanda:
@damianavila yep good catch.

GeorgianaElena (approved):

PR #3047, opened 2023-08-29
Enable `unlisted_choice` in LEAP hub

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/2146

Just brings in @GeorgianaElena's work in https://github.com/jupyterhub/kubespawner/pull/769

### DIFF
--- a/config/clusters/leap/common.values.yaml
+++ b/config/clusters/leap/common.values.yaml
@@ -37,6 +37,9 @@ basehub:
             name: LEAP
             url: https://leap-stc.github.io
     hub:
+      image:
+        name: quay.io/2i2c/unlisted-choice-experiment
+        tag: "0.0.1-0.dev.git.6801.h3f4f0c4a"
       allowNamedServers: true
       config:
         Authenticator:
@@ -189,6 +192,13 @@ basehub:
                     cpu_limit: 16
             image: &profile_list_profile_options_image
               display_name: Image
+              unlisted_choice: &profile_list_unlisted_choice
+                enabled: True
+                display_name: "Custom image"
+                validation_regex: "^.+:.+$"
+                validation_message: "Must be a valid public docker image, including a tag"
+                kubespawner_override:
+                  image: "{value}"
               choices:
                 pangeo_new:
                   display_name: Base Pangeo Notebook ("2023.07.05")
@@ -268,6 +278,7 @@ basehub:
           profile_options:
             image:
               display_name: Image
+              unlisted_choice: *profile_list_unlisted_choice
               choices:
                 tensorflow_new:
                   display_name: Pangeo Tensorflow ML Notebook ("2023.07.05")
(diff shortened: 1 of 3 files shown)


consideRatio (approved):

yuvipanda:
This unfortunately still has the problem previously reported, where once you use an image as the other choice, you can't use any other image :( This shouldn't be deployed to the LEAP prod hub until we figure that out.

consideRatio:
> This unfortunately still has the problem previously reported, where once you use an image as the other choice, you can't use any other image :(

I reproduce this once before in staging.leap.2i2c.cloud earlier today, but I've not been able to reproduce it later or with a local k3s cluster with jupyterhub 4.0.2 + kubespawner's main branch.

GeorgianaElena:
I've just pushed a commit that I think fixes it. I've manually deployed it to staging https://staging.leap.2i2c.cloud and seems to be working.

It looks like it was the same bug, with the same fix as the one in https://github.com/jupyterhub/kubespawner/pull/766.  But because we were using a callable `profile_list` defined by us, that fix was not applying for our case.

GeorgianaElena (approved):

yuvipanda:
Yesss, I can confirm that fixes it @GeorgianaElena!!!! \o/

consideRatio:
Nice work @GeorgianaElena!!!

I think there is another fix to make in kubespawner also. I think [this logic](https://github.com/jupyterhub/kubespawner/blob/dc119c20db846d5cff786ed9b47ee64b02a73746/kubespawner/spawner.py#L3234-L3239) shouldn't just take the value returned by profile_list, but should make a copy of the value returned by a profile_list callable.

```python
        if callable(self.profile_list):
            profile_list = await maybe_future(self.profile_list(self))
        else:
            # Use a copy of the self.profile_list dict,
            # otherwise we might unintentionally modify it
            profile_list = copy.deepcopy(self.profile_list)

        profile_list = self._populate_profile_list_defaults(profile_list)
```

This upstream fix isn't a problem if the callable profile_list returns a copy at all times as done in the leap hub - but we don't return a copy at all times, for example when we return this:

```
            # Only apply to GitHub Authenticator
            if not isinstance(spawner.authenticator, GitHubOAuthenticator):
                return original_profile_list
```

consideRatio:
I'll submit a kubespawner PR to make a copy of the returned value, and then we should be fine - but we will need the fix in [b62a46a](https://github.com/2i2c-org/infrastructure/pull/3047/commits/b62a46a019f3b6b76dc5807e07f51883f0a1368c) as well I think.

consideRatio:
> @consideRatio I just looked into that, and I _don't_ think that works - the problem is in

I agree that the problem we have for the leap hub is the specific one patched by @GeorgianaElena, and would be a problem even if we patched what I consider patching in KubeSpawner.

But, I think there is another bug, and I think we are exposed to it because we break the advice of "if you are using profile_list as a callable, make sure to make a copy!".

1. We end up with a callable profile_list
2. We end up storing a single reference to what the callable profile_list will return
3. KubeSpawner doesn't make a copy of the return value of a callable profile_list as it does for a non-callable dictionary config value

Issue #3053, opened 2023-08-29
Deploy `unlisted_choice` to NASA VEDA hub


yuvipanda:
From @j08lue on slack:

Great to see this released, @yuvipanda and @batpad! This is a power feature for very advanced users but maybe also less technical teams will use this in the future as a shortcut to sharing reproducible environments - we will see.
So, yes, please deploy this on the VEDA hub.
Do you have any 2i2c-side documentation of this feature yet? If not, we can just add some to the VEDA user docs for our needs.

yuvipanda:
This has been deployed, and it was demo'd at the VEDA meeting today as well.

j08lue:
I hope I did it justice - would like to prepare another demo for next time (early October), where I have an actual image ready.

yuvipanda:
@j08lue yeah, a menu of custom images for people to choose from would be quite nice!

PR #3054, opened 2023-08-29
Deploy `unlisted_choice` to NASA VEDA hub

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/3053

### DIFF
--- a/config/clusters/nasa-veda/common.values.yaml
+++ b/config/clusters/nasa-veda/common.values.yaml
@@ -32,6 +32,9 @@ basehub:
             name: "NASA"
             url: https://www.earthdata.nasa.gov/esds
     hub:
+      image:
+        name: quay.io/2i2c/unlisted-choice-experiment
+        tag: "0.0.1-0.dev.git.6801.h3f4f0c4a"
       allowNamedServers: true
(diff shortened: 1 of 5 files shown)


j08lue:
"bring your own image"

config/clusters/nasa-veda/common.values.yaml
```diff
@@ -79,6 +82,13 @@ basehub:
           profile_options:
             image: &image_options
               display_name: Image
+              unlisted_choice:
+                enabled: True
+                display_name: "Custom image"
+                validation_regex: "^.+:.+$"
+                validation_message: "Must be an image location, matching ^.+:.+$"
```
sgibson91, line 89: Do we expect users to know what we mean when we present them with regex like this, or should we provide something more verbose/a link out to further resources?

j08lue: Btw, is that regex not just `something:something` and even matches `:::` or ` : `, for example? 

If the allowed pattern is indeed that wide, we can probably describe it as `image:tag`?

consideRatio (approved):

PR #3062, opened 2023-08-30
[CloudBank] SBCC New Image

sean-morris:
SBCC is updating its notebooks now and needs the new image for testing while the current classes are using the old image


damianavila:
I am surprised that testing new notebooks with a new image needs a new hub deployment... am I missing something?

PR #3067, opened 2023-09-01
Use custom homepage for victor

sgibson91:
`templateVars` is required by the schema, so I have left it as is. I'm not sure how this affects the custom homepage branch in https://github.com/2i2c-org/default-hub-homepage

ref: https://2i2c.freshdesk.com/a/tickets/925

### DIFF
--- a/config/clusters/victor/common.values.yaml
+++ b/config/clusters/victor/common.values.yaml
@@ -17,6 +17,7 @@ basehub:
         add_staff_user_ids_to_admin_users: true
         add_staff_user_ids_of_type: "github"
       homepage:
+        gitRepoBranch: "victor"
         templateVars:
           org:
             name: Victor


GeorgianaElena (commented):
> templateVars is required by the schema, so I have left it as is. I'm not sure how this affects the custom homepage branch in https://github.com/2i2c-org/default-hub-homepage

@sgibson91, I believe whatever is in `templateVars` will override what has been set in the custom homepage branch.

So, we either modify them here directly, or we make them not be required in the schema? I'm not sure, but I would go with the first one for simplicity.

Either one, I believe what has been set on that branch already existsted in the templateVars here, agree? https://github.com/2i2c-org/default-hub-homepage/commit/eec9ea55fe1cff81080deb23b3c2a5f531b0030c

sgibson91:
Agree!

cc @volcanocyber for future reference ☝🏻

consideRatio (approved):

PR #3070, opened 2023-09-04
Update openscapes staging python image

GeorgianaElena:
Per https://2i2c.freshdesk.com/a/tickets/935

This puts the profileList config in each hub's config file staging/prod and updates the staging python image tag per the request in the support ticket.


consideRatio (approved):
I appreciate how you moved the config out of common.values.yaml instead of just overriding it in staging, that will help us not get the wrong expectations when making future config changes!

betolink:
Thanks @GeorgianaElena @consideRatio, we used to have a command to update the image using the hub REST interface but that doesn't work with the multi profile mode. Would there be a way to scan the latest builds from https://hub.docker.com/r/openscapes/python/tags and update **staging** accordingly?

GeorgianaElena:
@betolink, I am not aware of a recommended way to automatically pick up latest tag for a image and automatically deploy it to a hub.

But, there are a couple of ways we could achieve this without having you request it every time.
**Option 1**
We are experimenting on a few hubs with a new feature "unlisted choice" (that is still unreleased and actively updated) that allows users to manually input an image and a tag to use on a server. It is currently documented at https://infrastructure.2i2c.org/howto/features/allow-unlisted-profile-choice.

We could deploy this on the staging hub and you could use that to try out a new image version. This has the advantage that it doesn't require any 2i2c intervention after we deploy the feature.

**Option 2**
We could enable an action that automatically checks for new tags of a (public) image every _x_ interval of time and opens a PR to update it every time a new tag is found. It is documented at https://infrastructure.2i2c.org/reference/ci-cd/auto-bumping/#bumping-image-tags.

This approach still requires a 2i2c engineer to review, approve and merge the PR in order to deploy.

What do you think? Would any of these options work for you?

betolink:
Thanks @GeorgianaElena, option 1 sounds great, can we try it on staging?

betolink:
quick question, I'd like to test a new image today, should I wait for the `unlisted image` update or should I send a new ticket PR to update it?

PR #3065, opened 2023-08-31
[nasa-ghg] Add custom login page template and fix titles

j08lue:
- [x] Fix metadata for nasa-ghg hub
- [x] Point to customized login page template for the `staging` hub

Ready for review/merging.

PR #3081, opened 2023-09-05
[nasa-ghg] Set custom homepage for production and staging

j08lue:
After successfully testing custom homepage templates on staging (https://github.com/2i2c-org/infrastructure/pull/3065), we now want to do the same for production, please.

Ready for review/merging.


yuvipanda (approved):

yuvipanda:
Thanks @j08lue!

Issue #3019, opened 2023-08-23
Provide OAuth credentials to use hub as OAuth provider for cosmicds


yuvipanda:
Based on prior discussion with CosmicDS, they would need us to provide them with OAuth credentials to the JupyterHub to run a service that needs to auth with the JupyterHub.

```[tasklist]
- [x] Document why someone would do this
- [x] Document how 2i2c engineers should do this
- [x] Actually do it for CosmicDS
```

/cc @nmearl

yuvipanda:
We had a very good meeting with @nmearl today, where we eventually sorted this out. Follow-up from this meeting:

- [ ] https://github.com/jupyterhub/jupyterhub/pull/4549
- [ ] File issue about JupyterHub saying 'client_secret is missing' when the problem is that oauth_redirect_url is missing
- [ ] Document how to use a JupyterHub as the *source* of auth via auth0

PR #3083, opened 2023-09-05
Enable using cosmicds jupyterhub as an authentication *provider*

yuvipanda:
OAuth2 credentials for the CosmicDS portal, which uses this JupyterHub as authentication *provider*. So when users hit "Login" in the CosmicDS portal, they're actually redirected to this JupyterHub (via auth0). This ensures that the portal knows exactly the (anonymized) usernames of these users, and can do additional work on their part to track them as necessary.

From perspective of 2i2c engineers, this is just another JupyterHub service - just one that we don't have to maintain or manage! Our responsibility is to *provide the credentials*, and no more.

Documentation of this feature is coming in a separate PR, so as to not block this one from getting merged.

Ref https://github.com/2i2c-org/infrastructure/issues/3019

### DIFF
--- a/config/clusters/2i2c-aws-us/enc-cosmicds.secret.values.yaml
+++ b/config/clusters/2i2c-aws-us/enc-cosmicds.secret.values.yaml
@@ -1,22 +1,25 @@
 jupyterhub:
     hub:
+        services:
+            cosmicds-portal:
+                api_token: ENC[AES256_GCM,data:QY4XnFjLgb8KqvVcXignYeJg7Xi5S5IWZ9dAFYy326XFo/oeds7DSDdj2rKam6zRX00ZC1+Cvx+FHlhL0FsjYg==,iv:dgw2az+wRJMt4uwklBcuxaYK64ETnKYH8VX0vS6RzRs=,tag:GP2yWFm9u69iUCogRoCkoQ==,type:str]
         extraEnv:
-            USERNAME_DERIVATION_PEPPER: ENC[AES256_GCM,data:AXMgK5+Gzojb2j65OA87X0BEs4JxjZr1jkemLRNhMp5pxdvt40YyMEO2fyhx+nfNwrvMf9DV6z9Hl7l2XEsbTQ==,iv:B9EBaac4VFOkU+nzxcm7LUzqJRJ4N38o4BbsZqxW69Q=,tag:cERvKEh9TfxyoDyzzVrb1Q==,type:str]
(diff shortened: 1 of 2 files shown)


config/clusters/2i2c-aws-us/enc-cosmicds.secret.values.yaml
```diff
@@ -1,22 +1,25 @@
...
     hub:
+        services:
+            cosmicds-portal:
+                api_token: ENC[AES256_GCM,data:QY4XnFjLgb8KqvVcXignYeJg7Xi5S5IWZ9dAFYy326XFo/oeds7DSDdj2rKam6zRX00ZC1+Cvx+FHlhL0FsjYg==,iv:dgw2az+wRJMt4uwklBcuxaYK64ETnKYH8VX0vS6RzRs=,tag:GP2yWFm9u69iUCogRoCkoQ==,type:str]
         extraEnv:
-            USERNAME_DERIVATION_PEPPER: ENC[AES256_GCM,data:AXMgK5+Gzojb2j65OA87X0BEs4JxjZr1jkemLRNhMp5pxdvt40YyMEO2fyhx+nfNwrvMf9DV6z9Hl7l2XEsbTQ==,iv:B9EBaac4VFOkU+nzxcm7LUzqJRJ4N38o4BbsZqxW69Q=,tag:cERvKEh9TfxyoDyzzVrb1Q==,type:str]
+            USERNAME_DERIVATION_PEPPER: ENC[AES256_GCM,data:mH/CtGOgBMPqUfy74r1sOdxxS2UcUAs/JkSItBgNKguTkV6WmBkd1VvEiRbiFmh86/VZrgFMj9k3B+eUfBSvMw==,iv:ZBhhx3mMrydErwPTNRQ391yUcZo4UIaEooz3exBY3c8=,tag:4eTclvSgvqby9qthqjAEcA==,type:str]
         config:
             CILogonOAuthenticator:
-                client_id: ENC[AES256_GCM,data:bG3o+fgg9Un2YzPxgBisMSpZ7mu0NnF3u7fbHFf3TErMRSNZdbKYne9InZfntOSt9CFP,iv:K+L30QknUdByGTTgs/Xo7xdWAl3ceUjyRi09PFFq0Us=,tag:GRtn+QCLaBjToj4Wk42kEg==,type:str]
```
consideRatio, line 7: Is the client id/secret change using CILogon as an identity provider an intended change?

consideRatio (approved):
This looks great, excellent comments!

I asked a question about something that seemed unrelated, but it should be fine i think!

config/clusters/2i2c-aws-us/enc-cosmicds.secret.values.yaml
yuvipanda: @consideRatio nope, and the information itself hasn't changed. I think sops re-encrypted here.

PR #3086, opened 2023-09-06
Document how to provide OAuth2 credentials to a community

yuvipanda:
While we can't *run* arbitrary third party services for our communities, we should provide them with appropriate credentials so they can do so themselves. This allows them to independently build their own portals or other kinds of services on top of JupyterHub using the mechanism meant for this - services.

This accompanies https://github.com/2i2c-org/infrastructure/pull/3083, which sets this up for CosmicDS.

Ref https://github.com/2i2c-org/infrastructure/issues/3019


docs/howto/features/auth-provider.md
```diff
@@ -0,0 +1,75 @@
+# Providing credentials for externally managed services
+
+Sometimes, our communities want to run *their own* [external JupyterHub service](https://jupyterhub.readthedocs.io/en/stable/reference/services.html#externally-managed-services),
+often just to use JupyterHub as an OAuth2 provider for user logins. This
+allows them to have the same shared users between the hub and their 
+service.
```
consideRatio, line 6: I think we should let the title and the example focus solely on registering a jupyterhub service / oauth2 application, and avoid the broader "providing credentials for externally managed services" which should be done differently for all (?) other cases.

If they want to register a service to acquire credentials to interact with the JupyterHub REST API in order to stop user servers or similar, it would not involve `oauth_client_id` or `oauth_redirect_url`, it would instead only involve `api_token` and configuring `load_roles` defining more fine grained permissions to do that.

That makes me think we should scope the docs tighter that this is an example specifically about providing a credentials to use JupyterHub as an identity provider - where we register an OAuth2 application.

docs/howto/features/auth-provider.md
```diff
@@ -0,0 +1,75 @@
...
+   
+3. Define appropriate secrets in the hub's `enc-<hub-name>.secret.values.yaml` file:
+
+   ```yaml
+   jupyterhub:
+     hub:
+       services:
+         <name-of-service>:
+           # This will be the client secret
+           api_key: <output-of-step-2>
```
consideRatio, line 53:
```suggestion
           api_token: <output-of-step-2> 
```

docs/howto/features/auth-provider.md
```diff
@@ -0,0 +1,75 @@
...
+   user_profile_url: https://<hub-domain>/hub/api/user
+   ```
+   
+   In addition, the following pieces of information would be useful:
+   
+   a. The `client_secret` **must** be sent as a token in the `Authorization` header when making the token request to the `token_url`.
+   b. The `name` field from the user profile contains the user's name on the hub.
+   
+   This should be enough information for them to configure their oauth2 client to authenticate
+   against the JupyterHub!
```
consideRatio, line 75: It would be good if we can delegate some details to jupyterhub's docs:

- With regards to the authorization header, its documented in [how to make an api request](https://jupyterhub.readthedocs.io/en/stable/howto/rest.html#make-an-api-request), clarifying for example that you need a format of `Authorization: token <client_secret>`.
- There is also configuration references for all the endpoints
  - [GET `oauth2/authorize`](https://jupyterhub.readthedocs.io/en/stable/reference/rest-api.html#/default/get_oauth2_authorize)
  - [POST `oauth2/token`](https://jupyterhub.readthedocs.io/en/stable/reference/rest-api.html#/default/post_oauth2_token)
  - [GET `user`](https://jupyterhub.readthedocs.io/en/stable/reference/rest-api.html#/default/get_user)

  I've often looked at references for the userinfo endpoint as knowing what data is available and how its formatted is often very relevant and not standardized across identity providers.

docs/howto/features/auth-provider.md
```diff
@@ -0,0 +1,75 @@
...
+
+5. Send our community the following pieces of information to configure their
+   OAuth2 *client* with:
+   
+   ```
+   client_id: <oauth_client_id_from_step_1>
+   client_secret: <api_key_from_step_3>
+   authorization_url: https://<hub-domain>/hub/api/oauth2/authorize
+   token_url: https://<hub-domain>/hub/api/oauth2/token
+   user_profile_url: https://<hub-domain>/hub/api/user
```
consideRatio, line 66: These URLs can be referred to various things and has caused me a bit of confusion historically for various identity providers. I suggest we refer to them as openid connect has standardized: `authorization`, `token`, and `userinfo`.

```suggestion
   userinfo_url: https://<hub-domain>/hub/api/user
```

consideRatio (commented):
Great work on this!!

I think the key change I'd like to see is a narrowing of the focus towards registering an OAuth2 application with JupyterHub, as compared to provding credentials for any kind of externally managed service - as that involves `JupyterHub.load_roles`, `services[].url` `services[].display`, and does't involve `services[].oauth_client_id` for example.

docs/howto/features/auth-provider.md
yuvipanda: While in general I agree about delegation, I think in this case most people don't actually even know that OAuth2 requires 3 endpoints, so I think I would like to keep this here so a 2i2c engineer knows exactly what information to send to end user. I also forgot about the 3 endpoints while doing this with CosmicDS yesterday, and had to hunt around and look it up. Instead, this can be copy pasted and updated and sent to the community.

docs/howto/features/auth-provider.md
```diff
@@ -0,0 +1,75 @@
...
+         # Name of the service should be provided to us by the community
+         <name-of-service>:
+           name: <name-of-service>
+           # The client_id for this service - it *must* start with service-
+           oauth_client_id: service-<name-of-service>
+           # The OAuth2 redirect URI provided to us by the community
+           oauth_redirect_uri: <oauth-redirect-uri>
+           # Set to true if it should be displayed under 'services' in the
+           # hub control panel
+           display: true/false
```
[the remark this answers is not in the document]
yuvipanda: Ah, I didn't know that! Good to know, I'll omit it. Thanks for catching that, @consideRatio!

docs/howto/features/auth-provider.md
yuvipanda: Yeah, I had originally written this out as 'how to make JupyterHub an auth provider' (hence the file name!), but then tried to broaden it. But as you say, in most other service cases you only need api_token - I hadn't considered that. So I'll revert this back to the previous framing.

yuvipanda (commented):
Thanks @consideRatio.

docs/howto/features/auth-provider.md
consideRatio: I think by linking to configuration reference for the various endpoints are very relevant if we are also having a note about how to format the `Authorization` header.

In situations where I've setup authentication against a service, I've had great use of endpoint docs. In their shoes I may for example want to read the jupyterhub username from the userdata url - but what response should I expect and under what key should I look to find that?

I'll push a commit with a proposed resolution!

consideRatio:
@yuvipanda I pushed e2e51eb that includes a commit message with a summary of most changes, previewed below from https://2i2c-pilot-hubs--3086.org.readthedocs.build/howto/features/identity-provider/

image
image
image

consideRatio (approved):
With the pushed change, this LGTM!

yuvipanda (commented):
Thanks for rewriting the entire document, @consideRatio ;) I can't 'request changes' on my own PR, but I left a note. Can you amend that and then this can be merged?

docs/howto/features/identity-provider.md
```diff
@@ -0,0 +1,115 @@
...
+   jupyterhub:
+     hub:
+       services:
+         <name-of-service>:
+           # This will be the client secret
+           api_token: <output-from-openssl>
+   ```
+
+   ````{note}
+   If `api_token` isn't configured explicitly, it gets generated by the
```
yuvipanda, line 69: It was an intentional choice to explicitly set `api_token` for this specific case. So if we had to recreate the cluster or redeploy the hub, it doesn't change! If it does, their identity provider will immediately stop working, and we will need to communicate and co-ordinate this. So that was the reason for explicitly setting it, although right to replicate is also useful.

consideRatio: Thanks @yuvipanda, ensuring we can do fresh re-deploys without ensuring retention of this state seems like a more direct motivation worth focusing on!

I removed the details on how to read the k8s secret it and updated the motivation for explicit configuration to mention the need to recoordinate things.

While omitted from this text as it felt a bit out of scope, if the community wants to reduce the complexity of their configuration in the future to not manage such secret, they could opt to not set the `api_token` explicitly at any point in time, the same `api_token` will be still be used.

yuvipanda:
I appreciate the work you have done here, @consideRatio

consideRatio:
I appreciate the work you have done here as well @yuvipanda!

Thanks for tolerating me to butt in, let me know if its too much - I could absolutely see that. I wanted to help this go to completion and felt my wants was blocking it and ended up burdening you.

yuvipanda:
@consideRatio this played out perfectly from my perspective! Please feel no guilt

Issue #2930, opened 2023-08-05
Investigate explosion of disk usage of prometheus


yuvipanda:
<img width="1728" alt="image" src="https://github.com/2i2c-org/infrastructure/assets/30430/eb0391e5-551e-4003-9975-04d3cfd17ecc">

You can see that disk usage exploded around June 28. https://github.com/2i2c-org/infrastructure/pull/2730 coincides with that - a prometheus version bump

yuvipanda:
No corresponding explosion in time series

<img width="1726" alt="image" src="https://github.com/2i2c-org/infrastructure/assets/30430/a1cf6ba3-e49d-40a0-aaac-63680e036472">

Using prometheus_tsdb_head_series

yuvipanda:
Second inflection point was around when https://github.com/2i2c-org/infrastructure/pull/2621  was merged

yuvipanda:
Looks like this is actually the case for a lot of our prometheii

yuvipanda:
https://github.com/2i2c-org/infrastructure/pull/2932 bumps up everything to a miniumum of 200Gi.

yuvipanda:
I made a simple prometheus dashboard showing remaining disk space in all our prometheuses: https://grafana.pilot.2i2c.cloud/d/d560fa89-38bc-4253-964e-82573ea91d57/global-prometheus-health?orgId=1&var-cluster=All

<img width="1728" alt="image" src="https://github.com/2i2c-org/infrastructure/assets/30430/f36f28f7-f86a-4014-bd84-71bd716d19be">

yuvipanda:
<img width="1720" alt="image" src="https://github.com/2i2c-org/infrastructure/assets/30430/a88b25b0-d6af-4627-a625-978ac3f30b9b">

ok, most prometheuses have more disk space now! But this isn't a long term solution yet.

consideRatio:
These relevant changes were [made June 27-29](https://github.com/2i2c-org/infrastructure/pulls?q=is%3Apr+merged%3A2023-06-27..2023-06-29).

- https://github.com/2i2c-org/infrastructure/pull/2706
- https://github.com/2i2c-org/infrastructure/pull/2730
- https://github.com/2i2c-org/infrastructure/pull/2731

I wonder if it was prometheus version bump or config change.

yuvipanda:
[...]

PR #3090, opened 2023-09-07
Set default prometheus disk size to 400Gi

yuvipanda:
A lot of instances with 200Gi are in fact failing, as noticed in https://github.com/2i2c-org/infrastructure/actions/runs/6104903072, because their disks are full.

The goal is to try to bring these back up first, and then we can work on figuring out why they may be down.

Ref https://github.com/2i2c-org/infrastructure/issues/2930


yuvipanda:
Merging to unbreak prometheii

PR #3091, opened 2023-09-07
Increase shared cluster prometheus disk size to 1Ti

yuvipanda:
512Gi was full. We need to still investigate why, but this brings the prometheus back up

Ref https://github.com/2i2c-org/infrastructure/issues/2930


yuvipanda:
I've already deployed this

GeorgianaElena (approved):

Issue #2934, opened 2023-08-07
Timeouts when accessing prometheus metrics from grafana


consideRatio:
We have a few major issues with prometheus it seems:

- It seems prometheus has started to consumes more disk space
  #2930
- It seems prometheus regularly times out when grafana makes "last 30 days" requests for dashboards
  [Example query](https://grafana.pilot.2i2c.cloud/d/d560fa89-38bc-4253-964e-82573ea91d57/global-prometheus-health?orgId=1&from=now-30d&to=now)
- It seems we can no longer successfully access much historical data
  [Example query](https://grafana.pilot.2i2c.cloud/d/MMHgC_Qnz/cluster-information?orgId=1&from=now-7d&to=now&var-PROMETHEUS_DS=cloudbank)
  - Or maybe it was prometheus failing to scrape for a longer duration? It seems that it was a major gap between 2023-07-21 21:09 and 2023-08-05 18:58 UTC time for openscapes specifically.

yuvipanda:
First and last bullet points are the same - the disk space has been full for a while, so there was no data collected. Let's focus discussion about that on https://github.com/2i2c-org/infrastructure/issues/2930.

yuvipanda:
I've retitled this issue to focus purely on the grafana timeouts.

Looks like they are timing out at 30s mark. I looked in the nginx logs, and found that the error code is `499`, which means it's the *client* terminating the request. I raised the timeout in the grafana UI to 120s - the default for a prometheus *server* query termination. And now I see that this just caused a server restart. So at least, one part of the problem is that the grafana timeout is too small.

yuvipanda:
The increased timeout in https://github.com/2i2c-org/infrastructure/pull/2942 now allows some more queries to load that weren't able to load before!

I think our next step is a 60s default timeout with nginx-ingress that needs to be raised, as that currently masks issues as well.

yuvipanda:
<img width="1725" alt="image" src="https://github.com/2i2c-org/infrastructure/assets/30430/b819d669-db39-4c6c-a13b-a12faba60230">

This grafana panel showing CPU distribution in openscapes took 46s to load and was failing before this change.

yuvipanda:
Evidence for nginx timeouts:

```
[...]

PR #3093, opened 2023-09-07
Revert customized prometheus block duration config

yuvipanda:
Based on suspicion in
https://github.com/2i2c-org/infrastructure/issues/2934#issuecomment-1669619133 that this is contributing to the slowness of prometheus. Could also be related to why disk space usage is growing so much

Ref https://github.com/2i2c-org/infrastructure/issues/2930 Ref https://github.com/2i2c-org/infrastructure/issues/2934

Reverts config changes from #2731 and #2778

### DIFF
--- a/helm-charts/support/values.yaml
+++ b/helm-charts/support/values.yaml
@@ -180,21 +180,10 @@ prometheus:
         --storage.tsdb.path=/data \
         --web.console.libraries=/etc/prometheus/console_libraries \
         --web.console.templates=/etc/prometheus/consoles \
-        --web.enable-lifecycle \
-        --storage.tsdb.min-block-duration=60m \
-        --storage.tsdb.max-block-duration=60m"
+        --web.enable-lifecycle"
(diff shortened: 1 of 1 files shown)


consideRatio:
The defaults are 30-120 (min-max) if i recall correctly, can we make these have an interval as well? For example 30-60?

I think otherwise its possible that some operation continuously is triggering.

yuvipanda:
@consideRatio I've just removed the customization completely, so it's back to the defaults now.

yuvipanda:
@consideRatio haha, we clashed on the title change simultaneously :D I'll revert back to yours

consideRatio (approved):
:)

Okay lets be on the lookout for prom servers failing to start up because they run out of memory.

I hope that in 2 weeks, we'll be able to browse data going back at least two weeks then.

yuvipanda:
Tested on the 2i2c cluster, this deploys ok.

yuvipanda:
@consideRatio OK THIS DID FIX IT!

image

Issue #2024, opened 2023-01-06
Reflect over the current's `deployer`s package design


GeorgianaElena:
### Context

In https://github.com/2i2c-org/infrastructure/pull/1938 @sgibson91 correctly pointed out the naming of the files and possible their location in `deployer/grafana` and `deployer/generate` might need more thought put into it.

### Proposal

Reflect over the design of:
- `deployer/grafana`
- `deployer/generate`
- `deployer/utils.py`

sgibson91:
Copying my comment across for context:

> I have a small wondering on file naming. For example, now that we have a grafana subfolder, I would say that grafana/utils.py and grafana/manage_tokens.py would reduce redundancy compared to grafana/grafana_utils.py and grafana/grafana_tokens.py etc, particularly when importing, i.e. from grafana import utils versus from grafana import grafana_utils. In the utils-specific case we'd need to avoid naming clashes with the top-level utils though. Which indicates to me either (i) we are overusing utils and should be more creative in our filenaming, or (ii) I am overthinking package design :D

PR #3094, opened 2023-09-07
deployer: create separate new typer apps under the deployer and update the general structure of the directory

### LINKED ISSUE #971: Refactor for clarity that deployer/tests aren't just developer tests
## Description of problem and opportunity to address it

The `deployer` CLI rely on `pytest` as a production dependency - not just development dependency. This is because it use `pytest` and the tests in `deployer/tests` as a way to verify that hub's are functional.

From a glance, I would assume that those tests would be tests associated with testing the `deployer` CLI itself, because that is a pattern I've seen across the python ecosystem repeated. This is the first time I see `pytest` being used a real requirement for a project. Due to that, I suggest we are more explicit about this somehow.

### Should pytest be used at all?
@sgibson91 mentioned to me that @yuvipanda wants to avoid using `pytest` as a production dependency. That sounds reasonable to me and would solve this issue as well.

### Pytest and logging
@sgibson91 also noted that there has been issues with a hub health-check test that fails could lead to `pytest`'s verbosity exposes sensitive credentials. By doing a test manually, we would also avoid such complexity. This is related to:
- https://github.com/2i2c-org/infrastructure/issues/585
- https://github.com/2i2c-org/infrastructure/issues/694.

## Updates and ongoing work

- 2023-10-11, this issue is still relevant but low prio.

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/2024

## What the PR does
This PR refactors the deployer's directory structure and creates new  typer applications under the main `deployer` app.
The commands that we have, then register under these new applications based on their location in the package.

*NOTE*: the only new code that this PR adds is related to the creation of these typer apps. Other then this, this mostly is about moving files around, renaming them and extracting functions into their own files. More in the  commits' description.

## Benefits
We are now left with a more clear to follow structure of the main `deployer` directory instead of having an endless number of python files in a big pool. 
More, this directory structure maps to the deployer commands, so that it's easier to know where the code that describes a command is.

## A note to review more easily
I was very grateful each time @consideRatio split his work into individual logical commits and described what each commit did. That greatly improved my understanding of what the PR did and presented the changes gradually, hence the review was easier <3.

Although I didn't start with this thought in mind because this PR was supposed to be something else and smaller, I did my best to squash the relevant ones together and provide a description for each of them.

So, to review more easily, I suggest going to https://github.com/2i2c-org/infrastructure/pull/3094/commits and look at each commit and its description at a time. Thank you!!!

---

EDIT: fixes #971


deployer/README.md
```diff
@@ -17,172 +17,191 @@ pip install --editable .
...
-### `deploy`
+### The `commands` directory
 
-This function is used to deploy changes to a hub (or list of hubs), or install it if it does not yet exist.
-It takes a name of a cluster and a name of a hub (or list of names) as arguments, gathers together the config files under `/config/clusters` that describe the individual hub(s), and runs `helm upgrade` with these files passed as `--values` arguments.
+This is the directory where all of the code related to the main `deployer` sub-commands is stored.
 
-**Command line usage:**
+Each sub-commands's functions are stored:
+- either in a single Python file that ends in `_cmd` or `_commands`
```
yuvipanda, line 63: Given we're already in a subdirectory called `commands`, I think it's ok to not have this additional suffix. Less clutter that way.

deployer/README.md
```diff
@@ -17,172 +17,191 @@ pip install --editable .
...
+│   │   ├── deploy_dashboards_cmd.py
+│   │   ├── tokens_cmd.py
+│   │   └── utils.py
+│   └── validate
+│       ├── cluster.schema.yaml
+│       └── config_cmd.py
 ```
 
-### `validate`
+### The `tests` directory
```
yuvipanda, line 102: While we're moving this around, let's call these 'health-check-tests' or something like that for clarity, that it's not testing the deployer itself?

deployer/README.md
```diff
@@ -499,27 +669,64 @@ create/delete/get/get-all/update/ CILogon clients using the 2i2c administrative
...
+```bash
+ Usage: deployer exec [OPTIONS] COMMAND [ARGS]...                                                                       
+                                                                                                                        
+ Execute a shell in various parts of the infra. It can be used for poking around, or debugging issues.                  
+                                                                                                                        
+╭─ Options ────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
+│ --help          Show this message and exit.                                                                          │
+╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
+╭─ Commands ───────────────────────────────────────────────────────────────────────────────────────────────────────────╮
+│ debug                                                                                                                │
```
yuvipanda, line 695: Let's invert this, so `debug` is the top level, and `exec` is a sub-level? None of the things under `debug` are execing anything, while usually we `exec` things to debug something...

yuvipanda: So the layout would be:

```
- debug
  |- exec
  |- user-logs
  |- component-logs
  |- start-docker-proxy
```

yuvipanda: Alternatively, `debug` and `exec` can *both* be top level.

deployer/infra_components/cluster.py
```diff
@@ -3,11 +3,35 @@
...
+    get_decrypted_files,
+)
+from deployer.utils.rendering import print_colour
+
 from .hub import Hub
-from .utils import print_colour, unset_env_vars
+
+
+@contextmanager
+def unset_env_vars(vars):
```
yuvipanda, line 18: I feel this is generic enough it should continue to be in something under `utils` rather than here.

yuvipanda (changes requested):
Thanks @GeorgianaElena! Generally looks good to me, made some minor suggestions.

consideRatio (commented):
I love that tests is being renamed to involve health_checks, i consider that to fix #971!

GeorgianaElena:
> I love that tests is being renamed to involve health_checks, i consider that to fix https://github.com/2i2c-org/infrastructure/issues/971!

This was part of @yuvipanda's feedback <3 🚀

> Thanks @GeorgianaElena! Generally looks good to me, made some minor suggestions.

I've udpated based on your feedback, Yuvi and rebased & fix the conflicts.

deployer/infra_components/cluster.py
```diff
@@ -3,11 +3,16 @@
...
-from .file_acquisition import get_decrypted_file, get_decrypted_files
+from deployer.utils.env_vars_management import unset_env_vars
+from deployer.utils.file_acquisition import (
+    HELM_CHARTS_DIR,
+    get_decrypted_file,
+    get_decrypted_files,
+)
+from deployer.utils.rendering import print_colour
+
 from .hub import Hub
```
consideRatio, line 15: I expected `from .hub` to become `from deployer.infra_components.hub` given the other changes so that we consistently use either relative or absolute references.

I'm generally a bit uncertain about the choice between relative `.` references and absolute `package_name` references, not knowing the consequences of going for one, the other, or a mix.

consideRatio (commented):
The UI is far better with the sub-commands for sure, nice!

I expect this to be safe to merge for our production hubs operation since we have staging hubs also. I've not reviewed all commits or all changed lines, but skimmed through a bit.

To resolve:
- another merge conflict
- comment about `from .hub` import

deployer/infra_components/cluster.py
GeorgianaElena: @consideRatio, the main motivation for switching the majority of imports from relative to absolute, is because with this PR at least 2 or 3 more levels are added and it relative imports were not serving their purpose anymore when you have something like `....file_acquisition import find_absolute_path_to_cluster_file`

(Also, apparently there's a PEP that suggests using absolute ones https://peps.python.org/pep-0008/#imports)

But I believe relative imports are still useful for the `deployer` package when referring to the current directory.

For example, in `deployer/commands/generate/dedicated_cluster/aws.py` we have a very simple line that says:
```
from .dedicated_cluster_app import dedicated_cluster_app
```
if we turn it into an absolute import then it becomes:
```
from deployer.commands.generate.dedicated_cluster.dedicated_cluster_app import dedicated_cluster_app`
```

Although not recommended (relative imports are not recommended anyhow), I would leave these imports relative to the current directory be, for simplicity. @consideRatio, do you feel strongly about using absolute imports everywhere? If yes, I will update them.

I added a commit that uses absolute imports in a few more places, but haven't touched the ones in the `commands` directory for the reason above.

consideRatio: > do you feel strongly about using absolute imports everywhere? If yes, I will update them.

No, I'm mostly cautious because I recall issues that I didn't fully understand related to using one or the other kind.

> leave these imports relative to the current directory be

Sure!

consideRatio (approved):
Thank you for working this @GeorgianaElena!! Happy to see sub-commands adopted to compartmentalize the UI and code behind it!

I think we should review this further by testing it rather than by further code review. I don't expect this to break things in a way that we don't catch.

@GeorgianaElena maybe you could do a deploy-support and deploy a staging hub before merging though as a precaution, with a health check on the staging before we trigger the full deploy?

GeorgianaElena:
> @GeorgianaElena maybe you could do a deploy-support and deploy a staging hub before merging though as a precaution, with a health check on the staging before we trigger the full deploy?

Just ran them locally with no issue + a validate command. 

> I think we should review this further by testing it rather than by further code review. I don't expect this to break things in a way that we don't catch.

Thank you @consideRatio! Merging this now 🚀

PR #3097, opened 2023-09-07
singleuser post-start lifecycle hook for /veda-docs

ranchodeluxe:
### Changelog

* update `singleuser` image tag to get newest `stac_ipyleaeflet` plugin in there

* use rocker image tag `geospatial_unstable` to get newest GDAL in there based on discussion with rocker folks

* add post-start hook 

### Context about / veda-docs
VEDA (https://staging.nasa-veda.2i2c.cloud/ and https://nasa-veda.2i2c.cloud/) want to have all user pods start up [with the veda-docs](https://github.com/NASA-IMPACT/veda-docs) in a similar fashion to how MSPC is doing it. Even though [we currently have some entry points covered ](https://github.com/NASA-IMPACT/veda-config/pull/243/files#diff-5ae9e3ede6336192cb5241b5424b4f958a8d752aef072b6048add392c93227fbR9) on the dashboard we want to make sure all entrypoints have `/veda-docs`. So we are willing to take the risks with the `post-start` hook for that reason

### DIFF
--- a/config/clusters/nasa-veda/common.values.yaml
+++ b/config/clusters/nasa-veda/common.values.yaml
@@ -60,7 +60,7 @@ basehub:
         name: public.ecr.aws/nasa-veda/nasa-veda-singleuser
         # Based off pangeo/pangeo-notebook:2023.07.05 which uses JupyterLab <4, so jupyterlab-git and dask-dashboard work
         # If updating this tag, also update it in the profile_options section below
-        tag: "b807c7efa97c8df9ca38779f7e59d09f889fde9299b0d19de80389cf6b064f90"
+        tag: "5068290376e8c3151d97a36ae6485bb7ff79650b94aecc93ffb2ea1b42d76460"
       extraFiles:
         k8s-lifecycle-hook-post-start.sh:
           mountPath: "/etc/singleuser/k8s-lifecycle-hook-post-start.sh"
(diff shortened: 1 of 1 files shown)


ranchodeluxe:
@yuvipanda : based on our convo yesterday should we be using Python as the scripting language and `subprocess` calls for better tracebacks if errors do happen?

yuvipanda:
@ranchodeluxe yes! There's two things to do here:

1. The actual post start script should probably just be *in your image*, written in python. And then the hook here can just *call* that. This way, you can test the script without having to redeploy each time - which is quite cumbersome. By keeping it in the image, you can test it locally by just launching the image and running the command yourself.
2. I don't think the rocker image change works - The `rocker/binder` image has the extra bits required to make a rocker image work on a JupyterHub that the base images don't have. Let's put that in a separate PR and figure out how to make that work?

yuvipanda:
@ranchodeluxe and the way I'd suggest you *test* this on the hub is to use the new feature on the hub that lets you launch any image. So launch your image with the post-launch script inside, open a terminal in JupyterLab, and just run the command - see if that works.

ranchodeluxe:
thanks for the feedback @yuvipanda. can get to those later today

ranchodeluxe:
> @ranchodeluxe yes! There's two things to do here:
> 
>     1. The actual post start script should probably just be _in your image_, written in python. And then the hook here can just _call_ that. This way, you can test the script without having to redeploy each time - which is quite cumbersome. By keeping it in the image, you can test it locally by just launching the image and running the command yourself.
> 
>     2. I don't think the rocker image change works - The `rocker/binder` image has the extra bits required to make a rocker image work on a JupyterHub that the base images don't have. Let's put that in a separate PR and figure out how to make that work?

yeah, I was just realizing that I needed `rocker/binder:unstable` myself. Will figure that out

PR #3106, opened 2023-09-08
new singleuser image for VEDA JHs

ranchodeluxe:
This is a new singleuser image with an updated version of `stac_ipyleaflet` plugin

### DIFF
--- a/config/clusters/nasa-veda/common.values.yaml
+++ b/config/clusters/nasa-veda/common.values.yaml
@@ -60,7 +60,7 @@ basehub:
         name: public.ecr.aws/nasa-veda/nasa-veda-singleuser
         # Based off pangeo/pangeo-notebook:2023.07.05 which uses JupyterLab <4, so jupyterlab-git and dask-dashboard work
         # If updating this tag, also update it in the profile_options section below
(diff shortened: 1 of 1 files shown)


yuvipanda:
I had a quick chat with @ranchodeluxe and we're going to take a slightly different approach to this :)

consideRatio:
> The only question I'm unsure of is if the post-start user is jovyan as [I set it up over here](https://github.com/NASA-IMPACT/veda-jh-environments/blob/main/docker-images/custom/nasa-veda-singleuser/Dockerfile#L5-L6)

A k8s postStart lifecycle hook in user pods are executed as user `1000` (aka. jovyan), because that container is specified to `runAsUser: 1000` by default by the JupyterHub Helm chart via `singleuser.uid`.

consideRatio:
Marked as draft because Yuvi commented about another approach were to be taken.

ranchodeluxe:
@yuvipanda and @consideRatio: can we remove the draft designation here please. This is one needed change I need in. Everything regarding `/veda-docs` additions will be in a separate PR. I've updated the title and description :thankyou:

Issue #3111, opened 2023-09-09
Move 2i2c cluster's prometheus data from slow spinning disk to fast SSD


yuvipanda:
The 2i2c shared cluster's prometheus data is on the slowest possible pd-standard disk. We should move this to a pd-ssd with minimal data loss.

yuvipanda:
I JUST DID THIS AND OMFG EVERYTHING IS SO MUCH FASTER WTF

consideRatio:
Nice!!! :D

jmunroe:
That's great news.

Given that these Prometheus databases is a critical part of our infrastructure (especially for the shared cluster since they are used in for our billing), what's the backup strategy for these pd-balanced disks?  Is the data replicated elsewhere or backed-up outside of each cluster?

yuvipanda:
@jmunroe there isn't one! Can you open an issue so we don't lose track of it here? Thanks!

Issue #2847, opened 2023-07-20
Investigate gaps in prometheus data on the shared cluster


yuvipanda:
We use this for billing, so it needs to not have gaps in data. Investigate what the gaps are, why we have them, and try to make sure that doesn't happen.

PR #3112, opened 2023-09-09
Move 2i2c shared prometheus to pd-balanced disk

yuvipanda:
- Set up our own StorageClass for GKE clusters specifically for use with prometheus data.
- Sets retentionPolicy to 'Retain', so we don't accidentally kill the disk and lose all the data.
- Sets the disk type to 'Balanced', which is backed by SSDs and *much* faster than spinning disks. No more grafana timeouts!
- Move the existing data by manually attaching to a small VM I created, and then copying over to new PVC.
- Reduction in size, as https://github.com/2i2c-org/infrastructure/pull/3093
  drastically reduced the size of the data! We went from about
  512GB to only about 150GB after that. The size explosion has
  been solved! 512GB here still gives us enough room to grow.

Once this lands, I'll manually go through and do this for every single GCP cluster. Grafana timeouts BE GONE.

Ref https://github.com/2i2c-org/infrastructure/issues/2934 
Ref https://github.com/2i2c-org/infrastructure/issues/2717 
Ref https://github.com/2i2c-org/infrastructure/issues/2847 
Fixes https://github.com/2i2c-org/infrastructure/issues/3111

### DIFF
--- a/config/clusters/2i2c/support.values.yaml
+++ b/config/clusters/2i2c/support.values.yaml
@@ -1,11 +1,16 @@
 prometheusIngressAuthSecret:
   enabled: true
 
+prometheusStorageClass:
+  gke:
+    enabled: true
+
 prometheus:
   server:
     persistentVolume:
       # 100Gi filled up, and this is source of our billing data.
-      size: 1Ti
+      size: 512Gi
+      storageClass: balanced-rwo-retain
     ingress:
       enabled: true
       hosts:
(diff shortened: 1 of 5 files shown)


yuvipanda:
This has already been deployed to the 2i2c prometheus, and that grafana is now *snappy*

consideRatio (approved):

consideRatio:
Wieee thank you Yuvi for working this!!!

I figure defining a storaclass like this makes sense to avoid coupling too much with the k8s clusters pre-existing storageclasses.

yuvipanda:
Thanks @consideRatio. I've opened https://github.com/2i2c-org/infrastructure/issues/3119 to track moving everything else too.

PR #3125, opened 2023-09-11
deployer: make use-cluster-credentials less intrusive for gcp clusters

consideRatio:
Using `deployer use-cluster-credentials` on a GCP based cluster has previosly modified the system user state. This caused separate terminal sessions to conflict with each other.

To avoid this, we're now instead configuring an environment variable that makes us use a temporary file with the decrypted service account key.

credentials-unchanged

### DIFF
--- a/deployer/cluster.py
+++ b/deployer/cluster.py
@@ -222,26 +222,16 @@ class Cluster:
         # Else, it'll just have a `zone` key set. Let's respect either.
         location = config.get("zone", config.get("region"))
         cluster = config["cluster"]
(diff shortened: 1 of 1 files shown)


consideRatio:
@yuvipanda I'm working on trying to get GCP various clsuter off standard disks and would benefit from having this PR merged.

Do you have time to have a look?

yuvipanda (approved):

consideRatio:
Thank you @yuvipanda !!!!

Issue #3080, opened 2023-09-05
[Request deployment] New Hub: HHMI


colliand:
### The GitHub handle of the community representative

@colliand and pending

### Hub important dates

- **Target Start Date:** 2023-09-11
- **Required Start Data:** 2023-09-19

### Hub Authentication Type

GitHub (e.g., @mygithubhandle)

### First Hub Administrators

- Jim Colliander, @colliand

### [GitHub Auth only] How would you like to manage your users?

pending

### [GitHub Teams Auth only] Profile restriction based on team membership

pending

### Hub logo image URL

https://www.hhmi.org/sites/default/files/Logos/JPG/HHMI-vertical-signature-color.jpg

### Hub logo website URL

https://www.hhmi.org/

### Hub user image GitHub repository

pending

### Hub user image tag and name

pending

### Extra features you would like to enable

- [X] Dedicated Kubernetes cluster
- [x] Scalable Dask Cluster

### (Optional) Preferred cloud provider

None

### (Optional) Billing and Cloud account

None

### Other relevant information to the features above

This hub will be used for an emerging pilot with the Howard Hughes Medical Institute (HHMI). This engagement is different from 2i2c's prior partnerships. The project is led by central administration instead of by a Principal Investigator. As a result, we don't have input from a scientist requesting specific tools in the image, etc. 

The first big user of this hub is likely to be a neuroscientist. Therefore, the software image used on the Neurohackademy or the Allen Institute hubs are likely the best first choices. As the pilot evolves, I expect that the hub will be used by a variety of medical researchers. The hub should be configured to eventually support multiple hubs with different hardware configurations. 

### Tasks to deploy the hub

- [ ] 1. Deploy information filled in above
- [ ] 2. Engineer who will deploy the hub is assigned
- [ ] 3. If using GitHub Orgs/Teams Auth, Engineer is given Owner rights to the org to set this up.
- [ ] 4. Initial Hub deployment PR <link to PR>
- [ ] 5. Administrators able to log on -> Hub now in steady-state

colliand:
Next week, @colliand will travel to SF to meet with the [Loren Frank Group](https://franklab.ucsf.edu/) on Tuesday 2023-09-12 at UCSF. Ideally, the hub will be set up for demonstration on that afternoon.

colliand:
[...]

PR #3136, opened 2023-09-14
HHMI hub

GeorgianaElena:
For https://github.com/2i2c-org/infrastructure/issues/3080

This currently only has a staging hub, deployed at https://staging.hhmi.2i2c.cloud/.
Need to add a prod hub for it

### DIFF
--- a/config/clusters/hhmi/common.values.yaml
+++ b/config/clusters/hhmi/common.values.yaml
@@ -29,7 +29,6 @@ basehub:
             name: TODO
             url: https://www.hhmi.org/
     singleuser:
-      # User image: https://quay.io/repository/arokem/nh2023?tab=tags
       image:
         name: pangeo/pangeo-notebook
         tag: "2023.09.11"
--- a/config/clusters/hhmi/common.values.yaml
+++ b/config/clusters/hhmi/common.values.yaml
@@ -45,20 +45,12 @@ basehub:
         JupyterHub:
           authenticator_class: cilogon
         CILogonOAuthenticator:
+          scope:
+            - "profile"
+            - "email"
+          shown_idps:
+            - http://github.com/login/oauth/authorize
           allowed_idps:
             http://github.com/login/oauth/authorize:
               username_derivation:
                 username_claim: "preferred_username"
-        OAuthenticator:
-          # WARNING: Don't use allow_existing_users with config to allow an
-          #          externally managed group of users, such as
-          #          GitHubOAuthenticator.allowed_organizations, as it breaks a
-          #          common expectations for an admin user.
-          #
-          #          The broken expectation is that removing a user from the
-          #          externally managed group implies that the user won't have
-          #          access any more. In practice the user will still have
-          #          access if it had logged in once before, as it then exists
-          #          in JupyterHub's database of users.
-          #
-          allow_existing_users: True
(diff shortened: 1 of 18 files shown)


consideRatio (commented):
This LGTM and consider this approved if you provide oauth 15.1 functional config now that the oauth16 upgrade was reverted. Work towards oauth16 can be found in https://github.com/2i2c-org/infrastructure/pull/3144

config/clusters/hhmi/common.values.yaml
```diff
@@ -0,0 +1,66 @@
...
+      # User image: https://quay.io/repository/arokem/nh2023?tab=tags
+      image:
+        name: pangeo/pangeo-notebook
+        tag: "2023.09.11"
+      nodeSelector:
+        node.kubernetes.io/instance-type: n2-highmem-16
+      cpu:
+        guarantee: 0.5
+        # We're on n1-highmem-16 machines
+        limit: 14
```
yuvipanda, line 41: This probably needs a G after

GeorgianaElena: @yuvipanda, I think it's ok because this is the cpu limit. I just double-checked and the memory guarantee and limit have the G.

GeorgianaElena:
Thanks @consideRatio and @yuvipanda for the review. I've now also added (and deployed) a prod hub too. Also updated the auth config for oauthenticator < 16

> This LGTM and consider this approved if you provide oauth 15.1 functional config

Per Erik's comment above, I will merge this now.

terraform/gcp/projects/daskhub-template.tfvars
```diff
@@ -8,7 +8,7 @@
 prefix     = "{{ cluster_name }}"
 project_id = "{{ project_id }}"
 
-zone   = "{{ cluster_region }}"
+zone   = "{{ cluster_region }}-b"
```
damianavila, line 11: Quick question in a high-level post review (so I might be totally wrong :wink:): do we always want to deploy to the `-b` zone? Should not get the `zone` from `foo_cluster.tfvars` file instead of "building it" from the region?
cc @GeorgianaElena

colliand:
I believe this issue can be closed since the hub has been deployed.

terraform/gcp/projects/daskhub-template.tfvars
GeorgianaElena: > Should not get the zone from foo_cluster.tfvars file instead of "building it" from the region?

@damianavila, so this template is actually used by the deployer to generate a `foo_cluster.tfvars` file. This initial file generated by the deployer is more of a starting point with some "default" basics that are meant to be edited to fulfill all of the needs for the cluster.

> do we always want to deploy to the -b zone?

Not necessarily, but it is my understanding that this is the go-to zone for us-west2 (which I believe is the default region if none is requested) if we want to be able to have GPU machines (which for a daskhub is probable) available and a vaster CPU types machines, including AMDs.

damianavila: >@damianavila, so this template is 

I knew I was missing something (well-placed "so I might be totally wrong 😉" message).
Thanks for the clarification!!

> Not necessarily,

Yep, it makes sense now that I actually realize what I am seeing 😉
Again, thanks for your patience and feedback.

PR #3139, opened 2023-09-14
Revert "Upgrade to z2jh 3.0.2 from 3.0.0-beta.1 - oauthenticator 15.1 bumped to 16.0"

consideRatio:
Reverts 2i2c-org/infrastructure#3118 - utoronto and temple reported 403 unauthroized.


yuvipanda:
fwiw this was absolutely the right call!

PR #3146, opened 2023-09-15
utoronto: r-staging, increase size of hub's disk (was full)

consideRatio:
I've applied this to staging and ensured the resize worked out.


consideRatio:
Going for a self-merge on this to avoid issues about other deploys ending up down-sizing this PVC.

yuvipanda:
@consideRatio btw, disks can't be resized *down* in kubernetes, only up. So deployment would just fail, not size the disk down. 

Self-merging this is still ok though, as it prevents mystery failed deployment!

PR #3148, opened 2023-09-15
New init /veda-docs image hash

ranchodeluxe:
### Goal
The new image script now checks out `/veda-docs` as is instead of using the `gitpuller` arg `repo_dir` to try and rename to something more meaning such as `/veda-docs-examples`. Using a new `repo_dir` arg doesn't jive well with generated nbgitpuller links (even when using `targetpath`) b/c they are tightly coupled with `urlpath` and can't resolve properly. 

Even if I missed something above in my investigation I'm fine with this change b/c I don't want to modify this anymore 😏 

Let's push to staging and let me test and then I can tell you when I want to promote to production please and thankyou 👍

### DIFF
--- a/config/clusters/nasa-veda/common.values.yaml
+++ b/config/clusters/nasa-veda/common.values.yaml
@@ -95,7 +95,7 @@ basehub:
                   default: true
                   slug: pangeo
                   kubespawner_override:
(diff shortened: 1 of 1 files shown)


yuvipanda:
Just deployed!

ranchodeluxe:
Thanks @yuvipanda . Looking groovy to moi. Let's promote to production and call it doneso when you get a chance please

Issue #3130, opened 2023-09-13
2i2c/ucmerced have 34 user pods running for 6-7 days


consideRatio:
I figure the jupyterhub-idle-culling has failed for some reason, perhaps because they generate network activity regularly still, or because their kernels are considered active still?

I'm not looking into this right now, already working the "the cluster is slow" incident in https://github.com/2i2c-org/infrastructure/issues/2947

Related to #3042, where culling is configured to max 7 days as a precaution.

---

Here is one pod, with the latest log entry from several days back.

```
[I 2023-09-07 03:53:53.735 SingleUserLabApp kernelmanager:321] Starting buffering for 9ae455a9-c907-46a7-b914-f72475f8c58d:aabf4f6a4aa249ea88f0d5ff2b0ed2de
[W 2023-09-07 04:41:24.819 SingleUserLabApp kernelmanager:637] Culling 'idle' kernel 'python3' (7537ac89-0bcd-480b-9c3a-456ebc7b8c04) with 0 connections due to 3791 seconds of inactivity.
[I 2023-09-07 04:41:24.823 SingleUserLabApp multikernelmanager:499] Kernel shutdown: 7537ac89-0bcd-480b-9c3a-456ebc7b8c04
[W 2023-09-07 04:51:24.820 SingleUserLabApp kernelmanager:637] Culling 'idle' kernel 'python3' (9ae455a9-c907-46a7-b914-f72475f8c58d) with 0 connections due to 3731 seconds of inactivity.
[I 2023-09-07 04:51:24.827 SingleUserLabApp multikernelmanager:499] Kernel shutdown: 9ae455a9-c907-46a7-b914-f72475f8c58d
```

Here are some hub logs:

```
[W 2023-09-13 15:10:14.237 JupyterHub base:1436] Failing suspected API request to not-running server: /hub/user/gnebel@ucmerced.edu/api/kernels/87460ade-a29e-4419-8517-5ef1146adec3
[W 2023-09-13 15:10:14.238 JupyterHub log:191] 424 GET /hub/user/gnebel@ucmerced.edu/api/kernels/87460ade-a29e-4419-8517-5ef1146adec3?1694617812678 (gnebel@ucmerced.edu@10.128.0.98) 10.03ms
[...]

PR #3149, opened 2023-09-15
[DO NOT MERGE] Test if updated kubespawner resolves culler failure

sgibson91:
ref https://github.com/2i2c-org/infrastructure/issues/3130

I have opened this for visibility into what I have done, but shouldn't be merged. If updating kubespawner fixes the issue, then we should aim for a new release and update it in the main hub image, not maintain another image.

We are hoping that with the deployment of this image, some long-lived pods will be successfully culled.

### DIFF
--- a/config/clusters/2i2c/ucmerced.values.yaml
+++ b/config/clusters/2i2c/ucmerced.values.yaml
@@ -30,6 +30,9 @@ jupyterhub:
       name: quay.io/2i2c/2i2c-hubs-image
       tag: "14107b8a85fb"
   hub:
+    image:
+      name: quay.io/2i2c/culler-failure
+      tag: "0.0.1-0.dev.git.7009.h8f624235"
     config:
       Authenticator:
         admin_users:
--- a/helm-charts/chartpress.yaml
+++ b/helm-charts/chartpress.yaml
@@ -18,3 +18,9 @@ charts:
           REQUIREMENTS_FILE: "dynamic-image-building-requirements.txt"
         contextPath: "images/hub"
         dockerfilePath: "images/hub/Dockerfile"
+      culler-failure:
+        imageName: quay.io/2i2c/culler-failure
+        buildArgs:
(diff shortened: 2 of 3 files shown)


config/clusters/2i2c/ucmerced.values.yaml
```diff
@@ -27,8 +27,8 @@ jupyterhub:
           url: http://www.ucmerced.edu/
   singleuser:
     image:
-      name: quay.io/2i2c/2i2c-hubs-image
-      tag: "14107b8a85fb"
+      name: quay.io/2i2c/culler-failure
```
yuvipanda, line 30: This should actually be setting the image under `hub.image` rather than `singleuser.image`

yuvipanda (commented):
I've undeployed this by deploying from master temporarily! We can try again later

NessieCanCode:
Hi @sgibson91, I'm an admin at UC Merced, and I think this Test image is causing disruption to our users, which I would like to bring to your attention. https://2i2c.freshdesk.com/support/tickets/980 I was able to reproduce the issue as well.

yuvipanda:
Hey @NessieCanCode! I've undeployed this change, and things should be working now! I also replied on the ticket :)

yuvipanda:
@NessieCanCode also, welcome to our github :)

config/clusters/2i2c/ucmerced.values.yaml
yuvipanda: I opened https://github.com/2i2c-org/infrastructure/issues/3150 about having a separate staging hub for ucmerced perhaps, given they're still on the old user image and there may be other things we need to test.

sgibson91: Woof, is it too late in the game to change that image name? I was definitely lead to believe I was changing the right thing since it was the "hubs-image"

sgibson91:
I reverted this and redeployed from master again

Issue #3104, opened 2023-09-08
Wire up the prototype binderhub UI to the deployed binderhub service!


yuvipanda:
The goal here is to get this to a demoable state

```[tasklist]
- [x] Wire up prototype UI to hook into the binderhub service https://github.com/2i2c-org/infrastructure/pull/3145
- [x] Read returned built image name from binderhub service, hook it up to `unlisted_choice` https://github.com/2i2c-org/infrastructure/pull/3151
- [ ] Wire up appropriate authentication between jupyterhub and binderhub service
- [x] Wire up the built image to unlisted_choice so the built image is launched https://github.com/2i2c-org/infrastructure/pull/3151
- [x] Give @arnim access to the demo hub https://github.com/2i2c-org/infrastructure/pull/3152
- [x] Cleanup UX of imagebuilding-demo: https://github.com/2i2c-org/infrastructure/pull/3155
```

yuvipanda:
This is done!

PR #3151, opened 2023-09-15
Launch built images in imagebuilder-demo!

yuvipanda:
- Provides options for resource allocation that the user can select, showing that you *can* select what kinda resource you want while building image
- Update image choices to be more reflective of what we would have in a real hub
- Brings in a lot of changes from github.com/yuvipanda/prototype-kubespawner-dynamic-building-ui to support the
   following:
   - Built images are now *actually launched*! Yay!
   - The big piece of text proclaiming that this is a prototype that doesn't work is gone, as it does work
   - The build logs are collapsible
   - There's a poorly discsoverable message saying 'build is done, click to launch'.

Ref https://github.com/2i2c-org/infrastructure/issues/3104

Here's a demo!

https://github.com/2i2c-org/infrastructure/assets/30430/64c4b8ab-1669-4b59-8b6d-979e2fb82298

(dropdown slightly tweaked with suggestions from @choldgraf)
image

So all the options in the dropdown work *except* for wiring up the built image to be actually launched.

### DIFF
--- a/config/clusters/2i2c/imagebuilding-demo.values.yaml
+++ b/config/clusters/2i2c/imagebuilding-demo.values.yaml
@@ -121,7 +121,7 @@ jupyterhub:
         url: http://imagebuilding-demo-binderhub-service:8090
     image:
       name: quay.io/2i2c/dynamic-image-building-experiment
-      tag: "0.0.1-0.dev.git.7001.hf02ed7a1"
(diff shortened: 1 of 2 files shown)


GeorgianaElena (approved):
Yuhuuuuuu 🚀 🎉

PR #3155, opened 2023-09-19
Cleanup UX of imagebuilding-demo

yuvipanda:
- Get rid of the ancient bootstrap3 code for doing layouts, use moden native
   CSS grid instead. Makes things *much* easier.
- First real feature for profiles that isn't supported in the simple jinja2 template
  shipped with kubespawner - use https://react-select.com/ so we can actually
  have a useful *description* with each option the user can select. This is
  specified by admin when setting up profileList, and is ignored by the default
  profileList renderer. Very useful in setting context!
- UX is generally cleaner and melds in together nicer, vs being more disjointed
   before. 
- The ability to explicitly specify the GitHub repos' ref has been removed, and
   will be re-introduced in the near future. Same for ability to hide the build log,
   although that is less of an issue now that the logs are smaller.
- Brings in https://github.com/jupyterhub/kubespawner/pull/787, so resource
   allocations are oredered.
-

### DIFF
--- a/config/clusters/2i2c/imagebuilding-demo.values.yaml
+++ b/config/clusters/2i2c/imagebuilding-demo.values.yaml
@@ -121,7 +121,7 @@ jupyterhub:
         url: http://imagebuilding-demo-binderhub-service:8090
     image:
       name: quay.io/2i2c/dynamic-image-building-experiment
-      tag: "0.0.1-0.dev.git.7029.h44cd314e"
+      tag: "0.0.1-0.dev.git.7058.h013054f7"
     config:
       JupyterHub:
         authenticator_class: github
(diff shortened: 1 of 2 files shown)


GeorgianaElena (approved):
I like this improved UX a lot!

arnim:
This looks really nice 🎉 
Do we need a delete option for images as well?

yuvipanda:
@arnim to cleanup unused images in the registry? or somewhere else?

arnim:
To cleanup unused images.

arnim:
And only in the list of images the user has imported. Not sure if this makes sense. 

Would probably not delete images in the registry.

yuvipanda:
@arnim ah, so the list of images you see are coming from what the *admins* have set in config, not from the user level. So removing those is done by editing z2jh config. 

I do agree that separately, it would be useful to remember the images the user has built and allow them to rebuild / reuse. That functionality currently doesn't exist here though.

arnim:
In the old pBHub setup we also had the option to configure a "hallo world" image/repo, but did later not distinguish between admin configured images and user imported images. 

Not sure if something in this direction makes sense here as well.

Issue #3158, opened 2023-09-20
Give temple their own nodepool


yuvipanda:
From https://2i2c.freshdesk.com/a/tickets/983, temple is reporting server crashes as their users all come online at the same time. You can see this in the bump in their server startup times:

<img width="967" alt="image" src="https://github.com/2i2c-org/infrastructure/assets/30430/7badb108-8f57-46e1-b4cb-0d66f70a72a1">

Given they get to 100+ users, and they all come on suddenly enough, I think it makes sense to put them on their own nodepool. We now also can accurately attribute costs *per nodepool*, so this makes cost recovery easier too. This will also relieve pressure on the other hubs that are on the same shared cluster.

<img width="957" alt="image" src="https://github.com/2i2c-org/infrastructure/assets/30430/baac338c-6106-475f-9232-71f160cecb55">

yuvipanda:
For measuring the node, I saw that everyone has the basic default we set - 256M of guarantee and 1G of limit. Looking at actual memory usage:

<img width="1905" alt="image" src="https://github.com/2i2c-org/infrastructure/assets/30430/b69c84eb-33dc-4055-a517-03223a62fb79">

You can see that most users are roughly under about 512M. This is super rough, but based on this I choose n2-highmem-8. This has about 58ish GB of RAM. The scheduler will fit in 100 users here (as that's about the per-node pod limit), and overall I think that'll work out ok.

PR #3159, opened 2023-09-20
Move temple university to its own nodepool

yuvipanda:
Fixes https://github.com/2i2c-org/infrastructure/issues/3158


yuvipanda:
Self-merging because their class starts morning california time, and I don't want them on the common pool.

PR #3160, opened 2023-09-20
Increase size of prometheus disk for 2i2c-aws-us

yuvipanda:
Was full :(

### DIFF
--- a/config/clusters/2i2c-aws-us/support.values.yaml
+++ b/config/clusters/2i2c-aws-us/support.values.yaml
@@ -27,9 +27,7 @@ prometheus:
           hosts:
             - prometheus.aws.2i2c.cloud
     persistentVolume:
-      # Found to be out of space with at 100Gi disk at 2023-07-31, when it was
-      # increased to 250Gi.
-      size: 250Gi
+      size: 500Gi
 
 cluster-autoscaler:
   enabled: true


yuvipanda:
Already manually deployed

config/clusters/2i2c-aws-us/support.values.yaml
```diff
@@ -27,9 +27,7 @@ prometheus:
           hosts:
             - prometheus.aws.2i2c.cloud
     persistentVolume:
-      # Found to be out of space with at 100Gi disk at 2023-07-31, when it was
-      # increased to 250Gi.
-      size: 250Gi
```
consideRatio, line 32: The growth rate seems like ~100GB/month, so the increase of 250GB gives us ~2.5 month of time then I figure.

consideRatio (approved):

GeorgianaElena:
@consideRatio, let's also merge this to make sure it doesn't get overwritten?

consideRatio:
Good call @GeorgianaElena!

config/clusters/2i2c-aws-us/support.values.yaml
yuvipanda: @consideRatio I think that was mostly due to the small blocksize we had running, and I think this cluster never actually came back from that. If you look at the disk free space graph now:

image

Usage actually *shrunk* by about 100GB after prometheus came back and compacted all those blocks!

PR #3165, opened 2023-09-21
Specify tags & pullPolicy for alpine/git: image

yuvipanda:
Without this, the :latest tag was being used, and we were checking the registry each time even if the image is already present. Makes deployments *slightly* faster

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -405,7 +405,8 @@ jupyterhub:
                     description: An IDE For R, created by the RStudio company
     initContainers:
       - name: templates-clone
-        image: alpine/git
+        image: alpine/git:2.40.1
+        imagePullPolicy: IfNotPresent
         args:
           - clone
           - --
@@ -426,7 +427,8 @@ jupyterhub:
           - name: custom-templates
             mountPath: /srv/repo
       - name: templates-ownership-fix
-        image: alpine/git
+        image: alpine/git:2.40.1
+        imagePullPolicy: IfNotPresent
         command:
           - /bin/sh
         args:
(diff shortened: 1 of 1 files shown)


sgibson91 (approved):

consideRatio (changes requested):
Don't merge this yet - restarting the hub atm could cause issues until we have a juyterhub image using the latest kubespawner with a bugfix.

- we need #3165 merged first!

damianavila:
> we need https://github.com/2i2c-org/infrastructure/pull/3165 merged first!

I think you mislinked some other PR, right?

consideRatio (approved):
Yepp @damianavila, I meant https://github.com/2i2c-org/infrastructure/pull/3168 - now merged!

helm-charts/basehub/values.yaml
```diff
@@ -405,7 +405,8 @@ jupyterhub:
                     description: An IDE For R, created by the RStudio company
     initContainers:
       - name: templates-clone
-        image: alpine/git
+        image: alpine/git:2.40.1
+        imagePullPolicy: IfNotPresent
```
consideRatio, line 409: I think using a tag besides `latest`, or to use `latest` and specify `IfNotPresent`, is sufficient to ensure the behavior of `IfNotPresent` I think.

In a way I like that we use `:latest` here rather than having another image that can get outdated or that we automatically merge automated version bumps without further review because we assume its fine no matter what?

To pin:
- A minus: we ~need automation to bump it over time

To not pin:
- A minus: we don't know for sure when something updates, because `:latest` combined with `IfNotPresent` makes us not know that

In this case, I think I'm leaning towards favoring not needing the complexity of updating this automatically or the alternative drawback of possibly exposing security vulernabilities that we don't patch.

> - If the image tag is :latest, the imagePullPolicy will be automatically set to Always. 
> - If the image tag isn’t :latest, the imagePullPolicy will be automatically set to IfNotPresent. 
And if you don’t set any image tag, the imagePullPolicy will be automatically set to.

consideRatio: We can go for this as it is though, I mostly wanted to ensure that that we don't spread an understanding that we would need IfNotPresent specified alongside any image tag beisdes `:latest`

consideRatio (approved):

PR #3168, opened 2023-09-22
Use main branch of kubespawner with important bugfix

consideRatio:
I think without this merged before we restart any hub pod, we may end up making jupyterhub not realizing that an old user pod was still running - making jupyterhub end up re-starting the user pod and deleting the proxy's route to the user pod.

In other words, its quite critical to not restart our hub pods until this PR is merged.

## Related

- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3229
- https://github.com/jupyterhub/kubespawner/pull/742
- https://github.com/jupyterhub/kubespawner/issues/767

## Empirical feedback

To use the main branch of kubespawner seems to have worked out well in other contexts, I suggest we go for it: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3229#issuecomment-1732637727

### DIFF
--- a/config/clusters/2i2c-aws-us/researchdelight.values.yaml
+++ b/config/clusters/2i2c-aws-us/researchdelight.values.yaml
@@ -30,7 +30,7 @@ basehub:
     hub:
       image:
         name: quay.io/2i2c/unlisted-choice-experiment
-        tag: "0.0.1-0.dev.git.6863.h406a3546"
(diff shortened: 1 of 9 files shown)


GeorgianaElena (approved):

GeorgianaElena:
Thank you @consideRatio!!!

PR #3173, opened 2023-09-27
Update UBC EOAS resource requests based on observed usage

yuvipanda:
Ref https://2i2c.freshdesk.com/a/tickets/955 for measurement and reasoning

### DIFF
--- a/config/clusters/ubc-eoas/common.values.yaml
+++ b/config/clusters/ubc-eoas/common.values.yaml
@@ -60,13 +60,13 @@ jupyterhub:
   singleuser:
     defaultUrl: /lab
     memory:
-      # Based on measurements of observed usage
-      guarantee: 1G
-      limit: 1G
+      # https://2i2c.freshdesk.com/a/tickets/955
+      # Based on request from community rep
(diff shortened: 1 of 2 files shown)


config/clusters/ubc-eoas/common.values.yaml
```diff
@@ -59,25 +59,27 @@ jupyterhub:
 
   singleuser:
     defaultUrl: /lab
+    memory:
+      # Based on measurements of observed usage
+      guarantee: 1G
+      limit: 1G
```
lheagy, line 65: could we increase the limit a touch for both courses? There are a few notebooks that they will use later in the term that might exceed 1G. For EOSC 350, 2G should be fine.

consideRatio (approved):
After adjusting to two comments, this LGTM!

config/clusters/ubc-eoas/common.values.yaml
yuvipanda: @lheagy based on comments from Henryk, these are now set to 4G as they requested that.

Issue #3126, opened 2023-09-12
[EVENT] SWOT Pangeo Grenoble


jmunroe:
### Summary

Small demo of Dask XArray for a local event in Grenoble on Sept. 27

**_Configuration changes to profile required_** (please see next comment)

### Event Info

- **Community Representative:** @auraoupa
- **Event begin:** 27-Sept-2023
  - **In your timezone:** 
- **Event end:**  27-Sept-2023
  - **In your timezone:** 
- **Active times:** 
  - **In your timezone:** 
- **Number of attendees:** 60
- [**Hub Events Calendar**](https://calendar.google.com/calendar/u/2?cid=Y19rdDg0c2g3YW5tMHNsb2NqczJzdTNqdnNvY0Bncm91cC5jYWxlbmRhci5nb29nbGUuY29t)

### Hub info

- **Hub URL**: https://meom-ige.2i2c.cloud
- **Hub decommisioned after event?**:  no

### Task List

**Before the event**

- [ ] Dates confirmed with the community representative and added to Hub Events Calendar.
- [ ] Quotas from the cloud provider are high-enough to handle expected usage.
- [ ] **One week before event** Hub is running.
- [ ] Confirm with Community Representative that their workflows function as expected.
  - <details>
    <summary>👉Template message to send to community representative</summary>
    
    ```
    Hey {{ COMMUNITY REPRESENTATIVE }}, the date of your event is getting close!
    
    Could you please confirm that your hub environment is ready-to-go, and matches your hub's infrastructure setup, by ensuring the following things:
    - [ ] Confirm that the "Event Info" above is correct
    - [ ] On your hub: log-in and authentication works as-expected
    - [ ] `nbgitpuller` links you intend to use resolve properly
    - [ ] Your notebooks and content run as-expected
    ```
    
    </details>  
- [ ] **1 day before event**, either a separate nodegroup is provisioned for the event or the cluster is scaled up.

**During and after event**

- [ ] Confirm event is finished.
- [ ] Nodegroup created for the hub is decommissioned / cluster is scaled down.
- [ ] Hub decommissioned (if needed).
- [ ] Debrief with community representative.
  - <details>
    <summary>👉Template debrief to send to community representative</summary>
      
     ```
     Hey {{ COMMUNITY REPRESENTATIVE }}, your event appears to be over 🎉
     
[...]

PR #3179, opened 2023-09-27
meom-ige: restore previous profile_list after event

consideRatio:
We are to remove access for a group of users tomorrow Thursday, and I'm also commenting out the profile list entry.

Since this hub doesn't have node sharing yet in the other profile list option, I think its relevant to retain this profile list entry as a comment to make it quicker and easier to setup another event in the future.

### Related
- https://github.com/2i2c-org/infrastructure/issues/3126
- https://2i2c.freshdesk.com/a/tickets/958 and #grenooble


consideRatio:
To not merge this is a security matter, so I went for a merge without waiting for approval!

Issue #623, opened 2021-08-23
Enable EFS Backups on AWS


yuvipanda:
We use EFS for home directory storage on AWS. It offers a [built in backup solution](https://docs.aws.amazon.com/efs/latest/ug/awsbackup.html), which we must configure and enable for all our hubs.

yuvipanda:
I've applied this on everything except the NASA SMCE managed projects (GHG and VEDA), and I'll do that tomorrow.

PR #3182, opened 2023-09-28
Enable automatic backups for homedirs on AWS

yuvipanda:
"Automatic backups" are documented by AWS in
https://docs.aws.amazon.com/efs/latest/ug/awsbackup.html#automatic-backups, and are done *daily* with a 35 day retention period. We could customize this in the future, but this is a great start.

Fixes https://github.com/2i2c-org/infrastructure/issues/623


yuvipanda:
Also did https://github.com/2i2c-org/infrastructure/pull/3183 while I was here, so I can apply these together!

GeorgianaElena (approved):
THANK you @yuvipanda!!!

PR #3183, opened 2023-09-28
Enable AWS Infrequent access to save money on EFS

yuvipanda:
Transition files to a slower, cheaper backing medium 90 days after they were last *accessed*. They will be transferred back to regular backing medium immediately on access. This saves a *lot* of money with barely perceptible user performance hit. 90 days is the longest config, and we can start here to be conservative. IA (Infrequent Access) is documented in https://aws.amazon.com/efs/features/infrequent-access/

### DIFF
--- a/terraform/aws/efs.tf
+++ b/terraform/aws/efs.tf
@@ -46,6 +46,16 @@ resource "aws_efs_file_system" "homedirs" {
     Name = "hub-homedirs"
   }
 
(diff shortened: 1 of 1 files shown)


GeorgianaElena (approved):

yuvipanda:
I've applied this to everything except the NASA SMCE managed stuff, which I will do tomorrow.

PR #3184, opened 2023-09-28
Split EFS lifecycle into two blocks

yuvipanda:
Otherwise it succeeds on plan but fails on apply

https://github.com/hashicorp/terraform-provider-aws/issues/21862#issuecomment-1007115545

Follow-up to https://github.com/2i2c-org/infrastructure/pull/3183

### DIFF
--- a/terraform/aws/efs.tf
+++ b/terraform/aws/efs.tf
@@ -46,15 +46,20 @@ resource "aws_efs_file_system" "homedirs" {
     Name = "hub-homedirs"
   }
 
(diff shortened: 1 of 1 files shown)


yuvipanda:
terraform apply works with this now

yuvipanda:
image

yuvipanda:
Self-merging as a syntactic fixup

GeorgianaElena (approved):
🚀

Issue #3181, opened 2023-09-27
Write a command to help transfer home directories from one user to another


yuvipanda:
When users get renamed, it is important to be able to copy their files from their old to new home directories seamlessly. We can manually do this with `exec-homes-shell` now, but this should be a command of its own to reduce toil for the person doing it.

PR #3187, opened 2023-09-28
Add command to transfer files from one homedir to another

GeorgianaElena:
For https://github.com/2i2c-org/infrastructure/issues/3181


deployer/debug.py
```diff
@@ -120,18 +121,8 @@ def user_logs(
...
-    hub_name: str = typer.Argument(..., help="Name of hub to operate on"),
-):
-    """
-    Pop an interactive shell with the home directories of the given hub mounted on /home
-    """
-    # Name pod to include hub name so it is displayed as part of the default prompt
-    # This makes sure we don't end up deleting the wrong thing
-    pod_name = f"{cluster_name}-{hub_name}-shell"
-    pod = {
+def generate_pod_definition_with_home_mount(pod_name):
```
yuvipanda, line 124: I think we can't share pod definition, because the `cp` should be run in a container that's running as uid 1000, while exec-homes-shell probably as `root`. Otherwise the copies end up as owned by root and cause problems - see https://github.com/yuvipanda/prometheus-dirsize-exporter/issues/5.

deployer/debug.py
```diff
@@ -259,5 +264,128 @@ def start_docker_proxy(
...
+    # under a directory located at `/home/{destination_user}/{source_user}-homedir``
+    copy_cmd = [
+        "kubectl",
+        "-n",
+        hub_name,
+        "exec",
+        pod_name,
+        "--",
+        "cp",
+        "-r",
```
yuvipanda, line 345: Based on https://man7.org/linux/man-pages/man1/cp.1.html, I suggest adding the following options:

- `--archive`, to preserve permissions and stuff
- `--no-clobber`, so we don't accidentally delete files

deployer/debug.py
```diff
@@ -259,5 +264,128 @@ def start_docker_proxy(
...
+    destination_user: str = typer.Option(
+        prompt=True,
+        confirmation_prompt=True,
+        help="Name of destination user in who's home dir we will be moving the files",
+    ),
+):
+    """
+    Copy all the files from a user's old home dir to the new home directory
+    """
+    # Name pod to include hub name so it is displayed as part of the default prompt
```
yuvipanda, line 285: The engineer doesn't see this as the cp isn't done interactively.

deployer/debug.py
```diff
@@ -259,5 +264,128 @@ def start_docker_proxy(
...
+    create_pod_cmd = [
+        "kubectl",
+        "-n",
+        hub_name,
+        "run",
+        pod_name,
+        "--overrides",
+        json.dumps(pod),
+        "--image",
+        # Use ubuntu image so we get GNU rm and other tools
```
yuvipanda, line 300: Let's reference GNU `cp` here instead.

yuvipanda (changes requested):
Some suggested changes but otherwise good!

deployer/debug.py
GeorgianaElena: 💯 yes!!!! Thank you!

deployer/debug.py
GeorgianaElena: Yeah, definitely copied as root....

GeorgianaElena:
Thank you very much for the feedback @yuvipanda! I've now updated the PR based on it. LMK what you think

yuvipanda (approved):
Very minor code suggestions, feel free to merge after accepting them (if they feel acceptable, of course)!

Thanks for helping automate this, @GeorgianaElena!

Not part of this PR, but can you add a small `how to` document on doing this after, so it is more discoverable?

deployer/debug.py
```diff
@@ -259,5 +260,167 @@ def start_docker_proxy(
...
+            "securityContext": {"runAsUser": 1000, "runAsGroup": 1000},
+            "volumes": [
+                # This PVC is created by basehub
+                {"name": "home", "persistentVolumeClaim": {"claimName": "home-nfs"}}
+            ],
+            "containers": [
+                {
+                    "name": pod_name,
+                    # Use ubuntu image so we get better GNU cp
+                    "image": "ubuntu:jammy",
```
consideRatio, line 300: Its easier to keep track of the ubuntu versions if they are numerical compared to their names I think.

```suggestion
                    "image": "ubuntu:22.04",
```

deployer/debug.py
```diff
@@ -259,5 +260,167 @@ def start_docker_proxy(
...
+        "-n",
+        hub_name,
+        "run",
+        pod_name,
+        "--overrides",
+        json.dumps(pod),
+        "--image",
+        # Use ubuntu image so we get GNU cp and other tools
+        # Should match what we have in our pod definition
+        "ubuntu:jammy",
```
consideRatio, line 328:
```suggestion
        "ubuntu:22.04",
```

consideRatio: This was specifying `image` directly, but its also part of the pod override. Its nice to only use one so that we don't make mistakes updating only one but not the other in the future.

deployer/debug.py
```diff
@@ -259,5 +260,167 @@ def start_docker_proxy(
         subprocess.check_call(cmd)
 
 
+@app.command()
+def transfer_old_home_dir_to_new_location(
```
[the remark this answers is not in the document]
consideRatio: Transfer implies `mv` to me, or `copy` + `rm`, but its a plain `copy` operation. I'd appreciate this named with `copy` over `transfer`.

There is an assumption in the command of copying files from one homedir _into a subdir of_ another another users homedir, this also isn't obvious from either naming proposals either.

With this considered, what naming do I think makes sense? I propose `copy_homedir_into_another` as that seems quite short and self descriptive.

consideRatio:
## username vs homedir name clarification

I'm using this command atm for https://2i2c.freshdesk.com/a/tickets/1006, but when doing this I end up uncertain on how to use it related to the difference between usernames and home dir names.

The command sais `--source-user` and `--destination-user`, and the description is: `help="Name of source user who's home dir files we are transferring",`

This indicates that the jupyterhub usernames should be used, but then these will be escaped to relate to home dirs. For example the user `erik@2i2c.org` would translate to `erik-402i2c-2eorg` I think.

I'm not sure on the CLI to provide, but it must be clarified. Is specifying homedir names better (current behavior) than going for usernames that we escape first, as it avoids compelxity? Maybe? If we go for that, what we need to do is to adjust the CLI to make its working with homedir names, not usernames.

## (optional bonus) Info and validation of available homedir names

I think it would be very useful for a user to be provided help about the available homedirs before asking the user to enter info about. My desired CLI experience would be to optionally be able to pre-specify homedir source and destination via flags, but if they aren't specified be able to first be informed about whats available and then input a homedir source and/or destination.

I'd also like to validate the source/destination based on info about if they exist or not.
- If a source homedir that doesn't exist is chosen, then we should fail before trying to copy to get a clear error message etc
- If a destination homedir name exists or not is helpful to know is validated
  - if a destination homedir name doesn't exist, then what? Should we create it first? I think for now we should just fail instead.
  - if a destination homedir name does exist, I think we should do the check if the final destination already exists or not first as well, and fail if it already exists for now - a copy operation has already been made.

deployer/debug.py
```diff
@@ -259,5 +260,167 @@ def start_docker_proxy(
...
+    # Command that lists the contents of the source and destination home dirs
+    ls_dirs_cmd = [
+        "kubectl",
+        "-n",
+        hub_name,
+        "exec",
+        pod_name,
+        "--",
+        "ls",
+        "-lat",
```
consideRatio, line 355: Do we prefer seeing the exact bytes, or do we prefer improved readability?

With `-h` (`--human-readable`), the difference is like below, where its quite impotant for larger files where its hard to know if something is  100GB, or 10GB, 1GB or similar because of the large amount of numbers. The downside is that you can't know if its changed with a single byte, but we have the timestamp to indicate something about this.

I think I suggest we go for `-h`.

```
# with -h
-rw-rw-r--  1 erik erik 1,9K sep 25 16:56 .pre-commit-config.yaml

# without -h
-rw-rw-r--  1 erik erik 1865 sep 25 16:56 .pre-commit-config.yaml
```

```suggestion
        "-lath",
```

GeorgianaElena: Agree! Thank you @consideRatio

deployer/debug.py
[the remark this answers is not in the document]
GeorgianaElena: Agree that the original name I gave it was long and confusing. Thanks for the great suggestions @consideRatio and @yuvipanda. I will go with @consideRatio's proposed name which uses the "copy" instead of the "transfer" verb as that is more descriptive of what the command does.

@yuvipanda, please let me know if you disagree and we can all brainstorm to get to a common proposal

GeorgianaElena:
@consideRatio, I did the following updates based on your feedback:

- [x] created a global var UBUNTU_IMAGE="ubuntu:22.04" that I used everywhere
- [x] clarified that the command expects the escaped homedir name instead of the username
- [x] in terms of workflow you can now run `deployer copy-homedir-into-another cluster hub` (the `--source-homedir`  and ` --destination-homedir` flags are now optional) and the following will happen:
 1. a pod with `/home` mounted will be created and waited to be ready
 2. the script will notice that either the source homedir or dest homedir were not provided, so it will instead list the contents of `/home`
 3. it will then ask for confirmation from the user if they now wish to provide the name of the source and dest homedir
 4. if they don't, the pod is deleted and the scripts exists 
 5. if they do, they will be presented with prompts where they can input the names of the dirs
 6. if the names are left empty again, then 4 happens
 7. if we have names, then list the contents of these dirs
 8. notify that the copy operation will start and ask for confirmation to continue with it
 9. if confirm, the copy is performed
 10. then the contents of the source and destination directories are again listed
 11. the pod is deleted
 
 And in screenshots:
 
Screenshot 2023-10-03 at 17 18 08
(...)

Screenshot 2023-10-03 at 17 18 23
(...)

Screenshot 2023-10-03 at 17 18 36
(...)

Screenshot 2023-10-03 at 17 18 49

Does this look ok to you?

consideRatio (approved):
It looks amazing @GeorgianaElena, thanks for the care and effort into this!! I used the command successfully when resolving https://2i2c.freshdesk.com/a/tickets/1006 btw!

GeorgianaElena:
Yaaaaay 🎉 Thank you for the great feedback.
No more shooting myself in the foot with a `cp -r` again 🤦🏼‍♀️ 😅 

I think I will merge this now to get it off the table, but will:
- keep an eye on @yuvipanda's feedback about the new command name
- open an issue about documenting this command

consideRatio:
> open an issue about documenting this command

:+1: about referencing the command to raise awareness about it being available, but I think its good if we rely on the `deployer copy-homedir-into-another --help` text to avoid duplicating docs on how to use it and how it works to make future maintenance easier.

image

PR #3192, opened 2023-09-29
basehub/daskhub: cleanup orphaned pods from z2jh 3.0/kubespawner 6.0

consideRatio:
I've tested this cleanup logic across all hubs in the 2i2c and utoronto cluster, and it seems to work out fine. It is meant to be applied, and then removed once its run once.

I've also documented this strategy for any z2jh users via https://discourse.jupyter.org/t/how-to-cleanup-orphaned-user-pods-after-bug-in-z2jh-3-0-and-kubespawner-6-0/21677.

- Closes #3180

PR #3193, opened 2023-09-29
Remove the cleanup logic that has now cleaned up orphaned pods

consideRatio:
This reverts commit 39e6b2c985540ecfe7203833fdbdbf8b8e9e4c39, the main part of #3192. This was something intented to be reverted once we had let it execute in all hubs - it was just a one off cleanup operation.


yuvipanda (approved):
This looks good, but GHG center is in the middle of a demo right now, so let's hold off on merging it until after.

I'm happy to merge this in a few hours once that's over!

consideRatio:
Thanks, your most welcome to merge it!
The PR adding it had a 100% success rate so I think it could work out without any issues to merge this as well.

Issue #3194, opened 2023-09-29
Notes on outage in nasa-ghg


consideRatio:
Yuvi restarted the hub pod, but looking at the other pods, they all had their containers restart `Fri, 29 Sep 2023 17:43:27 +0200` - what happened?

```
api-prod-dask-gateway-84ddd5b8-5mgpv            1/1     Running   1 (3h36m ago)   59d
controller-prod-dask-gateway-554bb6c787-vqmkl   1/1     Running   1 (3h36m ago)   59d
hub-865d686bb4-hkgxp                            2/2     Running   0               70m
proxy-67dc5db685-r64w6                          1/1     Running   1 (3h36m ago)   15d
shared-dirsize-metrics-757595f7b4-f8slg         1/1     Running   1 (3h36m ago)   10d
shared-volume-metrics-5b9bc49c59-nxkzx          1/1     Running   1 (3h36m ago)   59d
traefik-prod-dask-gateway-659d8b4598-mlx7q      1/1     Running   1 (3h36m ago)   59d
user-scheduler-6d67457c96-9vvgc                 1/1     Running   1 (3h36m ago)   15d
user-scheduler-6d67457c96-sb4mh                 1/1     Running   1 (3h36m ago)   15d
```

Looking at nodes I see that a single core node has been running for a long time, and there is one 112m old user node as well.

```
kubectl get node                                                  
NAME                                           STATUS   ROLES    AGE    VERSION
ip-192-168-31-131.us-west-2.compute.internal   Ready    <none>   112m   v1.27.3-eks-a5565ad
ip-192-168-8-94.us-west-2.compute.internal     Ready    <none>   59d    v1.27.3-eks-a5565ad
```

Looking with `kubectl describe pod` to see the "Last state" on these pods, we see this for several (all?) restarted pods:

```
    Last State:     Terminated
      Reason:       Unknown
      Exit Code:    255
      Started:      Tue, 01 Aug 2023 06:20:55 +0200
      Finished:     Fri, 29 Sep 2023 17:43:27 +0200
```

It seems that something major crashed on the node for some reason, all pods including kube-system restarted - but it recovered to some degree at least.

yuvipanda:
The hub pod didn't recover from the crash. Looking at describe, I saw:

```
[...]

Issue #1695, opened 2022-09-09
Templates sync don't work


GeorgianaElena:
### Context

The [`templates-sync` container](https://github.com/2i2c-org/infrastructure/blob/7aa7442ab67a0dd040789ccb823d4c83f659be2d/deployer/hub.py#L128-L146) is currently not working, which means that any change to https://github.com/2i2c-org/default-hub-homepage won't be picked up "real-time', without a re-deploy of the hub. 

The container logs show the following errors:

```
fatal: detected dubious ownership in repository at '/srv/repo'
To add an exception for this directory, call:

        git config --global --add safe.directory /srv/repo
fatal: 'origin' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

### Proposal

Upon investigation, it looks like the ownership of `/srv/repo` is:

```
drwxrwsrwx    5 root     1000        4.0K Sep  9 09:36 /srv/repo
```

Which is a bit strange, since the container is ran with the following security context:

https://github.com/2i2c-org/infrastructure/blob/7aa7442ab67a0dd040789ccb823d4c83f659be2d/deployer/hub.py#L147-L151

Also, we're seeing the error now, because of a vulnerability fix from GitHub, which catches these same ownership miss-matches: https://github.blog/2022-04-12-git-security-vulnerability-announced/

### Updates and actions

- [ ] I managed to fix this by creating a new init container that fixes the ownership of the repo dir (similar with what we're doing for the home dirs). I will open up a PR in a couple of minutes and get feedback on whether this seems the optimal fix.

PR #3195, opened 2023-09-29
Simplify git cloning hub templates

yuvipanda:
- Git clone into a subdirectory of the `emptyDir` mount, as the mount itself is default owned by root. This was the cause of https://github.com/2i2c-org/infrastructure/issues/1695, which was previously solved by adding another initContainer. This PR removes that, speeding up starts.
- Remove existing directory if it exists before cloning. This prevents the container from failing when the *pod* is restarted due to node issues, as that doesn't clear out `emptyDirs`. And git freaks out if the repo already exists
- Remove redundant `IfNotPresent` imagePullPolicy from alpine/git, as that is not needed when we specify a non :latest tag. Thanks to @consideRatio's investigation in https://github.com/2i2c-org/infrastructure/pull/3165#discussion_r1336351693

Ref https://github.com/2i2c-org/infrastructure/issues/3194

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -526,12 +526,21 @@ jupyterhub:
     initContainers:
       - name: templates-clone
         image: alpine/git:2.40.1
-        imagePullPolicy: IfNotPresent
+        command:
+          - /bin/sh
         args:
-          - clone
-          - --
-          - $(GIT_REPO_URL)
-          - /srv/repo
+          - -c
+          # Remove the existing repo first if it exists, as otherwise we will
+          # error out when the pod restarts. /srv/extra-templates-dir is an
+          # emptyDir volume, so it is *not* cleaned up when the pod restarts -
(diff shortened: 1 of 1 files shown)


helm-charts/basehub/values.yaml
```diff
@@ -526,12 +526,21 @@ jupyterhub:
...
+          - /bin/sh
         args:
-          - clone
-          - --
-          - $(GIT_REPO_URL)
-          - /srv/repo
+          - -c
+          # Remove the existing repo first if it exists, as otherwise we will
+          # error out when the pod restarts. /srv/extra-templates-dir is an
+          # emptyDir volume, so it is *not* cleaned up when the pod restarts -
```
consideRatio, line 535:
```suggestion
          # emptyDir volume, so it is *not* cleaned up when the pod's containers restarts -
```

consideRatio (approved):
Nice this LGTM!

PR #3207, opened 2023-10-02
2i2c, temple: update memory requests/limit on request

consideRatio:
Requsted in https://2i2c.freshdesk.com/a/tickets/1003, indirectly referencing https://2i2c.freshdesk.com/a/tickets/643 that previously had increased requests/limits to the double of basehub's default.


consideRatio:
Since the request arrived late and is relevant ~now, I'd like to self-merge this and not worry about it rather than request quick help from an engineer to review it.

PR #3208, opened 2023-10-02
Cleanup redudant cluster config that was re-setting default values

consideRatio:
- refactor: cleanup redundant config about dask-gateway idle_timeout
- refactor: cleanup redundant config about matchNodePurpose

### DIFF
--- a/config/clusters/carbonplan/common.values.yaml
+++ b/config/clusters/carbonplan/common.values.yaml
@@ -237,8 +237,3 @@ dask-gateway:
       limits:
         cpu: 2
         memory: 4Gi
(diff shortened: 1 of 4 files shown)


yuvipanda (approved):
I'm assuming idle_timeout is now specified upstream? If so, good to go

consideRatio:
> I'm assuming idle_timeout is now specified upstream? If so, good to go

Yepp in our own daskhub helm chart's default values - see https://github.com/2i2c-org/infrastructure/blob/62238918b3fab769115bb53259ff18d7acc9a133/helm-charts/daskhub/values.yaml#L212-L214

PR #3209, opened 2023-10-02
support chart, node-exporter: tolerate 2i2c.org/community tainted nodes

consideRatio:
The prometheus-node-exporter chart declares node taint tolerations by default in a way that it allows node-exporter to schedule on nodes with a "NoSchedule" taint, no matter what kind of key/value that goes with the effect of "NoSchedule".

This makes more sense for us than to declare tolerations for all individual taints we may declare, such as 2i2c.org/community.

Without this fix, we end up without node-exporter running on 2i2c.org/community tainted nodes, which in turn makes us unable to get statistics about user pods cpu and memory usage etc for pods on such nodes.

### DIFF
--- a/helm-charts/support/values.yaml
+++ b/helm-charts/support/values.yaml
@@ -85,21 +85,6 @@ prometheus:
   # values ref: https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-node-exporter/values.yaml
   #
   prometheus-node-exporter:
-    tolerations:
-      # Tolerate tainted jupyterhub user nodes
-      - key: hub.jupyter.org_dedicated
-        value: user
-        effect: NoSchedule
-      - key: hub.jupyter.org/dedicated
-        value: user
-        effect: NoSchedule
-      # Tolerate tainted dask worker nodes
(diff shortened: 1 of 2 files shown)


helm-charts/support/values.yaml
```diff
@@ -85,21 +85,6 @@ prometheus:
...
-      - key: hub.jupyter.org/dedicated
-        value: user
-        effect: NoSchedule
-      # Tolerate tainted dask worker nodes
-      - key: k8s.dask.org_dedicated
-        value: worker
-        effect: NoSchedule
-      - key: k8s.dask.org/dedicated
-        value: worker
-        effect: NoSchedule
```
consideRatio, line 102: [These are the chart's defaults](https://github.com/prometheus-community/helm-charts/blob/3205e454685a5de2a90d27748280160b96884516/charts/prometheus-node-exporter/values.yaml#L357-L359):

```yaml
tolerations:
  - effect: NoSchedule
    operator: Exists
```

yuvipanda (approved):
Good catch!

consideRatio:
Thank you @yuvipanda for reviewing!!

Issue #3150, opened 2023-09-15
Add a staging hub for UCMerced


yuvipanda:
I think it's useful to add a separate staging hub for UCMerced - they have enough users in there, and are still unfortunately on the old image. So having a separate testing ground will help.

Issue #3188, opened 2023-09-28
Enable UC Merced hub users to select between R and Python focused images


colliand:
### Context

UC Merced is working through a pilot with a 2i2c operated education hub service. The pilot involves courses that use R and Python. The software image that 2i2c uses to delivery R and Python is bloated and difficult to keep up to date. The situation is reminiscent of prior experience with the University of Toronto.  Merced may wish to move over to multi-hub service after the pilot but wants to complete the current engagement using a single hub. The 2i2c team discussed some scenarios and shared recommendations with Merced.

### Proposal

Change from the current single R + Python image to offer an image selector after login with a menu offering Merced hub users to choose between two images: an up-to-date Python-focused image and an up-to-date R-focused image. 

Merced personnel have been notified that the menu is likely to create some confusion among hub users. Some students intending to work on Python related work will mistakenly select the R image, etc. Merced is developing communications and local support plans to address these risks. Merced requests that 2i2c provide a precise data and time when this change will be deployed.  Merced asked if this change could be done during the week of October 2 or October 9. @colliand indicated he'd respond with a precise date pending a capacity review by 2i2c's Engineering Team. FYI @damianavila.

yuvipanda:
Let's use the `jupyter/scipy-notebook` image for python, and perhaps the `rocker/binder` image for R. However, neither of these actually have `nbgitpuller` in them, which can be problematic if they are using it. I had opened https://github.com/jupyter/docker-stacks/pull/2000 to put nbgitpuller in jupyter docker-stacks.

There's no staging hub for UC merced - I think perhaps we should change that too #3150 so we can deploy things there for them to test. I think that's the easiest way to do this.

yuvipanda:
Here's what I think is the easiest, least stressful thing for engineers to do:

1. Deploy a uc-merced staging hub (#3150)
2. Set it up to have new images
3. Have UC Merced folks to test it
4. Once they are satisfied, move that to the prod hub

Given the primary reason this is happening is because we're retiring the image we maintain that has both R and python, I think this is the easiest way to do.

yuvipanda:
I'm happy to provide review but would like someone else to be assigned to this :)

colliand:
Thanks @yuvipanda! FYI @damianavila I need to reply to Sarvani with a date when the service will change to include the R + Python images drop-down menu. She asked me via text today.

damianavila:
[...]

Issue #2674, opened 2023-06-19
Reach out to communities to help them transition from 2i2c-hubs-image


consideRatio:
After #2671 the following communities config files still reference quay.io/2i2c/2i2c-hubs-image:

- config/clusters/2i2c/aup.values.yaml
- config/clusters/2i2c/pfw.values.yaml (done in #3052)
- config/clusters/2i2c/ucmerced.values.yaml (staging hub without it established via #3218)
- config/clusters/ubc-eoas/common.values.yaml (done in #3020)

@yuvipanda suggested that:

> we should still reach out to the 4 communities still using this and move them to a different, more appropriate image.

yuvipanda:
pfw has been decomissioned, so aup and ucmerced are the only ones left.

PR #3218, opened 2023-10-04
Add ucmerced staging hub with profileList for Python and R images

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/3150

## Related

- #2674
- #2583
- https://github.com/2i2c-org/infrastructure/issues/3188

### DIFF
--- a/config/clusters/2i2c/ucmerced-staging.values.yaml
+++ b/config/clusters/2i2c/ucmerced-staging.values.yaml
@@ -6,25 +6,6 @@ jupyterhub:
       - secretName: https-auto-tls
         hosts:
           - staging.ucmerced.2i2c.cloud
-  custom:
-    2i2c:
-      add_staff_user_ids_to_admin_users: true
-      add_staff_user_ids_of_type: "google"
-    homepage:
-      templateVars:
-        org:
-          name: University of California, Merced
-          url: http://www.ucmerced.edu/
-          logo_url: https://brand.ucmerced.edu/sites/brand.ucmerced.edu/files/images/ucm-logo-text.png
-        designed_by:
-          name: 2i2c
-          url: https://2i2c.org
-        operated_by:
-          name: 2i2c
-          url: https://2i2c.org
-        funded_by:
-          name: University of California, Merced
-          url: http://www.ucmerced.edu/
   singleuser:
     profileList:
       - display_name: Python
@@ -55,20 +36,5 @@ jupyterhub:
               readOnly: true
   hub:
     config:
-      JupyterHub:
-        authenticator_class: cilogon
       CILogonOAuthenticator:
         oauth_callback_url: https://staging.ucmerced.2i2c.cloud/hub/oauth_callback
-        allowed_idps:
-          urn:mace:incommon:ucmerced.edu:
-            username_derivation:
-              username_claim: "eppn"
-            allow_all: true
-          http://google.com/accounts/o8/id:
-            username_derivation:
-              username_claim: "email"
-            allowed_domains:
-              - "2i2c.org"
-      Authenticator:
-        admin_users:
-          - schadalapaka@ucmerced.edu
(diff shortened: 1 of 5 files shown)


config/clusters/2i2c/ucmerced-staging.values.yaml
```diff
@@ -0,0 +1,76 @@
...
+          - https://accounts.google.com/o/oauth2/auth
+        allowed_idps:
+          urn:mace:incommon:ucmerced.edu:
+            username_derivation:
+              username_claim: "eppn"
+          http://google.com/accounts/o8/id:
+            username_derivation:
+              username_claim: "email"
+            allowed_domains:
+              - "2i2c.org"
```
consideRatio, line 76: - relocated `Authenticator` config to the bottom to be consistent in how other config is ordered
- removed `shown_idps` for oauth16
- added `allow_all` for the ucmerced.edu idp for oauth16 equivalent config

```suggestion
      JupyterHub:
        authenticator_class: cilogon
      CILogonOAuthenticator:
        oauth_callback_url: https://staging.ucmerced.2i2c.cloud/hub/oauth_callback
        allowed_idps:
          urn:mace:incommon:ucmerced.edu:
            username_derivation:
              username_claim: "eppn"
          http://google.com/accounts/o8/id:
            username_derivation:
              username_claim: "email"
            allowed_domains:
              - "2i2c.org"
      Authenticator:
        admin_users:
          - schadalapaka@ucmerced.edu
```

Related to #3144 being merged or not, and I'm looking to merge it in 30 minutes or so after reviewing all config once more

sgibson91 (commented):
Is it worth having a `ucmerced-common.values.yaml` file to reduce repetition, e.g., homepage config?

config/clusters/2i2c/ucmerced-staging.values.yaml
GeorgianaElena: Thanks @consideRatio! I'll wait for the merge (🚀 🚀 🚀 ) to rebase this and manually deploy to double check config.

GeorgianaElena:
Thanks @consideRatio and @sgibson91 for the review. I've updated the PR based on the feedback. Also, I've manually deployed this and is runnitng at https://staging.ucmerced.2i2c.cloud/ What do you think?

consideRatio (approved):
I agree with the adding a staging hub as a way to help the many users of ucmerced test the experience of choosing one of these images while migrating away from the one we have maintained.

Can we add a note that this is to be considered an exception, and that once they want the same experience in the prod hub, we remove this staging hub?

sgibson91:
@GeorgianaElena what was the motivation for only using profile lists on staging? I actually don't see this mentioned in the original issue

consideRatio:
@sgibson91 as a strategy to help them move away from using the old 2i2c maintained image, this relates to https://github.com/2i2c-org/infrastructure/issues/2674 and https://github.com/2i2c-org/infrastructure/issues/2583

sgibson91:
Right, I just didn't think that was why the original issue was opened. I think it was opened because I tanked prod while trying to debug something right before a class 😅

GeorgianaElena:
Sorry for not linking the main issue @sgibson91 (https://github.com/2i2c-org/infrastructure/issues/3188)

I have now also added a comment in `cluster.yaml` about this as well.

consideRatio:
Oops I understood this initially as a "let this hub be deployed first to catch issues via our automation", but its more like a production hub that is deployed side by side with the real production ucmerced hub - used as a staging hub for the end users.

image

sgibson91:
I'm surprised it didn't launch as a staging job given this logic

https://github.com/2i2c-org/infrastructure/blob/e9a7ac7802c5bc8855767ae03c2954743ff974f4/deployer/helm_upgrade_decision.py#L315-L322

PR #3227, opened 2023-10-04
Allow named servers in imagebuilding-demo hub

yuvipanda:


yuvipanda:
Self-merging as it's for a demo hub. I'll open a PR later adding 'for demo hubs, when actively working on them' to the criteria for self-merges.

PR #3231, opened 2023-10-05
[CloudBank] Add Cal Maritime

sean-morris:


config/clusters/cloudbank/csum.values.yaml
```diff
@@ -0,0 +1,78 @@
...
+          http://google.com/accounts/o8/id:
+            username_derivation:
+              username_claim: "email"
+            # allow_all is a partial authorization, username_pattern is enforced also
+            allow_all: true
+      Authenticator:
+        admin_users:
+          - ericvd@berkeley.edu
+          - sean.smorris@berkeley.edu
+          - jteoh@csum.edu
```
consideRatio, line 53: This configuration didn't include `username_pattern`, and with `allow_all` to true for the Google IdP, all users google users are allowed access @sean-morris.

I'll propose a fix for this in a PR

sean-morris: ahhhh.. OK Thank you! so it should be allowed_domains: 2i2c and berkeley

consideRatio: Yes that is an improvement, but that is also not optimal as many berkeley users not intended to have access are then still allowed.

I've proposed what i think is a suitable solution in #3232 where we dont have allowed_domains or allow_all, instead its now enough with listing the usernames to allow access via admin_users.

If you find time time, please review that pull request!

PR #3144, opened 2023-09-15
Upgrade to z2jh 3.1.0 with oauthenticator 16.1

consideRatio:
## Summary

This PR updates from our use of z2jh 3.0.0-beta.1 (where we manually had upgraded KubeSpawner) to z2jh 3.1.0, and with it we get oauthenticator 16.1.0, up from 15.1.0. With this, we have significant amount of changes to review manually as the authentication configuration is not something we test automatically (its very hard).

## Pre-merge review

- Ensure we have all newly added hubs get their configs updated as well, for example #3136
  - [x] Ensured initially
  - [x] Ensured before merge
- Go though _all_ CILogonOAuthenticator config manually
  - [x] Verified to seem OK by Erik
  - [x] Verified to seem OK by at least another engineer (edit: or by Erik again at a later time?)
- Go though _all_ GitHubOAuthenticator config manually
  - [x] Verified by Erik
  - [x] Verified by at least another engineer (edit: or by Erik again at a later time?)
- [x] Deploy to cluster's staging hubs (Saturday 2023-09-30, 15:00 UTC+2)
  - [x] utoronto (cilogon idp https://idpz.utorauth.utoronto.ca/shibboleth)
  - [x] ubc-eoas (cilogon idp https://authentication.ubc.ca)
  - [x] openscapes (github, allow_existing_users)
  - [x] jupyter-meets-the-earth (github, allow_existing_users)
  - [x] gridsst (github, allow_existing_users)
  - [x] cloudbank (a lot of different hubs)
  - [x] carbonplan (cilogon with github idp, allow_existing_users)
  - [x] callysto (cilogon with special authenticator customization)
- [x] Check ourselves the access in staging hubs, or ask for community to check if needed
  - [x] utoronto (cilogon idp https://idpz.utorauth.utoronto.ca/shibboleth)
    I checked with two utoronto accounts
  - [x] ubc-eoas (cilogon idp https://authentication.ubc.ca)
    Asked representatives in https://github.com/2i2c-org/infrastructure/pull/3144#issuecomment-1741769515
  - [x] openscapes (github, allow_existing_users)
    I checked with `consideratio-demo`, first not getting access, then addig that user and getting access, then deleting the user and no longer having access. This was the intented behavior.
  - [x] jupyter-meets-the-earth (github, allow_existing_users)
    I performed a check like for openscapes
  - [x] gridsst (github, allow_existing_users)
    I performed a check like for openscapes
  - [x] cloudbank (a lot of different hubs)
[...]

PR #3232, opened 2023-10-05
cloudbank: revision of auth config

### LINKED ISSUE #3196: Cloudbank followup to review auth config
## Should users of `urn:mace:incommon:berkeley.edu` generally be authorized?

@sean-morris in https://github.com/2i2c-org/infrastructure/pull/3144 I've migrated a lot of auth configuration from oauthenticator v15.1.0 config to oauthenticator v16.1.0 confg. Doing so I observed something that may not be intented in the auth config for cloudbank hubs and wanted to check with you the intent.

In almost all hubs configured to authenticate with the CILogon IdP `urn:mace:incommon:berkeley.edu`, all of the idps users get authorized access to the hub. The exception is those hub that `username_pattern` is configured as that is a separate check that needs to be passed to be authorized.

Is is your intent to authorize all users from that identity provider? If not, additional options arise in oauthenticator 16.1 that we are transitioning to now.

I've for now configured this previous behavior to remain the same by adding `allow_all` under that idp, but after this we can make any individual jupyterhub user listed in `admin_users` be authorized without authorizing all users from an idp like `urn:mace:incommon:berkeley.edu`.

consideRatio:
This PR updates cloudbank cluster's hubs configuration of CILogonOAuthenticator, and various changes made here should also be made in other hubs in other clusters but this PR is scoped for cloudbank specifically for now.

### Now authorized _either by_ ...

With oauthenticator v16, a user can be authorized by different config and doesn't have to be authorized by all config. For example, `allowed_domains` previously rejected admin users signing in if it was configured, but in oauthenticator v16 you can be authorized by being part of either `allowed_domains` or `admin_users`.

### berkeley.edu and 2i2c.org users access reduced

Previously any user with a berkeley.edu email or 2i2c.org email was granted access by many hubs configurations via `allowed_domains`. This is now tightened to only allow specific berkeley.edu usernames and 2i2c.org usernames.

### Berkeley idp removed, Google idp is sufficient

Both berkeley and google has been configured as identity providers, where the berkeley idp was meant to provide access for admin users at berkeley and google idp was meant to provide access for admin users at 2i2c. Since the admin users of berkeley have google accounts as well, we can rely soley on the google idp instead to grant access to all admins.

To reduce the complexity of auth config and help us avoid granting more access than wanted, the berkeley idp is removed entirely.

### csum open access tightened

I saw that the csum hub was configured to grant access to all google users, this access is now reduced. This resolves the comment in https://github.com/2i2c-org/infrastructure/pull/3231#pullrequestreview-1659245312.

### Reordered entries allowed_idps entries

I moved the user relevant idp to the top, and the admin relevant idp to the bottom. This was done for consistency and in preparation for a future improvement about pre-selecting a choice in a CILogon dropdown list. This needs https://github.com/jupyterhub/oauthenticator/issues/690 to be resolved first though.

## Related

- Fixes #3196
- Followup to #3144

### DIFF
--- a/config/clusters/cloudbank/ccsf.values.yaml
+++ b/config/clusters/cloudbank/ccsf.values.yaml
@@ -41,9 +41,6 @@ jupyterhub:
               username_claim: "email"
             allowed_domains:
               - mail.ccsf.edu
-          urn:mace:incommon:berkeley.edu:
-            username_derivation:
-              username_claim: "email"
       Authenticator:
         allowed_users:
           - clare.alice.heimer@gmail.com
--- a/config/clusters/cloudbank/ccsf.values.yaml
+++ b/config/clusters/cloudbank/ccsf.values.yaml
@@ -39,14 +39,14 @@ jupyterhub:
           http://google.com/accounts/o8/id:
             username_derivation:
               username_claim: "email"
-            # allow_all is a partial authorization, username_pattern is enforced also
-            allow_all: true
+            allowed_domains:
+              - mail.ccsf.edu
           urn:mace:incommon:berkeley.edu:
             username_derivation:
               username_claim: "email"
-            # allow_all is a partial authorization, username_pattern is enforced also
-            allow_all: true
       Authenticator:
+        allowed_users:
+          - clare.alice.heimer@gmail.com
         admin_users:
           - ericvd@berkeley.edu
           - sean.smorris@berkeley.edu
@@ -54,7 +54,6 @@ jupyterhub:
           - craig.persiko@mail.ccsf.edu
           - efuchs@mail.ccsf.edu
           - amy.mclanahan@mail.ccsf.edu
-        username_pattern: '^(.+@2i2c\.org|.+@berkeley\.edu|.+@mail\.ccsf\.edu|clare\.alice\.heimer@gmail\.com|deployment-service-check)$'
     extraFiles:
       configurator-schema-default:
         data:
--- a/config/clusters/cloudbank/csum.values.yaml
+++ b/config/clusters/cloudbank/csum.values.yaml
@@ -53,6 +53,7 @@ jupyterhub:
           - ericvd@berkeley.edu
           - sean.smorris@berkeley.edu
           - jteoh@csum.edu
+          - jsimons@csum.edu
     extraFiles:
       configurator-schema-default:
         data:
--- a/config/clusters/cloudbank/csum.values.yaml
+++ b/config/clusters/cloudbank/csum.values.yaml
@@ -39,7 +39,6 @@ jupyterhub:
           https://cma-shibboleth.csum.edu/idp/shibboleth:
             username_derivation:
               username_claim: "email"
-            # allow_all is a partial authorization, username_pattern is enforced also
             allow_all: true
           http://login.microsoftonline.com/common/oauth2/v2.0/authorize:
             username_derivation:
@@ -49,8 +48,6 @@ jupyterhub:
           http://google.com/accounts/o8/id:
             username_derivation:
               username_claim: "email"
-            # allow_all is a partial authorization, username_pattern is enforced also
-            allow_all: true
       Authenticator:
         admin_users:
           - ericvd@berkeley.edu
--- a/config/clusters/cloudbank/demo.values.yaml
+++ b/config/clusters/cloudbank/demo.values.yaml
@@ -42,21 +42,17 @@ jupyterhub:
           http://google.com/accounts/o8/id:
             username_derivation:
               username_claim: "email"
-            # allow_all is a partial authorization for this hub because
-            # username_pattern configured and enforced also, allowing only users
-            # with *.edu suffixed domains besides specific admin users.
-            allow_all: true
       Authenticator:
         admin_users:
           - ericvd@berkeley.edu
           - sean.smorris@berkeley.edu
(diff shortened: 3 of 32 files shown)


config/clusters/cloudbank/ccsf.values.yaml
```diff
@@ -39,22 +39,18 @@ jupyterhub:
...
+        allowed_users:
+          - clare.alice.heimer@gmail.com
         admin_users:
           - ericvd@berkeley.edu
           - sean.smorris@berkeley.edu
           - shawn.wiggins@mail.ccsf.edu
           - craig.persiko@mail.ccsf.edu
           - efuchs@mail.ccsf.edu
           - amy.mclanahan@mail.ccsf.edu
-        username_pattern: '^(.+@2i2c\.org|.+@berkeley\.edu|.+@mail\.ccsf\.edu|clare\.alice\.heimer@gmail\.com|deployment-service-check)$'
```
consideRatio, line 57: Thanks to being able to allow users either by `allowed_domains` and `allowed_users`, we can now grant an individual user access without introducing the more hacky way of configuring `username_pattern`.

Previously, `allowed_domains` once configured would require us to also allow `gmail.com`, but now we can configure `allowed_domains` to authorize a group of users and then surgically adding an individual user via `allowed_users`.

This is the same mechanism that allows us to reduce something like `allowed_domains: ["2i2c.org", "berkeley.edu", "peralta.edu"]` to just `allowed_domains: ["peralta.edu"]` in other hub configurations.

config/clusters/cloudbank/csum.values.yaml
```diff
@@ -39,13 +39,10 @@ jupyterhub:
           https://cma-shibboleth.csum.edu/idp/shibboleth:
             username_derivation:
               username_claim: "email"
-            # allow_all is a partial authorization, username_pattern is enforced also
             allow_all: true
           http://google.com/accounts/o8/id:
             username_derivation:
               username_claim: "email"
-            # allow_all is a partial authorization, username_pattern is enforced also
-            allow_all: true
```
consideRatio, line 48: This hub didn't have username_pattern configured, so declaring `allow_all` led to authorizing all google users. Without having `allow_all`, only google users with certain usernames listed in `admin_users` or similar would be allowed.

config/clusters/cloudbank/demo.values.yaml
```diff
@@ -42,15 +42,11 @@ jupyterhub:
...
-            # allow_all is a partial authorization, username_pattern is enforced also
-            allow_all: true
-          urn:mace:incommon:berkeley.edu:
-            username_derivation:
-              username_claim: "email"
-            # allow_all is a partial authorization, username_pattern is enforced also
+            # allow_all is a partial authorization for this hub because
+            # username_pattern configured and enforced also, allowing only users
+            # with *.edu suffixed domains besides specific admin users.
             allow_all: true
```
consideRatio, line 48: This is now the only situation `username_pattern` is adopted, and I'd like us to minimize this to avoid complexity and rely instead on more establishes practices whenever possible.

In this case, we wanted to allow a range of domains suffixied with `*.edu` and for that we have no tool.

config/clusters/cloudbank/humboldt.values.yaml
```diff
@@ -39,23 +39,16 @@ jupyterhub:
...
             username_derivation:
               username_claim: "email"
             allow_all: true
-          urn:mace:incommon:berkeley.edu:
+          http://google.com/accounts/o8/id:
             username_derivation:
               username_claim: "email"
-            allow_all: true
+            allowed_domains:
+              - humboldt.edu
```
consideRatio, line 50: Here we can see also that I ordered the google idp used for admin access below the https://sso.humboldt.edu/idp/metadata idp. This is in preparation for https://github.com/jupyterhub/oauthenticator/issues/690 that could help us ensure the non-admin idp is pre-selected by default rather than the admin-only relevant idp.

consideRatio (commented):
@sean-morris I wrote some comments to help you understand the motivation behind the different kinds of changes I'm making in this PR

config/clusters/cloudbank/demo.values.yaml
sean-morris: OK, I think we just lock demo down to just us and when people want to use it I can add their Gmail address to allowed_users temporarily. The only caveat would be that in June for the Data Science National Conference, we would want to open it up for that week

config/clusters/cloudbank/ccsf.values.yaml
sean-morris: ok got it

config/clusters/cloudbank/demo.values.yaml
consideRatio: @sean-morris okay! I've commented out the username_pattern and made sure that its easy to enable back again for you temporarily in the future.

consideRatio:
@sean-morris do you approve merging this PR at this point? Since last review, I've added the admin user you added as part of #3239 that I closed in favor of this, and I've reduced the demo hub access to just the listed admin users for now but ensured its easy to enable ~public access temporarily when needed in the future.

config/clusters/cloudbank/demo.values.yaml
sean-morris: Sounds great.

PR #3233, opened 2023-10-05
Grant access for 2i2c members only via admin_users

consideRatio:
With oauthenticator 16, being listed in `admin_users` makes 2i2c members have access even if they are now listed in `allowed_domains` or `allowed_organizations`. Due to this, we can avoid the need to declare `2i2c.org` and `2i2c-org` there and rely on being listed in `admin_users`.

I've provided a few comments, one per kind of change made.

## Related

- This is done for cloudbank cluster's hub in #3232
- #3144

### DIFF
--- a/config/clusters/2i2c-aws-us/dask-staging.values.yaml
+++ b/config/clusters/2i2c-aws-us/dask-staging.values.yaml
@@ -37,7 +37,6 @@ basehub:
           authenticator_class: "github"
         GitHubOAuthenticator:
           oauth_callback_url: "https://dask-staging.aws.2i2c.cloud/hub/oauth_callback"
-          allowed_organizations:
-            - 2i2c-org
+          allowed_organizations: []
           scope:
             - read:org
--- a/config/clusters/2i2c-aws-us/dask-staging.values.yaml
+++ b/config/clusters/2i2c-aws-us/dask-staging.values.yaml
@@ -37,6 +37,3 @@ basehub:
           authenticator_class: "github"
         GitHubOAuthenticator:
           oauth_callback_url: "https://dask-staging.aws.2i2c.cloud/hub/oauth_callback"
-          allowed_organizations: []
-          scope:
-            - read:org
--- a/config/clusters/2i2c-aws-us/go-bgc.values.yaml
+++ b/config/clusters/2i2c-aws-us/go-bgc.values.yaml
@@ -33,7 +33,6 @@ jupyterhub:
       GitHubOAuthenticator:
         oauth_callback_url: https://go-bgc.2i2c.cloud/hub/oauth_callback
         allowed_organizations:
-          - 2i2c-org:hub-access-for-2i2c-staff
           - go-bgc
         scope:
           - read:org
--- a/config/clusters/2i2c-uk/staging.values.yaml
+++ b/config/clusters/2i2c-uk/staging.values.yaml
@@ -43,5 +43,3 @@ jupyterhub:
           http://google.com/accounts/o8/id:
             username_derivation:
               username_claim: "email"
-            allowed_domains:
-              - "2i2c.org"
--- a/config/clusters/2i2c/mtu.values.yaml
+++ b/config/clusters/2i2c/mtu.values.yaml
@@ -36,18 +36,16 @@ jupyterhub:
       CILogonOAuthenticator:
         oauth_callback_url: "https://mtu.2i2c.cloud/hub/oauth_callback"
         allowed_idps:
-          # Allow 2i2c staff to login with Google
-          http://google.com/accounts/o8/id:
-            username_derivation:
-              username_claim: "email"
-            allowed_domains:
-              - "2i2c.org"
           # Allow MTU to login via Shibboleth
           https://sso.mtu.edu/idp/shibboleth:
             username_derivation:
               username_claim: "email"
             allowed_domains:
               - "mtu.edu"
+          # Allow 2i2c staff to login with Google accounts
+          http://google.com/accounts/o8/id:
+            username_derivation:
+              username_claim: "email"
       Authenticator:
         admin_users:
           - "dbkc@mtu.edu"
--- a/config/clusters/callysto/common.values.yaml
+++ b/config/clusters/callysto/common.values.yaml
@@ -19,6 +19,9 @@ jupyterhub:
       enabled: true
   custom:
     2i2c:
+      # add_staff_user_ids_to_admin_users is disabled because the usernames
+      # aren't github id or email based, individual 2i2c members have added
+      # their user to admin_users manually instead.
       add_staff_user_ids_to_admin_users: false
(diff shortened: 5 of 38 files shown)


config/clusters/2i2c-aws-us/go-bgc.values.yaml
```diff
@@ -33,7 +33,6 @@ jupyterhub:
       GitHubOAuthenticator:
         oauth_callback_url: https://go-bgc.2i2c.cloud/hub/oauth_callback
         allowed_organizations:
-          - 2i2c-org:hub-access-for-2i2c-staff
```
consideRatio, line 36: Removing ourselves from `allowed_organization` is fine, because we are added to `admin_users`.

Doing this also isn't influencing the filtering of profile_list entries using our basehub injected `allowed_teams` config.

config/clusters/2i2c-uk/staging.values.yaml
```diff
@@ -43,5 +43,3 @@ jupyterhub:
           http://google.com/accounts/o8/id:
             username_derivation:
               username_claim: "email"
-            allowed_domains:
-              - "2i2c.org"
```
consideRatio, line 47: Removing ourselves from a CILogonOAuthenticator `allowed_domains` is also fine, we are allowed by being added to `admin_users`.

config/clusters/2i2c/mtu.values.yaml
```diff
@@ -36,18 +36,16 @@ jupyterhub:
...
           # Allow MTU to login via Shibboleth
           https://sso.mtu.edu/idp/shibboleth:
             username_derivation:
               username_claim: "email"
             allowed_domains:
               - "mtu.edu"
+          # Allow 2i2c staff to login with Google accounts
+          http://google.com/accounts/o8/id:
+            username_derivation:
+              username_claim: "email"
```
consideRatio, line 48: This was a re-ordering of config entries. I want us to put the user-facing idp's first as a way to prepare for letting that lead to presenting that option by default in favor of presenting the option only relevant to us admin first.

For more details about this, see https://github.com/jupyterhub/oauthenticator/issues/690

config/clusters/callysto/common.values.yaml
```diff
@@ -19,6 +19,9 @@ jupyterhub:
       enabled: true
   custom:
     2i2c:
+      # add_staff_user_ids_to_admin_users is disabled because the usernames
+      # aren't github id or email based, individual 2i2c members have added
+      # their user to admin_users manually instead.
       add_staff_user_ids_to_admin_users: false
       # add_staff_user_ids_of_type: "google"
```
consideRatio, line 26: In this hub where we aren't added to admin_users, it remains important to not remove allowing us via `allowed_domains`.

config/clusters/meom-ige/common.values.yaml
```diff
@@ -92,7 +92,6 @@ basehub:
         - display_name: "Small"
           default: true
           allowed_teams: &allowed_teams_normal_use
-            - 2i2c-org:hub-access-for-2i2c-staff
             - meom-group:hub-users # long term users
```
consideRatio, line 95: Removing us from `allowed_teams` was a mistake, if this is defined and our github team isn't part of this, we won't be able to see this profile.

```suggestion
            - 2i2c-org:hub-access-for-2i2c-staff
            - meom-group:hub-users # long term users
```

config/clusters/smithsonian/common.values.yaml
```diff
@@ -76,7 +75,6 @@ basehub:
           description: &profile_list_description "Start a container with at least a chosen share of capacity on a node of this type"
           slug: small
           default: true
-          allowed_teams: *allowed_github_orgs
```
consideRatio, line 79: Its better if we just allow all users by default instead, because otherwise we need to make a few extra API calls to check for membership etc for no real reason.

Since https://github.com/2i2c-org/infrastructure/pull/3234#issuecomment-1749180174, removing `allowed_teams` makes all users logged in to the hub be able to see this like we want.

docs/hub-deployment-guide/configure-auth/github-orgs.md
```diff
@@ -97,7 +97,6 @@ You can remove yourself from the org once you have confirmed that login is worki
           GitHubOAuthenticator:
             oauth_callback_url: https://{{ HUB_DOMAIN }}/hub/oauth_callback
             allowed_organizations:
-              - 2i2c-org:hub-access-for-2i2c-staff
```
GeorgianaElena, line 100: Since we're updating the GitHub authentication example to not include the 2i2c org, let's also update the example in https://infrastructure.2i2c.org/hub-deployment-guide/configure-auth/cilogon/#configure-the-authenticator to not include 2i2c in the `allowed_domains`.

GeorgianaElena (approved):
I think this is good to go so it doesn't get into conflict with future work.

Note, that I believe documenting the oauthenticator > 16 new behaviour in the context of 2i2c config is really important, as I often find myself confused by the new defaults, even though I stayed pretty up to date with the work that into into that release.

So, I believe it is really important to have the docs up to date with the new version, so we can all relate to it when uncertain.

We should prioritize getting https://github.com/2i2c-org/infrastructure/issues/3247 ready soon, as atm, after we merge this PR, the examples in the docs are not complete and might cause confusion.

Until https://github.com/2i2c-org/infrastructure/issues/3247 gets implemented, since we are removing the need to explicitly allow the 2i2c org or the 2i2c domain from the examples in the docs in this PR, what do you think about adding a really quick warning message in the docs pages of both authentication options that says something like:

```warning
This examples assume that `add_staff_user_ids_to_admin_user` and `add_staff_user_ids_of_type` were configured for this hub to allow 2i2c staff access as admins.

If this is not the case, consider making the appropriate changes to allow such access.
```

docs/hub-deployment-guide/configure-auth/github-orgs.md
consideRatio: Resolved via 53c8e49

config/clusters/2i2c-aws-us/dask-staging.values.yaml
```diff
@@ -37,7 +37,6 @@ basehub:
           authenticator_class: "github"
         GitHubOAuthenticator:
           oauth_callback_url: "https://dask-staging.aws.2i2c.cloud/hub/oauth_callback"
-          allowed_organizations:
-            - 2i2c-org
+          allowed_organizations: []
```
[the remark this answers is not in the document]
consideRatio: Resolved via 2bd0faf

consideRatio:
Thank you for reviewing this @GeorgianaElena!!! I agree on the importance of following up in #3247, and I've done some steps towards it in [53c8e49](https://github.com/2i2c-org/infrastructure/pull/3233/commits/53c8e494e96e4d309ab91c3b6d500f0376e6bd10) as a start.

PR #3236, opened 2023-10-05
callysto: fix auth logic for use with oauthenticator v16

### LINKED ISSUE #3235: Fix Callysto's custom authenticator
Callysto has a custom OAuthenticator, and I think it was kind of broken before, but now its entirely broken for all users that isn't listed as `admin_users` :'(

The reason is that oauthenticator 16 is not using the `authenticate` method as it did before, so overriding it has a different effect. I'll work on fixing it now.

consideRatio:
- Closes #3235

Currently deployed to staging and prod as I figured it would help us avoid stressing getting this in.

### DIFF
--- a/config/clusters/callysto/common.values.yaml
+++ b/config/clusters/callysto/common.values.yaml
@@ -65,67 +65,68 @@ jupyterhub:
   hub:
     extraConfig:
       001-cilogon-email-auth: |
-        # Custom override of CILogonOAuthenticator to allow access control via one
-        # property (email) while the username itself is a different property (oidc).
-        # This allows us to restrict access based on the *email* of the authenticated
-        # user without having to store the actual email in our servers.
-        from oauthenticator.cilogon import CILogonOAuthenticator
         from fnmatch import fnmatch
-        from traitlets import List, Unicode
+
+        from oauthenticator.cilogon import CILogonOAuthenticator
+
 
         class EmailAuthenticatingCILogonOAuthenticator(CILogonOAuthenticator):
-          allowed_domains = List(
-            Unicode,
-            default=None,
-            # This must always be set!
-            allow_none=False,
-            help="""
-            Allow access to users with verified emails belonging to these domains.
+            """
+            Custom override of CILogonOAuthenticator to allow access control via
+            one property (email) while the username itself is a different
+            property (oidc), and to allow `allowed_domains` to involve wildcards
+            like *.
 
-            Allows glob patterns but not regexes. So you can allow access to
-            *.edu, for example. See https://docs.python.org/3/library/fnmatch.html for
-            list of allowed patterns.
-            """,
-            config=True
-          )
+            This allows us to restrict access based on the *email* of the
+            authenticated user without having to store the actual email in our
+            servers or use it as a username.
+            """
 
-          async def authenticate(self, *args, **kwargs):
-            authenticated_user = await super().authenticate(*args, **kwargs)
+            async def check_allowed(self, username, auth_model):
+                """
+                Replaces the default implementation of check_allowed as it
+                otherwise would error, not finding a `@` symbol in the non-email
+                username.
 
-            email = authenticated_user['auth_state']['cilogon_user']['email']
+                Mostly a copy of OAuthenticator 16.1, for comparison see:
+                https://github.com/jupyterhub/oauthenticator/blob/16.1.0/oauthenticator/cilogon.py#L349-L373
+                """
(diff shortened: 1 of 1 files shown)


config/clusters/callysto/common.values.yaml
```diff
@@ -65,67 +65,68 @@ jupyterhub:
...
+            authenticated user without having to store the actual email in our
+            servers or use it as a username.
+            """
 
-            Allows glob patterns but not regexes. So you can allow access to
-            *.edu, for example. See https://docs.python.org/3/library/fnmatch.html for
-            list of allowed patterns.
-            """,
-            config=True
-          )
```
consideRatio, line 90: Instead of defining a new trait, we rely on the existing on provided under the individual `allowed_idps` entries.

config/clusters/callysto/common.values.yaml
```diff
@@ -65,67 +65,68 @@ jupyterhub:
...
+            async def check_allowed(self, username, auth_model):
+                """
+                Replaces the default implementation of check_allowed as it
+                otherwise would error, not finding a `@` symbol in the non-email
+                username.
 
-          async def authenticate(self, *args, **kwargs):
-            authenticated_user = await super().authenticate(*args, **kwargs)
```
consideRatio, line 93: Instead of overriding `authenticate` we have to override `check_allowed` based on how it works now in oauthenticator v16.

config/clusters/callysto/common.values.yaml
```diff
@@ -65,67 +65,68 @@ jupyterhub:
...
+                if idp_allow_all:
+                    return True
+
+                user_info = auth_model["auth_state"][self.user_auth_state_key]
+                user_idp = user_info["idp"]
+
+                idp_allowed_domains = self.allowed_idps[user_idp].get("allowed_domains")
+                if idp_allowed_domains:
+                    # this was the part we wanted to replace
+                    email = auth_model["auth_state"]["cilogon_user"]["email"]
```
GeorgianaElena, line 110: Just to be consistent and to make sure this doesn't change [upstream](https://github.com/jupyterhub/oauthenticator/blob/2fa1c4818f7ba3444f3e2e44f8ddb38127553ffd/oauthenticator/cilogon.py#L45) and forget about it here:

```suggestion
                    email = auth_model["auth_state"][self.user_auth_state_key]["email"]
```

GeorgianaElena (commented):
Thanks for fixing this @consideRatio!

I am not sure if you are aware of https://github.com/jupyterhub/oauthenticator/issues/547, that I opened right after Yuvi implemented this special case authenticator for Callysto.

What do you think about upstreaming this change to authenticator directly?
I still believe this is something useful to have by default in `oauthenticator`.

Also, we're still not enforcing to define `allowed_domains` only when the claim is `email`. So people can still define an `allowed_idp.allowed_domains` even if their `usernama_claim` is not `email`, which will make `check_allowed` to fail, with a potentially confusing error message.

What do you think?

consideRatio:
> I am not sure if you are aware of https://github.com/jupyterhub/oauthenticator/issues/547, that I opened right after Yuvi implemented this special case authenticator for Callysto.

Thanks for linking that and opening it in the first place! I wasn't aware and had it on my todo list to write such issue :)

> Also, we're still not enforcing to define `allowed_domains` only when the claim is `email`. So people can still define an `allowed_idp.allowed_domains` even if their `usernama_claim` is not `email`, which will make `check_allowed` to fail, with a potentially confusing error message.

Yes, this confusion makes me consider it almost a bugfix to make the breaking change of checking against `email` claim instead of username. I think it may not end up being a breaking change, but I'm not sure... Let's follow up such discussion in the linked issue.

> What do you think about upstreaming this change to authenticator directly?

I'd like to take this step first separately because callysto's deployment is broken currently, and I don't want to make us feel forced to push through something upstream because we have put us in a situation needing that.

consideRatio:
Thank you for reviewing this @GeorgianaElena!!!!

GeorgianaElena (approved):
> I'd like to take this step first separately because callysto's deployment is broken currently, and I don't want to make us feel forced to push through something upstream because we have put us in a situation needing that.

Agree! Then let's merge this and remove it once we've upstreamed it 🚀 🚀 
Actually, @consideRatio, can you please also add a FIXME comment with a link to the upstream issue before merging, so we know where this is being tracked?

consideRatio:
@GeorgianaElena 100% agree about the comment. I linked to two issues, one about wildcards to handle `*.edu` and one about checking against `email` claim and not the chosen username.

PR #3239, opened 2023-10-05
[CloudBank] CSUM: Fixed Google Auth and Admin

sean-morris:


consideRatio:
This is done a bit differently in #3232 so I'll close this in favor of that PR, making sure that the admin user `jsimons@csum.edu` is added there as well as done as part of this PR.

PR #3243, opened 2023-10-07
Bump image tag for callysto hub on 2i2c

ianabc:
Adding CollabTurtlePlus and various other minor updates.


ianabc:
Forgot I can do this via the configurator now. Closing.

Issue #3068, opened 2023-09-01
[Request deployment] New Hub: NASA ESDIS Hub


colliand:
### The GitHub handle of the community representative

@bilts

### Hub important dates

- **Target Start Date**: 2023-09-22
- **Required Start Date**: 2023-09-30

### Hub Authentication Type

GitHub (e.g., @mygithubhandle)

### First Hub Administrators

- Patrick Quinn (@bilts)

### [GitHub Auth only] How would you like to manage your users?

pending

### [GitHub Teams Auth only] Profile restriction based on team membership

pending

This hub should launch with configuration and image based on the OpenScapes hub. 

### Hub logo image URL

pending

### Hub logo website URL

https://www.earthdata.nasa.gov/esdis

### Hub user image GitHub repository

pending

### Hub user image tag and name

pending; likely same as openscapes

### Extra features you would like to enable

- [ ] Dedicated Kubernetes cluster
- [ ] Scalable Dask Cluster

### (Optional) Preferred cloud provider

None

### (Optional) Billing and Cloud account

None

### Other relevant information to the features above

This hub deployment request is part of 2i2c's ongoing `NASA VEDA` project. The ESDIS hub will be launched with configuration and image based on the OpenScapes hub. The hub should be deployed in the NASA SMD cluster where 2i2c is currently operating the NASA VEDA hub. 

I request confirmation or edits to these configuration plans from @freitagb or @bilts. 

### Tasks to deploy the hub

- [ ] 1. Deploy information filled in above
- [ ] 2. Engineer who will deploy the hub is assigned
- [ ] 3. If using GitHub Orgs/Teams Auth, Engineer is given Owner rights to the org to set this up.
- [ ] 4. Initial Hub deployment PR <link to PR>
- [ ] 5. Administrators able to log on -> Hub now in steady-state

colliand:
Some information is required from @bilts to deploy this hub. 

- [ ] Please share a **URL pointing to the splash image** you'd like to appear on the login page for the ESDIS hub. See the [NASA VEDA hub login page](https://nasa-veda.2i2c.cloud/hub/login?next=%2Fhub%2F) for an example. 
- [ ] Following the pattern we've used with VEDA and GHG, I assume that access to the ESDIS hub will be managed by membership in a GitHub team or organization. Please share the URL for that GitHub team or org.

damianavila:
I have assigned @sgibson91 for the deployment of this hub.

[...]

PR #3245, opened 2023-10-09
Setup NASA ESDIS cluster and hub

sgibson91:
ref: https://github.com/2i2c-org/infrastructure/issues/3068


yuvipanda (approved):
This looks good to me. Thanks for working this through, @sgibson91.

I appreciate that the domain is not called nasa-esdis but just esdis - yay (https://github.com/2i2c-org/infrastructure/issues/3029). I think in the future, we should also not use the word 'nasa' in the name of the cluster itself, so the name matches the domain. But I think it's probably too much hassle to rename and redeploy this one, so I think it's fine to leave this as is. But I'd love for us to put this someplace as a policy so we don't use the 'nasa' prefix next time. Do you think you can help with that as a followup, @sgibson91?

PR #3254, opened 2023-10-11
nox: add docstring about nox and list command how to use it

### LINKED ISSUE #1000: Add description about what noxfile.py is
### Description of problem and opportunity to address it

**Context to understand the problem**

I'm trying to understand this code base, but find a file like `noxfile.py` without any comments in it making me stumble.

**Problem or idea**

I'd love to have a comment in the top of various standalone files that describes what they are about. For example, I've added that to the read the docs config file in #999 which also clarifies what RTD project should be inspected if something is acting up etc.

**What's the value and who would benefit**

Everyone would benefit - even the person added this file - because over time you forget.

- [ ] Add a comment of some kind to the noxfile.py file describing its purpose etc, allowing someone finding it to understand its role and purpose in this github project.
- [ ] If it was copy-pasted, a source of where it was copy-paste from should be included for example.
- [ ] If there is a known place where you can learn more about something etc that could be worth linking to as well.

### Updates and ongoing work

_No response_

consideRatio:
I think its a great practice to ensure that various things in a github repo is somewhat self-explanatory as that helps anyone trying to onboard do it more efficiently. I didn't know how to use `nox` myself before doing this, and at one point I didn't know for what reason this file was around or how it was related to a repo I found it in.

A docstring like this can resolve such issues for us!

- Closes #1000

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,3 +1,14 @@
+"""
+noxfile.py is a configuration file for the command-line tool nox that automates
+tasks in multiple Python environments. We use it to setup an environment to
+build our documentation.
+
+Config reference: https://nox.thea.codes/en/stable/config.html#noxfile
+
+Common tasks:
+- Install nox:                        pip install nox
+- Start a live reloading docs server: nox -- live
+"""
 import nox
 
 nox.options.reuse_existing_virtualenvs = True


choldgraf (approved):
This is a great idea. I'm kicking myself for not adding in a little docstring. I think the content looks great and I'd be fine with merging as-is, though I also wonder what @consideRatio thinks about simply linking to our `nox` documentation in the team compass:

https://compass.2i2c.org/reference/documentation/environment/#install-nox

(or potentially making that a dedicated section on Nox so that it's more discoverable). That way, we can point all of our `noxfile.py` files to that site, and update the site's content without needing to update each individual file.

consideRatio:
Thanks for reviewing and approving it as it is @choldgraf!

For this brief intro that I think by itselfs does a good enough job, I think its a good call to break the DRY principle and directly write it all noxfile.py files.

If we were to link to a centralized intro, I still think it would make sense to provide a brief intro like this before linking out as otherwise it may not be enticing enough to click the link in the first place.

consideRatio:
Ah missed that there was existing docs to link to already.

I think adding a link to that could make sense to learn a bit more, but that the first paragraph and config ref link is worth retaining.

PR #3234, opened 2023-10-05
basehub: fix added allowed_teams profile_list filtering for use in oauth v16

consideRatio:
I saw that smithsonian's prod hub is down for normal users (but not our deployment check), like this:

image

This is because our github profile_list filtering based on org/team membership fails, and that in turn is because we relied on a private API in oauthenticator that changed. This PR fixes this.

This PR also changes the behavior in a situation no hub is in, when a profile without `allowed_teams` key is defined. In such cases the profile wasn't presented, but after this PR it is for all users by default instead. Its possible to provide a profile nobody sees by declaring `allowed_profiles: []` still, so no functionality is lost.

```
User daisycm08 is part of teams 
[E 2023-10-04 21:13:10.275 JupyterHub web:1871] Uncaught exception GET /hub/spawn (192.168.51.250)
    HTTPServerRequest(protocol='https', host='smithsonian.2i2c.cloud', method='GET', uri='/hub/spawn', version='HTTP/1.1', remote_ip='192.168.51.250')
    Traceback (most recent call last):
      File "/usr/local/lib/python3.11/site-packages/tornado/web.py", line 1786, in _execute
        result = await result
                 ^^^^^^^^^^^^
      File "/usr/local/lib/python3.11/site-packages/jupyterhub/handlers/pages.py", line 207, in _get
        spawner_options_form = await spawner.get_options_form()
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/local/lib/python3.11/site-packages/jupyterhub/spawner.py", line 542, in get_options_form
        options_form = await maybe_future(self.options_form(self))
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/local/lib/python3.11/site-packages/kubespawner/spawner.py", line 2977, in _render_options_form_dynamically
        profile_list = await maybe_future(self.profile_list(current_spawner))
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "<string>", line 72, in profile_list_allowed_teams_filter
    TypeError: GitHubOAuthenticator._check_membership_allowed_organizations() missing 1 required positional argument: 'token_type'
```

PR #3261, opened 2023-10-11
basehub: fix bug in profile list filtering influencing a few users

consideRatio:
This is a followup to #3234 where I didn't account for auth_state could be retained from oauthenticator 15 even after we upgraded to oauthenticator 16.

The bug was reported by @jbusecke in https://2i2c.freshdesk.com/a/tickets/1037, thank you Julius!!!

I've deployed this to staging successfully, and then also deployed it to prod. Merging this ensures it remains in prod long term but otherwise we have time for more detailed review.

---

I think ths bug led to 500 errors if:

- GitHubOAuthenticator was used with the `allowed_teams` filtering of `profile_list` options
- A user had been logged in since oauthenticator 15 was used, needed to start a server and arrived to /hub/spawn, and didn't have access to all server options.

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -764,8 +764,15 @@ jupyterhub:
                     allowed_profiles.append(profile)
                     continue
 
-                access_token = auth_state["token_response"]["access_token"]
-                token_type = auth_state["token_response"]["token_type"]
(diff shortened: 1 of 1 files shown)


consideRatio:
Reviewed by yuvi on slack via screenshots as he didnt have github access

Issue #3217, opened 2023-10-03
[EVENT] OpenScapes


sgibson91:
### Summary

> We have an event coming up on October 17, 2023 and are expecting ~100 people using large python instances. I'm reaching out in advance to make sure the necessary computational resources are available on the front end so that things go smoothly.  

### Event Info

- **Community Representative:** 
- **Event begin:** October 17th, 2023
  - **In your timezone:** 
- **Event end:** October 17th, 2023
  - **In your timezone:** 
- **Active times:** 
  - **In your timezone:** 
- **Number of attendees:** ~100
- [**Hub Events Calendar**](https://calendar.google.com/calendar/u/2?cid=Y19rdDg0c2g3YW5tMHNsb2NqczJzdTNqdnNvY0Bncm91cC5jYWxlbmRhci5nb29nbGUuY29t)

### Hub info

- **Hub URL**: 
- **Hub decommisioned after event?**:  No

### Task List

**Before the event**

- [ ] Dates confirmed with the community representative and added to Hub Events Calendar.
- [ ] Quotas from the cloud provider are high-enough to handle expected usage.
- [ ] **One week before event** Hub is running.
- [ ] Confirm with Community Representative that their workflows function as expected.
  - <details>
    <summary>👉Template message to send to community representative</summary>
    
    ```
    Hey {{ COMMUNITY REPRESENTATIVE }}, the date of your event is getting close!
    
    Could you please confirm that your hub environment is ready-to-go, and matches your hub's infrastructure setup, by ensuring the following things:
    - [ ] Confirm that the "Event Info" above is correct
    - [ ] On your hub: log-in and authentication works as-expected
    - [ ] `nbgitpuller` links you intend to use resolve properly
    - [ ] Your notebooks and content run as-expected
    ```
    
    </details>  
- [ ] **1 day before event**, either a separate nodegroup is provisioned for the event or the cluster is scaled up.

**During and after event**

- [ ] Confirm event is finished.
- [ ] Nodegroup created for the hub is decommissioned / cluster is scaled down.
- [ ] Hub decommissioned (if needed).
- [ ] Debrief with community representative.
  - <details>
    <summary>👉Template debrief to send to community representative</summary>
      
     ```
     Hey {{ COMMUNITY REPRESENTATIVE }}, your event appears to be over 🎉
     
[...]

Issue #3177, opened 2023-09-27
Discussion towards documenting guidance for deciding on a users-per-node size


consideRatio:
Consider the of [active users from ubc-eoas last month](https://grafana.ubc-eoas.2i2c.cloud/d/hub-dashboard/jupyterhub-dashboard?orgId=1&from=1693208240488&to=1695800240488):

![image](https://github.com/2i2c-org/infrastructure/assets/3837114/41f28fca-6d2e-4015-9742-e9970bdd4690)

In this graph there is a peak of 86 users at one point in time, this peak looks like this:

![image](https://github.com/2i2c-org/infrastructure/assets/3837114/234915a1-5fc2-4cb9-98c3-02d6105a0649)

If we can choose the node size to fit any multiple of two amounts of users, what users-per-node size seems to be a good choice in this case?

**Extreme 1: all users on one node**

- :heavy_plus_sign: only at one point in time would there be startup time, when this node starts
- :heavy_plus_sign: with more users its easier to make effective use of CPU and Memory if limits are higher than requests
- :heavy_minus_sign: the cost incurred for a single user is as large as 100 users

**Extreme 2: user dedicated nodes**

- :heavy_plus_sign: the cost incurred for a single user is minimized
- :heavy_minus_sign: the startup time is significant for every user
- :heavy_minus_sign: the average cost per user will be high as CPU and Memory use will be very inefficient as CPU is typically idle

## Guidande to find the middle ground

Please join me in discussing what middle ground can make sense, so that we can _document guidance how to identify it_. 

Using ubc-eoas usage above as an example, how many users do you think could make sense for them to have per node, and based on what?

consideRatio:
I suggest we settle on providing guidance based on something simple enough, such as:

- **Running server statistics (min, average, median, max)**
  Basic statistics about the running user servers can be easily understood, guessed ahead of time, visually estimated from a dashboard, and even programatically measured.
- **User requests for CPU and memory**
  Without this we can't determine how many users could fit on a node of a certain size.

## Guidance draft 1

This is my initial guidance suggestion, put in three steps.

1. (lower bound) Choose a node size that at least fits the minimum amount of users
2. (upper bound) Choose a node size that makes peak usage result in having _at least_ around 3 nodes
3. (preference) Prefer a node size from one of the default size set suggested in #3176

[...]

PR #3262, opened 2023-10-12
openscapes: Use a bigger node to account for future event expected usage

GeorgianaElena:
For https://github.com/2i2c-org/infrastructure/issues/3217

### DIFF
--- a/config/clusters/openscapes/prod.values.yaml
+++ b/config/clusters/openscapes/prod.values.yaml
@@ -57,7 +57,7 @@ basehub:
                     cpu_guarantee: 1.875
                     cpu_limit: 3.75
                     node_selector:
-                      node.kubernetes.io/instance-type: r5.xlarge
+                      node.kubernetes.io/instance-type: r5.4xlarge
                 mem_29_7:
                   display_name: 29.7 GB RAM, upto 3.75 CPUs
                   kubespawner_override:
@@ -66,7 +66,7 @@ basehub:
                     cpu_guarantee: 3.75
                     cpu_limit: 3.75
                     node_selector:
-                      node.kubernetes.io/instance-type: r5.xlarge
+                      node.kubernetes.io/instance-type: r5.4xlarge
                 mem_60_6:
                   display_name: 60.6 GB RAM, upto 15.72 CPUs
                   kubespawner_override:


consideRatio (commented):
## More users per node during events - yes!

I think its great to schedule so that we expect 8 users to fit on the node instead of just 2 for an event. During events we expect significant amounts of users, and I figure ending up with using around ~4-12 nodes is about right with regards to startup time considerations and ability to scale up/down in a more fine grained way to reduce cloud costs.

## Warning about just changing node selector

The guarantees and limits are generated based on the specific node size, a node with 4 times the size will have slightly more than 4 times the allocatable capacity overall I think.

If this was done in the other way, where we used requests(aka guarantees)/limits based on a larger node, adopted for a smaller node, we could end up with request of 51% or 101% of a nodes capacity etc, making us schedule ineffectively.

So in practice - we must be careful changing this without re-generating the list, but in this case its okay I think.

## Related discussions

- When the ~same resource request can be made on two node sizes, we currently only provide one option. Sometimes one is better than the other, so it would be good to be able to configure what requests should use what instance size. This is exactly what is changed here, we want to use the larger node for requests that could also be made on a smaller node during an event where we expect many users.
  https://github.com/2i2c-org/infrastructure/pull/3030#issuecomment-1706777492
- How many users per nodes are suitable?
  https://github.com/2i2c-org/infrastructure/issues/3177

consideRatio (approved):

GeorgianaElena:
> When the ~same resource request can be made on two node sizes, we currently only provide one option. Sometimes one is better than the other, so it would be good to be able to configure what requests should use what instance size. This is exactly what is changed here, we want to use the larger n

Ah, I get it! Thank you @consideRatio. I am feeling a bit more confident in what we want to achieve with node sharing and how, but it's lots of info to consider. But would love to have a central place where we document all these different scenarios.

consideRatio:
>  But would love to have a central place where we document all these different scenarios.

I agree, I would have felt agency to write docs about this if we had clear guidance to go for. I've opened https://github.com/2i2c-org/infrastructure/issues/3177 to come up with preliminary guidance on how many users per node we may want to aim for. With such guidance we can not only document "this is complicated because..." but also "so here is a recommendation on what to do".

GeorgianaElena:
Thank you @consideRatio! https://github.com/2i2c-org/infrastructure/issues/3177 seems like Q4 material 🚀 🚀 and something I'm looking forward to have <3

PR #3276, opened 2023-10-17
Mark auth_token as a string, not integer

yuvipanda:
When marked as an integer, leading zeros get stripped.

### DIFF
--- a/deployer/cloud_access.py
+++ b/deployer/cloud_access.py
@@ -19,7 +19,7 @@ def exec_aws_shell(
     mfa_device_id: str = typer.Argument(
         ..., help="Full ARN of MFA Device the code is from"
     ),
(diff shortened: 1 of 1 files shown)


consideRatio (approved):

consideRatio:
I suggest cancelling deploy workflow after merge for this

yuvipanda:
@consideRatio i'm going to watch this one run through. I've caught hubs that have failing CI issues before via merges like this. Doing these once in a while isn't a bad idea.

yuvipanda:
There were a bunch of failures, but since https://github.com/2i2c-org/infrastructure/pull/3282 has to redeploy everything anyway, I'm going to just watch that instead now :)

PR #3255, opened 2023-10-11
Add jackeddy hub

GeorgianaElena:
For https://github.com/2i2c-org/infrastructure/issues/3166

It enables the same features the previous hub from https://github.com/2i2c-org/infrastructure/pull/1466 had + `unlisted_choice` and node sharing.

Note, that out of curiosity, I tried to link each "not-standard" config I found there to the engineering docs that explain them. This is why, there are some fixme comments and links to infra docs.

### The **documented** features this hub has:

- [setup custom, free-form user profile choices](https://infrastructure.2i2c.org/howto/features/allow-unlisted-profile-choice)
- [dedicated nodepool for a hub on a shared cluster](https://infrastructure.2i2c.org/howto/features/dedicated-nodepool/)
- [scratch buckets](https://infrastructure.2i2c.org/howto/features/buckets)
- [requestor Pays’ access to Google Cloud Storage buckets](https://infrastructure.2i2c.org/topic/features/#requestor-pays-access-to-google-cloud-storage-buckets)
- [gh-scoped-creds](https://infrastructure.2i2c.org/howto/features/github)
### Identified as not documented on the infra docs
- `custom_html` feature from https://github.com/2i2c-org/default-hub-homepage/pull/25
- allow jupyterlab option to show hidden files in browser
```yaml
    singleuser:
      extraFiles:
        jupyter_notebook_config.json:
          data:
            # Allow jupyterlab option to show hidden files in browser
            ContentsManager:
              allow_hidden: true
```
### Re-use
- Note that this hub will use the GitHubOAuthenticator and the credentials of the oauth app still existed in the repo under `enc-jackeddy.secret.values.yaml`
- Also, the GitHub app needed for gh-scoped-creds was still existing, so we are reusing that as well

PR #3282, opened 2023-10-18
Allow JupyterLab to show hidden files

yuvipanda:
This doesn't change any behavior, but just allows JupyterLab to show a 'View -> Show Hidden Files' option that folks can turn on. The only reason it isn't on by default is apparently a historical
accident (https://github.com/jupyterlab/jupyterlab/issues/11304#issuecomment-945466766)

Instead of turning it on piecemeal per-hub, this just turns it on for everyone. There's no reason to not.

Ref https://2i2c.freshdesk.com/a/tickets/1056

### DIFF
--- a/config/clusters/2i2c/jackeddy.values.yaml
+++ b/config/clusters/2i2c/jackeddy.values.yaml
@@ -31,14 +31,6 @@ basehub:
             url: ""
             custom_html: <a href="https://science.nasa.gov/heliophysics/programs/living-with-a-star/">NASA's Living with a Star program</a> and <a href="https://cpaess.ucar.edu/">UCAR/CPAESS</a>
     singleuser:
-      extraFiles:
-        # FIXME: document this feature
-        # Currently only documented in the context of culling https://infrastructure.2i2c.org/sre-guide/manage-k8s/culling/#kernel-culling-configuration
-        jupyter_notebook_config.json:
-          data:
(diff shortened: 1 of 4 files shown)


config/clusters/2i2c/jackeddy.values.yaml
```diff
@@ -31,14 +31,6 @@ basehub:
             url: ""
             custom_html: <a href="https://science.nasa.gov/heliophysics/programs/living-with-a-star/">NASA's Living with a Star program</a> and <a href="https://cpaess.ucar.edu/">UCAR/CPAESS</a>
     singleuser:
-      extraFiles:
-        # FIXME: document this feature
```
sgibson91, line 35: Can we open an issue and link it here? I think FIXME notes in code like this are far too hidden for them to be properly prioritised and actioned

sgibson91 (approved):

config/clusters/2i2c/jackeddy.values.yaml
yuvipanda: @sgibson91 I was not 100% sure what exactly the 'this' refers to, but based on https://github.com/2i2c-org/infrastructure/pull/3255#issue-1937880888 it's clear it's in reference to showing hidden files! So I've opened  https://github.com/2i2c-org/infrastructure/issues/3285. Thanks for the catch.

PR #3308, opened 2023-10-24
[2i2c-uk:lis] Enable profileList

GeorgianaElena:
### Update: waiting for feedback in https://2i2c.freshdesk.com/a/tickets/1066 before merging.

Follow-up to https://github.com/2i2c-org/infrastructure/pull/3302, also for https://2i2c.freshdesk.com/a/tickets/1066.

Matthew from LIS requested a guarantee of 1GB as they are still experiencing crashes, even after the increase to 512GB of guarantee from https://github.com/2i2c-org/infrastructure/pull/3302.

This PR enables profileLists for the lis hub, adding the new guarantee and reducing the limit of the initial one of 512G one from 2GB to 1GB.

I think it would be useful for them to enable https://github.com/jupyter-server/jupyter-resource-usage so they can better understand they mem needs, especially since from the grafana below, it looks like only a few users actually required around or more than 1GB.

This is also why the new server size option, instead of increasing it again.

Screenshot 2023-10-24 at 10 02 40

### DIFF
--- a/config/clusters/2i2c-uk/lis.values.yaml
+++ b/config/clusters/2i2c-uk/lis.values.yaml
@@ -45,10 +45,21 @@ jupyterhub:
       # https://hub.docker.com/r/lisacuk/lishub-base
       name: "lisacuk/lishub-base"
       tag: "581a714"
-    memory:
-      # increase as requested via https://2i2c.freshdesk.com/a/tickets/1066
-      guarantee: 512M
-      limit: 2G
+    # Added profileList per https://2i2c.freshdesk.com/a/tickets/1066
+    profileList:
+      - display_name: "Small: ~512 MB RAM / ~0.5 CPU"
+        slug: mem_512m
+        default: true
+        kubespawner_override:
+          # increase as requested via https://2i2c.freshdesk.com/a/tickets/1066
+          mem_guarantee: 512M
+          mem_limit: 1G
+      - display_name: "Large: ~1 GB RAM / ~0.5 CPU"
+        slug: mem_1g
+        kubespawner_override:
+          # increase as requested via https://2i2c.freshdesk.com/a/tickets/1066
+          mem_guarantee: 1G
+          mem_limit: 2G
   hub:
     config:
       JupyterHub:
--- a/config/clusters/2i2c-uk/lis.values.yaml
+++ b/config/clusters/2i2c-uk/lis.values.yaml
@@ -47,19 +47,42 @@ jupyterhub:
       tag: "581a714"
     # Added profileList per https://2i2c.freshdesk.com/a/tickets/1066
     profileList:
-      - display_name: "Small: ~512 MB RAM / ~0.5 CPU"
-        slug: mem_512m
-        default: true
-        kubespawner_override:
-          # increase as requested via https://2i2c.freshdesk.com/a/tickets/1066
-          mem_guarantee: 512M
-          mem_limit: 1G
-      - display_name: "Large: ~1 GB RAM / ~0.5 CPU"
-        slug: mem_1g
-        kubespawner_override:
-          # increase as requested via https://2i2c.freshdesk.com/a/tickets/1066
-          mem_guarantee: 1G
-          mem_limit: 2G
+      - mem_0_4:
+          display_name: 0.4 GB RAM, upto 3.485 CPUs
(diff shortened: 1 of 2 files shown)


config/clusters/2i2c-uk/lis.values.yaml
```diff
@@ -45,10 +45,21 @@ jupyterhub:
...
+        kubespawner_override:
+          # increase as requested via https://2i2c.freshdesk.com/a/tickets/1066
+          mem_guarantee: 512M
+          mem_limit: 1G
+      - display_name: "Large: ~1 GB RAM / ~0.5 CPU"
+        slug: mem_1g
+        kubespawner_override:
+          # increase as requested via https://2i2c.freshdesk.com/a/tickets/1066
+          mem_guarantee: 1G
+          mem_limit: 2G
```
consideRatio, line 62: I'd like to avoid use of relative words like "small" and "large" because it will make communities use these words that we won't understand without reading up on their configuration first.

I think it would make sense for them to use the existing 4 CPU / 32 GB nodes with node resource allocation script generated requests/limits where the limit is 2x the request. Is the script able to do that atm? Hmmm...

GeorgianaElena: I actually did a deploy with just the sizes as names and it felt very mathematical 😅 
Screenshot 2023-10-24 at 11 01 57

> I think it would make sense for them to use the existing 4 CPU / 32 GB nodes with node resource allocation script generated requests/limits where the limit is 2x the request. Is the script able to do that atm? Hmmm..

I am not familiar with what the script can do yet 😬

consideRatio (approved):
This is approved as is, including "Small" and "Large" mentions, but also with currently deployed change. Let's iterate over time on this instead!

GeorgianaElena:
@consideRatio, thank you! I was thinking just now, if we are moving to profileLists, why not add their initial memory guarantee as another, even smaller option (256MB)?

config/clusters/2i2c-uk/lis.values.yaml
consideRatio, line 62: I think this is a good resolution for now. They get requests == limits for memory, they still probably wont use more than one 4 CPU / 32 GB node in the cluster as they have done historically, and they are presented with something that is simple to understand.

We can optimize cloud costs from this going onwards by oversubscribing a bit, requesting less memory than the limit etc.

```suggestion
      - mem_0_4:
          display_name: 0.4 GB RAM, upto 3.485 CPUs
          kubespawner_override:
            mem_guarantee: 457785792
            mem_limit: 457785792
            cpu_guarantee: 0.054453125
            cpu_limit: 3.485
            node_selector:
              node.kubernetes.io/instance-type: n2-highmem-4
      - mem_0_9:
          display_name: 0.9 GB RAM, upto 3.485 CPUs
          kubespawner_override:
            mem_guarantee: 915571584
            mem_limit: 915571584
            cpu_guarantee: 0.10890625
            cpu_limit: 3.485
            node_selector:
              node.kubernetes.io/instance-type: n2-highmem-4
      - mem_1_7:
          display_name: 1.7 GB RAM, upto 3.485 CPUs
          kubespawner_override:
            mem_guarantee: 1831143168
            mem_limit: 1831143168
            cpu_guarantee: 0.2178125
            cpu_limit: 3.485
            node_selector:
              node.kubernetes.io/instance-type: n2-highmem-4
      - mem_3_4:
          display_name: 3.4 GB RAM, upto 3.485 CPUs
          kubespawner_override:
            mem_guarantee: 3662286336
            mem_limit: 3662286336
            cpu_guarantee: 0.435625
            cpu_limit: 3.485
            node_selector:
              node.kubernetes.io/instance-type: n2-highmem-4
```

consideRatio:
@GeorgianaElena I suggest we go for https://github.com/2i2c-org/infrastructure/pull/3308#issuecomment-1776735799 and ask them for feedback

consideRatio (approved):

GeorgianaElena:
### Update

Waiting for feedback before merging.

The profile list looks like:

Screenshot 2023-10-24 at 13 07 10

GeorgianaElena:
I will close this PR to de-clutter and will open up again if/when agreement about this one.

PR #3311, opened 2023-10-24
Rename researchdelight hub to showcase

sgibson91:
- https://github.com/2i2c-org/infrastructure/issues/3279

### Things changed in this PR

- filenames that used researchdelight
- specific parts of config that references research delight
- sets up a redirect from researchdelight.2i2c.cloud to showcase.2i2c.cloud

### Things not changed in this PR

- Kubernetes annotations that still use researchdelight. This PR does not update terraform in any way. These annotations cannot be seen by a user anyway so should not be problematic.
- Team names that contain research delight. Awaiting new teams.
- Image names that contain research delight. Awaiting new image.

### What to do AFTER approval, but BEFORE merge

- [x] Update A/CNAME records in Namecheap for new name
- [x] Update URLs in GitHub OAuth app

### What to do AFTER merge

- [ ] delete the old `researchdelight` namespace in helm

PR #3312, opened 2023-10-24
[support] Set allowSnippetAnnotations to true

sgibson91:
Deploy of #3311 failed with complaints of using snippet directives. This PR changes a default to permit such directives, but the default was set to false as the result of a CVE. So we should execute caution here.


consideRatio (approved):
:+1:

Issue #2541, opened 2023-05-16
Document how to pre-pull images to pre-started nodes during events


consideRatio:
### Update October 2023

To pre pull images to pre started nodes is a valid use case, but https://github.com/2i2c-org/infrastructure/issues/2344 is prioritized for now over this.

---

In #2539 I wanted to ensure that all nodes that I pre-started got images pulled to them ahead of users arriving. This may be automatic if the hub is setup with `prePuller` in the jupyterhub chart, but we don't always have that enabled.

We should not have any jupyterhub `prePuller` enabled when using the configurator, because then the configurator will be used to say what image to use which the jupyterhub chart's prePuller may not be aware of - and users may end up needing to wait for the prePuller image be downloaded before their actually used image when a new node starts up.

### Situation background

When nodes are pre-started before users have arrived to them, its a plus if they have the image to be used by users already downloaded when they arrive.

Challenges are that if a pre-started node starts downloading any unrelated image, a user may need to wait for that. Downloading images to a pre-started node is only a guaranteed win if they are fully downloaded before the user arrive. Enabling pre-pulling of images of no relevance to an arriving user can otherwise instead slow down startup time as images are downloaded one by one by the k8s node.

Issues can arise if:
- A jupyterhub chart profileList allows users to start a number of different images
- The configurator bypasses configuration on images to be used
- Nodes are not started ahead of time users arrive, but just as users arrive and can't schedule

### Pre-warming for an event where configurator is used

Assume the configurator is used, prePuller is disabled, nodes are pre-started. Then one can do the following trick to pre-pull images:

1. Start prePulling what's configured in `singleuser.image`, and ensure that is relevant to pre-pull by inspecting the configurator. This can be done [with a config change like this](https://github.com/2i2c-org/infrastructure/pull/2539/commits/45cd5cfc9324bd7b841ebbc8b6e4c6be13d83da3)
2. Stop prePulling again as the hub chooses an image based on the configurator, and may end up needing to pull an unrelated image which for a new user arriving to a fresh started node can slow down startup. This can be done [with a config change like this](https://github.com/2i2c-org/infrastructure/pull/2539/commits/5306b5934581ba496123f2c756f8375cef64e1a2).

### Related

- [z2jh docs about pre-pulling images before users arrive](https://z2jh.jupyter.org/en/stable/administrator/optimization.html#pulling-images-before-users-arrive)

PR #3313, opened 2023-10-24
Stop using continuous prePuller

### LINKED ISSUE #2344: Documented policy about use of JupyterHub chart's prePuller
Deciding if and when to setup prePuller is a tricky topic, and its partially 2i2c specific because:
- we have profileList's that allows users to configure images from a list which isn't standard jupyterhub chart behavior
- we have the configurator allowing users to update their default image

### Action points

- [ ] Establish and document a policy based on our current understanding
- [ ] Consider if some documentation should also be upstreamed

### Related
- [JupyterHub's chart's configuration reference](https://z2jh.jupyter.org/en/stable/resources/reference.html#prepuller)
- [JupyterHub's chart's documentation about pulling images ahead of time users arrive](https://z2jh.jupyter.org/en/stable/administrator/optimization.html#pulling-images-before-users-arrive)
- In https://github.com/2i2c-org/infrastructure/commit/a93926a037c84c3613b1c70ce036213418941874 I see it as a bug to have prePuller enabled, even though it was good for the a workshop when several nodes were started ahead of time.
- In this comment I clarify when I think we should not have prePuller activated any more to avoid a bug like above:
  https://github.com/2i2c-org/infrastructure/blob/f2eb7aa07670941260952aa56b4f2c1c05bdf6a1/config/clusters/nasa-veda/common.values.yaml#L53-L63

consideRatio:
## Update 2023-10-30

Based on feedback from Yuvi in https://github.com/2i2c-org/infrastructure/pull/3313#issuecomment-1777280535, I've pivoted to only disable the continuous prePuller in this PR, and not also the hook prePuller. I'm opening a followup PR to this to constrain the use of hook prePuller in another PR to help us avoid issues with it.

## Original PR

Fixes #2344 by inlining comments in basehub's default config that already disables prePuller logic by default, and by disabling both continous and hook prePuller where it is enabled for individual hubs.

---

prePuller is a somewhat complex machinery to reduce startup times in favorable conditions - conditions that also could change outside our chart configuration via the configurator for example. In unfavorable conditions, use of prePuller could instead increase startup times. This fragility and complexity has caused me some stress and cognitive load that I don't want to introduce to others, but instead help us all avoid.

This make me think __*the right call is to systematically disable both use of continious prePuller and hook prePuller*__.

<details>
<summary>Continue reading for some additional details</summary>

### Disabling continuous pulling

The continuous prePuller could be beneficial if nodes are pre-started before a user pod arrives to them. It was initially designed with intent to be used with `user-placeholder` / `node-placeholder` pods, something we are not using in any of our clusters.

We have successfully made use of the continuous prePuller in the past when pre-starting many nodes for events. With the introduction of the practice to schedule multiple users per node, this is not as relevant. I think if we want to use it for that time, we can do a one off image pull just after pre-starting nodes instead of having it enabled by default - this is to be clarified by #2541.

### Disabling hook pulling

The hook puller is active during `helm ugprade`, blocking the actual upgrade until pulling has been completed. I see some value in it compared to the current use of the continuous prePuller, but I think we are better off avoiding its complexity we introduce.

Read the [inlined config notes about disabling it](https://github.com/2i2c-org/infrastructure/pull/3313/files#diff-301db94aa87a33a484be445796919615b3eecdc883900e976be6870e645f6d48R160-R173) for some additional details.

### After disabling, these are no longer relevant caveats

When I've heard "slow startup" be mentioned in support tickets etc, I've often ended up considering caveats below to ensure use of prePuller wasn't causing it. If we disable prePulling systematically with this PR, none of us need to consider caveats like below going onwards to be the cause of slow startups! We could focus on fixing issues caused by other things without even first ruling out its the prePuller causing it.

- prePuller (continuous mainly) used for a hub scheduling user pods on nodes used by other hubs is likely to disturb other hubs, using a dedicated node pool has other caveats
- prePuller doesn't pull all images we configure for use - not those via the configurator or profile_options in some profile lists
- continuous prePuller pulling multiple images can even when pulling the right images delay startup times for non-pre-started nodes
- hook prePuller requires use of ClusterRoleBinding resources, something that some k8s admins isn't granted if they exercise the right to replicate in a cluster they only control a namespace within

</details>

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -143,9 +143,34 @@ jupyterhub:
           memory: 64Mi
         limits:
           memory: 1G
+  # prePuller is about pulling a one or more images identified via chart
+  # configuration, including singleuser.image, singleuser.profileList entries
+  # with a dedicated image, but not profileList entries with images' specified
+  # via profile_options.
   prePuller:
+    # continuous prePuller leads to the creation of a DaemonSet that starts a
+    # pod on each node to pull images.
+    #
+    # It is disabled as its only relevant for nodes started before user pods
+    # gets scheduled on them, in other cases it could delay startup and isn't
+    # expected to reduce startup times.
+    #
     continuous:
       enabled: false
+    # hook prePuller leads to the creation of a temporary DaemonSet and a pod
+    # awaiting pulling to complete before `helm upgrade` starts its main work.
+    #
+    # It is disabled as it adds notable complexity for a smaller benefit when
+    # correctly adopted. The added complexity includes:
+    #
+    # - risk of misconfiguration making image pulls not actually needed
+    # - risk of broken expectations and additional cognitive load
+    # - risk of causing significantly longer `helm upgrade` commands slowing
+    #   down our CI system
+    # - ClusterRoleBinding resources are needed for the image-awaiter Pod
+    #   involved, a resource that requires the highest k8s cluster permission
+    #   otherwise possibly not needed to deploy basehub
+    #
     hook:
       enabled: false
   proxy:
@@ -267,7 +292,7 @@ jupyterhub:
           MappingKernelManager: *server_config_mapping_kernel_manager
           NotebookApp: *server_config_server_app
           BaseFileIdManager: *server_config_base_file_id_manager
-    startTimeout: 600 # 10 mins, because sometimes we have too many new nodes coming up together
+    startTimeout: 600 # 10 mins, node startup + image pulling makes it relevant
     defaultUrl: /tree
     image:
       name: jupyter/scipy-notebook
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -292,7 +292,7 @@ jupyterhub:
           MappingKernelManager: *server_config_mapping_kernel_manager
           NotebookApp: *server_config_server_app
           BaseFileIdManager: *server_config_base_file_id_manager
-    startTimeout: 600 # 10 mins, node startup + image pulling makes it relevant
+    startTimeout: 600 # 10 mins, node startup + image pulling sometimes takes more than the default 5min
     defaultUrl: /tree
     image:
       name: jupyter/scipy-notebook
--- a/config/clusters/2i2c/climatematch.values.yaml
+++ b/config/clusters/2i2c/climatematch.values.yaml
@@ -1,11 +1,7 @@
 jupyterhub:
-  # pre-puller is necessary as the image is pretty big, and
-  # pulling during first user spawn might cause timeouts and poor user
-  # experience. Also helps with node pre-warming. This works ok here
-  # because they have a dedicated nodepool.
   prePuller:
-    continuous:
-      enabled: true
+    # hook prePuller shouldn't be enabled when configuring images in any other
+    # way than singleuser.image
     hook:
       enabled: true
   singleuser:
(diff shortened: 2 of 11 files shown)


yuvipanda:
Would you consider letting the hook prepuller be for what i consider the utoronto case? The specifics are:

1. configurator is not used
2. No profile list 
3. The images are fairly large

In particular, without this, if we change the image tag here on the repo, the next user who starts the server after a deploy will be responsible for pulling the image and get a very slow startup. With this enabled, that cost is rolled into the helm deploy. 

This was very poor UX for end users at berkeley, given that it happens “randomly” from their perspective,  which is what led to the creation of the hook prePuller.

consideRatio:
I've updated this PR to only disable the continuous prePuller for now, and I've opened a separate PR (#3341, currently in draft) to constrain the use of hook prePuller.

consideRatio:
@GeorgianaElena @sgibson91 do you agree on this change?

helm-charts/basehub/values.yaml
```diff
@@ -143,9 +143,34 @@ jupyterhub:
...
+  # with a dedicated image, but not profileList entries with images' specified
+  # via profile_options.
   prePuller:
+    # continuous prePuller leads to the creation of a DaemonSet that starts a
+    # pod on each node to pull images.
+    #
+    # It is disabled as its only relevant for nodes started before user pods
+    # gets scheduled on them, in other cases it could delay startup and isn't
+    # expected to reduce startup times.
+    #
```
GeorgianaElena, line 157: @consideRatio, what if we add these comments as a descriptions in the schema file?

I figure, the schema is where we currently "document" basehub defaults and having these in there kind of makes sense? Maybe in the future we can use that to generate some sort of API-like docs?

consideRatio: > Maybe in the future we can use that to generate some sort of API-like docs?

I agree on letting basehub generate config reference docs based on descriptions for basehub added config in values.schema.yaml like done by z2jh!

I favor the approach of inlining a comment next to the changed default value in this case though.

1. All charts require a values.yaml file, so helm chart users likely look in them to learn about default values. If we separate remarks about default values into a schema file / generated docs, they may be less findable from the values.yaml file.
2. We don't yet have generated docs from the basehub's values.schema.yaml
3. If/when we generate docs from the schema, should the schema really redefine parts of the schema of dependency charts to provide a description that the default value is changed?

GeorgianaElena (approved):

yuvipanda:
@consideRatio thoughts on how this would interact with how we 'pre-warm' nodes for an event? In those cases, I believe we increase the size of the nodepool to ensure faster startup, and the continuous prepuller would be required there (subject to the same constraints outlined in https://github.com/2i2c-org/infrastructure/pull/3313#issuecomment-1777280535).

consideRatio:
> @consideRatio thoughts on how this would interact with how we 'pre-warm' nodes for an event?

I've opened #2541 about this, my idea is that we do a single one-off pre-pull to pre-started nodes if we pre-start nodes.

helm-charts/basehub/values.yaml
```diff
@@ -267,7 +292,7 @@ jupyterhub:
           MappingKernelManager: *server_config_mapping_kernel_manager
           NotebookApp: *server_config_server_app
           BaseFileIdManager: *server_config_base_file_id_manager
-    startTimeout: 600 # 10 mins, because sometimes we have too many new nodes coming up together
+    startTimeout: 600 # 10 mins, node startup + image pulling makes it relevant
```
yuvipanda, line 295:
```suggestion
    startTimeout: 600 # 10 mins, node startup + image pulling sometimes takes more than the default 5min
```

yuvipanda (approved):

consideRatio:
Thank you for reviewing @GeorgianaElena and @yuvipanda!!!

Issue #3332, opened 2023-10-26
HHMI updates for demo


jmunroe:
In preparation for a meeting with the HHMI community, I making a few changes to the HHMI hub deployment.

- [x] Fixed broken logo image
- [x] Add support for multiple images
- [x] Create a binder-like repo for a Spyglass/NWB image

jmunroe:
The logo was originally pointing to https://www.hhmi.org/sites/default/files/Logos/JPG/HHMI-vertical-signature-color.jpg which no longer appears to be a valid URL.

All the HHMI logos are available for download as a .zip from https://www.hhmi.org/about/for-host-institutions#logo

I know we have had questions before of "where do we store a logo file" for a community if it not available from a public URL.  I know I can do it by dropping it a gist, or embedding it in a GitHub comment.  This time round I experimented with putting in the 2i2c-Team Google Drive under the Partnerships and making the link shareable:

https://drive.google.com/file/d/1tg5PRRT3_VjDaxq6Ax_JA3S92FuQ_NFt/view?usp=drive_link

This is not a link that can be embedded directly in an img tag.  I found that changing the link to

 https://drive.google.com/uc?export=view&id=1tg5PRRT3_VjDaxq6Ax_JA3S92FuQ_NFt

does allow it to be referenced directly within the `common.values.yaml` `logo_url` key.

jmunroe:
Most of communities that use ProfileList support both multiple machine types AND multiple images.  For this community (and for my demo on Friday), I wanted to focus only on presenting multiple images options and not first requiring the user to choose what size of machine to run this container on (thus leaving that as fixed, shared node for all users on this hub).  Here's the design I'm using for the moment:

<img width="583" alt="image" src="https://github.com/2i2c-org/infrastructure/assets/6181563/6a742c66-5883-4ed3-b639-b2fc9af966ff">

jmunroe:
For the `Spyglass-NWB` image, I want make available an environment that is based off the `environment.yml` file found in https://github.com/LorenFrankLab/spyglass

Assuming the demo goes well, I expect that this image repo will eventual migrate to something like
https://github.com/LorenFrankLab or https://github.com/hhmi or possibly a new org.  In the very short term, I'll create it within https://github.com/2i2c-org

jmunroe:
[...]

PR #3333, opened 2023-10-26
set up for a HHMI deploy re: #3332

jmunroe:


jmunroe:
This will go through additional iterations after the meeting tomorrow afternoon with HHMI.

I've already tested and deployed.  To keep the changes persistent, I feel I should self-merge this.  Is that the correct policy? (I think so -- but I don't want to overstep)

GeorgianaElena (approved):
@jmunroe, we have the self-merging policy defined at https://infrastructure.2i2c.org/contributing/code-review/#self-merging-as-a-2i2c-engineer.

I suggest documenting this particular case there also if you believe qualifies.

Also, this PR looks good to me 🚀

consideRatio:
Nice @jmunroe!!!

Is this closing #3332 btw? Then you can add a `fixes #3332` or `closes #3332` in the PR description to have it close that issue. If you put it there instead of in the title, there will be a clickable link as well btw.

jmunroe:
> @jmunroe, we have the self-merging policy defined at https://infrastructure.2i2c.org/contributing/code-review/#self-merging-as-a-2i2c-engineer.
> 
> I suggest documenting this particular case there also if you believe qualifies.
> 
> Also, this PR looks good to me 🚀

Thanks @GeorgianaElena .  I think the changes in this PR go beyond the guidelines for a self-merge so I will leave as they are.

jmunroe:
> Nice @jmunroe!!!
> 
> Is this closing #3332 btw? Then you can add a `fixes #3332` or `closes #3332` in the PR description to have it close that issue. If you put it there instead of in the title, there will be a clickable link as well btw.

Aha! I knew there was something like that but didn't know that the keywords were "fixes" and "closes". Thanks!

PR #3337, opened 2023-10-28
deployer: add info generating commands under resource-allocation

### LINKED ISSUE #3330: Add script to collect information about daemonsets cpu/memory requests in various clusters
The resource allocation script added in #3030 would benefit from being used with information about daemonset's requests cpu/memory in various clusters one way or another. This issue represents the work to add a standalone script generating such overview.

### LINKED ISSUE #3338: Refine script to collect information about instances capacity
In #3030 we got the `deployer generate resource-allocation update-node-info` command. I'd like to incrementally refine it for the following reasons:

1. Make it easy to collect all our clusters available node's reported `capacity` / `allocatable` without knowing what cluster has what node's running currently
2. Validate our assumption that instance type's `capacity` and `allocatable` is more or less static, and if its not fully static help us overview the variation by keeping track of at least the lowest and highest values.

consideRatio:
This PR adds two commands of relevance towards refining the resource-allocation script to meet more needs. The added commands are updating their separate .yaml file's with collected information.

- Closes #3338 
- Closes #3330 

## Script overview

- `daemonset-requests` summarizes the requests from all daemonsets, and clarifies what daemonsets contributed to the requests and what daemonset didn't.

  The script makes us better understand and clarify differences observed. It helped me verify that the current overhead is mostly cloud provoider specific, to some extent cluster specific (feature and k8s version differences), and to a small extent instance type/node pool specific (gpu drivers).
  
  A caveat of this implementation strategy is that it doesn't intelligently rule out that a daemonset may not only be schedule on some nodes, for example core nodes only, or on nodes with gpus. I've filtered out the gpu related daemonsets manually, and not observed any core node pool only daemonsets.
- `instance-capacities` scrapes the `kubectl get node` reported `capacity` and `allocatable` (allocatable capacity), and saves both the lowest and highest reported value to help us overview the spread of these values across clusters. It turns out that there is a slight spread, but its very low.

## With `update-node-info`, why these commands?

These are iterations on `update-node-info`, but they are added separately as `update-node-info` directly hooks into the `resource-allocation choices` command, making an update to it break the existing behavior.

### DIFF
--- a/deployer/README.md
+++ b/deployer/README.md
@@ -88,8 +88,8 @@ The `deployer.py` file is the main file, that contains all of the commands regis
 │   │   |   ├── decision.py
 │   │   |   └── jobs.py
 |   |   └── resource_allocation
-│   │       ├── daemonset_overhead.py
-│   │       ├── daemonset_overhead.yaml
+│   │       ├── daemonset_requests.py
+│   │       ├── daemonset_requests.yaml
 │   │       ├── generate_choices.py
 │   │       ├── instance_capacity.py
 │   │       ├── instance_capacity.yaml
@@ -268,8 +268,8 @@ This sub-command can be used to generate the resource allocation choices for giv
 ##### `generate resource-allocation choices`
 This generates a custom number of resource allocation choices for a certain instance type, depending on a certain chosen strategy that can be used in the profile list of a hub.
 
-##### `generate resource-allocation daemonset-overhead`
-Updates `daemonset_overhead.yaml` with an individual cluster's DaemonSets with running pods combined requests of CPU and memory, excluding GPU related DaemonSets.
+##### `generate resource-allocation daemonset-requests`
+Updates `daemonset_requests.yaml` with an individual cluster's DaemonSets with running pods combined requests of CPU and memory, excluding GPU related DaemonSets.
 
 ##### `generate resource-allocation instance-capacity`
 Updates `instance_capacity.yaml` with instance types' total and allocatable capacity reported by `kubectl get node`.
--- a/deployer/README.md
+++ b/deployer/README.md
@@ -269,10 +269,20 @@ This sub-command can be used to generate the resource allocation choices for giv
 This generates a custom number of resource allocation choices for a certain instance type, depending on a certain chosen strategy that can be used in the profile list of a hub.
 
 ##### `generate resource-allocation daemonset-requests`
-Updates `daemonset_requests.yaml` with an individual cluster's DaemonSets with running pods combined requests of CPU and memory, excluding GPU related DaemonSets.
+Updates `daemonset_requests.yaml` with an individual cluster's DaemonSets' requests summarized.
+
+Only DaemonSet's with running pods are considered, and GPU related DaemonSets (with "nvidia" in the name) are also ignored.
+
+To run this command for all clusters, `xargs` can be used like this:
+
+    ls config/clusters | xargs -I {} deployer generate resource-allocation daemonset-requests {}
 
 ##### `generate resource-allocation instance-capacity`
-Updates `instance_capacity.yaml` with instance types' total and allocatable capacity reported by `kubectl get node`.
+Updates `instance_capacity.yaml` with an individual cluster's running instance types' total and allocatable capacity.
+
+To run this command for all clusters, `xargs` can be used like this:
+
+    ls config/clusters | xargs -I {} deployer generate resource-allocation instance-capacity {}
 
 ##### `generate resource-allocation node-info-update`
 This updates the json file `node-capacity-info.json` with info about the capacity of a node of a certain type. This file is then used for generating the resource choices.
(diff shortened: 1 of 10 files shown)


deployer/README.md
```diff
@@ -264,6 +268,12 @@ This sub-command can be used to generate the resource allocation choices for giv
 ##### `generate resource-allocation choices`
 This generates a custom number of resource allocation choices for a certain instance type, depending on a certain chosen strategy that can be used in the profile list of a hub.
 
+##### `generate resource-allocation daemonset-overhead`
+Updates `daemonset_overhead.yaml` with an individual cluster's DaemonSets with running pods combined requests of CPU and memory, excluding GPU related DaemonSets.
```
GeorgianaElena, line 272: This is currently a bit hard to understand. Maybe expand it a bit more with info from the PR's top comment?

```suggestion
This script updates information about a specific cluster in the `daemonset_overhead.yaml` file.

This file contains information about cpu/memory requests made by
DaemonSets with running pods in _all_ of our clusters but excluding GPU related DaemonSets.
```

deployer/README.md
```diff
@@ -264,6 +268,12 @@ This sub-command can be used to generate the resource allocation choices for giv
...
+
+##### `generate resource-allocation instance-capacity`
+Updates `instance_capacity.yaml` with instance types' total and allocatable capacity reported by `kubectl get node`.
```
GeorgianaElena, line 275: From what I understand this command requires a cluster, and it uses that cluster to get the running nodes and updates the file with info about that or those running nodes, right?

From this description I understand that it updates all instance types and I wouldn't expect the command to require a cluster.

So maybe rephrase to:
```suggestion
This script updates the information in the `instance_capacity.yaml` file about the instance types found running in a given cluster.

This file contains the total and allocatable capacity as reported by `kubectl get node` of the instance types'  used across our clusters.
```

deployer/commands/generate/resource_allocation/daemonset_overhead.py
```diff
@@ -0,0 +1,152 @@
...
+        text=True,
+    )
+    version_info = json.loads(output)
+    server_version_info = version_info["serverVersion"]["gitVersion"]
+    if "gke" in server_version_info:
+        return "gke", server_version_info
+    if "eks" in server_version_info:
+        return "eks", server_version_info
+    else:
+        return "aks", server_version_info
```
GeorgianaElena, line 42:
```suggestion
    return "aks", server_version_info
```

deployer/commands/generate/resource_allocation/daemonset_overhead.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+def get_daemon_sets_requests():
+    """
+    Returns a list of info about DaemonSets with pods desired to be scheduled on
```
GeorgianaElena, line 47:
```suggestion
    Returns a list of dicts with info about DaemonSets with pods desired to be scheduled on
```

deployer/commands/generate/resource_allocation/daemonset_overhead.yaml
```diff
@@ -0,0 +1,180 @@
...
+# - Cloud providers' managed k8s provides different DaemonSets by default
+# - DaemonSets may be coupled to managed k8s features (calico-node)
+# - DaemonSets' requests may be coupled to managed k8s version (netd)
+# - DaemonSets may have a vertical autoscaler changing requests dynamically over
+#   time if needed (calico-node-vertical-autoscaler)
+# - We may deploy or change a DaemonSet's requests over time (support-cryptnono,
+#   support-prometheus-node-exporter)
+#
+# This file isn't updated by automation, but can easily be updated by manually
+# by running a command once for each cluster:
```
GeorgianaElena, line 18:
```suggestion
# This file isn't updated by automation, but can easily be updated by manually
# running a command once for each cluster:
```

GeorgianaElena (approved):
🚀 Thanks @consideRatio

deployer/README.md
consideRatio: The command will not collect information from each cluster. I considered doing that, but I don't think the complexity added to `deployer` is justified when we can easily invoke it for all clusters one at the time with a command like `ls config/clusters | xargs -I {} deployer generate resource-allocation daemonset-overhead {}`.

I'll mention that in the docs here!

deployer/README.md
consideRatio: I'll add a commit to refine this part similar to your suggestion, both here and in the .py file where this text is also duplicated as a docstring!

consideRatio:
Thank you for reviewing this @GeorgianaElena! I've not resolved all comments, but will get it done soonish!

consideRatio:
- Rebased
- renamed command instance-capacity to instance-capacities
- renamed command daemonset-overhead to daemonset-requests
- updated docstrings to reflect feedback from @GeorgianaElena

---

I'll go for a merge here at this point!

Issue #3307, opened 2023-10-24
Iterate on what instance sizes we ensure is setup for all clusters


consideRatio:
This is a followup to https://github.com/2i2c-org/infrastructure/pull/3304 about node pools and associated instance sizes we ensure is setup for all our clusters to enable us flexibly make changes without needing additional infra work.

In https://github.com/2i2c-org/infrastructure/pull/3304#discussion_r1368816014 asked for another set of instance sizes than initially suggested in the PR. This issue represents looking back and refining the decision in the future. What set of instance sizes do we think makes sense to ensure all clusters have available?

### Context

- #3304
- #3256
- #3176

yuvipanda:
As another data point for why I think we should be using much smaller instances, I heard this from @jmunroe today:

> I found that during the course itself, the cloud costs are running is $1-$2 / user / day for two-week course itself but rise to more like $5-$10/user/day when the participants are doing project-related work. The tricky bit is during the course itself the number of unique users is a factor of ten larger than those who appear to be using the service for project work after the course.

I think in general our user node sizes are far too big, and we should definitely make them much smaller. The utilization information I provided in https://github.com/2i2c-org/infrastructure/pull/3304#discussion_r1368816014 is based upon reducing the total cost for things like this here, as 5-10$/user/day is an order of magnitude worse than what it should be.

jmunroe:
I have not looked into the details of what each unique user is doing. Assuming that each user is always using the the same share of node, I would though the increased cost is to due to usage being distributed throughout a 24 hour period. 24 user using a hub for the same one hour will have a much cheaper cost than each of those 24 using the same machine 1 hour each spread out over a day.  Or perhaps unique users are using the service for much longer periods when they were doing project work.  

I think that data about usage pattern (and actually amount of RAM used) is in our dataset -- we need to examine it in more detail.

consideRatio:
We went with 4 / 16 / 64 CPU highmem machines.

PR #3344, opened 2023-10-30
daskhub: provide worker resource options for 16CPU/128GB nodes on GKE/EKS

### LINKED ISSUE #2145: Verify cluster creation options for dask_gateway works
In #2144 we have a basic example on how to use dask_gateway the client to create a dask cluster. But, the cluster creation options we are passed doesn't work. They are configured by us as 2i2c admins when setting up the dask-gateway the server via the dask-gateway helm chart here.

https://github.com/2i2c-org/infrastructure/blob/cf7bfc3778001791bd83472b943fc695ea5e314d/helm-charts/daskhub/values.yaml#L154-L214

I assume we have a bug in these options, as I get this message.

image

### LINKED ISSUE #2364: daskhub / leap: optimize dask-gateway options for a 16 CPU / 128 GB mem worker

consideRatio:
All GKE/EKS clusters with daskhubs have `n2-highmem-16`/`r5.4xlarge` nodes setup for them (#2687). This makes us able to easily standardize pre-defined options to choose from for all clusters with regards to how large CPU/memory requests should be made for each pod.

- Fixes #2364
- Fixes #2145

## Review notes

- **daskhub's new default**
  The daskhub chart comes with these options by default now, adjusting to use `n2-highmem-16` or `r5.4xlarge` based on the k8s distribution that is detected automatically.
  With this, jupyter-meets-the-earth no longer specifies its own custom logic.
- **`basehub-cluster-info` ConfigMap added**
  It is what makes us have the `BASEHUB_K8S_DIST` environment variable available for the hub and dask-gateway pod, where the dask-gateway pod relies on it in this PR and the hub pod got it mounted as its likely useful.

  The ConfigMap gets labels aligning with modern seemingly now stable helm chart best practices.
- **Test with `pange/base-notebook:latest` at https://dask-staging.2i2c.cloud**

## UX for users

GKE|EKS (ugly extras resolves by in-image software update)
-|-
image|image

@jbusecke I've deployed this to leap's staging hub for now, if you have feedback its most welcome!


helm-charts/daskhub/values.yaml
```diff
@@ -177,38 +196,118 @@ dask-gateway:
                     chars.append(escaped_hex_char)
             return u''.join(chars)
 
+        # Decide on available instance types and their resource allocation
+        # choices to expose based on cloud provider. For each daskhub hub
+        # managed by 2i2c, there should be these instance types available.
```
GeorgianaElena, line 201: Maybe add a link to the docs https://infrastructure.2i2c.org/topic/infrastructure/cluster-design/#for-nodes-where-user-servers-will-be-scheduled-on and a note that if these ever change this piece of code needs to also be changed?

helm-charts/daskhub/values.yaml
```diff
@@ -177,38 +196,118 @@ dask-gateway:
...
+        instance_types = {
+            "gke": ["n2-highmem-16"],
+            "eks": ["r5.4xlarge"],
+        }
+        resource_allocations = {
+            # n2-highmem-16 nodes in our clusters have 15.89 allocatable cores
+            # and 116.549Gi allocatable memory, and daemonset are expected to
+            # not add more than 400m cores and 800Mi (0.781Gi) memory with some
+            # margin, so we get 15.49 cores and 115.768Gi available for worker
+            # pods to request.
```
GeorgianaElena, line 213: Maybe also add a note about these numbers coming from the deployer resource allocation helper scripts ?

helm-charts/daskhub/values.yaml
```diff
@@ -177,38 +196,118 @@ dask-gateway:
...
+                    default=ra_keys[0],
+                    label="Resources per worker container",
+                ),
+                # The default image is pre-specified by the dask-gateway client
+                # via the env var DASK_GATEWAY__CLUSTER__OPTIONS__IMAGE set on
+                # the jupyterhub user pods
                 String("image", label="Image"),
-                Mapping("environment", {}, label="Environment Variables"),
+                Mapping("environment", {}, label="Environment variables (YAML)"),
+                Integer("idle_timeout", 1800, min=0, label="Idle cluster terminated after (seconds)"),
```
GeorgianaElena, line 306: Does this value needs to be kept in sync with `c.KubeClusterConfig.idle_timeout` defined below?

GeorgianaElena (approved):
This looks good to me @consideRatio! Thank you!

I left a few comments about adding more notes about where certain numbers are coming from, so we can trace back in the future if need be.

Related to docs: do you believe it makes sense to add a small (rather generic) note about these new instance-dependent dask worker options in https://docs.2i2c.org/about/distributions/research/#a-brief-overview-of-scalable-computing-hubs?

helm-charts/daskhub/values.yaml
```diff
@@ -177,38 +196,118 @@ dask-gateway:
...
 
+        # Decide on available instance types and their resource allocation
+        # choices to expose based on cloud provider. For each daskhub hub
+        # managed by 2i2c, there should be these instance types available.
+        #
+        cloud_provider = os.environ["BASEHUB_K8S_DIST"] # gke, eks, or aks
+        instance_types = {
+            "gke": ["n2-highmem-16"],
+            "eks": ["r5.4xlarge"],
+        }
```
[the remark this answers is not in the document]
consideRatio: Yes we don't have any AKS daskhubs yet so I opted to avoid trying to pre-optimize this for now. Its better that it breaks than is setup without this for the first future AKS daskhub anyhow I think.

helm-charts/daskhub/values.yaml
consideRatio: This is related to the dask worker nodes, which are undocumented to be systematically set at 16 CPU nodes.

Since this is a controversial topic, I'd like to not touch it further. Yuvi requested that #3307 use smaller nodes etc.

helm-charts/daskhub/values.yaml
consideRatio: I don't think it needs, but I figured it made sense. I'll make a note to help them be kept in sync.

helm-charts/daskhub/values.yaml
GeorgianaElena: Fair enough! Thank you!

helm-charts/daskhub/values.yaml
GeorgianaElena: Thank you!

helm-charts/daskhub/values.yaml
consideRatio: I'll open a dedicated PR to follow this up

consideRatio:
Rebased - I'll go for a merge at this point!

PR #3348, opened 2023-10-31
basehub: make use of configurator opt-out

consideRatio:
Extracted from #3341 to enable isolated review of this. Merging this shouldn't have any impact because no hub is opt-out as part of making it possible to opt-out.

The current policy is that we have enabled configurator for all hubs unconditionally. I don't think it should be unconditionally, but I also don't think we have quick agreement all conditions when it shouldn't be used. Due to that, I suggest that we clarify when to opt-out of using the configurator, step by step in other PRs.

### DIFF
--- a/helm-charts/basehub/values.schema.yaml
+++ b/helm-charts/basehub/values.schema.yaml
@@ -282,6 +282,7 @@ properties:
           - cloudResources
           - 2i2c
           - auth
+          - jupyterhubConfigurator
         properties:
           auth:
             type: object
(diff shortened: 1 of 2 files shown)


GeorgianaElena (approved):
I believe being able to disable the configurator makes sense and it helps to reduce confusion for some cases that are very straightforward (like profile lists)

consideRatio:
I rebased this to handle a merge conflict and added and adjusted docstrings a bit. I did a test on a staging hub to verify I didn't mess things resolving the merge conflict also. I'll go for a merge at this point.

Issue #3240, opened 2023-10-06
Help cleanup openscapes access list


yuvipanda:
@erinmr reached out about helping clean up the openscapes user access list. It's currently 800+ folks manually managed, and a bit unwieldy.

Openscapes was set up before our understanding of how to easily manage GitHub authentication solidifed, and I think this is a good time to move 'em forward.

```[tasklist]
- [x] Create a new GitHub Org to be used for access control (to be done by openscapes folks)
- [x] Once done, we switch authentication to `GithubOAuthenticcator` and use org based access control (done by 2i2c)
- [x] Add whoever *still* needs access to this github org (to be done by openscapes folks)
- [ ] We also just delete all existing users from the hub db as cleanup (done by 2i2c)
```

This isn't disruptive because usernames will still be the same (as we're switching to github directly than github via cilogon). It allows easier long term maintenance of access control (via GitHub org), as well as allowing us to show different profiles to subsets of users based on GitHub *team* membership. For temporary workshops and stuff, we can also grant access for the duration of the workshop to a different GH team, that can be then be revoked when they no longer need access.

The question to be determined is what happens to the home directories of all the users who will no longer have access? The openscapes folks must make a choice here before we can proceed with this. The options are:

1. Delete it all
2. Zip it up and give it to the openscapes folks, who can give it to users *if they ask*

yuvipanda:
Note that an existing GitHub org with a team can also be used instead :)

erinmr:
Hi @yuvipanda - this looks great. We know our list of users to remove, let me know how best to get this to you all. I think we'd want to use teams set up in the NASA-Openscapes Github org (unless there is some limit b/c we're talking about ~300 people who will likely maintain access)

yuvipanda:
@erinmr I think you should just add the people you wanna *keep* to specific teams in the `NASA-Openscapes` org, and then we can just delete everyone else. 

What do you wanna do with the home directories of people who will no longer have access?

erinmr:
What's the cost of keeping home directories for folks we remove in terms of size and $ range?

erinmr:
@yuvipanda - I am not seeing options for bulk addition to Github teams. We are looking at adding about 300+ to a team. I can do this manually, but don't want to. They also will need to accept the invitation to the team which is another step that I'm not sure we want to add. Let me know what you think.

yuvipanda:
@erinmr I can probably look into a script that could do bulk addition.

[...]

PR #3357, opened 2023-11-01
openscapes: Setup GitHub Auth for openscapes staging

yuvipanda:
- Switching to using orgs & teams for authentication
- Don't allow hand-entered usernames access anymore

Ref https://github.com/2i2c-org/infrastructure/issues/3240

### DIFF
--- a/config/clusters/openscapes/common.values.yaml
+++ b/config/clusters/openscapes/common.values.yaml
@@ -51,7 +51,6 @@ basehub:
         GitHubOAuthenticator:
           oauth_callback_url: https://staging.openscapes.2i2c.cloud/hub/oauth_callback
           allowed_organizations:
-            - 2i2c-org:hub-access-for-2i2c-staff
             - NASA-Openscapes:workshopaccess-2i2c
             - NASA-Openscapes:longtermaccess-2i2c
             - NASA-Openscapes:championsaccess-2i2c


yuvipanda:
Will wait on @erinmr to provide list of orgs / teams to allow access

erinmr:
GitHub teams are: 

- https://github.com/orgs/NASA-Openscapes/teams/workshopaccess-2i2c
- https://github.com/orgs/NASA-Openscapes/teams/longtermaccess-2i2c
- https://github.com/orgs/NASA-Openscapes/teams/championsaccess-2i2c

yuvipanda:
ok, @erinmr has granted access and folks should be able to login to staging.openscapes.2i2c.cloud with these teams! We'll test then merge.

config/clusters/openscapes/common.values.yaml
```diff
@@ -47,41 +47,17 @@ basehub:
...
-          #          access any more. In practice the user will still have
-          #          access if it had logged in once before, as it then exists
-          #          in JupyterHub's database of users.
-          #
-          allow_existing_users: True
+          authenticator_class: github
+        GitHubOAuthenticator:
+          oauth_callback_url: https://staging.openscapes.2i2c.cloud/hub/oauth_callback
+          allowed_organizations:
+            - 2i2c-org:hub-access-for-2i2c-staff
```
consideRatio, line 54: With oauthenticator 16, we don't need this - its resolved by:

```yaml
        add_staff_user_ids_to_admin_users: true
        add_staff_user_ids_of_type: "github"
```

```suggestion
```

consideRatio (approved):

config/clusters/openscapes/common.values.yaml
consideRatio: Oh wait... I'll confirm my thoughts first. I made a search to confirm it by it was made incorrectly.

yuvipanda: Thanks @consideRatio, I'll hold off for now.

yuvipanda: Thanks @consideRatio. Just to verify, is that true when `OAuthenticator.allow_existing_users` is not set to True as well?

consideRatio: https://github.com/2i2c-org/infrastructure/pull/3358 makes us 100% consistent

consideRatio: > Thanks @consideRatio. Just to verify, is that true when `OAuthenticator.allow_existing_users` is not set to True as well?

Yes! With Oauthenticator 16 - a user allowed in any way is allowed, the union of ways to allow users as compared to the intersection.

`OAuthenticator.allow_existing_users` is one way to allow users, `admin_users`, `allowed_users`, `allowed_organizations` are other ways as well. This PR looks correct to me with the intent I've understood.

yuvipanda: Thanks @consideRatio

PR #3360, opened 2023-11-01
openscapes: Use GitHubOAuthenticator for prod as well

yuvipanda:
This also fixes an issue in https://github.com/2i2c-org/infrastructure/pull/3357, where config that should've been in staging.values.yaml was instead in common.values.yaml. I cancelled the run even though I only caught it after I hit merge, so no harm done.

Ref https://github.com/2i2c-org/infrastructure/issues/3240


yuvipanda:
Gotta merge https://github.com/2i2c-org/infrastructure/pull/3362 first, that should take care of the merge conflict.

erinmr:
I approve - let's do this :-)

PR #3371, opened 2023-11-03
Disable hook prePuller for three communites using the jupyterhub configurator

consideRatio:
Since callysto, qcl, and 2i2c/climatematch are all using the configurator, they don't have any benefit from the hook prePuller so this PR disables it for these hubs to reduce unessecary complexity and false expectations from the config.

These are commits extracted from #3341 that as a whole can't be merged short term

### DIFF
--- a/config/clusters/callysto/common.values.yaml
+++ b/config/clusters/callysto/common.values.yaml
@@ -9,11 +9,6 @@ nfs:
     # Name of Google Filestore share
     baseShareName: /homes/
 jupyterhub:
-  prePuller:
-    # hook prePuller shouldn't be enabled when configuring images in any other
-    # way than singleuser.image
-    hook:
-      enabled: true
   custom:
     2i2c:
       # add_staff_user_ids_to_admin_users is disabled because the usernames
@@ -37,9 +32,6 @@ jupyterhub:
           name: Callysto
           url: https://www.callysto.ca
   singleuser:
-    image:
-      name: callysto/2i2c
-      tag: 0.1.4
     extraFiles:
       tree.html:
         mountPath: /usr/local/share/jupyter/custom_template/tree.html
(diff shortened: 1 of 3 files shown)


config/clusters/qcl/common.values.yaml
```diff
@@ -48,13 +43,6 @@ jupyterhub:
           - gizmo404
           - jtkmckenna
   singleuser:
-    image:
-      # pangeo/pangeo-notebook is maintained at: https://github.com/pangeo-data/pangeo-docker-images
-      name: pangeo/pangeo-notebook
-      # pullPolicy set to "Always" because we use the changing over time tag
-      # "latest".
-      pullPolicy: Always
-      tag: "latest"
```
consideRatio, line 57: QCL are using `latest` tags for their configured images, but pullPolicy defaults to the `Always` behavior when unspecified and the tag is `latest` anyhow.

config/clusters/callysto/common.values.yaml
```diff
@@ -37,9 +32,6 @@ jupyterhub:
           name: Callysto
           url: https://www.callysto.ca
   singleuser:
-    image:
```
yuvipanda, line 40: I consistently see PRs bumping the image tag from @ianabc (https://github.com/2i2c-org/infrastructure/pull/2823). Did you switch to using the configurator, @ianabc?

ianabc: I used the configurator for the last two of image updates we did. I remember where we discussed this, but I thought this was the intention, but I haven't been following the prePuller discussions so if I'm messing something up, please let me know and I'll change it.

yuvipanda (approved):
This looks good to me :)

consideRatio:
Thank you for reviewing @yuvipanda !!

config/clusters/qcl/common.values.yaml
yuvipanda: We just discovered that this is actually not true, due to a kubespawner bug. https://github.com/jupyterhub/kubespawner/pull/807 has details.

yuvipanda: https://github.com/2i2c-org/infrastructure/pull/3396 temporarily sets imagePullPolicy again

Issue #3250, opened 2023-10-10
Document a k8s upgrade procedure for GKE


consideRatio:
## Action points

- Document an upgrade procedure for GKE clusters (Equivalent of #2142)
  - Preliminary notes in https://github.com/2i2c-org/infrastructure/issues/2388#issuecomment-1479153571.
  - Its likeley relevant to consider https://github.com/2i2c-org/infrastructure/issues/2490, and ensure we don't get two core replicas

consideRatio:
```bash
# useful in order to work with a different workspace in another terminal
# by default, the workspace is otherwise found in `./terraform/environment
export TF_WORKSPACE=...
```

consideRatio:
It seems that terraform can perform a node surge upgrade by updating node pools version, this makes us more easily able to upgrade core nodes but doesn't influence the user nodes very much as we can't evict users etc.

PR #3375, opened 2023-11-04
qcl, awi-ciroh, leap, 2i2c: gke k8s clusters upgraded to 1.27

consideRatio:
These clusters were opportunistically upgraded as part of me trying to gain experience to document an upgrade policy for GKE clusters (#3250).

### Semi-related extras

388eae8 - I used `deployer generate resource-allocation daemonset-requests <cluster name>` to update information about daemonsets' resource requests in clusters as well

d1e193d - I made the deploy hubs workflow not redeploy all hubs because of changes to unrelated `generate` commands not used in the workflow

### DIFF
--- a/terraform/gcp/projects/leap.tfvars
+++ b/terraform/gcp/projects/leap.tfvars
@@ -5,10 +5,10 @@ project_id = "leap-pangeo"
 core_node_machine_type = "n2-highmem-4"
 
 k8s_versions = {
-  min_master_version : "1.25.6-gke.1000",
-  core_nodes_version : "1.25.6-gke.1000",
-  notebook_nodes_version : "1.25.6-gke.1000",
-  dask_nodes_version : "1.25.6-gke.1000",
+  min_master_version : "1.27.4-gke.900",
+  core_nodes_version : "1.27.4-gke.900",
+  notebook_nodes_version : "1.27.4-gke.900",
+  dask_nodes_version : "1.27.4-gke.900",
 }
 
 # GPUs not available in us-central1-b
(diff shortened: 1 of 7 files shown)


consideRatio:
NOTE TO SELF: when merging, cancel the workflow as the full redeploy isn't relevant when updating the daemonset-requests.yaml file

terraform/gcp/projects/leap.tfvars
```diff
@@ -76,22 +76,26 @@ notebook_nodes = {
...
     max : 100,
     machine_type : "n2-highmem-16",
+    node_version : "1.25.6-gke.1000",
   },
   "n2-highmem-64" : {
     min : 0,
     max : 100,
     machine_type : "n2-highmem-64"
   }
+  # FIXME: Remove node pool specific node_version pin when given the chance and no such nodes are running
```
GeorgianaElena, line 93: Removing the pin + terraform apply will delete the current node pool and recreate one with the newest k8s version? Or are extra steps needed to upgrade these machine types' k8s version?

terraform/gcp/projects/pilot-hubs.tfvars
```diff
@@ -19,11 +19,11 @@ enable_filestore      = true
 filestore_capacity_gb = 5120
 
 notebook_nodes = {
-  # FIXME: Rename this to "n2-highmem-4" when given the chance and no such nodes are running
-  "user" : {
+  "n2-highmem-4" : {
     min : 0,
     max : 100,
     machine_type : "n2-highmem-4",
+    node_version : "1.27.4-gke.900",
```
GeorgianaElena, line 26: Is this pin needed if it's the same with the  `notebook_nodes_version : "1.27.4-gke.900", "1.27.4-gke.900"`?

GeorgianaElena (approved):
Thanks for doing these upgrades @consideRatio

terraform/gcp/projects/leap.tfvars
consideRatio: No removing it will actually not need to recreate the node pool, a rolling upgrade will be done to use the non-node-pool-specific version specified in terraform config above, with for 0 active nodes means its quick and a simple "update". But, only if no active users are on it.

terraform/gcp/projects/pilot-hubs.tfvars
consideRatio: Nice catch, nope - removed it in a dedicated commit

consideRatio:
Thank you @GeorgianaElena!!

I rebased this on two other PRs merged, but otherwise since your review I only removed a fixme note that you caught to be resolved (b8090f0).

PR #3377, opened 2023-11-05
leap and carbonplan: tune hub/proxy pods' requests/limits

### LINKED ISSUE #3376: Tune hub / proxy pod resource requests/limits for LEAP and Carbonplan
Among all hubs, only LEAP and Carbonplan seems to adjusts their `hub` and `proxy` pods' requested CPU/memory allocation. I think this is a bit too conservative and should be tweaked to be not overly conservative, but still conservative. By doing that, we ensure we don't propegate a too conservative practice that could lead to more core nodes than needed in some cases, which in turn will increase cloud costs, make k8s upgrades a bit slower if "node surge" is 1 as it is by default, and increase complexity without notable benefit.

### Leap

The leap hub's pods CPU/memory requests have been very conservatively configured, requesting 1 CPU and 2 GB of memory. Looking at the last 90 days of operations I conclude we can probably be less conservative.

image

image

### Carbonplan

Last 180 days, hub pod:

image

Last 180 days, proxy pod:

image

## Suggested new conservative limits

I suggest the new requests and limits for both Leap and Carbonplan
- CPU hub: 100m to 1000m
- CPU proxy: 100m to 1000m
- Memory hub: 512Mi to 1Gi
- Memory proxy: 256Mi to 512Mi

These are very conservative still, but less conservative.

consideRatio:
Fixes #3376 by tuning proxy/hub pods' resources requests/limits for cpu/memory that has been increased historically to ensure it won't be a problem. This tunes those conservative requests/limits based on data to help us avoid setting a too conservative precedent for other hubs and potentially reducing cloud costs.
- still requesting more memory than has ever been seen used, and allowing the limit of memory to be 2x of that request
- still requesting more CPU than has been seen the last 90 days, and allowing the limit of CPU to 10x of that request

### DIFF
--- a/config/clusters/carbonplan/common.values.yaml
+++ b/config/clusters/carbonplan/common.values.yaml
@@ -170,19 +170,19 @@ basehub:
       chp:
         resources:
           requests:
-            cpu: 0.5
+            cpu: 100m
             memory: 256Mi
           limits:
             cpu: 1
-            memory: 4Gi
+            memory: 512Mi
     hub:
       resources:
         requests:
-          cpu: 0.5
-          memory: 256Mi
+          cpu: 100m
+          memory: 512Mi
         limits:
           cpu: 1
-          memory: 4Gi
+          memory: 1Gi
       allowNamedServers: true
       config:
         JupyterHub:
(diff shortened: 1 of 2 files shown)


yuvipanda (approved):

consideRatio:
Thank you @yuvipanda !!

Issue #3334, opened 2023-10-27
Mitigating AWS running out on GPU nodes in an individual zone


consideRatio:
This issue is a followup to https://github.com/2i2c-org/infrastructure/issues/3334 where CryoCloud wanted GPU but AWS couldn't deliver, and how to mitigate it in the future.

---

The k8s based cluster-autoscaler controls the node pools that in AWS are called Auto Scaling Groups (ASG). The cluster-autoscaler's requests to scale up/down can sometimes fail because of AWS didn't have the capacity to meet the demand.

When this happens (as in https://2i2c.freshdesk.com/a/tickets/1076 for CryoCloud recently), the spawn failure for a user will be that the last thing logged is "triggered scale up" but then the start attempts times out.

In the AWS web console, we can identify the ASG/node pool and look under the "Activity" tab. There we could find this information.

> Launching a new EC2 instance. Status Reason: We currently do not have sufficient g4dn.xlarge capacity in the Availability Zone you requested (us-west-2a). Our system will be working on provisioning additional capacity. You can currently get g4dn.xlarge capacity by not specifying an Availability Zone in your request or choosing us-west-2b, us-west-2c. Launching EC2 instance failed.

![image](https://github.com/2i2c-org/infrastructure/assets/3837114/c49350a5-d964-4f42-9f67-13040ab8e749)

As seen in the error, they suggest that we could try request the same instance from another zone. This leads to the action point of this issue.

## Action points

- Investigate if we can change/re-create cryocloud's existing AWS GPU node pool to run in more than one zone via the eksctl config
- Figure out what additional zone(s) to run in that has the instance, and ideally capture notes on how

PR #3381, opened 2023-11-06
cryo: Allow GPU nodes to spawn across AZs

yuvipanda:
We already do this for GCP, now we do it on AWS too.

Fixes https://github.com/2i2c-org/infrastructure/issues/3334

### DIFF
--- a/docs/howto/features/gpu.md
+++ b/docs/howto/features/gpu.md
@@ -114,6 +114,9 @@ AWS, and we can configure a node group there to provide us GPUs.
         tags+: {
             "k8s.io/cluster-autoscaler/node-template/resources/nvidia.com/gpu": "1"
         },
+        // Allow provisioning GPUs across all AZs, to prevent situation where all
+        // GPUs in a single AZ are in use and no new nodes can be spawned
+        availabilityZones: masterAzs,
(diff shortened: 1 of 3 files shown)


yuvipanda:
Deployed and tested on cryo

GeorgianaElena (approved):

yuvipanda:
I'm going to terraform apply this across our AWS clusters

consideRatio:
I think there may be an issue if we specify a region where the GPU isn't available, but I'm not sure - maybe not. Could also have been resolved with updates to cluster-autoscaler understanding it shouldn't attempt to scale up in a zone where the relevant GPU to attach isn't available.

yuvipanda:
The `terraform apply` only creates EFS Mount Target in all zones, so nodes *can* spawn there if needed. Only cryocloud is updated now.

Was there a previous issue with cluster-autoscaler around this?

damianavila:
> I'm going to terraform apply this across our AWS clusters

@yuvipanda, can you confirm if this ⬆️ happened?

consideRatio:
I did a sweep, this is now consistently applied across all AWS projects!

PR #3383, opened 2023-11-06
GCP: Allow public access to buckets via IAM, not ACL

yuvipanda:
Previously, we were mixing ACL and IAM, which led to basically the bucket *not* being accessible publicly - only to authenticated users.

This switches everything to using IAM, which *does* make the bucket properly publicly accessible.

In addition, there's now a policy that we only enable this when 2i2c is *not* handling billing, as there can be disastrous cost consequences.

We fix this for the LEAP bucket.

Config is moved to `user_buckets` rather than `hub_permissions`, as the config is purely set on the bucket and not related to which hub we are configuring.

Ref https://2i2c.freshdesk.com/a/tickets/954

### DIFF
--- a/docs/howto/features/buckets.md
+++ b/docs/howto/features/buckets.md
@@ -73,9 +73,31 @@ on why users want this!
 
 4. Get this change deployed, and users should now be able to use the buckets!
    Currently running users might have to restart their pods for the change to take effect.
-   
-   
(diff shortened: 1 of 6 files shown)


consideRatio (approved):
It would be good if the title clarifies this is GCP specific, otherwise this looks great to me!

Issue #3267, opened 2023-10-13
Fix for CILogon to provide a default idp to select


consideRatio:
https://2i2c.freshdesk.com/a/tickets/1038 reported what is tracked in https://github.com/jupyterhub/oauthenticator/issues/690 which should be easy to fix with the planning done already.

I think we should try to resolve this quickly without waiting for / rushing an oauthenticator release by:

1. develop a fix, trial the fix, open a PR to oauthenticator
2. update our default hub image to reference PR branch
   - I think we shouldn't use an experiment because its a regression impacting many hubs rather than a feature we need feedback and look to iterate on
3. report to Robert in https://2i2c.freshdesk.com/a/tickets/1038 that this is now resolved

yuvipanda:
Looking at the generated `values.yaml` used with `k -n ucmerced get secret hub -o json | jq -r '.data."values.yaml" ' | base64 -D`

```
    CILogonOAuthenticator:
      allowed_idps:
        http://google.com/accounts/o8/id:
          username_derivation:
            username_claim: email
        urn:mace:incommon:ucmerced.edu:
          allow_all: true
          username_derivation:
            username_claim: eppn
```

So helm is doing something with the values.yaml that's causing this.

yuvipanda:
And we use the helm `toYaml` function to write this out, and it sorts keys lexicographically. And since 'u' comes after 'h' we end up in this situation.

damianavila:
Given that 3385 (workaround) was merged (and communicated via FD), I am gonna to close this one for now.
Feel free to re-open if you disagree.

yuvipanda:
I think we should keep this open until the upstream issue is sorted

PR #3385, opened 2023-11-07
ucmerced: Show ucmerced login option before Google

yuvipanda:
Temporary but not particularly hacky fix, until
https://github.com/jupyterhub/oauthenticator/issues/690

Ref https://github.com/2i2c-org/infrastructure/issues/3267


yuvipanda:
Tested on UC Merced staging.

GeorgianaElena (approved):
Makes sense! Thanks @yuvipanda

PR #3389, opened 2023-11-08
openscapes: Fix extraVolumeMounts being overriden

yuvipanda:
Follow-up to https://github.com/2i2c-org/infrastructure/pull/3388, which otherwise basically just didn't mount home directories at all!

Fixes https://github.com/2i2c-org/infrastructure/pull/3388


yuvipanda:
Merging this as otherwise openscapes prod folks could lose data. Currently I manually deployed from my laptop to prevent that.

PR #3391, opened 2023-11-08
hhmi: add custom_image option

jmunroe:


jmunroe:
Self-merging since only changing the profile options.

PR #3399, opened 2023-11-10
[New Hub] MUST on Catalyst Project Africa cluster

AIDEA775:
- ref: https://github.com/2i2c-org/infrastructure/issues/3305

This will be deployed to must.af.catalystproject.2i2c.cloud


config/clusters/catalystproject-africa/must.values.yaml
```diff
@@ -0,0 +1,40 @@
...
+      JupyterHub:
+        authenticator_class: cilogon
+      CILogonOAuthenticator:
+        oauth_callback_url: https://must.af.catalystproject.2i2c.cloud/hub/oauth_callback
+        allowed_idps:
+          http://google.com/accounts/o8/id:
+            username_derivation:
+              username_claim: email
+            allowed_domains:
+              - nm-aist.ac.tz
```
sgibson91, line 37: Should this be `must.ac.mw` instead?

AIDEA775: Oh, yes! It slipped my mind.

sgibson91:
So visiting https://must.af.catalystproject.2i2c.cloud is returning a 404 for me 😕

AIDEA775:
> So visiting https://must.af.catalystproject.2i2c.cloud is returning a 404 for me 😕

Okay, so it will be available when it's merged? I edited the PR comment.

Also, do I need to change any other configurations in the load balancer? :thinking:

sgibson91:
> Okay, so it will be available when it's merged?

No, if you've run `deployer deploy $CLUSTER_NAME $HUB_NAME` locally it should be available immediately. We tend to deploy new hubs manually first and check they work, ask for feedback from the hub champion, before merging, as we can sometimes make tweaks as part of the setup.

If you have deployed locally, then we need to do more investigation as to why this is not working, but I cannot see any obvious issues from the files changed (though I am on my phone).

AIDEA775:
Fantastic! No, I didn't deploy it locally. Now it's running!

AIDEA775:
Aah! I didn't see that. Done!

jmunroe:
After logging in, I get bounced to https://must.af.catalystproject.2i2c.cloud/user/jmunroe@2i2c.org/rstudio? (which gives a 404)

I think I should be redirect to https://must.af.catalystproject.2i2c.cloud/user/jmunroe@2i2c.org/lab (which does work).

I haven't yet looked at why this occurring.

sgibson91:
I think now we have a hub on this cluster using a different image, it makes sense to move the RStudio image definition from `common.values.yaml` and into the relevant hub-specific values files. I was going to do this after this PR was merged, but we may have to do it beforehand if there's a problem with this hub.

sgibson91:
@jmunroe We are picking up rstudio as the default URL from these lines in common.values.yaml

https://github.com/2i2c-org/infrastructure/blob/31d7b50302aea67a83642eebcf88202c2bc2a30a/config/clusters/catalystproject-africa/common.values.yaml#L81-L82

@AIDEA775 I opened https://github.com/AIDEA775/2i2c-infrastructure/pull/1 which you should be able to merge into this PR. Then if you redeploy, I think the James reported will be fixed.

AIDEA775:
Okay, redeployed, and now it's working properly.

Note: The default url is https://must.af.catalystproject.2i2c.cloud/user/asilva@2i2c.org/tree no `/lab`.

config/clusters/catalystproject-africa/must.values.yaml
```diff
@@ -0,0 +1,220 @@
...
+        allowed_idps:
+          http://google.com/accounts/o8/id:
+            username_derivation:
+              username_claim: email
+            allowed_domains:
+              - must.ac.mw
+      Authenticator:
+        admin_users:
+          - "bkankuzi@must.ac.mw"
+  singleuser:
```
sgibson91, line 41:
```suggestion
  singleuser:
    defaultUrl: "/lab"
```

This might work for setting the default interface to JupyterLab

AIDEA775: Hmm, okay. Do all JupyterHubs need to have that default URL? I tested some hubs, for example, https://2i2c.callysto.ca/, and the default URL is "/tree".

sgibson91: I don't know. I guess it depends on if the community expect /lab or /tree, I also think it depends on what environment image the communities use and what entrypoint they define. This may be a preference we need to establish when partnerships are gathering the new hub info. @jmunroe are this community expecting to work in a /lab interface?

jmunroe: Let's go with the `/lab` interface.

AIDEA775: I deployed the change, but it doesn't work until I changed it in the configurator UI. I guess it will apply only on the first deploy?

AIDEA775:
@jmunroe @sgibson91 Any update on this? Can we proceed?

sgibson91:
@AIDEA775 Have you commented on the related issue and invited the community champion to try it and provide feedback?

AIDEA775:
The community champion gave the okay. I rebased the branch to maintain a clean history :D

sgibson91 (approved):
LGTM, merge when you like :)

PR #3402, opened 2023-11-11
hhmi: add ephemeral mysql db sidecar container

jmunroe:


jmunroe:
Already deployed and tested.  Adding an ephemeral MySQL database as a sidecar container (no persistent storage).

jmunroe:
Self-merging so it doesn't get overwritten before the community sees it and can try it themselves (see Slack #hhmi-pilot for context)

damianavila:
I presume this could be added as a tiny section in https://infrastructure.2i2c.org/howto/features/per-user-db/#setup-the-database-server, right?

Issue #3405, opened 2023-11-13
Collaboratively and incrementally introduce `node-purpose` resource label


consideRatio:
This is a followup to #3397 where our node pools got a `node-purpose = core|notebook|dask-worker` label added to them unconditionally. The unconditional change will force our node pools to upgrade, which often leads to disruption. Due to this, the change hasn't yet been applied to all our clusters node pools.

Here is how I think we should make a hard-to-do-all-at-once infra change collaboratively over time:
- at one point in time: introduce a temporarily opt-out change and opting out of the change systematically
- incrementally over time and by all engineers: incrementally stop opting out of the change across our clusters
- cleanup the temp opt-out config when its not longer referenced

### Action points

- [x] Introduce temp opt-out config, apply where needed along with a fixme note
  #3406
- [ ] Collaboratively and opportunistically stop opting out
- [ ] Cleanup temp opt-out config

### Clusters with partial upgrades remaining currently

- [ ] cloudbank
- [x] leap
- [x] pangeo-hubs
- [ ] pilot-hubs
- [x] qcl

PR #3397, opened 2023-11-10
gcp: Tag nodes with their purpose for costing purposes

yuvipanda:
Helps more easily separate the cost of various parts of the infrastructure.
    
Unfortunately this *does* recreate the nodes in the node-pool,
so will have to be rolled out slowly with appropriate timing.
    
Projects to apply this on:
- [x] 2i2c-uk
- [x] calysto
- [x] awi-ciroh
- [ ] catalystproject-latam
- [ ] cloudbank
- [x] hhmi
- [ ] leap
- [ ] linked-earth
- [ ] m2lines
- [ ] meom-ige
- [ ] pangeo-hubs
- [ ] pilot-hubs
- [ ] qcl

 

Ref https://github.com/2i2c-org/infrastructure/issues/3350

PR #3406, opened 2023-11-13
terraform, gcp: add opt-out var to collaboratively and opportunistically apply node-purpose infra labels

consideRatio:
This is for #3405, a followup to #3397 that is work I think all @2i2c-org/engineering needs to opportunistically apply over time when given the chance. @2i2c-org/engineering the ask is that when you apply terraform config, consider if there are a fixme that you'd like to opportunistically resolve while you are at it.

This PR adds variables and fixme notes in individual GCP clusters' .tfvar files, allowing us to see where we haven't yet managed to apply what #3397 introduced.

The purpose of this is to:
- ensure our terraform state is reflect in our config
- set us up to collectively and opportunistically work towards the desired state over time
- unblock other changes that may not be disruptive in the terraform config

### DIFF
--- a/terraform/gcp/cluster.tf
+++ b/terraform/gcp/cluster.tf
@@ -218,7 +218,7 @@ resource "google_container_node_pool" "core" {
     # Faster disks provide faster image pulls!
     disk_type = "pd-balanced"
 
-    resource_labels = {
+    resource_labels = var.temp_opt_out_node_purpose_label_core_nodes ? {} : {
       "node-purpose" : "core"
     }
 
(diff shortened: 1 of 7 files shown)


yuvipanda (approved):
This is pretty clever, thanks for doing this @consideRatio.

I've just applied tf to pangeo-hubs (and ticked it off in the PR I was using to track) so can you remove that from here before merging?

consideRatio:
> I've just applied tf to pangeo-hubs (and ticked it off in the PR I was using to track) so can you remove that from here before merging?

Ah nice @yuvipanda, I rebased and removed that commit!

consideRatio:
Thank you for reviewing and working on this @yuvipanda!!

consideRatio:
Argh, removed `pilot-hubs` instead of `pangeo-hubs`... =/ I'll open a new PR to fix it.

PR #3416, opened 2023-11-14
Add GitHub Actions workflow to automatically provide readthedocs preview links in PRs that touch docs

sgibson91:
This PR adds a GitHub Action workflow that will automatically add a link to the readthedocs preview to PR descriptions when we make changes within the `docs/` folder. An example PR can be seen here: https://github.com/sgibson91/test-multilingual-sphinx/pull/4

### DIFF
--- a/.github/workflows/autolink-readthedocs-previews-to-prs.yaml
+++ b/.github/workflows/autolink-readthedocs-previews-to-prs.yaml
@@ -1,0 +1,21 @@
+name: Add readthedocs preview link to pull requests
+
+on:
+  pull_request:
+    types:
+      - opened
+    branches:
+      - main
+    paths:
+      - "docs/**"
+
+jobs:
+  autolink-rtd-previews:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/autolink-readthedocs-previews-to-prs.yaml
```diff
@@ -0,0 +1,21 @@
+name: Add readthedocs preview link to pull requests
+
+on:
+  pull_request:
+    types:
+      - opened
+    branches:
+      - main
```
sgibson91, line 8:
```suggestion
      - master
```

consideRatio (approved):
Nice, looks great! :)

sgibson91:
FYI, I suspect we have the same permission issue here as the other 2 PRs but we don't see it because we have the additional trigger modifier of changes to the docs path

sgibson91:
Solution found in https://github.com/2i2c-org/docs/pull/194#discussion_r1404439410

Issue #3224, opened 2023-10-04
Upgrade our k8s clusters to k8s 1.25+


consideRatio:
## Current status

```shell
# collect clusters current versions by running a `kubectl version` command
ls config/clusters | xargs -I {} deployer use-cluster-credentials {} 'echo {} $(kubectl version -o json | jq -r .serverVersion.gitVersion) >> k8s-versions.txt'
```

Filtered to show only <=1.24 clusters, these are:

```
# 1.24 - EKS
carbonplan v1.24.16-eks-2d98532
openscapes v1.24.16-eks-2d98532
ubc-eoas v1.24.16-eks-2d98532
```

## Action points

- Upgrade the three clusters currently at 1.24 to the latest available minor versions.
  These clusters are all EKS clusters, and there is established documentation on how to upgrade these [here](https://infrastructure.2i2c.org/howto/upgrade-cluster/aws/).
  - [x] carbonplan v1.24.16-eks-2d98532
  - [x] openscapes v1.24.16-eks-2d98532
  - [x] ubc-eoas v1.24.16-eks-2d98532
- [x] Improve the AWS EKS cluster upgrade docs if needed

## Related
- #3249
- #3412

GeorgianaElena:
@consideRatio, I went through the upgrade docs (yay for having them, very useful) and have some questions:

1.  This step is not in the docs, but I've noticed, that for leap for example, you coordinated with the community before going for an upgrade https://github.com/2i2c-org/infrastructure/pull/2318
    - should we update the docs to add such a step?
    - for these particular 3 clusters, do you suggest getting in touch with the communities before? If so,  I remember a discussion about creating a list of community reps, but I don't know where we're at with that atm?
2. Should I update these clusters to the latest version `1.28`?
3. I feel like we should also add a step about checking the release notes, so that we don't accidentally miss a breaking change. But maybe that should instead go into a policy like https://github.com/2i2c-org/infrastructure/issues/3248. Wondering if we should add some docs about the policy before going on with this one. WDYT?
3. Also, "one minor version at a time?" 😱

consideRatio:
- 1a - "should we update the docs to add such a step?"
  There is this part currently:
  ![image](https://github.com/2i2c-org/infrastructure/assets/3837114/0c464512-3782-46fd-a5df-148bda38b560)
  For now, I don't want this work to include thinking about policy because its a big task by itself. With this work done, you would also have more practical experience which can help when thinking about policy.
[...]

PR #3420, opened 2023-11-14
Upgrade carbonplan version and update aws upgrade docs

GeorgianaElena:
For https://github.com/2i2c-org/infrastructure/issues/3224

### DIFF
--- a/docs/howto/upgrade-cluster/aws.md
+++ b/docs/howto/upgrade-cluster/aws.md
@@ -195,7 +195,7 @@ cluster is unused or that the maintenance is communicated ahead of time.
       current version of the control plane and this version is **no more
       than three minor versions** ahead of the node groups versions.
 
-   #### 5.2. Rename part 1: add a new core node group (like `core-b`)
+   #### 5.2. Renaming node groups part 1: add a new core node group (like `core-b`)
 
       Rename the `CLUSTER_NAME.jsonnet` config file's entry for the core node
       group temporarily when running this command, either from `core-a` to `core-b` or
@@ -212,7 +212,7 @@ cluster is unused or that the maintenance is communicated ahead of time.
       eksctl create nodegroup --config-file=$CLUSTER_NAME.eksctl.yaml --include="core-b"
       ```
 
-   #### 5.3. Rename part 2: delete all old node groups (like `core-a,nb-*,dask-*`)
+   #### 5.3. Renaming node groups part 2: delete all old node groups (like `core-a,nb-*,dask-*`)
 
       Rename the core node group again in the config to its previous name,
       so the old node group can be deleted with the following command,
@@ -232,7 +232,7 @@ cluster is unused or that the maintenance is communicated ahead of time.
       Rename (part 3/3) the core node group one final time in the config to its
       new name, as that represents the state of the EKS cluster.
 
-   #### 5.4. Rename part 3: re-create all non-core node groups (like `nb-*,dask-*`)
+   #### 5.4. Renaming node groups part 3: re-create all non-core node groups (like `nb-*,dask-*`)
 
       ```bash
       eksctl create nodegroup --config-file=$CLUSTER_NAME.eksctl.yaml --include="nb-*,dask-*"
--- a/docs/howto/upgrade-cluster/aws.md
+++ b/docs/howto/upgrade-cluster/aws.md
@@ -194,10 +194,10 @@ If you upgrade k8s multiple minor versions, consider repeating step 3,
 up to maximum three times, incrementing the control plane one minor version
 at the time.
 
-This is because a node's k8s software (`kubelet`) can be up to three minor
-versions **behind** the control plane version[^2] if kublet is at least
-at version 1.25. Due to this, you can plan your cluster upgrade to only
-involve the minimum number of node group upgrades.
+This is because the control plane version can be **ahead** of
+the node's k8s software (`kubelet`) by up to three minor versions [^2]
+if kublet is at least at version 1.25. Due to this, you can plan your
+cluster upgrade to only involve the minimum number of node group upgrades.
 
 So if you upgrade from k8s 1.25 to 1.28, you can for example upgrade the k8s
 control plane three steps in a row, from 1.25 to 1.26, then from 1.26
--- a/docs/howto/upgrade-cluster/aws.md
+++ b/docs/howto/upgrade-cluster/aws.md
@@ -86,200 +86,191 @@ cluster is unused or that the maintenance is communicated ahead of time.
 
 ## Cluster upgrade
 
-### 1. Ensure you are using the correct cluster's credentials
+### 1. Ensure in-cluster permissions
 
-   Future `eksctl` commands will need info from the kubeconfig file
-   to be able to run successfully.
+The k8s api-server won't accept commands from you unless you have configured
+a mapping between the AWS user to a k8s user, and `eksctl` needs to make some
+commands behind the scenes.
 
-   ```bash
-   deployer use-cluster-credentials $CLUSTER_NAME
-   ```
+This mapping is done from a ConfigMap in kube-system called `aws-auth`, and
+we can use an `eksctl` command to influence it.
 
-### 2. Ensure in-cluster permissions
+```bash
+eksctl create iamidentitymapping \
+   --cluster=$CLUSTER_NAME \
+   --region=$CLUSTER_REGION \
+   --arn=arn:aws:iam::<aws-account-id>:user/<iam-user-name> \
+   --username=<iam-user-name> \
+   --group=system:masters
+```
 
-   The k8s api-server won't accept commands from you unless you have configured
-   a mapping between the AWS user to a k8s user, and `eksctl` needs to make some
-   commands behind the scenes.
+### 2. Acquire and configure AWS credentials
 
-   This mapping is done from a ConfigMap in kube-system called `aws-auth`, and
-   we can use an `eksctl` command to influence it.
+Visit https://2i2c.awsapps.com/start#/ and acquire CLI credentials.
 
-   ```bash
-   eksctl create iamidentitymapping \
-      --cluster=$CLUSTER_NAME \
-      --region=$CLUSTER_REGION \
-      --arn=arn:aws:iam::<aws-account-id>:user/<iam-user-name> \
-      --username=<iam-user-name> \
-      --group=system:masters
-   ```
+In case the AWS account isn't managed there, inspect
+`config/$CLUSTER_NAME/cluster.yaml` to understand what AWS account number to
+login to at https://console.aws.amazon.com/.
 
-### 3. Acquire and configure AWS credentials
+Configure credentials like:
 
-   Visit https://2i2c.awsapps.com/start#/ and acquire CLI credentials.
+```bash
+export AWS_ACCESS_KEY_ID="..."
+export AWS_SECRET_ACCESS_KEY="..."
+```
 
-   In case the AWS account isn't managed there, inspect
-   `config/$CLUSTER_NAME/cluster.yaml` to understand what AWS account number to
-   login to at https://console.aws.amazon.com/.
+### 3. Upgrade the k8s control plane's one minor version for three times in a row
 
-   Configure credentials like:
+```{important}
+The k8s control plane can only be upgraded one minor version at the time.[^1]
+```
 
-   ```bash
-   export AWS_ACCESS_KEY_ID="..."
-   export AWS_SECRET_ACCESS_KEY="..."
-   ```
+Update the cluster's jsonnet config's version field **one minor version.**
+In `$CLUSTER_NAME.jsonnet` there should be an entry like the one below,
+where the version must be updated.
 
-### 4. Upgrade the k8s control plane's one minor version for three times in a row
+```yaml
+apiVersion: 'eksctl.io/v1alpha5',
+   kind: 'ClusterConfig',
+   metadata+: {
+      name: "carbonplanhub",
+      region: clusterRegion,
+      version: '1.27'
+   },
+```
 
-   ```{important}
-   The k8s control plane can only be upgraded one minor version at the time.[^1]
-   ```
+Then regenerate the `eksctl` config file:
 
-   Update the cluster's jsonnet config's version field **one minor version.**
-   In `$CLUSTER_NAME.jsonnet` there should be an entry like the one below,
-   where the version must be updated.
+```bash
+# re-generate an eksctl config file for use with eksctl
+jsonnet $CLUSTER_NAME.jsonnet > $CLUSTER_NAME.eksctl.yaml
+```
 
-   ```yaml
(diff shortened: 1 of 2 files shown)


docs/howto/upgrade-cluster/aws.md
```diff
@@ -86,7 +86,16 @@ cluster is unused or that the maintenance is communicated ahead of time.
...
 ## Cluster upgrade
 
-1. *Ensure in-cluster permissions*
+### 1. Ensure you are using the correct cluster's credentials
+
+   Future `eksctl` commands will need info from the kubeconfig file
+   to be able to run successfully.
+
+   ```bash
+   deployer use-cluster-credentials <hub>
```
sgibson91, line 95:
```suggestion
   deployer use-cluster-credentials <cluster>
```

docs/howto/upgrade-cluster/aws.md
```diff
@@ -119,10 +128,32 @@ cluster is unused or that the maintenance is communicated ahead of time.
    export AWS_SECRET_ACCESS_KEY="..."
    ```
 
-3. *Upgrade the k8s control plane's one minor version*
+### 4. Upgrade the k8s control plane's one minor version for three times in a row
```
sgibson91, line 131:
```suggestion
### 4. Upgrade the k8s control plane's minor version incrementally through consecutive versions
```

I'm trying to generalise this statement a bit, because it will depend on what the starting version is and what the desired version is as to whether it's two, three, or four times!

sgibson91: Reading further on, I see we cannot have node groups more than 3 minor versions behind, so maybe this should be:

```
Upgrade the k8s control plane incrementally through consecutive minor versions, to a maximum of three minor versions
```

(This is getting very verbose but it's a complex topic!)

docs/howto/upgrade-cluster/aws.md
```diff
@@ -134,62 +165,86 @@ cluster is unused or that the maintenance is communicated ahead of time.
...
 
-   2. Add a new core node group (like `core-b`)
+      Make sure the cluster's jsonnet config's version field matches the
+      current version of the control plane and this version is **no more
+      than three minor versions** ahead of the node groups versions.
 
-      Rename (part 1/3) the config file's entry for the core node group
-      temporarily when running this command, either from `core-a` to `core-b` or
-      the other way around.
+   #### 5.2. Rename part 1: add a new core node group (like `core-b`)
```
sgibson91, line 198:
```suggestion
   #### 5.2. Renaming node groups part 1: add a new core node group (like `core-b`)
```

docs/howto/upgrade-cluster/aws.md
```diff
@@ -134,62 +165,86 @@ cluster is unused or that the maintenance is communicated ahead of time.
...
+      jsonnet $CLUSTER_NAME.jsonnet > $CLUSTER_NAME.eksctl.yaml
+      ```
 
       ```bash
+      # create a copy of the current nodegroup
       eksctl create nodegroup --config-file=$CLUSTER_NAME.eksctl.yaml --include="core-b"
       ```
 
-   3. Delete all old node groups (like `core-a,nb-*,dask-*`)
+   #### 5.3. Rename part 2: delete all old node groups (like `core-a,nb-*,dask-*`)
```
sgibson91, line 215:
```suggestion
   #### 5.3. Renaming node groups part 2: delete all old node groups (like `core-a,nb-*,dask-*`)
```

docs/howto/upgrade-cluster/aws.md
```diff
@@ -134,62 +165,86 @@ cluster is unused or that the maintenance is communicated ahead of time.
...
       ```bash
+      # delete the original nodegroup
       eksctl delete nodegroup --config-file=$CLUSTER_NAME.eksctl.yaml --include="core-a,nb-*,dask-*" --approve --drain=true
       ```
 
       Rename (part 3/3) the core node group one final time in the config to its
       new name, as that represents the state of the EKS cluster.
 
-   4. Re-create all non-core node groups (like `nb-*,dask-*`)
+   #### 5.4. Rename part 3: re-create all non-core node groups (like `nb-*,dask-*`)
```
sgibson91, line 235:
```suggestion
   #### 5.4. Renaming node groups part 3: re-create all non-core node groups (like `nb-*,dask-*`)
```

sgibson91 (approved):
LGTM! I've left some comments trying to clean up the language a bit, but I think this is a complex topic so if I've misunderstood anything feel free to ignore it.

docs/howto/upgrade-cluster/aws.md
```diff
@@ -216,11 +271,17 @@ cluster is unused or that the maintenance is communicated ahead of time.
    ```
    ````
 
-6. *Repeat steps 3 and 5 if needed*
+### 7. Repeat steps 4 and 6 if needed
 
-   If you upgrade k8s multiple minor versions, repeat step 3 and 5, where you
+   If you upgrade k8s multiple minor versions, repeat step 4 and 6, where you
    increment it one minor version at the time.
```
consideRatio, line 278: I think this step stops making sense after the change to let a step upgrade the k8s control plane multiple times in a row, so I figure this should be removed!

docs/howto/upgrade-cluster/aws.md
```diff
@@ -134,62 +165,86 @@ cluster is unused or that the maintenance is communicated ahead of time.
...
+
+      Rename the `CLUSTER_NAME.jsonnet` config file's entry for the core node
+      group temporarily when running this command, either from `core-a` to `core-b` or
+      the other way around, then regenerate the `$CLUSTER_NAME.eksctl.yaml` file and
+      then create the new nodegroup.
+
+      ```bash
+      # re-generate an eksctl config file for use with eksctl
+      jsonnet $CLUSTER_NAME.jsonnet > $CLUSTER_NAME.eksctl.yaml
+      ```
```
consideRatio, line 208: I've appreciated doing temporary changes directly from the generated .yaml file instead of re-generating it repeatedly. Consider if you think that is an easier or more complicated strategy to use and document.

docs/howto/upgrade-cluster/aws.md
```diff
@@ -205,7 +260,7 @@ cluster is unused or that the maintenance is communicated ahead of time.
    eksctl utils update-coredns --config-file=$CLUSTER_NAME.eksctl.yaml --approve
 
    # upgrade the aws-ebs-csi-driver addon's deployment and daemonset
-   eksctl update addon --config-file=$CLUSTER_NAME.eksctl.yaml
+   eksctl update addon --name aws-ebs-csi-driver --version <version-to-upgrade-to> --cluster $CLUSTER_NAME
```
consideRatio, line 263:
```suggestion
   eksctl update addon --config-file=$CLUSTER_NAME.eksctl.yaml
```

- With #3421 we can use the previous approach!

docs/howto/upgrade-cluster/aws.md
```diff
@@ -134,62 +165,86 @@ cluster is unused or that the maintenance is communicated ahead of time.
...
-   5. Restore the k8s version in the config
+   #### 5.5. Restore the k8s version in the config
 
       We adjusted the k8s version in the config to influence the desired version
       of our created nodegroups. Let's restore it to what the k8s control plane
-      currently have.
+      currently have if not already.
 
-5. *Upgrad EKS add-ons (takes ~3*5s)*
+### 6. Upgrade EKS add-ons (takes ~3*5s)
```
consideRatio, line 247: I'm not sure if the addons needs to be updated directly once per k8s minor version upgraded, I'm not sure. The safer call is to bundle this with "upgrade k8s control plane one minor version at the time (up to three times)" step.

I see downsides in doing it once per minor version as well. For example, one can get stuck debugging something that would have gone away if upgrading all the way first.

docs/howto/upgrade-cluster/aws.md
```diff
@@ -86,7 +86,16 @@ cluster is unused or that the maintenance is communicated ahead of time.
...
-1. *Ensure in-cluster permissions*
+### 1. Ensure you are using the correct cluster's credentials
+
+   Future `eksctl` commands will need info from the kubeconfig file
+   to be able to run successfully.
+
+   ```bash
+   deployer use-cluster-credentials <hub>
+   ```
+
```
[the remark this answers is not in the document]
GeorgianaElena: Yes, before using the deployer, I was getting
```Error: getting list of API resources for raw REST client: Unauthorized```
 when running the `create iamidentitymapping` command.

Actually, checking again now, it works outside of deployer as well even though `kubectl config view` doesn't show the carbonplan cluster.

So it was probably some other config that I was missing outside of deployer 🤔  I will remove it from the steps as it was probably something going wrong on my end.

Thanks for pointing it out @consideRatio. I thought the kubeconfig needs to point to the cluster for eksctl to work.

File: docs/howto/upgrade-cluster/aws.md [hunk 11]
```diff
@@ -104,7 +113,7 @@ cluster is unused or that the maintenance is communicated ahead of time.
       --group=system:masters
    ```
 
-2. *Acquire and configure AWS credentials*
+### 3. Acquire and configure AWS credentials
 
    Visit https://2i2c.awsapps.com/start#/ and acquire CLI credentials.
```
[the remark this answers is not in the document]
GeorgianaElena: @consideRatio, you're right about the steps being more readable without the headers. The motivation was the need to link the step that I was at when getting errors when we tried to debug it toghether.

With headers saying something like: "running the command at https://2i2c-pilot-hubs--3420.org.readthedocs.build/howto/upgrade-cluster/aws/#rename-part-2-delete-all-old-node-groups-like-core-a-nb-dask failed with x error" without too many words.

docs/howto/upgrade-cluster/aws.md
GeorgianaElena: Yes, yes! Agree! I actually did the same, but felt the need to run a diff to make sure I got the node pool name right and I was not changing something else by mistake.

Hmm, instead, maybe temporarily git adding the eksctl.yaml file in the beginning might be more straightforward, as then you get to have the diff. Wdyt?

GeorgianaElena: @consideRatio, just realized you had this written all along: _If you update the .jsonnet file, make sure to re-generate the .yaml file
   before using `eksctl`. Respectively if you update the .yaml file directly,
   remember to update the .jsonnet file._

docs/howto/upgrade-cluster/aws.md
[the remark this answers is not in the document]
GeorgianaElena: Removed the step in https://github.com/2i2c-org/infrastructure/pull/3420/commits/997afa724980bbe533e8dccefdec7efaaa6e4406

docs/howto/upgrade-cluster/aws.md
GeorgianaElena: Addressed this in https://github.com/2i2c-org/infrastructure/pull/3420/commits/5684ef9a4c8f71fcd3a0b3913879d22329540b66

docs/howto/upgrade-cluster/aws.md
GeorgianaElena: 👍🏼 I moved it under the control plane update step in https://github.com/2i2c-org/infrastructure/pull/3420/commits/f5ca3276a01c2af20cc32ad242811b569e4812dc

docs/howto/upgrade-cluster/aws.md
GeorgianaElena: I've made the repeat part its own step in https://github.com/2i2c-org/infrastructure/pull/3420/commits/decdb37c343aa1de6c860668b20c18e6c7bf185a

docs/howto/upgrade-cluster/aws.md
```diff
@@ -134,62 +165,86 @@ cluster is unused or that the maintenance is communicated ahead of time.
    If you see the error `Error: the server has asked for the client to provide credentials` don't worry, if you try it again you will find that the cluster is now upgraded.
    ```
 
-4. *Upgrade node groups up to two minor versions above the k8s control plane*
+### 5. Upgrade node groups up to three minor versions until it matches the version of the k8s control plane
```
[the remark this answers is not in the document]
GeorgianaElena: Yes! Thank you! Updated it in https://github.com/2i2c-org/infrastructure/pull/3420/commits/decdb37c343aa1de6c860668b20c18e6c7bf185a

docs/howto/upgrade-cluster/aws.md
```diff
@@ -134,62 +165,86 @@ cluster is unused or that the maintenance is communicated ahead of time.
...
+   So if you upgrade from k8s 1.25 to 1.28, you can for example upgrade the k8s
+   control plane  three steps in a row, from 1.25 to 1.26, then from 1.26
+   to 1.27 and then from 1.27 to 1.28. This way, the node groups were left
+   behind the control plane by three minor versions, which is acceptable.
 
-   So if you upgrade from k8s 1.21 to 1.24, you can for example upgrade the k8s
-   control plane from 1.21 to 1.22, then upgrade the node groups from 1.21 to
-   1.24, followed by upgrading the control plane two steps in a row.
+   Then, you can upgrade the node groups from 1.25 to 1.28.
+   ```
```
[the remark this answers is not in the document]
GeorgianaElena: I've now transformed this into its own step in https://github.com/2i2c-org/infrastructure/pull/3420/commits/decdb37c343aa1de6c860668b20c18e6c7bf185a

GeorgianaElena:
@sgibson91 and @consideRatio, many thanks for your feedback. I have added a bunch of commits this morning that I hope made the info a bit more clear. 

Check it out at https://2i2c-pilot-hubs--3420.org.readthedocs.build/howto/upgrade-cluster/aws/ and let me know what you think

consideRatio:
image

I think the commit part is complicated because we don't version control it, and I don't think we should either. I think it simply doesn't get easier than the somewhat tricky procedure that requires attention to make sure to keep track of what is being applied - it will always be the yaml file as written to disk. If you introduce work on committing, it adds additional steps and complexity, especially since its not a version controlled file.

Due to this, I'd favor the removal of this optional step and nesting the `important` admonition under the third step.

---

Here "the config file" is mentioned to be the .jsonnet file specifically. I think this should be left not explicit, letting the reader either update the .yaml file, or the jsonnet file + generate like done in the step below. Compare:

image

image

---

With three minor version upgradable at the time, I think we should lean on the reduced complexity we can arrive at and assume we never need to upgrade more then three versions at the time.

This doesn't belong any more i think, after we upgraded the control plan version 1-3 times, we should simply have stayed at there going onwards. This change was also the only change to commit btw.

image

consideRatio (approved):
Thanks for working on this @GeorgianaElena!! I figure we can merge, and if needed we can do another iteration at some later time. Its really tricky to document this procedure I think!

GeorgianaElena:
Thanks @consideRatio. I've added one more commit and going for a merge now.

PR #3421, opened 2023-11-14
eksctl: declare an addon version to latest, so it can get updated

consideRatio:
Specifying "latest" here only influences us when we do `eksctl update addon --config-file=...`, then it will update to the latest version instead of doing nothing.

Note that just because it was set to `latest` doesn't mean it gets updated at all times, only when `eksctl update addon` is called, not for example when doing a node pool upgrade or cluster upgrade. This aligns with how `eksctl` typically works with the cluster config file - only reading whats relevant to the specific command, where no command is applying the whole config.


GeorgianaElena (approved):
Thanks @consideRatio

PR #3422, opened 2023-11-15
[QCL] Increase kernel culling interval

GeorgianaElena:
For https://2i2c.freshdesk.com/a/tickets/1120

### DIFF
--- a/config/clusters/qcl/common.values.yaml
+++ b/config/clusters/qcl/common.values.yaml
@@ -43,6 +43,13 @@ jupyterhub:
           - gizmo404
           - jtkmckenna
   singleuser:
+    extraFiles:
+      jupyter_notebook_config.json:
+        data:
+          MappingKernelManager:
+            # Cull idle kernels after 24h (24 * 60 * 60)
+            # Ref https://2i2c.freshdesk.com/a/tickets/1120
+            cull_idle_timeout: 86400
     image:
       # Required when using :latest, until https://github.com/jupyterhub/kubespawner/pull/807
       # is merged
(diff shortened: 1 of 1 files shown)


config/clusters/qcl/common.values.yaml
```diff
@@ -43,6 +43,13 @@ jupyterhub:
           - gizmo404
           - jtkmckenna
   singleuser:
+    extraFiles:
+      jupyter_notebook_config.json:
```
consideRatio, line 47: https://github.com/2i2c-org/infrastructure/blob/bd5f92a95dc7dff9757498be737c989f7bc7bc32/helm-charts/basehub/values.yaml#L255-L264

I think at this point, almost all cases involves configuring the jupyter_server_config.json rather than the jupyter_notebook_config.json file at this point. For QCL, I saw this when starting a user server and looking at its startup logs.

```
[I 2023-11-15 10:28:09.485 SingleUserLabApp mixins:547] Starting jupyterhub single-user server version 4.0.1
[I 2023-11-15 10:28:09.485 SingleUserLabApp mixins:561] Extending jupyterlab.labhubapp.SingleUserLabApp from jupyterlab 3.6.5
[I 2023-11-15 10:28:09.486 SingleUserLabApp mixins:561] Extending jupyter_server.serverapp.ServerApp from jupyter_server 1.23.6
```

With `Extending jupyter_server.serverapp.ServerApp from jupyter_server 1.23.6` I think its safe to say we should especially configure the `jupyter_server_config.json` file, and with very high likeleyhood don't need to also configure the `jupyter_notebook_config.json` file.

```suggestion
      jupyter_server_config.json:
```

GeorgianaElena: Ah, true. I copy-pasted it from leap config. Maybe it's worth updating that too. Thank you!

consideRatio: Ah, opened https://github.com/2i2c-org/infrastructure/pull/3423

PR #3423, opened 2023-11-15
Configure jupyter_server instead of notebook when configuring kernel culling

consideRatio:
This influences our docs and m2lines. m2lines is not being actively used and should to my knowledge be decommissioned, but it is updated anyhow to avoid retaining an outdated configuration that could be seen as an example to be copied to other hub configs.


GeorgianaElena (approved):
Thanks @consideRatio!

Also, you're right about m2lines https://github.com/2i2c-org/infrastructure/issues/3398

PR #3431, opened 2023-11-16
Clarify what the requestor_pays flag does

yuvipanda:
Reference: https://2i2c.slack.com/archives/C055A1J1DRP/p1700050896750479

### DIFF
--- a/docs/howto/features/buckets.md
+++ b/docs/howto/features/buckets.md
@@ -106,6 +106,35 @@ user_buckets = {
 }
 ```
 
+## Enabling Requester Pays flag on buckets
+
+Some hubs want to expose a particular bucket to the broad internet
+but not be billed for the charges associated with making and executing the
+requests on this bucket.
+
+By enabling the [Requester Pays flag](https://cloud.google.com/storage/docs/using-requester-pays#using),
+the requesters are required to include a billing project in their requests,
+which will mean that the billing will happen on the requester's project.
+
+Enabling Requester Pays is useful, for example, if the communities have a lot
+of data that they want to make available to others, but don't want to be
+charged for their access to that data.
+
+This can be enabled by setting the `requester_pays` parameter in `user_buckets`
+for the appropriate bucket, and running `terraform apply`.
+
+Example:
+
+```terraform
+user_buckets = {
+   "persistent": {
+      "delete_after": null,
+      "public_access": true,
+      "requester_pays": true
+   }
+}
+```
+
 ## Enable access logs for objects in a bucket
 
 ### GCP
--- a/docs/howto/features/buckets.md
+++ b/docs/howto/features/buckets.md
@@ -106,35 +106,6 @@ user_buckets = {
 }
 ```
 
-## Enabling Requester Pays flag on buckets
-
-Some hubs want to expose a particular bucket to the broad internet
-but not be billed for the charges associated with making and executing the
-requests on this bucket.
-
-By enabling the [Requester Pays flag](https://cloud.google.com/storage/docs/using-requester-pays#using),
-the requesters are required to include a billing project in their requests,
-which will mean that the billing will happen on the requester's project.
-
-Enabling Requester Pays is useful, for example, if the communities have a lot
-of data that they want to make available to others, but don't want to be
-charged for their access to that data.
-
-This can be enabled by setting the `requester_pays` parameter in `user_buckets`
-for the appropriate bucket, and running `terraform apply`.
-
-Example:
-
-```terraform
-user_buckets = {
-   "persistent": {
-      "delete_after": null,
-      "public_access": true,
-      "requester_pays": true
-   }
-}
-```
-
 ## Enable access logs for objects in a bucket
 
 ### GCP
--- a/docs/topic/features.md
+++ b/docs/topic/features.md
@@ -34,8 +34,22 @@ the data. This is very commonly used by organizations that provide big datasets
 on Google Cloud storage, to sustainably share costs of maintaining the data.
 
 When this feature is enabled, users on a hub accessing cloud buckets from
-other organizations marked as 'requestor pays' will increase our cloud bill.
+other organizations marked as 'Requester Pays' will increase our cloud bill.
 Hence, this is an opt-in feature.
+
+```{important}
+This feature enables the hub users to access `Requester Pays` buckets,
+**outside** of their project.
+
+However, note that this feature **does not** control which buckets **inside**
+the project will have `Requester Pays` enabled for themselves.
+
+This can be checked from the console following these steps in the
+[GCP docs](`Requester Pays`). Enabling/disabling the `Requester Pays`
+setting on a bucket can be achieved following
+[this other section](https://cloud.google.com/storage/docs/using-requester-pays#set)
+ of the GCP docs.
+```
 
 (topic:features:cloud:scratch-buckets)=
 ## 'Scratch' buckets on object storage
--- a/docs/topic/features.md
+++ b/docs/topic/features.md
@@ -54,7 +54,7 @@ Hence, this is an opt-in feature.
 The buckets that we set for communities, inside their projects can also have this flag enabled on them, which means that other people outside will be charged for their usage.
 
 ```{warning}
-This is not supported yet by our terraform. Follow (todo: insert issue link) for when support will be added.
+This is not supported yet by our terraform. Follow https://github.com/2i2c-org/infrastructure/issues/3746 to check when support will be added.
 ```
 
 (topic:features:cloud:scratch-buckets)=
--- a/docs/topic/features.md
+++ b/docs/topic/features.md
@@ -39,7 +39,7 @@ on Google Cloud storage, to sustainably share costs of maintaining the data.
 
 If buckets outside the project have the `Requester Payes` flag, then we need to:
 - set `hub_cloud_permissions.allow_access_to_external_requester_pays_buckets`
-  in the terraform config of the cluster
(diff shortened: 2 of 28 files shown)


GeorgianaElena:
I just pushed a small commit updating the docs as well and believe it might be ready for review. Did you want to do add anything else to this @yuvipanda ?

docs/topic/features.md
```diff
@@ -34,9 +34,23 @@ the data. This is very commonly used by organizations that provide big datasets
...
 Hence, this is an opt-in feature.
 
+```{important}
+This feature enables the hub users to access `Requester Pays` buckets,
+**outside** of their project.
+
+However, note that this feature **does not** control which buckets **inside**
+the project will have `Requester Pays` enabled for themselves.
+
+This can be checked from the console following these steps in the
```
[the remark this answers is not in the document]
GeorgianaElena: Ahhh!! The last missing puzzle piece in my head. Thanks for pointing this out. I've added two commits trying to clarify this. There's also a technical example, because _I don't think_ we have done this before?

(I will also follow-up on the freshdesk ticket as well).

docs/howto/features/buckets.md
```diff
@@ -106,6 +106,35 @@ user_buckets = {
...
+for the appropriate bucket, and running `terraform apply`.
+
+Example:
+
+```terraform
+user_buckets = {
+   "persistent": {
+      "delete_after": null,
+      "public_access": true,
+      "requester_pays": true
```
GeorgianaElena, line 133: @yuvipanda, do you think `requester_pays` has sense outside of a public access? Should these two flags be documented togheter?

docs/topic/features.md
```diff
@@ -34,9 +34,26 @@ the data. This is very commonly used by organizations that provide big datasets
 on Google Cloud storage, to sustainably share costs of maintaining the data.
 
 When this feature is enabled, users on a hub accessing cloud buckets from
-other organizations marked as 'requestor pays' will increase our cloud bill.
+other organizations marked as 'Requester Pays' will increase our cloud bill.
 Hence, this is an opt-in feature.
 
+```{important}
+This feature enables the hub users to access `Requester Pays` buckets,
+**outside** of their project.
```
sgibson91, line 42: Here, we talk about using requester pays so hub users can access buckets **outside** their hub... (thought continued in another comment)

docs/howto/features/buckets.md
```diff
@@ -106,6 +106,35 @@ user_buckets = {
...
+but not be billed for the charges associated with making and executing the
+requests on this bucket.
+
+By enabling the [Requester Pays flag](https://cloud.google.com/storage/docs/using-requester-pays#using),
+the requesters are required to include a billing project in their requests,
+which will mean that the billing will happen on the requester's project.
+
+Enabling Requester Pays is useful, for example, if the communities have a lot
+of data that they want to make available to others, but don't want to be
+charged for their access to that data.
```
sgibson91, line 121: ... but this section sounds like the host hub is providing access to buckets **inside** their hub. Which is it? Can it be both? In what scenario is the bucket **inside** and what scenario is the bucket **outside**?

GeorgianaElena: Yes, it can be both. `Requestor Pays` is a feature that a bucket can have.

- If buckets **outside** the project have this flag, then we need to:
     - set `hub_cloud_permissions.allow_access_to_requestor_pays_buckets` in the terraform config of a cluster (this config is simply called `requestor_pays` without this PR, which was the source of confusion for me).
     - This is to allow them to be charged on their project for access of such **outside** buckets
- But buckets that we set for communities, **inside** their projects can also have this feature enabled on them, which means that other people outside will be charged for their usage.
    - This can be enabled by setting `user_buckets.<bucket_name>.requester_pays` to true in the terraform config of the cluster.
    - I don't think we have any buckets like this yet, but was recently requested by leap in https://2i2c.freshdesk.com/a/tickets/1316

GeorgianaElena: Does the above has more sense @sgibson91? I will update the language in the docs as well to make the split and differences more clear.

sgibson91: Yeah, thank you - though I'm wondering if it should be `hub_cloud_permissions.allow_access_to_external_requestor_pays_buckets` to be crystal clear? Though I realise it's getting kinda long.

docs/howto/features/buckets.md
yuvipanda: Ah, great question @GeorgianaElena.

Right now, we don't actually have a terraform variable called `requstor_pays` (see list for [gcp](https://github.com/2i2c-org/infrastructure/blob/083419739d40e4cb5fbd0e8ebdae9a59465b46ce/terraform/gcp/variables.tf#L255) and [aws](https://github.com/2i2c-org/infrastructure/blob/083419739d40e4cb5fbd0e8ebdae9a59465b46ce/terraform/aws/variables.tf#L25) for what we do have). So we currently *don't* have a way to mark a bucket as requestor pays via our terraform. 

It's not super hard to implement (pass through to [gcp config](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_bucket#requester_pays) or [this aws resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_request_payment_configuration)), so we should implement it if / when a community asks to make one of their buckets into requestor_pays.

It may also be helpful to rename `public_access` to `public_anonymous_access` - basically, it currently means that *anyone* can access that bucket, without needing to provide any authentication at all. But since `requestor_pays` *requires* that the entity requesting this be part of an AWS or GCP project, it by definition is *authenticated* - so GCP or AWS knows who to charge! This 'authentication' is what our hubs provide when we turn on the `allow_access_to_requestor_pays_buckets` flag.

So `public_access` (or `public_anonymous_access`) is actually *mutually exclusive* with `requestor_pays`. The first is fully unauthenticated, while the second requires authentication.

I hope this is helpful!

terraform/aws/variables.tf
```diff
@@ -44,17 +44,23 @@ variable "user_buckets" {
 }
 
 variable "hub_cloud_permissions" {
-  type        = map(object({ requestor_pays : bool, bucket_admin_access : set(string), extra_iam_policy : string }))
+  type = map(object({
+    allow_access_to_requestor_pays_buckets : optional(bool, false),
```
yuvipanda, line 48: Turns out we actually never implemented this on AWS! So we should just remove this from here.

docs/howto/features/buckets.md
```diff
@@ -106,6 +106,35 @@ user_buckets = {
 }
 ```
 
+## Enabling Requester Pays flag on buckets
```
yuvipanda, line 109: We currently don't have this functionality at all, so we should turn this into a feature request issue and remove this section. See https://github.com/2i2c-org/infrastructure/pull/3431/files#r1498313106 for some more info.

docs/howto/features/buckets.md
yuvipanda: I just want to say that https://github.com/2i2c-org/infrastructure/pull/3431/files#r1495651158 is an accurate description of the situation! Only difference is that right now,

> This can be enabled by setting user_buckets.<bucket_name>.requester_pays to true in the terraform config of the cluster.

is not true because we don't have that implemented.

yuvipanda (commented):
Let's simplify this PR to:

- [x] Renaming the existing `requestor_pays` to `allow_access_to_external_requestor_pays_buckets` (as @sgibson91 suggested in https://github.com/2i2c-org/infrastructure/pull/3431/files#r1495801966) for clarity
- [x] Document what exactly that does in clearer terms. I think your words here (https://github.com/2i2c-org/infrastructure/pull/3431/files#diff-022e962ded687634ecdad02e21584d886b7e0ed8002cd91e2e9b94f5637c289cR58) are great, and maybe can be ported to the features page?
- [ ] Open an issue to track the new feature request of marking a bucket we manage as `requestor_pays` to the outside world.

I hope this helps! I can't 'request changes' because this is my own PR, but thanks for picking this up, @GeorgianaElena

docs/howto/features/buckets.md
GeorgianaElena: Extremely helpful @yuvipanda! Thank you very much for taking the time to clarify this <3

GeorgianaElena:
I've updated the PR per instructions and the terraform config of the clusters. I've also opened the follow-up issue, so this should be ready for review again 🚀

terraform/gcp/variables.tf
```diff
@@ -415,9 +415,11 @@ variable "hub_cloud_permissions" {
   Key is name of the hub namespace in the cluster, and values are particular
   permissions users running on those hubs should have. Currently supported are:
 
-  1. requestor_pays: Identify as coming from the google cloud project when accessing
-     storage buckets marked as  https://cloud.google.com/storage/docs/requester-pays.
-     This *potentially* incurs cost for us, the originating project, so opt-in.
+  1. allow_access_to_external_requester_pays_buckets: Allow code running in user servers from this
+     hub to identify as coming from this particular GCP project when accessing GCS buckets
```
sgibson91, line 419:
```suggestion
     hub to identify as coming from this particular GCP project when accessing GCS buckets in other projects
```

sgibson91 (approved):
LGTM!

GeorgianaElena:
Yuhuuu! Thank you @sgibson91!

GeorgianaElena:
And thank you @yuvipanda for shining light over this confusing topic 🎉

yuvipanda:
This is great work @GeorgianaElena!

Issue #3430, opened 2023-11-16
[EVENT] Openscapes 10 December


GeorgianaElena:
Requested in https://2i2c.freshdesk.com/a/tickets/1121

### Event Info

- Space Station Synergies (full day, 8:00 am-3:00 pm PST) 

    expected usage: 14.8 GB per person

- Earthdata Cloud (half day, 1:00 - 4:30 pm PST)

    expected usage: 7.4 GB per person

- **Community Representative:**  [BriannaLind](https://github.com/BriannaLind)
- **Number of attendees:** ~50

damianavila:
@GeorgianaElena, given https://github.com/2i2c-org/infrastructure/pull/3262 and https://github.com/2i2c-org/infrastructure/issues/3403#issuecomment-1814695693, both without issues, do you think we need any further preparation (ie. bigger box if the 7.4GB per person)?

GeorgianaElena:
### Event Todos
I believe the todo list for this event is:

- [ ] Do something like https://github.com/2i2c-org/infrastructure/pull/3262 again. 

- [ ] Also, they've requested to make the `shared` directory be read-only for the duration of the event, as atm it is writable by all users (they requested this change in the past).

### Node choice reasoning

The reasoning for the node type choice is:

There are two events happening one with 14.8GB expected usage and another one with 7.4GB and they will also overlap. At 50 users, we would need:

  - ~25 `r5.xlarge` nodes to come up for the ~14.8GB profile. I believe the number of nodes it's pretty high and half of the users will experiment longer startup times because of this.
   - ~13 `r5.xlarge` nodes to come up for the 7,4GB profile. This number is decent enough, but because there are two events, we should consider choosing based on that.

In total, there will be around 40 nodes running by the time the second event starts.

But if we go with `r5.4xlarge` nodes with 4 times as much memory, then at 50 users, we would need:
   - ~7 `r5.4xlarge` nodes nodes to come up for the ~14.8GB profile
   - ~3 `r5.4xlarge` nodes to come up for the 7,4GB profile.
 
In total, there will be around 10 nodes running by the time the second event starts.

PR #3436, opened 2023-11-17
[maint] Simplify and change format of the event new issue template

GeorgianaElena:
This is how it looks rendered https://github.com/GeorgianaElena/.github/issues/new?assignees=&labels=event&projects=&template=event.yml&title=%5BEVENT%5D+%7B%7B+HUB+NAME+%7D%7D

While opening [this recent event issue](https://github.com/2i2c-org/infrastructure/issues/3430), I realized that our issue template for new events is not matching current expectations:

- support triagers are the ones that open these type of issues now and not communtiy reps
-  we do not encourage creating separate node-pools before events anymore, and altough we still don't have clear documentation about events, I figured we should remove it entirely to not cause confusion. 
     - although it might make sense to add separate nodepools for hubs with events on shared clusters 🤔 https://infrastructure.2i2c.org/howto/features/dedicated-nodepool/
- we don't ask for feedback from the communities consistently and we don't do anything with that feedback yet, so I removed it. Also, since the communication should happen on support, this feedback questionare I believe should be stored as a template ticket in freshdesk instead.

Follow-up taks after merging this PR:
- [ ] Move the feedback request in a template in freshdesk instead.
- [ ] Follow-up issue about moving the `Engineer task list` from the issue template to the infra docs, where some reasoning about creating a separate nodepool on a shared cluster and resource allocation strategies could be added.


consideRatio (approved):
I agree with the decisions motivated in the PR description, and this seems like a great incremental improvement!

Issue #1924, opened 2022-11-16
Get rid of eksctl from our infrastructure


yuvipanda:
Almost 3y after filing, https://github.com/aws/containers-roadmap/issues/724 is finally fixed on the EKS side! This means we can switch away from using eksctl, and towards using terraform for our AWS setup - same as we do for GCP, Azure, etc.

I'd suggest we pick a single cluster and move it, and see how that goes.

# TODO

- [ ] Write Terraform Code that uses [managed nodegroups](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group) for nodegroups (EKS equivalent of nodepools)
- [ ] Upgrade to k8s 1.24 on EKS, as that is necessary for managed nodegroups to work properly
- [ ] Make sure our nvidia driver install is moved back into the support chart - this was moved to be handled by eksctl in https://github.com/2i2c-org/infrastructure/pull/1787

Once this is done, we can evaluate moving all the other ones gradually.

PR #3445, opened 2023-11-21
eksctl: set nodegroup name based on namePrefix, nameBase, and nameSuffix

consideRatio:
This feature enables us to create new node pools of the same instance type, allowing us to create a new node pool while waiting for another to become phased out. I wanted this for `nasa-cryo` to finalize a k8s upgrade.

### No phaseout guarantee

To add a new side by side with an old isn't enough, we would want to _ensure that the next scale up happens on the new node pool rather than the old_. I tried finding a way to guarantee this for a long time unsuccessfully using `tags` on AWS auto scaling groups, launch templates, and launch template versions. It is possible to constrain the old node pool by declaring `maxSize` to its current size, but that just stops it from growning and doesn't guarantee it shrinks but I'm not sure what else can be done.

### Misc refactors for jsonnet beginners like myself

As part of this, I also learned a lot about jsonnet and tried to help make the .jsonnet files a bit more easy to understand going onwards for jsonnet newbies like me.

### Related

- https://github.com/2i2c-org/infrastructure/issues/1924

### DIFF
--- a/eksctl/libsonnet/nodegroup.jsonnet
+++ b/eksctl/libsonnet/nodegroup.jsonnet
@@ -10,10 +10,15 @@
 
 // Exported object
 {
-  // If using spot instances where a Auto Scaling Group (ASG) has multiple
-  // instances associated with it, label using the first instance type.
-  // FIXME: Clarify the limitations of picking one of multiple instance types
+  // Note that spot instances configured via an Auto Scaling Group (ASG) can have
+  // multiple instance types associated with it (with the same CPU/RAM/GPU), we
+  // label them using the first instance type in the ASG as a compromise.
+  //
+  // More details at:
+  // https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md#using-mixed-instances-policies-and-spot-instances
+  //
   local instanceType = if 'instanceType' in $ then $.instanceType else $.instancesDistribution.instanceTypes[0],
+  // NodeGroup names can't have a '.' in them as instanceTypes has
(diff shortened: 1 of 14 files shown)


eksctl/libsonnet/nodegroup.jsonnet
```diff
@@ -1,34 +1,54 @@
...
+*/
 
+// Exported object
 {
-  name: '',
+  // If using spot instances where a Auto Scaling Group (ASG) has multiple
+  // instances associated with it, label using the first instance type.
+  // FIXME: Clarify the limitations of picking one of multiple instance types
+  local instanceType = if 'instanceType' in $ then $.instanceType else $.instancesDistribution.instanceTypes[0],
+  local escapedInstanceType = std.strReplace(instanceType, ".", "-"),
```
GeorgianaElena, line 17: Let's move this comment that was in cluster's jsonnet before, here since we've changed where we do the name escaping.
```suggestion
  // // NodeGroup names can't have a '.' in them, while instanceTypes always have a .
  local escapedInstanceType = std.strReplace(instanceType, ".", "-"),
```

GeorgianaElena (approved):
I looked over `nodegroup.jsonent` and it looked correct to me. Also, it was easier to follow than the original version thanks to the comments and the splitting of some of the complex logic. Thank you!

eksctl/libsonnet/nodegroup.jsonnet
consideRatio: Thank you for reviewing @GeorgianaElena, resolved by f1b7c8a

Issue #1510, opened 2022-07-11
2i2c datasource showing up twice in central grafana


GeorgianaElena:
We currently use the script at https://github.com/2i2c-org/infrastructure/blob/master/deployer/commands/grafana/central_grafana.py to add all the (authenticated) prometheus servers as datasources of the central grafana.

However, since the central Grafana is actually the 2i2c grafana, its default datasource is set through the config at https://github.com/2i2c-org/infrastructure/blob/7a1e75a1b8bb726e873e495944323d732870e06f/helm-charts/support/values.yaml#L86-L96 as being the same prometheus server, but without using basic auth since they're running the same private network.

This results in the central grafana showing the same info about the 2i2c cluster, one labeled as 2i2c and one labeled as prometheus.

We should find a way to only show info about the 2i2c cluster once

PR #3450, opened 2023-11-24
Central grafana datasources and cli fixes

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/1510.

This PR:
- renames the `prometheus` datasource of the 2i2c's Grafana to `2i2c`
- creates a new deployer sub-command called `grafana central-ds` to manage the datasources in the central Grafana
- replaces the `deployer grafana update-central-datasources` command that added all clusters in the repo as datasources, in bulk to four individual commands:
   - one to add a cluster as datasource
   - one to delete a cluster from the datasources list
   - one to list all of the clusters that are not marked as datasources in the 2i2c grafana, but doesn't take any action
   - one to list all of the datasources in the central grafana that don;t have a cluster counterpart in this repo
- updates the add cluster / rm cluster with central grafana update steps.

Note, that "renaming" the `prometheus` data source was a bit tricky:

- this was a config that was provisioned through config, and it's marked as not editable in our support chart https://github.com/2i2c-org/infrastructure/blob/5e12d77b0037706f42cbba1ba88d96aa8794262b/helm-charts/support/values.yaml#L373-L386, so it had to be deleted through config
- Followed the docs at https://grafana.com/docs/grafana/latest/administration/provisioning/#data-sources to add a `deleteDatasources` list entry
- Then renamed the default provisioned datasource of the 2i2c cluster to be 2i2c instead of `prometheus`
- Then removed the  `deleteDatasources` entry.
- Note that the datasources in `deleteDatasources` are first deleted, then whatever is listed under `datasouces` is added.

----
:books: Documentation preview :books:: https://2i2c-pilot-hubs--3450.org.readthedocs.build/en/3450/

### DIFF
--- a/deployer/README.md
+++ b/deployer/README.md
@@ -317,11 +317,20 @@ If such a file doesn't already exist, it will be created by this function.
   Updates:
   - the content of `enc-grafana-token.secret.yaml` with the new token if one already existed
 
-#### `grafana update-central-datasources`
+#### The `grafana central-ds` sub-command
+This is a sub-command meant to help engineers to programmatically manage the datasources registered in the 2i2c central Grafana at https://grafana.pilot.2i2c.cloud.
 
-Ensures that the central grafana at https://grafana.pilot.2i2c.cloud is
-configured to use as datasource the authenticated prometheus instances of all
-the clusters that we run.
+##### `grafana central-ds add`
+Adds a new cluster as a datasource to the central Grafana.
+
+##### `grafana central-ds remove`
+Removes a datasource from the central Grafana. In the unlikely case, the datasource has a different name than the cluster's name, pass it through the optional `datasource-name` flag.
+
+##### `grafana central-ds get-add-candidates`
+Gets the clusters that are in the infrastructure repository but are NOT registered in central grafana as datasources. Usually this happens when a new clusters was added, but its prometheus server was not mapped to a datasource of the central 2i2c Grafana. This list can then be used to know which datasources to add.
+
+##### `grafana central-ds get-rm-candidates`
+Gets the list of datasources that are registered in central grafana but are NOT in the list of clusters in the infrastructure repository. Usually this happens when a clusters was decommissioned, but its prometheus server was not removed from the datasources of the central 2i2c Grafana. This list can then be used to know which datasources to remove.
 
 ### The `validate` sub-command
 
--- a/deployer/commands/grafana/central_grafana.py
+++ b/deployer/commands/grafana/central_grafana.py
@@ -27,6 +27,16 @@ from .utils import (
 
 yaml = YAML(typ="safe")
 
+# Creates a new typer application, called "central"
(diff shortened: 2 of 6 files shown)


deployer/commands/grafana/central_grafana.py
```diff
@@ -98,69 +119,129 @@ def get_clusters_used_as_datasources(cluster_name, datasource_endpoint):
...
+    ),
+    datasource_name=typer.Option(
+        "",
+        help="The name of the datasource we're removing. This is usually the same with the name of cluster.",
+    ),
+):
+    """
+    Remove a datasource that maps to a cluster, from the central Grafana.
+    In the unlikely case, the datasource has a different name than the cluster's name,
+    pass it through the optional `datasource-name` flag.
```
consideRatio, line 223:
```suggestion
    pass it through the optional `--datasource-name` flag.
```

deployer/README.md
```diff
@@ -317,11 +317,20 @@ If such a file doesn't already exist, it will be created by this function.
...
+This is a sub-command meant to help engineers to programmatically manage the datasources registered in the 2i2c central Grafana at https://grafana.pilot.2i2c.cloud.
 
-Ensures that the central grafana at https://grafana.pilot.2i2c.cloud is
-configured to use as datasource the authenticated prometheus instances of all
-the clusters that we run.
+##### `grafana central-ds add`
+Adds a new cluster as a datasource to the central Grafana.
+
+##### `grafana central-ds remove`
+Removes a datasource from the central Grafana. In the unlikely case, the datasource has a different name than the cluster's name, pass it through the optional `datasource-name` flag.
```
consideRatio, line 327:
```suggestion
Removes a datasource from the central Grafana. In the unlikely case, the datasource has a different name than the cluster's name, pass it through the optional `--datasource-name` flag.
```

sgibson91 (approved):

consideRatio (approved):
Wieee @GeorgianaElena, looks great to me!!

GeorgianaElena:
I haven't merged this yet, because apparently apart from having `2i2c` and `prometheus` point to the same prometheus data source there is also something called `default` that appears in the list of clusters in each of the Grafana panels.

However this `default` is not present in the list of datasource (like `prometheus` was), so I've been trying to figure out if I can get rid of it.

GeorgianaElena:
Ok, finally found it 🎉 I guess looking at the same problem with fresh eyes, really does help.

Apparently having a datasource be marked as the default one, causes a placeholder called `default` be shown in the Grafana panels, but not in the datasources. Not having a default datasource explicitly set, solves it, and the first datasource (alphabetically) is shown as the default one in panels.

Also, I've updated the support chart to not set a default for either of the cluster's Grafana, otherwise, you are presented with two possible `clusters` in the panels, which is confusing.

Screenshot 2023-11-30 at 10 29 17

PR #3455, opened 2023-11-27
carbonplan, qcl, ubc-eoas: cleanup no longer needed pullPolicy

consideRatio:
With KubeSpawner 6.2.0 we can cleanup declaring an explicit pullPolicy to revert KubeSpawner's previous behavior that was to explicitly set this to IfNotPresent and instead rely on the k8s default behavior of `Always` for all `:latest` tags. For details see https://github.com/jupyterhub/kubespawner/pull/807.


GeorgianaElena (approved):
🎉

Issue #3461, opened 2023-11-27
LEAP had a temporary outage due to core node being upgraded


yuvipanda:
Reported in https://2i2c.freshdesk.com/a/tickets/1141, and I investigated immediately, and could catch it in the act.

I noticed that the core nodes were new, and looking at the versions, see that it's uneven:

<img width="710" alt="image" src="https://github.com/2i2c-org/infrastructure/assets/30430/3db9f3e5-f974-4c03-aa94-ad3e5ce6af1e">

And the core nodes had *just* updated, even though auto upgrade is set to false

<img width="356" alt="image" src="https://github.com/2i2c-org/infrastructure/assets/30430/f0551acf-672d-4f18-9e6a-8c6cda776051">

Unfortunately, `gcloud container operations list` returns empty, and looking at cloud audit events with the following query also returns nothing:

```
logName="projects/leap-pangeo/logs/cloudaudit.googleapis.com%2Factivity"
resource.type="k8s_cluster"
-protoPayload.methodName=~"io.k8s.*"
```

So the question is:

- [ ] Why did this auto upgrade happen?
- [ ] What can we do to prevent this?

yuvipanda:
It happened again, and not due to an upgrade. It's possible the previous time was not an upgrade either - I had just assumed that due to the fact that it was a different version, but entirely possible that was the prior version too. I did see the instancegroup change, but that may have been just the autoscaler bringing nodes up.

yuvipanda:
Almost all of them say `Deleting node gke-leap-cluster-core-pool-1592e1d3-nshx because it does not exist in the cloud provider` *after* the post NodeNotReady, so let's go look at the cloud provider to see what is up.

yuvipanda:
And at the instance group's page (https://console.cloud.google.com/compute/instanceGroups/details/us-central1-c/gke-leap-cluster-core-pool-1592e1d3-grp?project=leap-pangeo), you can see this oscillation:

<img width="1628" alt="image" src="https://github.com/2i2c-org/infrastructure/assets/30430/038da8df-084d-46ff-a8e8-4835e4c5a4c4">

yuvipanda:
I looked at all audit events that involved this particular instance group with:

```
resource.type="gce_instance_group"
resource.labels.instance_group_name="gke-leap-cluster-core-pool-1592e1d3-grp"
```

and I could find 'removeInstance' calls here:
```
{
  "protoPayload": {
[...]

PR #3462, opened 2023-11-28
Tell cluster autoscaler to leave the hub pod alone

yuvipanda:
Prevents the autoscaler from evicting the hub pod, and resolves an outage caused by the autoscaler bouncing around a lot.

Ref: https://github.com/2i2c-org/infrastructure/issues/3461


yuvipanda:
I'm going to merge this now to prevent further possible outages.

PR #3471, opened 2023-11-29
[openscapes] Bump matlab image tag

sgibson91:
Requested in https://2i2c.freshdesk.com/a/tickets/1142


yuvipanda (approved):

yuvipanda:
Thanks @sgibson91! I think image tag updates can generally be self merged as well, especially when they are explicitly asked for by the community.

sgibson91:
Yeah, Openscapes catches me off guard because their profile list isn't centralised in the common file, and then it's like "do they want staging, or, prod, or both?" It's extra decision fatigue.

PR #3478, opened 2023-11-30
Use GKE's native NVidia driver installer for GPUs

yuvipanda:
- Remove our custom GPU installer daemonset, as GKE now supports automatically doing it (like eksctl does).
- Switch from installing 'latest' to using default driver, which is slightly older (version 470 with CUDA 11.4, vs version 530 with CUDA 12). There seems to be a bug with the latest driver causing the GPU to not be usable by non-root users, so let's stick to this until that is resolved.
- Apply these changes to LEAP hub already. m2lines is about to be decomissioned, so not necessary.


sgibson91 (approved):
🎉

PR #3490, opened 2023-12-01
Add an announcement to UToronto's homepage about maintenance

sgibson91:
Requested in https://2i2c.freshdesk.com/a/tickets/1150

---

Note: I'm having difficulty deploying this locally to staging to see what it looks like. I would appreciate anyone with the appropriate access to help me out on this.

### DIFF
--- a/config/clusters/utoronto/common.values.yaml
+++ b/config/clusters/utoronto/common.values.yaml
@@ -38,13 +38,15 @@ jupyterhub:
         announcements:
           - |
             <div>
-              <strong>ARC will be making changes to our educational JupyterHub on January 9, 2023</strong>: R/RStudio will be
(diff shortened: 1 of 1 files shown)


config/clusters/utoronto/common.values.yaml
```diff
@@ -36,6 +36,16 @@ jupyterhub:
           name: University of Toronto
           url: https://www.utoronto.ca/
         announcements:
+          - |
+            <div>
+              <strong>ARC will be making changes to our educational JupyterHub on January 9, 2023</strong>: R/RStudio will be
```
yuvipanda, line 41: Let's turn this into an `<h4>` to match what is in the prior announcement? We can then put the other two paragraphs inside `<p>` tags so it looks ok. Looks good to merge after those changes.

sgibson91:
Ended up manually deploying this with #3492

PR #3492, opened 2023-12-02
Regenerate utoronto kubernetes cluster credentials

yuvipanda:
Looks like this may have expired. This was regenerated with:

- `az login`, and logging into Azure with appropriate credentials
- `sops -i -d <path-to-credential-file>`, to decrypt the credentials
- `export KUBECONFIG=<path-to-credential-file>`, to tell `kubectl` to use our checked in credentials. You can test that the credentials no longer work with `kubectl get node` - you should get an error message about authorization
- Get a list of running clusters with `az aks list -o table`
- Regenerate credentials with `az aks get-credentials --name <resource-group> --resource-group <resource-group>`. For the utoronto cluster, this is `az aks get-credentials --name=hub-cluster --resource-group 2i2c-utoronto-cluster`. You may be asked if you want to overwrite the current credentials - answer 'yes'.
- Re-encrypt the file, with `sops -i -e <path-to-credential-file>`.
- Tada


sgibson91:
I am going to merge this so my deploy succeeds

Issue #3484, opened 2023-12-01
[Decommission Hub] M2LInES


colliand:
### Summary

The 2i2c + M2LInES contract expired a while ago. The hub should be decommissioned to prevent continued accrual of cloud costs. On 2023-11-30, @colliand sent email to M2LInES community representative Professor Laure Zanna forecasting the decommission. 

### Info

- **Community Representative:** @LaureZanna 
- **Link to New Hub issue:** #1168 
- **Proposed end date:** 2023-12-04
#### Phase 1: Hub Removal

(These steps are described in more detail in the docs at https://infrastructure.2i2c.org/en/latest/hub-deployment-guide/hubs/other-hub-ops/delete-hub.html)

- [x] Manage existing data (**delete data**)
- [x] Delete the hub's authentication application on GitHub or CILogon (note CILogon removal requires the hub config in place)
- [ ] Remove the appropriate `config/clusters/<cluster_name>/<hub_name>.values.yaml` files. A complete list of relevant files can be found under the appropriate entry in the associated `cluster.yaml` file.
- [ ] Remove the associated hub entry from the `config/clusters/<cluster_name>/cluster.yaml` file.
- [ ] Remove the hub deployment
  - `helm --namespace HUB_NAME delete HUB_NAME`
  - `kubectl delete namespace HUB_NAME`

#### Phase 2 - Cluster Removal

_This phase is only necessary for single hub clusters._

- [ ] Remove the cluster's datasource from the central Grafana with:
  - `deployer grafana central-ds remove CLUSTER_NAME`
- [ ] Run `terraform plan -destroy` and `terraform apply` from the [appropriate workspace](https://infrastructure.2i2c.org/en/latest/topic/terraform.html#workspaces), to destroy the cluster
- [ ] Delete the terraform workspace: `terraform workspace delete <NAME>`
- [ ] Remove the associated `config/clusters/<cluster_name>` directory and all its contents
- Remove the cluster from CI:
  - [ ] [`deploy-hubs.yaml`](https://github.com/2i2c-org/infrastructure/blob/HEAD/.github/workflows/deploy-hubs.yaml)
  - [ ] [`validate-clusters.yaml`](https://github.com/2i2c-org/infrastructure/blob/HEAD/.github/workflows/validate-clusters.yaml)
- [ ] Remove the cluster from the list of grafana datasources at https://grafana.pilot.2i2c.cloud/datasources
- [ ] Remove A record from Namecheap account

[...]

PR #3496, opened 2023-12-04
Decomission m2lines

yuvipanda:
- `terraform destroy` already run
- Buckets were deleted from the web console
- GitHub apps were deleted

Fixes https://github.com/2i2c-org/infrastructure/issues/3484


GeorgianaElena (approved):
LGTM! @yuvipanda, I believe `deployer grafana central-ds remove m2lines` needs to be also ran so that the m2lines cluster will not appear in the 2i2c central grafana.

Hmm, I don't think it will work without the config files, so we might need to be doing it manually. I'll look into it.

yuvipanda:
Thank you, @GeorgianaElena!

PR #3502, opened 2023-12-06
[merge this on Dec 10 morning CET] Openscapes event

GeorgianaElena:
Closes https://github.com/2i2c-org/infrastructure/issues/3430

- I didn't get a confirmation about changing the `shared` directory to `readonly` persistently on the hub, so I've assumed we should only do it for the duration of the event
- I've changed the node type from `r5.xlarge` to `r5.4xlarge` for the `14.8GB` profile to have the 50 expected users on ~7 nodes, this means only 14% of the users will experiment longer startup times due to a new node coming up as opposed to 50% of them with the original node type
- I've left the `7.4GB` profile on `r5.xlarge`, as this will mean the 50 expected users of the second event on ~13 nodes, which means ~26 users of the users will experiment longer startup time, which I believe is acceptable.

### DIFF
--- a/config/clusters/openscapes/common.values.yaml
+++ b/config/clusters/openscapes/common.values.yaml
@@ -42,11 +42,15 @@ basehub:
           - name: home
             mountPath: /home/jovyan/shared
             subPath: _shared
-            readOnly: false
+            # To be changed back to r-w after Dec 10th
+            # https://2i2c.freshdesk.com/a/tickets/1121
+            readOnly: true
           - name: home
             mountPath: /home/rstudio/shared
             subPath: _shared
-            readOnly: false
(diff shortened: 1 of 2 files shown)


sgibson91 (approved):

yuvipanda:
I love that this requires no eksctl changes, only helm changes.

GeorgianaElena:
I've missed merging this because I've miscalculated and thought Dec 10th was today, on a Monday, but it was yesterday :((((

The event has now passed, so closing this one 😞

yuvipanda:
I spent last evening with them at a happy hour and they had no complaints :) don’t worry @GeorgianaElena!

Issue #2434, opened 2023-03-29
Regenerate NASA SMCE AWS account credentials every 60 days


yuvipanda:
## Cluster status

Every 60 days / two months, we need to re-generate the deployer credentials for each nasa AWS account we manage, this issue describes the status of the accounts and how to do it. Update the status below if you have re-generated credentials.

Cluster | Re-generation history
-|-
`nasa-esdis` | April 23 via #3979, Feb 28 via #3747, Jan 5 via #3575, Dec 6 2023 via #3503
`nasa-ghg` | April 23 via #3979, Feb 28 via #3747, Jan 5 via #3575, Dec 12 2023 via #3528, Nov 19 2023 via #3442, Oct 4 2023 via #3220
`nasa-veda` | April 23 via #3979, Feb 28 via #3747, Jan 5 via #3575, Dec 6 2023 via #3506, Nov 19 2023 via #3442, Oct 4 2023 via #3219

Upcoming regeneration: 23 April

## How to re-generate credentials for deployer

https://github.com/2i2c-org/infrastructure/issues/2339 is related, but this is specifically for the continuous deployer key for individual nasa AWS accounts.

Here's the how to on doing this:

1. Authenticate yourself - see https://repost.aws/knowledge-center/authenticate-mfa-cli on how to
2. ```shell
   cd terraform/aws
   ```
3. ```shell
   CLUSTER_NAME=...
   ```
4. ```shell
   # put yourself in the right workspace
   export TF_WORKSPACE=$CLUSTER_NAME
   ```
5. ```shell
   # replace previous credentials with new
   terraform apply -replace=aws_iam_access_key.continuous_deployer -var-file=projects/$CLUSTER_NAME.tfvars
   ```
6. ```shell
   # write new credentials to file, and then encrypt them in place
   terraform output -raw continuous_deployer_creds > ../../config/clusters/$CLUSTER_NAME/enc-deployer-credentials.secret.json
   sops -i -e ../../config/clusters/$CLUSTER_NAME/enc-deployer-credentials.secret.json
   ```
7. ```shell
   # verify function of new credentials
   deployer use-cluster-credentials $CLUSTER_NAME
   ```
8. ```shell
   # add relevant files updated, and then...
   git commit -m "nasa smce clusters: re-generate deployer credentials"
   ```

yuvipanda:
This needs to be done every 60 days :(

jmunroe:
[...]

PR #3528, opened 2023-12-12
Regenerate nasa-ghg deployer access key

sgibson91:
ref: https://github.com/2i2c-org/infrastructure/issues/2434


GeorgianaElena (approved):
Thanks @sgibson91

PR #3529, opened 2023-12-13
Bump qcl homedir size again

GeorgianaElena:
Per https://2i2c.slack.com/archives/C041E05LVHB/p1702435988936249


GeorgianaElena:
self-merging as I already performed the change from the console

PR #3535, opened 2023-12-16
cloudbank: Explicitly specify what nodepools users should go to

yuvipanda:
Otherwise, they still end up on the old 'user' nodepool, and it can not be decomissioned. With this, they will end up on the equivalent 'new' nodepool, and then I can get rid of the old one.

I have also had to adjust the CPU limit, as at some point I suppose we switched from 8 CPU nodes to 4 but did not adjust this.

### DIFF
--- a/config/clusters/cloudbank/common.values.yaml
+++ b/config/clusters/cloudbank/common.values.yaml
@@ -9,12 +9,15 @@ nfs:
 jupyterhub:
   singleuser:
     cpu:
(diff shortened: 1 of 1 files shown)


yuvipanda:
This also ensures users don't end up accidentally on one of the large nodepools, which would cost a lot more!

yuvipanda:
Self-merging as there aren't any other engineers around until end of year.

PR #3537, opened 2023-12-16
leap, 2i2c, cloudbank: Cleanup some unused nodepools

yuvipanda:
- Also removes the temp opt out variable in use
- Also handles a FIXME about core node size on cloudbank. This brought us from 2
  n2-highmem-4 nodes to 3 n2-highmem-2 nodes, which is cheaper by 25%!

Fixes https://github.com/2i2c-org/infrastructure/issues/3405

These changes have all been applied


yuvipanda:
Self merging as there aren't any other engineers around until end of year.

PR #3542, opened 2023-12-20
Update org url for GHG

slesaad:
# Changes
Updates template variables for US GHG Center to use the new earth.gov/ghgcenter url


yuvipanda:
Thanks @slesaad!

Issue #3562, opened 2024-01-02
Unused home directory cleanup feature


yuvipanda:
While helping openscapes folks cleanup their user access list (https://github.com/2i2c-org/infrastructure/issues/3240), I also helped them cleanup unused home directories. This isn't a service 2i2c offers to users yet, but I wanted to open this to track work to be done here long term - otherwise it just increases costs monotonically for our users (https://github.com/2i2c-org/infrastructure/issues/3240#issuecomment-1751218146). This issue mostly currently tracks the one-off job I helped do for them, but needs more work to be rolled out sustainably to others.

I wrote an impromptu script that helped us archive home directories of a list of users given in the commandline to S3.

<details>
<summary>python script</summary>
<pre>
#!/usr/bin/env python3
"""
Archive home directories onto object storage (like S3 or GCS).

Designed to be run manually, and takes care to not delete anything without a lot of
confirmation.
"""
import hashlib
import string
import sys
import shutil
import os
import argparse
import boto3
from botocore.exceptions import ClientError
from escapism import escape
from pathlib import Path
from contextlib import contextmanager
import tempfile
import time
import subprocess
from functools import cache

@cache
def get_tar_command() -> str:
    """
    Return the `tar` command to use.

    We use `gnu` tar for compressing files, and Mac OS ships with bsd tar by
    default. We detect this, and tell users to get gnu tar if needed for local
    testing. Should not be an issue when running on containers.
    """
    out = subprocess.check_output(["tar", "--version"]).decode()
    if out.startswith("tar (GNU tar)"):
        return "tar"
    else:
        # We may be on Mac OS, and GNU Tar is not installed by default
        # It can be installed from homebrew with `brew install gnu-tar`,
        # which provides `gtar`
        if shutil.which("gtar"):
            return "gtar"
        else:
            print("Could not find GNU Tar on the system", file=sys.stderr)
            print(
                "If on Mac OS, please install gnu-tar with the following command (if using homebrew) and try again",
                file=sys.stderr,
            )
            print("brew install gnu-tar", file=sys.stderr)
            sys.exit(1)

def validate_homes_exist(basedir: Path, usernames: list[str], ignore_missing: bool):
    """
    Validate that all given homedirectories for users exist
    """
    errors = []
    for username in usernames:
        escaped_username = escape(
            username, safe=set(string.ascii_lowercase + string.digits), escape_char="-"
        ).lower()
        # We should still protect against directory traversal attacks
[...]

PR #3563, opened 2024-01-02
Allow specifying an archival policy for buckets

yuvipanda:
Only enables this for openscapes, which is using it as described in https://github.com/2i2c-org/infrastructure/issues/3562.

### DIFF
--- a/docs/howto/features/buckets.md
+++ b/docs/howto/features/buckets.md
@@ -15,6 +15,9 @@ on why users want this!
       },
       "bucket2": {
          "delete_after": null
+      },
+      "bucket3": {
+         "archival_storageclass_after": 3
       }
    }
    ```
@@ -27,6 +30,12 @@ on why users want this!
    time* the object will be automatically cleaned up - this is
    very helpful for 'scratch' buckets that are temporary. Set to
    `null` to prevent this cleaning up process from happening, e.g., if users want a persistent bucket.
+
+   `archival_storageclass_after` (available only for AWS currently) transitions objects
+   created in this bucket to a cheaper, slower archival class after the number of days
+   specified in this variable. This is helpful for archiving user home directories or similar
+   use cases, where data needs to be kept for a long time but rarely accessed. This should
+   not be used for frequently accessed or publicly accessible data.
 
 2. Enable access to these buckets from the hub or make them publicly accessible from outside
    by [editing `hub_cloud_permissions`](howto:features:cloud-access:access-perms)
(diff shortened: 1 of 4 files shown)


yuvipanda:
I have already applied this.

consideRatio:
I did a `terraform plan` on a cluster where this hasn't been applied yet and where `archival_storageclass_after` wasn't configured (`victor`) and got this.

```terraform

  # aws_s3_bucket_lifecycle_configuration.user_bucket_expiry["scratch"] will be updated in-place
  ~ resource "aws_s3_bucket_lifecycle_configuration" "user_bucket_expiry" {
        id     = "victor-scratch"
        # (1 unchanged attribute hidden)

      + rule {
          + id     = "archival-storageclass"
          + status = "Enabled"

          + transition {
              + storage_class = "GLACIER_IR"
            }
        }

        # (1 unchanged block hidden)
    }

  # aws_s3_bucket_lifecycle_configuration.user_bucket_expiry["scratch-staging"] will be updated in-place
  ~ resource "aws_s3_bucket_lifecycle_configuration" "user_bucket_expiry" {
        id     = "victor-scratch-staging"
        # (1 unchanged attribute hidden)

      + rule {
          + id     = "archival-storageclass"
          + status = "Enabled"

          + transition {
              + storage_class = "GLACIER_IR"
            }
        }

        # (1 unchanged block hidden)
    }

Plan: 0 to add, 2 to change, 0 to destroy.
```

It seems off that the rule is created and put in a `Enabled` state even if `archival_storageclass_after` is an optional defaulting to `null` hasn't been explicitly configured. It also renders to an ambugious rule if the `transition.days` is omitted.

consideRatio:
@yuvipanda if the final form of this PR will make `terraform apply` in other clusters apply something harmless, it could be good if you ping engineers about it to provide a heads up that this is a change they  can expect to show and not worry about.

yuvipanda:
Thanks for the check, @consideRatio. I'm actually going to investigate if that stanza can be just made conditional entirely (instead of using enabled / disabled), so it's a no-op elsewhere. If not, I'll post to make sure other engineers are aware of a no-op change.

yuvipanda:
@consideRatio ok, updated so it is a no-op on clusters that don't use this feature.

consideRatio (approved):
Thank you @yuvipanda!

I merged this branch to master locally and tested applying this on 2i2c-aws-us: `No changes. Your infrastructure matches the configuration.`.

consideRatio:
I'll go for a merge here to ensure we get this in and we don't drop the ball on this - as I did now for a month.

yuvipanda:
Thank you, @consideRatio!

Issue #3565, opened 2024-01-03
Automatically pull ghg-docs repo on server start on GHG hub (same as VEDA hub)


yuvipanda:
Via https://github.com/2i2c-org/infrastructure/pull/3124 and follow-up PRs, the NASA VEDA hub automatically clones the [GHG Docs](https://github.com/US-GHG-Center/ghgc-docs) repo each time a user server starts. Per https://2i2c.freshdesk.com/a/tickets/1209, this behavior is desired for the GHG center hub as well.

batpad:
Copied this issue over to the `veda-jupyterhub` repo: https://github.com/NASA-IMPACT/veda-jupyterhub/issues/7 .

@yuvipanda I think you can close here if that looks good / seems like the more appropriate place to track this.

yuvipanda:
\o/ Great!

PR #3566, opened 2024-01-03
Use generated resource allocation options for veda hub

yuvipanda:
While starting to work on https://github.com/2i2c-org/infrastructure/issues/3565, I realized that VEDA was still using the older style 'node share' rather than the generated 'resource allocation' options. I've swapped over the options to now be based on images for users to choose and resource allocation options generated by our resource allocation script. This matches openscapes, and there has generally been pretty big positive feedback on this mode. It also gives more visibility to the R & QGIS options.

I've kept the initial cloning to only happen on the pangeo image as it currently exists, without making any changes. That should be cleaned up as part of https://github.com/2i2c-org/infrastructure/issues/3565

I also had to update the node-info.json file by running the appropriate command. 

Old:

image

image

New:

image

image

### DIFF
--- a/config/clusters/nasa-veda/common.values.yaml
+++ b/config/clusters/nasa-veda/common.values.yaml
@@ -102,7 +102,7 @@ basehub:
                   runAsUser: 1000
                   runAsGroup: 1000
           profile_options: &profile_options
-            resource_allocation:
+            resource_allocation: &profile_options_resource_allocation
               display_name: Resource Allocation
               choices:
                 mem_1_9:
@@ -208,6 +208,7 @@ basehub:
                 kubespawner_override:
                   image: "{value}"
               choices: {}
+            resource_allocation: *profile_options_resource_allocation
 
     scheduling:
       userScheduler:
--- a/config/clusters/nasa-veda/common.values.yaml
+++ b/config/clusters/nasa-veda/common.values.yaml
@@ -59,31 +59,146 @@ basehub:
             - wildintellect
     singleuser:
       defaultUrl: /lab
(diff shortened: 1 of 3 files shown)


yuvipanda:
I've pinged @freitagb and @slesaad on slack to take a look.

wildintellect:
I'd suggest just rounding down the numbers to the nearest 0.5?
e.g 1.9 = 1.5, 28.937 = 28.5 etc
As long as your not over promising it's accurate enough.

yuvipanda:
Positive feedback from Alex Mandel on slack:

image

Although there was a suggestion to do better rounding of the numbers.

wildintellect:
This does seems like an improvement, though I wonder why more than one section is necessary:
box1 - What image type
box2 (based on box1) - what machine size
box3 - only if needed for a custom image

Also linking to external descriptions/docs on the what software is installed or how to use the custom image choice.

yuvipanda:
@wildintellect that issue is pending https://github.com/jupyterhub/kubespawner/pull/778 getting fixed.

sgibson91 (approved):

sgibson91:
I've approved pending any tweaks asked for by the community :)

yuvipanda:
I'll open another issue to deal with the rounding as it's reasonably complex, and deploy this now. The community seemed to not object to this on slack :)

yuvipanda:
@wildintellect I opened https://github.com/2i2c-org/infrastructure/issues/3584 with the explanation for *why* we show the odd numbers in memory.

PR #3569, opened 2024-01-04
Bring in newer cryptnono version

yuvipanda:
I've been upgrading cryptnono quite a bit over the last few months, bringing in new detectors that have been quite effective on mybinder.org. We automatically bump cryptnono on our clusters (https://github.com/2i2c-org/infrastructure/pull/3482), but recent progress have included some breaking changes to the helm chart config.

This PR just brings in the new config changes, but does not change behavior in any real way. No new detectors are enabled.

I've re-measured resource usage for the individual daemonset container (rather than the initContainer) as that can now be set separately. This probably requires us to redo some of the resource allocation generated profiles, which I'll do once this is merged. However, it is an overall reduction in daemonset requests, so deploying this shouldn't result in any profile being undeployable.

Merging this should allow https://github.com/2i2c-org/infrastructure/pull/3482 to move forward as well.

### DIFF
--- a/helm-charts/support/values.yaml
+++ b/helm-charts/support/values.yaml
@@ -428,13 +428,14 @@ cryptnono:
         # Measured with the following prometheus queries:
         # Memory: sum(container_memory_usage_bytes{container="monero", namespace="support"}) by (instance)
         # CPU: sum(rate(container_cpu_usage_seconds_total{container="trace", namespace="support"}[5m])) by (instance)
-        # Seems to hover mostly around the 60Mi mark for memory, and generally less than 0.0002 in CPU
+        # Seems to hover mostly around the 60Mi mark for memory, and generally less than 0.0002 in CPU. But
+        # 1m (or 0.001) is the lowest that can be specified in kubernetes, so we use that.
         limits:
           memory: 128Mi
-          cpu: 0.005
+          cpu: 5m
         requests:
           memory: 64Mi
-          cpu: 0.0001
+          cpu: 1m
 
 # Configuration of templates provided directly by this chart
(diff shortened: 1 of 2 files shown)


helm-charts/support/values.yaml
```diff
@@ -396,30 +396,45 @@ cryptnono:
...
+        # Measured with the following prometheus queries:
+        # Memory: sum(container_memory_usage_bytes{container="monero", namespace="support"}) by (instance)
+        # CPU: sum(rate(container_cpu_usage_seconds_total{container="trace", namespace="support"}[5m])) by (instance)
+        # Seems to hover mostly around the 60Mi mark for memory, and generally less than 0.0002 in CPU
+        limits:
+          memory: 128Mi
+          cpu: 0.005
+        requests:
+          memory: 64Mi
+          cpu: 0.0001
```
consideRatio, line 437: I think the lowest resolution is 1/1024 of a CPU, and I recall that 2m was the practical minimum value for dockerd and containerd. I'm not sure if there are any enforcement to prevent specification of 0.1m or extremely low values, but I'd be inclined to not optimize this extreme and go for a value of at least 1m to avoid issues.

consideRatio (approved):
I made a better-safe-than-sorry comment on a detail, but this lgtm!

helm-charts/support/values.yaml
yuvipanda: Good catch, @consideRatio. Based on the 'note' in https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu, I switched to using `m` units and specified 1m - 0.0001 is definitely not enforceable.

PR #3571, opened 2024-01-04
Add an issue template for tickets that will not be immediately resolved by the triagers

### LINKED ISSUE #3477: [maint-support] Create an issue template for support tickets
I believe we should have an issue template for the issues that a support triager needs to open if a ticket cannot be resolved in the first 30m time frame.

It should collect information about:

- [ ] link towards the freshdesk ticket
- [ ] type of ticket (roughly one of the big categories in https://compass.2i2c.org/projects/managed-hubs/timeboxed-initial-ticket-evaluation/#support-timeboxed-evaluation)
- [ ] any useful information that was found during the initial 30m investigation 
- [ ] a label

Having such an issue template, will help gather all the useful info needed to later assign and resolve the ticket. Also, I would like to use this template to create an automation that automatically does the manual tasks a triager needs to do, as described at step 3 in https://compass.2i2c.org/projects/managed-hubs/support/#non-incident-response-process

GeorgianaElena:
Fixes #3477 

After merging this, the docs at https://compass.2i2c.org/projects/managed-hubs/support/#non-incident-response-process should be updated to instruct opening such in issue instead of a general one like it it the case right now.

Also, I want to investigate if it's possible to have a workflow that adds issues with this label to the appropriate board, with the appropriate metadata.


consideRatio (approved):
This looks great!

sgibson91:
> Also, I want to investigate if it's possible to have a workflow that adds issues with this label to the appropriate board, with the appropriate metadata.

You don't need to write a workflow. If the Project board is of the new style (not the classic boards), you can use an automation. Checkout my screen capture here: https://www.loom.com/share/d072c4ebff7846a78f38174cd3177b88?sid=c2103fe2-b98f-44d9-8ba8-9fe93e6378a0

sgibson91:
> I don't think it's possible to set project board metadata

The difficulty is that for the new project boards there's no REST API endpoint, only a GraphQL endpoint. And the GraphQL language is very counterintuitive. I was trying to learn it so that the End Date and Start Date fields in the Time Off project board could be automatically parsed from the title of the issue.

If folk have already done the hard work and written actions, that's great!

PR #3572, opened 2024-01-04
Include initContainers when calculating pod overhead

yuvipanda:
https://github.com/2i2c-org/infrastructure/pull/3569 changed the cryptnono daemonset to have different resource requests for the init containers as well as the container. While working on https://github.com/2i2c-org/infrastructure/pull/3566, I noticed this was generating wrong choices - the overhead was calculated wrong (too small).

We were intentionally ignoring init containers while calculating overhead, and turns out the scheduler and the autoscaler both do take it into consideration. The effective resource request for a pod is the higher of the resource requests for the containers *or* the init containers - this ensures that a pod with higher requests for init containers than containers (like our cryptnono pod!) will actually run. This is documented at
https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#resource-sharing-within-containers, and implemented in Kubernetes itself at
https://github.com/kubernetes/kubernetes/blob/9bd0ef5f173de3cc2d1d629a4aee499d53690aee/pkg/api/v1/resource/helpers.go#L50 (this is the library code that the cluster autoscaler uses).

This PR updates the two places we currently have that calculate effective resource requests (I assume eventually these will be merged into one - I haven't kept up with the team's work last quarter here).

I've updated the node-capacity-info.json file, which is what seems to be used by the generator script right now.

### DIFF
--- a/deployer/commands/generate/resource_allocation/daemonset_requests.py
+++ b/deployer/commands/generate/resource_allocation/daemonset_requests.py
@@ -64,23 +64,50 @@ def get_daemon_sets_requests():
     info = []
     for ds in daemon_sets:
         name = ds["metadata"]["name"]
-        req_mem = req_cpu = lim_mem = lim_cpu = 0
+        # From https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#resource-sharing-within-containers
+        # > - The highest of any particular resource request or limit defined on
+        # >   all init containers is the effective init request/limit. If any
+        # >   resource has no resource limit specified this is considered as the
+        # >   highest limit.
+        # > - The Pod's effective request/limit for a resource is the higher of:
+        # >  - the sum of all app containers request/limit for a resource
+        # >  - the effective init request/limit for a resource
+        #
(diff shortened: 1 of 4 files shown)


yuvipanda:
Put up the adjusted memory limits in https://github.com/2i2c-org/infrastructure/pull/3573 - I'm not sure why an update for r5.4xlarge was needed (seems unrelated to the fixes here, as the fix should be a no-op).

consideRatio:
@yuvipanda this looks great, I added a commit where I re-executed the generate command for all clusters based on a comment in the daemonset-requests.yaml file:

```yaml
# This file isn't updated by automation, but can easily be updated by manually
# running a command once for each cluster:
#
#     ls config/clusters | xargs -I {} deployer generate resource-allocation daemonset-requests {}
#
```

yuvipanda:
Ah, thank you @consideRatio! I'll do that next time I change any code that affects these generated.

consideRatio (approved):
Thank you for working this Yuvi!!

PR #3573, opened 2024-01-04
Adjust memory requests so largest instances spawn on r5.4xlarge

yuvipanda:
Brings in the new memory / cpu limits set up for
https://github.com/2i2c-org/infrastructure/pull/3572 - without that, the largest set up size doesn't actually spawn on r5.4xlarge due to insufficient memory.


yuvipanda:
I'm actually just going to wait for review, I think these have been broken for a little bit now, so no point rushing this fix.

sgibson91 (approved):

yuvipanda:
Thanks @sgibson91!

PR #3575, opened 2024-01-04
Re-generate deployer's credentials for nasa-esdis/veda/ghg and document this as self-mergable

consideRatio:
For some reason the esdis deployer creds had been deleted, which caused deploy failures observed by @yuvipanda recently - this fixes that and makes next update needed for any of these align.

I think it makes sense if we try to do these generations all at once to reduce the times we need to attend to this matter.

----
📚 Documentation preview 📚: https://2i2c-pilot-hubs--3575.org.readthedocs.build/en/3575/

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,7 @@
 repos:
   # Autoformat: markdown, yaml
   - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v4.0.0-alpha.8
+    rev: v3.1.0
     hooks:
       - id: prettier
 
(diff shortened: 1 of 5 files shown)


GeorgianaElena (approved):

consideRatio:
Thank you @GeorgianaElena!!

.pre-commit-config.yaml
```diff
@@ -11,7 +11,7 @@
 repos:
   # Autoformat: markdown, yaml
   - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v4.0.0-alpha.8
+    rev: v3.1.0
```
consideRatio, line 14: Oops this was not intentionally added, but let's not revert it because it was causing issues for me using the v4 version of prettier - the pre-commit hooks failed if committing no file processed by prettier.

PR #3577, opened 2024-01-05
Refine and detail k8s cluster upgrade policy

### LINKED ISSUE #412: Decide on a cluster upgrade policy
# Background

We already have 6 clusters across 3 cloud providers, and this number is only going to increase. Currently they are on different k8s versions based on the cloud provider, even though they are mostly deployed the same way. This will eventually bite us.

I propose we adopt the following rules:

1. We try to keep all our clusters to the same version that's the default on [new GKE clusters](https://cloud.google.com/kubernetes-engine/docs/release-notes). GKE is often the last to adopt a kubernetes master version, so this gives us time to adapt.
2. Do upgrades on a 2 month cycle. So every two months, we find the current GKE default version, and upgrade all our clusters to it.
3. Upgrades take place over the course of a week
4. We find some way to communicate this 'upgrade window' to our users. It should be as unobtrusive to them as possible. For centrally managed hubs (pilot-hubs and cloudbank right now), we need to just provide notice.
5. Do both master and node upgrades as part of the same window.

# Expected timeline

This issue is only about discussing and agreeing on a timeline, and not actually doing anything. So it's complete once we document it.

consideRatio:
Builds on work by @GeorgianaElena by detailing and tightening up an k8s upgrade policy. To tighten up the policy is possible now that we have managed to get all our clusters upgraded to 1.26+ (after next week, all are at 1.27+).

- closes #412

Documentation preview at https://2i2c-pilot-hubs--3577.org.readthedocs.build/howto/upgrade-cluster/, inlined below:

image

### DIFF
--- a/config/clusters/nasa-esdis/enc-deployer-credentials.secret.json
+++ b/config/clusters/nasa-esdis/enc-deployer-credentials.secret.json
@@ -1,23 +1,23 @@
 {
 	"AccessKey": {
-		"AccessKeyId": "ENC[AES256_GCM,data:kZTskz6SEOAG7Gs/XJWgb3/rBP8=,iv:0lt2hvWTvim6BQbOHRK1mEM4hOiWaYn1NMk+D6X50uE=,tag:h4kg7OhCjPSkP944bfwl9g==,type:str]",
-		"SecretAccessKey": "ENC[AES256_GCM,data:2reeLNQI/3AoYUvLOt2YMS6rfDif6FKkKy/P1jydTYCOp8WxLJQAWw==,iv:yCFDlAlUMhT++132DYsNj33o1GQeQzf//7812PUeGf0=,tag:chPGkunqefsBCJYktR/4fw==,type:str]",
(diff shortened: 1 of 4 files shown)


GeorgianaElena (approved):
Thank you @consideRatio

consideRatio:
@damianavila this policy declares that we should check in on k8s upgrades at the beginning of each quarter to ensure we schedule required actions to get upgrades done. Do you think this is acceptable also?

damianavila:
> @damianavila this policy declares that we should check in on k8s upgrades at the beginning of each quarter to ensure we schedule required actions to get upgrades done. Do you think this is acceptable and do you have ideas on how we ensure to check in on scheduling required upgrade actions at the beginning of the quarter?

That's a good question about recurring tasks on the maintenance load (I think another example of that is the recurrent key/secrets updates), let me think a little bit more about it and talk with @haroldcampbell to discuss how to make sure we follow the policy successfully.

PR #3578, opened 2024-01-05
terraform, azure and utoronto: fixes and terraform state sync

consideRatio:
- docs: fix notes about AKS machine types to use
- Add docstring to terraform/azure/proxycommand.py
- terraform, utoronto: reorganize for readability and add fixme notes
- terraform, utoronto: align files with current state
- terraform, azure: complete earlier cleanup by hardcoding NFS over SMB
- terraform, azure: stop declaring core node pool node count
- terraform, azure: fix capitalization
- terraform, azure: bump lower bound on some providers
- terraform, azure: cleanup unused azure-file k8s namespace

### DIFF
--- a/terraform/azure/projects/utoronto.tfvars
+++ b/terraform/azure/projects/utoronto.tfvars
@@ -18,8 +18,8 @@ core_node_vm_size = "Standard_E4s_v3"
 
 notebook_nodes = {
   "default" : {
-    min : 1,
-    max : 100,
+    min : 0,
+    max : 86,
     # FIXME: upgrade user nodes vm_size to Standard_E8s_v5
     vm_size : "Standard_E8s_v3",
   }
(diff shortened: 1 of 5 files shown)


terraform/azure/projects/utoronto.tfvars
```diff
@@ -1,23 +1,26 @@
...
-location          = "canadacentral"
+# FIXME: upgrade core_node_vm_size to Standard_E4s_v5
 core_node_vm_size = "Standard_E4s_v3"
+
 notebook_nodes = {
   "default" : {
-    min : 1,
-    max : 100,
+    min : 0,
+    max : 86,
```
yuvipanda, line 22: Is there a specific reason 86 was chosen?

yuvipanda (commented):
Generally looks good, one question about max number!

terraform/azure/projects/utoronto.tfvars
consideRatio: They align with the current state, but I don't know why they were set to that - it was done without using terraform.

yuvipanda: @consideRatio ah, ok. Can you just add a comment about that here? Once that's done, this is good to go. Thanks for cleaning this up!

yuvipanda (approved):

consideRatio:
Thank you for reviewing and answering questions about this @yuvipanda. This is now applied, and terraform plan now reports `No changes. Your infrastructure matches the configuration.` :)

yuvipanda:
Thanks for going through this in detail and getting all this set up before the upgrade, @consideRatio!

Issue #2729, opened 2023-06-27
Move R users of jupyter.utoronto.ca to r.datatools.utoronto.ca


yuvipanda:
We moved UToronto to have multiple hubs as part of https://github.com/2i2c-org/infrastructure/issues/1712, and set up r.datatools.utoronto.ca with a rocker based image. As part of https://github.com/2i2c-org/infrastructure/issues/2583, we want to simplify the images we maintain. We had already communicated to UToronto users that R is going to go away from jupyter.utoronto.ca, and we should use this opportunity to do that now.

Based on in-person conversations with @ntaback, the following schedule should work:

1. UToronto announces to all users (via their own mechanisms) that R will go away on **August 15**. This should be sent out by **first week of July** so they have time. Auguest 14 is last day of class for summer of Utoronto, and september 7 is first day of fall classes ([source](https://www.artsci.utoronto.ca/current/dates-deadlines/academic-dates#academic-dates-deadlines-accordion-3)). 
2. We merge https://github.com/2i2c-org/utoronto-image/pull/54, getting rid of R and simplifying everything else, **on August 15**.
3. If there are missing R packages, we sort them out over the next few weeks as requests come in. @ntaback should be able to corral them to us as needed.

yuvipanda:
https://2i2c.freshdesk.com/a/tickets/952 is the linked ticket. There is a plan outlined there on how to do this + timelines.

yuvipanda:
I tried to find a list of users who have ever used rstudio, and found that this way:

1. `deployer exec-homes-shell utoronto prod`. This spins up a pod with the utoronto homes be mounted under `/home`
2. Run this command to find all user home directories with a `.rstudio` directory created in them. I think this happens only if the user has *ever* used rstudio: `k -n prod exec utoronto-prod-shell  -- /bin/bash -c "cd /home; find . -name '.rstudio' -maxdepth 2 -mindepth 2" > rstudio-homedirs-utoronto`. This must be run in a different tab, while (1) is still running. It'll output to a file *locally* named `rstudio-homedirs-utoronto`.
3. The home directories are escaped, unescape them by first making this following script and calling it `doirstudio.py`:
    ```python
    import sys
    from escapism import unescape

    for l in sys.stdin:
        print(unescape(l.split('/')[1], '-'))
    ```
4. Run it with: `cat rstudio-homedirs-utoronto| python doirstudio.py > users-with-rstudio.txt`
[...]

PR #3583, opened 2024-01-06
[Don't merge until Jan 9]Redirect all UToronto home pages to unified homepage

yuvipanda:
UToronto will have a new URL (to be determined...) that they will manage, and will act as the 'unified' home page. In https://2i2c.freshdesk.com/a/tickets/952 they have asked us to redirect anyone landing just on the home page of the hubs to this new URL. I've setup the [utoronto
branch](https://github.com/2i2c-org/default-hub-homepage/tree/utoronto) of the home pages repo to redirect to the new URL.

This just deploys that change. The new URL hasn't been provided to us yet, so the utoronto branch in the default-hub-homepage repo will need to be updated once that URL is present.

### DIFF
--- a/config/clusters/utoronto/common.values.yaml
+++ b/config/clusters/utoronto/common.values.yaml
@@ -20,8 +20,11 @@ jupyterhub:
       enabled: true
   custom:
     homepage:
+      # Everyone hitting the home page directly is redirected to
+      # https://datatools.utoronto.ca as the unified home page. The various parameters
+      # here under templateVars are no-op - UToronto manages the unified homepage themselves.
+      gitRepoBranch: "utoronto"
       templateVars:
-        interface_selector: true
         org:
           name: University of Toronto
           logo_url: https://raw.githubusercontent.com/2i2c-org/default-hub-homepage/utoronto-prod/extra-assets/images/home-hero.png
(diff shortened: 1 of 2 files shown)


config/clusters/utoronto/common.values.yaml
```diff
@@ -35,26 +38,6 @@ jupyterhub:
         funded_by:
           name: University of Toronto
           url: https://www.utoronto.ca/
-        announcements:
-          - |
-            <div>
-              <h4>ARC will be making changes to our educational JupyterHub on January 9, 2023</h4>
```
consideRatio, line 41: "Will be making", but when they read this the changes are made - should this be revised?

config/clusters/utoronto/default-staging.values.yaml
```diff
@@ -4,12 +4,6 @@ jupyterhub:
     tls:
       - hosts: [staging.utoronto.2i2c.cloud]
         secretName: https-auto-tls
-  custom:
-    homepage:
-      gitRepoBranch: "utoronto-staging"
-  singleuser:
-    image:
-      tag: "14320bae73a0"
```
consideRatio, line 12: Also in the other PR, but probably fine?

config/clusters/utoronto/common.values.yaml
yuvipanda: @consideRatio the home page will just not be visible at all as soon as this change is made (due to https://github.com/2i2c-org/default-hub-homepage/commit/20ea7925734f00340af4eccfc41eaacdfe894df8) so this is fine.

config/clusters/utoronto/default-staging.values.yaml
yuvipanda: @consideRatio ah correct. But I've removed it from here anyway just for cleanliness.

Issue #3584, opened 2024-01-06
Clarify and document rounding for resource allocation list


yuvipanda:
The list of resource allocations that end users can choose from on the VEDA hub looks like this:

<img width="330" alt="image" src="https://github.com/2i2c-org/infrastructure/assets/30430/3c35b87c-915a-44e9-bf3f-69c928388709">

# Why are these numbers odd?

The reason these are a bit odd (1.9GB instead of 2GB, etc) is so we account for overhead of running various system and support components on each node, so not all the resources of the node are available for end users to use. The resources that *are* available, we split into chunks that fit together proportionally to minimize the amount of possible wasted space. https://github.com/2i2c-org/infrastructure/pull/3030 is the PR that originally added this, and has a lot of rationale in the description about how they are calculated.

# Why do we *show* these numbers?

Why do we **show** these odd numbers? It's because if I log in, and in the bottom of JupyterLab, I see the current *memory limit*:

<img width="231" alt="image" src="https://github.com/2i2c-org/infrastructure/assets/30430/b706b802-0595-47f8-bc6e-2e27b45096f8">

When an end user *hits* this memory limit, they start seeing their kernels die. Kernels dying due to memory exhaustion is perhaps the number one confusing thing that end users experience on a JupyterHub that they don't on their local machine. If you exhaust memory on your local machine, everything freezes. Here, the kernel dies. 

So when building the script to generate these choices (https://github.com/2i2c-org/infrastructure/pull/3030), I made the *display* of memory match the actual memory limit, so it is easy for the user to understand where this number was coming from. It does mean that the numbers look 'odd', but I personally think it's better to have *one* number be in the server startup screen and in jupyterlab and that look odd, than to have *two different numbers* here simply so that they look more rounded out.

However, everyone always asks why this number looks odd, so we should at least document this someplace nice.

yuvipanda:
For CPU though, I think we don't need to show 2 decimals after the ., 1 is probably fine. I'll make a PR for that.

PR #3585, opened 2024-01-06
Round CPU display numbers in resource allocation list

yuvipanda:
Brings it down to one digit after decimal, rather than 2.

Ref https://github.com/2i2c-org/infrastructure/issues/3584


consideRatio (approved):

yuvipanda:
Thanks @consideRatio

Issue #3588, opened 2024-01-08
Unpin dind version pin for dynamic image building & binderhub charts


yuvipanda:
When merging https://github.com/2i2c-org/infrastructure/pull/3585, I saw that it failed for binderhub-service (https://github.com/2i2c-org/infrastructure/actions/runs/7451299461/job/20273916986). Looking deeper, it seems to be due to https://github.com/docker-library/docker/issues/463.

I've temporarily pinned the dind version to the previous one (https://github.com/2i2c-org/infrastructure/pull/3589), which works.

# Action Items

- [x] Check if upstream (docker) has merged and deployed the change that fixes the issue we were facing
- [x] Test the change by removing the pin in our config
- [x] If the change works, remove the pin in our config (essentially reverting #3589)

yuvipanda:
I also documented this debugging session in https://youtube.com/live/lWp5BItlde4?feature=share

consideRatio:
This is unpinned in binderhub-service and the binderhub chart is removed - resolved!

PR #3589, opened 2024-01-08
Fix crashing docker daemon in binderhub-staging

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/3588


yuvipanda:
The process of debugging this was documented in https://youtube.com/live/lWp5BItlde4?feature=share

yuvipanda:
Thanks @sgibson91

PR #3595, opened 2024-01-09
utoronto: reduce prometheus requested memory to fit better on smaller nodes

consideRatio:
While we can't change the node type currently in utoronto's k8s clusters core node pool, this is a pre-requisite to using 2 CPU / 16 GB RAM nodes (Standard_E2s_v5).

Prometheus memory footprint looks like this the last 6 months, so on average ~6 Gi used, and always below 10Gi, so it should be robust to declare a maximum of 12Gi.
image


yuvipanda (approved):

consideRatio:
Thank you @yuvipanda!

Issue #3592, opened 2024-01-09
AKS / UToronto: calico-typha scaled to three replicas, forcing three core nodes


consideRatio:
This is similar to https://github.com/2i2c-org/infrastructure/issues/2490 where the same problem is considered for GKE clusters. In the AKS clusters though, calico is deployed in the "calico-system" namespace via the "tigera-operator" in the namespace named the same thing.

I've not yet figured out how to re-configure it to not require 3 replicas, but it would be good to reduce that number or allow at least two out of three pods to co-locate on a node to avoid forcing use of 3 nodes.

consideRatio:
### Findings

- calico-typha will fail to schedule next to another calico-typha pod due to use of node ports (when the pod is to schedule `# node(s) didn't have free ports for the requested pod ports`).
- aks installs calico by using the tigera-operator, and the installation of calico is controller by an [`Installation` resource-type](https://docs.tigera.io/calico/latest/reference/installation/api#operator.tigera.io/v1.Installation) named `default`
- while `Installation` resources allows typha config under `typhaDeployment`, we can't declare the number of replicas directly or influence how the number of replicas may be calculated behind the scenes by the tigera-operator
- the developers of tigera-operator describes in https://github.com/projectcalico/calico/issues/6100#issuecomment-1129178462 that the replica count for calico-typha is calculated based on cluster size
- the calico-typha autoscaling logic is declared [here to reference](https://github.com/tigera/operator/blob/ff575488cca3fed1113219b5cd8f77f594c23eed/pkg/controller/installation/typha_autoscaler.go#L180-L201) a [calculated amount of replicas](https://github.com/tigera/operator/blob/ff575488cca3fed1113219b5cd8f77f594c23eed/pkg/common/autoscale.go#L17-L56) based on the nodes in the cluster.

### A way forward

1. there is no great solution, only upstream feature request and mitigation is reasonable at the moment
2. utoronto should use 2 CPU / 16 GB nodes instead of the 4 CPU / 32 GB nodes to get three cheaper replicas
   - I'll add a fixme note in the utoronto cluster config, allowing that to be tweaked when we can during cluster re-creation
3. we should suggest that the tigera-operator start supporting configuration where the autoscaling ladder is configurable and not hardcoded
   - I'll submit a issue

PR #3596, opened 2024-01-09
terraform, azure and utoronto: an upgrade, misc to support it, and misc opportunistic details

consideRatio:
The terraform changes are applied already. I've done some things besides just upgrading the k8s cluster version that I believed were resonable to go for opportunistically.

### Changes overviewed

- a user node pool switch from `Standard_E8s_v3` to `_v5`, which is like `r3.2xlarge` to `r5.2xlarge` in AWS. This price of these are the same.
- a user node pool switch and a core node pool a prepared transition from use of persistent machine disks on user nodes to ephemeral machine disks (like in GKE also for example)
- (reverted, autoscaling and 3 replicas remains) a pinning of core node pool count to 2 in order to avoid paying for an empty core node pool due to #3592
- added resolution of configuration in order to
  - enable upgrade of control plan and node pools separately
  - enable a non-disruptive transitions from misc machine types
- misc fixme notes and comments with notable research behind them
- removal of no longer relevant label (`hub.jupyter.org/node-size`)
- an AKS transition towards referencing "notebook nodes" as "user nodes"
- (added post-review) refactored use of multiple node pool variables to a single variable to reduce duplication of node pool variable definitions

### DIFF
--- a/terraform/azure/projects/utoronto.tfvars
+++ b/terraform/azure/projects/utoronto.tfvars
@@ -42,25 +42,9 @@ core_node_pool = {
 }
 
 user_node_pools = {
-  "default" : {
-    name : "nbdefault",
-    # NOTE: min-max below was set to 0-86 retroactively to align with
-    #       observed state without understanding on why 0-86 was picked.
-    min : 0,
-    max : 86,
-    # FIXME: upgrade user nodes vm_size to Standard_E8s_v5
-    vm_size : "Standard_E8s_v3",
-    # FIXME: remove this label
-    labels : {
-      "hub.jupyter.org/node-size" = "Standard_E8s_v3",
-    },
-    kubernetes_version : "1.26.3",
-    # FIXME: stop using persistent disks for the nodes, use Temporary instead
-    kubelet_disk_type : "OS",
-  },
   "usere8sv5" : {
     min : 0,
     max : 100,
     vm_size : "Standard_E8s_v5",
-  }
+  },
 }
--- a/terraform/azure/projects/utoronto.tfvars
+++ b/terraform/azure/projects/utoronto.tfvars
@@ -20,10 +20,6 @@ core_node_pool = {
   #        reasons like three calico-typha pods. See
   #        https://github.com/2i2c-org/infrastructure/issues/3592#issuecomment-1883269632.
   #
-  #        Transitioning to E2s_v5 would require reducing the requested memory
-  #        by prometheus-server though, but that should be okay since
-  #        prometheus has reduced its memory profile significant enough recently.
-  #
   vm_size : "Standard_E4s_v3",
 
   # FIXME: stop using persistent disks for the nodes, use the variable default
--- a/terraform/azure/projects/utoronto.tfvars
+++ b/terraform/azure/projects/utoronto.tfvars
@@ -1,3 +1,11 @@
+# IMPORTANT: Due to a restrictive network rule from storage.tf, we can't perform
+#            "terraform plan" or "terraform apply" without a workaround.
+#
+#            One known workaround is to allow your public IP temporarily as
+#            discussed in https://github.com/2i2c-org/infrastructure/issues/890#issuecomment-1879072422.
+#            This workaround is problematic as that may temporarily allow access
+#            to storage by other actors with the same IP.
+#
 tenant_id                      = "78aac226-2f03-4b4d-9037-b46d56c55210"
 subscription_id                = "ead3521a-d994-4a44-a68d-b16e35642d5b"
 resourcegroup_name             = "2i2c-utoronto-cluster"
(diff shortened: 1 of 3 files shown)


terraform/azure/projects/utoronto.tfvars
```diff
@@ -8,21 +8,39 @@ location                       = "canadacentral"
...
+  kubelet_disk_type : "OS",
+
+  # FIXME: use a larger os_disk_size_gb than 40, like the default of 100, to
+  #        avoid running low when few replicas are used
+  os_disk_size_gb : 40,
+
+  # FIXME: its nice to use autoscaling, but we end up with three replicas due to
+  #        https://github.com/2i2c-org/infrastructure/issues/3592#issuecomment-1883269632
+  #        and its a waste at least using Standard_E4s_v3 machines.
+  enable_auto_scaling : false,
```
yuvipanda, line 36: I'm concerned that we don't have the capacity to fully deal with the fact that this would be the *only* nodepool in our entire fleet that doesn't have autoscaling turned on, and all the changes that entails. For example, I think this would mean that cordoning and draining the node won't automatically bring up another node, so I'm not exactly sure how we would deal with nods failing (that do happen). Given other constraints on our engineering capacity, I think we should avoid special casing unless we absolutely must.

So my suggestion is to make the cost trade-off here, and not try to optimize away the third node. Even if we want to optimize cost for UToronto, I think there are other avenues to tackle (like their huge home directory for example). Let's turn autoscaling back on, and we can deal with the extra node once upstream fixes their configurability issues.

terraform/azure/variables.tf
```diff
@@ -92,27 +78,47 @@ variable "ssh_pub_key" {
   EOT
 }
 
-variable "notebook_nodes" {
+variable "core_node_pool" {
+  type = object({
+    name : optional(string, ""),
```
yuvipanda, line 83: Since `""` is not a valid name, I think this should be marked as required rather than optional (same as `vm_size`)

terraform/azure/variables.tf
```diff
@@ -92,27 +78,47 @@ variable "ssh_pub_key" {
...
+    os_disk_size_gb : optional(number, 100),
+    kubernetes_version : optional(string, ""),
+    kubelet_disk_type : optional(string, "Temporary"),
+  })
+  description = "Core node pool"
+}
+
+variable "user_node_pools" {
   type = map(object({
+    name : optional(string, ""),
```
yuvipanda, line 100: Same comment as `name` for the core node..

yuvipanda (changes requested):

terraform/azure/projects/utoronto.tfvars
consideRatio: > I'm concerned that we don't have the capacity to fully deal with the fact that this would be the only nodepool in our entire fleet that doesn't have autoscaling turned on, and all the changes that entails.

I think you are right, this is just too strong of a point!

consideRatio:
@yuvipanda I went back to autoscaling the core node pool and applied the change, I also refactored use of multiple node pool _variables_ to a single variable in order to go from having three definitions of a node pool to a single.

yuvipanda (changes requested):
Thanks a lot @consideRatio! This *mostly* looks good to me - however, I'd like us to do one of two things:

1. Have separate `core_node_pool`, `notebook_nodepools` and `dask_nodepool`, to match what we have on GCP
2. Encode via type definition somehow that the *only* supported keys for the `node_pools` map are "core", "user" and "dask"

Currently, (2) is encoded in the code - if we use any other key, it will just silently fail. 

I think (2) is preferable (here and in GCP too, although that's out of scope) but if that's not possible I'd like us to do (1) than leave it implicit.

I appreciate your cleanup efforts, @consideRatio!

consideRatio:
> if we use any other key, it will just silently fail

I think I can come up with a way to make this fail loudly, but not via the type definition. I tried to do this myself but failed to find a way to do it without breaking the DRY principle and verbosly duplicating the type definition for any individual node pool. I don't want to break DRY in this case, I find it gets too messy and hard to maintain.

I'm aiming to meet your request for changes by failing loudly if:
- node_pools variable includes keys besides `core`, `user`, and `dask`
- node_pools value for `core` doesn't include exactly one node pool (since one and only one node pool maps to `default_node_pool`)

consideRatio:
Ahh googling for "terraform assertions" I ended up finding that there can be a "custom validation" block for variables like documented in https://developer.hashicorp.com/terraform/language/values/variables#custom-validation-rules - I'll go for that.

consideRatio:
Thank you for reviewing @yuvipanda !!

yuvipanda:
@consideRatio thank you for patiently doing a couple rounds with me async on this!

PR #3598, opened 2024-01-09
utoronto: Don't force usage of notebookapp

yuvipanda:
https://github.com/2i2c-org/infrastructure/pull/3582 has been merged, so we can stop forcing NotebookApp in prod.

Without this, JupyterLab actually fails to launch, fixed upstream in https://github.com/jupyter-server/jupyter_server/pull/1384


yuvipanda:
self-merging to restore JupyterLab on the prod utoronto hub

PR #3599, opened 2024-01-09
Revert "[Merge Jan 9] utoronto: Enable automatic login on utoronto hubs"

yuvipanda:
Reverts 2i2c-org/infrastructure#3531

More detail in https://github.com/2i2c-org/infrastructure/issues/2729#issuecomment-1883571591.

We don't want to *unconditionally* autologin. We want to autologin when nbgitpuller is being
used, and to redirect to datatools.utoronto.ca any other time. This is now handled by the JS
in https://github.com/2i2c-org/default-hub-homepage/pull/28


yuvipanda:
Self-merging to make sure the redirects are in place at least on the same day as originally planned.

PR #3600, opened 2024-01-09
Use same home page branch for utoronto prod hub too

yuvipanda:
Everything should use the `utoronto` branch now, as set in the common config file


yuvipanda:
Self merging so users don't see the homepage on jupyter.utoronto.ca

PR #3606, opened 2024-01-10
Revert "Move the " UToronto JupyterHub Private Cloner"  GitHub app's details under the default hubs common"

yuvipanda:
Reverts 2i2c-org/infrastructure#3604


yuvipanda:
Maybe https://github.com/2i2c-org/utoronto-r-image/pull/17 was *all* that was needed?

Issue #3603, opened 2024-01-10
[Support] nbgitpuller link breaks after update - University of Toronto


GeorgianaElena:
### The Freshdesk ticket link

https://2i2c.freshdesk.com/a/tickets/1229

### Ticket request type

Something is not working

### Ticket impact

🟥 Critical

### Short ticket description

Nbgitpuller links not working.

When the hub is:
- pointing to[ https://jupyter.utoronto.ca](https://jupyter.utoronto.ca/) or[ https://datatools.utoronto.ca](https://datatools.utoronto.ca/) it gives 404 or page not found
- pointing to[ https://r.datatools.utoronto.ca](https://r.datatools.utoronto.ca/) it thows  the following (see screenshot below)
![utoronto-nbgitpuller-error](https://github.com/2i2c-org/infrastructure/assets/7579677/7d863221-db5f-4e7d-9736-558158f9b8a5)

### (Optional) Investigation results

No investigation happened yet as I wanted to first open the issue for visibility.

GeorgianaElena:
nbgitpuller works with https://r.datatools.utoronto.ca/ if the repo is public.

However the repo in the ticket is private and it was reported that it worked on Monday, I believe before the migration?

GeorgianaElena:
I am confused of the fact that I don't see evidence that private GitHub pulling was setup for the toronto hub https://infrastructure.2i2c.org/howto/features/private-nbgitpuller/#allow-nbgitpuller-to-pull-from-private-github-repos

GeorgianaElena:
Ahh, the GitHub app is not under the 2i2c org, this is why I couldn't find it. `git-credentials-helper` is setup at https://github.com/2i2c-org/infrastructure/blob/b6846fb9814bc613b8f442fc00f5faf2699ebe28/config/clusters/utoronto/common.values.yaml#L59-L65

GeorgianaElena:
`git-credentials-helper` is available in the original utoronto hub image https://github.com/2i2c-org/utoronto-image/blob/14320bae73a0570aa1ab1b19c68758894b291979/environment.yml#L42 but not in the new r-image.

https://github.com/2i2c-org/utoronto-r-image/pull/17 should make it available

GeorgianaElena:
[...]

PR #3608, opened 2024-01-10
Document we may share GitHub apps for private nbgitpuller use

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/3603

----
📚 Documentation preview 📚: https://2i2c-pilot-hubs--3608.org.readthedocs.build/en/3608/


GeorgianaElena (approved):
Yay! 🚀

PR #3611, opened 2024-01-12
Update hhmi user image tag and name

GeorgianaElena:
Ref: https://2i2c.freshdesk.com/a/tickets/1233


GeorgianaElena:
I believe the validate workflow failure is related to https://github.com/cryptnono/cryptnono. So I'm investigating that now

yuvipanda (approved):
This can be self merged without review too

PR #3616, opened 2024-01-16
dask-gateway: bump from 2023.9.0 to 2024.1.0

consideRatio:
This makes `latest` tag not come with `IfNotPreset` pull policies, which easily could cause issues if a `latest` tag was used. That in turn is something I want before suggesting use of `latest` tag in https://2i2c.freshdesk.com/a/tickets/794 as a compromise.


sgibson91 (approved):

consideRatio:
Thank you @sgibson91!!

PR #3626, opened 2024-01-19
[showcase] Enable env var access to scratch bucket

sgibson91:
requested in https://2i2c.freshdesk.com/a/tickets/1255

A scratch bucket already existed, but the env var wasn't set up to link it to the hub


sgibson91:
I deployed this already to make sure it works

GeorgianaElena (approved):

Issue #3306, opened 2023-10-24
[Request deployment] New Hub: Uganda Virus Research Institute/Makerere University


jmunroe:
### The GitHub handle of the community representative

pending

### Hub important dates

Waiting on additional information from the community.

### Hub Authentication Type

pending

### First Hub Administrators

- Edward Lukyamuzi, elukyamuzi2010@gmail.com
- Alfred Ssekagiri, assekagiri@gmail.com

### [GitHub Auth only] How would you like to manage your users?

None

### Hub logo image URL

pending

### Hub logo website URL

pending

### Hub user image GitHub repository

pending

### Hub user image tag and name

pending

### Extra features you would like to enable

- [ ] Dedicated Kubernetes cluster
- [ ] Scalable Dask Cluster

### (Optional) Preferred cloud provider

None

### (Optional) Billing and Cloud account

None

### Other relevant information to the features above

This hub will be deployed on the af.catalystproject.2i2c.cloud cluster.  I am currently waiting to get additional information from the community champions before this deployment can be scheduled.

### Tasks to deploy the hub

- [x] 1. Deploy information filled in above
- [x] 2. Engineer who will deploy the hub is assigned
- [x] 3. If using GitHub Orgs/Teams Auth, Engineer is given Owner rights to the org to set this up.
- [x] 4. Initial Hub deployment PR <link to PR>
- [ ] 5. Administrators able to log on -> Hub now in steady-state

damianavila:
@jmunroe, do you think it is possible to have some preliminary information for next week?
I would like @AIDEA775 (Alejandro) to try deploying a new hub in a pre-existing cluster and this one would be a good target if the complexity is not major.

jmunroe:
What about #3305 since that's ready to go?

jmunroe:
I am not sure why this community has not been responding to us over email.  In the meantime, I propose we can move it forward by taking the same approach I outlined in https://github.com/2i2c-org/infrastructure/issues/3515:

- Have both a R and Python image available
- Only have a single hardware profile (4GB / 1CPU on shared small 32GB/4CPU machine)
- Use GitHub teams for authentication with the the hub champions invited to join a GitHub org called CatalystProject-hubs
  - Edward Lukyamuzi, [elukyamuzi2010@gmail.com](mailto:elukyamuzi2010@gmail.com)
  - Alfred Ssekagiri, [assekagiri@gmail.com](mailto:assekagiri@gmail.com)
[...]

Issue #3349, opened 2023-10-31
Come up with when the jupyterhub-configurator is to be disabled


consideRatio:
Here is at least one situation when it could make sense to opt-out of jupyterhub-configurator, which is technically enabled by #3348.

- When profileList overrides make the configurator choices pointless (image overrides, default_url overrides)
  https://github.com/2i2c-org/infrastructure/pull/3341#discussion_r1377641920

PR #3629, opened 2024-01-19
Add UVRI hub on catalystproject-africa cluster

AIDEA775:
ref: #3306 
url: https://uvri.af.catalystproject.2i2c.cloud

Note: the admins list is empty because we are waiting for them to accept the invitation to the GitHub organization.

Edit: @eddUG has accepted the invitation and is now an admin.

### DIFF
--- a/config/clusters/catalystproject-africa/uvri.values.yaml
+++ b/config/clusters/catalystproject-africa/uvri.values.yaml
@@ -1,0 +1,70 @@
+jupyterhub:
+  ingress:
+    hosts: [uvri.af.catalystproject.2i2c.cloud]
+    tls:
+      - hosts: [uvri.af.catalystproject.2i2c.cloud]
+        secretName: https-auto-tls
+  custom:
+    2i2c:
+      add_staff_user_ids_to_admin_users: true
+      add_staff_user_ids_of_type: "github"
+    homepage:
+      templateVars:
+        org:
+          name: Catalyst Project, Africa - UVRI
+          url: https://2i2c.org
+          logo_url: https://2i2c.org/media/logo.png
+        designed_by:
+          name: "2i2c"
+          url: https://2i2c.org
+        operated_by:
+          name: "2i2c"
+          url: https://2i2c.org
+        funded_by:
+          name: Chan Zuckerberg Initiative - Open Science
+          url: "https://chanzuckerberg.com/science/programs-resources/open-science/"
+  hub:
+    config:
+      JupyterHub:
+        authenticator_class: github
+      GitHubOAuthenticator:
+        oauth_callback_url: https://uvri.af.catalystproject.2i2c.cloud/hub/oauth_callback
+        allowed_organizations:
+          - CatalystProject-Hubs:uvri
+        scope:
+          - read:org
+      # Authenticator:
+      #   admin_users:
+      #     -
+  singleuser:
+    profileList:
+      - display_name: Jupyter
+        slug: jupyter-scipy
+        description: "Python environment"
+        default: true
+        kubespawner_override:
+          image: jupyter/scipy-notebook:2023-06-27
+          default_url: /lab
+        profile_options: &profile_options
+          resource_allocation:
+            display_name: Resource Allocation
+            choices:
+              mem_8:
+                default: true
+                display_name: ~8 GB RAM, ~1.0 CPU
+                kubespawner_override:
+                  mem_guarantee: 6.684G
+                  cpu_guarantee: 0.75
+                  mem_limit: null
+                  cpu_limit: null
+                  node_selector:
+                    node.kubernetes.io/instance-type: r5.xlarge
+      - display_name: RStudio
+        description: R environment
+        kubespawner_override:
+          image: rocker/binder:4.3
+          default_url: /rstudio
+          # Ensures container working dir is homedir
+          # https://github.com/2i2c-org/infrastructure/issues/2559
+          working_dir: /home/rstudio
+        profile_options: *profile_options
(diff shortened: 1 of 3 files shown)


config/clusters/catalystproject-africa/uvri.values.yaml
```diff
@@ -0,0 +1,70 @@
...
+        slug: jupyter-scipy
+        description: "Python environment"
+        default: true
+        kubespawner_override:
+          image: jupyter/scipy-notebook:2023-06-27
+          default_url: /lab
+        profile_options: &profile_options
+          resource_allocation:
+            display_name: Resource Allocation
+            choices:
```
sgibson91, line 51: I think using choices here is overkill if you are only providing one option, you could probably just merge the `kubespawner_override` section below here with the one above that sets the image (same in the R profile as well)

sgibson91 (approved):
One comment, but otherwise LGTM

config/clusters/catalystproject-africa/uvri.values.yaml
GeorgianaElena: I agree with @sgibson91 about not having `profile_options` in this case. 

Also, @AIDEA775, if you want to get more insight into configuring profile lists, you can checkout https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html#kubespawner.KubeSpawner.profile_list.

config/clusters/catalystproject-africa/uvri.values.yaml
```diff
@@ -0,0 +1,70 @@
...
+              mem_8:
+                default: true
+                display_name: ~8 GB RAM, ~1.0 CPU
+                kubespawner_override:
+                  mem_guarantee: 6.684G
+                  cpu_guarantee: 0.75
```
GeorgianaElena, line 57: I'm a bit confused about where these numbers come from. In the issue, it looks like the request was for 4GB / 1CPU.

I suspect these numbers were taken from a hub that had a more extensive list of resource options and are the closest to the resources requested by the community (4GB 1CPU)? Which is a fair assumption, I think, given we don't have docs about the resource allocation script nor guidelines about how to choose these :(

I am not 100% confident to recommend a change here, but would love for @consideRatio's input!  In this particular case, would one use the resource allocation script to generate the numbers or should the exactly requested numbers be used?

config/clusters/catalystproject-africa/uvri.values.yaml
consideRatio: I'd advocate that we provide this with multiple options, defaulting to a lower request in https://github.com/2i2c-org/infrastructure/issues/3631#issuecomment-1905365573.

Ideally we'd have lots of guidance about this and discussed requests/limits already, but we haven't yet found time to conclude such guidance.

This doesn't have to be a PR blocker, but I think there is UX value to provide a single option already to not introduce it later as a change.

config/clusters/catalystproject-africa/uvri.values.yaml
```diff
@@ -0,0 +1,70 @@
...
+    hosts: [uvri.af.catalystproject.2i2c.cloud]
+    tls:
+      - hosts: [uvri.af.catalystproject.2i2c.cloud]
+        secretName: https-auto-tls
+  custom:
+    2i2c:
+      add_staff_user_ids_to_admin_users: true
+      add_staff_user_ids_of_type: "github"
+    homepage:
+      templateVars:
```
GeorgianaElena, line 12: This is lacking docs :( but let's also disable the configurator, per https://github.com/2i2c-org/infrastructure/issues/3349.
```suggestion
    jupyterhubConfigurator:
      enabled: false
    homepage:
      templateVars:
```
(I'll search for a place in the docs to add thi  so that is a bit more findable)

GeorgianaElena (approved):
This looks good to me @AIDEA775, thank you 🚀 ✨  I've left a few comments to restart some conversations about resource allocations and also about deactivating the configurator, but I don't think any of those are blockers.

config/clusters/catalystproject-africa/uvri.values.yaml
consideRatio: > In this particular case, would one use the resource allocation script to generate the numbers or should the exactly requested numbers be used?

I'd say no, but I don't know the state of things - its providing requests==limits right? That isn't suitable here I'd say, but of course it can be used to provide something to tweak etc...

Yikes, this is as always a big topic for each single hub.

My proposal is to use this formula and provide options requesting 0.5, 1, 2, and 4 GB of memory, representing ~1/64, ~1/32, ~1/16, and ~1/8 of the node.

- mem request 0.5G
- mem limit 1G (twice the requested memory)
- cpu request 3.6 / 64 (~share of _allocatable_ CPU)
- cpu limit 4 / 64 * 8 (something like eight times the requested CPU excluding the ~400m headroom, but at least 1CPU)

Something like....

```
{
   "mem_guarantee": 0.5G,
   "mem_limit": 1.0G,
   "cpu_guarantee": 0.05625,
   "cpu_limit": 1,
}
{
   "mem_guarantee": 1G,
   "mem_limit": 2G,
   "cpu_guarantee": 0.1125,
   "cpu_limit": 1,
}
{
   "mem_guarantee": 2G,
   "mem_limit": 4G,
   "cpu_guarantee": 0.225,
   "cpu_limit": 2,
}
{
   "mem_guarantee": 4G,
   "mem_limit": 8G,
   "cpu_guarantee": 0.45,
   "cpu_limit": 4,
}
```

Generated from just-in-time writeup:

```python
def scale(n):
    print({
        "mem_guarantee": f"{n}G",
        "mem_limit": f"{n * 2}G",
        "cpu_guarantee": 1/(32/n) * (4-0.4),
        "cpu_limit": max(1, 1/(32/n) * 4 * 16),
    })

scale(0.5)
scale(1)
scale(2)
scale(4)
```

consideRatio (approved):

AIDEA775:
I will merge this, and the discussion of the profileList will continue in #3631. Thank you all for your reviews and comments!

PR #3630, opened 2024-01-22
[victor] enable unlisted choice on staging for testing purposes

sgibson91:
ref: https://2i2c.freshdesk.com/a/tickets/1257

I copied the original profiles from the common values file because helm doesn't append lists when combining values files, instead overwriting the list


sgibson91:
I'm gonna go ahead an self-merge here and I'm pretty confident this works

Issue #3048, opened 2023-08-29
Procedure for emailing all community representatives


consideRatio:
There are situations where we may want to email all community representatives to inform them of an upcoming change or something else. But how do we do this practically? Let's assume there is agreement that we want to communicate something to all community representatives via email for now, then how do we go about it?

### Draft idea on procedure to email all community reps

1. An email text is drafted via a google document
2. The email text is reviewed by partnership, and engineering if relevant
3. The email is sent - but how?
   - FROM: support@2i2c.freshdesk.com
   - TO: support@2i2c.freshdesk.com
   - BCC: all the individual community representatives emails from a list somewhere

The outcomes of this idea is:
- community reps emails remain secret from each other
- anyone pressing reply-all can't cause a mess, being sent FROM and TO from the same email, only that email gets a response
- anyone replying (to support) gets an individual ticket created for themselves, which is probably good

### Missing information

If this procedure is agreed on, it still is missing details about:
- [ ] where we put the google document in our shared drive
- [ ] where we have a list of all the community reps emails

damianavila:
cc @2i2c-org/partnerships-and-community-guidance, can you please provide feedback on the proposed idea?

yuvipanda:
@consideRatio would you consider https://github.com/2i2c-org/infrastructure/pull/3042 blocked on this issue? I'm not sure who is responsible for currently pushing this forward. I don't think it should be you, as I think you're already pretty overloaded.

damianavila:
@jmunroe and @colliand, can you please provide feedback on the proposed idea?
Additionally, we will need your intervention/participation as well to unblock this one.

colliand:
Yes. I agree with this idea. 2i2c should manage a list of hub champions. The Google/bcc approach will likely work but requires 2i2c personnel to manually manage the list. With GDPR and no-spam rules, there will be risks with the manual management of the list. I suggest we use a list serve or another system (MailChimp, Campaign Monitor) for managing transactional emails. This kind of resource can also be used for other kinds of messages (quarterly updates, marketing). 

What is the priority on this? Echoing Yuvi's message above, is the absence of a system to announce breaking changes blocking something?

damianavila:
> What is the priority on this? Echoing Yuvi's message above, is the absence of a system to announce breaking changes blocking something?

Yep, we currently have this one blocked: https://github.com/2i2c-org/infrastructure/pull/3042#issuecomment-1696633530

consideRatio:
[...]

PR #3634, opened 2024-01-24
Check for technical support contacts as part of new hub turn-up

yuvipanda:
As part of https://github.com/2i2c-org/infrastructure/issues/3048#issuecomment-1907195217, we now have an airtable be the 'source of truth' for who is allowed to initiate support requests. This PR adds an item in the checklist for new hub turn up, making sure we validate that there *is* a technical contact for this hub.

The airtable in use may change soon as well as the terminology (as partnerships works on it https://github.com/2i2c-org/infrastructure/issues/3048#issuecomment-1909011115). But it is important to have a verification step here, so we don't let whatever airtable it is be out of date.

Ref https://github.com/2i2c-org/infrastructure/issues/3048#issuecomment-1907335901

### DIFF
--- a/.github/ISSUE_TEMPLATE/2_new-hub-provide-info.yml
+++ b/.github/ISSUE_TEMPLATE/2_new-hub-provide-info.yml
@@ -23,6 +23,20 @@ body:
     validations:
       required: true
 
+  - type: checkboxes
+    attributes:
+      label: Technical support contacts listed
+      description: |
+        Technical contacts are the folks who are authorized to open support tickets about this hub.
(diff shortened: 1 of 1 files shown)


sgibson91 (commented):
I might add this further up to the community information section? I feel like it's getting lost amongst technical detail (such as what image are they using, what authentication method, etc). And since it's Partnerships responsibility to act on updating the airtable (even if an engineer confirms that the information is missing), then it will probably benefit from being higher up in the list of tasks that need checking.

yuvipanda:
@sgibson91 great suggestion! I moved it up - what do you think?

sgibson91 (approved):
Looks great!

damianavila (approved):

Issue #3635, opened 2024-01-24
Adjust our schema to turn jupyterConfigurator off if profileList is not empty


yuvipanda:
(lifted from https://github.com/2i2c-org/infrastructure/pull/3625#pullrequestreview-1841482496)

In our [schema](https://github.com/2i2c-org/infrastructure/blob/285faeb899b7e10623e9464269d90497d8a97a13/helm-charts/basehub/values.schema.yaml#L496), I believe we can set it up so that jupyterConfigurator can be `true` *only* if `singleuser.profileList` is empty or unset (via one of the methods described in https://stackoverflow.com/a/38781027). This should ensure that we can't have both profile lists and the configurator enabled at the same time.

PR #3641, opened 2024-01-25
Make sure the configurator is disabled when appropriate

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/3635

> In our [schema](https://github.com/2i2c-org/infrastructure/blob/285faeb899b7e10623e9464269d90497d8a97a13/helm-charts/basehub/values.schema.yaml#L496), I believe we can set it up so that jupyterConfigurator can be true only if singleuser.profileList is empty or unset (via one of the methods described in https://stackoverflow.com/a/38781027). This should ensure that we can't have both profile lists and the configurator enabled at the same time.

Unfortunately I couldn't implement @yuvipanda's suggestion above because we do not do any validation for the jupyterhub config and we rely on it's own schema. So instead I wrote this function which is also how we validate some of the authentication configuration. 

Does this make sense or should I continue to try and fiddle around with the actual schema?

### DIFF
--- a/deployer/commands/validate/config.py
+++ b/deployer/commands/validate/config.py
@@ -242,8 +242,8 @@ def configurator_config(
 ):
     """
     For each hub of a specific cluster:
-     - It asserts that when the JupyterHub GitHubOAuthenticator is used,
-       then `Authenticator.allowed_users` is not set.
+     - It asserts that when the singleuser configuration overrides the same fields like the configurator,
+       the later must be disabled.
        An error is raised otherwise.
     """
     _prepare_helm_charts_dependencies_and_schemas()
@@ -254,7 +254,7 @@ def configurator_config(
 
     for i, hub in enumerate(hubs):
         print_colour(
-            f"{i+1} / {len(hubs)}: Validating authenticator config for {hub.spec['name']}..."
+            f"{i+1} / {len(hubs)}: Validating configurator and profile lists config for {hub.spec['name']}..."
         )
 
         configurator_enabled = False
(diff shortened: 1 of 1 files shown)


deployer/commands/validate/config.py
```diff
@@ -236,6 +235,73 @@ def authenticator_config(
...
 
+@validate_app.command()
+def configurator_config(
+    cluster_name: str = typer.Argument(..., help="Name of cluster to operate on"),
+    hub_name: str = typer.Argument(None, help="Name of hub to operate on"),
+):
+    """
+    For each hub of a specific cluster:
+     - It asserts that when the singleuser configuration overrides the same fields like the configurator,
+       the later must be disabled.
```
sgibson91, line 246:
```suggestion
       the latter must be disabled.
```

deployer/commands/validate/config.py
```diff
@@ -236,6 +235,73 @@ def authenticator_config(
...
 
 
+@validate_app.command()
+def configurator_config(
+    cluster_name: str = typer.Argument(..., help="Name of cluster to operate on"),
+    hub_name: str = typer.Argument(None, help="Name of hub to operate on"),
+):
+    """
+    For each hub of a specific cluster:
+     - It asserts that when the singleuser configuration overrides the same fields like the configurator,
```
sgibson91, line 245:
```suggestion
     - It asserts that when the singleuser configuration overrides the same fields like the configurator, specifically kubespawner_override and profile_options,
```

I think we could make this a little bit more explicit, especially if someone reading this doesn't know what the configurator does

sgibson91 (approved):
I'm very +1 on not trying to spend too much time trying to hack together something in an upstream schema for a 2i2c-specific feature. So long as an error is thrown _at some point_ before deployment, this feels good enough to me.

deployer/commands/validate/config.py
GeorgianaElena: Yes, very good point! Thank you! ✨

PR #3618, opened 2024-01-16
Enable auth0 authentication for earthscope

yuvipanda:
This adds documentation for receiving credentials and
setting up auth0 authentication: https://github.com/2i2c-org/infrastructure/issues/3533

https://github.com/2i2c-org/infrastructure/issues/3636 tracks adding documentation on how
to pass credentials from `auth_state` to spawned user servers.

https://github.com/2i2c-org/infrastructure/issues/3637 tracks getting the PRs to OAuthenticator
merged upstream and bringing them back into the fold here.

----
📚 Documentation preview 📚: https://2i2c-pilot-hubs--3618.org.readthedocs.build/en/3618/

PR #3642, opened 2024-01-25
Handle deployment-service-check user for auth0

yuvipanda:
That user (which we use for our health checks) does not have any auth state, so this was causing health checks to fail.

Follow-up to https://github.com/2i2c-org/infrastructure/pull/3618


yuvipanda:
Self-merging to unbreak

PR #3645, opened 2024-01-26
Fix a link that wasn't rendering properly

sgibson91:


sgibson91:
This is uncontroversial so will self-merge

Issue #3643, opened 2024-01-25
[Request deployment] HHMI Spyglass Ephemeral Hub


colliand:
### The GitHub handle of the community representative

pending (@colliand for now)

### Hub important dates

- as soon as possible and certainly by 2024-01-31

The template below does not exactly apply. This is a request for a publicly accessible binder designed (initially) to render notebooks in the Spyglass repository: https://github.com/LorenFrankLab/spyglass 

Eventually, this hub will likely be used to serve repos generated by various other research groups affiliated with HHMI. The current focus is to deploy infrastructure to enable a "magic link" that interactively renders Spyglass tutorial content. 

### Hub Authentication Type

pending

### First Hub Administrators

- edeno

### [GitHub Auth only] How would you like to manage your users?

None

### [GitHub Teams Auth only] Profile restriction based on team membership

Profile should be set up to work for Spyglass.

Spyglass uses an adjacent database. This was briefly discussed in a call with @yuvipanda .

### Hub logo image URL

should use same image on hhmi.2i2c.cloud hub which is not rendering right now; 

![HHMI-vertical-signature-color](https://github.com/2i2c-org/infrastructure/assets/1879041/76419ba9-6d1a-41fe-b9b7-56fd89e0da40)

### Hub logo website URL

https://hhmi.org

### Hub user image GitHub repository

to work with Spyglass for now. build on the fly via binder/GESIS tooling?

### Hub user image tag and name

pending

### Extra features you would like to enable

- [ ] Dedicated Kubernetes cluster
- [ ] Scalable Dask Cluster

### (Optional) Preferred cloud provider

None

### (Optional) Preferred cloud region

as with hhmi.2i2c.cloud

### (Optional) Billing and Cloud account

None

### Tasks to deploy the hub

- [ ] 1. Deploy information filled in above
- [ ] 2. Engineer who will deploy the hub is assigned
- [ ] 3. If using GitHub Orgs/Teams Auth, Engineer is given Owner rights to the org to set this up.
- [ ] 4. Initial Hub deployment PR <link to PR>
- [ ] 5. Administrators able to log on -> Hub now in steady-state

jmunroe:
There is overlap here with the request in 
- https://github.com/2i2c-org/infrastructure/issues/3624

[...]

PR #3646, opened 2024-01-29
Add ephemeral spyglass hub to the hhmi cluster

sgibson91:
- ref: https://github.com/2i2c-org/infrastructure/issues/3643

hub can be tested at https://spyglass.hhmi.2i2c.cloud

### DIFF
--- a/config/clusters/hhmi/spyglass.values.yaml
+++ b/config/clusters/hhmi/spyglass.values.yaml
@@ -9,6 +9,8 @@ jupyterhub:
         hosts:
           - spyglass.hhmi.2i2c.cloud
   custom:
+    jupyterhubConfigurator:
+      enabled: false
     singleuserAdmin:
       # Turn off trying to mount shared-readwrite folder for admins
       extraVolumeMounts: []
--- a/config/clusters/hhmi/spyglass.values.yaml
+++ b/config/clusters/hhmi/spyglass.values.yaml
@@ -1,13 +1,5 @@
 nfs:
-  enabled: true
-  dirsizeReporter:
-    enabled: true
-  pv:
-    mountOptions:
-      - soft
-      - noatime
-    serverIP: 10.55.112.74
-    baseShareName: /homes/
+  enabled: false
 jupyterhub:
   ingress:
     hosts:
--- a/config/clusters/hhmi/spyglass.values.yaml
+++ b/config/clusters/hhmi/spyglass.values.yaml
@@ -33,6 +33,9 @@ jupyterhub:
           name: ""
           url: ""
           custom_html: <a href="https://www.hhmi.org/">HHMI</a> and <a href="https://strategiesos.org/about/">Stratos</a>
+  prePuller:
+    hook:
+      enabled: true
   singleuser:
     initContainers: []
     storage:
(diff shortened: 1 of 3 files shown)


config/clusters/hhmi/spyglass.values.yaml
```diff
@@ -0,0 +1,84 @@
...
+  dirsizeReporter:
+    enabled: true
+  pv:
+    mountOptions:
+      - soft
+      - noatime
+    serverIP: 10.55.112.74
+    baseShareName: /homes/
+jupyterhub:
+  ingress:
```
yuvipanda, line 12: I've also documented now that we should enable the hook prepuller and disable the configurator. https://github.com/2i2c-org/infrastructure/pull/3653

yuvipanda (approved):
Two straightforward suggestions, I think you can merge once those are in @sgibson91!

config/clusters/hhmi/spyglass.values.yaml
sgibson91: I get a validation error and cannot deploy if I don't define the rest, so there is a bug in our schema:

```
1 / 1: Validating non-encrypted hub values files for spyglass...
Error: values don't meet the specifications of the schema(s) in the following chart(s):
basehub:
- nfs.pv: Must validate at least one schema (anyOf)
- nfs.pv: serverIP is required
- nfs.pv: baseShareName is required
```

sgibson91: See [`bcb736c` (#3646)](https://github.com/2i2c-org/infrastructure/pull/3646/commits/bcb736cabc7c993d5dddb029aac2b45a056d48c0) for change

sgibson91: This is the related schema section https://github.com/2i2c-org/infrastructure/blob/cfb4e641b7993922c640f523bae87010d844b309/helm-charts/basehub/values.schema.yaml#L189-L201

sgibson91: This config fixed it:

```
nfs:
  enabled: false
  pv:
    enabled: false
```

But I am surprised that the top level `nfs.enabled = false` didn't automatically turn off the nested `pv` section

consideRatio: Hmmm, so there is an unconditional requirement to have `nfs.pv` declared:

https://github.com/2i2c-org/infrastructure/blob/cfb4e641b7993922c640f523bae87010d844b309/helm-charts/basehub/values.schema.yaml#L156-L163

basehub's default values declares `nfs.pv`, and has `nfs.pv.enabled=True`:

https://github.com/2i2c-org/infrastructure/blob/cfb4e641b7993922c640f523bae87010d844b309/helm-charts/basehub/values.yaml#L43-L48

The schema section for `nfs.pv` is conditional, where more fields are required when `nfs.pv.enabled=True`.

https://github.com/2i2c-org/infrastructure/blob/cfb4e641b7993922c640f523bae87010d844b309/helm-charts/basehub/values.schema.yaml#L189-L204

---

I think the improvement is to not default to `nfs.pv.enabled=true` since other fields are required to be set for that, but instead default nfs.pv.enabled=false` and set it to true alongside other required fields.

Opening an issue about this.

consideRatio: Tracked in https://github.com/2i2c-org/infrastructure/issues/3654

PR #3651, opened 2024-01-30
Disable NFS completely when setting up ephemeral hub

yuvipanda:
We disallow all persistent storage anyway, so we don't need the PersistentVolume and other base infrastructure setup by the nfs stanza


consideRatio (approved):

yuvipanda:
Thanks, @consideRatio

yuvipanda:
@sgibson91 ah yeah i agree it should be fixed in schema!

I opened https://github.com/2i2c-org/infrastructure/pull/3656 anyway though, in case the schema takes a while to fix.

Issue #3654, opened 2024-01-30
basehub: switch default value of nfs.pv.enabled to false (without changing any final values)


consideRatio:
@sgibson91 discovered that setting `nfs.enabled=false` where you wouldn't want to declare anything else still required `nfs.pv.enabled=false` to be set.

The reason for this is detailed in https://github.com/2i2c-org/infrastructure/pull/3646#discussion_r1471011667, and I think the fix is explicitly start setting `nfs.pv.enabled=true` alongside setting other fields that are required under `nfs.pv` such as `nfs.pv.serverIP`, and defaulting to `nfs.pv.enabled=false` in basehub.

This aligns with the Helm chart best practice of making charts renderable without being passed any config. Now it wouldn't be, because `nfs.enabled=true` and `nfs.pv.enabled=true` by default while we still havn't set the then required `nfs.pv.serverIP`.

yuvipanda:
When this is fixed, this documentation change should be undone (https://github.com/2i2c-org/infrastructure/pull/3656)

PR #3656, opened 2024-01-30
Mention that nfs.pv should also be set to false

yuvipanda:
Until https://github.com/2i2c-org/infrastructure/issues/3654 gets fixed, we update the ephemeral hub docs so folks who are setting up an ephemeral hub can just copy paste


consideRatio:
The relevant changes looks good, but the PR includes unrelated changes!

yuvipanda:
@consideRatio I've extracted out the extra bit into its own PR correctly in https://github.com/2i2c-org/infrastructure/pull/3684, and this one is clean now.

sgibson91 (approved):

PR #3666, opened 2024-01-31
basehub: default to nfs.enabled / nfs.pv.enabled to false

### LINKED ISSUE #3654: basehub: switch default value of nfs.pv.enabled to false (without changing any final values)

consideRatio:
Closes #3654.

We are still not following the best practice of providing a chart (basehub) that works out of the box with its default config, because at least we still have `jupyterhub.singleuser.extraVolumeMounts` and `jupyterhub.custom.singleuserAdmin.extraVolumeMounts` that mounts NFS folders even if `nfs.enabled` and `nfs.pv.enabled` aren't true, which would make startup fail.

This PR does not revert #3656 because I think many clusters will have hubs enabling and configuring `nfs` stuff in a common.values.yaml, while individual hubs may then go back to opting out of it.

Overall, this is something I consider an improvement, but it isn't as clearcut as I initially thought.

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -64,14 +64,14 @@ staticWebsite:
       privateKey: ""
 
 nfs:
-  enabled: true
+  enabled: false
   dirsizeReporter:
     enabled: true
   shareCreator:
     enabled: true
     tolerations: []
   pv:
-    enabled: true
+    enabled: false
     mountOptions:
       - soft
       - noatime
--- a/config/clusters/2i2c-aws-us/basehub-common.values.yaml
+++ b/config/clusters/2i2c-aws-us/basehub-common.values.yaml
@@ -1,5 +1,7 @@
 nfs:
+  enabled: true
   pv:
+    enabled: true
     # from https://docs.aws.amazon.com/efs/latest/ug/mounting-fs-nfs-mount-settings.html
     mountOptions:
       - rsize=1048576
--- a/config/clusters/2i2c-aws-us/cosmicds.values.yaml
+++ b/config/clusters/2i2c-aws-us/cosmicds.values.yaml
@@ -1,5 +1,6 @@
 nfs:
-  # No persistent storage for cosmicds
+  # nfs functionality explicitly disabled in case a common.values.yaml
+  # file is used to enable it for all hubs in the cluster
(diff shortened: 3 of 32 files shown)


consideRatio:
This change is a systematic change, so it can quickly get out of sync and needs to be updated/reviewed/merged in quick succession. I'd like to update this PR to be mergabe once the PR is reviewed in outs outdated form, where focus is on the kind of changes, and the retroactive update is focused on making sure the changes are systematic once again.

helm-charts/basehub/values.yaml
```diff
@@ -64,14 +64,14 @@ staticWebsite:
...
-  enabled: true
+  enabled: false
   dirsizeReporter:
     enabled: true
   shareCreator:
     enabled: true
     tolerations: []
   pv:
-    enabled: true
+    enabled: false
```
consideRatio, line 74: Since this switches these from true to false, I'm updating all cluster config explicitly configuring nfs.

consideRatio:
Thank you @GeorgianaElena!

I've now rebased and verified the change is sound twice by searching for `nfs:` across the repo and ensured that configuring `nfs:` in any way goes hand in hand with configuring `nfs.enabled=true` and `nfs.pv.enabled=true`. If we would miss enabling `nfs` somewhere it were configured, it would be a breaking change.

consideRatio:
Actially I think any mistakes here will be caught by the health check tests, because a user server will be started with storage mounted - storage that can't be mounted unless nfs stuff has been enabled.

I'll go for a merge before new hub PRs are being created as they will be soon.

Issue #3670, opened 2024-01-31
[Support] Increase RAM size on spyglass.hhmi.2i2c.cloud to 3.7GB


jmunroe:
### Context

This request is from the #hhmi_pilot Slack channel and is related to 
- https://github.com/2i2c-org/infrastructure/issues/3624

### Proposal

I think I just need to add the following to `config/clusters/hhmi/spyglass.values.yaml`

```
kubespawner_override:
  mem_guarantee: 3644869120
  mem_limit: 3644869120
  cpu_guarantee: 0.434625
  cpu_limit: 3.477
  node_selector:
    node.kubernetes.io/instance-type: n2-highmem-4
```

### Updates and actions

I'll assign myself to complete this task.

yuvipanda:
@jmunroe yes, that sounds good to me. We'll have to tweak the limit values eventually based on observed usage for this though. 

You can also self merge this when done.

PR #3687, opened 2024-02-02
hhmi:spyglass change available RAM on spyglass to 3.4GB

jmunroe:
I've added an `kubespawner_override`  to that the memory is set to 3.4GB.  It looks like `kubespawner_override` has to be under a `profile_list` but i this application, there is only one item in that list.  

Is there a way for skipping the UI that shows the list of profile options and defaulting to this one and only option?

### DIFF
--- a/config/clusters/hhmi/spyglass.values.yaml
+++ b/config/clusters/hhmi/spyglass.values.yaml
@@ -66,6 +66,15 @@ jupyterhub:
     image:
       name: quay.io/2i2c/hhmi-spyglass-image
       tag: "67523d9ea855"
+    profileList:
+      - display_name: "Spyglass"
+        description: "3.4GB RAM environment for spyglass notebooks"
+        default: true
+        kubespawner_override:
+          mem_guarantee: 3.4G
+          mem_limit: 3.4G
+          node_selector:
+            node.kubernetes.io/instance-type: n2-highmem-4
     extraContainers:
       - name: mysql
         image: datajoint/mysql # following the spyglass tutorial at https://lorenfranklab.github.io/spyglass/latest/notebooks/00_Setup/#existing-database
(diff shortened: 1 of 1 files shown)


jmunroe:
Resolves #3670

sgibson91:
@jmunroe I think you can set mem/cpu limits/guarantees directly under the `singleuser` key according to these docs: https://z2jh.jupyter.org/en/stable/jupyterhub/customizing/user-resources.html#set-user-memory-and-cpu-guarantees-limits I don't know what that would mean for the node_selector though

sgibson91:
Ok, I think you can do node selection like this: https://github.com/2i2c-org/infrastructure/commit/2dda0ec31ec2e4b5294f41aa3a2faeed80528ffd

sgibson91:
Here's a more complete example from the COESSING values file https://github.com/2i2c-org/infrastructure/blob/4ddad9b3dfa0da35aa209f03dddffb6174f4a58f/config/clusters/pangeo-hubs/coessing.values.yaml#L28-L34

jmunroe:
Thanks @sgibson91 !  Here was my confusion: in many hubs we use `profileList` and to set the memory limits and requirements requires something like:

```
singleuser:
  profileList:
    - kubespawner_override:
        mem_guarantee: 3.4G
        mem_limit: 3.4G
        node_selector:
          node.kubernetes.io/instance-type: n2-highmem-4
```

while memory requirements/limits can also be set (without profiles) as

```
singleuser:
  memory: 
    limit: 3.4G 
    guarantee: 3.4G 
  nodeSelector: 
    node.kubernetes.io/instance-type: n2-highmem-4
```

The difference in syntax is odd to me (e.g. `node_selector` in one case and `nodeSelector` in the other!) 

But your suggestion of the COESSING hub as example is perfect. Thank you.

sgibson91:
The difference between `nodeSelector` and `node_selector` is that `nodeSelector` is read directly as a z2jh helm chart value, whereas `kubespawner_override` and therefore `node_selector` are parsed directly to the KubeSpawner python code. These are two separate GitHub repos (though connected by the helm chart), so may have different syntax expectations. I don't know the historical or conventional reasons why, but I know that using the override means we're dropping more directly into the Python code of KubeSpawner and so that's why the syntax changes.

yuvipanda (approved):
One suggestion but otherwise good to go

Issue #3703, opened 2024-02-06
Cleanup and validate that we don't overspecify image config


consideRatio:
From [this support ticket](https://2i2c.freshdesk.com/a/tickets/794).

> One final question about manually updating the tags in `config/clusters/leap/common.values.yaml`: I have been confused about [these lines](https://github.com/2i2c-org/infrastructure/blob/cfdab1b7710add1d89ca49a833a5736f67832a5e/config/clusters/leap/common.values.yaml#L79-L81)
>
> https://github.com/2i2c-org/infrastructure/blob/cfdab1b7710add1d89ca49a833a5736f67832a5e/config/clusters/leap/common.values.yaml#L79-L81
>
> and whether they actually have an effect on the images available. I have not changed these recently but wanted to double check if that is the correct way to do it.

If the profileList options all provide an image override, the `singleuser.image` config will not have an effect. Due to this, it would be good if we managed to cleanup this config where its ignored if we for example also provide a profileList config overriding the image. Its hard to write robust logic that ensures this though.

This complexity is related to ensuring we don't specify the image via jupyterhub-configurator, which then makes a change to `singleuser.image` irrelevant as well.

PR #3704, opened 2024-02-06
leap, all hubs: remove ignored singleuser.image config for clarity

consideRatio:
This is addressing #3703 for leap specifically.


consideRatio:
Thank you @AIDEA775!!

Issue #3695, opened 2024-02-03
[Support] Add `nasa-openscapes-workshops` org to access list for openscapes


yuvipanda:
### The Freshdesk ticket link

https://2i2c.freshdesk.com/a/tickets/1284

### Ticket request type

Feature Request

### Ticket impact

🟨 Medium

### Short ticket description

This came out of https://github.com/2i2c-org/infrastructure/issues/3614

Add a GitHub org + some specific teams to the access list for openscapes.

This is a new org, so we may need to do the additional steps required to make sure that we have access to those.

yuvipanda:
I've temporarily invited @sgibson91, @GeorgianaElena and @consideRatio to be owners of this org so it is easier for them to get this done.

sgibson91:
I've taken this on, but @yuvipanda is it also possible to get access to the https://github.com/nasa-openscapes org as well? I've nervous that in reforcing the auth flow to grant permissions to the new org, I've revoked permissions to the existing org and want to double check.

yuvipanda:
@sgibson91 you have been invited to that org as well!

sgibson91:
Thanks Yuvi! I confirmed that both orgs have maintained access to the OAuth app. Erik had a question in #3705 about whether user servers should be restricted to specific teams. That's a question for the community and I don't really want to be interacting with Freshdesk when I'm not on rotation.

yuvipanda:
The community has reported that this is not working, via freshdesk.

sgibson91:
@yuvipanda did #3716 fix this? Can it be closed again?

yuvipanda:
Yes it did!

PR #3705, opened 2024-02-06
Add nasa-openscapes-workshops to list of allowed orgs to access openscapes hubs

sgibson91:
- related https://github.com/2i2c-org/infrastructure/issues/3695

### DIFF
--- a/config/clusters/openscapes/common.values.yaml
+++ b/config/clusters/openscapes/common.values.yaml
@@ -46,7 +46,11 @@ basehub:
             - NASA-Openscapes:workshopaccess-2i2c
             - NASA-Openscapes:longtermaccess-2i2c
             - NASA-Openscapes:championsaccess-2i2c
-            - nasa-openscapes-workshops  # Requested in: https://2i2c.freshdesk.com/a/tickets/1284
+            # Requested in: https://2i2c.freshdesk.com/a/tickets/1284
+            - nasa-openscapes-workshops:AdminTeam
+            - nasa-openscapes-workshops:ChampionsAccess-2i2c
+            - nasa-openscapes-workshops:LongtermAccess-2i2c
+            - nasa-openscapes-workshops:WorkshopAccess-2i2c
           kubespawner_override:
             image: openscapes/python:4f340eb
           profile_options: &profile_options
--- a/config/clusters/openscapes/common.values.yaml
+++ b/config/clusters/openscapes/common.values.yaml
@@ -160,6 +160,9 @@ basehub:
           allowed_teams:
             - NASA-Openscapes:longtermaccess-2i2c
             - 2i2c-org:hub-access-for-2i2c-staff
+            # Requested in: https://2i2c.freshdesk.com/a/tickets/1284
+            - nasa-openscapes-workshops:AdminTeam
+            - nasa-openscapes-workshops:LongtermAccess-2i2c
           profile_options:
             image:
               display_name: Image
--- a/config/clusters/openscapes/common.values.yaml
+++ b/config/clusters/openscapes/common.values.yaml
@@ -46,6 +46,7 @@ basehub:
             - NASA-Openscapes:workshopaccess-2i2c
             - NASA-Openscapes:longtermaccess-2i2c
             - NASA-Openscapes:championsaccess-2i2c
+            - nasa-openscapes-workshops  # Requested in: https://2i2c.freshdesk.com/a/tickets/1284
           kubespawner_override:
             image: openscapes/python:4f340eb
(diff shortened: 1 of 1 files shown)


consideRatio:
I think the community request is too vaguely specified still, and that this isn't ready to be merged yet.

1. Who should be allowed to login?

   Currently, the following org:teams are allowed to login. Note how the org `NASA-Openscapes` is mentioned instead of `nasa-openscapes-workshops`.

   ```yaml
          allowed_organizations:
            - NASA-Openscapes:workshopaccess-2i2c
            - NASA-Openscapes:longtermaccess-2i2c
            - NASA-Openscapes:championsaccess-2i2c
   ```

   Do they want to transition from providing access via one github org to another github org? Then I figure a transitionary period should list both orgs and teams in the orgs.

1. What server profiles should be visible to what teams?

   There are 4 profiles: Python, R, Matlab, and "Bring your own image".
   
   I guess they want to limit access to "Bring your own image", because that lists fewer teams listed currently (Only `NASA-Openscapes:longtermaccess-2i2c`). But this should include the `AdminTeam` as well I figure, which isn't mentioned anywhere.

sgibson91:
There was a list of specific teams in the freshdesk ticket that were alluded to but not explicitly listed in the issue.

I don't know about servers, but they are not restricted for any of the other teams.

sgibson91:
Actually, "bring your own image" is the only profile that is restricted by teams, and by doing nothing/not changing that config, the new teams added won't be able to see it anyway. So I do think the PR in it's current state is fit for merging.

yuvipanda (changes requested):
(Edit because I forgot to write it): This PR only seems to add them to the python profile, rather than to the authenticator's allowed list or to all the profiles.

The teams need to also be added in the following places:

https://github.com/2i2c-org/infrastructure/pull/3705/files#diff-3e6a6aaa229aa8ec507fc308feb5b502a0683705283d6c0497e3d991339f6a0fR176 (without this, they can't actually log in)
https://github.com/2i2c-org/infrastructure/pull/3705/files#diff-3e6a6aaa229aa8ec507fc308feb5b502a0683705283d6c0497e3d991339f6a0fR139
https://github.com/2i2c-org/infrastructure/pull/3705/files#diff-3e6a6aaa229aa8ec507fc308feb5b502a0683705283d6c0497e3d991339f6a0fR126
(without the last two, they can't see the non-python profiles)

https://github.com/2i2c-org/infrastructure/pull/3705/files#diff-3e6a6aaa229aa8ec507fc308feb5b502a0683705283d6c0497e3d991339f6a0fR151 - I think here the longtermaccess and admin teams should be added. 

We should eventually have a template / runbook for this.

yuvipanda:
I think @sgibson91 has identified a part that's missing in our documentation - we don't have anything on how to ask the community which profiles they would like teams to be given access to. In this specific case, I can answer that clearly because it was discussed during the meeting they alluded to in the ticket with me and @jmunroe. But that's not generally going to be true.

sgibson91:
> (Edit because I forgot to write it): This PR only seems to add them to the python profile, rather than to the authenticator's allowed list or to all the profiles.

Yeah, it's because I'm finding the config very difficult to skim. I guess I would've expected the auth stuff to be higher up so as I read down the file my brain would think "okay, these are all the teams/orgs that are allowed, then we move down into specific profiles and which specific teams can access them".

sgibson91:
> [#3705 (files)](https://github.com/2i2c-org/infrastructure/pull/3705/files#diff-3e6a6aaa229aa8ec507fc308feb5b502a0683705283d6c0497e3d991339f6a0fR151) - I think here the longtermaccess and admin teams should be added.

I'm a bit unclear on what you mean here. Do you mean they should be the _only_ ones added? You say "should be added" but they are already there?

Other than that point, I have addressed the other places you mentioned.

yuvipanda:
@sgibson91 

> You say "should be added" but they are already there?

Currently, in https://github.com/2i2c-org/infrastructure/pull/3705/files#diff-3e6a6aaa229aa8ec507fc308feb5b502a0683705283d6c0497e3d991339f6a0fR161, it lists `NASA-Openscapes:longtermaccess-2i2c` and `2i2c-org:hub-access-for-2i2c-staff` as the teams that get access. Here, we should add `nasa-openscapes-workshops:AdminTeam` and `nasa-openscapes-workshops:LongtermAccess-2i2c`.

sgibson91:
My brain is just not operating. I don't think I ever really came back from Christmas.

sgibson91:
Done

yuvipanda (approved):
ty @sgibson91 :)

Issue #3693, opened 2024-02-03
[Support] Setup `gh-scoped-creds` for Openscapes


yuvipanda:
### The Freshdesk ticket link

https://2i2c.freshdesk.com/a/tickets/1277

### Ticket request type

Feature Request

### Ticket impact

🟨 Medium

### Short ticket description

This came out of https://github.com/2i2c-org/infrastructure/issues/3614.

Set up `gh-scoped-creds` for use by openscapes. We already have some engineering facing documentation for this in our repo.

consideRatio:
I marked the freshdesk ticket as pending, linking to this issue. Respond to the freshdesk issue after resolving this issue.

sgibson91:
I have done this in #3706 but have not deployed it yet as I don't want to overwrite changes made in #3705 (for #3695)

- GitHub app public link: https://github.com/apps/openscapes-github-push-access

colliand:
FYI @jnywong as this issue overlaps with a documentation task you are working on during the upcoming week.

jnywong:
Thanks @colliand ! Yeah installing a GH App to the GH org of a user community is not well explained in our user docs. T

There's an associated task with updating the `infrastructure` repo with the GH App client ID whenever a community wishes to use `gh-scoped-creds` (see [Infrastructure Guide](https://infrastructure.2i2c.org/howto/features/github/)). Will 2i2c engineering continue to take ownership of this task? And does this generally get picked up via FreshDesk?

sgibson91:
closed by https://github.com/2i2c-org/infrastructure/pull/3706

I'm closing this as the technical work has been completed. @jnywong would you like to track your question in a different/dedicated issue?

jnywong:
> closed by #3706
> 
> I'm closing this as the technical work has been completed. @jnywong would you like to track your question in a different/dedicated issue?

Done!

PR #3706, opened 2024-02-06
Add gh-scoped-creds env vars for openscapes hubs

sgibson91:
- related https://github.com/2i2c-org/infrastructure/issues/3693

This is not deployed yet as I did not want to overwrite the change in #3705

### DIFF
--- a/config/clusters/openscapes/common.values.yaml
+++ b/config/clusters/openscapes/common.values.yaml
@@ -37,6 +37,9 @@ basehub:
     singleuser:
       serviceAccountName: cloud-user-sa
       defaultUrl: /lab
+      extraEnv:
+        GH_SCOPED_CREDS_CLIENT_ID: "Iv1.6981e043b45f042f"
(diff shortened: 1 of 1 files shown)


config/clusters/openscapes/common.values.yaml
```diff
@@ -37,6 +37,9 @@ basehub:
     singleuser:
       serviceAccountName: cloud-user-sa
       defaultUrl: /lab
+      extraEnv:
+        GH_SCOPED_CREDS_CLIENT_ID: "Iv1.6981e043b45f042f"
+        GH_SCOPED_CREDS_ALL_URL: https://github.com/apps/openscapes-github-push-access
```
yuvipanda, line 42:
```suggestion
        GH_SCOPED_CREDS_APP_URL: https://github.com/apps/openscapes-github-push-access
```

sgibson91: thank you!

PR #3720, opened 2024-02-13
hhmi: update image tags for hhmi and spyglass

jmunroe:
Updated image tags upon request for LorenFrankLab


jmunroe:
Self-merging since only image is changed.

Issue #3631, opened 2024-01-22
[Epic] Standardize Catalyst Project Hubs


jmunroe:
We are making progress deploying Catalyst Project hubs.  Here's what we have to deployed to date:

**Catalyst Project, LatAm**(GCP: southamerica-east1)

- https://staging.latam.catalystproject.2i2c.cloud
- https://unitefa-conicet.latam.catalystproject.2i2c.cloud
- https://gita.latam.catalystproject.2i2c.cloud
- https://cicada.latam.catalystproject.2i2c.cloud

**Catalyst Project, Africa** (AWS: af-south-1)

- https://staging.af.catalystproject.2i2c.cloud
- https://must.af.catalystproject.2i2c.cloud
- https://nm-aist.af.catalystproject.2i2c.cloud
- https://uvri.af.catalystproject.2i2c.cloud

I like to normalize what we are offering to Catalyst communities (until a particular community has different needs).

1. Standardize on a ~8 GB RAM, ~1.0 CPU machine type per user with four users / node (using "small" machines).
2. Make both a community maintained RStudio and an JupyterLab environment available.
3. Include a unlisted_choice option so communities can test deploying their own images.
4. Deploy a landing page (in English, Spanish, or Portuguese) that references back to the Catalyst Project website for additional information
5. Use GitHub authentication (and team membership in the Catalyst-Hubs GitHub organization) as the default unless a community has elected to use a Google Suite style authentication
6. Use staging.latam and staging.af as 'reference' deployments and GitHub authentication through Catalyst-Hubs to allow investigation of the platform by potential new community champions.

Our Collaborative Lesson Training Development can then make the assumption that this is the configuration of hub a community is starting from.  Potential items to train community champions on:
1. User intro to a JupyterHub. The difference between RStudio or JupyterLab 'images/environments' and the overall JupyterHub platform. Cloud computing concepts like machine type, node, autoscaling. When a hub is automatically shut down.  Memory limits. Example cloud workflow demonstration.
2. Admin intro a JupyterHub.  Explain authentication via Google Suite or via GitHub Teams.  Activity: add a new user to a GitHub team. Using the JupyterHub admin page to see activity of other users.  Explain running multiple servers for one user. Explain memory/cpu limits and the options available for larger machine types.
3. Image management.  Some combination of  repo2docker, binder, Docker, and configuring both Python and R environments.  Give example of QGIS as example that hubs can also support full desktop applications as well.
[...]

PR #3721, opened 2024-02-14
[catalyst-latam] Standarize profileList

AIDEA775:
This changes was deployed in staging hub: https://staging.latam.catalystproject.2i2c.cloud

ref: #3631


config/clusters/catalystproject-latam/common.values.yaml
```diff
@@ -12,134 +12,80 @@ jupyterhub:
...
                 kubespawner_override:
-                  mem_guarantee: 57.775G
-                  cpu_guarantee: 7.5
-              mem_128:
-                display_name: ~128 GB, ~16.0 CPU
-                kubespawner_override:
-                  mem_guarantee: 115.549G
-                  cpu_guarantee: 15.0
+                  mem_guarantee: 8G
+                  mem_limit: 16GB
```
consideRatio, line 62: Checking the logs on the hub pod after the spawn failure, I could see the following error.

```
    traitlets.traitlets.TraitError: 16GB is not a valid memory specification. Must be an int or a string with suffix K, M, G, T
```

```suggestion
                  mem_limit: 16G
```

consideRatio (approved):
Nice!! I suggest we try to go for a merge and iterate from there!

There are some inefficiencies around requesting 8GB exactly for example, as that would only fit 3 users on a 32 GB node due to some capacity is reserved by k8s and some is needed by a few system pods running on the node besides the user pods. I discussed this in https://github.com/2i2c-org/infrastructure/issues/3631#issuecomment-1943198542.

I was when I wrote that more positive towards use of `n2-highmem-2` nodes, but I think now that I've remembered the complexities about overhead that becomes relatively large on smaller node makes me think its good to avoid use of a combination of 2 CPU and 4 CPU nodes, and figure it may very well be that 4 CPU nodes is needed to fit a few users per node in order to avoid regular startup delays.

### Deliberation

- **Should memory requests be lowered to account for 32 GB not being available?**
  I think let's not try to do it and make the UX simple, if you request 1, 2, 4, 8GB, you get exactly that. At worst we use ~25% of the capacity inefficiently on a 4 CPU node, which happens if all users start the 8 GB servers.
- **Should we have a node running?**
  I think having one node running at all time would be a reasonable optimization to ensure startup times. It adds cloud costs, but its quite a big deal I think to get quick access.

Issue #3667, opened 2024-01-31
[Request deployment] New Hub: sciencecore.opensci.2i2c.cloud


jmunroe:
This hub is part of
- https://github.com/2i2c-org/meta/issues/788

Functionality requested:

- Dynamic image building -- YES via binderhub-service + fancy profile stuff as seen in imagebuilding-demo
- Authenticated users -- YES
- Persistent user storage -- YES
- dask-gateway -- NO
- magic-links -- No requirement to do something here.

### The GitHub handle of the community representative

@jmunroe

### Technical support contacts listed

- [ ] Technical contact present

### Hub important dates

- **Target Date** Feb 14, 2024 
- To be demo'd at a NASA TOPS-T meeting Feb 27-29

### Hub Authentication Type

GitHub (e.g., @mygithubhandle)

### First Hub Administrators

@jmunroe 

### [GitHub Auth only] How would you like to manage your users?

team membership under https://github.com/ScienceCore

### Hub logo image URL

2i2c logo

### Hub logo website URL

2i2c.org

### Hub user image tag and name

I assume we'll need a "default" image even if we support dynamic image building.

Let's deploy a jupyter-scipy image and a rocker-binder image

### Extra features you would like to enable

- [x] Dedicated Kubernetes cluster -- depends on our ability to do cloud billing (see below)
- [ ] Scalable Dask Cluster

### (Optional) Preferred cloud provider

AWS

### (Optional) Preferred cloud region

us-west-2 

### (Optional) Billing and Cloud account

We could use the existing cluster: 2i2c-aws-us IF there is way to bill internally charges on this hub back to the HPOSS grant.

If this is not possible, this hub may need to go on its own cluster.

DECISION: dedicated cluster. 

### Tasks to deploy the hub

- [ ] 1. Deploy information filled in above
- [ ] 2. Engineer who will deploy the hub is assigned
- [ ] 3. If using GitHub Orgs/Teams Auth, Engineer is given Owner rights to the org to set this up.
- [ ] 4. Initial Hub deployment PR <link to PR>
- [ ] 5. Administrators able to log on -> Hub now in steady-state

GeorgianaElena:
@jmunroe, just double-checking that this will be a `daskhub`, right?
Also, there are is a [redeem code that will to be applied](https://infrastructure.2i2c.org/sre-guide/support/credits/) on this new aws account, right?

GeorgianaElena:
[...]

PR #3733, opened 2024-02-21
[New] Opensci cluster and hub

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/3667


eksctl/opensci.jsonnet
```diff
@@ -0,0 +1,128 @@
...
+
+
+{
+    apiVersion: 'eksctl.io/v1alpha5',
+    kind: 'ClusterConfig',
+    metadata+: {
+        name: "opensci",
+        region: clusterRegion,
+        version: "1.28",
+    },
```
consideRatio, line 42: Nice that you push to use k8s 1.28!

This is the first time we get an EKS cluster at 1.28 - does it seem to work OK without issues? I'm confident our k8s resources etc should work in k8s 1.28 - but I'm not fully confident on not running into issues with misc EKS specific things like  `aws-ebs-csi-drivers` and such things.

Looking at https://eksctl.io/usage/schema/, it seems that k8s 1.29 is now supported, but I figure its perfectly fine if we stay on the next newest k8s version.

Based on https://infrastructure.2i2c.org/howto/upgrade-cluster/#upgrade-policy, we should go for upgrades to the latest minor version that is supported by EKS and is also considered mature based on our GKE policy - which is that its part of GKE's regular release channel - I see that 1.29 meets that criteria since [a few weeks ago](https://cloud.google.com/kubernetes-engine/docs/release-notes-regular).

Okay, this is in general off topic - but I'm glad to see we pushed up from eks 1.27 with this!!

GeorgianaElena:
Ah, there's also this in the binderhub pod:

```
[W 240223 11:18:49 registry:112] No docker config at /.docker/config.json
[W 240223 11:18:49 registry:154] No username for docker registry at https://index.docker.io/v1
[W 240223 11:18:49 registry:179] No password for docker registry at https://index.docker.io/v1
```

Which is similar to https://github.com/2i2c-org/infrastructure/pull/2699#issuecomment-1631051210

GeorgianaElena:
I managed to get a workaround on the situation above. @jmunroe, you should now be able to test and use the hub at https://sciencecore.opensci.2i2c.cloud

So, this hub now builds and pushes images to our quay.io account under the prefix `2i2c/opensci-sciencecore/` (https://quay.io/organization/2i2c?tab=repos). However, there's no way to make the images being pushed to quay be `public` and somehow the push defaults to private images.

So, the notebook pods need credentials to be able to pull these images in order to start the server. A temporary fix to this is to put the robot's account's credentials under `jupyterhub.imagePullSecret`. Note: this robot's account creds are also mounted as a secret into the binderhub-service build pods.

consideRatio:
@Georgiana nice work it seems like we can get an image built and that we then can launch it.

I figure we shouldn't use the 2i2c organization on quay.io long term, but that was the workaround you referred to right?

GeorgianaElena:
> I figure we shouldn't use the 2i2c organization on quay.io long term, but that was the workaround you referred to right?

@consideRatio, yes! This and the fact that the images being pushed to quay default to being private images, so we need `imagePullSecret` set.

I couldn't find a way to signal to quay to default any new image to be a public image :( I believe there is a `visibility` query string that we could set as `public` when sending the docker push request, but that wasn't super straightforward to solve.

config/clusters/opensci/sciencecore.values.yaml
```diff
@@ -0,0 +1,172 @@
...
+  hub:
+    allowNamedServers: true
+    services:
+      binder:
+        # FIXME: ref https://github.com/2i2c-org/binderhub-service/issues/57
+        # for something more readable and requiring less copy-pasting
+        url: http://sciencecore-binderhub-service:8090
+    image:
+      name: quay.io/2i2c/dynamic-image-building-experiment
+      tag: "0.0.1-0.dev.git.7567.ha4162031"
```
GeorgianaElena, line 119: I believe the hub image we're using by default has everything it needs and we don't need the "experimental" one amymore.

Will check tomorrow, but leaving a note to don't forget.

jmunroe:
People have been added to the hub and hopefully we'll get some 'real world' testing from other ScienceCore teams over the next few weeks.

The only 'issue' encountered is the domain name: looks like sciencecore.opensci.2i2c.cloud is blocked through NASA firewall or security settings.  Speculation is that *.2i2c.cloud may be the issue.

I'll be working with the NASA folks to establish a preferred domain to use but we shouldn't be blocked waiting for a decision.

I think this PR should be merged and say this sciencecore hub has been deployed!  We can create new issues/support issues for future iterations.

yuvipanda (commented):
This is incredible work, @GeorgianaElena. I've a couple suggestions, but otherwise this is good.

config/clusters/opensci/sciencecore.values.yaml
yuvipanda: If it does still need this image, I would be open to us just merging it in to our standard image now anyway. Doesn't need to happen as part of this PR though.

GeorgianaElena: Just checked and the only difference between the default hub image and this one is the presence of `jupyterhub-fancy-profiles`. Will open an issue for bringing it into the default one.

config/clusters/opensci/sciencecore.values.yaml
```diff
@@ -0,0 +1,172 @@
...
+        value: user
+        effect: NoSchedule
+      - key: hub.jupyter.org/dedicated
+        value: user
+        effect: NoSchedule
+  config:
+    BinderHub:
+      base_url: /services/binder
+      use_registry: true
+      image_prefix: quay.io/2i2c/opensci-sciencecore
```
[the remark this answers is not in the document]
GeorgianaElena: I thought ECR is not yet supported by binderhub https://github.com/jupyterhub/binderhub/issues/705. Ah, you mean, once its supported, we should use that instead of quay right?

config/clusters/opensci/sciencecore.values.yaml
```diff
@@ -0,0 +1,172 @@
...
+          - ScienceCore
+        scope:
+          - read:org
+
+    extraConfig:
+      enable-fancy-profiles: |
+        from jupyterhub_fancy_profiles import setup_ui
+        setup_ui(c)
+
+binderhub-service:
```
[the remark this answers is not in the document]
GeorgianaElena: Ah, great idea <3

config/clusters/opensci/sciencecore.values.yaml
GeorgianaElena: https://github.com/2i2c-org/infrastructure/issues/3757

GeorgianaElena:
Thanks for the feedback @yuvipanda. I believe I've addressed all the comments 🚀

config/clusters/opensci/sciencecore.values.yaml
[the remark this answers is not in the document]
yuvipanda: @GeorgianaElena ah, I had thought it is now supported given that we have a mybinder.org federation member running on AWS (thanks to Simon's work). I've pinged him in the ticket to ask.

yuvipanda (approved):
Thanks @GeorgianaElena! This looks good to me!

config/clusters/opensci/cluster.yaml
```diff
@@ -0,0 +1,23 @@
+name: opensci
+provider: aws
```
consideRatio, line 2:
```suggestion
provider: aws # https://2i2c.awsapps.com/start#/
```

consideRatio: Is `account` only needed in cluster.yaml when we don't assume our aws management account is used to manage this cluster's AWS account?

>```python
>        # We mostly use our 2i2c account to login into cloud provider UI's
>        # with a few exceptions
>        account = cluster.get("account", "2i2c")
>```

config/clusters/opensci/common.values.yaml
```diff
@@ -0,0 +1,12 @@
+nfs:
+  pv:
```
consideRatio, line 2: This is to align with https://github.com/2i2c-org/infrastructure/pull/3666/files - it is the current default made explicit.

```suggestion
  enabled: true
  pv:
    enabled: true
```

consideRatio (approved):
Thank you @GeorgianaElena for working this and pulling in some cluster unspecific improvements!

config/clusters/opensci/cluster.yaml
GeorgianaElena: @consideRatio, yes.

GeorgianaElena:
Thank  you @yuvipanda and @consideRatio! Merging 🚀

PR #3734, opened 2024-02-21
terraform: default to k8s 1.29 for new GKE/EKS clusters

consideRatio:
This is aligning with the policy on what k8s version to upgrade to in https://infrastructure.2i2c.org/howto/upgrade-cluster/#upgrade-policy.

At some point in time, we will trial use of a minor k8s version not already tried - I think the best time to do that is for new clusters as they don't have existing users. We can absolutely upgrade existing clusters to arrive at a new minor version never before used by us at 2i2c, but I figure we should be especially brave doing it for new clusters.


GeorgianaElena (approved):

consideRatio:
Thank you @GeorgianaElena !

Issue #3689, opened 2024-02-02
[Decommission Hub] Callysto


colliand:
### Summary

Callysto's funding from the Government of Canada ends soon. 

### Info

- **Community Representative:** @byrcyb
- **Link to New Hub issue:**  #1439 
- **Proposed end date:** 2024-02-29

#### Phase I

- [x] Confirm with Community Representative that the hub is no longer in use and it's safe to decommission
- [x] Confirm if there is any data to migrate from the hub before decommissioning
  - [ ] If yes, confirm where the data should be migrated to
    - [ ] Confirm a 2i2c Engineer has access to the destination in order to complete the data migration
  - [x] If no, confirm it is ok to delete all the data stored in the user home directories

#### Phase II - Hub Removal

(These steps are described in more detail in the docs at https://infrastructure.2i2c.org/en/latest/hub-deployment-guide/hubs/other-hub-ops/delete-hub.html)

- [x] Manage existing home directory data (migrate data from the hub or delete it)
- [ ] Manage existing cloud bucket data (migrate data, or delete it)
- [x] Delete the hub's authentication application on GitHub or CILogon (note CILogon removal requires the hub config in place)
- [x] Remove the appropriate `config/clusters/<cluster_name>/<hub_name>.values.yaml` files. A complete list of relevant files can be found under the appropriate entry in the associated `cluster.yaml` file.
- [x] Remove the associated hub entry from the `config/clusters/<cluster_name>/cluster.yaml` file.
- [x] Remove the hub deployment
  - `helm --namespace HUB_NAME delete HUB_NAME`
  - `kubectl delete namespace HUB_NAME`

#### Phase III - Cluster Removal

_This phase is only necessary for single hub clusters._

- [x] Remove the cluster's datasource from the central Grafana with:
  - `deployer grafana central-ds remove CLUSTER_NAME`
- [x] Run `terraform plan -destroy` and `terraform apply` from the [appropriate workspace](https://infrastructure.2i2c.org/en/latest/topic/terraform.html#workspaces), to destroy the cluster
- [x] Delete the terraform workspace: `terraform workspace delete <NAME>`
- [x] Remove the associated `config/clusters/<cluster_name>` directory and all its contents
- Remove the cluster from CI:
  - [x] [`deploy-hubs.yaml`](https://github.com/2i2c-org/infrastructure/blob/HEAD/.github/workflows/deploy-hubs.yaml)
[...]

PR #3742, opened 2024-02-27
Update the hub decommission issue template

sgibson91:
This PR provides some gotchas to the template I came across while fulfilling #3689

### DIFF
--- a/.github/ISSUE_TEMPLATE/3_decommission-hub.md
+++ b/.github/ISSUE_TEMPLATE/3_decommission-hub.md
@@ -40,18 +40,19 @@ Usually, it is because it was a hub that we created for a workshop/conference an
 - [ ] Remove the hub deployment
   - `helm --namespace HUB_NAME delete HUB_NAME`
   - `kubectl delete namespace HUB_NAME`
+  - TIP: Run `deployer use-cluster-credentials <cluster_name>` before running the above commands
 
 #### Phase III - Cluster Removal
 
 _This phase is only necessary for single hub clusters._
 
 - [ ] Remove the cluster's datasource from the central Grafana with:
-  - `deployer grafana central-ds remove CLUSTER_NAME`
+  - `deployer grafana central-ds remove <cluster_name>`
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/3_decommission-hub.md
```diff
@@ -40,18 +40,19 @@ Usually, it is because it was a hub that we created for a workshop/conference an
...
+  - `deployer grafana central-ds remove <cluster_name>`
 - [ ] Run `terraform plan -destroy` and `terraform apply` from the [appropriate workspace](https://infrastructure.2i2c.org/en/latest/topic/terraform.html#workspaces), to destroy the cluster
 - [ ] Delete the terraform workspace: `terraform workspace delete <NAME>`
+- [ ] Delete the terraform values file under the `projects` folder associated with the relevant cloud provider (e.g. `terraform/gcp/projects/` for GCP)
 - [ ] Remove the associated `config/clusters/<cluster_name>` directory and all its contents
 - Remove the cluster from CI:
   - [ ] [`deploy-hubs.yaml`](https://github.com/2i2c-org/infrastructure/blob/HEAD/.github/workflows/deploy-hubs.yaml)
-  - [ ] [`validate-clusters.yaml`](https://github.com/2i2c-org/infrastructure/blob/HEAD/.github/workflows/validate-clusters.yaml)
-- [ ] Remove the cluster from the list of grafana datasources at https://grafana.pilot.2i2c.cloud/datasources
+  - [ ] [`deploy-grafana-dashboards.yaml`](https://github.com/2i2c-org/infrastructure/blob/HEAD/.github/workflows/deploy-grafana-dashboards.yaml)
```
sgibson91, line 57: We dynamically generate which clusters to validate in a similar way to the deploy-hubs workflow. However, we do hardcode the cluster names in the deploy-grafana-dashboards workflow.

consideRatio (approved):
Thank you for taking the time to update these steps!

There is also one missing about uptime-checks I think, which can be done after the cluster.yaml is no longer listed. I think the procedure should be:

This step ought to be common for adding a hub I think, and I suspect we lack this also for that docs.

```shell
terraform init
terraform workspace select default
terraform apply
```

---

I opened #3744 to track this.

PR #3745, opened 2024-02-27
leap, lis, and climatematch hubs: avoid overriding shared-readwrite mount

### LINKED ISSUE #3738: [Support] LIS: Admins cannot write to shared-readwrite folder
### The Freshdesk ticket link

https://2i2c.freshdesk.com/a/tickets/1349

### Ticket request type

Feature Request

### Ticket impact

🟧 High

### Short ticket description

Matthew needs to write a dataset to the shared-readwrite folder but does not have the correct permissions, despite being an admin.

I then used `deployer exec homes` to move the dataset (which had been added to Matthew's home dir) into shared-readwrite, but it is not visible via the shared folder when I log in to my server.

The dataset needs to be available by 13:30 GMT on Monday.

### (Optional) Investigation results

I am also suffering the same permissions error :/

```
jovyan@jupyter-sgibson91:~$ mkdir -p shared-readwrite/test
mkdir: cannot create directory ‘shared-readwrite/test’: Permission denied
jovyan@jupyter-sgibson91:~$ sudo mkdir -p shared-readwrite/test
sudo: The "no new privileges" flag is set, which prevents sudo from running as root.
sudo: If sudo is running in a container, you may need to adjust the container configuration to disable the flag.
jovyan@jupyter-sgibson91:~$ 
```

I used the following command

```
deployer exec homes 2i2c-uk lis
```

This gave me a shell with the home directories mounted.
I found Matt's home directory and the `data/acsids` directory he mentioned, and used the following command to copy it into `shared-readwrite`

```
# From within Matt's home dir
cp -r data/acsids shared-readwrite
```

An `ls` showed the folder was now there

```
root@2i2c-uk-lis-shell:/home/matthew-2dbrett# ls shared-readwrite/
acsids
```

I then restarted my server on the lis hub, but could not find the folder in the `shared-readwrite` or `shared` folders, I don't know what's going on here :/

consideRatio:
Fix override of shared-readwrite mounting config

basehub defines the list `custom.singleuserAdmin.extraVolumeMounts` and
provides two entries to it.

If a hub re-specifies this list with a single new entry, it will end up
overriding entire list, making the two entries defined by default be
lost. This is a challenge with using helm values that are list, they
combine by replacing eachother rather than appending to each other.

Due to that, whenever we want to add a mount, we need to re-specify the
mounts that are in the list by default.

This commit impacts the following cluster, hubs:

- 2i2c-uk, lis
- 2i2c, climatematch
- leap, staging and prod

---

- Fixes #3738

### DIFF
--- a/config/clusters/2i2c-uk/lis.values.yaml
+++ b/config/clusters/2i2c-uk/lis.values.yaml
@@ -33,13 +33,22 @@ jupyterhub:
         funded_by:
           name: London Interdisciplinary School
           url: https://www.lis.ac.uk
-    # Extra mount point for admins to access to all users' home dirs
-    # Ref https://2i2c.freshdesk.com/a/tickets/228
     singleuserAdmin:
       extraVolumeMounts:
+        # /allusers is an extra mount point for admins to access to all users'
+        # home dirs, ref: https://2i2c.freshdesk.com/a/tickets/228.
         - name: home
           mountPath: /home/jovyan/allusers
           readOnly: false
+        # mounts below are copied from basehub's values that we override by
+        # specifying extraVolumeMounts (lists get overridden when helm values
+        # are combined)
+        - name: home
+          mountPath: /home/jovyan/shared-readwrite
+          subPath: _shared
+        - name: home
+          mountPath: /home/rstudio/shared-readwrite
+          subPath: _shared
   singleuser:
     image:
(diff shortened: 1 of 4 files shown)


sgibson91 (approved):

consideRatio:
Thank you @sgibson91!!

Issue #3732, opened 2024-02-21
Document how to deploy an `imgebuilding-demo` type of hub


GeorgianaElena:
### Context

The hub running at https://imagebuilding-demo.2i2c.cloud/ will be "replicated" as part of https://github.com/2i2c-org/infrastructure/issues/3667.

### Proposal

We should have docs similar to https://infrastructure.2i2c.org/howto/features/ephemeral/ on how to deploy such a hub and which are the config gotchas so we don't end up copying hub config from one hub to another that is not relevant.

### Updates and actions

- [ ] Document while deploying https://github.com/2i2c-org/infrastructure/issues/3667

PR #3749, opened 2024-02-29
Document how to deploy an `imagebuilding` type of hub

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/3732

I've also removed all the "layers" from the features index as they weren't rendering ok and we didn't standardize them anyway.

Rendered at https://2i2c-pilot-hubs--3749.org.readthedocs.build/howto/features/imagebuilding/

### DIFF
--- a/docs/howto/features/imagebuilding.md
+++ b/docs/howto/features/imagebuilding.md
@@ -39,33 +39,34 @@ jupyterhub:
 2. Since `jupyterhub-fancy-profiles` adds on to the [`profileList` feature of `KubeSpawner`](https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html#kubespawner.KubeSpawner.profile_list), we need to configure a profile list here as well. Edit and update the following configuration per the community requesting the hub needs.
 
     ```yaml
-    singleuser:
-      profileList:
-        - display_name: "Only Profile Available, this info is not shown in the UI"
-          slug: only-choice
-          profile_options:
-            image:
-              display_name: Image
-              unlisted_choice:
-                enabled: True
-                display_name: "Custom image"
-                validation_regex: "^.+:.+$"
-                validation_message: "Must be a publicly available docker image, of form <image-name>:<tag>"
-                display_name_in_choices: "Specify an existing docker image"
-                description_in_choices: "Use a pre-existing docker image from a public docker registry (dockerhub, quay, etc)"
-                kubespawner_override:
-                  image: "{value}"
-              choices:
-                pangeo:
-                  display_name: Pangeo Notebook Image
-                  description: "Python image with scientific, dask and geospatial tools"
+    jupyterhub:
+      singleuser:
+        profileList:
+          - display_name: "Only Profile Available, this info is not shown in the UI"
+            slug: only-choice
+            profile_options:
+              image:
+                display_name: Image
+                unlisted_choice:
+                  enabled: True
+                  display_name: "Custom image"
+                  validation_regex: "^.+:.+$"
+                  validation_message: "Must be a publicly available docker image, of form <image-name>:<tag>"
+                  display_name_in_choices: "Specify an existing docker image"
+                  description_in_choices: "Use a pre-existing docker image from a public docker registry (dockerhub, quay, etc)"
                   kubespawner_override:
-                    image: pangeo/pangeo-notebook:2023.09.11
-                scipy:
-                  display_name: Jupyter SciPy Notebook
-                  slug: scipy
-                  kubespawner_override:
-                    image: jupyter/scipy-notebook:2023-06-26
+                    image: "{value}"
+                choices:
+                  pangeo:
+                    display_name: Pangeo Notebook Image
+                    description: "Python image with scientific, dask and geospatial tools"
+                    kubespawner_override:
+                      image: pangeo/pangeo-notebook:2023.09.11
+                  scipy:
+                    display_name: Jupyter SciPy Notebook
+                    slug: scipy
+                    kubespawner_override:
+                      image: jupyter/scipy-notebook:2023-06-26
     ```
 
 ## The `binderhub-service` chart
@@ -102,39 +103,12 @@ The `build pods` are created as a result of an image build request, and they mus
 
   ```yaml
   binderhub-service:
-    nodeSelector:
-      hub.jupyter.org/node-purpose: core
     enabled: true
-    # Set a port on which the binderhub-service ClusterIP Service will run
-    service:
-      port: 8090
-    dockerApi:
-      nodeSelector:
-        # The dockerApi pods need to run on the same user nodes as the user pods
-        hub.jupyter.org/node-purpose: user
-      tolerations:
-        # Tolerate tainted jupyterhub user nodes
-        - key: hub.jupyter.org_dedicated
-          value: user
-          effect: NoSchedule
-        - key: hub.jupyter.org/dedicated
-          value: user
-          effect: NoSchedule
     config:
       BinderHub:
-        # jupyterhub-fancy-profile expects binderhub to be available under http://{{hub url}}/services/binder
-        base_url: /services/binder
-        use_registry: true
         # something like <region>-docker.pkg.dev/<project-name>/<repository-name> for grc.io
         # or quay.io/org/repo/cluster-hub/ for quay.io
         image_prefix: <repository_path>
-      KubernetesBuildExecutor:
-        # Get ourselves a newer repo2docker!
-        build_image: quay.io/jupyterhub/repo2docker:2023.06.0-8.gd414e99
-        node_selector:
-          # Schedule builder pods to run on user nodes only
-          hub.jupyter.org/node-purpose: user
-    # The password to the registry is stored encrypted in the hub's encrypted config file
     buildPodsRegistryCredentials:
       # registry server address like https://quay.io
       server: <server_address>
--- a/docs/howto/features/imagebuilding.md
+++ b/docs/howto/features/imagebuilding.md
@@ -101,40 +101,41 @@ The `build pods` are created as a result of an image build request, and they mus
 
 2. Setup the `binderhub-service` config
 
-  ```yaml
-  binderhub-service:
-    enabled: true
-    config:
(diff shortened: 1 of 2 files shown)


docs/howto/features/imagebuilding.md
```diff
@@ -0,0 +1,153 @@
...
+              slug: scipy
+              kubespawner_override:
+                image: jupyter/scipy-notebook:2023-06-26
+```
+
+## The `binderhub-service` chart
+
+We will use the [binderhub-service](https://github.com/2i2c-org/binderhub-service/) Helm chart to run BinderHub, the Python software, as a standalone service to build and push images with [repo2docker](https://github.com/jupyterhub/repo2docker), next to JupyterHub.
+
+### How it works
```
sgibson91, line 62: To be more in-line with the diataxis framework, I try and keep the "explanation of how stuff works" in topics and limit the how-to guides to the specific steps an engineer would take to deploy it, then add cross-refs between the two. E.g. the topic document could contain _brief_ explanations of what jupyterhub-fancy-profiles and binderhub-service are/do with links to their docs (if/when they exist).

docs/howto/features/index.md
```diff
@@ -5,85 +5,24 @@ There are several extra features and services that can be enabled on the 2i2c Ju
 These add extra functionality, connect with cloud services, etc.
 See the sections below for more details.
 
-## Layers
```
sgibson91, line 8: I really appreciate this, thank you. This page has been bothering me - mainly because it's called "enable specific hub features" and the first entry was "deploy a kubernetes cluster" 🙄

sgibson91 (approved):
LGTM! I have a minor comment about keeping with the diataxis framework

docs/howto/features/imagebuilding.md
```diff
@@ -0,0 +1,153 @@
...
+  - specifying another image, different from the ones in the pre-defined list
+  - building and pushing an image to a registry from a GitHub repository
+
+## Connect with `jupyterhub-fancy-profiles`
+
+1. The [jupyterhub-fancy-profiles](https://github.com/yuvipanda/jupyterhub-fancy-profiles) project provides a user facing frontend for connecting the JupyterHub to the BinderHub service, allowing the users to build their own images similar with how they would on mybinder.org.
+
+```yaml
+hub:
+  extraConfig:
```
yuvipanda, line 20: Just to note that if we still need to keep a separate image for this, that should be mentioned as well.

I'm also not 100% sure, but this code block may need to be indented 3 spaces to line up with the numbered list, so it is counted as *part* of the item rather than just a separate code block?

docs/howto/features/imagebuilding.md
```diff
@@ -0,0 +1,153 @@
...
+
+The `build pods` are created as a result of an image build request, and they must run on the same node as the builder pods to make use of the docker daemon. These pods mount a k8s Secret with the docker config file holding the necessary registry credentials.
+
+### How to set it up
+
+1. Setup a registry and safely store the credentials to push to it
+
+  - If the hub will be deployed on a GCP cluster, follow the steps 4 to 8 from the [binderhub-service installation guide](https://github.com/2i2c-org/binderhub-service?tab=readme-ov-file#installation). When you're done, you should have a gcr.io registry setup and the appropriate credentials to push images to it.
+
+  - If the hub will be deployed on another cloud provider than GCP, then you should setup the desired quay.io repository and robot account to push and pull from it. The steps below describes how to create one under the 2i2c quay.io organization, assuming this is where we wish the repositories that will be build by `binderhub-service` to be pushed.
```
yuvipanda, line 76: I would suggest that we actually create perhaps *one* new quay.io organization (like 2i2c-imagebuilding-hubs perhaps?) that can be used for this, instead of the existing 2i2c org. This strikes a decent balance between creating one new org per hub vs just re-using the same one we have.

docs/howto/features/imagebuilding.md
```diff
@@ -0,0 +1,153 @@
...
+    - Once done, click the name of the robot account again. This will give you its username and password.
+    - Select the 'Team and Membership' option on the left menu.
+    - Click on the 'Options' wheel of the `owners` team, then select 'Manage Team Members'
+    - Type in the name of the robot account that you created, select it from the list and add it to the `owners` team
+
+
+2. Setup the `binderhub-service` config
+
+  ```yaml
+  binderhub-service:
```
yuvipanda, line 91: I think some chunk of this can be simplified per https://github.com/2i2c-org/infrastructure/pull/3733/files#r1508207972

docs/howto/features/imagebuilding.md
yuvipanda: The code block formatting suggestion applies to all code blocks here. But if it renders visually ok, I don't particularly care otherwise :)

docs/howto/features/imagebuilding.md
```diff
@@ -0,0 +1,153 @@
...
+  extraConfig:
+    enable-fancy-profiles: |
+      from jupyterhub_fancy_profiles import setup_ui
+      setup_ui(c)
+```
+
+2. Since `jupyterhub-fancy-profiles` adds on to the [`profileList` feature of `KubeSpawner`](https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html#kubespawner.KubeSpawner.profile_list), we need to configure a profile list here as well. Edit and update the following configuration per the community requesting the hub needs.
+
+```yaml
+singleuser:
```
yuvipanda, line 29: This should probably be nested under `jupyterhub`

docs/howto/features/imagebuilding.md
```diff
@@ -0,0 +1,153 @@
...
+
+## The `binderhub-service` chart
+
+We will use the [binderhub-service](https://github.com/2i2c-org/binderhub-service/) Helm chart to run BinderHub, the Python software, as a standalone service to build and push images with [repo2docker](https://github.com/jupyterhub/repo2docker), next to JupyterHub.
+
+### How it works
+
+The `binderhub-service` installation starts a `docker-api` pod on each of the user nodes via the following [DaemonSet definition](https://github.com/2i2c-org/binderhub-service/blob/main/binderhub-service/templates/docker-api/daemonset.yaml).
+
+The `docker-api` pod setups and starts the [dockerd](https://docs.docker.com/engine/reference/commandline/dockerd/) daemon, that will then be accessible via a mounted unix socket, by the `build pods`.
```
yuvipanda, line 66:
```suggestion
The `docker-api` pod setups and starts the [dockerd](https://docs.docker.com/engine/reference/commandline/dockerd/) daemon, that will then be accessible via a mounted unix socket on the node, by the `build pods`.
```

docs/howto/features/imagebuilding.md
```diff
@@ -0,0 +1,153 @@
...
+
+The `build pods` are created as a result of an image build request, and they must run on the same node as the builder pods to make use of the docker daemon. These pods mount a k8s Secret with the docker config file holding the necessary registry credentials.
```
yuvipanda, line 68:
```suggestion
The `build pods` are created as a result of an image build request, and they must run on the same node as the builder pods to make use of the docker daemon. These pods mount a k8s Secret with the docker config file holding the necessary registry credentials so they can push to the container registry.
```

docs/howto/features/imagebuilding.md
```diff
@@ -0,0 +1,153 @@
...
+
+### How to set it up
+
+1. Setup a registry and safely store the credentials to push to it
+
+  - If the hub will be deployed on a GCP cluster, follow the steps 4 to 8 from the [binderhub-service installation guide](https://github.com/2i2c-org/binderhub-service?tab=readme-ov-file#installation). When you're done, you should have a gcr.io registry setup and the appropriate credentials to push images to it.
```
yuvipanda, line 74: In our infrastructure for GCP, we actually have https://github.com/2i2c-org/infrastructure/blob/master/terraform/gcp/registry.tf in terraform that sets this up.

It has to be enabled in terraform [like this](https://github.com/2i2c-org/infrastructure/blob/a18937fbe413428b98515d0b5c713aacb5a0d56b/terraform/gcp/projects/pilot-hubs.tfvars#L83). One entry per registry we want to have.

And then, you can get the password for the credentials by running `terraform output  registry_sa_keys`. The username is `_json_key` ([source](https://cloud.google.com/artifact-registry/docs/docker/authentication#json-key)), and the password is the service account JSON that `terraform output` provides.

This is what we have for the imagebuilding-demo hub: https://github.com/2i2c-org/infrastructure/blob/a18937fbe413428b98515d0b5c713aacb5a0d56b/config/clusters/2i2c/imagebuilding-demo.values.yaml#L177.

So I suggest we split this up into its own subsection on 'setting up the image registry'. *That* can have two subsections - one for GCP (with terraform), and one for everything else (with quay.io). As we add native support (via terraform) for other cloud providers, we can add those here.

docs/howto/features/imagebuilding.md
```diff
@@ -0,0 +1,153 @@
...
+      # registry server address like https://quay.io
+      server: <server_address>
+      # robot account namer or "_json_key" if using grc.io
+      username: <account_name>
+  ```
+
+3. Sops-encrypt and store the password of the registry account, in the `enc-<hub>.secret.values.yaml` file.
+
+  ```yaml
+  buildPodsRegistryCredentials:
```
yuvipanda, line 135: Let's explicitly specify this is under `binderhub-service`.

yuvipanda (changes requested):
This is awesome work, @GeorgianaElena! Some suggested changes, but overall this is great.

docs/howto/features/imagebuilding.md
GeorgianaElena: Thanks @sgibson91! I've opened https://github.com/2i2c-org/binderhub-service/pull/90 added this info bit there in the explanation section and took advantage to move the installation guide from the readme to the docs too.

GeorgianaElena:
Thank you @yuvipanda and @sgibson91 for the review. I think I've addressed all of the feedback 🚀

yuvipanda (approved):
Great work, @GeorgianaElena!

PR #3752, opened 2024-02-29
Cleanup unused bits in the 2i2c shared cluster

yuvipanda:
- Jackeddy has been decomissioned for a while
- The agu binder was decomissioned but the associated GAR was not
- The pilot hubs artifact registry has never been used

I saw these while providing feedback on https://github.com/2i2c-org/infrastructure/pull/3749


consideRatio (approved):

yuvipanda:
Applied! Thanks @consideRatio

PR #3766, opened 2024-03-05
[Cloudbank] El Camino - 3GB memory limit

sean-morris:


config/clusters/cloudbank/ccsf.values.yaml
```diff
@@ -10,7 +10,14 @@ jupyterhub:
       # https://2i2c.freshdesk.com/a/tickets/567
       guarantee: 384M
       limit: 1.5G
+    image:
```
yuvipanda, line 13: this seems accidental?

config/clusters/cloudbank/ccsf.values.yaml
```diff
@@ -10,7 +10,14 @@ jupyterhub:
...
+      repository: "your-registry/your-image"
+      tag: "latest"
   custom:
+    singleuserAdmin:
+      image:
+        repository: "your-registry/your-image"
+        tag: "latest"
```
consideRatio, line 20:
```suggestion
  custom:
```

PR #3782, opened 2024-03-06
[2i2c-uk : lis] Add a warning message for users selecting the large instance

GeorgianaElena:
For https://2i2c.freshdesk.com/a/tickets/1348

Had to hack into how kubespawner renders the description text to be able to get the warning a new line  😅 Not sure if there's a better way to do it.

https://github.com/jupyterhub/kubespawner/blob/f99b98fc731517129ebfb7ef3fee293f28d75a75/kubespawner/templates/form.html#L16

### DIFF
--- a/config/clusters/2i2c-uk/lis.values.yaml
+++ b/config/clusters/2i2c-uk/lis.values.yaml
@@ -63,7 +63,10 @@ jupyterhub:
           mem_limit: 2G
           mem_guarantee: 512M
       - display_name: Large Instance
-        description: "Provides 1G - 4G of RAM"
+        # Requested in https://2i2c.freshdesk.com/a/tickets/1348
+        description: |
+          Provides 1G - 4G of RAM</p>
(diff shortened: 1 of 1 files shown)


config/clusters/2i2c-uk/lis.values.yaml
```diff
@@ -63,7 +63,10 @@ jupyterhub:
           mem_limit: 2G
           mem_guarantee: 512M
       - display_name: Large Instance
-        description: "Provides 1G - 4G of RAM"
+        # Requested in https://2i2c.freshdesk.com/a/tickets/1348
+        description: |
```
consideRatio, line 67: I'm not sure, it didn't seem to work by adding a </br> inside the <p> tag. I figure maybe this will lead to kubespawner to use something else than <p> to wrap the description text? I'm not sure - for now, lets inline the comment about this I figure.

```suggestion
        # FIXME: The </p><p> is a hack for kubespawner to render a new line, maybe there are better ways to do this?
        description: |
```

consideRatio (approved):

Issue #3743, opened 2024-02-27
[EVENT] Openscapes, large single day event, March 14th


sgibson91:
### The link towards the Freshdesk ticket this event was reported

https://2i2c.freshdesk.com/a/tickets/1344

### The GitHub handle or name of the community representative

Brianna Lind, blind@contractor.usgs.gov

### The date when the event will start

March 14, 2024

### The date when the event will end

March 14, 2024

### What hours of the day will participants be active? (e.g., 5am - 5pm US/Pacific)

12:00pm PST (https://arewemeetingyet.com/Los%20Angeles/2024-03-14/12:00)

### Are we three weeks before the start date of the event?

- [X] Yes (it was when the ticket was opened, but not at the time I've transferred it to an issue)
- [ ] No

### Number of attendees

230, expecting 16GB usage each, maybe more

### Make sure to add the event into the calendar

- [ ] Done

### Does the hub already exist?

- [X] Yes
- [ ] No

### The URL of the hub that will be used for the event

https://openscapes.2i2c.cloud

### Will this hub be decommissioned after the event is over?

- [ ] Yes
- [X] No

### Was all the info filled in above?

- [ ] Yes
- [ ] No

### Quotas from the cloud provider are high-enough to handle expected usage?

- [ ] Yes
- [ ] No

sgibson91:
From the freshdesk ticket:

> - Date: March 14, 2024
> - Time: 12:00 pm PST
> - Language: Python
> - No. participants: 230
> - Expected usage: 16 GB (per/participant)..... possibly slightly larger
> 
> Do you think there is anything specific that needs to be done in order to facilitate this? We'd also like to request that during the workshop, we can have files in the shared folder be read only on the hub.

sgibson91:
Assignee can see if this guide is helpful: https://infrastructure.2i2c.org/howto/prepare-for-events/event-prep/#pre-warm-the-hub-to-reduce-wait-times

consideRatio:
I'm not completing this or assigning myself, but note:

- Their resource allocation choices have requests == limits, making them not share ram - safe but expensive.
- The resource allocation choices provides the following choices, none of them map directly to "16 GB" or "slightly more".
  ```
                  mem_14_8:
                    display_name: 14.8 GB RAM, upto 3.7 CPUs
                    kubespawner_override:
                      mem_guarantee: 15941615616
                      mem_limit: 15941615616
                      cpu_guarantee: 1.875
[...]

PR #3791, opened 2024-03-12
Fix 29GB RAM profile option not starting in openscapes

yuvipanda:
Looks like the overhead of resources in AWS nodes has changed since we last set these, causing the largest pod on r5.xlarge to not fit anymore. We should eventually dynamically generate these
\- but in the meantime, I've just regenerated these numbers using our existing deployer command so mem_29_7 starts again.


yuvipanda:
Came up during discussion of https://github.com/2i2c-org/infrastructure/issues/3743

sgibson91 (approved):

PR #3792, opened 2024-03-12
Openscapes: Change the node selector instance type to r5.16xlarge for all profile options

sgibson91:
- addresses: https://github.com/2i2c-org/infrastructure/issues/3743

As discussed in https://github.com/2i2c-org/infrastructure/issues/3743#issuecomment-1989387520


yuvipanda (approved):
I think this is good to go! I suggest we merge and test at least 24-48h before start of event (we can encode this info elsewhere)

Issue #3434, opened 2023-11-17
Stop unconditionally setting `cmd` to `jupyterhub-singleuser`


yuvipanda:
in `basehub/values.yaml`, we set `cmd` to `jupyterhub-singleuser`, to account for images that may not have that correctly set up. However, this means that the `ENTRYPOINT` set up in the docker image is *not* run, causing issues with logging when using repo2docker (https://github.com/2i2c-org/infrastructure/pull/3433) as well as startup when using desktop-proxy.

We should evaluate if this is actually still needed, and if not, unset `jupyterhub-singleuser` as `cmd` in `basehub/values.yaml`

consideRatio:
https://github.com/jupyterhub/kubespawner/issues/493#issuecomment-1148712835 has related discussion summarized.

I think the role of `jupyterhub-singleuser` has changed notably over time, where a key point is to configure `notebook` and `jupyter_server` to be jupyterhub aware, but that such awareness in at least `jupyter_server` of modern versions is handled by environment variables and a `jupyter_server` extension.

So, unsetting it is probably fine for all images that use `jupyter_server` of a modern enough version, but probably not for those using `notebook` as a server still.

consideRatio:
> in `basehub/values.yaml`, we set `cmd` to `jupyterhub-singleuser`, to account for images that may not have that correctly set up. However, this means that the `ENTRYPOINT` set up in the docker image is _not_ run, causing issues with logging when using repo2docker (#3433) as well as startup when using desktop-proxy.

Currently z2jh's default is to set `singleuser.cmd` (`KubeSpawner.cmd`) to `jupyterhub-singleuser`, we don't do it in basehub. Setting it overrides docker images' `CMD`, but won't override docker images's ENTRYPOINT. I think we are fine with the current state of things and no action is required.

yuvipanda:
fwiw we still definitely do it in basehub: https://github.com/2i2c-org/infrastructure/blob/a6e5a196b76719b20afaa4e6a551733db3b5863f/helm-charts/basehub/values.yaml#L282

yuvipanda:
@consideRatio can this be closed now? I remember you took some actions here, but I'm not sure what.

consideRatio:
I don't advocate any action related to this issue, so if you think its OK to close lets close!

Logs etc is resolved, but we have via z2jh set singleuser.cmd explicitly, which I think we should.

yuvipanda:
[...]

PR #3797, opened 2024-03-14
nasa and hhmi hubs: update QGIS image

consideRatio:
Bumps to the image built via https://github.com/2i2c-org/nasa-qgis-image/pull/10.

- Fixes https://github.com/2i2c-org/meta/issues/931
- Removes a workaround to set `singleuser.cmd: null`
  This can be done as the docker image, based on quay.io/jupyter/minimal-notebook, since 2024-01-22 declares an ENTRYPOINT that includes executing the `start.sh` script.
  
  A Dockerfile's `ENTRYPOINT` / `CMD` can be overridden by specifying `command` / `args` in a k8s pod's container specification. The z2jh charts `singleuser.cmd` maps to `KubeSpawner.cmd`, and that configuration together with `KubeSpawner.args` is combined to set k8s pod's container `args`. This means KubeSpawner doesn't touch `ENTRYPOINT` aka. `command` in k8s specifications, making the `start.sh` script still be executed, while when it was set in jupyter/docker-stacks images `CMD` before 2024-01-22 it would get overridden.

### DIFF
--- a/config/clusters/hhmi/common.values.yaml
+++ b/config/clusters/hhmi/common.values.yaml
@@ -174,15 +174,10 @@ basehub:
                   display_name: Linux desktop
                   slug: desktop
                   kubespawner_override:
-                    # Explicitly unset this - we set this to 'jupyterhub-singleuser'
-                    # in basehub/values.yaml. We instead want to leave this unset,
-                    # so the default command for the docker image is used instead.
-                    # This is required for .desktop files to show up correctly.
-                    cmd: null
                     # Launch people directly into the Linux desktop when they start
                     default_url: /desktop
                     # Built from https://github.com/jupyterhub/jupyter-remote-desktop-proxy/pull/51
-                    image: "quay.io/jupyter-remote-desktop-proxy/qgis:2023-09-27"
+                    image: quay.io/2i2c/nasa-qgis-image:babb0fa15dfa
     hub:
       allowNamedServers: true
       config:
--- a/config/clusters/nasa-esdis/common.values.yaml
+++ b/config/clusters/nasa-esdis/common.values.yaml
@@ -133,13 +133,8 @@ jupyterhub:
       - display_name: QGIS
         description: QGIS environment
         kubespawner_override:
-          # Explicitly unset this - we set this to 'jupyterhub-singleuser'
-          # in basehub/values.yaml. We instead want to leave this unset,
-          # so the default command for the docker image is used instead.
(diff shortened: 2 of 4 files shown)


yuvipanda:
@consideRatio can you help me understand why `cmd: null` is not needed anymore? It's primarily set for https://github.com/2i2c-org/infrastructure/issues/3434, and that particular image requires `start.sh` to run (due to https://github.com/2i2c-org/nasa-qgis-image/blob/babb0fa15dfa5149016a019be595dffc483b08f8/Dockerfile#L14).

yuvipanda:
If this is based on https://github.com/jupyter/docker-stacks/pull/2087#event-11551562606, my understanding is that that shouldn't affect us here because Dockerfile's `ENTRYPOINT` is actually Kubernetes' `cmd` and Dockerfile's `CMD` is actually Kubernetes `args`.

consideRatio:
Hehe yes but what is kubespawners cmd? :) kubespawners cmd and kubespawners args combines into setting k8s args on the pod i think / from mobile

consideRatio:
I've updated the PR description to clarify this further @yuvipanda, and I've also updated the image tags. I consider this PR ready to be merged from a technical standpoint, but we are interfering with the communities user environment which we have no procedure around.

Do you think its OK to bump these images @yuvipanda?

consideRatio:
Verified my understanding that kubespawner `cmd` maps to k8s `args`:

```
kubectl get pod -n redacted jupyter-redacted -o yaml | grep -E -A5 "command|args"

  - args:
    - jupyterhub-singleuser
    env:
    - name: CPU_GUARANTEE
      value: "0.05"
    - name: JPY_API_TOKEN
--
  - command:
    - sh
    - -c
    - id && chown 1000:1000 /home/jovyan /home/jovyan/shared && ls -lhd /home/jovyan
    image: busybox:1.36.1
    imagePullPolicy: IfNotPresent
--
  - command:
    - iptables
    - --append
    - OUTPUT
    - --protocol
    - tcp
```

yuvipanda:
Ah, somewhere somehow my understanding needs to be rethought, but we don't need to wait for that to merge this. I tested this on staging and it works. Thanks for thinking this through, @consideRatio.

consideRatio:
Thank you @yuvipanda!!

I checked that repo2docker based images also doesn't have an issue with setting `singleuser.cmd` which led to #3807.

PR #3798, opened 2024-03-14
nasa-veda: add s3 bucket access to pangeo-forge-veda-output

ranchodeluxe:
This PR adds a bucket of pangeo-forge outputs that we want to be able to play with on the veda hub


consideRatio (approved):
LGTM, applied via terraform!

PR #3800, opened 2024-03-14
hhmi/spyglass: set default URL to pull spyglass demo notebooks

jmunroe:


jmunroe:
This change modifies the defaultURL so that spyglass.hhmi.2i2c.cloud automatically does a nbgitpuller link. I think I am ok to self-merge this change.

PR #3805, opened 2024-03-15
basehub: don't declare singleuser.cmd - its z2jh's default

consideRatio:
Its a breaking change for z2jh to change this, so if it changes we'll be
well aware about it from changelogs when upgrading.

I mainly want to remove the now outdated comment, but I also appreciate
not re-specifying defaults in general as that makes it easier to
overview and trust whats declared in basehub's default values are actual
changes from the default.


consideRatio:
Outdated because we now use the vuln-patch script

Issue #3801, opened 2024-03-14
For all hubs using `rocker/binder` images, declare `KubeSpawner.image_pull_policy = "Always"`


consideRatio:
Even when we specify `rocker/binder:4.3.3`, its an image that gets updated over time. Let's always start the latest variant if it has been rebuilt instead of starting old/new depending if the specific k8s node had the image cached or not.

PR #3813, opened 2024-03-19
Always check for latest rocker/binder image as tags get rebuilt

consideRatio:
- Fixes https://github.com/2i2c-org/infrastructure/issues/3801

These images have tags that get updated over time, even 4.3.3 gets updated over time. Due to that, we should check against the container registry at all times if there is a new version to avoid old versions being used on long-running nodes. If that happens - we end up with multiple versions of the same image tag running, for example missing out on a security patch.

I verified this works in 2i2c-aws-us/showcase.


consideRatio:
I'll go for a merge to unblock a security PR

yuvipanda:
Good work @consideRatio

PR #3816, opened 2024-03-20
[Support: Earthscope] Remove authentication special casing

GeorgianaElena:
According to https://2i2c.freshdesk.com/a/tickets/1416, earthscope's Auth0 integration with CILogon was updated and no longer returns `sub` claims like `oauth|CILogon|https://cilogon.com/serverA/users/BCDEF` but `oauth2|CILogon|ABCDEF`.

This means we can get rid of this special casing.

### DIFF
--- a/config/clusters/earthscope/common.values.yaml
+++ b/config/clusters/earthscope/common.values.yaml
@@ -22,7 +22,7 @@ basehub:
         templateVars:
           org:
             url: https://www.earthscope.org/
-            logo_url: https://drive.google.com/uc?export=view&id=1UUStqv7PBcxiIkzECUFKIdQKKIU8mXeb
+            logo_url: https://github.com/2i2c-org/infrastructure/assets/7579677/589da909-86c2-4440-a42b-e3f1b59f49d5
           designed_by:
             name: "2i2c"
             url: https://2i2c.org
--- a/config/clusters/earthscope/common.values.yaml
+++ b/config/clusters/earthscope/common.values.yaml
@@ -82,18 +82,6 @@ basehub:
                     url = f"{url}?{params}"
                 return url
 
-            async def authenticate(self, *args, **kwargs):
-              auth_model = await super().authenticate(*args, **kwargs)
-              username = auth_model["name"]
-              # This is required until https://github.com/jupyterhub/oauthenticator/pull/717
-              # gets merged, can be removed after that.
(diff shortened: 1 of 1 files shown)


consideRatio (approved):
Thank you @GeorgianaElena!

For a change like this, it would be good to followup with a verification that it works to login still and without getting a new username.

GeorgianaElena:
@consideRatio, I did a manual deploy on the staging hub and tried to login, but got a 403. However, it does look that the username gets built ok from the logs:

```bash
[I 2024-03-20 08:57:36.209 JupyterHub <string>:56] Denying access to user oauth2|cilogon|a51530946 - scopes {'geolab'} were not granted, only ['openid', 'offline_access'] were granted
[W 2024-03-20 08:57:36.209 JupyterHub auth:533] User 'oauth2|cilogon|a51530946' not allowed.
[W 2024-03-20 08:57:36.209 JupyterHub base:843] Failed login for unknown user
```

GeorgianaElena:
I don't think I ever logged into the hub and it required me to register with the earthscope consortium. Curious if you already registered in the past and were successful to login if it still works now?  Link to the staging hub https://staging.geolab.earthscope.cloud

consideRatio:
I've not logged in, and I struggled to conclude this is a change that makes users loose access to their old home directories by possibly getting a new username.

I figure they can be asked to confirm its okay as it is now or if they want to request an intervention to migrate home directory storage etc?

I posted some notes on inspecting the database without having personal admin access on slack.

consideRatio:
I suggest merge -> communication via freshdesk (-> maybe a followup action, hopefully not)

GeorgianaElena:
>  and I struggled to conclude this is a change that makes users loose access to their old home directories by possibly getting a new username.

Hmm, looking at the database users that you shared on slack, and revisiting the code that this PR removes, it does look like the users will loose access to their old home dirs since they're usernames were `oauth2|cilogon|server<some-id>|<some-number>` and now they'll be `oauth2|cilogon|<some-id><some-number>`.

I think I'll double check first on freshdesk, then merge the PR

GeorgianaElena:
Got approval that it's ok to merge this PR because hub access hasn't been rolled out yet.

Also, the logo from google drive isn't rendering per @consideRatio's observation, so I'll update it here and reference it from this comment:
gage_sage_stacked_color_small

PR #3823, opened 2024-03-21
update image tag for nasa-singleuser image

batpad:
Refs https://github.com/NASA-IMPACT/veda-jh-environments/issues/46

Updates the nasa-veda-singleuser image to an image that uses an updated version of the `pangeo-notebook` base image (now using `pangeo/pangeo-notebook:2024.03.13`)

This update is needed to address https://github.com/jupyterhub/jupyter-server-proxy/security/advisories/GHSA-w3vc-fx9p-wp4v

cc @yuvipanda @ranchodeluxe @wildintellect @abarciauskas-bgse

### DIFF
--- a/config/clusters/nasa-ghg/common.values.yaml
+++ b/config/clusters/nasa-ghg/common.values.yaml
@@ -66,7 +66,7 @@ basehub:
           description: Pangeo based notebook with a Python environment
           default: true
           kubespawner_override:
-            image: public.ecr.aws/nasa-veda/nasa-veda-singleuser:5068290376e8c3151d97a36ae6485bb7ff79650b94aecc93ffb2ea1b42d76460
+            image: public.ecr.aws/nasa-veda/nasa-veda-singleuser:2023-03-20
           profile_options: &profile_options
             resource_allocation: &profile_options_resource_allocation
               display_name: Resource Allocation
--- a/config/clusters/nasa-ghg/common.values.yaml
+++ b/config/clusters/nasa-ghg/common.values.yaml
@@ -66,7 +66,7 @@ basehub:
           description: Pangeo based notebook with a Python environment
           default: true
           kubespawner_override:
-            image: public.ecr.aws/nasa-veda/nasa-veda-singleuser:2023-03-20
+            image: public.ecr.aws/nasa-veda/nasa-veda-singleuser:2024-03-20
           profile_options: &profile_options
             resource_allocation: &profile_options_resource_allocation
               display_name: Resource Allocation
--- a/config/clusters/nasa-veda/common.values.yaml
+++ b/config/clusters/nasa-veda/common.values.yaml
@@ -68,7 +68,7 @@ basehub:
           description: Pangeo based notebook with a Python environment
           default: true
           kubespawner_override:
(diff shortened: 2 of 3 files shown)


config/clusters/nasa-veda/common.values.yaml
```diff
@@ -68,7 +68,7 @@ basehub:
           description: Pangeo based notebook with a Python environment
           default: true
           kubespawner_override:
-            image: public.ecr.aws/nasa-veda/nasa-veda-singleuser:2024-03-07
```
yuvipanda, line 71: You should update the config for the GHG and ESDIS hubs as well here.

yuvipanda (changes requested):
Thanks for opening the pR!

config/clusters/nasa-veda/common.values.yaml
slesaad: let's wait for the esdis hub until the end of the webinar tomorrow Mar 21st at 1 pm CST

batpad: @yuvipanda updated the PR to include the GHG and ESDIS hubs. Let's wait for a review from @ranchodeluxe as well as Mar 21st 1pm CST as per @slesaad's comment :)

slesaad: we're good to go with this!

config/clusters/nasa-ghg/common.values.yaml
```diff
@@ -66,7 +66,7 @@ basehub:
           description: Pangeo based notebook with a Python environment
           default: true
           kubespawner_override:
-            image: public.ecr.aws/nasa-veda/nasa-veda-singleuser:5068290376e8c3151d97a36ae6485bb7ff79650b94aecc93ffb2ea1b42d76460
+            image: public.ecr.aws/nasa-veda/nasa-veda-singleuser:2023-03-20
```
wildintellect, line 69: is it supposed to be 2024-03-20?

wildintellect (commented):
2 say 2023-03-20 and one says 2024-03-20 <- please check and fix

batpad:
@wildintellect oops! Thank you so so much for the catch there 🙏 - fixed the two wrong dates on the tags and pushed.

@wildintellect I did a very basic test of the image working on the staging VEDA hub - do you know if there's a process / more thorough tests we should be running before merging?

config/clusters/nasa-ghg/common.values.yaml
yuvipanda: This probably also needs to be updated, @batpad

yuvipanda (changes requested):

batpad:
Oops, must remember to actually `git add` and `git push` after making a change :-/ - thanks for the catch @yuvipanda and apologies all for all the typos. This should be actually done now.

consideRatio (approved):

wildintellect:
@batpad we do not have a testing framework at the moment but it's in the works, don't let it hold up this PR. @jsignell can comment if there's anything we can do now.

Looks like you now need to:
- [ ] update from `master`
- [ ] request re-review from @yuvipanda

sgibson91:
I actually think we can just merge this - I can override the change requested blockage and Erik has already approved

PR #3825, opened 2024-03-21
hhmi: update image tags

jmunroe:
Updating hhmi-spyglass-image tag based on request from LorenFrank lab via Slack.


jmunroe:
Image tag change only. Self-merging.

PR #3826, opened 2024-03-21
Allow ESDIS hub access to S3

yuvipanda:
We were setting up the name of the service account in daskhub/values.yaml, so hubs that were set up without dask don't actually have access to the IAM roles set up for them!


yuvipanda:
There's a workshop running on this hub tomorrow, so I'm going to self-merge.

Issue #3780, opened 2024-03-06
Tune AWS management accounts default quota requests ?


consideRatio:
This is a followup to https://2i2c.freshdesk.com/a/tickets/1368.

After Georgiana setup a new cluster under the AWS management account / organization account, we received an email:

```
While processing your quota increase request, we noticed your requested EC2 vCPU amount was less than or equal to the current quota value in your requested region.

Your current quota is: 64 for EC2 Instances / All G and VT instances [US West (Oregon)]
Your requested quota was: 16 for EC2 Instances / All G and VT instances [US West (Oregon)]
```

We think this happens because:
- we have automation to request quota for new accounts under the management account
- maybe AWS have defaulted new accounts to get 64 in quota instead of 0 as before, and now our request lowers it instead of raises it?

## Action points

- [ ] Identify our aws management accounts automated quota limit requests
- [ ] Look for AWS announcing a change in the default quota for new project, and if its increased from 0 to 64  in all regions or similar we should then just drop our automated rule, otherwise perhaps learn to handle such followup emails.

yuvipanda:
Our default quota request template (you can see them in https://us-east-1.console.aws.amazon.com/servicequotas/home/template?region=us-east-1#) asked for 16 CPUs for use with instance types that have GPUs, but only in the us-west-2. It looks like we now have enough history with AWS that our accounts are getting 64 CPUs for instance types with GPUs, *across all regions*. This is helpful, because as in https://github.com/2i2c-org/infrastructure/issues/3827#issuecomment-2015444460, we are also spinning up hubs in places that aren't us-west-2 now.

I've removed the request template, and I'm also going to work on reflecting this in our documentation.

yuvipanda:
Fixed by https://github.com/2i2c-org/infrastructure/pull/3836

PR #3836, opened 2024-03-23
Document that we don't need to ask for extra quota on AWS

yuvipanda:
Fixes https://github.com/2i2c-org/infrastructure/issues/3780

### DIFF
--- a/docs/hub-deployment-guide/cloud-accounts/new-aws-account.md
+++ b/docs/hub-deployment-guide/cloud-accounts/new-aws-account.md
@@ -80,7 +80,7 @@ of these are all 'Total CPUs' and hence larger nodes consume more quota.
   High Memory instances`) is requested specifically to be able to use GPU
   powered machines or machines with high amounts of RAM memory.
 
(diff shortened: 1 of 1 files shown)


docs/hub-deployment-guide/cloud-accounts/new-aws-account.md
```diff
@@ -84,7 +80,8 @@ of these are all 'Total CPUs' and hence larger nodes consume more quota.
   High Memory instances`) is requested specifically to be able to use GPU
   powered machines or machines with high amounts of RAM memory.
 
-  Our custom tailored hubs will require an increase in this quota.
+  By default, AWS grants us 640 quota here for GPU instances and 448 for
```
consideRatio, line 83: 640? Is that a typo that should be 64? I recall seeing this to be 64, but i may very well be mistaken

yuvipanda: haha, it is a typo! The other two we do get 640. I'll fix.

consideRatio (approved):
Nice!!

Issue #3273, opened 2023-10-16
Get rid of eksctl produced `cloud-user-sa`


yuvipanda:
The following three hubs use a `cloud-user-sa` service account provisioned by eksctl to give hub users *full* S3 access:

- [x] uwhackweeks
- [x] carbonplan
- [ ] openscapes

This predates our more modern terraform based service accounts. We should get rid of these eksctl generated accounts and bring these hubs in line with the rest.

### Stop using the `cloud-user-sa` service account

- [ ] Stop using `cloud-user-sa` serviceaccount by removing `jupyterhub.singleuser.serviceAccountName` under `config/clusters/openscapes/common.yaml`. Make a PR, get this deployed. 
- [ ] Test to make sure you can still access `$SCRATCH_BUCKET` S3 bucket, by running `aws s3 ls $SCRATCH_BUCKET` in a jupyterlab terminal in the hub. This should not error out.

### Remove the IAM service account from AWS

- [ ] Render the `.jsonnet` file into a `.yaml` file with the `jsonnet` command in https://infrastructure.2i2c.org/hub-deployment-guide/new-cluster/aws/#create-and-render-an-eksctl-config-file
- [ ] Delete the iamserviceaccount with `eksctl delete iamserviceaccount -f openscapes.eksctlyaml --name cloud-user-rsa`
- [ ] Repeat the test from earlier to make sure that `aws s3 ls $SCRATCH_BUCKET` still works inside a terminal in the jupyterhub.
- [ ] Remove `namespaces` variable in `eksctl/openscapes.jsonnet`
- [ ] Remove `iam.serviceAccounts` in the main cluster definition in `eksctl/opencsapes.jsonnet`
- [ ] Make this into a PR, get it merged

### Celebrate

This is a celebration of the fact that we've been serving openscapes so long that they're using deprecated features! When done, we can close this issue :)

consideRatio:
uwhacksweeks is removed already via https://github.com/2i2c-org/infrastructure/issues/2173, I opened #3291 to finalize the cleanup.

### Related code

https://github.com/2i2c-org/infrastructure/blob/682ef1f38cd6eb944761e65824c6d4d58688600e/eksctl/carbonplan.jsonnet#L70-L82

### Issue scope question

@yuvipanda is the scope narrowed to replacing how the permissions are granted? Currently carbonplan and openscapes has `arn:aws:iam::aws:policy/AmazonS3FullAccess` which includes creation of new buckets etc.

[...]

Issue #3694, opened 2024-02-03
[Support] Setup persistent S3 bucket for openscapes


yuvipanda:
### The Freshdesk ticket link

https://2i2c.freshdesk.com/a/tickets/1278

### Ticket request type

Feature Request

### Ticket impact

🟨 Medium

### Short ticket description

This also came out of https://github.com/2i2c-org/infrastructure/issues/3614

Setup a persistent S3 bucket for use by users on this hub.

### (Optional) Investigation results

https://github.com/2i2c-org/infrastructure/issues/3273 is related, and may need to be resolved first.

yuvipanda:
This needs https://github.com/2i2c-org/infrastructure/issues/3273 first.

yuvipanda:
We don't actually need #3273 here I think - we can just follow the current instructions as they are.

sgibson91:
PR https://github.com/2i2c-org/infrastructure/pull/3832 is ready for review and has been deployed to the staging hub for testing

PR #3842, opened 2024-03-25
Remove use of cloud-user-sa from openscapes hub

yuvipanda:
Without this, https://github.com/2i2c-org/infrastructure/issues/3694 does not quite take, as the `user-sa` service account is not attached to the user pod.

This unifies all our infrastructure so nobody is using service accounts created by `eksctl`. Since openscapes is *just about* to start using S3, this is a good time to make sure it's set up in a standard way.

Fixes https://github.com/2i2c-org/infrastructure/issues/3273

### DIFF
--- a/config/clusters/openscapes/common.values.yaml
+++ b/config/clusters/openscapes/common.values.yaml
@@ -35,7 +35,6 @@ basehub:
             name: NASA Openscapes
             url: https://nasa-openscapes.github.io/
     singleuser:
-      serviceAccountName: cloud-user-sa
       defaultUrl: /lab
       extraEnv:
         GH_SCOPED_CREDS_CLIENT_ID: "Iv1.6981e043b45f042f"
--- a/eksctl/openscapes.jsonnet
+++ b/eksctl/openscapes.jsonnet
@@ -20,11 +20,6 @@ local clusterRegion = "us-west-2";
 local masterAzs = ["us-west-2a", "us-west-2b", "us-west-2c"];
 local nodeAz = "us-west-2b";
 
-// List of namespaces where we have hubs deployed
-// Each will get a ServiceAccount that will get credentials to talk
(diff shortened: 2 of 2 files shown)


yuvipanda:
Without this, the openscapes hub is not using the service account created for it:

```
(notebook) jovyan@jupyter-yuvipanda:~$ aws sts get-caller-identity
{
    "UserId": "AROA3M4ZSJJNXDIFKYZKA:botocore-session-1711408989",
    "Account": "783616723547",
    "Arn": "arn:aws:sts::783616723547:assumed-role/eksctl-openscapeshub-addon-iamserviceaccount-Role1-VCQZHI6KZIQ9/botocore-session-1711408989"
}
```

yuvipanda:
After merging, this still requires running an `eksctl delete iamserviceaccount` to be run. I'll do that locally.

Before working on it, I considered if I should just write documentation instead, or try to pair with someone. I chose to just do this instead as it leads to this concept of 'eksctl created service account' just disappearing from our infrastructure, thus reducing overall cognitive load.

consideRatio (approved):

consideRatio:
Since I'm very happy about cleanup like this, I've performed a small dance to celebrate!

Thank you @yuvipanda!!

yuvipanda:
Thanks @consideRatio :)

I've deleted the staging service account with:

```
eksctl delete iamserviceaccount --namespace staging --name cloud-user-sa --cluster openscapeshub --region=us-west-2
```

There are currently users on the prod hub, and I'm not sure that deleting an in-use service account is even possible. So I'm going to wait that out for a bit, as new users will use the `user-sa` (than `cloud-user-sa`). Then I can run the same thing for the prod hub.

yuvipanda:
I ran the command for prod too, and it's all gone now.

PR #3843, opened 2024-03-26
Document how to test if buckets are fully set up

yuvipanda:


docs/howto/features/buckets.md
```diff
@@ -83,6 +83,75 @@ on why users want this!
 4. Get this change deployed, and users should now be able to use the buckets!
    Currently running users might have to restart their pods for the change to take effect.
 
+### Testing access to buckets
+
+Once bucket access has been set up, we should test to make sure users can write
+to and read from it.
```
consideRatio, line 89: Maybe also mention that we expect certain environment variables to be defined, matching a pattern like `<NAME>_BUCKET`. I'm thinking a first step could be `env | grep _BUCKET` or similar to help overview those.

consideRatio (approved):
I just had an improvement suggestion - this looks great!

docs/howto/features/buckets.md
yuvipanda: Good idea @consideRatio. I've added that as well.

PR #3849, opened 2024-03-26
Add inline comments for consistency about cloud infra accounts etc

consideRatio:
A small PR for consistency of information available in cluster.yaml - even though its via comments currently.


consideRatio:
Self merging as trivial

Issue #3828, opened 2024-03-21
[New Hub] LINC (MIT Brain)


yuvipanda:
Copied over from https://github.com/2i2c-org/meta/issues/913

## Process Note

I'm using this as a way to try to rejig our new hub request process. See https://github.com/2i2c-org/meta/issues/897 (particularly https://github.com/2i2c-org/meta/issues/897#issuecomment-2010984904) for more information.

https://miro.com/app/board/uXjVNjUP3iQ=/, describes the various 'phases' of new hub turn-up. Each phase will be marked as "READY" or "NOT READY" when all information needed for it is available. Each section should also link to an appropriate runbook.

There will be customizations after this is all set up, but this is pathway towards a *standardized hub turn up*.

## Phase 1: Account setup (**READY**)

This is applicable for cases where this is a dedicated cluster. The following table lists the information before this phase can start.

| Question | Answer |
| - | - |
| Cloud Provider | AWS |
| Will 2i2c pay for cloud costs? | Yes |
| Name of cloud account | `linc` |

Appropriate runbook: https://infrastructure.2i2c.org/hub-deployment-guide/cloud-accounts/new-aws-account/

## Phase 2: Cluster setup (**READY**)

This assumes all engineers have access to this new account, and will be able to set up the cluster + support, without any new hubs being set up.

| Question | Answer | 
| - | - |
| Region / Zone of the cluster | us-east-1 | 
| Name of cluster | linc | 
| Is GPU required? | yes | 

Appropriate runbooks:
- New cluster setup: https://infrastructure.2i2c.org/hub-deployment-guide/new-cluster/aws/
- Support deployment: https://infrastructure.2i2c.org/hub-deployment-guide/deploy-support/

## Phase 3 : Hub setup (**READY**)

There's going to be a number of hubs, and this starts specifying them.

#### Phase 3.1: Initial setup 
| Question | Answer | Notes |
| - | - | - |
| Name of the hub | staging | |
| Dask gateway? | no | |
| Splash Image | https://github.com/lincbrain/linc-artwork/blob/main/linc.logo.color+white.notext.png?raw=true | |
| URL | https://lincbrain.org/ | |

#### Phase 3.2: Authentication

| Question | Answer |
| - | - |
| Authentication Mechanism | GitHub (via GitHubOAuthenticator) |
| Org based access? | No |
| Admin Users | `@kabilar, @aaronkanzer, @asmacdo, @satra` | 

#### Phase 3.3: Object storage access

[...]

PR #3854, opened 2024-03-26
linc: add cluster, and deploy support chart and staging hub

consideRatio:
Resolves part 1 and 2 of https://github.com/2i2c-org/infrastructure/issues/3828, and parts of part 3 which was about deploying a staging hub. Details about that hub isn't locked in place yet, so I did parts.

## Unrelated test failure

Note that the test failure is unrelated, which stems from me removing autoformat removed whitespace in a file inside `config/clusters/templates`. I tried to patch a job to not trigger incorrectly on those files but failed. Opened #3855 related to this.


.github/workflows/deploy-hubs.yaml
```diff
@@ -180,32 +180,33 @@ jobs:
     #
     # If you are adding a new cluster, please remember to list it here!
     outputs:
-      failure_2i2c: "${{ env.failure_2i2c }}"
+      failure_2i2c-aws-us: "${{ env.failure_2i2c-aws-us }}"
       failure_2i2c-uk: "${{ env.failure_2i2c-uk }}"
+      failure_2i2c: "${{ env.failure_2i2c }}"
```
consideRatio, line 185: I sorted these alphabetically, makes it less likeley that we get merge conflicts compared to appending - but mostly because its easier to read.

GeorgianaElena (approved):

PR #3856, opened 2024-03-26
Move service account specification to basehub

yuvipanda:
Follow-up to https://github.com/2i2c-org/infrastructure/pull/3826 to prevent that from happening again in other hubs. The service account is *created* in basehub, so it should just be set there too.


consideRatio (approved):

consideRatio:
Or wait, doesn't this make misc cleanup also?

consideRatio:
I opened https://github.com/2i2c-org/infrastructure/issues/3864 as a followup to ensure consistent function of things we have setup already. I spotted for example 3/6 hubs in 2i2c cluster without function with buckets setup for them.

Issue #3223, opened 2023-10-04
Improve stability of shared-dirsize-metrics pods


consideRatio:
They are crashing if there is permission issues like described in https://github.com/yuvipanda/prometheus-dirsize-exporter/issues/5

I've seen quite a few pods in various clusters' hub namespaces be in crashloop backoff, so it would be good if we work to make this more robust and ensure that we get all our such pods up and running reliably. If not, it ends up being debugging overhead when its relevant for some other situation. Resolving this ahead of time systematically is far more efficient use of our time.

As an example, the victor's prod hub has a failure here.

PR #3857, opened 2024-03-26
Bump version of prometheus-dirsize-exporter

yuvipanda:
Brings in https://github.com/yuvipanda/prometheus-dirsize-exporter/pull/7

Ref https://github.com/2i2c-org/infrastructure/issues/3223

### DIFF
--- a/config/clusters/gridsst/common.values.yaml
+++ b/config/clusters/gridsst/common.values.yaml
@@ -1,9 +1,6 @@
 basehub:
   nfs:
     enabled: true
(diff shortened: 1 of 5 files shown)


consideRatio (commented):
Lets enable hubs where we disabled the reported due to errors making it crash, searching for dirsizeReporter: should help spot those places, there is a comment next to each disabled reporter also

consideRatio (approved):
I'll keep an eye on for:
- gridsst
- nasa-ghg
- nasa-veda
- ubc-eoas

They are hubs that have had their dirsize reported disabled historically where deployments started failing due to crashing pod.

Issue #3796, opened 2024-03-14
[Request deployment] New Hub: Jupyter Health Hub


colliand:
### The GitHub handle of the community representative

@fperez @colliand @yuvipanda

### Technical support contacts listed

- [ ] Technical contact present

### Hub important dates

- **Target Start Date:** 2024-03-31 (...or as soon as reasonably possible within 2i2c Engineering's sprint process)

### Hub Authentication Type

GitHub (e.g., @mygithubhandle)

### First Hub Administrators

- @colliand 
- @fperez 
- @maryamv
- @ryanlovett

### [GitHub Auth only] How would you like to manage your users?

Allowing members of a specific GitHub organization

### [GitHub Teams Auth only] Profile restriction based on team membership

https://github.com/jupyterhealth

### Hub logo image URL

pending

### Hub logo website URL

pending

### Hub user image GitHub repository

pending

### Hub user image tag and name

pending

### Extra features you would like to enable

- [x] Dedicated Kubernetes cluster
- [ ] Scalable Dask Cluster

### (Optional) Preferred cloud provider

AWS

### (Optional) Preferred cloud region

us-east-2 <---This was updated by @colliand on 2024-03-14.

### (Optional) Billing and Cloud account

None

### Other relevant information to the features above

Let's set this hub on its own billing account so that the project can track and assign cloud costs effectively. 

### Tasks to deploy the hub

- [ ] 1. Deploy information filled in above
- [ ] 2. Engineer who will deploy the hub is assigned
- [ ] 3. If using GitHub Orgs/Teams Auth, Engineer is given Owner rights to the org to set this up.
- [ ] 4. Initial Hub deployment PR <link to PR>
- [ ] 5. Administrators able to log on -> Hub now in steady-state

colliand:
I suggest this be deployed as a dedicated cluster. Thanks @consideRatio.

minrk:
Trying to answer some of these questions, to help me get up to speed on 2i2c process

> technical support contacts listed

Should this be @minrk and/or @ryanlovett ?

> hub logo image url

Don't have one yet, so start with default Jupyter logo?

>  hub logo website url

For now, https://github.com/jupyterhealth

> hub user image github repository

@colliand should we create a public repo on @jupyterhealth just for the user image? I assume we're at least going to install the TCP client package in there once we can get access to it.

> hub user image tag and name

let's start with `quay.io/jupyter/scipy-notebook:2024-03-14` while we wait for the user image repo.

[...]

PR #3859, opened 2024-03-27
Add Jupyter Health Hubs

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/3796


config/clusters/jupyter-health/cluster.yaml
```diff
@@ -0,0 +1,28 @@
+name: jupyter-health
+provider: aws
```
consideRatio, line 2: cluster.yaml and this comment has been the goto place to figureout where a cluster is available - consistent across all current cluster.yaml files

```suggestion
provider: aws # https://2i2c.awsapps.com/start/#/
```

config/clusters/jupyter-health/common.values.yaml
```diff
@@ -0,0 +1,181 @@
...
+                runAsUser: 0
+              volumeMounts:
+                - name: home
+                  mountPath: /home/jovyan
+                  subPath: "{username}"
+                # mounted without readonly attribute here,
+                # so we can chown it appropriately
+                - name: home
+                  mountPath: /home/jovyan/shared
+                  subPath: _shared
```
consideRatio, line 82: We are already providing an init container by default to do this - aren't we?

consideRatio (commented):
I left a comment about a init_container override that I think isn't relevant to specify, but otherwise this looks great.

consideRatio:
Fixed a merge conflict

config/clusters/jupyter-health/common.values.yaml
yuvipanda: Yep! Another reason to move away from copy pasting hubs! This was leftover from copy pasting from nasa-veda

yuvipanda:
CI caught that I hadn't added the `paid_by_us` field. Thanks for working on that, @consideRatio.

Issue #3824, opened 2024-03-21
[New Hub] DANDI (MIT Brain)


yuvipanda:
Copied over from https://github.com/2i2c-org/meta/issues/913

## Process Note

I'm using this as a way to try to rejig our new hub request process. See https://github.com/2i2c-org/meta/issues/897 (particularly https://github.com/2i2c-org/meta/issues/897#issuecomment-2010984904) for more information.

https://miro.com/app/board/uXjVNjUP3iQ=/, describes the various 'phases' of new hub turn-up. Each phase will be marked as "READY" or "NOT READY" when all information needed for it is available. Each section should also link to an appropriate runbook.

There will be customizations after this is all set up, but this is pathway towards a *standardized hub turn up*.

## Phase 1: Account setup (**READY**)

This is applicable for cases where this is a dedicated cluster. The following table lists the information before this phase can start.

| Question | Answer |
| - | - |
| Cloud Provider | AWS |
| Will 2i2c pay for cloud costs? | Yes |
| Name of cloud account | `dandi` |

Appropriate runbook: https://infrastructure.2i2c.org/hub-deployment-guide/cloud-accounts/new-aws-account/

## Phase 2: Cluster setup (**READY**)

This assumes all engineers have access to this new account, and will be able to set up the cluster + support, without any new hubs being set up.

| Question | Answer | 
| - | - |
| Region / Zone of the cluster | us-east-2 | 
| Name of cluster | dandi | 
| Is GPU required? | yes | 

Appropriate runbooks:
- New cluster setup: https://infrastructure.2i2c.org/hub-deployment-guide/new-cluster/aws/
- Support deployment: https://infrastructure.2i2c.org/hub-deployment-guide/deploy-support/

## Phase 3 : Hub setup (**READY**)

There's going to be a number of hubs, and this starts specifying them.

#### Phase 3.1: Initial setup 
| Question | Answer | Notes |
| - | - | - |
| Name of the hub | staging | |
| Dask gateway? | no | |
| Splash image | https://raw.githubusercontent.com/dandi/artwork/3f287d3ae53154a66f8b50711549740719a23fdb/pics/dandi-logo.svg | |
| URL | https://dandiarchive.org/ | | 

#### Phase 3.2: Authentication

| Question | Answer |
| - | - |
| Authentication Mechanism | GitHub (via GitHubOAuthenticator) |
| Org based access? | No |
[...]

PR #3866, opened 2024-03-27
Add new cluster DANDI

GeorgianaElena:
PHASE 1 and 2 of https://github.com/2i2c-org/infrastructure/issues/3824


yuvipanda (approved):
@GeorgianaElena this lgtm, and I think we should actually merge this if it's been deployed so far - so we can count 'phase 1 and phase 2' as completed.

config/clusters/dandi/cluster.yaml
```diff
@@ -0,0 +1,8 @@
+name: dandi
+provider: aws
```
consideRatio, line 2:
```suggestion
provider: aws # https://2i2c.awsapps.com/start#/
```

eksctl/dandi.jsonnet
```diff
@@ -28,6 +28,15 @@ local notebookNodes = [
     { instanceType: "r5.xlarge" },
     { instanceType: "r5.4xlarge" },
     { instanceType: "r5.16xlarge" },
+    {
+     instanceType: "g4dn.xlarge",
```
consideRatio, line 32: @yuvipanda should dandi have 2xlarge also like the linc hub?

The cost is ~50% higher (0.75 USD/hour compared to 0.53)

consideRatio: Ahhttps://github.com/2i2c-org/infrastructure/issues/3824#issuecomment-2024299035 declared that to be the case.

GeorgianaElena: @consideRatio, I've added https://github.com/2i2c-org/infrastructure/pull/3866/commits/392979ce325791b8c90240d635b656f3c5897cd8 to also support 2xlarge gpu machines. Does that look right to you or should we have _only_ 2xlarge machines?

GeorgianaElena:
With the grafana setup, phase 1 and 2 should now be ready. Merging PR

GeorgianaElena:
thank you @consideRatio and @yuvipanda for the feedback

eksctl/dandi.jsonnet
yuvipanda: @GeorgianaElena yep, we should have both.

Issue #3711, opened 2024-02-09
Provide a single source of truth for wether a community's GCP cloud costs is paid by us or not


yuvipanda:
### Context

There's a `gcp.billing.paid_by_us` boolean field in our `cluster.yaml` that tells us whether a GCP project is being paid for with the 2i2c credit card or the account is provided to us by the community. While this *is* documented, it's not well documented. Partnerships has some automation that looks at various hubs we run and scrapes it into airtable, but that doesn't pick up this field.

### Proposal

- [x] Go through all current clusters, and validate the `gcp.billing.paid_by_us` field is accurate (see https://github.com/2i2c-org/infrastructure/issues/3711#issuecomment-1977818547 for broad information on how)
  - Done via https://github.com/2i2c-org/infrastructure/pull/3872 and https://github.com/2i2c-org/infrastructure/pull/3873
- [x] Add an item in new hub turnup to verify this information is accurate https://github.com/2i2c-org/infrastructure/issues/3853
  - Done via https://github.com/2i2c-org/infrastructure/pull/3946
- [x] #4068
  - Done via #3959

#### Edit: Added to proposal

- [x] Ensure our "phase 1" about cluster creation ensures we have info about who should pay etc so that we set this up correctly initially.
  - Until there is a version controlled place to update "phase 1" which is a work in progress, this item is hard to get started with.

#### Edit: Removed from proposal

- [ ] ~Modify the automation that partnerships has to look for this boolean value and represent it in airtable~

damianavila:
@haroldcampbell, I think this one is well-specified and I see a clear definition of done, do you agree?
If that is not the case, can you drop your thoughts/requests so we can specify it better?
Thanks!

yuvipanda:
Just a note on how someone can verify if a particular project if being paid for by 2i2c or not.

1. On GCP, if the billing account attached to the project is https://console.cloud.google.com/billing/0157F7-E3EA8C-25AC3C?project=two-eye-two-see, we are paying for it. If not, we are not paying for it.
2. On AWS, if it's not part of our SSO (and hence accessible by starting at https://2i2c.awsapps.com/), it's not paid for by us.

haroldcampbell:
@damianavila this feels well-defined

consideRatio:
> On AWS, if it's not part of our SSO (and hence accessible by starting at https://2i2c.awsapps.com/), it's not paid for by us.

[...]

PR #3873, opened 2024-03-27
leap: change cluster.yaml's `gcp.billing.paid_by_us` to false

consideRatio:
- part of #3711

It seems that LEAP isn't paid by us, but via credits + a credit card I didn't recognize the last digits of to be CH's.


yuvipanda (approved):
Thanks for catching this, @consideRatio!

Issue #2187, opened 2023-02-09
Document how to use `deployer generate dedicated-cluster` in  the creation proces of new clusters


GeorgianaElena:
### Context

The template configuration files in `config/cluster/templates`, the:
```
├── aws
├── common
│   ├── support.secret.values.yaml
│   └── support.values.yaml
└── gcp
    └── cluster.yaml
```
 Can  now be used through the `deployer` to auto-generate specific files for new clusters on AWS and for new clusters and an initial hub on GCP.

### Proposal

- [x] Document how these templates, together with the `deployer generate dedicated-cluster` command can be used to make the process of adding a new cluster easier.
- [ ] Update or remove existing documentation like https://infrastructure.2i2c.org/hub-deployment-guide/deploy-support/configure-support/#create-a-support-values-yaml-file-in-your-chosen-cluster-folder as this is now replaced by the command completely.

Ref: https://github.com/2i2c-org/infrastructure/issues/1917

GeorgianaElena:
Useful feedback from @pnasrat  about this one:

- 1. **We need to document it**
- 2. Originally posted by @pnasrat in https://2i2c.slack.com/archives/C01GLCC1VCN/p1677843574698329?thread_ts=1677772541.271159&cid=C01GLCC1VCN
_"I think in general having templated config and tooling makes sense to reduce operational work, I had to comment out the terraform generation as I had already done that
Features that would have been useful - skipping a step (eg provisioning for terraform/eksctl), a flag allowing overwriting a cluster.yaml,
This means integration with new cloud providers might ideally look like doing once manually with a test providers, write up manual steps, and writing deployer code necessary to do - a  test create/destroy with deployer and create the actual community cluster"_

GeorgianaElena:
I updated the top comment to only cover the documentation part of the generator command as the command has been already extended to account for more providers (GCP and AWS currently). Also, I opened https://github.com/2i2c-org/infrastructure/issues/3294 as a follow-up to the comment above

damianavila:
Ask @sgibson91 if she can pair with @GeorgianaElena on this one.

GeorgianaElena:
> Document how these templates, together with the deployer generate dedicated-cluster command can be used to make the process of adding a new cluster easier.

This was actually solved by #3878 that standardized all three providers in the same document -> https://infrastructure.2i2c.org/hub-deployment-guide/new-cluster/new-cluster/

[...]

PR #3878, opened 2024-03-28
Allow to default to a basehub infrastructure setup during phase2 of hub deploy

GeorgianaElena:
When generating the infrastructure files using the deployer, information about basehub/daskhub type was needed.
In order to be able to better separate the deployment phases, we need to remove this requirement.

This PR, defaults to defaulting the deployer to basehub and documents what changes to make in order to accomodate a daskhub (when this information is needed).

Also fixes https://github.com/2i2c-org/infrastructure/issues/2187


docs/howto/features/dask.md
```diff
@@ -0,0 +1,73 @@
...
+   ```
+    dask_nodes = {
+    #  A not yet fully established policy is being developed about using a single
+    #  node pool, see https://github.com/2i2c-org/infrastructure/issues/2687.
+      "n2-highmem-16" : {
+        min : 0,
+        max : 200,
+        machine_type : "n2-highmem-16",
+      },
+    }
```
consideRatio, line 52: This is how it could look in a .jsonnet file after updatig the `local daskNodes = []` variable

```suggestion
local daskNodes = [
    // Node definitions for dask worker nodes. Config here is merged
    // with our dask worker node definition, which uses spot instances.
    // A `node.kubernetes.io/instance-type label is set to the name of the
    // *first* item in instanceDistribution.instanceTypes, to match
    // what we do with notebook nodes. Pods can request a particular
    // kind of node with a nodeSelector
    //
    // A not yet fully established policy is being developed about using a single
    // node pool, see https://github.com/2i2c-org/infrastructure/issues/2687.
    //
    { instancesDistribution+: { instanceTypes: ["r5.4xlarge"] }},
];
```

terraform/gcp/projects/cluster.tfvars.template
```diff
@@ -81,8 +81,8 @@ notebook_nodes = {
   }
 }
 
-dask_nodes = {
 {% if hub_type == "daskhub" %}
+dask_nodes = {
```
consideRatio, line 85: Like this, the logic below may reference an uninitilaized variable - making the logic below possibly more convoluted to handle it. Maybe its cleaner if we instead always initialize it?

GeorgianaElena: @consideRatio, I've updated the deployer command so that the hub_type will be requested as a prompt if no provided at all, where the default is `basehub`. I don't believe it can ever be uninitialized. It will look like this:

Screenshot 2024-04-03 at 16 52 36

docs/howto/features/dask.md
```diff
@@ -0,0 +1,78 @@
...
+terraform apply -var-file projects/$CLUSTER_NAME.tfvars
+```
+````
+
+### AWS
+
+We use `eksctl` with `jsonnet` to provision our kubernetes clusters on
+AWS, and we can configure a node group there for the dask pods to run onto.
+
+1. In the appropriate `.jsonnet` file, add a `dask_nodes` definition
```
consideRatio, line 40: Suggests updating the local variable instead of adding it, because it should already be there but initialized as an empty list.

```suggestion
1. In the appropriate `.jsonnet` file, update the `local daskNodes` definition 
```

terraform/gcp/projects/cluster.tfvars.template
```diff
@@ -22,40 +22,47 @@ regional_cluster = true
 #       Then use that version to explicitly set all k8s versions below, and
 #       finally decomment the k8s_versions section and removing this comment.
 #
-#k8s_versions = {
+# k8s_versions = {
```
consideRatio, line 25: With this, we avoid indentation errors on de-commenting

```suggestion
#k8s_versions = {
```

terraform/aws/projects/template.tfvars
```diff
@@ -1,26 +1,32 @@
...
-  },
-  "scratch" : {
-    "delete_after" : 7
-  },
-}
-
+# Tip: uncomment and fill the missing info in the lines below if you want
+#       to setup scratch buckets for the hubs on this cluster.
+#
+# user_buckets = {
```
consideRatio, line 16:
```suggestion
#user_buckets = {
```

terraform/aws/projects/template.tfvars
```diff
@@ -1,26 +1,32 @@
...
+# Tip: uncomment and fill the missing info in the lines below if you want
+#       to setup specific cloud permissions for the buckets in this cluster.
+#
+# hub_cloud_permissions = {
+#   "{{ hub_name }}" : {
+#     bucket_admin_access : ["scratch-{{ hub_name }}"],
+#     extra_iam_policy : ""
+#   },
+#   # Tip: add more namespaces below, if this cluster will be multi-tenant
+# }
```
consideRatio, line 32: Inconsistent indentation on commented out stuff here compared to other places

terraform/gcp/projects/cluster.tfvars.template
```diff
@@ -22,40 +22,47 @@ regional_cluster = true
...
 filestore_capacity_gb = 1024
 
-user_buckets = {
-  "scratch-{{ hub_name }}" : {
-    "delete_after" : 7
-  },
+# Tip: uncomment and fill the missing info in the lines below if you want
+#       to setup scratch buckets for the hubs on this cluster.
+#
+# user_buckets = {
```
consideRatio, line 47:
```suggestion
#user_buckets = {
```

terraform/gcp/projects/cluster.tfvars.template
```diff
@@ -22,40 +22,47 @@ regional_cluster = true
...
+#       to setup specific cloud permissions for the buckets in this cluster.
+#
+# hub_cloud_permissions = {
+#   "{{ hub_name }}" : {
+#     allow_access_to_external_requester_pays_buckets : true,
+#     bucket_admin_access : ["scratch-{{ hub_name }}"],
+#     hub_namespace : "{{ hub_name }}"
+#   },
+#   # Tip: add more namespaces below, if this cluster will be multi-tenant
+# }
```
consideRatio, line 64: Inconsistent indentation compared to other places (`#` + `space`, instead of just `#`).

terraform/gcp/projects/cluster.tfvars.template
```diff
@@ -74,15 +81,14 @@ notebook_nodes = {
...
 dask_nodes = {
+#  A not yet fully established policy is being developed about using a single
+#  node pool, see https://github.com/2i2c-org/infrastructure/issues/2687.
   "n2-highmem-16" : {
     min : 0,
     max : 200,
     machine_type : "n2-highmem-16",
   },
 }
+{% endif %}
```
consideRatio, line 94: If we make this commented out instead, do we then need to provide hub_type still in this context?

consideRatio (approved):
Nice, I think its good we work to avoid this differentiation.

I provided code suggestions to get consistent indentation within commented out blocks of code, and to get consistent indentation between commented out code blocks.

terraform/gcp/projects/cluster.tfvars.template
GeorgianaElena: Yes, I think so, but we'll also need to comment it out in the aws template https://github.com/2i2c-org/infrastructure/blob/d8018e18e41fadd2b9956934ca3cfbc1c1f2055b/eksctl/template.jsonnet#L43-L59.

But I think it's worth keeping with the change in this PR, where you just need to hit enter when you don't know the hub type:
Screenshot 2024-04-03 at 16 52 36

I also like the advantages that you get if you know from the beginning there will be daskhubs, i.e. don't change and regenerate afterwards. I think it's a decent code complexity compromise. Do you agree?

consideRatio: Absolutely, lets go! Thanks for reflecting on this!

PR #3879, opened 2024-03-28
[New hubs] Dandi

GeorgianaElena:
Closes https://github.com/2i2c-org/infrastructure/issues/3824

Also, updates the eksctl jsonnet file to not have dask nodes (I assumed there will be daskhubs on setup) and the terraform to use the requested buckets name (staging and prod were assumed by the template).


config/clusters/dandi/common.values.yaml
```diff
@@ -0,0 +1,146 @@
...
+    baseShareName: /
+jupyterhub:
+  hub:
+    config:
+      JupyterHub:
+        authenticator_class: github
+      Authenticator:
+        enable_auth_state: true
+        admin_users:
+          - asdf
```
yuvipanda, line 23:
```suggestion
```

GeorgianaElena:
I don't know what's wrong, but I can't get the autoscaler to trigger when starting a new server. It's just stuck in:

```
2024-03-29T11:50:12.467559Z [Warning] 0/1 nodes are available: 1 node(s) didn't match Pod's node affinity/selector. preemption: 0/1 nodes are available: 1 Preemption is not helpful for scheduling..
```

I even recreated the cluster, thinking I missed something during the first phases :|

yuvipanda:
@GeorgianaElena i can take a look at this today.

yuvipanda:
On AWS, we unfortunately have to manually install and manage the cluster autoscaler. I looked to see if it was crashing, with `k -n support get pod`, and noticed it wasn't there. We need to explicitly enable it in https://github.com/2i2c-org/infrastructure/blob/main/config/clusters/dandi/support.values.yaml, like in https://github.com/2i2c-org/infrastructure/blob/40fca243d0b8c246cf45a81e989286f815ffcff1/config/clusters/bican/support.values.yaml#L30.

I have opened https://github.com/2i2c-org/infrastructure/issues/3881 to automate this. Otherwise, I think if you enable the cluster autoscaler in the support chart and redeploy it, that should work!

And sorry for not catching it when I was reviewing https://github.com/2i2c-org/infrastructure/pull/3866

config/clusters/dandi/common.values.yaml
```diff
@@ -0,0 +1,145 @@
...
+                    node.kubernetes.io/instance-type: r5.xlarge
+              mem_29_7:
+                display_name: 29.7 GB RAM, upto 3.7 CPUs
+                kubespawner_override:
+                  mem_guarantee: 31861460992
+                  mem_limit: 31861460992
+                  cpu_guarantee: 3.75
+                  cpu_limit: 3.75
+                  node_selector:
+                    node.kubernetes.io/instance-type: r5.xlarge
```
yuvipanda, line 101: I went and re-read the requirements document again, and noticed that they wanted options upto 512GB of RAM. So I've updated the issue with a few more options here. Should be copy-pasteable.

GeorgianaElena:
> On AWS, we unfortunately have to manually install and manage the cluster autoscaler. 

Thank you @yuvipanda. I think this bit was recently removed from docs probably under the assumption that was already automated. Will work on fixing https://github.com/2i2c-org/infrastructure/issues/3881. Thank you 🙏🏼

GeorgianaElena:
I believe this PR is now ready to be reviewed/deployed 🚀

config/clusters/dandi/common.values.yaml
yuvipanda: I *think* this still needs to be removed

terraform/aws/projects/dandi.tfvars
```diff
@@ -8,19 +8,18 @@ user_buckets = {
   "scratch-staging" : {
     "delete_after" : 7
   },
-  "scratch" : {
+  "scratch-prod" : {
```
yuvipanda, line 11: I think we should keep these just as `scratch`, and not let the word `prod` leak into them.

yuvipanda (changes requested):
Two changes but otherwise looks good to me!

config/clusters/dandi/common.values.yaml
GeorgianaElena: Ah, I think I've overwritten the commit 🤦🏼‍♀️  Thank you, again :D

yuvipanda (approved):
Good to go!

yuvipanda:
Looks like I missed spotting the equivalent of https://github.com/2i2c-org/infrastructure/pull/3895 here too!

PR #3882, opened 2024-03-29
include new QGIS image, with improved outer UI and clipboard handling

batpad:
Includes a much improved UI for the shell around the QGIS desktop in Jupyter. Thanks much to @yuvipanda for the work in https://github.com/jupyterhub/jupyter-remote-desktop-proxy that added this.

This is what the UI now looks like:

Screenshot 2024-03-29 at 1 50 48 PM

 - Improves how Copy - Paste works
 - Gives access in the UI to go back to JupyterHub launcher + Server manager
 - Greatly improves the look of the shell UI, to make this look and behave well integrated into the JupyterHub setup.

I tested this image on staging and it seems to work well.

cc @yuvipanda


yuvipanda:
Thanks @batpad. Can you let the folks in the #veda-ap channel know about these improvements too once this goes out?

Issue #3827, opened 2024-03-21
[New Hub] BICAN (MIT Brain)


yuvipanda:
Copied over from https://github.com/2i2c-org/meta/issues/913

## Process Note

I'm using this as a way to try to rejig our new hub request process. See https://github.com/2i2c-org/meta/issues/897 (particularly https://github.com/2i2c-org/meta/issues/897#issuecomment-2010984904) for more information.

https://miro.com/app/board/uXjVNjUP3iQ=/, describes the various 'phases' of new hub turn-up. Each phase will be marked as "READY" or "NOT READY" when all information needed for it is available. Each section should also link to an appropriate runbook.

There will be customizations after this is all set up, but this is pathway towards a *standardized hub turn up*.

## Phase 1: Account setup (**READY**)

This is applicable for cases where this is a dedicated cluster. The following table lists the information before this phase can start.

| Question | Answer |
| - | - |
| Cloud Provider | AWS |
| Will 2i2c pay for cloud costs? | Yes |
| Name of cloud account | `bican` |

Appropriate runbook: https://infrastructure.2i2c.org/hub-deployment-guide/cloud-accounts/new-aws-account/

## Phase 2: Cluster setup (**READY**)

This assumes all engineers have access to this new account, and will be able to set up the cluster + support, without any new hubs being set up.

| Question | Answer | 
| - | - |
| Region / Zone of the cluster | us-east-2 | 
| Name of cluster | bican | 
| Is GPU required? | yes | 

Appropriate runbooks:
- New cluster setup: https://infrastructure.2i2c.org/hub-deployment-guide/new-cluster/aws/
- Support deployment: https://infrastructure.2i2c.org/hub-deployment-guide/deploy-support/

## Phase 3 : Hub setup (**READY**)

There's going to be a number of hubs, and this starts specifying them.

#### Phase 3.1: Initial setup 
| Question | Answer | Notes |
| - | - | - |
| Name of the hub | staging | |
| Dask gateway? | no | |
| Splash image | https://static.wixstatic.com/media/c57c96_a9e46e43008349c8b65dcacc3ceaba35~mv2.png/v1/fill/w_402,h_150,al_c,q_85,usm_0.66_1.00_0.01,enc_auto/_BICAN-full-logo-final.png | |
| URL | https://www.portal.brain-bican.org/ | | 

#### Phase 3.2: Authentication

| Question | Answer |
| - | - |
[...]

PR #3891, opened 2024-04-02
Add BICAN hub

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/3827

### DIFF
--- a/config/clusters/bican/common.values.yaml
+++ b/config/clusters/bican/common.values.yaml
@@ -42,11 +42,11 @@ jupyterhub:
         authenticator_class: github
       Authenticator:
         admin_users:
-        - kabilar
-        - aaronkanzer
-        - asmacdo
-        - satra
-        - djarecka
+          - kabilar
+          - aaronkanzer
+          - asmacdo
+          - satra
+          - djarecka
   singleuser:
     profileList:
       - display_name: "DANDI (CPU)"
--- a/config/clusters/bican/common.values.yaml
+++ b/config/clusters/bican/common.values.yaml
@@ -1,0 +1,183 @@
+nfs:
+  enabled: true
+  pv:
+    enabled: true
+    # from https://docs.aws.amazon.com/efs/latest/ug/mounting-fs-nfs-mount-settings.html
+    mountOptions:
+      - rsize=1048576
+      - wsize=1048576
+      - timeo=600
+      - soft # We pick soft over hard, so NFS lockups don't lead to hung processes
+      - retrans=2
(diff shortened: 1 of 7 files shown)


consideRatio (approved):

config/clusters/bican/common.values.yaml
```diff
@@ -0,0 +1,183 @@
...
+      Authenticator:
+        admin_users:
+          - kabilar
+          - aaronkanzer
+          - asmacdo
+          - satra
+          - djarecka
+  singleuser:
+    profileList:
+      - display_name: "DANDI (CPU)"
```
GeorgianaElena, line 52: Not sure, as the tracking issue also says `dandi` but shouldn't this be Bican instead of Dandi??

eksctl/bican.jsonnet
```diff
@@ -37,6 +37,15 @@ local notebookNodes = [
       // GPUs in a single AZ are in use and no new nodes can be spawned
       availabilityZones: masterAzs,
     },
+    {
```
GeorgianaElena, line 40: I _think_ we're missing a  `g4dn.xlarge` here as well? I see it being specified in the profile options, so we should remove it from there if it's not required

GeorgianaElena (commented):
This looks good @yuvipanda, thanks for setting it up 🚀 . I've left a few comments with some questions where I wasn't sure if this hub should be different than dandi.

eksctl/bican.jsonnet
consideRatio: image

It is specified since before, now we specify both here.

I think its reasonable that we transition to a default of specifying both and allowing for a choice of 4CPU and 8CPU, the cost difference is 50%+ for 8 CPU, and when whats matters is just basic testing it could be nice to opt cheap by default or similar.

config/clusters/bican/common.values.yaml
yuvipanda: @GeorgianaElena these are DANDI images, so I think we can call them that for now. I think this will be modified in the future.

eksctl/bican.jsonnet
yuvipanda: Yeah, Sarah already set that up! https://github.com/2i2c-org/infrastructure/pull/3891/files#diff-0067a696fa6484b21a3052fb489c0e6e09703df1b5c700f5a379d6641e8ff73cR32

yuvipanda:
Thanks @GeorgianaElena!

PR #3892, opened 2024-04-02
Don't specify oauth_callback_url in config for GitHub

yuvipanda:
Turns out this is not actually needed, I accidentally discovered this while setting up the BICAN hub


yuvipanda:
Thanks @consideRatio. I'm just going to close this one now, as I don't think we have the capacity to properly review and set this up.

PR #3893, opened 2024-04-02
Remove setting 'oauth_callback_url' step when using github auth

yuvipanda:
It's not needed. Separated the doc change out from the config cleanup in https://github.com/2i2c-org/infrastructure/pull/3892 for easier review and merging.

### DIFF
--- a/docs/hub-deployment-guide/configure-auth/github-orgs.md
+++ b/docs/hub-deployment-guide/configure-auth/github-orgs.md
@@ -82,7 +82,6 @@ You can remove yourself from the org once you have confirmed that login is worki
           JupyterHub:
             authenticator_class: github
           GitHubOAuthenticator:
-            oauth_callback_url: https://{{ HUB_DOMAIN }}/hub/oauth_callback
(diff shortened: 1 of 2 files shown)


consideRatio:
I think this is a fine change assuming our docs ensures we don't break the conditions outlined in https://github.com/2i2c-org/infrastructure/pull/3892#issuecomment-2031169000 when doing misc things over time.

- _Updates to documentation about adopting a community managed domain_
   - a redirect to the new domain is configured
   - an allowed redirect url isn't added to the github oauth application, but instead replaces the old or put at the top of the list

It seemed like the community docs may not have notes on the need to update the github apps with a redirect url, or configuring redirection rules currently. I think such updates are pre-requisites for this PR to land robustly. I think either we do this as part of this PR or issues are opened to track doing it later if we believe we can really get it done somewhat soonish.

yuvipanda:
Thanks @consideRatio. I'm just going to close this one now, as I don't think we have the capacity to properly review and set this up.

PR #3895, opened 2024-04-02
Add missing `failure_bican` to github workflow

yuvipanda:


yuvipanda:
Fixes the error in https://github.com/2i2c-org/infrastructure/actions/runs/8527031771/job/23357757638

PR #3901, opened 2024-04-03
linc: setup staging/prod hubs (and fixes for bican/dandi)

consideRatio:
I tested ability to startup the biggest resource allocation request, as we don't have any headroom its very sensitive to changes. I observed that EKS k8s 1.29 has eaten up a bit of headroom, so we were unable to start those instances. I've updated node info stuff so that future generation attempts won't fail in EKS and fixed the resource allocation stuff for bican/dandi/linc.

As soon as other hubs upgrade to k8s 1.29 (maybe 1.28), they will run into these kinds of issues again as memory/cpu headroom is reduced.

- PR for #3828 

## Remaining

- `funded_by` not filled in


GeorgianaElena (approved):
Thanks for fixing the allocation issue on the other mit hubs as well @consideRatio ❤️ 
LGTM 🚀

consideRatio:
Thanks for reviewing @GeorgianaElena! I force pushed to filter out the deployer stuff into a dedicated PR (#3903).

consideRatio:
The funding_url should be declared, and maybe there are more things, but i'll open a new PR for that and go for a merge.

Thanks for reviewing @GeorgianaElena!

PR #3905, opened 2024-04-03
binderhub-service deploy: hide irrelevant services dropdown entry

consideRatio:
This is an artifact of how we configure things, and we don't want it to show up like this:
image

By using `display: false`, we avoid it. See https://jupyterhub.readthedocs.io/en/stable/reference/services.html#properties-of-a-service


GeorgianaElena (approved):
Good idea. LGTM

PR #3911, opened 2024-04-03
automation: also deploy grafana's global dashboard for 2i2c cluster, and deploy all dashboards daily

consideRatio:
The global dashboard wasn't getting updated over time through our automation. I think this is the long-term fix for #3911, where I previously just manually patched it after finding it outdated and malfunctioning due to it.

### DIFF
--- a/.github/workflows/deploy-grafana-dashboards.yaml
+++ b/.github/workflows/deploy-grafana-dashboards.yaml
@@ -70,4 +70,4 @@ jobs:
 
       - name: Deploy grafana dashboards for ${{ matrix.cluster_name }}
         run: |
-          deployer grafana deploy-dashboards ${{ matrix.cluster_name }} ${{ matrix.deploy_flags }}
+          deployer grafana deploy-dashboards ${{ matrix.deploy_flags }} ${{ matrix.cluster_name }}
--- a/.github/workflows/deploy-grafana-dashboards.yaml
+++ b/.github/workflows/deploy-grafana-dashboards.yaml
@@ -1,6 +1,10 @@
 name: Deploy grafana dashboards
 
-on: workflow_dispatch
+on:
+  schedule:
+    # Run at 05:00 every day, ref: https://crontab.guru/#0_5_*_*_*
+    - cron: "0 5 * * *"
+  workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-grafana-dashboards.yaml
```diff
@@ -1,16 +1,24 @@
...
     strategy:
       # Don't stop other deployments if one fails
       fail-fast: false
       matrix:
         include:
-          # The grafana for 2i2c cluster holds also info about all other clusters
+          # The grafana for 2i2c cluster holds also info about all other
+          # clusters and gets an additional dashboard deployed for that
+          - cluster_name: 2i2c
+            deploy_flags: --dashboards-dir=global-dashboards
```
GeorgianaElena, line 20: Because the `deployer grafana deploy-dashboards` command [wraps the `deploy` script](https://github.com/2i2c-org/infrastructure/blob/c168f862b0890916cd9cd1cd589c671b6371d2d5/deployer/commands/grafana/deploy_dashboards.py#L42-L49) of jupyterhub-grafana-dashboards I believe you also need to add support for this flag in the deployer, otherwise the flag won't be recognized.

https://github.com/2i2c-org/infrastructure/blob/c168f862b0890916cd9cd1cd589c671b6371d2d5/deployer/commands/grafana/deploy_dashboards.py#L13-L16

GeorgianaElena (commented):
I believe this is a good idea as it is now stable enough to be deployed automatically. Hopefully nobody will be caught by surprise by the fact that any custom dashboard will be overwritten.

Left a comment about adding support for the `dashboards-dir` flag in the deployer.

consideRatio:
> Hopefully nobody will be caught by surprise by the fact that any custom dashboard will be overwritten.

If edits have been made to the dashboards deployed by jupyterhub/grafana-dashboards, then they will be overwritten - but other dashboards should be left intact I think!

.github/workflows/deploy-grafana-dashboards.yaml
consideRatio: Ah nice catch, resolved by 19e1ff2

consideRatio:
Approved via slack chat by Georgiana

PR #3916, opened 2024-04-04
Keep two years of prometheus data for our main shared cluster

yuvipanda:
We haven't fully billed this out yet I think, and I don't want to lose historical data right now.

### DIFF
--- a/config/clusters/2i2c/support.values.yaml
+++ b/config/clusters/2i2c/support.values.yaml
@@ -7,6 +7,17 @@ prometheusStorageClass:
 
 prometheus:
   server:
+    defaultFlagsOverride:
+      - -c
+      - "rm -rvf /data/wal/* && \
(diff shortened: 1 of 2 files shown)


consideRatio (approved):

consideRatio:
Off topic meta question:

I'm looking at a huge amount of PR titles a day via notifications, and for this repo, I find it helpful to have a prefix of the cluster name if its a change specific to a cluster for example. What do you think about adopting a pattern of leading with cluster name for cluster specific changes, like `2i2c: <title>`?

I've adopted such practice, and when for example a community rep opens a PR that I look at that may be named like "Updated image to latest version" I'll add `cluster, hub: ` or `cluster:` as a prefix to the title to help me and others orient quicker what its about.

yuvipanda:
That sounds like a good idea.

PR #3920, opened 2024-04-08
support chart: update grafana chart 6.61.2 to 7.3.7 (grafana 10.1.5 to 10.4.0)

### LINKED ISSUE #3619: Upgrade to latest Grafana
In https://github.com/2i2c-org/infrastructure/pull/3482, we updated all the components of our support setup *except* Grafana - there was some breaking change, and i didn't want it to hold the other components back.

This issue is to track upgrading that.

consideRatio:
This is not a breaking change for us now, grafana itself still supports our dashboard definitions from jupyterhub/grafana-dashboards, and the chart major version bump relates to a [for us unused chart configuration](https://github.com/grafana/helm-charts/blob/main/charts/grafana/README.md#to-700).

There were upgrade errors stemming from grafana introducing (without a major version bump, but should have been!) a [strict check for secrets being passed introduced in chart version 7.1.0](https://github.com/grafana/helm-charts/pull/2867) - this PR handles that by disabling the (too) strict check - see Yuvi's comment for example in https://github.com/2i2c-org/infrastructure/pull/3482#issuecomment-1894722474, which I also align with.

- Fixes #3619

### DIFF
--- a/helm-charts/support/Chart.yaml
+++ b/helm-charts/support/Chart.yaml
@@ -21,7 +21,7 @@ dependencies:
   # Grafana for dashboarding of metrics.
   # https://github.com/grafana/helm-charts/tree/main/charts/grafana
   - name: grafana
-    version: 6.61.2
+    version: 7.3.7
     repository: https://grafana.github.io/helm-charts
 
   # ingress-nginx for a k8s Ingress resource controller that routes traffic from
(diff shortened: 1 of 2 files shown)


GeorgianaElena (approved):
Thanks for getting to the bottom of this @consideRatio

Issue #3881, opened 2024-03-29
Autogenerating support structure should enable cluster autoscaler on AWS


yuvipanda:
On AWS, we have to explicitly enable the cluster autoscaler. Our generator for generating support should do this when the cluster is on AWS automatically, as we always need this.

PR #3924, opened 2024-04-09
[deployer] Explicitly enable the cluster-autoscaler when generating the support chart for aws clusters

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/3881

### DIFF
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -23,6 +23,8 @@ on:
       - .github/actions/setup-deploy/**
       - helm-charts/**
       - config/clusters/**
+      # Exclude the template configuration files
+      - "!config/clusters/templates"
   pull_request:
     branches:
       - main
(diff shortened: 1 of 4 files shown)


consideRatio (approved):
Nice!!

I assume there may be related docs that is now outdated, but we can merge this as it is also as i expect no significant negative consequence of that

GeorgianaElena:
Thanks @consideRatio ❤️ 

> I assume there may be related docs that is now outdated, but we can merge this as it is also as i expect no significant negative consequence of that.

So, actually there is no info in our docs about this anymore and this is why I had trouble when deploying the dandi hub 😅  I believe it was removed under the assumption it is already being autogenerated by the deployer, which is what this PR does.

We do have outdated info about the support chart, mainly about the need to manually generate its contents, but that is tracked in https://github.com/2i2c-org/infrastructure/issues/2187

> I'd cancel the job to deploy on merge this PR btw

Good idea. I will try to also exclude the template files from the worflow too. Commit incoming.

Issue #3906, opened 2024-04-03
[Request deployment] New Hub: Kitware


colliand:
## Phase 1: Account setup (**READY**)

This is applicable for cases where this is a dedicated cluster. The following table lists the information before this phase can start.

| Question | Answer |
| - | - |
| Cloud Provider | AWS |
| Will 2i2c pay for cloud costs and recover them via invoice? |  yes |
| Target Start Date | 2024-04-30 |
| Community Representatives | @annehaley @aashish24 |
| Technical Contacts added to Airtable? | No |

Docs at https://infrastructure.2i2c.org/hub-deployment-guide/cloud-accounts/

## Phase 2: Cluster setup (**READY**)

This assumes all engineers have access to this new account, and will be able to set up the cluster + support, without any new hubs being set up.

| Question | Answer | 
| - | - |
| Region / Zone of the cluster | us-west-2 | 
| Name of cluster | `kitware` | 
| Is GPU required? | Yes | 

Docs at https://infrastructure.2i2c.org/hub-deployment-guide/new-cluster/new-cluster/
and https://infrastructure.2i2c.org/hub-deployment-guide/deploy-support/configure-support/

## Phase 3 : Hub setup

There's going to be a number of hubs, and this starts specifying them.

#### Phase 3.1: Initial setup (**READY**)

| Question | Answer | Notes |
| - | - | - |
| Name of the hub | staging | |
| Dask gateway? | no | |
| Splash image | https://www.kitware.com/main/wp-content/uploads/2021/10/Kitware-Wordmark.svg | |
| URL | https://www.kitware.com/  | Used for both URL and funded_by | 
| Authentication Mechanism | GitHub (via GitHubOAuthenticator) |
| Admin Users | @aashish24 @annehaley (**needs confirmation**) | 

At the end of this phase, both 2i2c engineers and the admin users mentioned can login to the hub.

#### Phase 3.2: Object storage access (**READY**)

| Question | Answer | Notes |
| - | - | - |
| Scratch bucket enabled? |  Yes |  |
| Persistent bucket enabled? | no |  |
| Requestor pays requests to external buckets allowed? | no | |  

#### Phase 3.3: Profile List (**READY**)

These are the standard profile list options we want to start with. Can be customized later

| Question | Answer | Notes |
| - | - | - |
[...]

PR #3929, opened 2024-04-10
[New cluster] Kitware

GeorgianaElena:
Ref #3906 

Also updates some things in the docs while walking through them.


docs/hub-deployment-guide/deploy-support/configure-support.md
```diff
@@ -51,19 +51,6 @@ prometheus:
...
-Include the following in your `support.values.yaml` file:
-
-```yaml
-cluster-autoscaler:
-  enabled: true
-  autoDiscovery:
-    clusterName: <cluster-name>
-  awsRegion: <aws-region>
-```
-````
```
GeorgianaElena, line 65: @consideRatio, apparently it was still documented, I might have missed it when deploying dandi. Anyway, I've removed it now that we autogenerate it via the deployer.

consideRatio (approved):
Maybe add the  8cpu gpu option, then copy paste from recent hub setups will work well. I think it makes sense to provide both in general i guess. Out of scope from doing it by the docs though.

Lgtm no matter what!

GeorgianaElena:
Thank you @consideRatio 🚀 

> Maybe add the 8cpu gpu option, then copy paste from recent hub setups will work well. I think it makes sense to provide both in general i guess.

Wasn't sure what to do about this, but tried not to assume and followed the docs as close as possible as an experiment. As long as we can add it later, if it's explicitly required, I think it's fine.

We can also add it las a general good practice in the docs as you said.

Issue #3038, opened 2023-08-28
[Support] 2i2c-aws-us/itcoocean: help to setup s3 bucket (admin/non-admin with readwrite/readonly access)


consideRatio:
In https://2i2c.freshdesk.com/a/tickets/906 Eli asks for help to setup a s3 bucket for the itcoocean.2i2c.cloud hub part of the 2i2c-aws-us cluster.

The goals are:

- Admin users should have read/write permissions
- non-admin users should just have read access
- Allowing anyone access the data, but to not need to pay for users accessing the data.

> Goal is to not have egress charges from our S3 bucket (in us-west-2). The data are not "private". We'd make the S3 bucket public but the egress charges would rack up.

## Related

- Looking at [s3 pricing](https://aws.amazon.com/s3/pricing/), I see they mention you don't have to pay for:
  > Data transferred from an Amazon S3 bucket to any AWS service(s) within the same AWS Region as the S3 bucket (including to a different account in the same AWS Region).
- 2i2c infrastructure docs about buckets: https://infrastructure.2i2c.org/howto/features/cloud-access/
- Notes from Eli about setting up a bucket and declaring permissions: https://2i2c.freshdesk.com/a/tickets/906

## Todo

- Distinguishing admin users from non-admin users access to buckets
  - [x] Enable us to configure admin users to have different k8s ServiceAccounts (that should be associated with AWS credentials in s3 buckets) than normal users
    #3039
  - Enable creation of another set of cloud credentials to access buckets with
    - [x] For AWS
    - [ ] For GCP
  - [x] Enable creation of another k8s ServiceAccount that we can annotate as well
- [ ] Allowing anyone access the data, but to not need to pay for users accessing the data

jmunroe:
<div class="panel panel-warning" style="border-color: #faebcc;margin-bottom: 20px;background-color: #fff;border: 1px solid transparent;border-radius: 4px;-webkit-box-shadow: 0 1px 1px rgb(0 0 0 / 5%);box-shadow: 0 1px 1px rgb(0 0 0 / 5%);font-size: 14px;">
                <div class="panel-heading" style="color: #8a6d3b;background-color: #fcf8e3;border-color: #faebcc;padding: 10px 15px;border-bottom: 1px solid transparent;border-top-left-radius: 3px;border-top-right-radius: 3px;">Comment from Freshdesk:</div>
                <div class="panel-body" style="padding: 15px;">
[...]

Issue #2886, opened 2023-07-28
[Support] cryocloud: Allow bucket access based on hub admin status


GeorgianaElena:
### Context

In https://2i2c.freshdesk.com/a/tickets/847 a persistent bucket is requested. This bucket should be accessible only by hub admins.

It is my understanding that this is not possible given the current setup because we map all the hub user pods to a cloud user account (`user-sa`) and there's no further granular acccess.

Another way to manage more granular access is [through Google Groups](https://infrastructure.2i2c.org/howto/features/buckets/#allowing-access-to-buckets-from-outside-the-jupyterhub) but that's only setup for GCP.

### Question
What does @2i2c-org/engineering think about a new hub-admin <-> specific cloud user account for more granular access setup on the hubs? Is this feasible? Or is there an easier way that I'm missing?
```[tasklist]
### Tasks
- [ ] https://github.com/2i2c-org/infrastructure/issues/2893
- [x] Decide if granular access based on admin/non-admin status of users is achievable
- [ ] Follow-up to https://2i2c.freshdesk.com/a/tickets/847 with an answer
```

damianavila:
Bumping ping to @2i2c-org/engineering for feedback on @GeorgianaElena's question.

yuvipanda:
I think we're currently not planning on doing this for cryocloud. If that changes, we can reopen this.

PR #3932, opened 2024-04-11
aws, terraform: support creation of more than a single IAM Role, and add `bucket_readonly_access` config

consideRatio:
## New feature

- Adds terraform config `bucket_readonly_access` under `hub_cloud_permissions`

## Changed

The AWS terraform variable `hub_cloud_permissions` is changed from having a structure of:

```yaml
hub_cloud_permissions:
  <hub name>:
    bucket_admin_access: []
    extra_iam_policy: ""
```

to:

```yaml
hub_cloud_permissions:
  <hub name / k8s namespace>:
    <aws iam role name / k8s service account name>
      bucket_admin_access: []
      bucket_readonly_access: []
      extra_iam_policy: ""
```

Where the fields `bucket_admin_access`, `bucket_readonly_access`, and `extra_iam_policy` have also been made optional.

## terraform apply changes

This change is meant to not cause issues when `terraform apply` is used for existing clusters. It will however cause a change to a resource - the bucket access policy.

<details><summary>`terraform apply` safe-to-apply changes</summary>

```terraform

  # aws_s3_bucket_policy.user_bucket_access["prod.scratch"] will be destroyed
  # (because key ["prod.scratch"] is not in for_each map)
  - resource "aws_s3_bucket_policy" "user_bucket_access" {
      - bucket = "bican-scratch" -> null
      - id     = "bican-scratch" -> null
      - policy = jsonencode(
            {
              - Statement = [
                  - {
                      - Action    = "s3:*"
                      - Effect    = "Allow"
                      - Principal = {
                          - AWS = "arn:aws:iam::533267153382:role/bican-prod"
                        }
                      - Resource  = [
                          - "arn:aws:s3:::bican-scratch/*",
                          - "arn:aws:s3:::bican-scratch",
                        ]
                      - Sid       = ""
                    },
                ]
              - Version   = "2012-10-17"
            }
        ) -> null
    }

  # aws_s3_bucket_policy.user_bucket_access["scratch"] will be created
  + resource "aws_s3_bucket_policy" "user_bucket_access" {
      + bucket = "bican-scratch"
      + id     = (known after apply)
      + policy = jsonencode(
            {
              + Statement = [
                  + {
                      + Action    = "s3:*"
                      + Effect    = "Allow"
                      + Principal = {
                          + AWS = "arn:aws:iam::533267153382:role/bican-prod"
                        }
                      + Resource  = [
                          + "arn:aws:s3:::bican-scratch/*",
                          + "arn:aws:s3:::bican-scratch",
                        ]
                      + Sid       = ""
                    },
                ]
              + Version   = "2012-10-17"
            }
        )
    }

  # aws_s3_bucket_policy.user_bucket_access["scratch-staging"] will be created
  + resource "aws_s3_bucket_policy" "user_bucket_access" {
      + bucket = "bican-scratch-staging"
      + id     = (known after apply)
      + policy = jsonencode(
            {
              + Statement = [
                  + {
                      + Action    = "s3:*"
                      + Effect    = "Allow"
                      + Principal = {
                          + AWS = "arn:aws:iam::533267153382:role/bican-staging"
                        }
                      + Resource  = [
                          + "arn:aws:s3:::bican-scratch-staging/*",
                          + "arn:aws:s3:::bican-scratch-staging",
                        ]
                      + Sid       = ""
                    },
                ]
              + Version   = "2012-10-17"
            }
        )
    }

  # aws_s3_bucket_policy.user_bucket_access["staging.scratch-staging"] will be destroyed
  # (because key ["staging.scratch-staging"] is not in for_each map)
  - resource "aws_s3_bucket_policy" "user_bucket_access" {
      - bucket = "bican-scratch-staging" -> null
      - id     = "bican-scratch-staging" -> null
      - policy = jsonencode(
            {
              - Statement = [
                  - {
                      - Action    = "s3:*"
                      - Effect    = "Allow"
                      - Principal = {
                          - AWS = "arn:aws:iam::533267153382:role/bican-staging"
                        }
                      - Resource  = [
                          - "arn:aws:s3:::bican-scratch-staging/*",
                          - "arn:aws:s3:::bican-scratch-staging",
                        ]
                      - Sid       = ""
                    },
                ]
              - Version   = "2012-10-17"
            }
        ) -> null
    }

Plan: 2 to add, 0 to change, 2 to destroy.
```

</details>

## Limitations and possible issues

- `hub_cloud_permissions` structure is now changed for AWS but not for GCP, making them have different structure but otherwise very similar structure.
- `bucket_readonly_access` is only implemented for AWS with this PR
- `dask-gateway` is hardcoded to use `user-sa` and can't securely use whatever the user is having I think.
  So, if we configure `jupyterhub.custom.singleuserAdmin.serviceAccountName: admin-sa`, users requests for dask clusters (dask-scheduler pod + dask-worker pods) they won't have the same service account.

## Reamining work before merge

- [x] There is a docs helper script creating a table of features, this code needs to be updated as the structure of
- [x] Check misc docstrings etc
- [x] `terraform apply` across all clusters so this change doesn't cause confusion for other people later

## Remaining work after merge

We decided that we'll get this merged first, and then write docs as we wanted to prio other things.

- [ ] Write documentation about this functionality

## Related

- Fixes https://github.com/2i2c-org/meta/issues/979
- It can also fix other similar requests:
  - https://github.com/2i2c-org/infrastructure/issues/2886
  - https://github.com/2i2c-org/infrastructure/issues/3038


terraform/aws/irsa.tf
```diff
@@ -1,20 +1,38 @@
...
 
-  assume_role_policy = data.aws_iam_policy_document.irsa_role_assume[each.key].json
+
+locals {
+  hub_role = flatten([
+    for hub, hub_value in var.hub_cloud_permissions : [
+      for role, role_value in hub_value : {
+        // id is conservatively adjusted to not change any previous resource
+        // name set to the hub's name when only "user-sa" roles were around
+        id   = role == "user-sa" ? hub : "${hub}-${role}"
```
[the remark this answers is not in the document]
yuvipanda: Thank you for thinking this through and minimizing the amount of work needed here!

terraform/aws/irsa.tf
```diff
@@ -1,20 +1,38 @@
...
 data "aws_partition" "current" {}
 
-resource "aws_iam_role" "irsa_role" {
-  for_each = var.hub_cloud_permissions
-  name     = "${var.cluster_name}-${each.key}"
 
-  assume_role_policy = data.aws_iam_policy_document.irsa_role_assume[each.key].json
+
+locals {
+  hub_role = flatten([
```
yuvipanda, line 16: Let's call this `hub_to_roles_mapping` for clarity? `hub_role` sounds like a singular hub's role to me.

terraform/aws/irsa.tf
```diff
@@ -1,20 +1,38 @@
...
+    for hub, hub_value in var.hub_cloud_permissions : [
+      for role, role_value in hub_value : {
+        // id is conservatively adjusted to not change any previous resource
+        // name set to the hub's name when only "user-sa" roles were around
+        id   = role == "user-sa" ? hub : "${hub}-${role}"
+        hub  = hub
+        role = role
+        data = role_value
```
yuvipanda, line 24: `data` feels too generic here. Can we call it `cloud_permissions` to match what is in the configuration variable name (`hub_cloud_permissions`)?

terraform/aws/irsa.tf
```diff
@@ -1,20 +1,38 @@
...
-resource "aws_iam_role" "irsa_role" {
-  for_each = var.hub_cloud_permissions
-  name     = "${var.cluster_name}-${each.key}"
 
-  assume_role_policy = data.aws_iam_policy_document.irsa_role_assume[each.key].json
+
+locals {
+  hub_role = flatten([
+    for hub, hub_value in var.hub_cloud_permissions : [
+      for role, role_value in hub_value : {
```
yuvipanda, line 18: My understanding is that the *key* here refers to the name of the Kubernetes Service Account (because that is what is specified in the variable). The name of the kubernetes service account is the same as the name of the IAM role (with a prefix) in some cases (`user-sa`) but not others. And `id` is here our own creation, and not the same as the `id` of the `iam_role` resource (https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_role#id).

I would suggest the following fields:

1. `iam_role_name` (currently called `id`)
2. `ksa_name` (currently called `role`)

I believe this would make this intermediate data structure easier to understand when being called.

yuvipanda: This would make things like https://github.com/2i2c-org/infrastructure/pull/3932/files#diff-9f223d043fe1fe9d6f6e3f3a909a3e8d633b2fbbc20f388af72c20069d53cc5aR49 much more obvious for example - we're using that in the context of the kubernetes service account name inside a namespace, not the name of the role.

terraform/aws/irsa.tf
yuvipanda: This would also make lines like https://github.com/2i2c-org/infrastructure/pull/3932/files#diff-9f223d043fe1fe9d6f6e3f3a909a3e8d633b2fbbc20f388af72c20069d53cc5aR73 clearer. Just reading that I wasn't sure if `data` was another computed field or not.

terraform/aws/bucket-access.tf
```diff
@@ -0,0 +1,145 @@
...
+      - bucket: persistent-sciencecore
+        role: sciencecore
+        actions: ["s3:ListBucket", "s3:GetObject", "s3:GetObjectVersion"]
+      - bucket: persistent-sciencecore
+        role: sciencecore
+        actions: ["s3:*"]
+  */
+  bucket_role_actions = flatten([
+    for hub, hub_value in var.hub_cloud_permissions : [
+      for role, role_value in hub_value : flatten([
```
yuvipanda, line 51: Can we keep the credit for the `flatten` trick from https://github.com/2i2c-org/infrastructure/pull/3932/files#diff-c1bb77fe872eda5a7e42a026322e595cf40d3595a0605c60be0793a362310639L40

yuvipanda (changes requested):
Thanks for scoping and working on this, @consideRatio. The overall structure looks good to me - it causes minimal churn right now while also allowing us to expand this in the future if we wish. Splitting the files is also a good call.

I've left some comments about naming, as well as made two PRs to your PR with respect to some doc strings and comments. With those addressed, I think this is good to go.

I didn't test the actual effects of these, although I did test `terraform apply` and observed the changes to be ok. I think you've already tested them in various places. I would say we should make sure the `extra_iam_permission` hubs work correctly, and make sure we verify those after apply.

Thanks!

terraform/aws/irsa.tf
consideRatio: Thanks for thinking about naming, I agree fully that these are improvements and I think in this case - quite important improvements for anyone trying to understand this code.

terraform/aws/irsa.tf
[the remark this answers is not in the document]
consideRatio: Thank you for going for the code suggestion / PR, merged!

consideRatio:
Thank you for reviewing this carefully and making it easy for me to address your input @yuvipanda! All review comments addressed in pushed commits.

I've previously checked that these changes work end to end in opensci/sciencecore - and I think they have also successfully been used by the community. I've also checked after the new commits have been made, that `terraform apply` leads to `No changes` in opensci cluster as expected.

If this looks good to you now, I've outlined the next steps I intend to take in https://github.com/2i2c-org/infrastructure/pull/3932#issuecomment-2066894289

consideRatio:
Thank you for meeting with me about what led to this and reviewing efforts @yuvipanda!!

PR #3934, opened 2024-04-11
update default image for jupyter-health

minrk:
we now have an [image-building repo](https://github.com/jupyterhealth/singleuser-image), so use that

- set default image to https://quay.io/repository/jupyterhealth/singleuser-premvp:latest
- remove large node size choices (we aren't doing anything big yet)
- remove geospatial image choice


minrk:
Ultimately, I think we'll want to remvoe the image choice and make this a simple single-image deployment, but having the escape hatch for now is useful for prototyping

consideRatio (approved):

PR #3939, opened 2024-04-12
automation: grafana dashboard deployment requires go-jsonnet

consideRatio:
`jsonnet` is available in a c++ version and a golang version, and only the golang version works with the library grafonnet versions 10.2.0+ (maybe it was 10.3.0+ or so, but around here) which is used by jupyterhub/grafana-dashboards.

So this change installs the golang based jsonnet instead so that our automation to re-deploy grafana dashboards starts working again.

I've hardcoded the version to 0.20.0, but releases are infrequent and I figure we don't need to automate bumping that.

### DIFF
--- a/.github/workflows/deploy-grafana-dashboards.yaml
+++ b/.github/workflows/deploy-grafana-dashboards.yaml
@@ -58,7 +58,12 @@ jobs:
       - name: Install deployer
         run: |
           python3 -m pip install --editable .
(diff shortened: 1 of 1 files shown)


GeorgianaElena (approved):
🚀

consideRatio:
image

Thank you @GeorgianaElena!!

PR #3941, opened 2024-04-12
[Kitware] Setup grafana

GeorgianaElena:
Ref #3906


GeorgianaElena:
Thank you @consideRatio 🚀 

> but perhaps its already setup? Like root_url etc

Yes, this just adds the last part of setting up Grafana for the first time like generating the api token so the dashbords can be deployed

PR #3942, opened 2024-04-13
[ucmerced] update images

jmunroe:
Re: https://2i2c.freshdesk.com/a/tickets/1407


jmunroe:
updating images. Self-merging.

PR #3946, opened 2024-04-16
docs: ensure docs capture details about paid_by_us field

consideRatio:
- Fixes #3946

Ends up not doing anything really as it was already done, besides cleanup of other things.


GeorgianaElena (approved):
🚀

PR #3948, opened 2024-04-16
automation: add workflow to fail on warnings during docs build

### LINKED ISSUE #3945: Add strict check of sphinx build to catch broken references etc
Typically when sphinx build, warning tends to be relevant and worth fixing. Due to that, we should have a strict check during build.

image

Note that this isn't the same as a linkcheck - that is a separate kind of check.

consideRatio:
- Fixes #3945

I saw we had many sphinx build warnings that I figured we shouldn't have, because they are typically very relevant and introduced as part of changes into docs. This workflow will ensure we don't introduce warnings.


GeorgianaElena (approved):
Thank you @consideRatio

consideRatio:
Thank you @GeorgianaElena!

PR #3951, opened 2024-04-16
opensci, sciencecore: update pangeo-notebook and scipy-notebook image tags

consideRatio:
The pangeo-notebook image is vulnerable (but dynamically patched), it felt like a good idea to get them updated anyhow.

### DIFF
--- a/config/clusters/opensci/sciencecore.values.yaml
+++ b/config/clusters/opensci/sciencecore.values.yaml
@@ -62,7 +62,7 @@ jupyterhub:
                 display_name: Pangeo Notebook Image
                 description: "Python image with scientific, dask and geospatial tools"
                 kubespawner_override:
-                  image: pangeo/pangeo-notebook:2023.09.11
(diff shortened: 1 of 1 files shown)


config/clusters/opensci/sciencecore.values.yaml
```diff
@@ -48,7 +48,7 @@ jupyterhub:
                 display_name: Pangeo Notebook Image
                 description: "Python image with scientific, dask and geospatial tools"
                 kubespawner_override:
-                  image: pangeo/pangeo-notebook:2023.09.11
+                  image: pangeo/pangeo-notebook:2024.04.08
```
sgibson91, line 51: Should this be quay.io too?

sgibson91: Nope. Because the other image is jupyter... Jet lagged eyeessss

sgibson91 (approved):

PR #3963, opened 2024-04-18
[kitware-staging] Fix deployment

GeorgianaElena:
Forgot to push this commit in #3961


GeorgianaElena:
Self merging to fix the deployment

PR #3972, opened 2024-04-20
More explicitly list safe config changes to merge

yuvipanda:
Code review work is also work, and especially for a split TZ team like ours, latency here can be a real velocity killer. A lot of config changes only take a 'rubber stamp' approval, because someone still needs to verify that the change actually propagated.

This PR more explicitly lists some of these config changes that don't need review, as there isn't any effective way for someone to really provide review here. This list will get bigger over time, and should give us more confidence to move faster.


consideRatio (approved):
Good separation on "when asked by the community"

sgibson91 (approved):

PR #3976, opened 2024-04-22
Update version of cryptnono image

yuvipanda:
Brings in https://github.com/cryptnono/cryptnono/pull/33 temporarily, until it gets reviewed and merged.

### DIFF
--- a/helm-charts/support/values.yaml
+++ b/helm-charts/support/values.yaml
@@ -432,6 +432,10 @@ cryptnono:
         cpu: 5m
         memory: 100Mi
 
+  image:
+    # Brings in https://github.com/cryptnono/cryptnono/pull/33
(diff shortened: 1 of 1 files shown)


helm-charts/support/values.yaml
```diff
@@ -432,6 +432,10 @@ cryptnono:
         cpu: 5m
         memory: 100Mi
 
+  image:
+    # Brings in https://github.com/cryptnono/cryptnono/pull/33
```
sgibson91, line 436: Maybe should have a FIXME note to update/change again once that PR is merged?

sgibson91 (approved):

helm-charts/support/values.yaml
yuvipanda: @sgibson91 good idea, added!

Issue #3864, opened 2024-03-27
Ensure our buckets are setup consistently for all hubs


consideRatio:
There are more buckets setup than we have misc other things setup.

For each hub with buckets setup via terraform, we should ensure the config aligns to make use of it. There are three things I know needs to be setup.

- environment variables
   ```yaml
     singleuser:
       extraEnv:
         SCRATCH_BUCKET: s3://scratch-go-bgc/$(JUPYTERHUB_USER)
   ```
   - [x] All AWS hubs (via #3977)
   - [ ] All GCP hubs
- k8s ServiceAccount cloud specific annotations to grant access
   Via `userServiceAccount.annotations`
   - [x] All AWS hubs (via #3977)
   - [ ] All GCP hubs
- singleuser.cloudMetadata.blockCloudMetadata
   ```yaml
     singleuser:
       cloudMetadata:
         blockWithIptables: false
   ```
   - [ ] All AWS hubs
   - [ ] All GCP hubs

PR #3977, opened 2024-04-22
aws hubs: consistent setup of cloud permissions and bucket envs

consideRatio:
I saw that we had several hubs where buckets were setup in terraform, but not made available to hubs. This PR setups access to the buckets.

Does parts of whats needed for #3864 - looked at all AWS clusters' hubs.

- This PR hasn't verified access works, and for several hubs I think it won't work due to `singleuser.cloudMetadata.blockWithIptables` defaulting to false, for example for opensci's staging hub.
- This PR has used terraform as the source of truth if hubs should have access to buckets - so if a scratch bucket was defined for staging, then staging should have access to if for example. So, we could opt to instead delete the buckets from terraform as an alternative to setting up access to it.


consideRatio:
Still relevant but not championed further

Issue #3978, opened 2024-04-23
[Feature Enablement] Allow access to an external S3 bucket from CryoCloud


yuvipanda:
# Source of request

https://2i2c.freshdesk.com/a/tickets/1508

# Feature to enable

Allow access to buckets in other AWS projects from the hub. https://infrastructure.2i2c.org/howto/features/buckets/#granting-access-to-cloud-buckets-in-other-cloud-accounts-projects

# Hub

`nasa-cryo`, staging and production hubs.

# Information needed for feature enablement to be completed

| Question | Answer | Note |
| - | - | - |
| Name of bucket | `sliderule-public` | |
| Has the appropriate policy been attached at the source? | yes | step 1 & 2 have been completed, we can go to 3 |

# How to test

We can test on the staging hub, to avoid disruption

1. Terraform apply and manually deploy to staging
2. Install [aws cli](https://infrastructure.2i2c.org/howto/features/buckets/#testing-access-to-buckets) in the image if needed.
3. Run `aws s3 ls s3://<name-of-bucket>` to see if it works

yuvipanda:
I'm also testing how our new 'definitons of ready' described via https://github.com/2i2c-org/meta/issues/953 can help here. This is a possible template for 'Feature enablement', that we can tweak.

consideRatio:
I think we lack some docs to help the community managing the external bucket to grant access to the specific IAM role for users on the hub.

I think historically, maybe the assumption has been that they grant public __read__ access as long as __requester pays__ for it or similar, but they may want to grant only the hub user's associated IAM role access (read __or readwrite__ etc), __not necessarily to anyone__.

I suspect that the docs currently written as...

```
3. In the `.tfvars` file for the cluster hosting the hub, add `extra_iam_policy`
   as a key to the hub under `hub_cloud_permissions`. This is used to set any additional
   IAM permissions granted to the users of the hub. In this case, you should copy the
   exact policy that was applied to the bucket in step 2, but remove the "Principal" key.
   So it would look something like:
```

... we may actually just be needing to specify that we allow our IAM role to pay for access costs to the external bucket or similar, and that the permissions on what kind of access we have, is determined in the external bucket's access policy.

PR #3983, opened 2024-04-24
Turn documentation on how to do cross-bucket S3 access into a runbook

yuvipanda:
Rewrite documentation on how to do cross-org S3 access to be a runbook.

Based on lessons learnt from https://github.com/2i2c-org/infrastructure/issues/3978

### DIFF
--- a/docs/howto/features/buckets.md
+++ b/docs/howto/features/buckets.md
@@ -282,8 +282,7 @@ On AWS, we would need to set up [cross account S3 access](https://aws.amazon.com
    find this under `userServiceAccount.annotations.eks.amazon.com/role-arn` in
    the `values.yaml` file for your hub. It should look something like
    `arn:aws:iam::<account-id>:role/<hub-name>`.
-2. In the AWS account *with the S3 bucket*, [create an IAM
-   policy](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_create-console.html)
+2. In the AWS account *with the S3 bucket*, you need to [attach a bucket policy](https://docs.aws.amazon.com/AmazonS3/latest/userguide/add-bucket-policy.html)
    that grants appropriate access to the S3 bucket from the hub. For example, the
    following policy grants readonly access to the bucket for users of the hub
 
@@ -316,6 +315,13 @@ On AWS, we would need to set up [cross account S3 access](https://aws.amazon.com
    You can list as many buckets as you want, but each bucket needs two entries -
    one with the `/*` and one without so both listing the bucket as well as fetching
    data from it can work
+   ```
+
+   ```{warning}
+
+   You must add this as a bucket policy, not a regular IAM Policy. If you try to add
+   this as a regular IAM Policy, you may get an error about `Principal` not being
+   allowed in the policy document.
    ```
 
 3. In the `.tfvars` file for the cluster hosting the hub, add `extra_iam_policy`
--- a/docs/howto/features/buckets.md
+++ b/docs/howto/features/buckets.md
@@ -276,15 +276,26 @@ other cloud providers when needed.
 
 ### AWS
 
-On AWS, we would need to set up [cross account S3 access](https://aws.amazon.com/premiumsupport/knowledge-center/cross-account-access-s3/).
+On AWS, we would need to set up [cross account S3 access](https://aws.amazon.com/premiumsupport/knowledge-center/cross-account-access-s3/). It requires action from both the 2i2c engineering team as
+well as the the community representative.
 
-1. Find the ARN of the service account used by the *users* on the hub. You can
-   find this under `userServiceAccount.annotations.eks.amazon.com/role-arn` in
-   the `values.yaml` file for your hub. It should look something like
+1. **Who**: 2i2c Engineering Team
+
+   Find the ARN of the service account that is assumed by the users on the hub,
+   under `userServiceAccount.annotations.eks.amazon.com/role-arn` in the YAML
+   config specific to the hub. It should look something like
    `arn:aws:iam::<account-id>:role/<hub-name>`.
-2. In the AWS account *with the S3 bucket*, you need to [attach a bucket policy](https://docs.aws.amazon.com/AmazonS3/latest/userguide/add-bucket-policy.html)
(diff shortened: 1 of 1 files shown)


docs/howto/features/buckets.md
```diff
@@ -282,8 +282,7 @@ On AWS, we would need to set up [cross account S3 access](https://aws.amazon.com
    find this under `userServiceAccount.annotations.eks.amazon.com/role-arn` in
    the `values.yaml` file for your hub. It should look something like
    `arn:aws:iam::<account-id>:role/<hub-name>`.
-2. In the AWS account *with the S3 bucket*, [create an IAM
-   policy](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_create-console.html)
+2. In the AWS account *with the S3 bucket*, you need to [attach a bucket policy](https://docs.aws.amazon.com/AmazonS3/latest/userguide/add-bucket-policy.html)
    that grants appropriate access to the S3 bucket from the hub. For example, the
```
consideRatio, line 286:
```suggestion
   that grants appropriate access to the IAM role used by the hub users to work against the bucket. For example, the
```

docs/howto/features/buckets.md
```diff
@@ -282,8 +282,7 @@ On AWS, we would need to set up [cross account S3 access](https://aws.amazon.com
...
    following policy grants readonly access to the bucket for users of the hub
```
consideRatio, line 287:
```suggestion
   that grants appropriate access to the IAM role used by the hub users to work against the bucket. For example, the
   following bucket policy grants readonly access to the bucket for users of the hub via an IAM role they work with.
```

consideRatio (dismissed):

sgibson91 (dismissed):

yuvipanda:
I've completely rewritten this, to follow the runbook pattern. Another request of the same sort just came in (https://2i2c.freshdesk.com/a/tickets/1547), and I'm responding to that using the template provided here. Would appreciate a -re-review.

docs/howto/features/buckets.md
```diff
@@ -310,19 +320,60 @@ On AWS, we would need to set up [cross account S3 access](https://aws.amazon.com
    }
    ```
 
-   You can add additional permissions to the bucket if needed here.
+   here,
+   1. `Statement.Principal.AWS` is the ARN determined from step 1. This is the only item that needs
+      to be supplied by the 2i2c engineer.
```
consideRatio, line 325:
```suggestion
   1. `Statement.Principal.AWS` lists the ARNs determined from step 1. This is the only info that needs
      to be supplied by the 2i2c engineer.
```

docs/howto/features/buckets.md
```diff
@@ -276,16 +276,26 @@ other cloud providers when needed.
...
+   If this hub has an equivalent *staging* hub, find the ARN of the service
+   account for the staging hub too. This helps us keep staging and prod in sync.
+
+2. **Who**: 2i2c Engineering Team
+
+   Generate a [Bucket Policy](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-iam-policies.html)
+   that can be communicated to the community champion. You need to generate one policy
+   for each ARN determined in step 1. The `Action` section lists various permissions to be
+   granted. The default is to allow hub users to list all objects and get data - so readonly
+   access. The community champion is welcome to modify this to fit their needs.
```
consideRatio, line 298:
```suggestion
   for each ARN determined in step 1.
```

The description of action ocurring twice (but not principal/resource) - I suggest removing a mention of what action does here and relying on the description below instead saying the same thing.

docs/howto/features/buckets.md
```diff
@@ -310,19 +320,60 @@ On AWS, we would need to set up [cross account S3 access](https://aws.amazon.com
...
+   You need to fill in the list of S3 buckets you want granted access under the "Resource" section.
+   Note that there are two entries for each bucket.
+
+   This policy, as written, will grant readonly access to the buckets from the hub. Users
+   on the hub will be able to list objects and retrieve them. If you'd like to grant other
+   kinds of access, adjust the list under 'Actions' accordingly.
+
+   AWS has [some documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/add-bucket-policy.html)
+   on how to attach this bucket policy. Once attached, can you give us the full content of the
+   policy you attached? That would help us take next steps in granting access to these buckets.
```
consideRatio, line 358: Since its only possible to have one bucket policy (but multiple statements in it), the bucket admins are required to merge any existing policy with the policy we provide - by appending statements and avoiding statement duplicates.

The text needs to be updated to reflect this.

docs/howto/features/buckets.md
```diff
@@ -310,19 +320,60 @@ On AWS, we would need to set up [cross account S3 access](https://aws.amazon.com
...
+
+   In the `.tfvars` file for the cluster hosting the hub, add `extra_iam_policy`
    as a key to the hub under `hub_cloud_permissions`. This is used to set any additional
-   IAM permissions granted to the users of the hub. In this case, you should copy the
-   exact policy that was applied to the bucket in step 2, but remove the "Principal" key.
-   So it would look something like:
+   IAM permissions granted to the users of the hub. In this case, we use `s3:*` to allow
+   all S3 actions on the given list of buckets, so the exact permissions granted are
+   controlled by the bucket policy set in earlier steps. This allows the community to grant
+   more (or less) permissions in the future without any intervention on our part.
```
consideRatio, line 376: I understand that this works, but I don't feel confident on what permissions we need to declare on the IAM Role already being granted permissions to work against the bucket in the bucket policy.

- Do we need to grant the bucket actions on the IAM role as well?
- Do we need to grant permissions to the IAM role to pay for access of this bucket if requestor pays is enabled on the bucket - if so, how is paying for bucket access configured? Is that perhaps under `s3:` something, and that is the reason we have needed `s3:*`?

If you are uncertain about this as well, I think it would be good to detail that this is done as a "ah this works" kind of solution as compared to "this is required" kind of solution.

consideRatio (changes requested):

yuvipanda:
I'm just going to close this as it isn't a high priority for me right now. Next time a support ticket comes in requesting this, I'll get someone else to do this (and fix the docs along the way).

Issue #3989, opened 2024-04-24
Automate cloud billing CSV transformation (GCP + AWS)


yuvipanda:
Part of https://github.com/2i2c-org/infrastructure/issues/3958

Given we know what CS&S needs to bill, and we already have manual instructions for cloud cost data (insert link), write a deployer subcommand that takes as input:

1. A local path to CSV file generated from the cloud UI
2. Transposes that to the format CS&S wants
3. Outputs a CSV file to be manually uploaded to appropriate google drive location

After this, we update the manual billing instructions with these two extra steps.

These will be two different commands, so they will output two different CSVs - one for GCP, one for AWS. We will ignore Azure. 

We will create a google drive directory that will be linked to from the instructions as the place to upload. A naming convention will be codified by the output file name of the script, but it will also be documented as well.

This will likely be a different 'verb' / command (transform), rather than be added on to the 'generate' verb we have

# Considerations

1. Outputting a new CSV file instead of merging with existing billing google sheet, to avoid clobbering that sheet and to avoid it becoming unmanageble over time as more stuff gets apended to it. As this is not scoped to integrate with the google sheets api (with the `gsheets` package or otherwise), doing this would be out of scope as well.
2. We want to keep this simple and scoped, so we are not going to automate uploading this to google sheets
3. We can come back to the bigquery automation later if needed, but this will remove chances of copy paste errors without adding another human into the loop (for verification / independent control) or depending on more complex automation (bigquery).

# Schema of the output CSV

1. Name of project
2. One column for each month found in the input CSV - direct copy of the 'subtotal' column from the input CSV (GCP), for AWS (to be determined)
3. One column with total $$$ for all the things found

(@sgibson91 wrote most of this, I was just transcribing)

sgibson91:
Super rough draft of this for GCP available in https://github.com/2i2c-org/infrastructure/pull/3992

sgibson91:
The technical aspect of this issue is ready to be reviewed in https://github.com/2i2c-org/infrastructure/pull/3992
I will open another PR to update our process docs https://infrastructure.2i2c.org/howto/bill/

sgibson91:
Documentation is ready to be reviewed https://github.com/2i2c-org/infrastructure/pull/4063

sgibson91:
[...]

PR #3992, opened 2024-04-25
Automate cloud billing CSV transformation for dedicated clusters on GCP and AWS

sgibson91:
- https://github.com/2i2c-org/infrastructure/issues/3989

---

### 2024-05-09

New functions for AWS and GCP have been added under a `transform cost-table` sub-command that will transform a CSV file generated in the cloud provider's UI into the format required by CS&S and dump this file locally. These new commands have been documented in the deployer's readme.

Ready for review! 🎉 

### 2024-05-07

I tried migrating the toy script I wrote into the deployer. I am having difficulty setting up the Typer application to give the expected CLI functionality.

I successfully added a new top-level verb called `transform`:

```
$ deployer --help
                                                                                                                                                                                        
 Usage: deployer [OPTIONS] COMMAND [ARGS]...                                                                                                                                            
                                                                                                                                                                                        
╭─ Options ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --install-completion          Install completion for the current shell.                                                                                                              │
│ --show-completion             Show completion for the current shell, to copy it or customize the installation.                                                                       │
│ --help                        Show this message and exit.                                                                                                                            │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Commands ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ cilogon-client            Manage cilogon clients for hubs' authentication.                                                                                                           │
│ debug                     Debug issues by accessing different components and their logs                                                                                              │
│ decrypt-age               Decrypt secrets sent to `support@2i2c.org` via `age`                                                                                                       │
│ deploy                    Deploy one or more hubs in a given cluster                                                                                                                 │
│ deploy-support            Deploy support components to a cluster                                                                                                                     │
│ exec                      Execute a shell in various parts of the infra. It can be used for poking around, or debugging issues.                                                      │
│ generate                  Generate various types of assets. It currently supports generating files related to billing, new dedicated clusters, helm upgrade strategies and resource  │
│                           allocation.                                                                                                                                                │
│ grafana                   Manages Grafana related workflows.                                                                                                                         │
│ run-hub-health-check      Run a health check on a given hub on a given cluster. Optionally check scaling of dask workers if the hub is a daskhub.                                    │
│ **transform                 Programatically transform datasets, such as cost tables for billing purposes.**                                                                              │
│ use-cluster-credentials   Pop a new shell or execute a command after authenticating to the given cluster using the deployer's credentials                                            │
│ validate                  Validate configuration files such as helm chart values and cluster.yaml files.                                                                             │
╰───────────────────────────────────────────
```

However, the nested sub-command I am trying to add (`cost-table` is not showing up:

```
$ deployer transform --help
                                                                                                                                                                                        
 Usage: deployer transform [OPTIONS] COMMAND [ARGS]...                                                                                                                                  
                                                                                                                                                                                        
 Programatically transform datasets, such as cost tables for billing purposes.                                                                                                          
                                                                                                                                                                                        
╭─ Options ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --help          Show this message and exit.                                                                                                                                          │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

```
$ deployer transform cost-table --help
Usage: deployer transform [OPTIONS] COMMAND [ARGS]...
Try 'deployer transform --help' for help.
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ No such command 'cost-table'.                                                                                                                                                        │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

I have been copying code from `generate dedicated-cluster` and I'm not sure why this isn't working.

### 2024-04-25

This is a _super_ rough draft that just handles GCP CSV input and does the corresponding pandas stuff to get it into the CS&S desired schema. Output looks like (blurred out the specific numbers just in case):

![transorm-gcp-cost-tables-output](https://github.com/2i2c-org/infrastructure/assets/44771837/a0b26902-7560-42e6-862c-0ff40ebdcac8)

I've written this in a very "brain logical" way (at least for my brain), which might not be very Pythonic or utilising pandas to it's fullest. So any tips or tricks in that regard are welcome, now that I've verified the output we want is possible.

Should we maintain a list excluding some of these project names, e.g., two-eye-two-see since that has the shared clusters on it and this work is for the dedicated clusters? **ETA:** In discussion with Yuvi, we came to the conclusion that it's probably best to minimise the amount of manipulation to do since we are already running the risk that GCP will change the column headings in it's output. Instead we will create documentation for CS&S that points out the inclusion of _all_ projects and notes any shared or "default" ones that can be ignored.

#### Example of what the script is trying to achieve
For GCP billing, we aim to transform an input CSV which looks like:

```
Project name, Month, Subtotal ($), ...
project-A, 2024-01, 123, ...
project-A, 2024-02, 456, ...
project-B, 2024-01, 789, ...
project-B, 2024-02, 147, ...
```

into an output CSV that looks like:

```
Project name, 2024-01, 2024-02, Total
project-A, 123, 456, 579
project-B, 789, 147, 936
```

where the `Total` column is the sum of the `2024-01` and `2024-02` columns.

Exact months and project names are subject to change.


transform-gcp-cost-tables.py
```diff
@@ -0,0 +1,82 @@
...
+# Read input file
+input_cost_table = pd.read_csv(
+    input_filepath,
+    usecols=["Month", "Project name", "Subtotal ($)"],
+)
+
+# Drop NaN values
+input_cost_table = input_cost_table.dropna()
+
+# Rename the projects column. This makes using the 'query' function later easier
```
yuvipanda, line 41: Filipe suggests:

> You can also rename a column with the .rename instead of copying and dropping. Not that different but it will look more "panda-ish."

yuvipanda:
Filipe from the pyopensci slack suggests:

> I use https://pypi.org/project/pandas-vet/ to help with some basic stuff. Like avoiding inplace, and checking some transformation that can be problematic.

Titus Brown from the pyopensci slack suggests:

> It looks really nice and clear to me. There are things in there that I might research to figure out if there’s a more pandas-onic way of doing it (the sorted(…tolist()) stuff for example) but it matches what I would do and I think it’s really nice.

yuvipanda:
From Dave Nicholson on the pyopensci slack

>  I also cannot wrap my brain around the pandas API but I think the more idiomatic way to do it would be something like pivot_table -- to use "tidy data" lingo, you are going from a "long" table format (where Month is the variable and you have values for it) to a "wide" format (where every month becomes a column, and then you "aggregate" across them)
> See https://pandas.pydata.org/docs/user_guide/reshaping.html and https://jakevdp.github.io/PythonDataScienceHandbook/03.09-pivot-tables.html
> So what Sarah has works fine but the more Pandas native way would be chaining some set of operations that starts with a
df.pivot_table
> Since I am an imposter Pandas user my first impulse would be to do something like
> `pd.groupby('Month').agreggate('sum')` kind of like Jake says at the top of that chapter.

yuvipanda:
Code sample from Neils Bantilan on the pyopensci slack:

```python
import pandas as pd
from io import StringIO


data = StringIO("""Project name, Month, Subtotal ($)
project-A, 2024-01, 123
project-A, 2024-02, 456
project-B, 2024-01, 789
project-B, 2024-02, 147
""")

df = (
    pd.read_csv(data)
    .rename(columns=lambda col: col.strip().lower().replace(" ", "_"))
)

assert [*df.columns] == ["project_name", "month", "subtotal_($)"]

agg = (
    df.groupby(["project_name", "month"]).sum()
    .pivot_table(index="project_name", columns="month", values="subtotal_($)")
    .assign(total=lambda df: df.sum(axis=1))
)
print(agg)
"""
month          2024-01   2024-02  total
project_name
project-A          123       456    579
project-B          789       147    936
"""
```

with this comment:

> I’m somewhat of a method-chaining maximalist, but feel free to break it out into separate variable-assignment steps if it’s more readable

NickleDave:
:+1: for @cosmicBboy (Neils)'s version! (Although what @sgibson91 had worked fine of course. We can at least pretend that Pythonistas are not several communities divided by a common language :stuck_out_tongue:)

sgibson91:
Thank you everyone for your feedback! I'm looking forward to working on this more and learning some new things!

GeorgianaElena (commented):
> However, the nested sub-command I am trying to add (cost-table is not showing up:

@sgibson91 I believe the bit that's missing is importing the sub-command in this file:

https://github.com/2i2c-org/infrastructure/blob/main/deployer/__main__.py

I also think it's worth mentioning this in the readme here https://github.com/2i2c-org/infrastructure/blob/main/deployer/README.md#the-directory-structure-of-the-deployer-package? (Also sorry for not have documented this more clearly :( )

sgibson91:
> I also think it's worth mentioning this in the readme here [`main`/deployer/README.md#the-directory-structure-of-the-deployer-package](https://github.com/2i2c-org/infrastructure/blob/main/deployer/README.md?rgh-link-date=2024-05-07T14%3A22%3A31Z#the-directory-structure-of-the-deployer-package)?

@GeorgianaElena I tried to address this in [`69cacc4` (#3992)](https://github.com/2i2c-org/infrastructure/pull/3992/commits/69cacc4dacc0b077012fa93cd34b3e9c3f792b5b)

deployer/commands/transform/cost_table.py
```diff
@@ -0,0 +1,161 @@
...
+    # Ensure values of the linked_account_name column are lower case and any
+    # whitespace is replaced with underscores.
+    df["linked_account_name"] = df["linked_account_name"].apply(
+        lambda val: val.strip().lower().replace(" ", "_")
+    )
+
+    # Set the linked_account_name column as the index
+    df.set_index("linked_account_name", drop=True, inplace=True)
+
+    # Drop the total_costs column
```
sgibson91, line 64:
```suggestion
    # Drop the total_costs column. This column is the total across all linked accounts,
    # and hence not necessary. We use the linked_account_total column for the total
    # per linked account.
```

deployer/commands/transform/cost_table.py
```diff
@@ -0,0 +1,163 @@
...
+    # project names.
+    df = pd.read_csv(
+        input_path,
+        skiprows=1,
+    ).rename(columns=lambda col: col.strip().lower().replace(" ", "_").strip("_($)"))
+
+    # Ensure values of the linked_account_name column are lower case and any
+    # whitespace is replaced with underscores.
+    df["linked_account_name"] = df["linked_account_name"].apply(
+        lambda val: val.strip().lower().replace(" ", "_")
```
yuvipanda, line 58: spaces are allowed here apparently (https://docs.aws.amazon.com/organizations/latest/APIReference/API_Account.html). But am curious about the reason for us doing this transformation. Wouldn't it cause confusion if we were ever to look for the account name in the UI based on what's here?

yuvipanda (commented):
As someone who doesn't know a lot of pandas, i appreciate all the comments in here, @sgibson91. I've one question about underscores, but otherwise this looks good.

deployer/commands/transform/cost_table.py
sgibson91: This is a quirk of how AWS outputs the data, how pandas infers columns, and the fact we haven't transposed the dataframe yet.

AWS outputs the linked account name as a _row_, but pandas reads the first row as column headers. So this line of code is actually affecting the column that has the months in it and some other random stuff we don't care about, not the account names.

Screenshot 2024-05-14 at 10 26 21

deployer/commands/transform/cost_table.py
```diff
@@ -0,0 +1,163 @@
...
+    # Rename the columns so that they are all lower case and any whitespace is
+    # replaced with underscores. Then strip '_($)' to ensure we just have the
+    # project names.
+    df = pd.read_csv(
+        input_path,
+        skiprows=1,
+    ).rename(columns=lambda col: col.strip().lower().replace(" ", "_").strip("_($)"))
+
+    # Ensure values of the linked_account_name column are lower case and any
+    # whitespace is replaced with underscores.
```
sgibson91, line 56:
```suggestion
    # Ensure values of the linked_account_name column are lower case and any
    # whitespace is replaced with underscores.
    # Note that because AWS outputs the linked account names as rows, this
    # column name is misleading - we are not affecting linked account names here.
```

deployer/commands/transform/cost_table.py
sgibson91: I expanded the comment [`3458571` (#3992)](https://github.com/2i2c-org/infrastructure/pull/3992/commits/345857182d86e5879061c6b4f44b51aa0ec116f9)

sgibson91: I also pushed [`96f4dfd` (#3992)](https://github.com/2i2c-org/infrastructure/pull/3992/commits/96f4dfdb04eefaad3c1051dfd3031fb4834604ab) to address your concern that in the future we may have an account name with whitespace in it

sgibson91: Generally with pandas, it's nice not to have whitespace in column names and indices, because then syntax like `df.column_name` works, as well as `df["column_name"]` - but this _only_ works if there is no whitespace.

PR #3998, opened 2024-04-27
Update new image tag

benlee0423:
1. Use newer version of base image which fixes [jupyter-server-proxy](https://github.com/jupyterhub/jupyter-server-proxy/security/advisories/GHSA-w3vc-fx9p-wp4v) issue.


sgibson91 (approved):

benlee0423:
@sgibson91 
Yes, this is for staging only. I like to verify the change in staging.
It would be nice if we can verify the change even before this PR.

benlee0423:
@yuvipanda
Can you merge this PR please?

Issue #3999, opened 2024-04-29
Disable shared volume reporter for all AWS hubs


consideRatio:
We have two core nodes in 2i2c-aws-us because we have run out of pods per node on the first, even though only one would suffice for our CPU/memory needs there. EKS nodes fit fewer pods than GCP nodes so this can be a problem on EKS. So, this is a reason to reduce the number of pods there.

We have the shared-volume-reporter pods, one per hub, that reports the available disk space remaining in the shared volume, but for our EKS hubs, this is always 100% as the EFS system we rely on is unbounded. So, in practice, we can disable that on all AWS hubs until that situation changes.

![image](https://github.com/2i2c-org/infrastructure/assets/3837114/7a72b30d-424d-4c1d-9853-775f8872324d)

PR #4000, opened 2024-04-29
basehub: allow shared-volume-reporter to be disabled, and disable it on 2i2c-aws-us and catalystproject-africa

consideRatio:
Technically why this is acceptable is detailed in #3999.

2i2c-aws-us runs two core nodes, but only requires one for its CPU/memory needs. What makes us have two is the fact that we need room for 66 pods, but each core node only allows for 57 per node.

With this change, we reduce the pods to run on the core nodes with one pod per hub we have deployed --- 7 pods. And that in turn combined with other pod reductions like #3869 will get us below 57 pods to run on the core nodes, allowing us to cut cost for communities on 2i2c-aws-us (I think with almost 200USD total) and help us demonstrate more cost effective cloud operations.

## Related
- #3999 
- Another pod reduction PR: #3869

### DIFF
--- a/config/clusters/2i2c-aws-us/basehub-common.values.yaml
+++ b/config/clusters/2i2c-aws-us/basehub-common.values.yaml
@@ -1,5 +1,7 @@
 nfs:
   enabled: true
+  volumeReporter:
+    enabled: false
   pv:
     enabled: true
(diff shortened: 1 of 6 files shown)


consideRatio:
I'll go for a self-merge here, it should be safe and can help me verify that we are indeed able to cut costs in 2i2c-aws-us and catalystproject-africa by reducing node pools form 2 to 1 nodes.

- Related #4024

PR #4010, opened 2024-04-30
Update spyglass image

choldgraf:
This updates the image tag for spyglass since we've just pushed an addition of jupyterlab-myst

I'm assuming this is the right tag based on [the GitHub action logs here](https://github.com/LorenFrankLab/hhmi-spyglass-image/actions/runs/8900298360/job/24441556790):

image

And the quay.io page here:

CleanShot 2024-04-30 at 12 24 26@2x


choldgraf:
Actually, I [read the instructions on self-merging](https://infrastructure.2i2c.org/contributing/code-review/#review-and-merge-guidelines), and I think that this is OK in this case because it's a routine change (image bump) and I can revert it if necessary.

Issue #4021, opened 2024-05-02
Unify & upstream our code for profile_list differentiation


yuvipanda:
We currently have code that allows us to show people different profile list options based on some notion of group membership. The initial implementation was tied to GitHub teams membership, and hence not upstreamable into KubeSpawner (where the profile_list code lives). However, with https://github.com/2i2c-org/infrastructure/pull/3883, we have a version of that code that uses JupyterHub *groups*, and hence is now actually upstreamable.

Unifying this code and upstreaming it allows us to maintain fewer things. There's no real rush, but it's useful to make sure that whatever we do is done with an eye to upstreaming. To that end, I've already been slowly working on things that can help make this happen. Here's a linear set of items that we can do to help.

- [ ] Get https://github.com/2i2c-org/infrastructure/pull/3883 deployed
- [ ] Get https://github.com/jupyterhub/oauthenticator/pull/735 merged
- [ ] Once https://github.com/jupyterhub/oauthenticator/pull/735 is merged, setup our hubs to turn GitHub team membership into JupyterHub groups. This motivating use case is driving the design of the PR forward.
- [ ] From this point on, new hubs can simply use `allowed_groups`, and `allowed_teams` can be considered legacy.
- [ ] Upstream the `allowed_groups` code into kubespawner. 
- [ ] Migrate existing hubs from `allowed_teams` to `allowed_groups` via some strategy (slowly, one every sprint, or quickly, in one go). It requires everyone to re-login so group membership is properly set up
- [ ] Rip out the `allowed_groups` code
- [ ] Pat ourselves on the back.

Issue #3900, opened 2024-04-03
[EPIC] EarthScope: Restrict access to profiles based on scopes granted


yuvipanda:
EarthScope wants to restrict access to profiles based on specific scopes granted to them.

I think the sustainable pathway to this is to do this based on *group* membership. This matches with a lot of the new work going on upstream. We can start this out in Earthscope, but eventually be able to expand this in other places. It also will eventually help us get rid of the hacky setup for restricting access to profiles based on GitHub Teams (https://github.com/2i2c-org/infrastructure/blob/120e88dfe77fe6b32520bbfc14297c5b2010cd3f/helm-charts/basehub/values.yaml#L979), but that's not a direct part of this ticket.

https://2i2c.freshdesk.com/a/tickets/1453 has additional information, and is our ongoing conversation with them.

The steps here are:

## ✅  Deploy latest OAuthenticator

https://github.com/jupyterhub/oauthenticator/pull/708 was merged a couple months ago, allowing OAuthenticator to manage JupyterHub groups based on results from the OAuth2 provider. This I believe will be deployed whenever https://github.com/2i2c-org/infrastructure/pull/3818 goes out. This allows us some more information about group management.

## Switch to using `GenericOAuthenticator` for using Auth0

https://github.com/jupyterhub/oauthenticator/pull/708 is specifically for GenericOAuthenticator. In https://github.com/2i2c-org/infrastructure/pull/3883, we move EarthScope from Auth0Authenticator to GenericOAuthenticator, so we can make use of this feature. 

Note: While there's upstream work from me to move this functionality away from just GenericOAuthenticator (https://github.com/jupyterhub/oauthenticator/pull/735), I definitely do **not** want us to block this work on that. Those changes will be made in a backwards compatible way, and we can roll them out as and when that gets released.

## Enable mapping granted scopes to JupyterHub user groups

This would be setting the equivalent of the following config:

```
GenericOAuthenticator:
  claims_groups_key: "scope"
  manage_groups: true
```

This should turn scopes granted by oauth2 into groups, and we should be able to see that in the JupyterHub admin UI.

## Enable restricting scopes based on JupyterHub user groups

This would likely be a straight port of our GitHub code (https://github.com/2i2c-org/infrastructure/blob/120e88dfe77fe6b32520bbfc14297c5b2010cd3f/helm-charts/basehub/values.yaml#L979) but to use JupyterHub groups. At this point, this will be only in earthscope, but primed for proper upstream contribution.
[...]

PR #4022, opened 2024-05-02
Remove feature for restricting profile options based on github org membership

yuvipanda:
This gets rid of restricting profile options based on github *org* membership, rather than *teams* membership. This is a complete no-op for any of our communities, as it turns out nobody is using this feature! Getting rid of this makes sure that new communities will not use this feature, as it relies on internal implementation details of the GitHub authenticator and may be hard to upstream when the time comes. See https://github.com/2i2c-org/infrastructure/issues/4021 for upstreaming plan.

Ref https://github.com/2i2c-org/infrastructure/issues/3900
https://github.com/2i2c-org/infrastructure/pull/3883 is based on this.

### DIFF
--- a/docs/hub-deployment-guide/configure-auth/github-orgs.md
+++ b/docs/hub-deployment-guide/configure-auth/github-orgs.md
@@ -66,30 +66,9 @@ You can remove yourself from the org once you have confirmed that login is worki
      ...
    ```
 
-4. **Edit the non-secret config under `config/clusters/<cluster_name>/<hub_name>.values.yaml`.**
-   You should make sure the matching hub config takes one of the following forms.
-
-   To authenticate against a GitHub organisation (Note the `read:user` scope. See comment box below.):
-
-    ```yaml
-    jupyterhub:
-      custom:
-        2i2c:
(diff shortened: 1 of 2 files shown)


docs/hub-deployment-guide/configure-auth/github-orgs.md
GeorgianaElena, line 1: I think there are a couple more places where this file needs changes like:

- the title, as it mentions organization based access (or keep the tile, but add a warning right below it as it is not longer supported)
- point 4 that says `**Edit the non-secret config under `config/clusters/<cluster_name>/<hub_name>.values.yaml`.**` has a `scopes` example specific to orgnanizations (line 72 to line 90)

GeorgianaElena (approved):
Thanks @yuvipanda 🚀

docs/hub-deployment-guide/configure-auth/github-orgs.md
yuvipanda: good catch @GeorgianaElena! I cleaned that out and removed a bunch of options there as well, as we have nowhere we use `read:user` at all - it's `read:org` everywhere. Let's just standardize to that.

PR #4026, opened 2024-05-03
docs: Add guide on how to 'reset' a hub

yuvipanda:
I just performed this based on request in https://2i2c.freshdesk.com/a/tickets/1521, and I think we would be getting more requests like this in the future.

### DIFF
--- a/docs/sre-guide/support/reset-hub.md
+++ b/docs/sre-guide/support/reset-hub.md
@@ -18,7 +18,7 @@ We are going to "reset" the hub. This would do the following:
 
    {{ list of admin users from the config.yaml file for this hub}}
 
-This will also forcibly log and stop the server of any user running at that time.
+This will also forcibly log out and stop the server of any user running at that time.
 
 Let us know if this sounds ok, and we will proceed!
 ```
@@ -30,9 +30,6 @@ Once they respond affirmatively, you can begin.
 This clears up space and reduces their cloud costs.
 
 1. Get a shell with the home directories mounted
-
-   ``b
-1. Delete *all* home directories.
 
    ```bash
    deployer exec homes <cluster-name> <hub-name>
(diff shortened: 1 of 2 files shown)


docs/sre-guide/support/reset-hub.md
```diff
@@ -0,0 +1,109 @@
...
+
+1. Remove **all** user home directories. The *shared* directories would not be
+   touched, but you can clean them up yourself if you need.
+
+2. Remove all the users from the admin view, except 2i2c engineers and those explicitly
+   marked as admins via our configuration. Those are:
+
+   {{ list of admin users from the config.yaml file for this hub}}
+
+This will also forcibly log and stop the server of any user running at that time.
```
sgibson91, line 21:
```suggestion
This will also forcibly log out and stop the server of any user running at that time.
```

docs/sre-guide/support/reset-hub.md
```diff
@@ -0,0 +1,109 @@
...
+This clears up space and reduces their cloud costs.
+
+1. Get a shell with the home directories mounted
+
+   ``b
+1. Delete *all* home directories.
+
+   ```bash
+   deployer exec homes <cluster-name> <hub-name>
+   ```
```
sgibson91, line 39: I think there's been some mis-formatting here. I think the deployer command needs to be above the `Delete *all* home directories` list point, and there should be a different command below it.

sgibson91, line 39:
```suggestion
1. Get a shell with the home directories mounted

   ```bash
   deployer exec homes <cluster-name> <hub-name>
   ```
```

Maybe this is the correct thing, as I see there is a check for `shared` directories below.

sgibson91 (approved):

yuvipanda:
Almost tested and verified while doing https://2i2c.freshdesk.com/a/tickets/1731 - however, I've opted to ask for reconfirmation from this community.

Issue #4014, opened 2024-05-01
[earthscope] Restrict access to dask-gateway JupyterHub service based on group membership


yuvipanda:
Part of https://github.com/2i2c-org/infrastructure/issues/3899

Not all users should be able to get access to dask-gateway - this should also be limited based on scopes granted to them.

Productboard link: https://2i2c.productboard.com/products/features/27020229/detail

This is something we have committed to as part of the SOW.

## Dependencies

https://github.com/2i2c-org/infrastructure/issues/3900, which provisions JupyterHub groups based on scopes granted to users. 

## Tasklist

```[tasklist]
- [ ] Figure out how to restrict access to a particular JupyterHub service based on group membership in JupyterHub (https://2i2c.slack.com/archives/CKJS000F4/p1712121603619569?thread_ts=1712108545.029909&cid=CKJS000F4)
- [ ] Cleanup our [loadRoles](https://github.com/2i2c-org/infrastructure/blob/6617eaed5f20babcbfffdc32a099375e524cd922/helm-charts/basehub/values.yaml#L593) to explicitly grant access for the services we care about (configurator and dask-gateway)
- [ ] Provide override in earthscope hub that only allows users of a specific group to access that service
```

consideRatio:
I ran into an issue with the strategy planned that require exploration of dask-gateway development, as outlined on a very high level in https://github.com/dask/dask-gateway/issues/829.

With that and no quick alterantives in my mind, this issue ballooned into something that won't resolve in this sprint =/

For now I'll work to offload insights about JupyterHub.load_roles and assumptions in our config for configurator/dask-gateway that doesn't seem true - without a crisp understanding about that we will struggle in the end anyhow with this issue.

yuvipanda:
@consideRatio ah damn. I'll communicate this to the community and see what we can do.

Gman0909:
Has there been any update on this from the communities? If not, does this issue need to be sent back to a scoping/refinement stage or is it best to park it for now?

Gman0909:
@yuvipanda is there an update to this issue?

PR #4030, opened 2024-05-07
basehub: reduce hub.loadRoles and clarify the situation

consideRatio:
We had configuration under `hub.loadRoles` that didn't make a difference. With this PR I'm removing it and adding related comments about it for the future instead.

- part of work towards #4014

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -601,29 +601,6 @@ jupyterhub:
     # c.JupyterHub.load_roles without overriding a list - use it instead of the
     # passthrough config hub.config.JupyterHub.load_roles.
     #
-    # - About jupyterhub-configurator service access
-    #
-    #   The JupyterHub admin users get the scope access:services by being
-    #   admins, which includes the scope access:services!service=configurator.
-    #   This makes them not need additional scope requests to work with
-    #   jupyterhub-configurator that also require users to be admins anyhow.
-    #
-    #   ref: https://github.com/yuvipanda/jupyterhub-configurator/blob/f46fb4e81b1de74c4fcaa5a7763fb230265bab90/jupyterhub_configurator/app.py#L100-L109
-    #
-    # - About dask-gateway service access
-    #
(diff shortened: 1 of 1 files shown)


helm-charts/basehub/values.yaml
```diff
@@ -590,15 +590,37 @@ jupyterhub:
...
-    loadRoles:
-      # Should use this, not hub.config.JupyterHub.load_roles - that will
-      # override any existing load_roles set by z2jh
-      service-use:
-        name: user
-        scopes:
-          # Allow all users access to 'services', which includes dask-gateway & configurator
-          - access:services
-          - self
+    # hub.loadRoles is z2jh native config to enable configuration of
```
yuvipanda, line 593: yay that it has no effect and we can remove it.

I'd suggest we just simply fully remove it, and not actually have this large comment here. It adds additional things that must be kept up to date whenever things elsewhere change, and makes this already large file even larger. Let's just remove that?

yuvipanda (changes requested):
Can merge once comment is addressed.

helm-charts/basehub/values.yaml
```diff
@@ -590,15 +590,37 @@ jupyterhub:
...
+    #   dask-gateway the client passes the jupyterhub user's jupyterhub api
+    #   token to the dask-gateway-server, which then just verifies that the api
+    #   token is associated with an actual user using the api token itself. Due
+    #   to that, we can't limit access to dask-gateway by providing that scope
+    #   only to some users.
+    #
+    #   Considerations of updating dask-gateway to gate access via
+    #   access:services!service=dask-gateway is considered in the linked issue.
+    #
+    #   ref: https://github.com/dask/dask-gateway/issues/829
```
yuvipanda, line 618: https://github.com/2i2c-org/infrastructure/issues/4014 is the right place for this and i see you've already put it there!

consideRatio:
Thank you for reviewing @yuvipanda!

I've tested as non-admin on imagebuilding-demo as well which worked ok

PR #4039, opened 2024-05-08
GHG Center: Add profile options for workshop

batpad:
Refs https://github.com/NASA-IMPACT/veda-jupyterhub/issues/32

Adding some compute-optimized options for containers.

This:

 - Adds a `c5.4xlarge` instance type to the `eksctl` config to create a nodepool with that instance type (compute optimized)
 - Adds resource options to the profile list for the compute optimized containers

Questions:

 - Should we include the workshop in the `display_name` or so to distinguish these options?
 - @yuvipanda please let me know if the way the resources are specified seems okay - I did some rough conversions between bits and GiB and put in values that seemed appropriate
 - Also, I've set the `cpu_limit` to be what's requested, but the `cpu_guarantee` a bit lower in line with the other resource options, lemme know if that seems okay

@yuvipanda let me know if these seem okay - happy to make any changes :)

cc @sunu @slesaad

### DIFF
--- a/config/clusters/nasa-ghg/common.values.yaml
+++ b/config/clusters/nasa-ghg/common.values.yaml
@@ -113,7 +113,7 @@ basehub:
                     cpu_guarantee: 12
                     cpu_limit: 12
                     node_selector:
-                      node.kubernetes.io/instance-type: r5.4xlarge                                                     
+                      node.kubernetes.io/instance-type: r5.4xlarge
                 mem_1_9:
                   display_name: 1.9 GB RAM, upto 3.7 CPUs
                   kubespawner_override:
--- a/config/clusters/nasa-ghg/common.values.yaml
+++ b/config/clusters/nasa-ghg/common.values.yaml
@@ -96,7 +96,7 @@ basehub:
                     cpu_limit: 12
                     node_selector:
                       node.kubernetes.io/instance-type: r5.4xlarge
-                cpu_12_16:
+                cpu_8_24:
                   display_name: 8 CPU, 24GB RAM
                   kubespawner_override:
                     mem_guarantee: 22.305Gi
--- a/config/clusters/nasa-ghg/common.values.yaml
+++ b/config/clusters/nasa-ghg/common.values.yaml
@@ -105,8 +105,7 @@ basehub:
                     cpu_guarantee: 6
                     cpu_limit: 11.8
                     node_selector:
-                      node.kubernetes.io/instance-type: c5.4xlarge
-                                                      
+                      node.kubernetes.io/instance-type: c5.4xlarge                                                     
                 mem_1_9:
                   display_name: 1.9 GB RAM, upto 3.7 CPUs
                   kubespawner_override:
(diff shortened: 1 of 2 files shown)


yuvipanda:
@batpad looks like the CI found a real issue too.

batpad:
@yuvipanda updated the PR based on our conversation:

 - Remove the additional node type and just use an existing node-pool for now while Sourish is testing
 - Added a comment to explain this and listing the steps to take after Sourish decides what resource allocations are ideal, including then figuring out the ideal node type, etc.

Thanks :-)

eksctl/nasa-ghg.jsonnet
```diff
@@ -27,7 +27,7 @@ local nodeAz = "us-west-2a";
 local notebookNodes = [
     { instanceType: "r5.xlarge" },
     { instanceType: "r5.4xlarge" },
-    { instanceType: "r5.16xlarge" },
+    { instanceType: "r5.16xlarge" }
```
yuvipanda, line 30:
```suggestion
    { instanceType: "r5.16xlarge" },
```

yuvipanda: @batpad i don't have permissions to accept this change on this PR apparently, can you?

yuvipanda:
hmm, not sure why pre-commit is failing :(

batpad:
> hmm, not sure why pre-commit is failing :(

oh, lol - maybe my YAML commenting style?

batpad:
@yuvipanda fixed all the things, thanks you and `pre-commit` for the catches.

yuvipanda:
ty for the fix!

config/clusters/nasa-ghg/common.values.yaml
```diff
@@ -70,6 +70,50 @@ basehub:
...
+                # See: https://github.com/NASA-IMPACT/veda-jupyterhub/issues/32
+                # Once we settle on the ideal resource allocation for the workshop, we should:
+                #  - Create a new profile for the workshop, with just the resource option needed
+                #  - Pick a more appropriate (compute-optimized) node type and add to eksctl
+                #  - Remove these options from the global resource select options
+                #  - Remove this comment.
+                cpu_8_16:
+                  display_name: 8 CPU, 16GB RAM
+                  kubespawner_override:
+                    mem_guarantee: 14.87Gi
```
yuvipanda, line 84: JupyterHub wants these to be 'G' rather than 'Gi'. Hence the issue. Can you change mem_guarantees to G?

Additionally, I'd love for you to also look at supporting the real 'i' suffixes in https://github.com/jupyterhub/jupyterhub/blob/fa538cfc65b199dd2947529e349abb9aa31fb9a3/jupyterhub/traitlets.py#L49 if you'd like, @batpad :)

batpad: Thanks - sorry I should have double checked the units there - I made another PR to fix this: https://github.com/2i2c-org/infrastructure/pull/4048

PR #4040, opened 2024-05-08
deployer: refactor to separate out test logic

consideRatio:


deployer/utils/file_acquisition.py
```diff
@@ -241,27 +240,14 @@ def get_decrypted_files(files):
...
-    if is_test:
-        # We are running a test via pytest. We only want to focus on the cluster
-        # folders nested under the `tests/` folder.
-        return set(REPO_ROOT_PATH.glob("tests/test-clusters/**/cluster.yaml"))
-
-    # We are NOT running a test via pytest. We only care about the clusters under config/clusters
     return {
         path
-        for path in REPO_ROOT_PATH.glob("config/clusters/**/cluster.yaml")
+        for path in CONFIG_CLUSTERS_PATH.glob("**/cluster.yaml")
```
consideRatio, line 251: The key change is that letting the deployer reference a constant to what path to glob cluster.yaml files from, and then have that constant be mocked during tests.

PR #4051, opened 2024-05-09
4/6 gke clusters still on 1.27: update to 1.29

consideRatio:
- **4/6 gke clusters still on 1.27: update k8s to 1.29**
- **deployer: update daemonset_requests.yaml**
- **4/6 gke clusters still on 1.27: update core/user nodes to 1.29**

### DIFF
--- a/config/clusters/cloudbank/common.values.yaml
+++ b/config/clusters/cloudbank/common.values.yaml
@@ -21,5 +21,5 @@ jupyterhub:
       name: quay.io/2i2c/cloudbank-data8-image
       tag: d2746e55a4ee
     nodeSelector:
-      # Put everything on the most appropriate nodepool for these users
-      cloud.google.com/gke-nodepool: nb-n2-highmem-4
+      # Put everything on the most appropriate instance type for these users
+      node.kubernetes.io/instance-type: n2-highmem-4
--- a/config/clusters/leap/common.values.yaml
+++ b/config/clusters/leap/common.values.yaml
@@ -284,7 +284,8 @@ basehub:
             environment:
               NVIDIA_DRIVER_CAPABILITIES: compute,utility
             node_selector:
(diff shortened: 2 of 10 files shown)


config/clusters/cloudbank/common.values.yaml
```diff
@@ -21,5 +21,5 @@ jupyterhub:
       name: quay.io/2i2c/cloudbank-data8-image
       tag: d2746e55a4ee
     nodeSelector:
-      # Put everything on the most appropriate nodepool for these users
-      cloud.google.com/gke-nodepool: nb-n2-highmem-4
+      # Put everything on the most appropriate instance type for these users
+      node.kubernetes.io/instance-type: n2-highmem-4
```
consideRatio, line 25: @yuvipanda @GeorgianaElena @sgibson91 note that if we pin to a specific node pool name instance of an instance type, we can't do rolling updates of k8s versions by adding a new node pool and tainting nodes of the old node pool.

Due to this, we should avoid pinning to the node pool name, and instead pin to the instance type.

config/clusters/leap/common.values.yaml
```diff
@@ -284,7 +284,8 @@ basehub:
             environment:
               NVIDIA_DRIVER_CAPABILITIES: compute,utility
             node_selector:
-              cloud.google.com/gke-nodepool: nb-gpu-t4
+              node.kubernetes.io/instance-type: n1-standard-8
+              cloud.google.com/gke-accelerator: nvidia-tesla-t4
```
consideRatio, line 288: For GPU nodes, you can pin like this - but this level of pinning is only relevant if either we have a mix of GPUs or a mix of instance sizes with GPUs.

If we only have a single GPU node pool, no node selector is required at all because k8s gets that it needs to use a GPU node pool.

PR #4055, opened 2024-05-09
[Support: openscapes] Mount allusers for openscapes admins

GeorgianaElena:
For https://2i2c.freshdesk.com/a/tickets/1572


GeorgianaElena:
I think this is a straight forward enough change that we could add to the self merge list?

Issue #4052, opened 2024-05-09
monero cryptnono detector does not start on GKE 1.29


consideRatio:
Tracked in https://github.com/cryptnono/cryptnono/issues/34

yuvipanda:
With #4056, I've disabled the bpftrace based monero detector and enabled the much more comprehensive execwhacker. execwhacker also is far more effective against cryptominers than the monero detector was.

I've also verified that there are no non-running pods with `deployer config get-clusters | xargs -I{} -L1 deployer use-cluster-credentials {} 'kubectl get pod -A | grep -v Running'`

This issue can be closed now.

PR #4056, opened 2024-05-09
cryptnono: Disable monero detector, enable execwhacker

yuvipanda:
execwhacker has run enough now that I think we can turn it on everywhere. It *should* also run more stably across kernel versions than the bpftrace based monero detector, and everything the monero detector catches should also be caught by our execwhacker config. It's also run on the HHMI hub for a while without issues.

Ref https://github.com/2i2c-org/infrastructure/issues/4052


yuvipanda:
I wanna monitor this roll out, so this shouldn't be merged until monday

Issue #4020, opened 2024-05-02
[Support] Enable GPUs for AWI-CIROH


yuvipanda:
### The Freshdesk ticket link

https://2i2c.freshdesk.com/a/tickets/1540

### Ticket request type

Feature Request

### Ticket impact

🟨 Medium

### Short ticket description

AWI CIROH wants to explore exporing GPUs. I've asked them if they'd like to restrict it to a particular subset of users. If so, we can use `allowed_teams`. If not, we can just turn it on.

Runbook available for this: https://infrastructure.2i2c.org/howto/features/gpu/

jameshalgren:
We can easily set up a separate group in the GitHub organization -- would that be sufficient to partition the users?

yuvipanda:
@jameshalgren yes that would be the way to go. Can you set up the team in GitHub and let us know the name?

jameshalgren:
@yuvipanda The team is created: 
https://github.com/orgs/AlabamaWaterInstitute/teams/2i2c_portal_users_gpu/members

yuvipanda:
Thanks, @jameshalgren! This has been assigned to @AIDEA775 to do this sprint.

@AIDEA775 I opened https://github.com/2i2c-org/infrastructure/pull/4027 with some clearer documentation on how to restrict GPUs to a specific subset of users . Based on https://github.com/2i2c-org/infrastructure/issues/4020#issuecomment-2091805709, only members of `AlabamaWaterInstitute:2i2c_portal_users_gpu` should be allowed access to the GPU profile option. Let me know if there are more questions :)

AIDEA775:
Hi @jameshalgren! We added a new option in staging with GPUs. Can you test it? Which images do you need?

https://staging.ciroh.awi.2i2c.cloud/

AIDEA775:
Hi @jameshalgren! We just deployed the same options to production. If you need any other changes, open a new ticket with support.

PR #4057, opened 2024-05-10
awi-ciroh: Enable GPUs

AIDEA775:
Ref: #4020 

Only in staging for now. If it is okay, we'll move forward to production.


yuvipanda:
@AIDEA775 i tested this, it's good to go I think!

PR #4061, opened 2024-05-12
automation: fix so pressing cancel workflow can cancel not yet started jobs

consideRatio:
If you try to cancel a "deploy and test hubs" workflow with a cancel button, what can happen is that new jobs keep starting still. This is because we have `if: always()` on the jobs as we want to allow them to run even if we have had a failure before - and this makes sense.

We can however accomplish this by using `!cancelled()` as well - which then would respect pressing cancel on the workflow to actually cancel it. Reading at [the docs for `always()`](https://docs.github.com/en/actions/learn-github-actions/expressions#always):

image

### DIFF
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -168,7 +168,6 @@ jobs:
       # Upload the pr-number.txt and comment-body.txt files as artifacts for the
       # comment-deployment-plan-pr.yaml workflow to access
       - name: Upload artifacts
-        # Only run this steps in PRs when the matrices are not empty
         if: >
           github.event_name == 'pull_request' &&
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-hubs.yaml
```diff
@@ -214,10 +213,9 @@ jobs:
       failure_utoronto: "${{ env.failure_utoronto }}"
       failure_victor: "${{ env.failure_victor }}"
 
-    # Only run this job on pushes to the default branch and when the job output is not
-    # an empty list
```
consideRatio, line 218: I removed some comments that were hard to maintain as they were very detailed, but the code described is quite clear by itself so i think it didnt hurt readability

Issue #3214, opened 2023-10-02
GitHub automation - workflows cancelled because other higher prio workloads


consideRatio:
We can't cancel the entire deployment workflow, as each workflow may deploy something different. For example, a support chart deployment workflow run could get cancelled because a hub deployment pr is merged - this logic can't be aware about this.

![image](https://github.com/2i2c-org/infrastructure/assets/3837114/1f3502b4-0d74-4731-b826-eaf73b22d6de)

yuvipanda:
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue maybe a solution

consideRatio:
I looked into the merge queue thing.

It seems to handle a situation when you want to merge for example 10 PRs, and want tests to rerun tests just before merge so they pass against an updated branch. I think however it won't help us with this issue =/

consideRatio:
Closing #2652 as a duplicate of this.

PR #4065, opened 2024-05-13
automation/deployer: respect PR label `deployer:skip-deploy`

### LINKED ISSUE #2009: A way to merge and not trigger CD pipeline

consideRatio:
- Fixes #2009

I wanted this a lot in order to:
1. Not need to press merge and then quickly manually cancel the triggered workflow when deciding to not deploy/test
2. Not create noise in `#gha-failures` for manually cancelled workloads, and to not need to comment that it was just a intentionally cancelled workload
3. Be able to visibilize a deploy skip is planned
4. Not incurr delays by queueing many deploys when only a one or few may be relevant

I also wanted this for some additional reasons, like reducing cloud costs for communities and reducing risk of #3214 further, but the main reasons motivating me now to implement this feature was those above.

## Review notes

- **About testing of changes**
  - Acquisition of PR labels in PR and [on merge has been tested in another repo](https://github.com/consideRatio/testing-gh-workflows/actions/runs/9070291918/job/24921758503)
  - `deployer` code changes is partially tested by added tests, where [this PR triggered a run that tested the remaining parts](https://github.com/2i2c-org/infrastructure/actions/runs/9069888846/job/24920451820?pr=4065)
- **About outdated deploy PR comments**
  The PR comment made listing what gets deployed won't get updated once the `deployer:skip-deploy` label gets added. This is also the case if you first push a commit that leads to a deploy, and then reverts the commit. There is no logic to delete an existing comment or update it to declare nothing will get deployed in those situations.

### DIFF
--- a/deployer/commands/generate/helm_upgrade/jobs.py
+++ b/deployer/commands/generate/helm_upgrade/jobs.py
@@ -26,13 +26,19 @@ yaml = YAML(typ="safe", pure=True)
 def helm_upgrade_jobs(
     changed_filepaths: str = typer.Argument(
         ..., help="Comma delimited list of files that have changed"
-    )
+    ),
+    pr_labels: str = typer.Argument(
+        "[]",
+        help="JSON formatted list of PR labels, where 'deployer:skip-deploy' is respected.",
+    ),
 ):
     """
-    Analyze added or modified files from a GitHub Pull Request and decide which
-    clusters and/or hubs require helm upgrades to be performed for their *hub helm
-    charts or the support helm chart.
+    Analyze added or modified files and labels from a GitHub Pull Request and
+    decide which clusters and/or hubs require helm upgrades to be performed for
+    their *hub helm charts or the support helm chart.
     """
+    pr_labels = json.loads(pr_labels)
+
     changed_filepaths = changed_filepaths.split(",")
     (
         upgrade_support_on_all_clusters,
@@ -87,6 +93,7 @@ def helm_upgrade_jobs(
                 cluster_config,
                 cluster_info,
                 set(changed_filepaths),
+                pr_labels,
                 upgrade_all_hubs_on_this_cluster=upgrade_all_hubs_on_this_cluster,
                 upgrade_all_hubs_on_all_clusters=upgrade_all_hubs_on_all_clusters,
             )
@@ -99,6 +106,7 @@ def helm_upgrade_jobs(
                 cluster_config,
                 cluster_info,
                 set(changed_filepaths),
+                pr_labels,
                 upgrade_support_on_this_cluster=upgrade_support_on_this_cluster,
                 upgrade_support_on_all_clusters=upgrade_support_on_all_clusters,
             )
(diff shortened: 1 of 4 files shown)


deployer/commands/generate/helm_upgrade/jobs.py
```diff
@@ -135,10 +143,9 @@ def helm_upgrade_jobs(
     if ci_env:
         # Add these matrix jobs as environment variables for use in another job
         with open(env_file, "a") as f:
-            f.write(f"prod-hub-matrix-jobs={json.dumps(prod_hub_matrix_jobs)}")
-            f.write("\n")
+            f.write(f"prod-hub-matrix-jobs={json.dumps(prod_hub_matrix_jobs)}\n")
             f.write(
-                f"support-and-staging-matrix-jobs={json.dumps(support_and_staging_matrix_jobs)}"
+                f"support-and-staging-matrix-jobs={json.dumps(support_and_staging_matrix_jobs)}\n"
```
consideRatio, line 148: This is an PR unrelated fix - when we write to the file path `$GITHUB_ENV`, it should always get full lines written to it. Otherwise we could get lines like `MY_ENV=testSOMEOTHER_ENV=value`

consideRatio:
@yuvipanda I requested your review since you were involved in discussion about this in #2009

GeorgianaElena (commented):
@consideRatio, sorry if it's a silly question, I didn't spent much time reasoning about it, but why have the label special cased in the deployer and not inside the workflow itself, maybe similar to https://github.com/2i2c-org/infrastructure/blob/296d8559c0a770b7a83c7ea669a8fb46ac4f94fc/.github/workflows/project-board-update.yml#L7-L11 ?

consideRatio:
> @consideRatio, sorry if it's a silly question, I didn't spent much time reasoning about it, but why have the label special cased in the deployer and not inside the workflow itself, maybe similar to
> 
> https://github.com/2i2c-org/infrastructure/blob/296d8559c0a770b7a83c7ea669a8fb46ac4f94fc/.github/workflows/project-board-update.yml#L7-L11
> 
> ?

The crux is that the information that is available in context under `github...` differ depending on how the workflow was triggered. The linked example above had a trigger of `on.issues.types[labeled]`, and that has context of a single label for example -- the one label being added.

For `pull_request` triggered workflows, there are information about the pull request, including its labels, but for `push` triggered workflows, github can't even know for sure the push stemmed from a merged pull request, so it doesn't provide a context describing information about a pull request.

Due to that, I ended up adding a job step to get the associated PRs labels adjusting for the context on how to acquire it - I'll add a comment clarifying this next to the step.

GeorgianaElena:
@consideRatio, would you be willing to document this label and when to use it (maybe in our SRE section)?

I'm not suggesting to think up-front about all possible scenarios, but while you notice yourself using it, I think it would be very useful for others on the team to have it documented and to understand when it it's beneficial to skip the deploy.

PR #4075, opened 2024-05-14
[openscapes-workshop] Use a branch that allows homepage customizations

GeorgianaElena:
This picks up the changes in https://github.com/2i2c-org/default-hub-homepage/pull/31

 https://workshop.openscapes.2i2c.cloud


yuvipanda:
I think this still counts as being within the 'new hub' period, so you can self merge this too.

PR #4076, opened 2024-05-14
leap: Add new devs only profile

jbusecke:
Adding a new profile just for devs to test out e.g. bigger instance types.

I basically just multiplied all the options of the 16 core choice below by 4 to adjust to the 64 core instance.

### DIFF
--- a/config/clusters/leap/common.values.yaml
+++ b/config/clusters/leap/common.values.yaml
@@ -183,7 +183,7 @@ basehub:
             mem_limit: null
             node_selector:
               node.kubernetes.io/instance-type: n2-highmem-64
-            
+
         - display_name: "CPU only"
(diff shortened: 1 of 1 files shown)


config/clusters/leap/common.values.yaml
```diff
@@ -128,6 +128,62 @@ basehub:
         #          causes more frequent need to wait on startups, increased cost,
         #          wasted energy.
         #
+        - display_name: "Devs Only"
+          description: &profile_list_description "Start a container limited to a chosen share of capacity on a node of this type"
+          slug: medium-full
```
consideRatio, line 133: This must be unique - it can also be omitted

```suggestion
```

yuvipanda:
Tested and it works

PR #4091, opened 2024-05-17
jupyter-health: remove profile list

minrk:
for smoother launch, since we only have one image

promotes default profile to sole singleuser configuration, unmodified

### DIFF
--- a/config/clusters/jupyter-health/common.values.yaml
+++ b/config/clusters/jupyter-health/common.values.yaml
@@ -60,7 +60,7 @@ jupyterhub:
     cpu:
       limit: 3.725
       guarantee: 0.2328125
-    node_selector:
+    nodeSelector:
       node.kubernetes.io/instance-type: r5.xlarge
 
   scheduling:
--- a/config/clusters/jupyter-health/common.values.yaml
+++ b/config/clusters/jupyter-health/common.values.yaml
@@ -53,69 +53,15 @@ jupyterhub:
           - ryanlovett
   singleuser:
     defaultUrl: /lab
-    profileList:
-      - display_name: "Default image"
-        slug: pangeo
-        description: Jupyter Health default image
-        default: true
-        kubespawner_override:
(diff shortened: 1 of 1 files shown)


config/clusters/jupyter-health/common.values.yaml
```diff
@@ -53,69 +53,15 @@ jupyterhub:
...
-            choices: {}
-          resource_allocation: *profile_options_resource_allocation
+    image: quay.io/jupyterhealth/singleuser-premvp:latest
+    memory:
+      limit: 1991244775
+      guarantee: 1991244775
+    cpu:
+      limit: 3.725
+      guarantee: 0.2328125
+    node_selector:
```
consideRatio, line 63:
```suggestion
    nodeSelector:
```

minrk:
Thanks!

minrk:
Launches are failing now, presumably a typo somewhere here?

minrk:
spawn is failing with: `No such profile: pangeo`

consideRatio:
@minrk I think what happens is that we have `user_options` state saved in the jupyterhub database about a profile slug, and when kubespawner receives it and there isn't a profile - it errors. I believe a fix could be to reset the jupyterhub DB for example. Do you want me to apply that fix short term?

- should jupyterhub retain and sending old user_options?
- should kubespawner handle this in a more fault tolerant way?

I think this is an upstream issue overall, but I'm not sure where.

consideRatio:
You get the same error if you use an incognito window right - because there isn't any user_option state in the frontend etc?

consideRatio:
@minrk I went ahead with what I think is the fix - hub-db-dir cleared and re-deployed

minrk:
Thanks! Works now. Something to think about in either kubespawner or JupyterHub: what happens when profilelist config is removed after being in use. Simplest version in kubespawner would be to ignore profile `if not profile_list`. Could be with a warning.

Issue #4085, opened 2024-05-15
Remove the `binderhub` chart from our infrastructure


yuvipanda:
We have decided to use https://github.com/2i2c-org/binderhub-service/ to deploy binderhub, rather than the existing upstream binderhub chart. This is primarily due to the fact that the upstream chart inherits its own copy of jupyterhub, which means we can not compose it easily with the other parts of what we offer in `basehub`. We had tried to deploy the `binderhub` chart in this repo as well, but that should be all ripped out now.

# Definition of done

This task is complete when there are no remnants of the `binderhub` chart in our code, documentation, testing infrastructure and CI.

PR #4096, opened 2024-05-17
Remove the binderhub chart and associated info

GeorgianaElena:
Closes https://github.com/2i2c-org/infrastructure/issues/4085


deployer/commands/generate/resource_allocation/daemonset_requests.yaml
```diff
@@ -40,7 +40,7 @@ gke:
   #
   2i2c:
     requesting_daemon_sets: calico-node,fluentbit-gke,gke-metadata-server,gke-metrics-agent,ip-masq-agent,netd,pdcsi-node,support-cryptnono,support-prometheus-node-exporter
-    other_daemon_sets: binder-staging-dind,binder-staging-image-cleaner,imagebuilding-demo-binderhub-service-docker-api
```
yuvipanda, line 43: I think the last two may still be running here?

yuvipanda (approved):
One comment, otherwise looks good to me!

deployer/commands/generate/resource_allocation/daemonset_requests.yaml
GeorgianaElena: Ah, good catch. The file was actually updated automatically by the deployer command: `deployer generate resource-allocation daemonset-requests 2i2c` and I think it removed the `imagebuilding-demo-binderhub-service-docker-api` because it was not running.

I fired an image build on the imagebuilding-demo hub and ran the command again and it picked it up. I don't think this command is documented though so not sure where to capture this shortcoming. (cc @consideRatio as tbe

consideRatio: For things running on the user nodes only, when there isn't a user node, this could happen.

Currently this is only used to provide information that is manually considered. I suggest doing nothing for now.

Issue #2766, opened 2023-07-06
paths-filter action detecting more changes than necessary when PR branch is out of date with default branch


sgibson91:
### Context

While working on https://github.com/2i2c-org/infrastructure/pull/2762, I noticed that at a stage the deployment plan comment was reporting that all the hubs would be redeployed even though my PR was not making any changes that should trigger a redeploy of everything.

I noticed two things:

1. In the action logs of the paths-filter step, there is the message: "Github token is not available - changes will be detected from PRs merge commit"
2. My branch was behind the default branch (though not in a way that prevented merge)

Once I updated my branch to be up-to-date with the default, the deployment plan refreshed to only include the one hub that I was adding, which is the expected behaviour. (Note that the message from (1) was still present.)

My intuition is thinking that, if the PR branch is behind the default branch and the paths-filter action is using the merge commit, then we will see deployment plans that do not make sense for our PRs because we have pulled in other changes into the analysis from the discrepancy. And these affected deployment plans *might* causes unnecessary redeploys (I'm not sure what would happen on merge).

### Proposal

I see two possible things to try here (let me know of any options I missed):

1. We could have a branch protection rule that requires branches be up-to-date before merge
   - **Pros:** I imagine this is a common practice in parts of industry and ensures changes are always tested on the most recent version of code in the default branch
   - **Cons:** I've never worked in a team that used this practice so I don't have a sense for how 'annoying' or 'disruptive' it'll be (though I don't think that much, especially as GitHub provides an "Update branch" button at the bottom of PRs). This also feels like a hack in that it doesn't address the problem which is unexpected behaviour from the paths-filter action.
2. We could try giving a token to the paths-filter action. We currently don't.
   - **Pros:** This solution actually targets the source of uncertainty
   - **Cons:** We don't (yet) know what paths-filter will use if not the merge commit. We may run into a resurgence of https://github.com/2i2c-org/infrastructure/issues/1686 (but there is a way to track that now)

sgibson91:
Needs investigation to determine if this is still the case or if it has resolved itself

PR #4097, opened 2024-05-17
Temporarily decom MIT Brain hubs (bican, dandi, linc)

yuvipanda:
Partnerships is working on reconfiguring this relationship, but it isn't going to look like 3 separate hubs on 3 separate clusters. Decom this to save our cloud costs, as nobody is currently using these.

Based on a meeting I just had with them (along with @colliand). Will have an associated issue soon.

Ref https://github.com/2i2c-org/leads/issues/344


yuvipanda:
I've terraform destroyed everything for now.

GeorgianaElena (approved):

consideRatio:
These clusters still incurring cloud costs from k8s clusters (72 USD / month) a core node (~180 USD / month), as this was controlled via eksctl rather than terraform.

@yuvipanda if you think its OK to proceed i'll just delete the eksctl resources as well.

yuvipanda:
@consideRatio ah, I missed that. yes, please do when you get a minute.

consideRatio:
```
jsonnet $CLUSTER_NAME.jsonnet > $CLUSTER_NAME.eksctl.yaml
eksctl delete cluster --config-file $CLUSTER_NAME.eksctl.yaml
```

```bash
# to help drain operation from another terminal without availability of deployer creds
eksctl utils write-kubeconfig --config-file=$CLUSTER_NAME.eksctl.yaml --auto-kubeconfig
export KUBECONFIG=/home/erik/.kube/eksctl/clusters/$CLUSTER_NAME

# eksctl reports: 6 pods are unevictable from node ip-192-168-8-78.us-east-2.compute.internal
# I think four of these referred to two coredns pods and two ebs-csi-controller
# pods with protection from PDB resources
# I think another pod was from the cryptnono daemonset, which as a daemonset
# shouldn't cause issues, but maybe it had a emptyDir volume or similar that maybe requires
# a flag that it was ok to delete also such pods?
```

consideRatio:
- we'll now start seeing pagerduty warnings I think
- we'll also until this is merged risk causing additional maintenance work and confusion about bican/dandi/linc during maintenance

Due to this, I'll go for a merge - then the pagerduty healthchecks are deleted I think via automation scanning our cluster config files.

consideRatio:
I applied the `deployer:skip-deploy` label as [166e53f](https://github.com/2i2c-org/infrastructure/pull/4097/commits/166e53f58acb52ea95ae0f0185c7f3a59c07219c) incorrectly seemed to have caused the wrong conclusion.

sgibson91:
> incorrectly seemed to have caused the wrong conclusion that all hubs need to re-deploy and test.

This can happen if the branch is not up-to-date with main https://github.com/2i2c-org/infrastructure/issues/2766

Issue #4009, opened 2024-04-30
k8s maintenance: k8s node pool upgrades on EKS clusters


consideRatio:
#4007 upgraded all control planes to 1.29, but we need to bring the node pools to 1.29 as well. Documentation on how to do this was updated in #4099 and is made available at https://infrastructure.2i2c.org/howto/upgrade-cluster/aws/ - where step 4 about upgrading the control plane is already done.

### Clusters with node groups to upgrade

- [x] 2i2c-aws-us
- [x] catalystproject-africa
- [x] earthscope
- [x] gridsst
- [x] jupyter-meets-the-earth
- [ ] nasa-cryo
   - Waiting for old user server node group to drain so it can then be deleted
- [x] nasa-esdis
- [ ] nasa-ghg
  - ~~Can't be upgraded due to missing public ssh key, tracked in #4122~~
  - Waiting for old user server node group to drain so it can then be deleted
- [x] nasa-veda
- [x] openscapes
- [x] opensci
- [x] smithsonian
- [x] ubc-eoas
- [x] victor

yuvipanda:
Should be paired with someone else (TBD)

consideRatio:
@sgibson91 agreed to do this work this during this sprint. I'm around to discuss anything if requested.

sgibson91:
Most clusters are done, there are only the remaining clusters which have some user servers still up

| Cluster | Number of running singleuser servers |
| :--- | :--- |
| nasa-cryo | 2 |
| nasa-ghg | 3 |

sgibson91:
2 of the nasa-cryo user servers have been up for >2days and 1 of the nasa-ghg user serves has been up for 36 hours. I suspect these might be abandoned so maybe they're fine to kick off?

consideRatio:
> 2 of the nasa-cryo user servers have been up for >2days and 1 of the nasa-ghg user serves has been up for 36 hours. I suspect these might be abandoned so maybe they're fine to kick off?

You could do a [rolling upgrade without using drain](https://infrastructure.2i2c.org/howto/upgrade-cluster/aws/#performing-rolling-upgrades-using-drain-or-not) on the node pool they are using, and in the "taint and wait step" you push the changes up to that point and get a PR merged - leaving only a comment saying we need to also delete a node pool, something that hopefully can be done next week at least.

sgibson91:
[...]

PR #4099, opened 2024-05-19
docs: iteration on aws k8s upgrade docs

consideRatio:
- **No longer assumes empty user nodes**
  Our docs previously assumed there were no users on the user nodes, assuming a re-creation upgrade strategy could be used. This assumption is removed in this iteration.
- **Removed out of scope section**
  I removed the pre-requiesite section "Consider changes to `template.jsonnet`" that I now consider too out of scope to be suggested in docs to be done during a k8s upgrade.
- **Refactoring to centralize misc content**
  Node upgrade strategies, AWS auth, notes on version skew, and maybe something more.
- **Check cluster status and activity**
  Added step to overview what goes on in the cluster before upgrading. This can for example be used to rule out that something broke because of the upgrade (because it was already broken).
- **Facilitate upgrading multiple clusters in parallell**
  Upgrading multiple clusters in parallell is reasonable, and I've now made it so that the guide is easier to scale to run in parallell.

## Related

This was worked in preparation for #4009. If it wasn't done now, it would be harder to handle #4009 even though it wasn't part of scope of #4009 to get this done.

## Review

I think the time efficient approach is to let this be practically reviewed by merging it and then using it - iterating on it further if needed to fix issues with it.

### DIFF
--- a/docs/howto/upgrade-cluster/aws.md
+++ b/docs/howto/upgrade-cluster/aws.md
@@ -2,74 +2,17 @@
 
 # Upgrade Kubernetes cluster on AWS
 
-```{warning}
-This upgrade will cause disruptions for users and trigger alerts for
-[](uptime-checks). To help other engineers, communicate that your are starting a
-cluster upgrade in the `#maintenance-notices` Slack channel and setup a [snooze](uptime-checks:snoozes)
-```
-
-```{warning}
-We haven't yet established a policy for planning and communicating maintenance
-procedures to users. So preliminary, only make a k8s cluster upgrade while the
-cluster is unused or that the maintenance is communicated ahead of time.
-```
-
 ## Pre-requisites
 
-1. *Install or upgrade CLI tools*
+1. *Install/upgrade CLI tools*
 
-   Install required tools as documented in [](new-cluster:prerequisites),
-   and ensure you have a recent version of eksctl.
+   Refer to [](new-cluster:prerequisites) on how to do this.
 
-   ```{warning}
-   Using a modern version of `eksctl` has been found important historically, make
-   sure to use the latest version to avoid debugging an already fixed bug!
+   ```{important}
+   Ensure you have the latest version of `eksctl`, it matters.
    ```
 
-2. *Consider changes to `template.jsonnet`*
-
-   The eksctl config jinja2 template `eksctl/template.jsonnet` was once used to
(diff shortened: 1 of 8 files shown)


docs/howto/upgrade-cluster/aws.md
```diff
@@ -2,74 +2,17 @@
...
-   git status
+   Refer to [](new-cluster:prerequisites) on how to do this.
 
-   # restore changes to all files but the .jsonnet files
-   git add *.jsonnet
-   git checkout ..  # .. should be the git repo's root
-   git reset
-
-   # inspect changes
-   git diff
```
consideRatio, line 66: This is out of scope for k8s maintenance, and more in scope for whenever something changes in the template.

docs/howto/upgrade-cluster/aws.md
```diff
@@ -2,74 +2,17 @@
 
 # Upgrade Kubernetes cluster on AWS
 
-```{warning}
-This upgrade will cause disruptions for users and trigger alerts for
-[](uptime-checks). To help other engineers, communicate that your are starting a
-cluster upgrade in the `#maintenance-notices` Slack channel and setup a [snooze](uptime-checks:snoozes)
```
consideRatio, line 8: We shouldn't need a snooze, so that was removed. I relocated the note on communicating in slack to a dedicated step.

docs/howto/upgrade-cluster/aws.md
```diff
@@ -2,74 +2,17 @@
...
-```
-
-```{warning}
-We haven't yet established a policy for planning and communicating maintenance
-procedures to users. So preliminary, only make a k8s cluster upgrade while the
-cluster is unused or that the maintenance is communicated ahead of time.
-```
```
consideRatio, line 15: We still haven't, but as long as we are only very little disruptive (brief networking etc) we can probably make upgrades while clusters are active anyhow.

I opted to remove this rather than refine this. I acknowledge we should have a policy, but I'd like us to collectively iterate and find agreement after collective experience rather than me declaring and motivating one in a PR about how to make a k8s upgrade technically.

GeorgianaElena (approved):

consideRatio:
Thanks for reviewing @GeorgianaElena!! I rebased and added a commit with some adjustments after trying them out myself.

PR #4103, opened 2024-05-21
Document the Dummy Authenticator

### LINKED ISSUE #4081: [docs] Document how to setup a hub with dummy authenticator
https://github.com/2i2c-org/infrastructure/issues/3984 required a new type of authentication for a 2i2c hub, using the dummy authenticator with a global password. We should make sure such an authentication is documented so we can replicate this hub pain-free when it's requested again.

Relevant info to be incorporated in the docs
- possible place for docs: https://github.com/2i2c-org/infrastructure/tree/main/docs/hub-deployment-guide/configure-auth
- the hub uses a specific branch of the homepage repository [https://github.com/2i2c-org/default-hub-homepage/tree/no-homepage-customizations that](https://github.com/2i2c-org/default-hub-homepage/tree/username-and-password-homepage) This allows us to have a username and password prompt on the homepage while also providing specific info about and for each community.
- the dummy authenticator can be used with or without a global password (we should recommend using it with)
- the password needs to be stored encrypted with sops 
- there can be no admin users because every user can use any username to login including the one of the presumed admins 
- this also means we need to disable the configurator

Things related more to processes:
- how the community should request changing the global password (https://infrastructure.2i2c.org/sre-guide/support/decrypt-age/)

sgibson91:
- fixes #4081


docs/hub-deployment-guide/configure-auth/dummy-auth.md
```diff
@@ -0,0 +1,110 @@
...
+    domain: ...
+    helm_chart: ...
+    helm_chart_values_files:
+      - ${HUB_NAME}.values.yaml
+      - enc-${HUB_NAME}.secret.values.yaml
+```
+
+```{note}
+To use the Dummy Authenticator without a password, the steps on including the password config and encrypting it can be skipped.
+```
```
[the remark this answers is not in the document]
consideRatio: I think so too

docs/hub-deployment-guide/configure-auth/dummy-auth.md
```diff
@@ -0,0 +1,110 @@
+# Dummy Authentication
```
yuvipanda, line 1: Let's call this 'shared password authentication', and keep 'Dummy authenticator' as an implementation detail. We should not set this up without a shared password.

docs/hub-deployment-guide/configure-auth/dummy-auth.md
```diff
@@ -0,0 +1,110 @@
...
+jupyterhub:
+  custom:
+    homepage:
+      gitRepoBranch: "username-and-password-homepage"
+      templateVars: [...]  # These values are as normal
+```
+
+## (Recommended) Using Dummy Authenticator with a global password
+
+We strongly recommend using a global password with the authenticator.
```
yuvipanda, line 31: In our infrastructure, we should *only* support this with a global password. For fully password-less access, there is documentation under 'ephemeral hub' that uses tmpauthenticator i think.

docs/hub-deployment-guide/configure-auth/dummy-auth.md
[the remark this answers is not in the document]
yuvipanda: Let's remove this, to not support this option.

docs/hub-deployment-guide/configure-auth/dummy-auth.md
```diff
@@ -0,0 +1,110 @@
...
+      - enc-${HUB_NAME}.secret.values.yaml
+```
+
+```{note}
+To use the Dummy Authenticator without a password, the steps on including the password config and encrypting it can be skipped.
+```
+
+### How the community should request changing the password
+
+If the community wishes to change the global password, they should do so by [submitting a ticket the support team](https://docs.2i2c.org/support/) and using `age` to encrypt the new password.
```
yuvipanda, line 80: For this specific case, I think it's ok to not ask them to encrypt the password with `age`.

yuvipanda (changes requested):
Thanks @sgibson91! I've left some comments here.

sgibson91:
Thank @yuvipanda - feedback addressed in the two most recent commits

yuvipanda (approved):

Issue #3957, opened 2024-04-17
[EVENT] [May 29 - June 1] AWI-CIROH 


yuvipanda:
### The link towards the Freshdesk ticket this event was reported

https://2i2c.freshdesk.com/a/tickets/1466

### The GitHub handle or name of the community representative

Arpita Patel

### The date when the event will start

May 29 2024, 9AM Central Time (https://arewemeetingyet.com/Center/2024-05-29/09:00)

### The date when the event will end

June 1 2024, 5PM Central Time (https://arewemeetingyet.com/Center/2024-06-01/17:00)

### What hours of the day will participants be active? (e.g., 5am - 5pm US/Pacific)

9AM - 5PM

### Are we three weeks before the start date of the event?

- [X] Yes
- [ ] No

### Number of attendees

TBD

### Make sure to add the event into the calendar

- [ ] Done

### Does the hub already exist?

- [X] Yes
- [ ] No

### The URL of the hub that will be used for the event

https://ciroh.awi.2i2c.cloud/

### Will this hub be decommissioned after the event is over?

- [ ] Yes
- [X] No

### Was all the info filled in above?

- [ ] Yes
- [X] No

### Quotas from the cloud provider are high-enough to handle expected usage?

- [ ] Yes
- [ ] No

consideRatio:
> Approximately 10 workshops (50 users per workshop) are set to utilize 2i2c, opting for either small or medium instances.

> The majority of workshops plan to use either Small instances (with 5GB RAM and 2 CPUs) or Medium instances (with 11GB RAM and 4 CPUs) for their sessions.

---

This implies up to 500 users, so this requires some planning and verification that its reasonable. I'd say they should use for example n2-highmem-16 machines.

Looking at their current setup:

```yaml
        - display_name: "Small"
          description: 5GB RAM, 2 CPUs
          default: true
          profile_options: &profile_options
            image: # ...
          kubespawner_override:
            mem_limit: 7G
            mem_guarantee: 5G
            cpu_limit: 2
            cpu_guarantee: 0.938
            node_selector:
              node.kubernetes.io/instance-type: n2-highmem-16
        - display_name: Medium
          description: 11GB RAM, 4 CPUs
          profile_options: *profile_options
          kubespawner_override:
            mem_limit: 15G
            mem_guarantee: 11G
            cpu_limit: 4
            cpu_guarantee: 1.875
            node_selector:
              node.kubernetes.io/instance-type: n2-highmem-16
```

[...]

PR #4106, opened 2024-05-21
AWI-CIROH: Add 5 profile entries for DevCon

benlee0423:
Ref https://github.com/2i2c-org/infrastructure/issues/3957

### DIFF
--- a/config/clusters/awi-ciroh/prod.values.yaml
+++ b/config/clusters/awi-ciroh/prod.values.yaml
@@ -49,6 +49,37 @@ basehub:
                   slug: "new"
                   kubespawner_override:
                     image: "quay.io/2i2c/awi-ciroh-image:79bb68961852"
+                devcon1:
+                  display_name: CNN for downscaling climate forcing variables
+                  default: true
+                  slug: "devcon1"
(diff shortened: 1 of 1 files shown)


config/clusters/awi-ciroh/prod.values.yaml
```diff
@@ -49,6 +49,36 @@ basehub:
                   slug: "new"
                   kubespawner_override:
                     image: "quay.io/2i2c/awi-ciroh-image:79bb68961852"
+                devcon1:
+                  display_name: CNN for downscaling climate forcing variables
+                  default: true
+                  slug: "devcon1"
+                  kubespawner_override:
+                    image: "quay.io/benlee7411/devcon24:swolvin"
```
yuvipanda, line 57:
```suggestion
                    image: "quay.io/benlee7411/devcon24:swolvin"
                    image_pull_policy: "Always"
```

I think you are planning on simply updating these tags as needed, is that right? If so, let's add this `image_pull_policy` to all of these as well, so that image tag caching is disabled.

benlee0423: Added

PR #4108, opened 2024-05-21
AWI-CIROH: remove extra unneeded config parameters in profile

benlee0423:
Follow-up to https://github.com/2i2c-org/infrastructure/pull/4106

Ref https://github.com/2i2c-org/infrastructure/issues/3957

### DIFF
--- a/config/clusters/awi-ciroh/prod.values.yaml
+++ b/config/clusters/awi-ciroh/prod.values.yaml
@@ -50,37 +50,27 @@ basehub:
                   kubespawner_override:
                     image: "quay.io/2i2c/awi-ciroh-image:79bb68961852"
                 devcon1:
-                  display_name: CNN for downscaling climate forcing variables
-                  default: true
-                  slug: "devcon1"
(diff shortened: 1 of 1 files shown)


config/clusters/awi-ciroh/prod.values.yaml
```diff
@@ -50,37 +50,27 @@ basehub:
                   kubespawner_override:
                     image: "quay.io/2i2c/awi-ciroh-image:79bb68961852"
                 devcon1:
-                  display_name: CNN for downscaling climate forcing variables
-                  default: true
-                  slug: "devcon1"
+                  display_name: DevCon24 workshop : CNN for downscaling climate forcing variables
```
yuvipanda, line 53: You can use a `description` field to display extra information here as well.

yuvipanda: CI is also failing because `:` is not allowed here in the content of the string. Either move this to the description, or put `"` around the whole string

benlee0423: Fixed

Issue #4086, opened 2024-05-15
[Spike] Determine what needs to happen for `binderhub-service` to expose the binderhub UI


yuvipanda:
Currently, the binderhub-service chart primarily exposes the binderhub API. Given that we're not going to be using the upstream binderhub chart (https://github.com/2i2c-org/infrastructure/issues/4085), we should explore what is needed for `binderhub-service` to expose the binderhub UI.

This should be done keeping in mind that we don't want to tie assumptions about how jupyterhub is deployed into this binderhub. The outcome should be primarily *documentation*, as well as *general extensibility improvements* to binderhub-service.

## Tasks
- [x] Schedule - Figure out who needs to be apart of this and the duration for the spike
- [x] Do the spike
- [x] Report the outcome

---

### 1. Actual spike
@consideRatio, @yuvipanda and @GeorgianaElena met for the spike and after the meeting the following steps were considered that need to happen for a binderhub ui style hub with binderhub-service (this is summarized from https://hackmd.io/G06D9YfOSsSf_mt13EeZoA?both):

- [ ] the jupyterhub is dedicated to be ephemeral (no persistent home dirs).
- [ ] binderhub server software needs to be aware about:
  - [ ] domain names
      - needs both local and public in order to redirect the user from web frontend
  - [ ] have an jupyterhub api token
      - autogenerated and exposed in a Secret that we can mount assuming its in the same namespace
      - Z2JH should define a service for binderhub
      - this needs additional work in binderhub-service
- [ ] [BinderSpawnerMixin](https://github.com/jupyterhub/binderhub/blob/c6c5dc8fe73f81ca538c47b420b33f317c3aa8ae/binderhub/binderspawner_mixin.py#L24) needs to be in the hub image

### 2. After-spike conclusion
Have a binderhub architecture walk-though for @GeorgianaElena and other interested 2i2c team members to spread the knowledge
- [ ] https://github.com/2i2c-org/infrastructure/issues/4113

### 3. After-spike work session

@consideRatio and @GeorgianaElena met after the spike for a working session. 
@GeorgianaElena's rough plan before the working session was to follow the instructions @yuvipanda wrote for installing and connecting binderhub-service https://binderhub-service.readthedocs.io/en/latest/tutorials/install.html  to a hub and from there, go with the plan @yuvipanda and Erik put toghether during the spike (referenced at point 1).

But:
- Georgiana's local install of minikube was erroring out and needed fidelling
[...]

PR #4119, opened 2024-05-22
[New Hub prototype] Binderhub UI demo

GeorgianaElena:
Related to https://github.com/2i2c-org/infrastructure/issues/4086 and https://github.com/2i2c-org/infrastructure/issues/4133

What happens in this PR:

- [x] disable hub's storage via `jupyterhub.storage.type = none`
- [x] swap the spawner with `BinderSpawnerMixin` and `enable_auth` on it
- [x] deploy binderhub as a jupyterhub externally managed service
- [x] enable the binderhub-service ingress and configure it so that we get a public domain for binderhub to run
   -  this means we ca stop specifying a service. This is because services.url  is meant to be added to the proxy at /services/:name and we don't need it because we intend to use the service's public url set by the ingress
- [ ] todo continue the list


consideRatio:
We were thinking about...

```yaml
        - name: JUPYTERHUB_SERVICE_NAME
          value: binder
        - name: JUPYTERHUB_API_URL
          value: {{ (print (.Values.config.BinderHub.hub_url_local | default .Values.config.BinderHub.hub_url | trimSuffix "/") "/hub/api/") }}
        - name: JUPYTERHUB_BASE_URL
          value: {{ .Values.jupyterhub.hub.baseUrl | quote }}
        - name: JUPYTERHUB_CLIENT_ID
          value: {{ .Values.jupyterhub.hub.services.binder.oauth_client_id | quote }}
        - name: JUPYTERHUB_OAUTH_CALLBACK_URL
          value: {{ .Values.jupyterhub.hub.services.binder.oauth_redirect_uri | quote }}
```

config/clusters/2i2c/binderhub-ui-demo.values.yaml
```diff
@@ -0,0 +1,124 @@
...
+  extraEnv:
+    - name: JUPYTERHUB_API_TOKEN
+      valueFrom:
+        secretKeyRef:
+          name: '{{ include "jupyterhub.hub.fullname" . }}'
+          key: hub.services.binder.apiToken
+    - name: JUPYTERHUB_CLIENT_ID
+      value: "service-binder"
+    - name: JUPYTERHUB_OAUTH_CALLBACK_URL
+      value: "/services/binder/oauth_callback"
```
consideRatio, line 110: I suspect this need to be the binders public domain, because otherwise we interact with the binder webserver both using the binder public domain and via /services/binder path on the hub domain.

GeorgianaElena:
There's almost a functional binderhub running at https://binderhub-ui-demo.2i2c.cloud. Last bit that's missing is launching that's currently failing due to a 403 related to scopes. Also last two commits were hacks to quickly fix a couple of issues I had with registry credentials. They weren't properly investigated yet.

consideRatio:
> Last bit that's missing is launching that's currently failing due to a 403 related to scopes. 

Looking at this now - we haven't provided permissions to the jupyterhub registered service "binder" to launch servers.

consideRatio:
I raised a question about binderhub docs for auth in https://github.com/jupyterhub/binderhub/issues/1854.

GeorgianaElena:
Yuhooooo 🚀

config/clusters/2i2c/binderhub-ui-demo.values.yaml
```diff
@@ -40,7 +40,6 @@ jupyterhub:
     redirectToServer: false
     services:
       binder:
-        display: false
```
consideRatio: This is relevant when a `url` is specified, but defaults to false whenever there isn't an `url` specified to link out to.

config/clusters/2i2c/binderhub-ui-demo.values.yaml
```diff
@@ -0,0 +1,114 @@
...
+      enable_api_only_mode: false
+      image_prefix: us-central1-docker.pkg.dev/two-eye-two-see/binderhub-ui-demo-registry/binderhub-service-
+  extraConfig:
+    # FIXME: set KubernetesBuildExecutor.push_secret again
+    #        without this for some reason the build pods
+    #        search after the wrong secret name (i.e. the default name)
+    #        set by binderhub in KubernetesBuildExecutor.push_secret
+    01-binderhub-service-set-push-secret: |
+      import os
+      c.KubernetesBuildExecutor.push_secret = os.environ["PUSH_SECRET_NAME"]
```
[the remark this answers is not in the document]
consideRatio: Every time I work with binderhub, the `push_secret` config is a big confusion - here are reasons for why its confusing observed in the last ~5 minutes looking:

1. `push_secret` should be configured to a k8s secret name having a `config.json` key providing a docker configuration including credentials to work against a registry (its not the typical k8s imagePullSecret or similar, its a docker client config)
2. `binder-build-docker-config` is a default value for `push_secret`, but its also hardcoded i the official BinderHub helm chart deployment, but in binderhub-service deployment it isn't hardcoded - instead `push_secret` is set by a `extraConfig` entry based on an env variable etc - we shouldn't touch `push_secret` config as users of binderhub-service I think.
3. In binderhub-service chart, we configure credentials that goes into the k8s Secret via chart config `buildPodsRegistryCredentials`, but in the official chart its under config.BinderHub.buildDockerConfig` which is confusing because it isn't a traitlet for the Python class `BinderHub` as expected if listed there.
4. The Deployment resource isn't changed if the credentials under buildDockerConfig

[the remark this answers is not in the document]
consideRatio: TL;DR - I was confused, and remain confused why you observed that.

config/clusters/2i2c/cluster.yaml
```diff
@@ -41,6 +41,14 @@ hubs:
       - basehub-common.values.yaml
       - imagebuilding-demo.values.yaml
       - enc-imagebuilding-demo.secret.values.yaml
+  - name: binderhub-ui-demo
+    display_name: "2i2c Binderhub UI demo"
+    domain: hub.binderhub-ui-demo-demo.2i2c.cloud
+    helm_chart: basehub
+    helm_chart_values_files:
+      - basehub-common.values.yaml
```
consideRatio, line 49: This just enables `nfs` things we won't use, should be removed I think.

consideRatio:
I'll go for a merge to flush what we have to ensure we get the terraform state changes represented - we haven't provided announcement, and **_there are things to refine and think about before the pattern of deploying a configuration like this is replicated_** even though that hasn't yet been specified.

consideRatio (approved):

PR #4123, opened 2024-05-23
[nasa-cryo] Add a new node group with newer k8s version

sgibson91:
Need to delete old node group once it has drained

- related to #4009

### DIFF
--- a/eksctl/nasa-cryo.jsonnet
+++ b/eksctl/nasa-cryo.jsonnet
@@ -25,6 +25,8 @@ local nodeAz = "us-west-2a";
 // A `node.kubernetes.io/instance-type label is added, so pods
 // can request a particular kind of node with a nodeSelector
 local notebookNodes = [
+    { instanceType: "r5.xlarge" },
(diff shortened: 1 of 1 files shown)


eksctl/nasa-cryo.jsonnet
```diff
@@ -25,6 +25,8 @@ local nodeAz = "us-west-2a";
 // A `node.kubernetes.io/instance-type label is added, so pods
 // can request a particular kind of node with a nodeSelector
 local notebookNodes = [
+    { instanceType: "r5.xlarge" },
+    // FIXME: tainted, to be deleted when empty, replaced by equivalent during k8s upgrade
     { instanceType: "r5.xlarge", nameSuffix: "b" },
```
sgibson91, line 30:
```suggestion
    { instanceType: "r5.xlarge" },
    { instanceType: "r5.xlarge", nameSuffix: "b" },  // FIXME: tainted, to be deleted when empty, replaced by equivalent during k8s upgrade
```

Issue #4089, opened 2024-05-16
[Support] Revise profile options for unam.latam.catalystproject.2i2c.cloud


jmunroe:
Re: https://2i2c.freshdesk.com/a/tickets/1588

We need add some additional resource allocations related for unam.latam.catalystproject.2i2c.cloud to support high memory (120G, 500G) science work.

The user-nodepool already has n2-highmem-16 and n2-highmem-64 machine types configured.  Please add these two machines as additional options under `resource_allocation` for catalystproject-latam/unam

This request is only for UNAM and not all Catalyst Project hubs. 

(If there is not engineering capacity for this request in this next sprint, @jmunroe is able to complete the task)

haroldcampbell:
@jmunroe this needs a bit more refinement for this card.

PR #4131, opened 2024-05-24
catalyst-latam: more ram to UNAM hub

AIDEA775:
Ref: https://github.com/2i2c-org/infrastructure/issues/4089

I added more nodes to `node-capacity-info.json` to run `deployer generate resource-allocation choices --num-allocations 1 n2-highmem-16` and `...-64`. However with these autogenerated values, the pods startup fails. The logs say "Insufficient memory". I had to subtract 1MiB to make it work.

The autogenerated numbers appear to be correct. However, I suppose it's good to leave a small margin just in case.

### DIFF
--- a/config/clusters/catalystproject-latam/unam.values.yaml
+++ b/config/clusters/catalystproject-latam/unam.values.yaml
@@ -58,28 +58,23 @@ jupyterhub:
                   cpu_guarantee: 1.5
                   cpu_limit: 3
               # guarantee = limit because we are requesting a complete node, no space for other user
-              # this was generated with deploy generate command
-              # TODO: this was generated with deploy generate command, but with mem -= 1024 * 1024,
-              #       otherwise the pod startup fail with "insufficient memory"
               mem_116_0:
-                display_name: Up to 116.0 GB RAM and 15.4 CPUs
+                display_name: Up to 115G RAM and 15 CPUs
                 kubespawner_override:
-                  mem_guarantee: 124555718656
-                  mem_limit: 124555718656
-                  cpu_guarantee: 15.443
-                  cpu_limit: 15.443
+                  mem_guarantee: 115G
+                  mem_limit: 115G
+                  cpu_guarantee: 15
+                  cpu_limit: 15
                   node_selector:
                     node.kubernetes.io/instance-type: n2-highmem-16
               # guarantee = limit because we are requesting a complete node, no space for other user
-              # TODO: this was generated with deploy generate command, but with mem -= 1024 * 1024,
-              #       otherwise the pod startup fail with "insufficient memory"
               mem_486_4:
-                display_name: Up to 486.4 GB RAM and 63.3 CPUs
+                display_name: Up to 486G RAM and 63 CPUs
                 kubespawner_override:
-                  mem_guarantee: 522271244288
-                  mem_limit: 522271244288
-                  cpu_guarantee: 63.323
-                  cpu_limit: 63.323
+                  mem_guarantee: 486G
+                  mem_limit: 486G
+                  cpu_guarantee: 63
+                  cpu_limit: 63
                   node_selector:
                     node.kubernetes.io/instance-type: n2-highmem-64
       - display_name: Rocker Geospatial with RStudio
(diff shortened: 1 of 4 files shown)


yuvipanda:
@AIDEA775 I regenerated those values, can you try again now?

deployer/commands/generate/resource_allocation/node-capacity-info.json
```diff
@@ -70,5 +70,41 @@
...
+        "capacity": {
+            "cpu": 16.0,
+            "memory": 135081324544
+        },
+        "allocatable": {
+            "cpu": 15.89,
+            "memory": 125139775488
+        },
+        "measured_overhead": {
+            "cpu": 0.447,
```
consideRatio, line 84: The current GKE overhead per node is 454m and 656Mi i think, i expect issues using this atm :/

yuvipanda:
I generated this by spawning an n2-highmem-64 node (by manually editing https://console.cloud.google.com/kubernetes/nodepool/southamerica-east1/latam-cluster/nb-n2-highmem-64?project=catalystproject-392106), and then running the following commands:

```bash
$ deployer generate resource-allocation node-info-update n2-highmem-64
$ deployer generate resource-allocation daemonset-requests catalystproject-latam
$ deployer generate resource-allocation instance-capacities catalystproject-latam
```

deployer/commands/generate/resource_allocation/node-capacity-info.json
yuvipanda: @consideRatio ah, I see. I would like us to not get too much into this, and will suggest a different method forward for now. I think working on this autogenerated part should be treated as a product effort and prioritized accordingly, and I'll bring that up with Giuliano tomorrow as well.

config/clusters/catalystproject-latam/unam.values.yaml
```diff
@@ -23,3 +23,87 @@ jupyterhub:
...
+                  cpu_guarantee: 1.5
+                  cpu_limit: 3
+              # guarantee = limit because we are requesting a complete node, no space for other user
+              # this was generated with deploy generate command
+              # TODO: this was generated with deploy generate command, but with mem -= 1024 * 1024,
+              #       otherwise the pod startup fail with "insufficient memory"
+              mem_116_0:
+                display_name: Up to 116.0 GB RAM and 15.4 CPUs
+                kubespawner_override:
+                  mem_guarantee: 124555718656
```
yuvipanda, line 67: Given @consideRatio's concerns in https://github.com/2i2c-org/infrastructure/pull/4131#pullrequestreview-2081777499, and the fact that we want to treat this as an issue of size '1', I think we shouldn't try to fiddle with the resource allocation script as part of this PR - especially given that the catalyst project hubs don't currently use it anyway!

Given that, I suggest the following values for the profiles:

1. `Up to 115GB RAM and 15 CPUs`, with mem_guarantee and mem_limit set to just `115G`, and cpu_guarantee and cpu_limit set to `15`. This should give us enough 'headroom' to not need tweaking for a while. This goes on the `n2-highmem-16` nodes
2. `Up to 486 GB RAM and 63 CPUs`, with mem_guarantee and mem_limit set to `486GB` and cpu_guarantee and cpu_limit set to `63`. Again, this should give us enough headroom to not need tweaking for a while. This goes on `n2-highmem-64` nodes.

This would allow us to service this request and get the community going, and not need everything to be 'perfect'. Does that seem ok to you, @AIDEA775?
4.

yuvipanda (changes requested):

deployer/commands/generate/resource_allocation/node-capacity-info.json
yuvipanda: Suggested https://github.com/2i2c-org/infrastructure/pull/4131#discussion_r1616621073

consideRatio: Thanks @yuvipanda I'll drop thinking about this!

config/clusters/catalystproject-latam/unam.values.yaml
AIDEA775: Okay! That sounds good!

PR #4136, opened 2024-05-24
AWI-CIROH: Fix missing auth_state config

yuvipanda:
Without this, allowed_groups was not being effective in restricting GPU users

Ref: #3957


yuvipanda:
This requires me to log people out, so I'm going to wait for a bit.

Issue #4144, opened 2024-05-29
[EVENT] OpenScapes Workshop, May 30th


yuvipanda:
### The link towards the Freshdesk ticket this event was reported

https://2i2c.freshdesk.com/a/tickets/1524

### The GitHub handle or name of the community representative

BriannaLind

### The date when the event will start

May 30th, 2024, 12PM EST

### The date when the event will end

May 30th, 2024, 5PM EST

### What hours of the day will participants be active? (e.g., 5am - 5pm US/Pacific)

12PM-5PM EST

### Are we three weeks before the start date of the event?

- [X] Yes
- [ ] No

### Number of attendees

250

### Make sure to add the event into the calendar

- [ ] Done

### Does the hub already exist?

- [X] Yes
- [ ] No

### The URL of the hub that will be used for the event

https://workshop.openscapes.2i2c.cloud/

### Will this hub be decommissioned after the event is over?

- [ ] Yes
- [X] No

### Was all the info filled in above?

- [X] Yes
- [ ] No

### Quotas from the cloud provider are high-enough to handle expected usage?

- [X] Yes
- [ ] No

yuvipanda:
This is basically the same as https://github.com/2i2c-org/infrastructure/issues/3743, so I'll take the same action as was done there.

yuvipanda:
Workshop over

PR #4145, opened 2024-05-29
openscapes/workshop: Put more users on r5.16xlarge nodes

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/4144


config/clusters/openscapes/workshop.values.yaml
```diff
@@ -80,7 +80,7 @@ basehub:
                     cpu_guarantee: 3.725
                     cpu_limit: 3.725
                     node_selector:
-                      node.kubernetes.io/instance-type: r5.xlarge
+                      node.kubernetes.io/instance-type: r5.4xlarge
```
consideRatio, line 83: Typo to declare 4x for this profile specifically?

```suggestion
                      node.kubernetes.io/instance-type: r5.16xlarge
```

config/clusters/openscapes/workshop.values.yaml
```diff
@@ -80,7 +80,7 @@ basehub:
...
                 mem_60_6:
                   display_name: 60.6 GB RAM, upto 15.6 CPUs
```
consideRatio, line 85: There are gow peofiles not set go 16x, but staying at 4x - but you know they won't use those peofiles during the workshop perhaps?

yuvipanda: Correct, they had mentioned which ones they're going to use and not these

PR #4203, opened 2024-06-11
Update global password for Openscapes Workshop hub

sgibson91:
- requested in: https://2i2c.freshdesk.com/a/tickets/1708

(Misclick, accidental push to this repo instead of my fork)


sgibson91:
I have deployed and tested this and it works with the new password, so self-merging

Issue #4169, opened 2024-05-30
Write a runbook for spinning up an ephemeral hub that can launch sessions


yuvipanda:
This would be a runbook that describes how to replicate the efforts in https://github.com/2i2c-org/infrastructure/issues/4133. It doesn't have to be *exactly* detailed - simply detailed enough for another engineer to follow. It should also describe choices made one way or another (such as multiple domains vs subpaths in single domain), with some (but not necessarily detailed) explanation for why (primarily so we can change our mind later if needed - the goal is not perfection)

## Decisions made so far

- two separate domains, one for binderhub UI and one for the jupyterhub. 
- A logged out homepage, and a logged-in homepage https://github.com/2i2c-org/infrastructure/issues/4168#issuecomment-2141701666

Issue #4134, opened 2024-05-24
[New Hub] binder.pythia.2i2c.cloud


yuvipanda:
A part of https://github.com/2i2c-org/infrastructure/issues/4109.

## Phase 2: Cluster

This should be on the Project Pythia cluster on **AWS**.

#### Phase 3.1: Initial setup (**READY**)

| Question | Answer | Notes |
| - | - | - |
| Name of the hub | pythia-binder | |
| Dask gateway? | no | |
| URL | binder.pythia.2i2c.cloud | | 
| BinderHub UI? | yes | | 
| Authentication Mechanism | None | |
| Admin Users | NA | Because BinderHub UI will be used | 
| Persistent Home Directory | no ||

#### Phase 3.2: Object storage access (**READY**)

| Question | Answer | Notes |
| - | - | - |
| Scratch bucket enabled? |  no |  |
| Persistent bucket enabled? | no |  |
| Requestor pays requests to external buckets allowed? | no | |  

#### Phase 3.3: Profile List (**READY**)

N/A, as BinderHub UI will be used

#### Phase 3.31: Resource configuration (**READY**)

No profile list, properties set under `singleuser`

| Question | Answer |
| - | -|
| memory limit | 1G |
| memory request | 256M |
| CPU limit | 0.25 |
 
#### Phase 3.4: Authentication tuning (**READY**)

| Question | Answer |
| - | - |
| Authentication Mechanism | None |

#### Phase 3.5: Profile List finetuning (**NA**)

NA, as BinderHub UI will be used

#### Phase 3.6: BinderHub config (**NOT READY**)

| Question | Answer |
| - | - |
| Allowed Repositories? | `^ProjectPythia/.*$` |
| Banner Message | "" | 
| BinderHub About Message | Binder for use with Project Pythia |

## Note

This differs from #4133 in two ways, described in https://github.com/2i2c-org/infrastructure/issues/4134#issuecomment-2164360396

yuvipanda:
@ktyle you may be interested in this issue :)

jmunroe:
I think the resources on this binder.projectpythia.2i2c.cloud hub should be the same as those in https://github.com/2i2c-org/infrastructure/issues/4005 (including GPU)

Also, for Project Pythia, access to this binder should be restricted. The main use case would be integration with a CI/CD pipeline using GitHub actions.

yuvipanda:
[...]

PR #4205, opened 2024-06-11
Add runbook on binderhub-ui hub

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/4169

### DIFF
--- a/docs/howto/features/binderhub-ui.md
+++ b/docs/howto/features/binderhub-ui.md
@@ -10,16 +10,22 @@ We call this a `binderhub-ui` style hub and the primary features offered would b
 
 ## Important architectural decisions made
 
-1. **Two separate domains, one for binderhub UI and one for the jupyterhub**
-1. **A logged out homepage, and a logged-in homepage**
+1. **Two separate domains, one for binderhub UI and one for JupyterHub**
+   Keeping them separate will help with having clean and correct sharing URLs without having them be based off the `hub/services/:name` path.
 
+1. **A logged out homepage, and a logged-in homepage**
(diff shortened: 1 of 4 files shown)


docs/howto/features/binderhub-ui.md
```diff
@@ -0,0 +1,181 @@
...
+        #        search after the wrong secret name (i.e. the default name)
+        #        set by binderhub in KubernetesBuildExecutor.push_secret
+        01-binderhub-service-set-push-secret: |
+          import os
+          c.KubernetesBuildExecutor.push_secret = os.environ["PUSH_SECRET_NAME"]
+      extraEnv:
+        - name: JUPYTERHUB_API_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: '{{ include "jupyterhub.hub.fullname" . }}'
```
yuvipanda, line 144: Let’s just say “hub” here insteaf of templating. It will never be a different value for us.

yuvipanda: This template name i also consider an internal implementation detail i don’t want to dive into and depend on

yuvipanda:
I'm writing out new code review guidelines that should help cover this (and we can discuss that during our 1:1 tomorrow). I haven't read through this, but I suggest the following course of action:

1. Merge it
2. Follow it to do https://github.com/2i2c-org/infrastructure/issues/4134
3. Make changes as you find issues, as well as make additions for the two ways in which that is different from binderhub-ui-demo
4. Make a PR with those changes

GeorgianaElena:
Good plan @yuvipanda! Thank you. Going for a merge then 🚀

Issue #4090, opened 2024-05-16
[Request deployment] New Hub: UW ISEA


colliand:
This hub will serve the ISEA Hackweek hosted by the University of Washington during July 2024. 
https://www.amplifylearn.ai/isea/

### Hub important dates

- **Target Start Data:** 2024-06-22
- **Required Start Date:** 2024-07-01
- **Shutdown Date:** 2024-07-31

## Phase 2: Cluster

Deploy on 2i2c-aws-us

#### Phase 3.1: Initial setup (**READY**)

| Question | Answer | Notes |
| - | - | - |
| Name of the hub | isea | |
| Dask gateway | no | |
| Splash image | AmplifyLearn.AI logo | see comment below   |
| URL | https://www.amplifylearn.ai/isea/ | |
| funded_by | [Institute of Education Sciences](https://ies.ed.gov/) (IES)  |  | 
| Authentication Mechanism | GitHub (via GitHubOAuthenticator) |
| Admin Users | @lovenoor  | 

At the end of this phase, both 2i2c engineers and the admin users mentioned can login to the hub.

#### Phase 3.2: Object storage access  (**READY**)

| Question | Answer | Notes |
| - | - | - |
| Scratch bucket enabled? |  no |  |
| Persistent bucket enabled? | no |  |
| Requestor pays requests to external buckets allowed? | no | |  

#### Phase 3.3: Profile List (**READY**)

These are the standard profile list options we want to start with. Can be customized later

| Question | Answer | Notes |
| - | - | - |
| profileList used? | no | | 
| Image? |  jupyter/scipy-notebook | Use latest available date tag for [jupyter/scipy-notebook](https://quay.io/repository/jupyter/scipy-notebook?tab=tags)
| RStudio (with Rocker)? |  no  |  |
| Allow users to specify any image they want to use? | no | Enable `unlisted_choice`| 
| Max RAM option allowed |   guarantee: 384M   limit: 1.5G  |  (this is the same as the cloudbank hubs) |
| GPU enabled? | no | |
| Default Interface | JupyterLab | |
| Allow multiple concurrent servers per user? | no | Translates to `allowNamedServers` |
 

#### Phase 3.4: Authentication tuning (**READY**)

| Question | Answer |
| - | - |
| Authentication Mechanism | GitHub (via GitHubOAuthenticator) |
| Org based access? | no |
| List of GitHub orgs / teams to be granted access |   |
| Initial hub admin user | @lovenoor |
[...]

PR #4207, opened 2024-06-12
2i2c-aws-us: add isea hub

AIDEA775:
Ref: #4090


AIDEA775:
Self-merging because it's a pretty standard hub.

Issue #3320, opened 2023-10-24
Alert when Azure Fileshare is about to run out of disk space


yuvipanda:
https://github.com/2i2c-org/infrastructure/pull/3085 adds alerts for disk storage being full to GCP clusters. We should add this for Azure as well - https://2i2c.freshdesk.com/a/tickets/1070 was just caused by this.

## Tasks

- [x] Make sure whoever is working on this has pagerduty rights to create appropriate webhooks
- [x]  Create a PagerDuty integration (https://www.pagerduty.com/docs/guides/azure-integration-guide/). This will give us a webhook URL
- [x] Use terraform to create an alert (https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_metric_alert). The example there has a place for the webhook URL to slot into
- [x] Figure out what metric to actually use in the alert. https://learn.microsoft.com/en-us/azure/storage/common/storage-insights-overview (or related documentation pages) may help find the metric to use
- [x] Test that it fires (by changing the alert % to low). Look in `#pagerduty-notifications` channel - make sure it comes there
- [ ] Communicate back when done via https://2i2c.freshdesk.com/a/tickets/1410 to Amar

## Relavant info

https://github.com/2i2c-org/infrastructure/blob/main/terraform/gcp/pagerduty.tf is related code we have for GCP. The `pagerduty` related parts (via `sops`) will be the **same**, but the notification part will be different.

yuvipanda:
We need to:

1. Create a PagerDuty integration (https://www.pagerduty.com/docs/guides/azure-integration-guide/). This will give us a webhook URL
2. Use terraform to create an alert (https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_metric_alert). The example there has a place for the webhook URL to slot into
3. tada!

yuvipanda:
This just happened again :(

yuvipanda:
We should prioritize this soon - we have this alerting on other cloud providers, and it has meant we haven't had this issue crop up in any hubs there for a while. https://2i2c.freshdesk.com/a/tickets/1410 was the last time this happened in UToronto (earlier this week)

sgibson91:
I will be picking this up when I have completed #3908

sgibson91:
I believe that I have enough rights to add the itegration in PagerDuty and have done the "correct thing", but would like a double check please.

[...]

PR #4234, opened 2024-06-17
Update UToronto terraform config to match UI

sgibson91:
This PR updates our terraform config for disk storage to match what's in the Azure UI for the Toronto cluster. I used the workaround [here](https://github.com/2i2c-org/infrastructure/issues/890#issuecomment-1879072422) to deploy this. It also explains a discrepancy in the numbers I was seeing when deploying the fileshare alerting in #3320


sgibson91:
This has been deployed

PR #4229, opened 2024-06-13
[New Hub] binder.pythia.2i2c.cloud

GeorgianaElena:
Closes https://github.com/2i2c-org/infrastructure/issues/4134

Couldn't make launching to work yet. Currently getting errors like:

```
[E 240613 18:47:14 launcher:208] Error creating user binder-examples-requirements-nm8uy25q: HTTP 403: Forbidden
    b'{"status": 403, "message": "Missing or invalid credentials."}'
[E 240613 18:47:14 builder:749] Retrying launch of https://github.com/binder-examples/requirements after error (duration=0s, attempt=1): HTTPError()
```

Which might indicate that the api token is not being passed correctly in the user creation request 🤔

PR #4243, opened 2024-06-18
[docs] Update binderhub style hub docs

GeorgianaElena:
Follow-up to https://github.com/2i2c-org/infrastructure/pull/4229 this improves the docs and adds a deployer command to generate some sample config to make our lives easier 😅.

This means that now the docs have a section that instruct the person doing the deploy to **check** some of the configuration that the deployer has generated, rather that manually adding it themselves.

There is also a section about setting up the registry  that needs to be done manually and the optional restriction of repos when the binderhub is not authemticated.


GeorgianaElena:
The failing tests are not related, see https://github.com/2i2c-org/infrastructure/issues/4248, so I will hit merge to unblock other people do the new binderhub hub requests.

Issue #4238, opened 2024-06-17
Migrate AWI-CIROH to a new GCP project with billing managed by them


yuvipanda:
Note: They did give us more than 6 weeks notice, in a meeting attended by me and @jmunro. Unfortunately while I got access, I didn't actually create an appropriate issue at that point.

AWI-CIROH is moving from a billing account + project managed by us to one managed by them. This is great news, because we no longer have to pass through cloud costs here. 

However, we will have to complete this migration before June 30th.

I've received access to this google cloud project, and invited everyone else to it as well with appropriate permissions.

https://2i2c.freshdesk.com/a/tickets/1687 has more information.

```[tasklist]
- [x] Get all engineers to accept permissions on this GCP project
- [x] Create new infrastructure on this new GCP project: https://github.com/2i2c-org/infrastructure/pull/4262
- [x] Transfer *initial* filestore contents
- [x] Do final migration (involving some downtime)
- [ ] Have the community validate this
```

sgibson91:
Invite accepted!

GeorgianaElena:
I don't see any invite in my email inbox :/ Wondering if it can be resent?

sgibson91:
@GeorgianaElena I have just added you again

GeorgianaElena:
Hmm, still can't see it :( . Was it sent to georgianaelena@2i2c.org?

sgibson91:
@GeorgianaElena Yes, have you checked spam?

sgibson91:
<img width="806" alt="Screenshot 2024-06-18 at 11 01 57" src="https://github.com/2i2c-org/infrastructure/assets/44771837/be0016ec-73a3-458f-b408-668c56077cf9">

I can see you in the IAM panel

Try visiting here? https://console.cloud.google.com/iam-admin/iam?authuser=1&project=awi-ciroh

sgibson91:
Hang on, I might be in the wrong project

sgibson91:
I have Slacked you an invite link directly based on what I received in my email

GeorgianaElena:
Invitation accepted. Thanks @sgibson91

sgibson91:
I'm getting a whole bunch of errors trying to deploy the terraform config to the new AWI-CIROH project.

I think these ones specifically are related to organisational setting:

```
 Error: googleapi: Error 412: Request violates constraint 'constraints/storage.uniformBucketLevelAccess', conditionNotMet
│ 
[...]

PR #4261, opened 2024-06-20
[terraform] Create a network in GCP instead of relying on the existence of an existing 'default' network

sgibson91:
Previously when we have deployed to GCP projects using terraform, we have read in a network called `default` that appears in each new project, rather than creating a new one. However, the organisational controls on the [new AWI-CIROH project](https://github.com/2i2c-org/infrastructure/issues/4238) mean there is no `default` network and our deployments were failing. This PR refactors our GCP terraform code to create a network instead of searching for `default`.

- I am not sure if there is a way in terraform to say "if this data source doesn't exist, create something instead", but my brief google search didn't surface much hope
- When we apply this to our existing projects, it may try to recreate filestores and clusters :(
- I am unsure what the implications are, particularly billing, for projects that will have a `default` network and we create another one

I'd appreciate guidance from @yuvipanda on these points

### DIFF
--- a/terraform/gcp/cluster.tf
+++ b/terraform/gcp/cluster.tf
@@ -73,10 +73,7 @@ resource "google_container_cluster" "cluster" {
     prevent_destroy = true
   }
 
-  // For private clusters, pass the name of the network and subnetwork created
-  // by the VPC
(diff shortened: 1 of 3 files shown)


sgibson91:
I'm just going to manually create a network called default in the UI and not apply these changes, because they will break other projects

Issue #4167, opened 2024-05-30
Handle healthchecks hubs with binderhub-service set up to launch


yuvipanda:
Per https://github.com/2i2c-org/infrastructure/issues/4133#issuecomment-2140722573, our health check scripts currently fail when used on a JupyterHub setup to be in service of binderhub for launching.

Here's a failing health check: https://github.com/2i2c-org/infrastructure/actions/runs/9307796142/job/25620397449

```[tasklist]
- [x] Temporarily skip these tests so that CI doesn't fail
- [x] Figure out what a real health check for this setup should be (See https://github.com/2i2c-org/infrastructure/pull/4263 and comments below on an alternative plan for binderhub ui and null authenticator setup)
- [x] Implement/skip real health check
```

GeorgianaElena:
> Figure out what a real health check for this setup should be

I've been thinking about this today and I couldn't come up with an answer to this, so instead of me going into various rabbit holes trying to come up with something, I'd really appreciate some direction from you @yuvipanda.

My findings so far are that:

- we need a policy on what goes into the `cluster.yaml` file under `hub.domain`. Is it the hub domain or the binderhub one? (this entry is currenlty being used in our health checks)

- running the current health checks on an authenticated binderhub like the demo one that's in the 2i2c cluster actually works but it bypasses binderhub I believe

_I think_ a solid health check would mean sending an API request to binderhub directly to build and launch an image then execute a simple notebook on the server. BUT, I don't think that's easy as we won't have a library that deals with cleaning up the users and servers afterwards and we will need to do that ourselves otherwise we'd run into troubles. So this sounds as a viable solution to me but will require more work than I'd expected initially I believe. Should I go ahead on this path?

yuvipanda:
First, I just *absolutely* love this comment @GeorgianaElena! This is exactly the kind of team conversation I want to be a part of :)

In ephemeral hubs where binderhub also controls launching, there are three things to test:

1. Does building repos work?
2. Can binderhub launch built repos?
3. Can JupyterHub spawn servers?

Our health checks currently only test for (3) and not (1) and (2). We do have upcoming work (with Project Pythia and VEDA) to do (1) and (2) as well - I'll have a productboard link soon.

So I would say that to complete *this* task, we should define it to only test (3). So it'll be on par with our other hubs.

I think the action items here are:

1. Make sure that our existing health checks (which only check for spawning) *do* work, with and without authentication.
[...]

PR #4263, opened 2024-06-20
Misc updates on testing hub server spawing of binderhubs and documentation

GeorgianaElena:
This is currently failing for unauthenticated binderhubs because of some issue with the token being passed to `/user/api/kernelspecs` or dropped. Seeing this in the notebook logs:

```bash
[I 2024-06-20 14:22:15.949 ServerApp] 302 GET /user/deployment-service-check/ (@192.168.15.175) 0.51ms
[W 2024-06-20 14:22:21.607 ServerApp] wrote error: 'Forbidden'
      File "/srv/conda/envs/notebook/lib/python3.10/site-packages/tornado/web.py", line 1788, in _execute
        result = method(*self.path_args, **self.path_kwargs)
      File "/srv/conda/envs/notebook/lib/python3.10/site-packages/tornado/web.py", line 3289, in wrapper
        url = self.get_login_url()
      File "/srv/conda/envs/notebook/lib/python3.10/site-packages/jupyter_server/base/handlers.py", line 782, in get_login_url
        raise web.HTTPError(403)
    tornado.web.HTTPError: HTTP 403: Forbidden
[W 2024-06-20 14:22:21.608 ServerApp] 403 GET /user/deployment-service-check/api/kernelspecs (@192.168.14.203) 1.89ms referer=None
```

But in that get request, the token should be passed because it's done under the same session here

https://github.com/Quansight/jhub-client/blob/860dff7365bdccc5208724390529b00b8198ccef/jhub_client/api.py#L226

Which gets called by https://github.com/Quansight/jhub-client/blob/860dff7365bdccc5208724390529b00b8198ccef/jhub_client/execute.py#L87
so not sure what's that about.

Can it be because of https://github.com/aio-libs/aiohttp/issues/5783? I do see a 302 in the logs

### DIFF
--- a/config/clusters/projectpythia/pythia-binder.values.yaml
+++ b/config/clusters/projectpythia/pythia-binder.values.yaml
@@ -23,7 +23,33 @@ jupyterhub:
     # The profile list in the common values files needs to be emptied
     # otherwise the launching will fail
     profileList: []
-    cmd: jupyter-lab
+    cmd:
+      - python3
+      - "-c"
+      - |
+        import os
+        import sys
+
+        try:
+            import jupyterlab
+            import jupyterlab.labapp
+            major = int(jupyterlab.__version__.split(".", 1)[0])
+        except Exception as e:
+            print("Failed to import jupyterlab: {e}", file=sys.stderr)
+            have_lab = False
+        else:
+            have_lab = major >= 3
+
+        if have_lab:
+            # technically, we could accept another jupyter-server-based frontend
+            print("Launching jupyter-lab", file=sys.stderr)
+            exe = "jupyter-lab"
+        else:
+            print("jupyter-lab not found, launching jupyter-notebook", file=sys.stderr)
+            exe = "jupyter-notebook"
+
+        # launch the notebook server
+        os.execvp(exe, sys.argv)
   hub:
     config:
       BinderSpawnerMixin:
--- a/deployer/health_check_tests/test_hub_health.py
+++ b/deployer/health_check_tests/test_hub_health.py
@@ -46,10 +46,6 @@ async def check_hub_health(hub_url, test_notebook_path, service_api_token):
             hub_url,
             test_notebook_path,
             username=username,
-            user_options={
-                "token": service_api_token,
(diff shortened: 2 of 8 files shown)


deployer/health_check_tests/test_hub_health.py
```diff
@@ -46,6 +46,10 @@ async def check_hub_health(hub_url, test_notebook_path, service_api_token):
             hub_url,
             test_notebook_path,
             username=username,
+            user_options={
+                "token": service_api_token,
+                "image": "quay.io/2i2c-projectpythia/binderhub-ui-binder-2dexamples-2drequirements-55ab5c:50533eb470ee6c24e872043d30b2fee463d6943f",
```
GeorgianaElena, line 51: This should be special-cased for hubs not requiring authentication because of where the Spawner expects to find the token:

https://github.com/2i2c-org/infrastructure/blob/6214ce186ef41c79b62f057073d92d456c9ca998/helm-charts/basehub/values.yaml#L901-L910

config/clusters/projectpythia/pythia-binder.values.yaml
```diff
@@ -23,7 +23,33 @@ jupyterhub:
...
+            have_lab = major >= 3
+
+        if have_lab:
+            # technically, we could accept another jupyter-server-based frontend
+            print("Launching jupyter-lab", file=sys.stderr)
+            exe = "jupyter-lab"
+        else:
+            print("jupyter-lab not found, launching jupyter-notebook", file=sys.stderr)
+            exe = "jupyter-notebook"
+
```
GeorgianaElena, line 50: This is what binderhub does too. Otherwise, in images that don't have jupyter-lab installed, spawning a server will fail

GeorgianaElena:
Based on @yuvipanda's feedback in #4167 I've updated this PR to:
- skip health tests for binderhubs that don't use authentication
- updated the singleuser cmd to cover the cases when jupyter-lab is not installed in the image being launched (similar to what binderhub does) and updated the docs
- updated the docs to mention that in the `cluster.yaml` we should enter the hub's domain rather than the binderhub's so that the tests work (for authenticated setups)

Issue #4162, opened 2024-05-30
[Support] catalystproject-latam, plnc: Add GPU profile (1 GPU + 4 GPU option)


consideRatio:
### The Freshdesk ticket link

https://2i2c.freshdesk.com/a/tickets/1591

### Ticket request type

Feature Request

### Ticket impact

🟩 Low

### Short ticket description

We should enable the community to start servers with 1 and 4 GPUs respectively:

- n1-highmem-4 + 1 nvidia-tesla-T4 GPU (for small workloads)
- n2-highmem-16 + 4 nvidia-tesla-T4 GPUs (for large workloads)

### Related documentation

- https://infrastructure.2i2c.org/howto/features/gpu/

### Additional notes

Practically, I propose a dedicated profile_list entry for GPU, where the machine type + GPU count is determined by a option under the profile.

The image choices etc for this profile isn't specified by the community or me yet, so I propose a call is made and we ask for feedback.

Below are examples of how I think we need to declare kubespawner_override for the options choosing the hardware within the GPU profile

```yaml
          kubespawner_override:
            # also here should be cpu/memory requests/limit to fill up the node picked
            environment:
              NVIDIA_DRIVER_CAPABILITIES: compute,utility
            node_selector:
              node.kubernetes.io/instance-type: n1-highmem-4
            extra_resource_limits:
              nvidia.com/gpu: "1"

          # .....

          kubespawner_override:
            # also here should be cpu/memory requests/limit to fill up the node picked
            environment:
              NVIDIA_DRIVER_CAPABILITIES: compute,utility
            node_selector:
              node.kubernetes.io/instance-type: n1-highmem-16
            extra_resource_limits:
              nvidia.com/gpu: "4"
```

jmunroe:
Thanks @consideRatio .  Your proposal for giving access to appropriately sized CPUs/GPUs looks good to me. Please make the changes.

AIDEA775:
@consideRatio I have added the GPU options to two new image options. Did I understand your proposal correctly?

Also, the `Pulling image "quay.io/pangeo/pytorch-notebook:2024.05.07"` phase takes a lot of time.

fvillena:
Can you add the new machines to the "Bring your own image" option please?.

And we don't need Pangeo, because we don't do geoscience. Can you replace those images with Jupyter-pytorch and Jupyter-tensorflow? 

I think these are the correct images:

- quay.io/jupyter/pytorch-notebook
[...]

PR #4269, opened 2024-06-21
catalystproject-latam: Fix PLNC GPUs

AIDEA775:
Ref: #4162 

The requested images (quay.io/jupyter/pytorch-notebook, quay.io/jupyter/tensorflow-notebook) don't recognize the GPUs and don't have `nvidia-smi` installed.

I tried using the `cuda` tags, but they don't work either.

However, the pangeo images are working correctly. What other images can we use for GPUs?

### DIFF
--- a/config/clusters/catalystproject-latam/plnc.values.yaml
+++ b/config/clusters/catalystproject-latam/plnc.values.yaml
@@ -35,9 +35,9 @@ jupyterhub:
           image: quay.io/jupyter/scipy-notebook:2024-03-04
           default_url: /lab
         profile_options: &profile_options
-          resource_allocation: &resource_allocation
+          resource_allocation:
             display_name: Resource Allocation
-            choices:
+            choices: &resource_choices
               mem_0_7:
                 display_name: Up to 2G of RAM and 1 CPU
                 kubespawner_override:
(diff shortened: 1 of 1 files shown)


AIDEA775:
@yuvipanda can you help me with this one?

According to this, it should work, right?
https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html#cuda-enabled-variant

consideRatio:
I've opened https://github.com/jupyter/docker-stacks/pull/2124 about this @AIDEA775

It can be very messy to try to adjust PATH and LD_LIBRARY_PATH because they are secure environment variables with special rules applicable to them - due to that I'm not sure we can workaround this quickly. Due to that my current idea is that we wait a week at least to see if the PR gets a review and is deemed reasonable to merge.

consideRatio:
@AIDEA775 I pushed a commit bumping the tag, and `nvidia-smi` now executes and I think everything should work fine in these images as well.

> The timeout error still happens sporadically.

It took me 6 minutes to start using the jupyter pytorch image, which included server startup + image pulling. I think it would be good to figure out what takes too long time when timeout happens. I expect it could be either scaling up or image pulling - where the scaling up part is hard to resolve, while the image pulling may be a network performance degradation that could be fixable somehow.

consideRatio (approved):

PR #4293, opened 2024-06-27
docs: reduce operational complexity of enabling auth state

consideRatio:
Instead of deleting the `hub` k8s Secret with a single command, we can instead delete just a key-value pair with a single command. Like this the `proxy` pod won't need to restart in order to enable and re-populate auth state.

Besides that technical change, I've updated the text to reflect that this isn't a disruptive operation. It will typically just be observed for pre-existing users already logged in as a brief back-and-forth redirect to github when visiting a `/hub` path.


consideRatio:
I think it can make sense to request this to be reviewed for knowledge sharing purposes mainly, but also fine to self-merge. I'll go for a self-merge.

I've tested that this no longer leads to the restart of proxy pod, and no disruptions for users etc.

PR #4317, opened 2024-07-01
opensci, big-binder: enable dask-gateway

consideRatio:


consideRatio:
@GeorgianaElena @sgibson91 nice work on the instructions! Setting up daskhub in the newly setup big-binder in opensci cluster worked great!

I added instructions on the need to configure dask-gateway with image pulling credentials when binderhub has pushed to a private registry as well - but beyond that, nothing else seems needed to handle the combination of binderhub based hubs + dask-gateway.

PR #4349, opened 2024-07-04
2i2c, binderhub-ui-demo: apply rebuild fix apparently also needed for GAR

consideRatio:
I believed our google artifact registry (GAR) was public, so I incorrectly specified in another issue that a bug only influenced quay.io registries among our binderhub-ui hubs I was wrong.

I observed this while quickly trying to re-deploy the `binderhub-ui-demo` hub, and figured applying this fix was fast enough to just do instead of opening an issue to get done.


consideRatio:
@sgibson91 this may impact your writing of docs, where I incorrectly had propegated a belief that this was only something we ran into when working against quay.io.

PR #4356, opened 2024-07-05
terraform, azure: fix deprecation warning and bump deps

### LINKED ISSUE #4233: When bumping azurerm terraform provider v3 -> v4, we will lose ability to taint nodes in default pool
### Context

As part of #4200, I bumped the version of the `azurerm` terraform provider from v2 to v3. This has generated a deprecation warning that when we bump again to v4, we will lose the ability to taint nodes in the default pool. This is problematic because we currently use the default node pool for our core nodes and taint them so that singleuser servers and dask workers are not scheduled on them

### Proposal

_No response_

### Updates and actions

_No response_

consideRatio:
- fixes #4233

We apparently wasn't declaring taints for the core node pool anyhow in our azure setup, not for utoronto or pchub, so removing it was fine. I think we've relied on node selectors to ensure user pods etc get started on non-core nodes.

I also went ahead and bumped misc versions of the terraform providers because its good to do from time and I'm looking for things to do. Note that the latest version is still v3, so I couldn't bump to v4 of the azure terraform provider.

### DIFF
--- a/terraform/azure/main.tf
+++ b/terraform/azure/main.tf
@@ -4,34 +4,27 @@ terraform {
   required_providers {
     azurerm = {
       # ref: https://registry.terraform.io/providers/hashicorp/azurerm/latest
-      #
-      # FIXME: In v4 of this module, we will lose the ability to taint the default
-      #        node pool. See comment in azurerm_kubernetes_cluster block and
-      #        deprecation warning in output.
-      #
-      #        Tracked in: https://github.com/2i2c-org/infrastructure/issues/4233
-      #
       source  = "hashicorp/azurerm"
-      version = "~> 3.107"
+      version = "~> 3.111"
     }
 
     azuread = {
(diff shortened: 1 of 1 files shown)


terraform/azure/main.tf
```diff
@@ -153,10 +146,6 @@ resource "azurerm_kubernetes_cluster" "jupyterhub" {
       "hub.jupyter.org/node-purpose" = "core",
       "k8s.dask.org/node-purpose"    = "core"
     }, var.node_pools["core"][0].labels)
-    # node_taints field will be removed in v4.0 of the Azure Provider since the AKS API
-    # doesn't allow arbitrary node taints on the default node pool. A warning that
-    # this field will be deprecated is produced.
-    node_taints = concat([], var.node_pools["core"][0].taints)
```
consideRatio, line 159: This is removed for `default_node_pool`, not for all node pools - so its just removing whats being deprecated and wasn't set.

PR #4411, opened 2024-07-10
terraform, gcp: grant Service Usage Consumer to buckets admin users

### LINKED ISSUE #4318: Update documentation about GCP bucket access via google groups to include granting permissions to incurr cost
We need to grant the google group of users the `Service Usage Consumer` role (via IAM view)  as well as `Storage Admin` on the bucket  in question (via Buckets view) - if not other users won't have such permissions.

### LINKED ISSUE #4319: Identify and fix existing setups of access GCP buckets via google groups
https://github.com/2i2c-org/infrastructure/issues/4318 was about documenting a permissions detail fix influencing non-2i2c users without other cloud permissions setup for them - this issue is about fixing the detail in existing setups.

I think leap and catalystproject-latam has this setup, and we can identify this by looking at GCP projects buckets and their permissions granted. We could also look at google groups we've created.

## Related

- Docs on setting up bucket access using google groups: https://infrastructure.2i2c.org/howto/features/buckets/#allowing-authenticated-access-to-buckets-from-outside-the-jupyterhub

consideRatio:
I think this fixed both of the following, and updates the docs to favor use of `@googlegroups.com` over `@2i2c.org`.

- fixes #4318
- fixes #4319

### DIFF
--- a/docs/howto/features/buckets.md
+++ b/docs/howto/features/buckets.md
@@ -233,9 +233,14 @@ primarily for large data transfer from on-premise systems. Since
 [Google Groups](https://groups.google.com) can be used to control access to
 GCS buckets, it can be used to allow arbitrary users to write to the bucket!
 
-1. With your `2i2c.org` google account, go to [Google Groups](https://groups.google.com) and create a new Google Group with the name
-   "{bucket-name}-writers", where "{bucket-name}" is the name of the bucket
-   we are going to grant write access to.
+1. With your `2i2c.org` google account, go to [Google Groups](https://groups.google.com)
+   and create a new Google Group with the name "{bucket-name}-writers" and email
+   "{bucket-name}-writers@googlegroups.com", where "{bucket-name}" is the name
+   of the bucket we are going to grant write access to.
+
+   Use of `@googlegroups.com` instead of `@2i2c.org` is suitable as this group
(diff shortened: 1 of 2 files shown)


docs/howto/features/buckets.md
```diff
@@ -233,9 +233,14 @@ primarily for large data transfer from on-premise systems. Since
...
-   "{bucket-name}-writers", where "{bucket-name}" is the name of the bucket
-   we are going to grant write access to.
+1. With your `2i2c.org` google account, go to [Google Groups](https://groups.google.com)
+   and create a new Google Group with the name "{bucket-name}-writers" and email
+   "{bucket-name}-writers@googlegroups.com", where "{bucket-name}" is the name
+   of the bucket we are going to grant write access to.
+
+   Use of `@googlegroups.com` instead of `@2i2c.org` is suitable as this group
+   will include non-2i2c members, who otherwise could get control of a 2i2c.org
+   email which isn't a necessary security compromise.
```
consideRatio, line 243: @yuvipanda do you think nudging to use `@googlegroups.com` instead of `@2i2c.org` was good to do here?

A followup question to this - we have two 2i2c.org groups setup atm, should we work to migrate them to a `@googlegroups.com` group?

consideRatio: (I linked to this from our meeting notes since its not tracked so well with a ping)

PR #4412, opened 2024-07-10
eksctl: default to not creating a unused NAT gateway to pay for

### LINKED ISSUE #2313: Remove NAT gateway option for our AWS clusters
### Background

We received this email about the `nasa-cryo` cluster, probably we'll receive it for all AWS clusters we have coupled to the support email soon. I'm not sure if this is a relevant optimization for us, but maybe.

I understand that a NAT often helps translate a range of local IP addresses to one or more publicly available IP addresses. So, if the k8s nodes (representing pods on it) for example doesn't have public IPs themselves, they can route traffic through the NAT which has a public IP.

### Action points

- [ ] Dig into if this is relevant for us to consider in order to minimize cost

### Email

Hello,

We have observed that your Amazon VPC resources are using a shared NAT Gateway across multiple Availability Zones (AZ). To ensure high availability and minimize inter-AZ data transfer costs, we recommend utilizing separate NAT Gateways in each AZ and routing traffic locally within the same AZ.

Each NAT Gateway operates within a designated AZ and is built with redundancy in that zone only. As a result, if the NAT Gateway or AZ experiences failure, resources utilizing that NAT Gateway in other AZ(s) also get impacted. Additionally, routing traffic from one AZ to a NAT Gateway in a different AZ incurs additional inter-AZ data transfer charges. We recommend choosing a maintenance window for architecture changes in your Amazon VPC.

The following is a list of your VPCs and NAT Gateways that are shared across AZ(s), in the format: 'VPC | NAT Gateway':

vpc-052f77e7a342aef18 | nat-080955a4715fbb99b

Please refer to the AWS public documentation on how to create a NAT Gateway [1], and how to configure routes for different NAT Gateway use cases [2].

Should you have any questions or concerns, please reach out to the AWS Support team [3].

[1] https://docs.aws.amazon.com/vpc/latest/userguide/vpc-nat-gateway.html#nat-gateway-working-with
[2] https://docs.aws.amazon.com/vpc/latest/userguide/nat-gateway-scenarios.html
[3] https://aws.amazon.com/support

Sincerely,
Amazon Web Services

Amazon Web Services, Inc. is a subsidiary of Amazon.com, Inc. Amazon.com is a registered trademark of Amazon.com, Inc. This message was produced and distributed by Amazon Web Services Inc., 410 Terry Ave. North, Seattle, WA 98109-5210

---
Reference: https://phd.aws.amazon.com/phd/home?region=us-east-1#/event-log?eventID=arn:aws:health:us-west-2::event/NATGATEWAY/AWS_NATGATEWAY_OPERATIONAL_NOTIFICATION/AWS_NATGATEWAY_OPERATIONAL_NOTIFICATION_f0c13ac4fdcc0cb01ff9db978e82a47681aa7ab6ad87020e5cf690603775e81a&amp;eventTab=details

consideRatio:
Enable/Disable NAT gateway can only be done during cluster creation by eksctl currently, let's default to not having it setup given @yuvipanda's comment https://github.com/2i2c-org/infrastructure/issues/2313#issuecomment-2030742704.

- I think this sufficiently fixes #2313

### DIFF
--- a/eksctl/template.jsonnet
+++ b/eksctl/template.jsonnet
@@ -78,6 +78,14 @@ local daskNodes = [];
     iam: {
         withOIDC: true,
     },
+    vpc: {
+        nat: {
+            // Use of a NAT gateway isn't required for our cluster with k8s
+            // nodes having public instanceTypes, so its disabled to reduce
+            // cost.
+            gateway: "Disable",
+        }
+    }
     // If you add an addon to this config, run the create addon command.
     //
     //    eksctl create addon --config-file=<< cluster_name >>.eksctl.yaml


consideRatio:
@yuvipanda I pinged you for review here because I think this if we are incorrect could lead to notable amount of extra work -- if we need to re-create a future eksctl cluster because we ended up needing to use the NAT gateway for something sometime.

I'm hesitating overall, maybe we should just stay consistent and not go for this?

yuvipanda:
I agree this needs to be somewhat carefully considered, and unfortunately I do not currently have time to review it - and I don't know if I will anytime soon, given various other priorities. So if you feel this needs additional review from me before we can merge it, I think that won't happen anytime soon. If you feel comfortable merging it with work that's commesurate with the story points given to this, you can go ahead.

yuvipanda:
Prioritywise, the binderhub stuff is definitely far higher priority.

consideRatio:
Thank you @yuvipanda for considering this! I went for a close of this PR and related issue - related comment in https://github.com/2i2c-org/infrastructure/issues/2313#issuecomment-2224901394.

PR #4416, opened 2024-07-11
basehub chart: resource limits for binderhub-service build & docker-api pods

### LINKED ISSUE #4371: Copy mybinder.org's build resource limits to all our binderhub-service deployments
I spoke to @jmunroe, and we decided that for now we want to simply copy the resource limits that mybinder.org employs for its build setup to all our binderhub-service deployments.

This includes:

1. https://github.com/jupyterhub/mybinder.org-deploy/blob/a820075a32266f6b84bc25c9eaea78f870ed39de/mybinder/values.yaml#L192 `memory_limit` and `memory_request`
2. https://github.com/jupyterhub/mybinder.org-deploy/blob/a820075a32266f6b84bc25c9eaea78f870ed39de/mybinder/values.yaml#L315 for the docker dind pods (we may have to configure this in a slightly different place)

consideRatio:
This is what can be done without breaking things as part of #4371 i think. The change in this PR is verified to work as intended for the 2i2c/binderhub-ui-demo hub.

Since this PR isn't honoring what I understood the task was meant to accomplish exactly, I'm requesting a review explicitly for this PR @yuvipanda. I realize now that the task title only mentions limits, but I ended up thinking it was about requests and limits.

- fixes #4371

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -37,6 +37,17 @@ binderhub-service:
       - key: hub.jupyter.org/dedicated
         value: user
         effect: NoSchedule
+    resources:
+      # requests are currently zeroed based on discussion in:
+      # https://github.com/2i2c-org/infrastructure/issues/4371#issuecomment-2223168854
+      requests:
+        cpu: 0 # 0.5 in mybinder.org-deploy
+        memory: 0 # 1Gi in mybinder.org-deploy
+      # limits are set to mimic mybinder.org-deploy:
+      # https://github.com/jupyterhub/mybinder.org-deploy/blob/6b921dca969ac2c4ed77fa322b8bf28245c3c4ba/mybinder/values.yaml#L316-L322
+      limits:
+        cpu: 4
+        memory: 4Gi
   config:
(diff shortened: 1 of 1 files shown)


helm-charts/basehub/values.yaml
```diff
@@ -45,6 +56,12 @@ binderhub-service:
       node_selector:
         # Schedule builder pods to run on user nodes only
         hub.jupyter.org/node-purpose: user
+      # memory_request are currently zeroed based on discussion in:
+      # https://github.com/2i2c-org/infrastructure/issues/4371#issuecomment-2223168854
+      memory_request: 0 # 1G in mybinder.org-deploy
```
consideRatio, line 61: The build pods aren't daemonset pods, so they won't cause as much trouble and could possibly have a request. They need to be scheduled on a node with a daemonset pod though, but it felt safer to just go 0 here to avoid complexity.

yuvipanda:
I don't really have much capacity to deeply review this, so I'll simply trust that your testing had gotten this to work, and if we run into actual problems, we can deal with it at that point.

Thanks @consideRatio.

PR #4417, opened 2024-07-11
Add python script and requirements to automatically create daily backups of gcp filestores

### LINKED ISSUE #4393: Create a python script (gcp-filestore-backups) to perform backups and retention of filestore instances
## Inputs

This script should take the following inputs (via commandline, parsed with `argparse`):

1. Name of the filestore instance to create a backup of
2. Name of the share inside the filestore instance to create a backup of (ours is called 'homes' I think)
3. Region (and / or) zone the filestore instance is in
4. Number of days to retain backups for

1-3 are based on what is needed to [create backups](https://cloud.google.com/filestore/docs/backup-restore#create-backup) and 4 is so we can figure out which backup to [delete](https://cloud.google.com/filestore/docs/backup-restore#list-view-update-delete)

## Control Loop

We should design this script using some principles from kubernetes, so that it's easy to maintain and monitor. The fundamental piece is to think of this as a [control loop](https://theithollow.com/2019/09/16/kubernetes-desired-state-and-control-loops/). So the script should do the following:

1. List all existing backups (`gcloud filestore backups list --format json`)
2. Find all the backups that match the filestore and fileshare we care about (by filtering based on inputs 1,2, 3)
3. If there has been no *successful* backup in the last 24h, create one (with `gcloud filestore backups create`). We pick 24h because these are 'daily' backups
4. If there are backups older than now - input#4 (number of days), delete those backups. 
5. Sleep for a period of time (maybe 10min?), go back to step#1

This allows us to run this script multiple times, interrupt it, etc without *too* much worry. If there are extra backups, they will simply get deleted at the end of the retention period. If creating a backup fails, step (3) picks it up when running again.

## Scope

This task is scoped to writing this script in such a way that any engineer with their local credentials will be able to run it from their computer after authenticating with `gcloud` and it would work. Deployment of this script and monitoring will be dealt with as further tasks under https://github.com/2i2c-org/infrastructure/issues/4391

## Client Library Use

Whoever does this task can either choose to parse the output of `gcloud` (with `--format=json`), or use the appropriate [python client library](https://cloud.google.com/python/docs/reference/file/latest). It's fairly minimal so the choice doesn't matter - do whatever feels comfortable to you, and don't spend too much time picking one or the other! If you have no opinions, pick using `gcloud`.

sgibson91:
- fixes #4393

### DIFF
--- a/helm-charts/images/gcp-filestore-backups/gcp-filestore-backups.py
+++ b/helm-charts/images/gcp-filestore-backups/gcp-filestore-backups.py
@@ -113,7 +113,7 @@ if __name__ == "__main__":
         type=str,
         help="",
     )
-    parser.add_argument("filestore_name", metavar="filestore-name", type=str, help="")
+    parser.add_argument("filestore_name", type=str, help="")
     parser.add_argument(
         "region",
         type=str,
--- a/helm-charts/images/gcp-filestore-backups/gcp-filestore-backups.py
+++ b/helm-charts/images/gcp-filestore-backups/gcp-filestore-backups.py
@@ -7,6 +7,13 @@ from datetime import datetime, timedelta
 import jmespath
 
 
+def extract_region_from_zone(zone: str):
+    """
+    Parse a GCP zone (e.g. us-central1-b) to return a region (e.g. us-central1)
+    """
+    return "-".join(zone.split("-"[:2]))
+
+
 def get_existing_backups(
     project: str, region: str, filestore_name: str, filestore_share_name: str
 ):
@@ -180,8 +187,9 @@ def delete_old_backups(backups: list, region: str):
 
 
 def main(args):
+    region = extract_region_from_zone(args.zone)
     filestore_backups = get_existing_backups(
-        args.project, args.region, args.filestore_name, args.filestore_share_name
+        args.project, region, args.filestore_name, args.filestore_share_name
     )
     recent_filestore_backups, old_filestore_backups = (
         filter_backups_into_recent_and_old(filestore_backups, args.retention_days)
@@ -191,10 +199,10 @@ def main(args):
         args.project,
         args.filestore_name,
         args.filestore_share_name,
-        args.region,
+        region,
         args.zone,
     )
-    delete_old_backups(old_filestore_backups, args.region)
+    delete_old_backups(old_filestore_backups, region)
 
 
 if __name__ == "__main__":
@@ -211,11 +219,6 @@ if __name__ == "__main__":
         "project",
         type=str,
         help="The GCP project the Filestore belongs to",
-    )
-    parser.add_argument(
-        "region",
-        type=str,
-        help="The GCP region the Filestore is deployed in, e.g. us-central1",
     )
     parser.add_argument(
         "zone",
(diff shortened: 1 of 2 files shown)


helm-charts/images/gcp-filestore-backups/gcp-filestore-backups.py
```diff
@@ -0,0 +1,134 @@
...
+        subprocess.check_call(
+            [
+                "gcloud",
+                "filestore",
+                "backups",
+                "create",
+                f"{args.filestore_name}-{args.filestore_share_name}-backup-{datetime.now().strftime('%Y-%m-%d')}",
+                f"--file-share={args.filestore_share_name}",
+                f"--instance={args.filestore_name}",
+                f"--instance-location={args.region}-b",
```
sgibson91, line 74: I need to create another input variable for `zone` here, and maybe use `--instance-zone` instead. Adding `-b` was a fix to get this working again the 2i2c shared cluster.

yuvipanda (commented):
The gist of this looks right to me for a reconciliation loop! Thanks for working on this, @sgibson91.

I'd suggest to make it clearer, the main function be split into separate functions that do the various parts of the loop (get_existing_backups, create_backup_if_necessary, delete_old_backups).

helm-charts/images/gcp-filestore-backups/gcp-filestore-backups.py
```diff
@@ -0,0 +1,134 @@
...
+                f"--region={args.region}",
+                # This operation can take a long time to complete and will only take
+                # longer as filestores grow, hence we use the `--async` flag to
+                # return immediately, without waiting for the operation in progress
+                # to complete. Given that we only expect to be creating a backup
+                # once a day, this feels safe enough to try for now.
+                "--async",
```
yuvipanda, line 81: I think calling it `--async` is fine, as long as our logic to detect 'do we need to create another backup right now?' takes into account incomplete backups. We want to avoid running into the [rate limit](https://cloud.google.com/filestore/docs/limits) and also avoid endlessly creating a ton of backups (as that can cost money).

sgibson91: I checked and it looks like this is handled on GCP's side. I ran the script twice in a row, and the returned JSON was already populated with the backup just created, even though it's state was "FINALIZING". So I didn't need to make any tweaks to the script here.

```
$ python gcp-filestore-backups.py pilot-hubs-homedirs two-eye-two-see us-central1-b
There have been no recent backups of the filestore for project two-eye-two-see. Creating a backup now...
Create request issued
Check operation [projects/two-eye-two-see/locations/us-central1/operations/operation-1720791941383-61d0d18660679-d749fa00-3bdf9958] for status. 
No outdated backups found.
...

$ python gcp-filestore-backups.py pilot-hubs-homedirs two-eye-two-see us-central1-b
[
  {
    "capacityGb": "5120",
    "createTime": "2024-07-12T13:45:41.576982250Z",
    "name": "projects/two-eye-two-see/locations/us-central1/backups/pilot-hubs-homedirs-homes-backup-2024-07-12",
    "sourceFileShare": "homes",
    "sourceInstance": "projects/two-eye-two-see/locations/us-central1-b/instances/pilot-hubs-homedirs",
    "sourceInstanceTier": "BASIC_HDD",
    "state": "FINALIZING"
  }
]
Recent backup found.
No outdated backups found.
...
```

PR #4421, opened 2024-07-12
[opensci, small-binder] Remove repo allowlist

jnywong:
~~I am trying to build documentation at https://2i2c-nasa-binder-access.github.io/docs/ and would like to connect to a the open science small binder service to it.~~

Closes https://github.com/2i2c-org/meta/issues/1299


yuvipanda:
@jnywong based on https://github.com/2i2c-org/meta/issues/1299, can you instead:

1. Remove the allow_list config completely
2. Update the banner text to say... something else?

That would allow anyone to launch anything, and the difference between this and mybinder.org is that this is always close to NASA data in us-west-2

jnywong:
Yep can do!

Issue #4394, opened 2024-07-08
Create a docker image for the GCP filestore script to run in


yuvipanda:

sgibson91:
quay.io repository lives at https://quay.io/repository/2i2c/gcp-filestore-backups

PR #4429, opened 2024-07-15
Containerise gcp-filestore-backups script and add to chartpress specification

sgibson91:
- fixes https://github.com/2i2c-org/infrastructure/issues/4394


sgibson91:
I actually don't think the entrypoint added here is needed. Will remove in a future PR relating to this work.

Issue #4396, opened 2024-07-08
Create cloud sevice accounts with permissions needed for gcp-filestore-backup to run


yuvipanda:
With terraform, we need to create a:

1. GCP cloud service account (resource `google_service_account`)
2. With permissions for `file.backups.*` (from https://cloud.google.com/filestore/docs/iam) (resource `google_project_iam_custom_role` and `google_project_iam_member`)
3. Binding to the service account created in https://github.com/2i2c-org/infrastructure/issues/4395 (resource `google_service_account_iam_binding`)
4. An output that specifies exactly what annotation must be put on the kubernetes service account

This is a *simpler* version of https://github.com/2i2c-org/infrastructure/blob/main/terraform/gcp/workload-identity.tf. It's simpler because `workload-identity.tf` needs to support multiple hubs per cluster, while in this case we do not. 

This should be disabled by default, and enable-able per-cluster as needed.

PR #4436, opened 2024-07-16
Create k8s resources in support chart to run gcp-filestores-backup

### LINKED ISSUE #4395: Create a deployment and service accounts in the support/ chart for gcp-filestore-backup to run
Once https://github.com/2i2c-org/infrastructure/issues/4394 is done, we need to create enough kubernetes resources that can deploy this service. That would need to be a deployment that's running the image as a container, and a service account attached to it (so we can grant it the required cloud permissions).

The service account must have configurable annotations (so we can use [workload identity](https://cloud.google.com/iam/docs/workload-identity-federation) to give it cloud permissions), and the deployment must set [strategy](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy) to 'Recreate' (so we minimize chances of two instances running together).

This should be disabled by default, and enable-able per-cluster as needed.

sgibson91:
- fixes #4395

### DIFF
--- a/helm-charts/support/values.yaml
+++ b/helm-charts/support/values.yaml
@@ -501,6 +501,10 @@ prometheusStorageClass:
     parameters:
       type: pd-balanced
 
+# Setup a pod that will periodically backup the Filestore contents
+gcp-filestore-backups:
+  enabled: false
+
 # A placeholder as global values that can be referenced from the same location
 # of any chart should be possible to provide, but aren't necessarily provided or
 # used.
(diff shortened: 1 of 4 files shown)


yuvipanda (commented):
The fundamentals of this look fine to me! I think now it's tweaking this until it runs ok in the container continuously.

sgibson91:
> I think now it's tweaking this until it runs ok in the container continuously.

I don't think it will run until gcloud has the necessary credentials - so I will have to complete #4396

helm-charts/support/values.yaml
```diff
@@ -501,6 +501,10 @@ prometheusStorageClass:
     parameters:
       type: pd-balanced
 
+# Setup a pod that will periodically backup the Filestore contents
```
sgibson91, line 504:
```suggestion
# Setup a deployment that will periodically backup the Filestore contents
```

Issue #3713, opened 2024-02-09
[Deadline: July 23] [New Hub] NeuroHackademy 2024


jmunroe:
**Note**: Refined by @yuvipanda. non-standard format of this new hub issue is due to the fact we are bringing back an old hub.

1. Revert https://github.com/2i2c-org/infrastructure/pull/3539
2. Remove all the bits about the configurator (because we will be using profiles)
3. Move the hub to `2i2c-aws-us` cluster (because GPUs are needed, and I don't think the 2i2c GCP cluster is in a region with GPUs)
4. Add a standard GPU profile

jmunroe:
See 
- https://github.com/2i2c-org/infrastructure/issues/2681

for the last time we set up a hub for this group.

yuvipanda:
@jmunroe can I ask you to make a new hub request here so we can get this prioritized?

arokem:
Can I make an additional request related to this hub? We'd love for there to be an option for users to commission GPU instances for some sessions (but not all). Is it possible to add a selection menu on startup with instance type?

CC: @noahbenson

yuvipanda:
@arokem hey! we can add a profile list at startup, and restrict users based on github team memberships. Does that work for you?

arokem:
A profile list at startup would be great. I don't think that we need to restrict this by group, though. We'll teach all the members of the hub how to use this list and how to make sure that they are using the right resources for each session.

yuvipanda:
@arokem sounds good. BTW, I know we have historically run this on Google Cloud - if needed, are you ok with us running this on AWS?

arokem:
Yes, that would be fine! In fact, many of the open datasets we work with are already in AWS, so there might even be some advantage to going with AWS.

yuvipanda:
@arokem great, I'll just move this to AWS.

GeorgianaElena:
@arokem, I've deployed the hub and it's running at https://neurohackademy.2i2c.cloud/ 🚀 

I've reused the info from last year on logo, links, funding, authentication and also the user image. Can you please look around and let me know what I should be updating and if there's anything that doesn't work so I can update it?
Thank you!

arokem:
Works great! A few things to update: 

- We need to grant access based on membership in this GitHub organization: https://github.com/NeuroHackademy2024. We'll add all our participants sometime next week. 
- Could you please remind me where I find the documentation for how to update the user image? We might need to introduce a couple of small changes to that. 
[...]

PR #4448, opened 2024-07-18
New neurohackademy hub

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/3713

Because the hub is now on aws rather than gcp, I had to also manually add a CNAME entry towards the ingress-nginx load balancer of the aws cluster in namecheap.

### DIFF
--- a/eksctl/2i2c-aws-us.jsonnet
+++ b/eksctl/2i2c-aws-us.jsonnet
@@ -40,11 +40,9 @@ local notebookNodes = [
         // GPUs in a single AZ are in use and no new nodes can be spawned
         availabilityZones: masterAzs,
     },
-];
-
-local dedicatedNotebookNodes = [
     { 
         instanceType: "r5.xlarge",
+        nameSuffix: 'dedicated',
         labels+: {
             "2i2c.org/community": "neurohackademy"
         },
@@ -57,11 +55,13 @@ local dedicatedNotebookNodes = [
     },
     {
         instanceType: "g4dn.xlarge",
+        nameSuffix: 'dedicated',
         labels+: {
             "2i2c.org/community": "neurohackademy"
         },
         tags+: {
-            "k8s.io/cluster-autoscaler/node-template/resources/nvidia.com/gpu": "1"
+            "k8s.io/cluster-autoscaler/node-template/resources/nvidia.com/gpu": "1",
+            "community": "neurohackademy"
         },
         taints+: {
             "nvidia.com/gpu": "present:NoSchedule",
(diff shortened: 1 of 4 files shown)


eksctl/2i2c-aws-us.jsonnet
```diff
@@ -42,6 +42,37 @@ local notebookNodes = [
     },
 ];
 
+local dedicatedNotebookNodes = [
```
yuvipanda, line 45: You can add these dirctly into `notebookNodes`, without needing a separate variable for them - exactly the same way as we add them to terraform.

eksctl/2i2c-aws-us.jsonnet
```diff
@@ -42,6 +42,37 @@ local notebookNodes = [
...
+        taints+: {
+            "2i2c.org/community": "neurohackademy:NoSchedule"
+        },
+    },
+    {
+        instanceType: "g4dn.xlarge",
+        labels+: {
+            "2i2c.org/community": "neurohackademy"
+        },
+        tags+: {
```
yuvipanda, line 63: Let's add the tag here as well. Tags help us keep track of costs!

eksctl/2i2c-aws-us.jsonnet
```diff
@@ -125,7 +156,28 @@ local daskNodes = [
                 "hub.jupyter.org/dedicated": "user:NoSchedule"
             },
         } + n for n in notebookNodes
-    ] + ( if daskNodes != null then
+    ] + ( if dedicatedNotebookNodes != null then
+        [
+        ng + {
```
yuvipanda, line 161: This can be removed by rolling the dedicated nodes directly into the `notebookNodes` definition.

eksctl/2i2c-aws-us.jsonnet
yuvipanda: The contents don't need to change.

GeorgianaElena: Thanks @yuvipanda!

The `dedicatedNotebookNodes` list added a different prefix to the nodes' name as the node name is constructed via prefix + instance type and needed a way to give them unique ids.

So I've now added them to notebookNodes but with a `dedicated` suffix. Does that sound good?

sgibson91: @GeorgianaElena (in lieu of Yuvi's time difference to help you move forward) this sounds excellent to me

GeorgianaElena: Thanks @sgibson91!

Issue #4479, opened 2024-07-23
[EVENT]  Amerigeo 2024 on binder.opensci.2i2c.cloud (Aug 1st)


GeorgianaElena:
### The link towards the Freshdesk ticket this event was reported

https://2i2c.freshdesk.com/a/tickets/1902

### The GitHub handle or name of the community representative

See freshdesk ticket

### The date when the event will start

August 1st 2024

### The date when the event will end

August 1st 2024

### What hours of the day will participants be active? (e.g., 5am - 5pm US/Pacific)

10am - 2pm Eastern

### Are we three weeks before the start date of the event?

- [X] Yes
- [ ] No

### Number of attendees

150

### Make sure to add the event into the calendar

- [X] Done

### Does the hub already exist?

- [X] Yes
- [ ] No

### The URL of the hub that will be used for the event

binder.opensci.2i2c.cloud

### Will this hub be decommissioned after the event is over?

- [ ] Yes
- [X] No

### Was all the info filled in above?

- [X] Yes
- [ ] No

### Quotas from the cloud provider are high-enough to handle expected usage?

- [x] Yes
- [ ] No

PR #4480, opened 2024-07-23
[To be merged on Aug 1] Update machine and resources for small binder opensci

GeorgianaElena:
Ref #4479


yuvipanda:
@GeorgianaElena this looks fine. I suggest also tuning the idle culler timeout - increase it to say 4h? You can merge this during your daytime on July 30 so there's enough time for them to play with it.

PR #4499, opened 2024-07-25
Move each hub to it's own nodegroup(s) on Openscapes

### LINKED ISSUE #4482: Create separate nodegroups for each hub on the openscapes cluster

sgibson91:
- fixes #4482

### DIFF
--- a/eksctl/openscapes.jsonnet
+++ b/eksctl/openscapes.jsonnet
@@ -25,9 +25,61 @@ local nodeAz = "us-west-2b";
 // A `node.kubernetes.io/instance-type label is added, so pods
 // can request a particular kind of node with a nodeSelector
 local notebookNodes = [
-    { instanceType: "r5.xlarge" },
-    { instanceType: "r5.4xlarge" },
-    { instanceType: "r5.16xlarge" },
+    { instanceType: "r5.xlarge" }, // FIXME: tainted, to be deleted when empty, replaced by equivalent during k8s upgrade
+    { 
+        instanceType: "r5.xlarge",
+        namePrefix: "nb-staging",
+        labels: { "2i2c/hub-name": "staging" },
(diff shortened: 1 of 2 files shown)


eksctl/openscapes.jsonnet
```diff
@@ -25,9 +25,61 @@ local nodeAz = "us-west-2b";
...
 local notebookNodes = [
-    { instanceType: "r5.xlarge" },
-    { instanceType: "r5.4xlarge" },
-    { instanceType: "r5.16xlarge" },
+    { instanceType: "r5.xlarge" }, // FIXME: tainted, to be deleted when empty, replaced by equivalent during k8s upgrade
+    { 
+        instanceType: "r5.xlarge",
+        namePrefix: "nb-staging",
+        labels: { "2i2c/hub-name": "staging" },
+        tags: { "2i2c:hub-name": "staging" }
```
GeorgianaElena, line 33: @sgibson91 I think I found the issue and it's not node selectors but rather the labels.

I've generated and inspected the eksctl.yaml file that gets generated by this jsonnet and the labels are not as expected. For example, for this nodegroup:

```
"instanceType": "r5.xlarge",
 "labels": {
    "2i2c/hub-name": "staging"
 },
```
The instance type is missing as you can see.

The fix I think it is to have:
```suggestion
    { 
        instanceType: "r5.xlarge",
        namePrefix: "nb-staging",
        labels+: { "2i2c/hub-name": "staging" },
        tags+: { "2i2c:hub-name": "staging" }
```

This is what we have for GPUs and https://infrastructure.2i2c.org/howto/features/dedicated-nodepool/.

I think it has to do with how this template works but don't have time to investigate further.

sgibson91:
Thanks @GeorgianaElena

PR #4500, opened 2024-07-25
Delete auto-author-assign-pull-request.yaml workflow

sgibson91:
We don't actually track assignments via PRs anymore, just issues, so removing this reduces the number of GitHub notifications we receive.


yuvipanda:
i approve!

PR #4501, opened 2024-07-25
[Neurohackademy] Add RAM to custom image instances.

arokem:
This is so that I can test our image drafts under realistic conditions.

Or more simply put, I am running out of RAM testing some of our lessons.


yuvipanda:
Let's try it!

Issue #4461, opened 2024-07-20
Switch `staging` hub on openscapes hub to its own EFS


yuvipanda:
As part of https://github.com/2i2c-org/infrastructure/issues/4453, we want to support running one EFS per hub. This allows accurate cost allocation per hub for home directory storage.

This task requires:

- [x] Modify our terraform code to allow creation of multiple EFS instances (taking care to not damage the existing ones). Equivalent work done on GCP in https://github.com/2i2c-org/infrastructure/issues/4334. https://github.com/2i2c-org/infrastructure/pull/4498
- [x] Setup an EFS instance for `staging`. The existing EFS can be retained for `prod` (and temporarily, `workshop`) https://github.com/2i2c-org/infrastructure/pull/4503
- [ ] Move the `staging` hub to use the new EFS instance set up for it. Existing data for the staging hub does not need to be migrated, and can be deleted.

## Definition of done

The staging hub uses a new EFS instance set up specifically for it.

GeorgianaElena:
I have finished the tasks of this issue. Only remaining bit is deleting the previous contents of the staging homedir. Will do so in a later iteration, after doing a workshop migration so I can still use it for testing.

### Note

This task was a bit more complex than initially anticipated and required more effort than planned (first sub-task). Not sure what to do with this info, but leaving it here for reference.

PR #4504, opened 2024-07-26
[openscapes-staging] Use the new filestores' address for homedirs

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/4461


GeorgianaElena:
I've deployed this and checked that the staging hub is using the new efs.

**I haven't deleted yet the staging hub's homedirs contents**

PR #4509, opened 2024-07-26
[NeuroHackademy] Change the default image to the 2024 image and new tag.

arokem:
While we are still working out some issues with our automated builds in https://github.com/NeuroHackademy2024/jhub-image, we'd like to adjust the default image to use to this version that has some new software installed and should be good enough for now. 

Let me know if there is another better way to request such changes (e.g., support email).


yuvipanda:
This is the fastest way, @arokem :)

PR #4512, opened 2024-07-28
[NeuroHackademy] Adds an R profile.

arokem:
Is this correct? I was not sure how to adjust the image just for one image, but I reasoned that maybe this is the way to over-ride the global setting?

### DIFF
--- a/config/clusters/2i2c-aws-us/neurohackademy.values.yaml
+++ b/config/clusters/2i2c-aws-us/neurohackademy.values.yaml
@@ -112,7 +112,7 @@ jupyterhub:
         description: "Start a container with R available"
         image:
           name: quay.io/arokem/nh2024-r
-          tag: "894883dfb3bd"
+          tag: "80b5d555defc"
       - display_name: "Bring your own image"
         description: Specify your own docker image (must have python and jupyterhub installed in it)
         slug: custom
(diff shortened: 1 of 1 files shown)


arokem:
Changed to WIP, because I will need to change the tag once I have the image correctly configured. I am still experimenting with that.

arokem:
Removed the WIP designation, because I have now tested this and this image/tag should work fine for our purposes.

yuvipanda:
I think it should work too - try it after this lands, we can tweak it monday if needed

arokem:
Doesn't look like it works. I get a new profile, but it uses the same image as the regular CPU instance.

config/clusters/2i2c-aws-us/neurohackademy.values.yaml
```diff
@@ -107,6 +107,12 @@ jupyterhub:
           extra_resource_limits:
             nvidia.com/gpu: "1"
           init_containers: *init_containers
+      - display_name: R machine
+        slug: R
+        description: "Start a container with R available"
+        image:
```
arokem, line 113: Does this need to be nested in a `profile_options` block? That's my guess just based on reading the profile for the GPU instance.

yuvipanda: Ah that is the correct diagnosis. I suppose this is the negative of looking at this while on my phone. Sorry for not catching it

arokem: Thanks! See #4513

PR #4516, opened 2024-07-29
[Neurohackademy] Modify R profile to use kubespawner_override options matching Python profile

sgibson91:
@arokem we needed a `kubespawner_override` block to mimic the other profiles. The "Choose your own image" profile with a `profile_options` block is a special case. I've deployed this locally and it works now. I've also set the other overrides to match the Python profile - let me know if that shouldn't be the case.


arokem:
It works! And I can get R/rstudio with this dedicated image. Thanks!

PR #4519, opened 2024-07-29
[Openscapes] Put dask-gateway workers (and schedulers) on their own nodegroups

### LINKED ISSUE #4485: Move dask-gateway workers to their own nodegroups on the openscapes cluster
Similar to https://github.com/2i2c-org/infrastructure/issues/4482 but for dask-gateway.

- [x] Create the same set of dask worker nodes for `staging`, `prod` and `workshop` 
- [x] Configure each of the hubs so their dask workers only spawn on the nodepools designated for them
- [x] Tag each of the nodepool with `2i2c:hub-name` to match the name of the hub
- [x] Tag each of the dask nodepools with `2i2c:node-purpose` set to `worker`

## Definition of done

- [x] Users across these hubs will spawn dask workers on to their own nodepools, and not share them with users on other hubs.
- [x] There is a tag  `2i2c:hub-name`, `2i2c:node-purpose` on nodes spawned by dask-gateway. You can verify this by looking at EC2 instances on the AWS console.

sgibson91:
- fixes #4485

### DIFF
--- a/config/clusters/openscapes/prod.values.yaml
+++ b/config/clusters/openscapes/prod.values.yaml
@@ -18,3 +18,14 @@ basehub:
           oauth_callback_url: "https://openscapes.2i2c.cloud/hub/oauth_callback"
         CILogonOAuthenticator:
           oauth_callback_url: "https://openscapes.2i2c.cloud/hub/oauth_callback"
+
+dask-gateway:
+  gateway:
+    scheduler:
+      extraPodConfig:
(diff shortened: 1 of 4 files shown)


sgibson91:
This needs more investigation as I don't believe the nodeSelectors are working. The dask worker I just span up from the staging hub, triggered a scale-up of the dask-workshop nodegroup.

Issue #4291, opened 2024-06-27
`deployer exec` feature - exec into a shell with the NFS storage filesystem root mounted


consideRatio:
When performing a EFS backup restoration, the restore was done to a new folder in the root of the EFS filesystem. In order to relocate that I had to get a shell where I had the full EFS filesystem mounted - not just a folder associated with a hub's files.

Currently we can quickly exec into a k8s pod with a hub's file system mounted via a command like `deployer exec homes nasa-cryo prod`, but we can't quickly get the cluster's entire filesystem mounted with a command like `deployer exec homes nasa-cryo` or similar.

I think a development of the deployer makes sense to do, to allow us to get a shell with the entire filesystem, as it enables us to relocate restored files in the root to where they should be restored.

## Workaround ideas

<details>

### Idea 1 (doesn't work)

A workaround until then is to update the deployer script's command `deployer exec homes <cluster> <hub>` to do it.

It doesnt work as the deployer references a PVC that is tied to the hub subpath, and there isn't a pvc not tied to the hub subpath.

### Idea 2

Declare the volume by not referencing the PVC, but something like this instead: https://kubernetes.io/docs/concepts/storage/volumes/#nfs
</details>

yuvipanda:
For the specific case that triggered this, I was able to complete the task with the following diff:

```diff
diff --git a/deployer/commands/exec/infra_components.py b/deployer/commands/exec/infra_components.py
index 7ea1fba3..edb33ee2 100644
--- a/deployer/commands/exec/infra_components.py
+++ b/deployer/commands/exec/infra_components.py
@@ -36,7 +36,7 @@ def homes(
             "automountServiceAccountToken": False,
             "volumes": [
                 # This PVC is created by basehub
-                {"name": "home", "persistentVolumeClaim": {"claimName": "home-nfs"}}
+                {"name": "home", "nfs": {"server": "fs-0872256335d483d5f.efs.us-west-2.amazonaws.com", "path": "/"}}
             ],
             "containers": [
                 {
```

I picked up the server name from our config.

We can prioritize this if it becomes a deeply recurring task.

yuvipanda:
We should implement this as a `deployer exec root-homes <cluster> <hub>` for simplicity.

Instead of binding the to the PVC, we should directly use NFS in the pod definition (https://kubernetes.io/docs/concepts/storage/volumes/#nfs)

The server and path information should be gleaned from:

[...]

PR #4545, opened 2024-07-31
[deployer] Update `exec root-homes` command so that it's per hub not per cluster

GeorgianaElena:
Needed this while doing the openscapes efs migration to double check transfer and was not working as expected. This fixes it.

### DIFF
--- a/deployer/commands/exec/infra_components.py
+++ b/deployer/commands/exec/infra_components.py
@@ -53,8 +53,6 @@ def root_homes(
                 .get("baseShareName", base_share_name)
             )
 
-    print(server_ip)
-    print(base_share_name)
     pod_name = f"{cluster_name}-root-home-shell"
     pod = {
         "apiVersion": "v1",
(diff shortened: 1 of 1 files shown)


consideRatio:
@GeorgianaElena there is `deployer exec homes` also, isn't that doing what you now made this command do?

```
# from `deployer exec --help`
homes                       Pop an interactive shell with the home directories of the given hub mounted on /home
root-homes                  Pop an interactive shell with the entire nfs file system of the given cluster mounted on /root-homes
```

A reason for adding this was when performing a EFS backup operation for example - because then you get a folder put in the root folder of the EFS system. Another reason, even though this one is only a convenience reason, is if you'd need to do something for all hub's in the same NFS server.

GeorgianaElena:
@consideRatio the command already existed for https://github.com/2i2c-org/infrastructure/issues/4291. I've just made it now to use the nfs ip of each hub rather than assuming there is only one per cluster, which now isn't the case.

GeorgianaElena:
I needed it when doing the backup in order to make sure that I hadn't transfer anything else but the homedir.

Issue #4576, opened 2024-08-06
[New Hub] [Aug 9] binder.nasa-veda.2i2c.cloud


yuvipanda:
A part of https://github.com/2i2c-org/meta/issues/1368

A part of https://github.com/NASA-IMPACT/veda-jupyterhub/issues/47

## Phase 1: Account

Done

## Phase 2: Cluster

Use existing nasa-veda cluster

#### Phase 3.1: Initial setup (**READY**)

| Question | Answer | Notes |
| - | - | - |
| Name of the hub | binder | |
| Dask gateway? | no | |
| URL | binder.nasa-veda.2i2c.cloud | (may change) | 
| BinderHub UI? | yes | | 
| Authentication Mechanism | None | |
| Admin Users | NA | Because BinderHub UI will be used | 
| Persistent Home Directory | no ||

#### Phase 3.2: Object storage access (**READY**)

| Question | Answer | Notes |
| - | - | - |
| Scratch bucket enabled? |  yes | retention set to 1 days, instead of 7 |
| Persistent bucket enabled? | no |  |
| Requestor pays requests to external buckets allowed? | no | |  

#### Phase 3.3: Profile List (**READY**)

N/A, as BinderHub UI will be used

#### Phase 3.31: Resource configuration (**READY**)

No profile list, properties set under `singleuser`

| Question | Answer |
| - | -|
| memory limit | 2G |
| memory request | 1G |
| CPU limit | 1 |
 
#### Phase 3.4: Authentication tuning (**READY**)

| Question | Answer |
| - | - |
| Authentication Mechanism | None |

#### Phase 3.5: Profile List finetuning (**NA**)

NA, as BinderHub UI will be used

#### Phase 3.6: BinderHub config (**READY**)

| Question | Answer |
| - | - |
| Allowed Repositories? |  all |
| Banner Message | default | 
| BinderHub About Message | default |

### Phase 4: Customizations

- [x] Use `binder.openveda.cloud` as the community manged domain. There's already a CNAME pointing from `binder.openveda.cloud` to `binder.nasa-veda.2i2c.cloud` so all the community managed steps are taken. (**Deadline: Aug 9**)
- [x] Apply same `extra_iam_policy` as the staging hub  (**Deadline: Aug9**)
[...]

PR #4595, opened 2024-08-08
[veda-binder] Update the underlying infra

GeorgianaElena:
This is for https://github.com/2i2c-org/infrastructure/issues/4576

### DIFF
--- a/eksctl/nasa-veda.jsonnet
+++ b/eksctl/nasa-veda.jsonnet
@@ -103,6 +103,9 @@ local daskNodes = [
                 "hub.jupyter.org/node-purpose": "core",
                 "k8s.dask.org/node-purpose": "core"
             },
+            tags+: {
+                "2i2c:node-purpose": "core"
+            },
         },
     ] + [
         ng + {
@@ -117,6 +120,9 @@ local daskNodes = [
             labels+: {
                 "hub.jupyter.org/node-purpose": "user",
                 "k8s.dask.org/node-purpose": "scheduler"
+            },
+            tags+: {
+                "2i2c:node-purpose": "user"
             },
             taints+: {
                 "hub.jupyter.org_dedicated": "user:NoSchedule",
@@ -135,6 +141,9 @@ local daskNodes = [
             },
             labels+: {
                 "k8s.dask.org/node-purpose": "worker"
+            },
+            tags+: {
+                "2i2c:node-purpose": "worker"
             },
             taints+: {
                 "k8s.dask.org_dedicated" : "worker:NoSchedule",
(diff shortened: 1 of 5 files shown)


sgibson91:
I may have broken this now as the build pod is no longer able to connect to the docker socket

```
Events:
  Type     Reason       Age                From               Message
  ----     ------       ----               ----               -------
  Normal   Scheduled    79s                default-scheduler  Successfully assigned binder/build-binder-2dexamples-2drequirements-55ab5c-505-2c to ip-192-168-30-60.us-west-2.compute.internal
  Warning  FailedMount  15s (x8 over 79s)  kubelet            MountVolume.SetUp failed for volume "docker-socket" : hostPath type check failed: /var/run/binder-binder/docker-api/docker-api.sock is not a socket file
```

sgibson91:
My problems above seem to be caused by trying to add the `2i2c:node-purpose` tag - I don't know why because tags should just be at the AWS-level, not the k8s level. But I reverted my attempts and the binder works again now. So I'm going to merge this as-is.

sgibson91:
Currently build pods are not being scheduled on the binder nodegroup. I think the node selector and toleration need to be added to the section below, but I don't know what the appropriate key is. I've tried `extraTolerations`, `extra_tolerations`, and `tolerations` - none of which worked.

https://github.com/GeorgianaElena/pilot-hubs/blob/263fc911b6ff5da4ed3cd22c244d2e00304248de/config/clusters/nasa-veda/binder.values.yaml#L111-L115

eksctl/nasa-veda.jsonnet
```diff
@@ -25,9 +25,50 @@ local nodeAz = "us-west-2a";
 // A `node.kubernetes.io/instance-type label is added, so pods
 // can request a particular kind of node with a nodeSelector
 local notebookNodes = [
-    { instanceType: "r5.xlarge" },
-    { instanceType: "r5.4xlarge" },
-    { instanceType: "r5.16xlarge" },
+    { instanceType: "r5.xlarge" },  // FIXME: tainted, to be deleted when empty, replaced by equivalent during k8s upgrade
+    { instanceType: "r5.xlarge", nameSuffix: "b" }, // FIXME: tainted, to be deleted when empty, replaced by equivalent during k8s upgrade
```
sgibson91, line 29:
```suggestion
    { instanceType: "r5.xlarge" },  // FIXME: tainted, to be deleted when empty, replaced by equivalent
    { instanceType: "r5.xlarge", nameSuffix: "b" }, // FIXME: tainted, to be deleted when empty, replaced by equivalent
```

Issue #4469, opened 2024-07-22
[EVENT] OceanHackWeek August 26-30


abkfenris:
### The link towards the Freshdesk ticket this event was reported

Discussed via email

### The GitHub handle or name of the community representative

@abkfenris, @cpsarason

### The date when the event will start

August 24th, 5 AM Eastern

### The date when the event will end

August 30th 5 PM Eastern

### What hours of the day will participants be active? (e.g., 5am - 5pm US/Pacific)

8-5, but hacking will likely continue for all hours

### Are we three weeks before the start date of the event?

- [X] Yes
- [ ] No

### Number of attendees

35

### Make sure to add the event into the calendar

- [X] Done

### Does the hub already exist?

- [X] Yes
- [ ] No

### The URL of the hub that will be used for the event

https://oceanhackweek.2i2c.cloud/hub/login?next=%2Fhub%2F

### Will this hub be decommissioned after the event is over?

- [ ] Yes
- [ ] No

### Was all the info filled in above?

- [ ] Yes
- [ ] No

### Quotas from the cloud provider are high-enough to handle expected usage?

- [ ] Yes
- [ ] No

abkfenris:
Could we get access to upload data to GCS via Google Groups auth for our tutorials? https://infrastructure.2i2c.org/howto/features/buckets/#allowing-authenticated-access-to-buckets-from-outside-the-jupyterhub

jmunroe:
Hi @abkfenris! I just noticed your question here:

> Could we get access to upload data to GCS via Google Groups auth for our tutorials? https://infrastructure.2i2c.org/howto/features/buckets/#allowing-authenticated-access-to-buckets-from-outside-the-jupyterhub

In future, could you please send requests like this through support@2i2c.org so it gets picked up by our regular support process? (I'll create a ticket on your behalf to capture this)

Is this still issue something you need assistance with?

abkfenris:
Sorry, I don't have a good handle of which requests go where now. I got back to you via email.

abkfenris:
#4687 updates our images for a cert change that is preventing login to Github.

abkfenris:
Random thought/feedback as we go:

We've had some hiccups using gh-scoped-creds this week. Tokens seemed to expire very quickly, or not be acquired at all with unclear errors. Most of the participants have switched to persistent auth with `gh auth login`.

[...]

PR #4615, opened 2024-08-13
Update OHW Python image and user groups

abkfenris:
First update to OceanHackWeek's Python image for 2024 (we're trying pixi!). Additionally updating our user groups and adding Christian as an admin.

xref #4469


yuvipanda:
ty @abkfenris!

PR #4658, opened 2024-08-22
terraform, gcp: update k8s minor versions info to include 1.30

consideRatio:
K8s 1.30 is now part of the regular release channel, we can create new clusters with it.

### DIFF
--- a/terraform/gcp/variables.tf
+++ b/terraform/gcp/variables.tf
@@ -50,9 +50,8 @@ variable "k8s_version_prefixes" {
   # channel, so we may want to remove or add minor versions here over time.
   #
   default = [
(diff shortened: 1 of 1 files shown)


terraform/gcp/variables.tf
```diff
@@ -50,9 +50,8 @@ variable "k8s_version_prefixes" {
   # channel, so we may want to remove or add minor versions here over time.
   #
   default = [
-    "1.27.",
-    "1.28.",
```
consideRatio, line 54: All our GKE clusters are at k8s 1.29, so we don't care about 1.27 and 1.28 versions.

PR #4687, opened 2024-08-26
Update OHW images for cert change blocking Github login

abkfenris:
We just found out that gh-scoped-creds and gh CLI tools weren't working in our images due to change in Github's TLS certs, so this bumps ca-certificates to try to fix that.


abkfenris:
Thanks James!

PR #4742, opened 2024-09-06
terraform, aws: ManagedBy and 2i2c.org/cluster-name tags for all clusters

consideRatio:
AWS tags managed via terraform could be applied to all clusters by updates, and no re-creation was needed etc.

In `eksctl` they are only applied on re-creation though, so there its trouble to adjust tags.

All changes are applied fully except in openscapes that needs to be applied again soonish.


consideRatio:
Docs failure unrelated.

PR #4745, opened 2024-09-09
terraform: add fixme comments about major versions bumps, and reference comments

consideRatio:
- **terraform: bump sops provider in uptime checks**
- **terraform, azure: add reference comments**
- **terraform: bump kubernetes providers lower version bound**
- **terraform: require terraform 1.9+ up from 1.5+**
- **terraform: add fixme comments about major version upgrades**
- **terraform: bump lower bound of google provider**


yuvipanda:
This has moved us away from open source terraform to BSL licensed terraform, and will have to be reverted

PR #4765, opened 2024-09-09
deployer: respect deployer:skip-deploy-hubs, :deploy-hubs, :deploy-support labels

consideRatio:
I want `deployer:deploy-support` to trigger support re-deploy when bumping cert manager version. It wasn't triggered in #4764 because it only re-deploys on support chart changes.

Merging this will test `deployer:deploy-support` label, here goes!


consideRatio:
The upside is that it worked, and the downside is that I forgot to also apply `deployer:skip-deploy-hubs`, making every single thing deploy and be tested... Arrrrrgh!

Issue #4735, opened 2024-09-05
[devseed] Add ability to provision an AWS EBS volume to terraform and deploy for VEDA cluster


sgibson91:
The jupyterhub-home-nfs helm chart expects an EBS volume to be available to act as in-cluster NFS server, but won't create it itself. Therefore, we should add to our terraform config, under `terraform/aws`, the ability to optionally provision an EBS volume. This should be an opt-in feature.

### Definition of Done

- [ ] Our terraform config for AWS provides support for optionally creating an EBS volume
- [ ] This config is used to deploy an EBS volume into VEDA's AWS account and associated with the cluster if required

sgibson91:
fixed by https://github.com/2i2c-org/infrastructure/pull/4783

PR #4783, opened 2024-09-11
Terraform config to provision EBS volumes for jupyter-home-nfs

sunu:
The jupyterhub-home-nfs helm chart expects an EBS volume to be available to act as in-cluster NFS server, but won't create it itself. This PR adds the ability to optionally provision an EBS volume to be used to store user home directories.

Addresses https://github.com/2i2c-org/infrastructure/issues/4735

Not entirely sure if I'm getting the Terraform syntax right here. Would really appreciate any feedback.

cc @sgibson91 @batpad


sgibson91 (approved):
Terraform syntax looks good and it applied successfully! Feel free to merge!

Issue #4706, opened 2024-08-28
(5/5) Ensure gcpFilestoreBackups documentation is up-to-date and the project is continuing to run on 2i2c cluster


sgibson91:
### Context

Now that gcpFilestoreBackups is it's own helm chart that can opted into through 2i2c's support chart, we should ensure that this project continues to run for the 2i2c cluster (GCP shared) and any documentation around enabling this is updated.

### Definition of Done

- [x] gcpFilestoreBackups continues to successfully run on 2i2c cluster
- [ ] Documentation related to gcpFilestoreBackups is up-to-date

PR #4798, opened 2024-09-13
Make gcpFilestoreBackups helm chart an opt-in dependency of support chart

### LINKED ISSUE #4705: (4/5) Reconfigure the support chart so gcpFilestoreBackups chart is an opt-in dependency
### Context

Now that gcpFilestoreBackups is shipped as it's own chart, we need to reconfigure 2i2c's support chart to use that helm chart as an opt-in _dependency_, as opposed to the current feature setup.

### Tasklist
```[tasklist]
- [ ] Clean up any remnants of previous feature code in the support chart
- [ ] Add gcpFilestoreBackups as an opt-in dependent chart of support
```

### Definition of Done

- [ ] 2i2c's support chart uses gcpFilestoresBackup as a dependency
- [ ] Any remnant code of the feature is removed

sgibson91:
- fixes #4705


sgibson91:
I'm not surprised that validation failed because I haven't updated 2i2c's support config yet. I will get a PR up to resolve #4706 so they can be merged quickly and reduce down time for the cluster.

Issue #4794, opened 2024-09-11
[devseed] Expand deployer exec homes command to optionally include EBS volumes for AWS clusters


sgibson91:
We have a `deployer exec homes` command that creates a throwaway k8s pod with the home directories mounted from NFS with read-write access. As we migrate our AWS hubs to using in-cluster NFS backed by EBS volumes, we will also need to mount these data sources. Being able to do this will also assist with the migration process tracked in #4728 as we tried to mount the EFS instance to an EC2 instance but failed.

This issue tracks expansion of the `deployer exec homes` command to optionally include EBS volumes for AWS clusters.

### Definition of Done

- [ ] `deployer exec homes` has an optional flag to also mount EBS volumes for AWS clusters
- [ ] when executed, the throwaway pod has access to both EFS and EBS file systems

sunu:
@sgibson91 I've created a PR at https://github.com/2i2c-org/infrastructure/pull/4803 to address this issue but it takes a slightly different approach.

I realized the jupyter-home-nfs NFS server pod and the throwaway pod could be running on different nodes. And we can't safely mount the EBS backed PVC created by jupyter-home-nfs to multiple nodes at once in read/write mode since EBS volumes don't support that. 

Instead, I've added some optional arguments to the root_homes command so that we can mount an extra NFS backed volume to the throwaway pod. This should allow us to mount a volume backed by the existing EFS instance and another volume backed by the NFS server of jupyter-home-nfs to the same throwaway pod and migrate the data from one to the other.

Let me know if that sounds reasonable.

sgibson91:
Yep, that sounds reasonable, thanks!

PR #4803, opened 2024-09-16
Add optional extra NFS mount to root_homes command

sunu:
Ability to mount an extra NFS volume is useful for data migration between NFS servers.

- fixes https://github.com/2i2c-org/infrastructure/issues/4794


sgibson91 (approved):

sgibson91:
LGTM! I've added a label that will prevent a redeploy of everything, because it's not required here.

Issue #4632, opened 2024-08-16
[16 Sep] Check disk space of prometheus servers for each cluster


sgibson91:
### Context

As part of #4584, we have increased the retention period of prometheus data to 3 years. This will take up more disk, so this issue tracks checking the disk space of each prometheus server and bumps the size if required.

### Task list

```[tasklist]
- [x] Each prometheus server on each cluster is checked for disk space
- [x] Any disks that are approaching full have their size increased
```

### Definition of Done

All prometheus servers for all clusters have enough free disk space to continue storing data for 3 years.

### Pre-defined Definition of Done

- [ ] All the tasks have been completed
- [ ] It has been tested to work and to not break any existing functionality
- [ ] Documentation was added so other engineers can implement this in the future if needed

sgibson91:
- Use grafana to check disk size
- Update disk size in support chart values, may require deleting pv/pvc to redeploy

sgibson91:
- [ ] 2i2c
- [ ] 2i2c-aws-us
- [ ] 2i2c-uk
- [ ] awi-ciroh
- [ ] catalystproject-africa
- [ ] catalystproject-latam
- [ ] cloudbank
- [ ] earthscope
- [ ] gridsst
- [ ] hhmi
- [ ] jupyter-health
- [ ] jupyter-meets-the-earth
- [ ] kitware
- [ ] leap
- [ ] nasa-cryo
- [ ] nasa-esdis
- [ ] nasa-ghg
- [ ] nasa-veda
- [ ] nmfs-openscapes
- [ ] openscapes
- [ ] opensci
- [ ] pangeo-hubs
- [ ] pchub
- [ ] projectpythia
- [ ] smithsonian
- [ ] ubc-eoas
- [ ] utoronto
- [ ] victor

sgibson91:
Merging #4806 broke deployment for 6/8 of the clusters (I don't know what's special about nmfs-openscapes or pchub).

The error is:

```
Error: UPGRADE FAILED: cannot patch "support-prometheus-server" with kind PersistentVolumeClaim: persistentvolumeclaims "support-prometheus-server" is forbidden: only dynamically provisioned pvc can be resized and the storageclass that provisions the pvc must support resize
```

I tried deleting just the pvc `support-prometheus-server` and rerun the support deploy, but that has left the pvc in a pending state (causing helm upgrade to timeout) and I don't want to touch anything more in case I blitz prometheus data.

sgibson91:
[...]

PR #4808, opened 2024-09-16
Increase size of pangeo's prometheus disk

sgibson91:
- related to #4632


sgibson91:
Successfully deployed locally

Issue #4390, opened 2024-07-08
 [EPIC] Ensure disaster recovery backups for home directories


yuvipanda:
(Part of the new internal engineering roadmap, see https://github.com/2i2c-org/meta/issues/1234 for more information)

User home directories contain very important data, and we *must* have backups for disaster recovery. We currently have it for AWS (and it has saved us multiple times), but we need to have that for all cloud providers.

sgibson91:
Are we tracking exploration for an Azure solution for this beyond the work already done?

I met with a Microsoft rep at UKRSECon24 who recommended [snapshots](https://learn.microsoft.com/en-us/azure/storage/files/storage-snapshots-files) as a DR solution for NFS. However, it is only available on the Premium storage tier.

yuvipanda:
@sgibson91 I think we should let Azure be for now, and perhaps plan on moving people to jupyterhub-home-nfs eventually.

PR #4842, opened 2024-09-19
Remove non-functional recovery vault and backup policy for Azure clusters

sgibson91:
In https://github.com/2i2c-org/infrastructure/issues/4390#issuecomment-2358036595, @yuvipanda stated that we were no longer pursuing backup of File Shares on Azure. In view of that, I'm removing the config setting up a recovery vault and backup policy from our Azure terraform config because 1) it is non-functioning, and 2) when we get around to deploying something for Queens University, I don't want to run into unexpected permissions issues relating to broken config.


sgibson91:
Successfully applied to utoronto and pchub

Issue #4730, opened 2024-09-04
[devseed] Document how to enable and configure the `jupyterhub-home-nfs` chart


sgibson91:
We need documentation on how to enable and configure the jupyterhub-home-nfs chart so that other engineers can share this knowledge. 2i2c's engineering documentation is under the docs folder in the infrastructure repo, and is automatically deployed to https://infrastructure.2i2c.org It roughly follows the [diataxis framework](https://diataxis.fr/)

### Definition of Done

- [ ] A how-to guide that engineers can follow to enable and configure jupyterhub-home-nfs chart is added to the engineering docs

PR #4845, opened 2024-09-20
Document how to configure per-user storage quota using jupyter-home-nfs

sunu:
fixes https://github.com/2i2c-org/infrastructure/issues/4730

@sgibson91 I'm not super confident about the commands and steps 😅 Would be great to do another end-to-end migration to test if these steps work or whether we need to change anything in the docs.

### DIFF
--- a/docs/howto/features/storage-quota.md
+++ b/docs/howto/features/storage-quota.md
@@ -1,3 +1,4 @@
+(howto:configure-storage-quota)=
 # Configure per-user storage quotas
 
 This guide explains how to enable and configure per-user storage quotas.
@@ -147,4 +148,4 @@ xfs_quota -x -c 'report -N -p'
 
 This should show us the list of directories being monitored for quota enforcement, and the current usage and quotas for each of the home directories.
 
-If a user exceeds their quota, the user's pod will not be able to start successfully. A hub admin will need to free up space in the user's home directory to allow the pod to start.
+If a user exceeds their quota, the user's pod will not be able to start successfully. A hub admin will need to free up space in the user's home directory to allow the pod to start, using [the `allusers` feature](topic:storage:allusers).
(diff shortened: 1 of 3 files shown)


sunu:
Thanks @sgibson91! I just added a section on how to provision the volume

docs/howto/features/storage-quota.md
```diff
@@ -0,0 +1,171 @@
+(howto:configure-storage-quota)=
+# Configure per-user storage quotas
+
+This guide explains how to enable and configure per-user storage quotas.
+
+```{note}
+Nest all config examples under a `basehub` key if deployingb this for a daskhub.
```
sgibson91, line 7:
```suggestion
Nest all config examples under a `basehub` key if deploying this for a daskhub.
```

docs/howto/features/storage-quota.md
```diff
@@ -0,0 +1,171 @@
...
+
+```{note}
+Nest all config examples under a `basehub` key if deploying this for a daskhub.
+```
+
+## Creating a pre-provisioned disk
+
+The in-cluster NFS server uses a pre-provisioned disk to store the users' home directories. We don't use a dynamically provisioned volume because we want to be able to reuse the same disk even when the Kubernetes cluster is deleted and recreated. So the first step is to create a disk that will be used to store the users' home directories.
+
+For hubs running on AWS, we can create a disk through Terraform by adding a block like the following to the [tfvars file of the hub](https://github.com/2i2c-org/infrastructure/tree/main/terraform/aws/projects):
```
sgibson91, line 14:
```suggestion
For infrastructure running on AWS, we can create a disk through Terraform by adding a block like the following to the [tfvars file of the hub](https://github.com/2i2c-org/infrastructure/tree/main/terraform/aws/projects):
```

Generalise this a bit

sgibson91:
Amazing, thanks!

Issue #4875, opened 2024-09-24
AWS cost attribution (3a/n): start version control the grafana dashboard definitions


consideRatio:

consideRatio:
@GeorgianaElena no as that project is part of the jupyterhub org, and its inclusion there means also aws-ce-grafana-backend needs go be available and open sourced properly.

Let this be entirely 2i2c specific for now, but do make use of the jupyterhub/grafana-dashboard's deploy script to render and deploy the dashboard.

consideRatio:
Should we go for a timeboxed 15 min chat to get started, with the goal to have an empty dummy dashboard to deploy?

GeorgianaElena:
The definition of done was respected for the first 3 panels. The reaming depend on the existence of a variable that I could not figure out and I've opened to track this work https://github.com/2i2c-org/infrastructure/issues/4944.

PR #4934, opened 2024-10-03
Add cost allocation dashboard grafonnet equivalent

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/4875

Only first two panels are ready. I've deployed them into the openscapes grafana under a [`testing` new directory](https://grafana.openscapes.2i2c.cloud/d/782b0b9b537d060d/cloud-cost-attribution?orgId=1&from=now-30d&to=now).

### DIFF
--- a/grafana-dashboards/cost-attribution.jsonnet
+++ b/grafana-dashboards/cost-attribution.jsonnet
@@ -1,0 +1,104 @@
+#!/usr/bin/env -S jsonnet -J ../vendor
+local grafonnet = import 'github.com/grafana/grafonnet/gen/grafonnet-v10.4.0/main.libsonnet';
+local dashboard = grafonnet.dashboard;
+local ts = grafonnet.panel.timeSeries;
+
+local totalDailyCosts =
+  ts.standardOptions.withMin(20)
+  + ts.options.withTooltip({ mode: 'multi' })
+  + ts.options.withLegend({
+    "calcs": [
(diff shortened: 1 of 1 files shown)


GeorgianaElena:
I give up!

The last panel that's supposed to present the costs per component and per hub doesn't work unfortunately.

And this is because I cannot find a way to specify the a variable of type `Query` and of subtype `Infinity` so i can pass it on the URL and all the other fields.

I couldn't get it to work with a custom variable either. API docs of vars is at https://grafana.github.io/grafonnet/API/dashboard/variable.html.

I will merge this PR as it is, so we can get the automation of deploying it ready as well and I will open a spike about figuring out is an `infinity` value is not available and what to do if it isn't and we can tackle this separately.

Btw, locally, I deployed this with the command below, ran from the jupyterhub/grafana-dashboards.
```
/deploy.py --dashboards-dir /Users/georgiana/dev/2i2c/pilot-hubs/grafana-dashboards --folder-name testing --folder-uid aws-ce-grafana-backend https://grafana.openscapes.2i2c.cloud
```

GeorgianaElena:
cc @consideRatio, if you want to fiddle with this more. I will open a separate spike for you to assign if you wish.

GeorgianaElena:
I opened https://github.com/2i2c-org/infrastructure/issues/4944

PR #4964, opened 2024-10-09
Add access for data and compute members separately

jbusecke:
I would like to ensure that our team members will have access to the hub. Thanks


config/clusters/leap/daskhub-common.values.yaml
```diff
@@ -91,6 +91,7 @@ basehub:
           allowed_organizations:
             - leap-stc:leap-pangeo-base-access
             - leap-stc:leap-pangeo-full-access
+            - leap-stc:data-and-compute
```
jbusecke, line 94: Unsure if this is actually needed?

consideRatio: This is needed, the later parts are only filtering whats presented - but this is the part that makes them able to login at all.

PR #4972, opened 2024-10-16
Update references to renamed jupyterhub-home-nfs chart

### LINKED ISSUE #4889: Rename jupyter-home-nfs into jupyterhub-home-nfs
In https://github.com/sunu/jupyter-home-nfs/issues/2 we agreed to rename the project from `jupyter-home-nfs` to `jupyterhub-home-nfs`. This task is to make that happen

## Definition of done

- [x] Repo is renamed
- [x] Helm chart is renamed
- [x] 2i2c-org/infrastructure repo is updated to use renamed helm chart
- [x] Any python code and documentation referencing old name is renamed

sgibson91:
- fixes #4889

### DIFF
--- a/helm-charts/basehub/Chart.yaml
+++ b/helm-charts/basehub/Chart.yaml
@@ -25,7 +25,7 @@ dependencies:
     version: "2024.1.0"
     repository: "https://helm.dask.org/"
     condition: dask-gateway.enabled
-  - name: jupyter-home-nfs
+  - name: jupyterhub-home-nfs
     version: 0.0.5
-    repository: oci://ghcr.io/sunu/jupyter-home-nfs
-    condition: jupyter-home-nfs.enabled
+    repository: oci://ghcr.io/2i2c-org/jupyterhub-home-nfs
+    condition: jupyterhub-home-nfs.enabled
--- a/helm-charts/basehub/Chart.yaml
+++ b/helm-charts/basehub/Chart.yaml
@@ -26,6 +26,6 @@ dependencies:
     repository: "https://helm.dask.org/"
     condition: dask-gateway.enabled
   - name: jupyterhub-home-nfs
-    version: 0.0.5
+    version: 0.0.6
     repository: oci://ghcr.io/2i2c-org/jupyterhub-home-nfs
     condition: jupyterhub-home-nfs.enabled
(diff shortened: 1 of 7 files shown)


sgibson91:
Hey @sunu could you help me understand what's going wrong with the GHCR package here please?

sunu:
@sgibson91 this is failing because the published packages in https://github.com/orgs/2i2c-org/packages?repo_name=jupyterhub-home-nfs are all private. Can we make them public?

sgibson91:
> Can we make them public?

Yes, please do!

sgibson91:
Is there a way to make the workflow in that repo push them publicly by default?

sgibson91:
Ah ok, I'm on it

sunu:
> Is there a way to make the workflow in that repo push them publicly by default?
Once the package visibility is set to public, any new package pushed by the workflow will be public as well. But I don't know how to make the package public by default from the beginning of the top of my head

sgibson91:
I changed the setting at the org level

sunu:
Cool, I changed all the packages to be public.
image

sgibson91:
GHA seems to be struggling to re-run the failed jobs, so let's leave that for a bit.

sgibson91:
> Once the package visibility is set to public, any new package pushed by the workflow will be public as well.

This is fine. It's a one-time action per package, and I don't think we'll be adding more soon :)

sgibson91:
Hmmmm, still can't pull the image :/

helm-charts/basehub/Chart.yaml
```diff
@@ -25,7 +25,7 @@ dependencies:
     version: "2024.1.0"
     repository: "https://helm.dask.org/"
     condition: dask-gateway.enabled
-  - name: jupyter-home-nfs
+  - name: jupyterhub-home-nfs
     version: 0.0.5
```
sunu, line 29: The `0.0.5` version isn't ported over to the new package under 2i2c Github org. I just pushed `0.0.6` tag to get a new version to use. Let's use that here.
```suggestion
    version: 0.0.6
```

sunu:
@sgibson91 I made a fix by creating a new release of the helm chart and using that here. The previous versions were only accessible under the old name.

sgibson91:
Thanks for the assist @sunu!

Issue #4691, opened 2024-08-27
Enable dynamic imagebuilding in NASA VEDA staging hub


yuvipanda:
### Context

Once https://github.com/2i2c-org/infrastructure/issues/4635 is completed, enable dynamic image building on the staging hub of NASA-VEDA

Works towards https://github.com/NASA-IMPACT/veda-jupyterhub/issues/45 objective.

### Tasklist

- [x] Enable dynamic image building feature on VEDA staging hub following relevant steps from https://infrastructure.2i2c.org/howto/features/imagebuilding/

### Definition of Done

- [x] Dynamic image building option is available and working
- [x] Previous profile options available continue to work as expected

yuvipanda:
https://github.com/2i2c-org/infrastructure/pull/4984 temporarily reverts this, so we could get https://github.com/2i2c-org/infrastructure/pull/4975 through. After that, we can put this back up, and then enable dynamic image building in the prod hub too!

PR #4975, opened 2024-10-16
nasa-veda: Enable GPU instance for select users

wildintellect:
How do I only push this to staging? should I be editing https://github.com/2i2c-org/infrastructure/blob/main/config/clusters/nasa-veda/staging.values.yaml instead?

@sgibson @batpad @sunu I think I could use a little more guidance than https://infrastructure.2i2c.org/howto/features/profile-list-restrict/

### DIFF
--- a/config/clusters/nasa-veda/common.values.yaml
+++ b/config/clusters/nasa-veda/common.values.yaml
@@ -266,7 +266,7 @@ basehub:
               choices:
                 pytorch:
                   display_name: Pangeo PyTorch ML Notebook
-                  default: true
+                  default: false
                   slug: "pytorch"
                   kubespawner_override:
                     image: "quay.io/pangeo/pytorch-notebook:2024.08.18"
--- a/config/clusters/nasa-veda/common.values.yaml
+++ b/config/clusters/nasa-veda/common.values.yaml
@@ -269,7 +269,13 @@ basehub:
                   default: true
                   slug: "pytorch"
                   kubespawner_override:
-                    image: "quay.io/pangeo/pytorch-notebook:2023.09.19"
+                    image: "quay.io/pangeo/pytorch-notebook:2024.08.18"
+                tensorflow2:
+                  display_name: Pangeo Tensorflow2 ML Notebook
+                  default: true
+                  slug: "tensorflow2"
+                  kubespawner_override:
+                    image: "quay.io/pangeo/ml-notebook:2024.08.18"
           kubespawner_override:
             environment:
               NVIDIA_DRIVER_CAPABILITIES: compute,utility
--- a/config/clusters/nasa-veda/common.values.yaml
+++ b/config/clusters/nasa-veda/common.values.yaml
@@ -248,6 +248,37 @@ basehub:
                     cpu_limit: 15.695
                     node_selector:
                       node.kubernetes.io/instance-type: r5.4xlarge
+        - display_name: NVIDIA Tesla T4, ~16 GB, ~4 CPUs
+          description: "Start a container on a dedicated node with a GPU"
+          slug: "gpu"
+          allowed_groups:
+            - veda-analytics-access:maap-biomass-team
+          profile_options:
+            image:
+              display_name: Image
+              unlisted_choice:
+                enabled: true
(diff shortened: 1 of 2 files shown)


sunu:
@wildintellect yes, we should edit https://github.com/2i2c-org/infrastructure/blob/main/config/clusters/nasa-veda/staging.values.yaml instead to make these changes only affect staging. Changes are deployed automatically once a PR is merged.

batpad:
This looks good to me. You will also need to add this instance type to the list of available nodepool groups here: https://github.com/2i2c-org/infrastructure/blob/main/eksctl/nasa-veda.jsonnet#L40

config/clusters/nasa-veda/common.values.yaml
```diff
@@ -248,6 +248,43 @@ basehub:
...
+              choices:
+                pytorch:
+                  display_name: Pangeo PyTorch ML Notebook
+                  default: true
+                  slug: "pytorch"
+                  kubespawner_override:
+                    image: "quay.io/pangeo/pytorch-notebook:2024.08.18"
+                tensorflow2:
+                  display_name: Pangeo Tensorflow2 ML Notebook
+                  default: true
```
yuvipanda, line 275: Only one of these can have `default: true`

wildintellect:
@yuvipanda @sgibson91 please review closely. I copy/pasted most of this, unclear if it's right.

sgibson91:
@wildintellect I have some local commits to push but I don't think I have write access to this branch?

Basically, I have updated the jsonnet file to match the convention we are using there. I think deployed this is staging but it's not currently working. You are using the jupyter-fancy-profile interface which only allows one `profileList` entry, then everything should be listed under `choices`. I've made that change locally and deployed it but it's not working. You are the first hub using jupyter-fancy-profiles and trying to add GPUs which is why you didn't have anything appropriate to copy from.

@yuvipanda I think we'll need some cycles here to figure out why my changes didn't work.

config/clusters/nasa-veda/common.values.yaml
```diff
@@ -248,6 +248,43 @@ basehub:
                     cpu_limit: 15.695
                     node_selector:
                       node.kubernetes.io/instance-type: r5.4xlarge
+        - display_name: NVIDIA Tesla T4, ~16 GB, ~4 CPUs
+          description: "Start a container on a dedicated node with a GPU"
+          slug: "gpu"
+          allowed_groups:
+            - veda-analytics-access:maap-biomass-team
```
yuvipanda, line 255:
```suggestion
            - veda-analytics-access:maap-biomass-team
            -  2i2c-org:hub-access-for-2i2c-staff
```

yuvipanda:
@sgibson91 I went through a bunch of other options, but turns out the issue actually is https://github.com/2i2c-org/infrastructure/pull/4975/files#r1807170311 :D So simple yet hadn't checked that first.

jupyterhub-fancy-profiles does also support multiple profiles now, thanks to work from folks at devseed!

image

yuvipanda:
And I see that @sgibson91 has already done the eksctl stuff needed, and this new profile does work!

image

yuvipanda:
profileList being overriden in the staging file (removed in https://github.com/2i2c-org/infrastructure/pull/4984) may have been the other reason it didn't work for you. I've re-opened https://github.com/2i2c-org/infrastructure/issues/4691 so we track enabling it there again.

yuvipanda:
Since I can't seem to push to this PR, I've merged this as is and opened https://github.com/2i2c-org/infrastructure/pull/4986 to merge my change in.

sgibson91:
> @sgibson91 I went through a bunch of other options, but turns out the issue actually is https://github.com/2i2c-org/infrastructure/pull/4975/files#r1807170311 :D So simple yet hadn't checked that first.

I actually _did_ have that change locally and it wasn't working for me! Clearly the work of computer gremlins.

yuvipanda:
@sgibson91 ah, do you think https://github.com/2i2c-org/infrastructure/pull/4984 was the cause then?

sgibson91:
@yuvipanda Maybe? Hard to say without a spike

PR #4977, opened 2024-10-17
jupyter-health: use CHCS auth provider for staging

minrk:
The JupyterHub CHCS client is a Public Application, meaning it has only a client id and requires PKCE, no client secret.

Not ready for prod since we only have test accounts on CHCS so far.

### DIFF
--- a/config/clusters/jupyter-health/staging.values.yaml
+++ b/config/clusters/jupyter-health/staging.values.yaml
@@ -9,5 +9,28 @@ jupyterhub:
         secretName: https-auto-tls
   hub:
     config:
-      GitHubOAuthenticator:
+      JupyterHub:
+        authenticator_class: generic-oauth
+      GenericOAuthenticator:
+        client_id: Ima7rx8D6eko0PzlU1jK28WBUT2ZweZj7mqVG2wm
         oauth_callback_url: https://staging.jupyter-health.2i2c.cloud/hub/oauth_callback
+        authorize_url: https://chcs.fly.dev/o/authorize/
+        token_url: https://chcs.fly.dev/o/token/
+        usredata_url: https://chcs.fly.dev/api/v1/users/profile
+        username_claim: email
+        login_service: CHCS
+        scope:
+          - openid
+        enable_auth_state: true
+    extraConfig:
+      # add access tokens via auth state
+      auth_state_env.py: |
+        def auth_state_env(spawner, auth_state):
+            if not auth_state:
+                spawner.log.warning(f"Missing auth state for user {spawner.user.name}")
+                return
+            spawner.environment["CHCS_TOKEN"] = auth_state["access_token"]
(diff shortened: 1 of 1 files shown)


consideRatio:
Manually deploying this, I see the following when pressing login:

image

Is there a way to provide access to 2i2c members to ensure we are able to help debug and support this - ideally by using either github / google identities? This deployment will stand out, because we otherwise have access to hubs as 2i2c employees using github / google identities.

minrk:
Since this is an oauth provider, we would need to create accounts for you on the oauth provider. Either that, or enable both with [multiauthenticator](https://github.com/idiap/multiauthenticator), which I'm happy to try as well. That's not in the default image either, but [it is proposed](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/3459). I'm happy to test it here, and potentially merge https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/3459

minrk:
Makes sense! I'll suggest trying out multiauthenticator, which should be a generic solution for 2i2c, since you could add GitHub login for any deployment where a cluster wants to use credentials where you don't have accounts.

consideRatio:
I want to avoid putting pressure on z2jh to include multiauthenticator right away to unblock this, so lets reduce that option to having a custom image for jupyter-health where its installed for now.

With that said, I see the following constructive short term choices remain:

1. Create users for 2i2c employees within this oauth provider
2. Create a single user for 2i2c employees to impersonate within this oauth provider
3. Use multiauthenticator with either github or google alongside the jupyter-health specific oauth provider

I've looped in @Gman0909, @aprilmj, and @yuvipanda about this for now in slack. 2i2c internal slack reference https://2i2c.slack.com/archives/C07SJJWVCAD/p1729169838322779

config/clusters/jupyter-health/staging.values.yaml
```diff
@@ -9,5 +9,28 @@ jupyterhub:
         secretName: https-auto-tls
   hub:
     config:
-      GitHubOAuthenticator:
+      JupyterHub:
+        authenticator_class: generic-oauth
```
yuvipanda, line 13:
```suggestion
        # Uses CHCS auth provider
        # Note: 2i2c engineers can not log in via this, so they can not provide support that
        # requires logging into this hub. But since Jupyter Health team members have access to this
        # repo, this is acceptable
        authenticator_class: generic-oauth
```

yuvipanda: My suggestion is we make this comment, and merge this as is, and accept that 2i2c engineers can't provide support that requires logging into the hub.

yuvipanda: @minrk if this looks good to you, I suggest you accept the suggestion and merge it!

Issue #4970, opened 2024-10-16
[8 hours of time] [Deadline 23th Oct] Erik doing good community citizen work


consideRatio:
This is the third issue of its kind, where the previous was #4884.

## Definition of done

- I put in 8h of work in the jupyterhub ecosystem purely nagivated by my JupyterHub team member hat on
- I provide an overview of work that was done in this issue before closing

Issue #4939, opened 2024-10-04
Including jupyterhub.services.binder.url by default in basehub leads to a warning every minute for non-binder deployments


consideRatio:
I figure we may want to not provide this config by default to avoid this. Just dumping this here for now.

```
kubectl logs -n staging hub-74b7569bcb-8td2w -c hub | grep binderhub 

[I 2024-10-04 07:37:39.514 JupyterHub app:3224] Adding external service binder at http://binderhub:8090
[E 2024-10-04 07:37:40.531 JupyterHub app:3246] Cannot connect to external service binder at http://binderhub:8090. Is it running?
[W 2024-10-04 07:38:41.631 JupyterHub app:2508] Cannot connect to external service binder at http://binderhub:8090
[W 2024-10-04 07:39:41.542 JupyterHub app:2508] Cannot connect to external service binder at http://binderhub:8090
[W 2024-10-04 07:40:41.450 JupyterHub app:2508] Cannot connect to external service binder at http://binderhub:8090
[W 2024-10-04 07:41:41.450 JupyterHub app:2508] Cannot connect to external service binder at http://binderhub:8090
[W 2024-10-04 07:42:41.465 JupyterHub app:2508] Cannot connect to external service binder at http://binderhub:8090
[W 2024-10-04 07:43:41.469 JupyterHub app:2508] Cannot connect to external service binder at http://binderhub:8090
[W 2024-10-04 07:44:41.497 JupyterHub app:2508] Cannot connect to external service binder at http://binderhub:8090
[W 2024-10-04 07:45:41.456 JupyterHub app:2508] Cannot connect to external service binder at http://binderhub:8090
[W 2024-10-04 07:46:41.516 JupyterHub app:2508] Cannot connect to external service binder at http://binderhub:8090
[W 2024-10-04 07:47:41.566 JupyterHub app:2508] Cannot connect to external service binder at http://binderhub:8090
[W 2024-10-04 07:48:41.473 JupyterHub app:2508] Cannot connect to external service binder at http://binderhub:8090
[W 2024-10-04 07:49:41.554 JupyterHub app:2508] Cannot connect to external service binder at http://binderhub:8090
[W 2024-10-04 07:50:41.631 JupyterHub app:2508] Cannot connect to external service binder at http://binderhub:8090
[W 2024-10-04 07:51:41.609 JupyterHub app:2508] Cannot connect to external service binder at http://binderhub:8090
[...]

PR #4979, opened 2024-10-18
Testing upgrade from jupyterhub 3.3.7 to 4.0.0-beta.4

consideRatio:
I'm using 2i2c infra (2i2c cluster's staging hub) during open source work to check that z2jh 4.0.0-beta.4 is mature enough to be released as 4.0.0. Part of #4970.

The changes made here reflect the kind of changes we may want to roll out systematically at some point. This PR isn't meant to be merged as we don't have a system to install different helm chart versions for different hubs. Instead, I'm just manually deploying this to 2i2c cluster's staging hub to and doing manual checks that for example users home folders remain accessible and servers aren't getting restarted during upgrade.

## Issues observed

- https://github.com/jupyterhub/jupyterhub/issues/4929
  - Mitigated for 2i2c in https://github.com/2i2c-org/infrastructure/pull/4981, but we also need to handle https://github.com/2i2c-org/infrastructure/issues/4939 unless its fixed
  - Fixed by https://github.com/jupyterhub/jupyterhub/pull/4930, released with jupyterhub 5.2.1
  - Included in z2jh via https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/3551, released with chart and image version `4.0.0-beta.4.git.6815.hcfafef2a`
- (not important) https://github.com/jupyterhub/jupyterhub/issues/4928
- Downgrading after upgrade couldn't be done without first replacing the jupyterhub.sqlite file with a backup.
  This can be manually handled by `kubectl debug <hub pod name> -it --copy-to=test-debug --image=ubuntu -c hub -- /bin/bash` and then looking at `/srv/jupyterhub for sqlite backup files and using a backup instead of the one named `jupyterhub.sqlite`

## What seem to work

Besides the issues, it seems that:

- kubespawner v7
  - nfs storage access was retained as intented by updating `subPath: "{username}"` to `subPath: "{escaped_username}"`
  - user servers running while the upgrade was made with JupyterHub 3, 4, and 5 in their images were accessible after jupyterhub was successfully restarted
- oauthenticator v17:
  - cilogon with google as an idp seem to work fine

### DIFF
--- a/config/clusters/2i2c-aws-us/itcoocean.values.yaml
+++ b/config/clusters/2i2c-aws-us/itcoocean.values.yaml
@@ -73,7 +73,7 @@ jupyterhub:
         volumeMounts:
           - name: home
             mountPath: /home/jovyan
-            subPath: "{username}"
+            subPath: "{escaped_username}"
           # Mounted without readonly attribute here,
           # so we can chown it appropriately
           - name: home
--- a/config/clusters/2i2c/imagebuilding-demo.values.yaml
+++ b/config/clusters/2i2c/imagebuilding-demo.values.yaml
@@ -117,7 +117,7 @@ jupyterhub:
     allowNamedServers: true
     image:
       name: quay.io/2i2c/dynamic-image-building-experiment
-      tag: "0.0.1-0.dev.git.10263.hc87b65cf"
+      tag: "0.0.1-0.dev.git.10722.h6ed14882"
     config:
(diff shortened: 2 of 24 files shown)


consideRatio:
Testing complete, no known exceptions to handle besides adjusting `subPath` config from `"{username}"` to `"{escaped_username}"` which I think will be mentioned in z2jh upgrade docs

PR #4983, opened 2024-10-18
Update 01_new-issue.yaml

Gman0909:
Updated our standard issue template with a standard DoD. 

It is understood that this DoD will not apply to every issue, but it should serve as a reminder of what we should be aiming for moving forward.

### DIFF
--- a/.github/ISSUE_TEMPLATE/01_new-issue.yaml
+++ b/.github/ISSUE_TEMPLATE/01_new-issue.yaml
@@ -18,7 +18,6 @@ body:
       description: |
         (optional) A clear and concise list of steps to take to close this issue.
       value: |
-        The Definition of Done. What must be true to consider this task complete?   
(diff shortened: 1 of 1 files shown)


yuvipanda:
This makes it the third 'definition of done' within this issue, the other two being:

https://github.com/2i2c-org/infrastructure/pull/4983/files#diff-01a30cc3f1ea54ff02c02c9a924b121346f0b52740c2b3a3a2bcaed9817962a9R37
https://github.com/2i2c-org/infrastructure/pull/4983/files#diff-01a30cc3f1ea54ff02c02c9a924b121346f0b52740c2b3a3a2bcaed9817962a9R43

Let's have only one?

Gman0909:
Edited. I opted to remove the predefined DoD in favour of an editable task list, which should be easier to convert into sub tasks.

PR #4988, opened 2024-10-19
nasa-veda: Enable dynamic image building

yuvipanda:
This enables dynamic image building for both staging and prod.

Fixes https://github.com/2i2c-org/infrastructure/issues/4691


yuvipanda:
Since this isn't part of the current 2i2c sprint, we don't need to work on this right now.

Issue #4844, opened 2024-09-19
[Support] Reduce awi-ciroh Filestore to 5 TB


jmunroe:
### The Freshdesk ticket link

https://2i2c.freshdesk.com/a/tickets/2170

### Ticket request type

Other

### Ticket impact

🟨 Medium

### Short ticket description

The filestore size for awi-ciroh has decreased from 9TB to 4.7TB. To avoid charges for unused storage, the community wants to update the filestore capacity to 5TB. 

### (Optional) Investigation results

Since this require creating a new filestore and transferring data over plus coordinating with the community, I think it should be schedule in the next engineering sprint.

sgibson91:
I had to request a quota increase to accommodate the size of the second filestore, so let's see how long that takes to come back.

sgibson91:
Quota increase approved

sgibson91:
Because AWI-CIROH are running in their own GCP project, I do not have the right access permissions to SSH into the VM I created to mount the 2 filestores and copy the data over.

> Insufficient IAM permissions. The instance belongs to an external organization. You must be granted the roles/compute.osLoginExternalUser IAM role on the external organization to configure POSIX account information.

sgibson91:
I worked around the above restriction by modifying our `deployer exec root-homes` command to mount the two filestores. The data copying has begun!

sgibson91:
I am having problems with my workaround modifying the `deployer exec root-homes` command to mount both filestores. The pod seems to be getting killed long before the transfer has completed.

sgibson91:
I set resource limits on the pod as well as a node selector and toleration to force the pod onto the largest node AWI-CIROH has available. It seems that the file transfer is moving much quicker now.

jmunroe:
Hi @sgibson91 . 

What is the current usage of data that we are trying to transfer?

I don't think you should be alone in this one. There is an interaction with AWI-CIROH that might need to happen -- if they are already using more than 3TB especially!

Is rsync the correct technology to use here? This sounds like an example of where Globus endpoints are the designed for.  (Globus is on my mind while I am at USRSE)

Let's make sure this issue is discussed in our iteration planning meeting tomorrow. It definitely sounds like it is much more complicated than originally expected.

sgibson91:
I have no idea what Globus is or how to use it, whereas the engineering team have extensively documented rsync in these situations.

I'm using a glob pattern to transfer. I'm not doing anything complicated or elegant to ignore files.

[...]

PR #4989, opened 2024-10-21
awi-ciroh: delete 3TB filestore instance that wasn't migrated to

consideRatio:
In https://github.com/2i2c-org/infrastructure/issues/4844 we introduced a 3TB filestore instance to migrate to, but we never succeeded and we are now to hold off doing something according to latest message in https://2i2c.freshdesk.com/a/tickets/2170. So, I figure we need to cleanup this now.


consideRatio:
@sgibson91 do you think this is the right call atm?

PR #4990, opened 2024-10-21
eksctl updates for VEDA cluster

sgibson91:
These were updates I applied to address #4975 but was unable to push to that branch before it was merged


yuvipanda:
Thank you, @sgibson91!

PR #5018, opened 2024-10-30
jupyter-health: set gh-scoped-creds env

minrk:
cc @fperez


fperez:
Excellent, thanks a lot @minrk!

Issue #4878, opened 2024-09-24
AWS cost attribution (3/4): deploy to all other AWS clusters


consideRatio:
## Definition of done

Cost attribution system is deployed, or offered to be deployed in case we are blocked by community cloud permissions, to all AWS clusters according to the four steps in https://infrastructure.2i2c.org/howto/cost-attribution/aws/

**2i2c AWS organization member accounts**
  - [x] 2i2c-aws-us
  - [x] catalystproject-africa
  - [x] gridsst
  - [x] jupyter-health
  - [x] kitware
  - [x] nasa-cryo
  - [x] opensci
  - [ ] smithsonian
  - [ ] strudel
  - [ ] ubc-eoas
  - [x] victor

**Standalone AWS accounts**
  - [x] nmfs-openscapes
  - [x] openscapes

**Community AWS organization's member accounts**
  - [x] earthscope
    - [x] Community contacted about enabling various cost allocation tags
    https://2i2c.freshdesk.com/a/tickets/2387
  - [x] jupyter-meets-the-earth
    - [x] Community contacted about enabling various cost allocation tags
      https://2i2c.freshdesk.com/a/tickets/2386
  - [x] nasa-ghg
    - [x] Community contacted about enabling various cost allocation tags
    (same as nasa-ghg) https://2i2c.freshdesk.com/a/tickets/2389
  - [x] nasa-veda
    - [x] Community contacted about enabling various cost allocation tags
    (same as nasa-veda) https://2i2c.freshdesk.com/a/tickets/2389
   - [x] projectpythia
     - [x] Community contacted about enabling various cost allocation tags
    (same as nasa-veda) https://2i2c.freshdesk.com/a/tickets/2408

Issue #4928, opened 2024-10-03
AWS cost attribution (feedback 4): Clarify situation on what is to be rolled out for all AWS clusters


consideRatio:
#4872 sais that we should roll out a system to all clusters, but this is a bit ambigious and doing it in the full sense is a compromise on cloud costs and startup times.

## Context

The grafana based AWS cost attribution system can capture most 2i2c attributed costs, and it can without trouble categorize the kinds of costs (compute, home storage, ...). This functionality can easily be rolled out to all AWS clusters.

However, the cost attribution _to specific hubs_ can only be done thoroughly with a compromise on cloud costs, startup times, and some complexity added to our cloud infra. Since its a tradeoff and not just a pure improvement, a decision on how we roll this out to communities should be made.

## Explanation to additional cloud costs and startup times

I'll explain this by analogy, where stones are user servers, buckets are nodes incurring cost when used, and the act of getting new buckets is to startup a node - which leads to startup time.

Let's say we have some amount of generic stones and are to fit them in some amount of generic buckets, we will _at worst need to leave up to one bucket unused_. However, if we need to treat stones differently based on Z amounts of different colors, and put them in buckets with matching color, we will instead _at worst have to leave up to Z buckets unused_.

This means that on average we'll use cloud resources less efficiently on average like this, and with the user of more buckets, we expect more addition/removal events to happen as well on average, which is what incurrs startup times.

## Overview of complexities to consider

- _Hub specific compute cost attribution_
  Introducing this means to hub specific colored buckets, which is a tradeoff in cloud costs, startup time, and some cloud infra complexity.
- _Hub specific home storage cost attribution_
  - Introducing this when we use AWS EFS is just incurring some cloud infra complexity, but it can also be a benefit from a maintenance perspective as backups can be restored for a subset of users instead of all users etc.
  - Introducing this when we use `jupyterhub-home-nfs` for storage, or if we are using GCP using filestore, then we'll have a tradeoff again.
  This is because with `jupyterhub-home-nfs` and GCP filestore, we pay for a fixed amount of storage capacity which we increase when needed, while when we use AWS EFS, we just pay for what we use no matter what.
  In practice, we'll end up with the colored buckets situation again if we transition to not use AWS EFS.

## Definition of done

- We have come to a decision on _when_ to roll out _what kind of cost attribution setup_ to _which communities_

consideRatio:
[...]

PR #5022, opened 2024-11-01
[victor] Upgrade k8s version, separate nodepools per hub, enable cost-allocation

GeorgianaElena:
Ref https://github.com/2i2c-org/infrastructure/issues/4878

### DIFF
--- a/config/clusters/victor/prod.values.yaml
+++ b/config/clusters/victor/prod.values.yaml
@@ -14,6 +14,8 @@ basehub:
         CILogonOAuthenticator:
           oauth_callback_url: https://hub.victorproject.org/hub/oauth_callback
     singleuser:
+      nodeSelector:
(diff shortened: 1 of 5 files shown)


consideRatio (approved):
Great @GeorgianaElena!

It isn't obvious if 2i2c ought to split apart node pools and home directory storage infra to become hub specific. I made a call with some discussion in https://github.com/2i2c-org/infrastructure/issues/4928 that we should not do so without explicit opt-in.

The documentation you followed was written before the decision was made about this, so its written uncertain about this and hasn't been updated to reflect something about it yet.

I figure its not a problem for victor to have a split node pool for staging/prod, but for the clusters going onwards I figure we should go with the minimalistic act of using shared node pools and shared EFS storage unless they already have a separation there.

PR #5034, opened 2024-11-05
Remove Sepehr and add Dan to the admin

benlee0423:
Sepehr leaves our organization and Dan joined us.


consideRatio:
@GeorgianaElena there is a very non-obvious behavior related to this that requires a manual intervention as well via JupyterHub UI.

Removing the user from the admin_users list is required to avoid re-adding the user as an admin, but the user also needs to be manually removed after this - the reason is that not being part of the admin_users list doesn't revoke admin permissions.

PR #5050, opened 2024-11-07
Revert terraform version bump

yuvipanda:
Partially reverts https://github.com/2i2c-org/infrastructure/pull/4745, as Terraform 1.9 is not an open source product.

### DIFF
--- a/terraform/aws/main.tf
+++ b/terraform/aws/main.tf
@@ -1,5 +1,5 @@
 terraform {
-  required_version = "~> 1.9"
+  required_version = "~> 1.5"
 
   required_providers {
     aws = {
--- a/terraform/azure/main.tf
+++ b/terraform/azure/main.tf
@@ -1,6 +1,6 @@
 
 terraform {
-  required_version = "~> 1.9"
+  required_version = "~> 1.5"
   required_providers {
     azurerm = {
       # FIXME: upgrade to v4, see https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/guides/4.0-upgrade-guide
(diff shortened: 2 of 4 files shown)


yuvipanda:
We will need to verify that this doesn't have any negative effects on the workspaces we currently maintain in the two months it's been in effect before merging it.

consideRatio:
I think this is just changing the lower bound, but that the upper bound would remain unchanged

terraform/aws/main.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = "~> 1.9"
+  required_version = "~> 1.5"
```
consideRatio, line 2:
```suggestion
  required_version = "~> 1.5.7"
```

This would change the lower bound from 1.x to 1.5.x

terraform/azure/main.tf
```diff
@@ -1,6 +1,6 @@
 
 terraform {
-  required_version = "~> 1.9"
+  required_version = "~> 1.5"
```
consideRatio, line 3:
```suggestion
  required_version = "~> 1.5.7"
```

terraform/gcp/main.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = "~> 1.9"
+  required_version = "~> 1.5"
```
consideRatio, line 2:
```suggestion
  required_version = "~> 1.5.7"
```

terraform/uptime-checks/main.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = "~> 1.9"
+  required_version = "~> 1.5"
```
consideRatio, line 2:
```suggestion
  required_version = "~> 1.5.7"
```

yuvipanda:
While this doesn't fix everything, without this change it's impossible for me to run terraform locally. So I'm going to merge this.

PR #5106, opened 2024-11-14
2i2c-aws-us: Nodegroup split and k8s update

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/5090


GeorgianaElena (commented):
@sgibson91, just fyi that this morning I came to the conclusion that we only need GPUs for the hubs that will use them and don't have to provision them for all hubs on a cluster.

Same for dask nodegroups.

PR #5108, opened 2024-11-14
fix: correct team names, and allowed orgs

wildintellect:
- Org:teams need to be in allowed orgs before gpu access is allowed
- Incorrect team was in the gpu section, the gpu team isn't listed in allowed orgs because a user needs to belong to one of the other allowed teams. Does this work @yuvipanda ?


consideRatio:
I think this should be fine and your understanding is correct; the typical issue here is allowing access to a profile choice but not allowing a login in the first place

Issue #5074, opened 2024-11-13
AWS cost attribution: split each hub in all of AWS clusters into their own user nodegroups


GeorgianaElena:
For each cluster in the list below, perform the following actions:

```[tasklist]
- [x] Create the same set of nodes for each hub of the cluster
- [x] Configure each of the hubs so their users only spawn on the nodepools designated for them
- [x] Tag each of the nodepool with `2i2c:hub-name` to match the name of the hub
```

  - [x] 2i2c-aws-us
  - [x] catalystproject-africa
  - [x] gridsst
  - [x] jupyter-health
  - [x] kitware
  - [x] nasa-cryo
  - [x] opensci
  - [x] smithsonian
  - [x] strudel
  - [x] ubc-eoas
  - [x] victor
  - [x] nmfs-openscapes
  - [x] openscapes
  - [x] earthscope
  - [x] jupyter-meets-the-earth
  - [x] nasa-ghg
  - [x] nasa-veda
  - [x] projectpythia (they are using other cost allocation system https://github.com/2i2c-org/meta/issues/1612)

### Reference 
See https://github.com/2i2c-org/infrastructure/issues/4482 for reference as at the end, [their config](https://github.com/2i2c-org/infrastructure/blob/287dd831038e100deecdda44e1eea7818478bf1c/eksctl/openscapes.jsonnet#L27-L81) should be similar to one of openscapes.

## Definition of done

- [ ] Users across the hubs of all clusters, will spawn on to their own nodepools, and not share them with users on other hubs.
- [ ] There is a `tag` `2i2c:hub-name`  and `2i2c:node-purpose` on all the *nodes* spawned when users log on to the hub. You can verify this by looking at EC2 instances on the AWS console.
- [ ] the cost of each hub shows up in its respective grafana

sgibson91:
@GeorgianaElena do you want to split up the list of clusters between us?

GeorgianaElena:
@sgibson91, I'll start open separate issues  for each cluster so that we can split them more easily: self-assign -> do the work -> close the issue. And to get that nice feeling of fixing an issue rather than have a single big one.

I will start with the nmfs-openscapes one so you can self-assign and start working on it if you want 🚀

GeorgianaElena:
@sgibson91, let's use this comment to record with what clusters we ran into issues so we can come back to them later on. So far we had issues with:

- nmfs-openscapes -> both of us had issues running eksctl commands against this 
[...]

PR #5110, opened 2024-11-14
catalyst-af: create hub-specific nodegroups

### LINKED ISSUE #5091: AWS cost attribution: put each hub in catalystproject-africa into its own user nodegroup

sgibson91:
- fixes #5091 
- sub-task of #5074


GeorgianaElena (commented):
@sgibson91, I believe this PR misses to update the hub's configs to have a node selector and actually put the hubs onto these nodegroups and https://github.com/2i2c-org/infrastructure/pull/5104#pullrequestreview-2438244630 too I think

PR #5111, opened 2024-11-14
aws docs: add back now conditional step on adding cluster permissions

consideRatio:
I now understand this better, and it turns out its still needed in k8s
clusters created for accounts where we aren't using 2i2c's AWS SSO as
then we don't access AWS with a single role but as separate users. This
is because the role or user creating the cluster gets access, where we
use the same role when working with 2i2c's AWS SSO credentials but
different users when working against other AWS accounts.

This reverts commit 0d4aa98ddeaa7907d79d5389df0a46366d878f8f and updates things further to make it conditionally used instead of always used.

### DIFF
--- a/docs/hub-deployment-guide/new-cluster/new-cluster.md
+++ b/docs/hub-deployment-guide/new-cluster/new-cluster.md
@@ -583,7 +583,7 @@ Commit this file to the repo.
 ````{tab-item} AWS
 :sync: aws-key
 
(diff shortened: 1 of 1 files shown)


consideRatio:
@GeorgianaElena and @sgibson91 I've now added access in nmfs-openscapes with this misconception corrected so you should be unblocked in nmfs-openscapes.

Issue #5061, opened 2024-11-13
Enable automated backups of EBS volumes on VEDA staging hub


sgibson91:
### Context

We will be moving forward on automated backups for AWS EBS volumes using the [Data Lifecycle Manager](https://docs.aws.amazon.com/ebs/latest/userguide/snapshot-lifecycle.html) option. 

- Automatically create daily backups
- Retention period of 5 days
- Deploy and test on VEDA staging hub first

### Task list

```[tasklist]
- [x] Create an IAM role for DLM service in terraform
- [x] Apply the target tag to existing EBS volumes for which automated backup has to be enabled.
- [x] Set up a DLM lifecycle policy, that defines the time and interval of backup creation, retention period and a target tag for selection of EBS volumes for backing up, in terraform
- [x] Set `Backup = true` on the existing EBS volume for staging
- [x] Wait a day and see if a backup has been created
- [x] Investigate how to restore from backups
- [x] Document feature and how to restore from backup
```

### Definition of Done

- [ ] The feature/service is technically complete
- [ ] Deployed to staging hub
- [ ] The feature/service been tested with one or more users (if applicable)
- [ ] The feature/service is well documented, and the documentation is accessible for the target user base

sgibson91:
#5189 applied successfully!

The policy will execute a backup snapshot at 23:45 UTC every day, so next items are:

- [ ] Check back tomorrow to see if a backup snapshot of the staging EBS volume has been created
- [ ] Check back in >5 days time to ensure we are only retaining 5 backup snapshots for the EBS volume

sgibson91:
AWS's docs on restoring EBS volume from a snapshot https://docs.aws.amazon.com/prescriptive-guidance/latest/backup-recovery/restore.html#restore-snapshot

sgibson91:
Gotcha: double check terraform state hasn't changed state due to the backup, see for example the GCP backup docs https://infrastructure.2i2c.org/howto/filesystem-backups/restore-filesystem/#gcp

sgibson91:
Woohoo! Another piece completed!

yuvipanda:
Great job, @sgibson91 and @sunu! 

Unfortunately I don't think this is quite complete yet. I've left a comment in https://github.com/2i2c-org/infrastructure/pull/5286/files#r1887447565.

sgibson91:
Did we close this in the end? @yuvipanda @sunu

yuvipanda:
[...]

PR #5189, opened 2024-11-26
Enable automatic backup of EBS volumes using DLM

sunu:
Addresses https://github.com/2i2c-org/infrastructure/issues/5061


sgibson91:
This applied successfully!

This policy will execute a backup snapshot at 23:45 UTC every day, so next items are:

- [ ] Check back tomorrow to see if a backup snapshot of the staging EBS volume has been created
- [ ] Check back in >5 days time to ensure we are only retaining 5 backup snapshots for the EBS volume

PR #5217, opened 2024-12-03
use oauthenticator 17.2 for pkce-experiment (jupyter-health)

minrk:
enables refreshing auth tokens, used in jupyter-health

### DIFF
--- a/config/clusters/jupyter-health/common.values.yaml
+++ b/config/clusters/jupyter-health/common.values.yaml
@@ -43,7 +43,7 @@ jupyterhub:
     #
     image:
       name: quay.io/2i2c/pkce-experiment
-      tag: 0.0.1-0.dev.git.10892.h37c70b2e
+      tag: 0.0.1-0.dev.git.11154.h8a6a8049
     allowNamedServers: true
     config:
       JupyterHub:
--- a/helm-charts/images/hub/pkce-requirements.txt
+++ b/helm-charts/images/hub/pkce-requirements.txt
@@ -1,8 +1,8 @@
 # Image lives at quay.io/2i2c/pkce-experiment
-# install oauthenticator 17.1,
-# which adds PKCE support.
-# experiment no longer needed when base chart is updated to z2jh 4.0.0
-oauthenticator>=17.1,<18
+# install oauthenticator 17.2,
+# which adds PKCE support (17.1) and refresh tokens (17.2).
+# experiment no longer needed when base chart is updated to z2jh 4.1.0
+oauthenticator>=17.2,<18
 
 
 # jupyterhub-configurator isn't maintained and its not intended to be developed


GeorgianaElena (commented):
Thanks @minrk, I believe this needs building and then updating the image tag in the hub config here ⬇️  but I'm not sure if you have push rights on 2i2c quay https://github.com/2i2c-org/infrastructure/blob/de31b5e617d6810ae3af6f92ad2a4e6bb40777f2/config/clusters/jupyter-health/common.values.yaml#L39-L46

I will try to do it

minrk:
@GeorgianaElena thanks! Looks like it deployed successfully and things are working (I see in the Hub logs that my token was expired, but rightly refreshed using a refresh token when I launched my server), but the health check failed. I don't see any information about the health check or its failure, though.

I suspect it's due to refresh_auth not succeeding for the fake `deployment-service-check` user and the fact that `refresh_pre_spawn` is True. I think OAuthenticator may need some handling of these fake users to bypass refresh_auth. Disabling `refresh_pre_spawn` altogether is one workaround.

GeorgianaElena:
@minrk, I've disabled `refresh_pre_spawn` and re-ran the test. It still fails.
With `refresh_pre_spawn` True, it failed to start the server, now it fails to start the kernel [here](https://github.com/Quansight/jhub-client/blob/860dff7365bdccc5208724390529b00b8198ccef/jhub_client/execute.py#L87)

And the actual error is:

```
    async def ensure_kernel(self, kernel_spec=None):
        kernel_specs = await self.list_kernel_specs()
        if kernel_spec is None:
>           kernel_spec = kernel_specs["default"]
E           KeyError: 'default'
```

Looking at the oauthenticator PR now.

minrk:
Okay, I think that makes sense, actually.

If we want to be able to use OAuthenticator with users that aren't actually users, I think we need an explicit bypass of the refresh_user machinery somewhere.

It is a generic issue, I suppose, but perhaps the fix belongs in OAuthenticator for now to somehow bypass either for specific users or some property of users, e.g. having no auth_state at all.

GeorgianaElena:
Agree! I was leaning towards the same solution. I think I will revert this PR for now, if that's ok and we can test it again once we have the oauthenticator bypass implemented.

minrk:
@GeorgianaElena actually, can I point the health check to another user rather than the fake one? Then the health check should work and we can keep testing.

minrk:
Ah, I guess I can't really do that because the test [hardcodes the username](https://github.com/2i2c-org/infrastructure/blob/cdf26509840357e04245c6523c79efd0b8b98878/deployer/health_check_tests/test_hub_health.py#L22) and makes quite a few assumptions about how users are handled by the Authenticator and that invalid users won't have any problems, which isn't necessarily the case, as seen here.

PR #5220, opened 2024-12-03
adjust opensci binder limits

jmunroe:


sgibson91:
@jmunroe lgtm and you should just hit merge :)

PR #5226, opened 2024-12-04
jupyter-health: test oauthenticator 17.2 and skip refresh for the test user

GeorgianaElena:
@minrk, how do you feel about experimenting with the changes in https://github.com/jupyterhub/oauthenticator/pull/780 on the jupyter-health staging?

I figured we could do this until that PR is merged and a new version is released with those changes.

I manually deployed this and confirmed that the health check tests work now.


minrk:
@GeorgianaElena that sounds great to me, thanks!

PR #5229, opened 2024-12-04
jupyter-health: deploy Exchange auth to prod

minrk:
promotes staging config to common

staging is working well after #5225, promote the same config to prod

only difference between staging and prod now is the callback url


GeorgianaElena (approved):
This looks good to me and I don't see any issues in merging it 🚀

minrk:
Thank you!

Issue #5062, opened 2024-11-13
Enable disk size alerting for AWS EBS volumes on VEDA staging hub


sgibson91:
### Context

 - Default JupyterHub dashboards live in https://github.com/jupyterhub/grafana-dashboards
 - Start in the infra repo, then have a discussion about upstreaming

### Task list

```[tasklist]
- [x] Use [Prometheus metric exporter](https://github.com/2i2c-org/infrastructure/blob/main/helm-charts/basehub/templates/home-space-reporter.yaml) and mount the whole NFS disk as a volume
- [x] Verify the graphs that report disk size continue to make sense when referring to EBS volumes, and both staging and prod volumes are captured in a way that is distinguishable from on another
- [x] Investigate and enable Grafana alerting based on some threshold, alerts sent to support[at]2i2c.org (alerting was manually enabled)
- [x] Deploy to VEDA staging hub
- [x] Enable alerting via Prometheus AlertManager to PagerDuty
- [x] Test by setting the threshold to something really low and see if an alert is triggered
- [x] Document the feature, the notifications, and how to resize the EBS volume
```

### Definition of Done

- [x] The feature/service is technically complete
- [ ] The feature/service been tested with one or more users (if applicable)
- [x] Deployed to VEDA staging hub
- [x] Feature has been tested to fire an alert
- [ ] The feature/service is well documented, and the documentation is accessible for the target user base

GeorgianaElena:
@sgibson91, raising here for visibility so you're not caught by surprise in case you haven't seen https://github.com/2i2c-org/infrastructure/pull/4923 (for more context https://2i2c.slack.com/archives/C055A1J1DRP/p1727942694562319)

sgibson91:
Thank you @GeorgianaElena I did see it and did remember about it, but hadn't done the work digging it out yet! Thank you!

@sunu, https://github.com/2i2c-org/infrastructure/pull/4923 will need to be reverted for Grafana alerting :)

sgibson91:
@sunu found it difficult to programatically enable grafana alerting via grafonnet. I have proposed that we add the enablement of alerts and a one-time manual step in our hub deployment guide for now to close this issue out. I will open an issue to track a spike investigating grafonnet further in the new year, maybe with more 2i2c folks helping him.

sgibson91:
Grafana needs access to an SMTP server to send emails

yuvipanda:
[...]

PR #5234, opened 2024-12-05
Add pagerduty and prometheus integration info to support chart secrets

sgibson91:
This information could be used by prometheus alertmanager to send automatic alerts

Set up following these instructions: https://www.pagerduty.com/docs/guides/prometheus-integration-guide/

related to #5062


yuvipanda:
I think this needs a possible addition to the json schema as well - trying to run `deploy-support nasa-veda` fails with:

```
Error: UPGRADE FAILED: values don't meet the specifications of the schema(s) in the following chart(s):
support:
- (root): Additional property pagerduty-prometheus-integration is not allowed
```

sgibson91:
@yuvipanda it's getting changed in another PR where Tarashish is actually turning on alert manager

PR #5244, opened 2024-12-06
jupyter-health: reapply refresh_user fix

minrk:
added in #5226 but got lost in #5229 because I started working on #5229 before #5226 was merged so it didn't get the updated changes


GeorgianaElena:
I missed it too, sorry

PR #5279, opened 2024-12-12
Refactoring deployer exec root-homes command

sgibson91:
- Addresses an action point in https://github.com/2i2c-org/meta/issues/1671
- Splits up the `kubectl run` command into a `create` and `exec` couplet. This means that the bash process we start will not be PID 1, therefore we can safely exit the pod without causing it to restart (which was happening before). This is because PID 1 is special, and killing it is the same as sending a shutdown command which was causing the pod to restart.
- Adds an `--persist` option to NOT cleanup the pod on completion if required

### DIFF
--- a/deployer/commands/exec/infra_components.py
+++ b/deployer/commands/exec/infra_components.py
@@ -40,12 +40,11 @@ def root_homes(
     with open(config_file_path) as f:
         cluster = Cluster(yaml.load(f), config_file_path.parent)
 
-    with cluster.auth():
-        hubs = cluster.hubs
-        hub = next((hub for hub in hubs if hub.spec["name"] == hub_name), None)
-        if not hub:
-            print_colour("Hub does not exist in {cluster_name} cluster}")
-            return
+    hubs = cluster.hubs
+    hub = next((hub for hub in hubs if hub.spec["name"] == hub_name), None)
+    if not hub:
+        print_colour("Hub does not exist in {cluster_name} cluster}")
+        return
 
     server_ip = base_share_name = ""
     for values_file in hub.spec["helm_chart_values_files"]:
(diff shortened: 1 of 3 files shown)


deployer/commands/exec/infra_components.py
```diff
@@ -112,27 +121,33 @@ def root_homes(
...
-    with cluster.auth():
-        subprocess.check_call(cmd)
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".json") as tmpf:
+        # Dump the pod spec to a temporary file
+        json.dump(pod, tmpf)
+        tmpf.flush()
+
+        with cluster.auth():
+            try:
+                # Ask kube-controller to create a pod
```
yuvipanda, line 144:
```suggestion
                # Ask api-server to create a pod
```

yuvipanda: Can you also add a comment here on *why* we're turning this into two steps than 1?

deployer/commands/exec/infra_components.py
```diff
@@ -112,27 +121,33 @@ def root_homes(
...
-        subprocess.check_call(cmd)
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".json") as tmpf:
+        # Dump the pod spec to a temporary file
+        json.dump(pod, tmpf)
+        tmpf.flush()
+
+        with cluster.auth():
+            try:
+                # Ask api-server to create a pod
+                subprocess.check_call(["kubectl", "create", "-f", tmpf.name])
```
yuvipanda, line 145: Can we add a comment here about why we're turning this into a two part action?

sgibson91: Done in https://github.com/2i2c-org/infrastructure/pull/5279/commits/86464bd805adf74cbca07c79a82a373a058e57d6

GeorgianaElena (approved):
Thanks for updating and documenting this @sgibson91!

PR #5284, opened 2024-12-13
Centralise docs on transferring data

sgibson91:
- fixes https://github.com/2i2c-org/meta/issues/1671

deploy preview: https://2i2c-pilot-hubs--5284.org.readthedocs.build/howto/filesystem-management/data-transfer/

### DIFF
--- a/docs/howto/filesystem-management/data-transfer.md
+++ b/docs/howto/filesystem-management/data-transfer.md
@@ -92,6 +92,7 @@ At this point, it is useful to have a few terminal windows open:
    ```
 
    If no resources are found, you can proceed to the next step.
+   If there are resources, you may wish to wait until these servers have stopped, or coordinate a maintenance window with the community when disruption and potential data loss should be expected.
 
 2. **Make the hub unavailable by deleting the `proxy-public` service.**
 
--- a/docs/howto/filesystem-management/data-transfer.md
+++ b/docs/howto/filesystem-management/data-transfer.md
@@ -94,16 +94,33 @@ At this point, it is useful to have a few terminal windows open:
    If no resources are found, you can proceed to the next step.
    If there are resources, you may wish to wait until these servers have stopped, or coordinate a maintenance window with the community when disruption and potential data loss should be expected.
 
-2. **Make the hub unavailable by deleting the `proxy-public` service.**
+1. **Make the hub unavailable by deleting the `proxy-public` service.**
 
    ```bash
    kubectl --namespace $HUB_NAME delete svc proxy-public
    ```
 
-3. **Re-run the `rsync` command in the data transfer pod.**
+1. **Re-run the `rsync` command in the data transfer pod.**
    This process should take much less time now that the initial copy has completed.
 
-4. **Delete the `PersistentVolume` and all dependent objects.**
+1. **Check the Reclaim Policy of the `Persistent Volume`.**
+
+   We should first verify the [reclaim policy](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reclaiming) of the persistent volume to ensure we will not lose any data.
+
+   The reclaim policy can be checked by running:
+
+   ```bash
+   kubectl get pv ${HUB_NAME}-home-nfs
+   ```
+
+   If the reclaim policy is `Retain`, we are safe to delete the pv without data loss.
+   Otherwise, you may need to patch the reclaim policy to change it to `Retain` with:
+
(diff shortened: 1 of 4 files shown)


docs/howto/filesystem-management/data-transfer.md
```diff
@@ -0,0 +1,138 @@
...
+
+   Once you have detached from `screen`, can now `exit` the pod and let the copy run.
+
+(migrate-data:reattach-pod)=
+## Reattaching to the data transfer pod
+
+You can regain access to the pod created for the data transfer using:
+
+```bash
+kubectl --namespace $HUB_NAME attach -i ${CLUSTER_NAME}-root-home-shell
```
yuvipanda, line 74: Would `attach` here cause problems when the user `exit`s, if it's pid 1?

docs/howto/filesystem-management/data-transfer.md
```diff
@@ -0,0 +1,138 @@
+(migrate-data)=
+# Transfer data across filestores
+
+This documentation covers how to transfer data between filestores.
```
yuvipanda, line 4: Let's note this works across cloud providers.

docs/howto/filesystem-management/data-transfer.md
```diff
@@ -0,0 +1,138 @@
...
+- A terminal that is attached to the data transfer pod to re-run the file transfer (see [](migrate-data:reattach-pod))
+
+1. **Check there are no active users on the hub.**
+   You can check this by running:
+
+   ```bash
+   kubectl --namespace $HUB_NAME get pods -l "component=singleuser-server"
+   ```
+
+   If no resources are found, you can proceed to the next step.
```
yuvipanda, line 94: What to do if resources are found?

docs/howto/filesystem-management/data-transfer.md
sgibson91: Resolved by https://github.com/2i2c-org/infrastructure/pull/5284/commits/3161f51f5da584463119efa9176accce5604ee89

docs/howto/filesystem-management/data-transfer.md
sgibson91: Resolved by https://github.com/2i2c-org/infrastructure/pull/5284/commits/edbe39ba098cc75f8743a2791d20cb7deef07cf7

docs/howto/filesystem-management/data-transfer.md
```diff
@@ -0,0 +1,138 @@
...
+1. **Make the hub unavailable by deleting the `proxy-public` service.**
+
+   ```bash
+   kubectl --namespace $HUB_NAME delete svc proxy-public
+   ```
+
+1. **Re-run the `rsync` command in the data transfer pod.**
+   This process should take much less time now that the initial copy has completed.
+
+1. **Delete the `PersistentVolume` and all dependent objects.**
```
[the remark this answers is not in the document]
sgibson91: Resolved by https://github.com/2i2c-org/infrastructure/pull/5284/commits/27a5fcfb2ca558a413e6c13257155189b6ac488f

docs/howto/filesystem-management/data-transfer.md
sgibson91: Good question that I'm not sure of the answer to.

I think `attach` would cause problems if it's PID 1, but we've taken steps to avoid getting assigned PID 1.

`attach` has the benefit of giving us the same process back. Does `exec` do the same if we do _not_ provide the `-t` flag? E.g. only `exec -i foo bar`

yuvipanda: I *think* `attach` will always give you the process the pod was opened with, which I think will be pid 1. This is how it gives us the same process back each time - by always giving us the same process that was used to start the pod. 

The easiest way to verify is to run `echo $$` ([some other similar vars](https://ryanstutorials.net/bash-scripting-tutorial/bash-variables.php#others)) (this prints the current pid) after `attach`, and see if  you get `1`, and repeat the process for `exec`. This should help you make the determination for wether we should use attach or exec here.

exec will *always* start a new process, regardless of whether `-i` is passed or not. So `screen` is required when using exec to get back to old terminal

sgibson91: fixed in https://github.com/2i2c-org/infrastructure/pull/5284/commits/e892e82fb5f55dc0c77feed51a36a06624d7035d

PR #5286, opened 2024-12-16
Document nfs home directories restore process

sunu:
Related to https://github.com/2i2c-org/infrastructure/issues/5061

### DIFF
--- a/docs/howto/filesystem-management/filesystem-backups/restore-filesystem.md
+++ b/docs/howto/filesystem-management/filesystem-backups/restore-filesystem.md
@@ -7,6 +7,8 @@ covers those steps for the cloud providers.
 (howto:filesystem-backup:restore:aws)=
 ## AWS
 
+### Restoring home directories from an EFS recovery point
+
 ```{note}
 We follow AWS's guidance for [restoring EFS from a recovery point](https://docs.aws.amazon.com/aws-backup/latest/devguide/restoring-efs.html#efs-restore-console)
 ```
--- a/docs/howto/filesystem-management/filesystem-backups/restore-filesystem.md
+++ b/docs/howto/filesystem-management/filesystem-backups/restore-filesystem.md
@@ -13,6 +13,34 @@ We follow AWS's guidance for [restoring EFS from a recovery point](https://docs.
 
 Due to the length of the steps listed in that document, we will not repeat them here.
 
+### Restoring home directories from EBS snapshots
+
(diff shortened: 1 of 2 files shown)


sgibson91:
Can we move this content into this part of the docs instead? https://infrastructure.2i2c.org/howto/filesystem-management/filesystem-backups/restore-filesystem/#aws

sunu:
@sgibson91 done!

docs/howto/filesystem-management/filesystem-backups/restore-filesystem.md
```diff
@@ -7,12 +7,42 @@ covers those steps for the cloud providers.
...
+1. Go to the EBS volumes page in the AWS console
+2. Find the volume ID of the existing EBS volume containing the NFS home directories
+3. Click on the volume ID and find the instance ID in the "Attached resources" section
+4. Once we have the instance ID, we can mount the new EBS volume to the EC2 instance by following the steps outlined in the [Attaching an Amazon EBS volume to an instance](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-attaching-volume.html) guide.
+
+Once we have mounted the new EBS volume to the EC2 instance, we can copy the contents from the restored EBS volume to the NFS home directories volume. This can be done by running the following commands on the EC2 instance:
+
+```bash
+# Copy the contents from the restored EBS volume to the NFS home directories volume
+rsync -av --info=progress2 </restored-volume/path-to-home-directories/> </existing-volume/path-to-home-directories/>
```
yuvipanda, line 39: I think this won't actually work for us, since we exclusively use EKS. It *does* create EC2 instances, but that's an implementation detail and we don't necessarily have packages in the base instance. I think we have to either:

1. Create a new pod, new PVC and PV that refers the recovered volumeId, and mount the existing volume via NFS into that pod.
2. Create a separate new EC2 instance (temporarily), and mount the recovered and existing home directory there to do a copy. However, we'll have to mount the existing home directory via NFS, which may be difficult to do outside the k8s cluster.

I think (1) is easier.

In addition, I think we shouldn't copy directly into the existing home directories, as that may delete new files that were created since the snapshot (default rsync behavior). So either:

1. We are recovering *everything* (disaster recovery scenario), and we don't have access to the existing volume at all. So here we have to do a full copy
2. We are recovering something partial, and we should copy into a subfolder and copy that into the users' home directory. We should do this sparingly.

PR #5289, opened 2024-12-16
Refactor CI/CD so support and staging jobs are separate, and multiple staging hubs are detected

### LINKED ISSUE #5186: Refactor the deployer and CI/CD pipeline to permit deployment of multiple staging hubs per cluster
### Context

When we designed the current CI/CD system for deploying hubs as it currently stands, we did so with the principle that there'd only be one staging hub per cluster. Shared clusters would share the staging hub as it was supposed to be a canary deployment. Hence the deployer _by design_ only looks for one staging hub per cluster.

However, we've been terrible at enforcing that principle, as evidenced by the existence of the ucmerced-staging hub. And that means that these hubs don't get automatically updated by the CI/CD pipeline. So we should probably do a refactor of the deployer and CI/CD pipeline to permit multiple staging hub deployments to get automatically updated.

### Task list

```[tasklist]
 - [ ] Decouple the support and staging upgrades into separate jobs in the CI/CD pipeline
 - [ ] Refactor the deployer to search for multiple staging hubs per cluster
 - [ ] Refactor the staging job in the CI/CD pipeline to accept multi-hub deployments per cluster, like the prod job
```

### Definition of Done

- [ ] If multiple staging hubs exist on a given cluster, then all of them are automatically updated by our CI/CD pipeline

sgibson91:
- fixes #5186


tests/test-clusters/cluster3/cluster.yaml
```diff
@@ -0,0 +1,15 @@
+name: cluster1
```
sgibson91, line 1:
```suggestion
name: cluster3
```

Issue #5209, opened 2024-11-29
Upgrade pathway to Z2JH 4


GeorgianaElena:
Z2JH 4.0.0 is out ↪ [changelog](https://z2jh.jupyter.org/en/latest/changelog.html) and we should upgrade to using it.

GeorgianaElena:
I intend to create sub-issues for each of them and further refine them and have this issue track the entire effort.

GeorgianaElena:
Everything that needed to be done is now done 🎉 The only remaining part is to actually merge this and see it working :D

PR #5292, opened 2024-12-17
Upgrade to z2jh 4

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/5209


helm-charts/basehub/values.yaml
```diff
@@ -1434,21 +1436,6 @@ jupyterhub:
...
-              if not auth_state or "teams" not in auth_state:
-                  print(f"User {spawner.user.name} does not have any auth_state set, profile_list filtering not available")
-
-              else:
-                # casefold teams to match what GitHub's API does when doing authorization calls
-                groups |= set([f'{team["organization"]["login"]}:{team["slug"]}'.casefold() for team in auth_state["teams"]])
-
-            print(f"User {spawner.user.name} is part of groups {' '.join(groups)}")
-
             # Filter out profiles with allowed_groups set if the user isn't part of the group
```
GeorgianaElena, line 1439: This, toghether with `GitHubOAuthenticator.auth_state_groups_key` = "teams" and `GitHubOAuthenticator.manage_groups = true` doesn't work unfortunately.

The logs say:
```bash
The auth_state_groups_key teams does not exist in the auth_model. Available keys are: dict_keys(['access_token', 'refresh_token', 'id_token', 'scope', 'token_response', 'github_user', 'teams'])
```
Which is weird.

After printing the auth_state, I also  tried to run https://github.com/jupyterhub/oauthenticator/blob/aadfbf5ff4f25c972bfa03ae65a24057c94f645c/oauthenticator/oauth2.py#L1240-L1248 locally and ran into:

 `TypeError: descriptor 'get' for 'dict' objects doesn't apply to a 'list' object`

Which looks like it's because auth_state[teams] is a list of dicts.

I'm now exploring with the `auth_state_groups_key` as a callable.

cc @yuvipanda

GeorgianaElena:
I tested this on staging and it worked. Then I tried testing it with profile options and GitHubOAuthenticator and ran into some issues.

helm-charts/basehub/values.yaml
yuvipanda: I would say for this to pick whatever is the least amount of work that unblocks z24jh 4.0 upgrade!

PR #5293, opened 2024-12-17
Use GITHUB_OUTPUT instead of GITHUB_ENV

sgibson91:
GITHUB_OUTPUT is defined per job, as opposed to GITHUB_ENV that is global. By making this switch, we can be more flexible with reusing variable names

- related to #5186


sgibson91:
I've confirmed this behaviour with a test workflow: https://github.com/sgibson91/testing-gh-actions/actions/workflows/multijob-output-file.yaml

PR #5294, opened 2024-12-18
[deployer exec root-homes]: wait for the pod to be ready before exec-ing into it

GeorgianaElena:
Without the command is not working as the pod is not ready immediately


sgibson91:
@sunu and I just found this this morning and he was going to fix it in another open PR 😆 

Tarashish, is this matching what you were planning on implementing?

GeorgianaElena:
Sure! Also, I now just realized the importance of opening issues when a bug is found rather than going ahead and try to fix it. (I meant to open one but thought it will be a quick fix 🙄 )

PR #5296, opened 2024-12-19
Collect GPU usage metrics with prometheus

yuvipanda:
We use [prometheus node exporter](https://github.com/prometheus/node_exporter), deployed as part of our prometheus chart, to collect metrics about CPU and memory usage.

This deploys NVIDIA's [dcgm-exporter](https://github.com/NVIDIA/dcgm-exporter) which collects information about GPU usage.

As we work towards more cost monitoring and usage monitoring, collecting this information should allow us to help users get more bang for the buck from their GPU use. Since we only collect information after the exporters are deployed, this starts the information collection process even if it's not directly visible to end users.

Works towards https://2i2c.productboard.com/entity-detail/features/30046512, initially requested as part of https://2i2c.freshdesk.com/a/tickets/2545.

### DIFF
--- a/helm-charts/support/Chart.yaml
+++ b/helm-charts/support/Chart.yaml
@@ -62,5 +62,6 @@ dependencies:
   # Provide metrics about GPU usage
   # https://github.com/NVIDIA/dcgm-exporter
   - name: dcgm-exporter
-    version: 3.6.1
+    version: 4.8.1
     repository: https://nvidia.github.io/dcgm-exporter/helm-charts
+    condition: dcgm-exporter.enabled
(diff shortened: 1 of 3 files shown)


yuvipanda:
Unfortunately this doesn't work on GCP yet:

```
  Warning  FailedCreate  9m (x18 over 19m)  daemonset-controller  Error creating: insufficient quota to match these scopes: [{PriorityClass In [system-node-critical system-cluster-critical]}]
```

yuvipanda:
We can set `.priorityClassName` to get around this on GCP. But we don't have a clean way to schedule this *only* on GPU nodes yet, as it will just crash and burn on non-GPU nodes.

yuvipanda:
I tried again and failed. Let's close this out for now, and rely on GPU metrics in cloudwatch / GCP metrics

Issue #4857, opened 2024-09-23
Provision an EBS volume for the VEDA prod hub


sgibson91:
An EBS volume needs to be pre-provisioned so that it can be setup as in-cluster NFS storage for the VEDA prod hub.

### Tasklist

Use the docs at the following URL to complete these tasks: https://infrastructure.2i2c.org/howto/features/storage-quota/#creating-a-pre-provisioned-disk

- [x] Update the terraform config for the nasa-veda cluster to create a new EBS volume
- [x] Provisionally enable the jupyter-nfs-home chart for the VEDA prod hub
	- Provide the config with the volume ID, but don't change the IP address yet. The deployer will need this config for `exec root-homes` to work for the migration step.

### Definition of Done

- [x] An EBS volume for the VEDA prod hub exists
- [x] The deployer can connect to the EBS volume ready for the migration

sgibson91:
Current size of the prod hub's file usage is 3.2TB. Though there is https://github.com/2i2c-org/infrastructure/issues/5021 to reduce this.

sgibson91:
- [ ] Check EBS can be resized _down_

sgibson91:
We didn't get a response from Alex regarding 1) whether the clen-up would be happening first or not, or 2) what initial size the disk should be given, given that it cannot be scaled down. Hence, this task is blocked.

sunu:
This is done! An EBS disk exists for VEDA prod hub home directories and ` jupyter-nfs-home` is configured to use that EBS disk.

PR #5376, opened 2025-01-13
Provision EBS disk for nasa veda prod home directories

sunu:
Related to https://github.com/2i2c-org/infrastructure/issues/4857.

Creates a 3 TB EBS disk for `jupyterhub-home-nfs` to store the VEDA prod hub home directories.

### DIFF
--- a/terraform/aws/projects/nasa-veda.tfvars
+++ b/terraform/aws/projects/nasa-veda.tfvars
@@ -218,6 +218,12 @@ ebs_volumes = {
     type        = "gp3"
     name_suffix = "staging"
     tags        = { "2i2c:hub-name" : "staging" }
+  },
+  "prod" = {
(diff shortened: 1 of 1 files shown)


sunu:
@sgibson91 The current EFS instance is 1.5 TB. I've set the EBS size to 3 TB to ensure we have adequate headroom and won't need to switch disks soon. Does this size sound reasonable, or should we go for a smaller volume?

sunu:
Ah, right, we don't actually have to switch disks when we need to increase the size

sunu:
I've deployed the changes through Terraform and created the volume. Merging this PR now. In a follow-up PR (https://github.com/2i2c-org/infrastructure/pull/5378), we will set up `jupyterhub-home-nfs` to use the newly created volume.

PR #5385, opened 2025-01-15
Fix `tmpauthenticator` for JupyterHub 5

yuvipanda:
While `tmpauthenticator` has been fixed to work with JupyterHub 5, no release was made and it's not in z2jh.

I've created a branch of `tmpauthenticator` with only one commit (https://github.com/jupyterhub/tmpauthenticator/compare/main...yuvipanda:tmpauthenticator:v2) that bumps the version. That's what's needed there.

In addition, `allow_all` needs to be explicilty enabled now. I think `tmpauthenticator` should default this to True (as this is the point of `tmpauthenticator`) but don't think we need to work on that right now.

I think the follow-ups here are:

- [ ] Check if we have other hubs using `tmpauthenticator`, and fix their config
- [ ] Merge this PR to make sure our users are doing ok
- [ ] Make a release of `tmpauthenticator` (we can call it 1.1, than 2.0). Note this would be a tactical contribution, rather than a strategic one
- [ ] Use the released version of `tmpauthenticator` in our image
- [ ] Open an issue (or PR) in z2jh to use newer version of `tmpauthenticator`

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -1054,7 +1054,7 @@ jupyterhub:
         oauth_no_confirm: true
     image:
       name: quay.io/2i2c/pilot-hub
-      tag: "0.0.1-0.dev.git.11322.hb977bf05"
+      tag: "0.0.1-0.dev.git.11481.heb8debe10"
     networkPolicy:
       enabled: true
       # interNamespaceAccessLabels=accept makes the hub pod's associated
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -1054,7 +1054,7 @@ jupyterhub:
         oauth_no_confirm: true
     image:
       name: quay.io/2i2c/pilot-hub
-      tag: "0.0.1-0.dev.git.11481.heb8debe10"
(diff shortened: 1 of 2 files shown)


GeorgianaElena:
Thanks for fixing this issue @yuvipanda!

I've just checked and noticed that just setting `allow_all` is enough to fix the issue for the ephemeral hubs that we have. I don't think that releasing a new tmp authenticator is needed, or at least not one that does not default to `allow_all=True`.

From https://github.com/jupyterhub/tmpauthenticator/compare/v1.0.0...main it looks like nothing relevant in the codebase has changed that should trigger a release _I think_.

I think we should:
- [ ] open a PR that sets `Authenticator.allow_all=True` for all the hubs that use the tmpauthenticator
- [ ] open a PR that makes `allow_all` True by default in tmpauthenticator
- [ ] Make a release of tmpauthenticator with this change. (Not sure if this would be a 1.1 or 2.0?)
- [ ] Use the released version of tmpauthenticator in our image and rm the allow_all config
- [ ] Open an issue (or PR) in z2jh to use newer version of tmpauthenticator

Do you agree with the updated plan?

Issue #5257, opened 2024-12-10
Migrate MAAP hub to use jupyterhub-home-nfs


yuvipanda:
The MAAP jupyterhub doesn't have any users yet, and it would be good to switch to jupyterhub-home-nfs before it gets users. 

## Information to be determined

- Initial size of the disk
- Initial per-user limit

## Definition of done

- [x] MAAP is using jupyterhub-home-nfs
- [ ] Existing EFS is decomissioned

wildintellect:
We'll need to get metrics from @sujen1412 on what the current home directory usage is, excluding mounted folders, in order to estimate a sensible quota.

yuvipanda:
@wildintellect we can also increase it later if needed (but not decrease it). Maybe we can pick something small to start and go from there?

wildintellect:
@yuvipanda we've agreed to start with 100 GB/user (initial guess of 20 users to start), so 2 TB total to start. The actual usage on MAAP is somewhere in the 32TB range but we think that's because users are storing copies of things that are also in buckets.

Can you remind me, how big the /tmp or /scratch on any given instance actually is?

sunu:
I'll need access to the MAAP AWS account to provision the EBS disk through Terraform. I'll talk to @wildintellect and get that set up.

wildintellect:
I raised this issue in VEDA meeting today. @Jjfrench (MAAP) and @slesaad (GHG) should be able to help you out.

wildintellect:
@sunu go ahead with this if you have access now. You can pick any day,  just post when, sooner is better before Devs start trying to work on things.

sunu:
@wildintellect I'm working on this today to get everything in place for the migration. We can make the switch tomorrow morning IST if that works.

wildintellect:
@sunu sounds good, go ahead.

sunu:
@wildintellect

sunu:
I have decommissioned the EFS instances. I think it's safe to close the issue now.

PR #5434, opened 2025-01-28
MAAP: Create EBS volumes to store home directories

sunu:
Initial step for the migration of MAAP staging and prod hubs from EFS to jupyterhub-home-nfs. The EBS volumes will be used by the in-cluster NFS server to store the home directories.

related to https://github.com/2i2c-org/infrastructure/issues/5257


sunu:
I've applied these changes and created the required EBS volumes.

PR #5487, opened 2025-02-04
Set resource constraints on jupyterhub-home-nfs containers

### LINKED ISSUE #5285: Set resource requests and limits on jupyterhub-nfs-home nfs-server pod
### Context

We should apply resource requests and limits (memory and cpu) to the NFS server pod running in jupyterhub-home-nfs enabled hubs.

Current deployments are:

- [ ] nmfs-openscapes: staging + prod
- [ ] nasa-cryo: staging + prod
- [ ] nasa-veda: staging

### Task list

```[tasklist]
- [ ] Apply some resource requests and limits to nfs server pod running in nasa-cryo
- [ ] Observe behaviour of nfs-server pod for some time (5 working days?) and make tweaks to requests and limits as needed
- [ ] When suitable requests and limits have been ascertained, roll these out to all hubs currently running jupyterhub-home-nfs
- [ ] Add ability to set resource contraints into jupyterhub-home-nfs chart
```

### Definition of Done

- [ ] The NFS server pod in all jupyterhub-home-nfs enabled hubs has some resource requests and limits for memory and cpu applied
- [ ] These requests and limits have been observed not to immediately break functionality
- [ ] Documentation has been updated to include setting the resource requests and limits

sgibson91:
- fixes #5285 

I have deployed this to CryoCloud staging and will observe for a couple of days before rolling out fully

### DIFF
--- a/helm-charts/basehub/Chart.yaml
+++ b/helm-charts/basehub/Chart.yaml
@@ -26,6 +26,6 @@ dependencies:
     repository: "https://helm.dask.org/"
     condition: dask-gateway.enabled
   - name: jupyterhub-home-nfs
-    version: 0.0.8
+    version: 0.0.9
     repository: oci://ghcr.io/2i2c-org/jupyterhub-home-nfs
     condition: jupyterhub-home-nfs.enabled
--- a/helm-charts/basehub/Chart.yaml
+++ b/helm-charts/basehub/Chart.yaml
@@ -26,6 +26,6 @@ dependencies:
     repository: "https://helm.dask.org/"
     condition: dask-gateway.enabled
   - name: jupyterhub-home-nfs
-    version: 0.0.7
+    version: 0.0.8
     repository: oci://ghcr.io/2i2c-org/jupyterhub-home-nfs
     condition: jupyterhub-home-nfs.enabled
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -1655,3 +1655,27 @@ jupyterhub:
 
 jupyterhub-home-nfs:
   enabled: false
+  nfsServer:
+    resources:
+      requests:
+        cpu: 0.1
+        memory: 30G
+      limits:
+        cpu: 0.2
+        memory: 32G
+  quotaEnforcer:
+    resources:
+      requests:
+        cpu: 0.01
+        memory: 10M
+      limits:
+        cpu: 0.02
+        memory: 11M
+  prometheusExporter:
+    resources:
+      requests:
+        cpu: 0.01
+        memory: 11M
+      limits:
+        cpu: 0.02
+        memory: 13M


sgibson91:
I am actually having trouble deploying this because PersistentVolumes are immutable objects, so I'm having to run kubectl controls manually. CI/CD won't be able to deploy this change automatically, I'll have to intervene manually for every hub with this enabled.

sgibson91:
The nfs-share-creator pod is now stuck creating, but there's nothing useful in the pod events

```
Events:
  Type    Reason     Age    From               Message
  ----    ------     ----   ----               -------
  Normal  Scheduled  8m56s  default-scheduler  Successfully assigned staging/nfs-share-creator-r9ggt to ip-192-168-12-131.us-west-2.compute.internal
```

sgibson91:
I am basically running through the commands listed here to try to resolve this https://infrastructure.2i2c.org/howto/filesystem-management/data-transfer/#switching-the-nfs-servers-over

sgibson91:
I have narrowed down the problem.

There are currently two core nodes up, so jupyterhub-home-nfs is trying to create a pod per node, and then the pod on the second node is failing to come up due to a MultiAttach error from the volume.

```
$ kdp staging-nfs-deployment-fb9869cf6-zbx2d
[...]
Events:
  Type     Reason              Age    From                     Message
  ----     ------              ----   ----                     -------
[...]
  Warning  FailedAttachVolume  2m56s  attachdetach-controller  Multi-Attach error for volume "staging-nfs-home-directories-volume" Volume is already used by pod(s) staging-nfs-deployment-8565f484cb-kkm44
```

sgibson91:
The access mode of the volume is set to `ReadWriteOnce` which explains why I'm seeing the MultiAttach error

https://github.com/2i2c-org/jupyterhub-home-nfs/blob/230c5bd569e1f7f204ccc6c372821174d991c252/helm/jupyterhub-home-nfs/values.yaml#L39-L40

My questions are: Do we actually need a pod per node of the nfs deployment? Is creating multiple pods intentional? Is this a rollout strategy issue? (Struggling to create the new version of the deployment before being able to delete the old one.) If we do need a pod per node, should we change the access mode?

sgibson91:
Replicas is set 1 which makes me think we _shouldn't_ have multiple pods here: https://github.com/2i2c-org/jupyterhub-home-nfs/blob/230c5bd569e1f7f204ccc6c372821174d991c252/helm/jupyterhub-home-nfs/templates/deployment.yaml#L6

sgibson91:
Looking at the below StackOverflow discussion, I think one approach we can take is to configure `maxSurge` and `maxUnavailable` for the `RollingUpdate` strategy. We could configure one more pod above the replica number required, and one pod to be unavailable during an upgrade. Therefore allowing the PVs/PVCs to unbind from one pod and bind to the new one.

https://stackoverflow.com/questions/65700396/how-to-handle-rolling-updates-deployment-strategy-for-rwo-volume-service

sgibson91:
`maxSurge` and `maxUnavailable` where already set to 25%, which I think is probably a default inherited from Kubernetes itself. I'm gonna try setting both to 1 and seeing if that resolves anything.

sgibson91:
Yep - with that edit, the old pod terminated successfully and the new pod came up!

Did a fresh deploy with the deployer and that was also successful. I will make these edits in the upstream chart and make a new release.

sgibson91:
I will merge this tomorrow following no issues with cryo staging

Issue #5455, opened 2025-01-30
Deploy kubernetes cluster with terraform on Jetstream2


jmunroe:
### Context

Using the lessons learned in 
- https://hackmd.io/DzgY3PW4TUOMAmpqvgOdZw?view#How-to-create-and-scale-a-cluster-on-Jetstream2

deploy and kubernetes cluster with terraform. 

There is an Openstack provider Terraform. See
- https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs

Also see [Terraform on Jetstream2 ](https://docs.jetstream-cloud.org/general/terraform/),

### Definition of Done

- [ ] New `openstack` directory in `terraform/`
- [ ] Create a cluster on openstack via Terraform, following similar pattern of `notebook_nodes` and `dask_nodes` with scale to 0 as we do for our other providers
- [ ] Create a `projectpythia-binder` `tfvars` file and setup a production cluster

julianpistorius:
This looks like it:

https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/containerinfra_cluster_v1

GeorgianaElena:
Issues I have learnt about so far:

- concurrent nodegroups creation requests fired by terraform fail
    - https://github.com/2i2c-org/infrastructure/pull/5518#issuecomment-2648324765
    - https://bugs.launchpad.net/magnum/+bug/2097946 might also be relevant
    - current workaround is to run `terraform apply` with `--parallelism=1` flag
- Nodegroup sizes cannot be less than 1 which is a big bummer :( 
    - I opened https://bugs.launchpad.net/magnum/+bug/2098002 about this
- I couldn't yet find a way to disable the creation of the default-worker nodegroup that gets automatically created with the cluster. My impression is that is not possible, but will dig more.

julianpistorius:
> I couldn't yet find a way to disable the creation of the default-worker nodegroup that gets automatically created with the cluster. My impression is that is not possible, but will dig more.

Could you try this approach?

https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1625#issuecomment-1902079012

> If you use Openstack version Wallaby or higher then you could create default-worker nodegroup scaled to 0 nodes and operate in terraform using additional nodegroups.

GeorgianaElena:
[...]

PR #5518, opened 2025-02-07
Add initial openstack magnum terraform config

GeorgianaElena:
This is a work in progress for https://github.com/2i2c-org/infrastructure/issues/5455. Everything is in one file for simplicity, harcoded nodegroups also for simplicity until the create command passes.

- Min count of nodegroups cannot be 0 :(

- Nodegroup creation currently fails with:

```
CREATE_FAILED                                                                                                                                                          |
| status_reason      | Unexpected error while running command.                                                                                                                                |
|                    | Command: helm upgrade test-cluster-jde45zh7dcud openstack-cluster --history-max 10 --install --output json --timeout 5m --values - --namespace                         |
|                    | magnum-390542082bd74fa6abcde82f8c7ded89 --repo https://azimuth-cloud.github.io/capi-helm-charts --version 0.4.0                                                        |
|                    | Exit code: 1                                                                                                                                                           |
|                    | Stdout: ''                                                                                                                                                             |
|                    | Stderr: 'Error: UPGRADE FAILED: release: already exists\n'      
```

Also, when redeploying, the failed nodegroups cannot be deleted and they just hang and the error is the same.

### DIFF
--- a/terraform/openstack/cluster.tf
+++ b/terraform/openstack/cluster.tf
@@ -1,0 +1,112 @@
+variable "prefix" {
+  type        = string
+  description = <<-EOT
+  Prefix for all objects created by terraform.
+
+  Primary identifier to 'group' together resources created by
+  this terraform module. Prevents clashes with other resources
+  in the cloud project / account.
+
+  Should not be changed after first terraform apply - doing so
+  will recreate all resources.
+
+  Should not end with a '-', that is automatically added.
+  EOT
+}
+# https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/containerinfra_clustertemplate_v1#master_flavor-1
+variable "master_flavor" {
+  type        = string
+  default     = "m3.small"
+  description = "Machine type for the master nodes"
+}
+
+variable "core_machine_type" {
+  type        = string
+  default     = "m3.small"
+  description = "Machine type for the core nodes"
+}
+
+variable "nb_machine_type" {
+  type        = string
+  default     = "m3.small"
+  description = "Machine type for the notebook nodes"
+}
+
+# Picked latest ubuntu image from `openstack image list`
+variable "image" {
+  type    = string
+  default = "ubuntu-jammy-kube-v1.31.0-240828-1652"
+}
+
+# https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/containerinfra_clustertemplate_v1
+resource "openstack_containerinfra_clustertemplate_v1" "template" {
+  name                = "${var.prefix}-{var.image}-template"
+  coe                 = "kubernetes"
+  server_type         = "vm"
+  tls_disabled        = false
+  image               = var.image
+  external_network_id = "public"
+  # disable load balancer for master nodes since we're having only one master node
+  master_lb_enabled   = false
+  floating_ip_enabled = false
+  network_driver      = "calico"
+  flavor              = var.master_flavor
+  master_flavor       = var.master_flavor
+  labels = {
+    monitoring_enabled               = "false" # these will handled by the support chart
+    prometheus_monitoring            = "false" # these will handled by the support chart
+    influx_grafana_dashboard_enabled = "false" # these will handled by the support chart
+    auto_scaling_enabled             = "true"
+    min_node_count                   = 1
+  }
+}
+
+output "template" {
+  value = openstack_containerinfra_clustertemplate_v1.template
+}
+
+resource "openstack_containerinfra_cluster_v1" "cluster" {
+  name                = "${var.prefix}-cluster"
+  cluster_template_id = openstack_containerinfra_clustertemplate_v1.template.id
+  master_count        = 1
+  # node_count          = 1
+
+  # lifecycle {
+  #   # An additional safeguard against accidentally deleting the cluster.
+  #   # The databases for the hubs are held in PVCs managed by the cluster,
+  #   # so cluster deletion will cause data loss!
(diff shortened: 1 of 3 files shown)


GeorgianaElena:
@julianpistorius, do you have any ideas about why this error might be showing up when creating the nodepools? 🤔 

The cluster is created successfully, but the nodegroups fail with the error above.

julianpistorius:
Hmm... No idea. I'll have to dig a bit.

julianpistorius:
While I dig, do you mind trying again? We had a networking problem this morning, possibly around the same time you had run into this problem.

julianpistorius:
We are still seeing intermittent problems with networking. Please stand by.

julianpistorius:
Never mind. The networking problems had been solved, so you should be able to try again.

GeorgianaElena:
Thanks for looking into it @julianpistorius! I've retried again today and it fails with an error which **I think** indicates that communication with the management cluster failed?

I see the error both when I try to re-apply the terraform but also when trying to delete the cluster or nodegroups with `openstack coe cluster ...`

```
| stack_id             | test-cluster-b6ppkxavrpgq                                                                                            |
| status_reason        | Unexpected error while running command.                                                                              |
|                      | Command: helm uninstall test-cluster-b6ppkxavrpgq --timeout 5m --namespace magnum-390542082bd74fa6abcde82f8c7ded89   |
|                      | Exit code: 1                                                                                                         |
|                      | Stdout: ''                                                                                                           |
|                      | Stderr: 'Error: Kubernetes cluster unreachable: Get "https://149.165.173.80:6443/version": dial tcp                  |
|                      | 149.165.173.80:6443: i/o timeout\n'                                                                                  |
```

GeorgianaElena:
It looks like if I want to create only one nodepool via `terraform` it work, but having more than one results in this `release: already exists` error.

It might be some kind of race condition somewhere.

GeorgianaElena:
It looks like I can force the nodegroups to fire create requests in sequence with [`depends_on`](https://developer.hashicorp.com/terraform/language/meta-arguments/depends_on).

julianpistorius:
Hi @GeorgianaElena. I'll ask somebody who should know and get back to you.

GeorgianaElena:
Thank you @julianpistorius! 

For now, I've learnt that I can unblock myself by telling terraform to disable creation of resource in parallel with `--parralelism=1`.

julianpistorius:
Great! So according Scott from
StackHPC you have apparently hit a known bug in Magnum:

> I think this is probably the same as the bug which was raised by another one of our clients which Stig has recorded as an upstream bug here: https://bugs.launchpad.net/magnum/+bug/2097946

They'll work on fixing it. In the meantime your workaround will get you by.

GeorgianaElena:
Thank you @julianpistorius! I think I came across the bug report today too https://github.com/2i2c-org/infrastructure/issues/5455#issuecomment-2651346327

Now I'm battling a different issue. The labels I set on the nodegroups through terraform are not propagated to the actual node instances and scheduling fails.

julianpistorius:
Have you been able to set the labels manually using the `openstack coe nodegroup update` command? Because this is something I'm currently also struggling with.

julianpistorius:
@GeorgianaElena 

> The labels I set on the nodegroups through terraform are not propagated to the actual node instances and scheduling fails.

Is the terraform trying to set the labels at the time of creating the nodegroups? Or afterwards on existing nodegroups?

GeorgianaElena:
@julianpistorius, I believe it does it during creation. Based the [code](https://github.com/gophercloud/gophercloud/blob/a5a70f7f0c1e7a89b3ef8da8d6e0339a689a61e3/openstack/containerinfra/v1/nodegroups/requests.go#L104-L113), they do it all in one request, i.e. they pass the labels in the post request creating the nodegroups.

For context, I can see the labels on the actual nodegroups as  `labels_added` with:
```
openstack coe nodegroup show js-cluster core-js
```
Screenshot 2025-02-12 at 11 44 13

But then when I run `k describe node` on a node in that `core-js` nodegroup I don't see these custom labels

julianpistorius:
Interesting. I'm going to see if I can reproduce this using the OpenStack Magnum CLI.

julianpistorius:
This is indeed a bug (with a workaround) as I mentioned in https://github.com/2i2c-org/infrastructure/issues/5455#issuecomment-2654287226

terraform/openstack/cluster.tf
```diff
@@ -1,3 +1,22 @@
...
+#    nodes. This means that we cannot add the node-purpose labels that z2jh
+#    expects and instead we need to rely on the only label that gets added by
+#    the magnum driver and can identify the node, capi.stackhpc.com/node-group.
+#    This in turn, forces us to only offer one machine type per notebook/dask/gpu
+#    for the users.
+#    https://github.com/azimuth-cloud/capi-helm-charts/issues/84
+# 3. The node count and min node count cannot be set to 0 and each nodegroup has
+#    to have at least 1 node.
+#    https://bugs.launchpad.net/magnum/+bug/2098002
+# 4. A default-worker is created apart from the default-controlplane nodegroup
```
yuvipanda, line 17: @jmunroe is dask a requirement for project pythia? if it is, we can leave this be. If not we can disable this by default.

jmunroe: Yes, Dask is used within Project Pythia

GeorgianaElena:
I've deployed the cluser here to the project pythia allocation, so merging this one 🚀

julianpistorius:
That's great news @GeorgianaElena!

PR #5536, opened 2025-02-13
Migrate LEAP staging on to NFS server with storage quotas enforced

### LINKED ISSUE #5257: Migrate MAAP hub to use jupyterhub-home-nfs

sgibson91:
- fixes #5257 5472

### DIFF
--- a/config/clusters/leap/staging.values.yaml
+++ b/config/clusters/leap/staging.values.yaml
@@ -25,3 +25,6 @@ basehub:
     gke:
       enabled: true
       volumeId: projects/leap-pangeo/zones/us-central1-c/disks/hub-nfs-homedirs-staging
+    quotaEnforcer:
+      hardQuota: 0.1  # in GB
+      path: "/export/staging"
--- a/config/clusters/leap/staging.values.yaml
+++ b/config/clusters/leap/staging.values.yaml
@@ -2,6 +2,7 @@ basehub:
   nfs:
     pv:
       serverIP: staging-nfs-service.staging.svc.cluster.local
+      baseShareName: /
   userServiceAccount:
     annotations:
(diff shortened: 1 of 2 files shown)


sgibson91:
```bash
  Warning  FailedMount  1s     kubelet                                MountVolume.SetUp failed for volume "home-base" : mount failed: exit status 1
Mounting command: /home/kubernetes/containerized_mounter/mounter
Mounting arguments: mount -t nfs 10.3.246.86:/homes/ /var/lib/kubelet/pods/59d30fb9-515b-40d3-abec-0958b8ac5d43/volumes/kubernetes.io~nfs/home-base
Output: Mount failed: mount failed: exit status 32
Mounting command: chroot
Mounting arguments: [/home/kubernetes/containerized_mounter/rootfs mount -t nfs 10.3.246.86:/homes/ /var/lib/kubelet/pods/59d30fb9-515b-40d3-abec-0958b8ac5d43/volumes/kubernetes.io~nfs/home-base]
Output: mount.nfs: Connection timed out
```

sgibson91:
Resolved! This was mostly an order of operations issue on my behalf. I was deleting the nfs server pod because I was nervous about needing to patch immutable volumes, but the nfs-share-creator pod needs the server to be running in order to do it's thing. However, because the nfs-share-creator runs _first_ and blocks everything else until it completes, it was waiting for a server to come up that would never come up, so would just time out.

We should improve our documentation to talk about toggling `nfs.enabled` when deploying jupyterhub-home-nfs (especially on fresh hubs that don't have any storage attached yet), to make the deployment experience smoother for engineers.

Issue #5494, opened 2025-02-04
[New Hub] [Phase 3] Hub Setup - Jetstream2, project-pythia with jupyterhub-home-nfs


GeorgianaElena:
### Context

Apart from the binderhub deployed for project-pythia on Jetstream2, we should also deploy a JupyterHub for 2i2c on Jetstream2 with jupyterhub-home-nfs enabled.

### How many hubs will be deployed?

2 (staging and prod)

### Which cluster will the hub(s) be deployed on?

tbd

### Hub Setup Information

For each hub to be deployed, copy the tables below and fill them in.
The tables must be complete for each hub to be considered READY to be
deployed. For a staging/prod pair, a statement such as "Same as staging
but for production" will suffice for the prod hub's specification. Use
the Notes column to provide any contextual information.

Available runbooks:
- https://infrastructure.2i2c.org/hub-deployment-guide/runbooks/phase3/

#### Phase 3.1: Initial setup

| Question | Answer | Notes |
| :--- | :--- | :--- |
| Name of the hub | staging | |
| Dask gateway? | `Yes/No` | no |
| Splash image URL | 2i2c | |
| Homepage URL | | | 
| Funded by? | | |
| Authentication Mechanism |GitHub | |
| Admin Users (GitHub handles or emails, depending on Mechanism) | | |

At the end of this phase, both 2i2c engineers and the admin users mentioned can login to the hub.

#### Phase 3.2: Object storage access

| Question | Answer | Notes |
| :--- | :--- | :--- |
| Scratch bucket enabled? | `Yes/No` | yes |
| Persistent bucket enabled? | `Yes/No` | no |
| Requester pays requests to external buckets allowed? | `Yes/No` |  no |  

At the end of this phase, both 2i2c engineers and the admin users mentioned can access any object storage setup.

#### Phase 3.3: Profile List

These are the standard profile list options to start with. They can be customized later.
Use the Notes column to provide extra information, such as specific image tags to use, or `deployer generate resource-allocation choices` command to use for RAM, etc.

| Question | Answer | Notes |
| :--- | :--- | :--- |
| Pangeo Notebook?  | `Yes/No` | |
| RStudio (with Rocker)? | `Yes/No` |  |
| Allow users to specify any image they want to use? | `Yes/No` | If Yes, enable `unlisted_choice`. | 
| Max RAM option allowed | | |
| GPU enabled? | `Yes/No` | |
| Default Interface | `Classic/JupyterLab/Other` | |
[...]

PR #5597, opened 2025-02-25
[new-hub] 2i2c jetstream2

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/5494


GeorgianaElena:
See https://github.com/2i2c-org/infrastructure/issues/5601 about the issue with this cluster. For now, I'm merging as it is because the hub has been deployed.

PR #5640, opened 2025-03-04
add custom display_name_in_choices for veda hub unlisted_choice options

batpad:
refs https://github.com/2i2c-org/jupyterhub-fancy-profiles/issues/104

This should work - am not sure if we've used this option in production before, so it might be nice to deploy to staging first "just in case".

cc @maxrjones @yuvipanda @sgibson91


GeorgianaElena:
I tested this on staging and it now shows `Bring your own image` instead of `Other...`.

Thank you @batpad!

PR #5759, opened 2025-03-19
Disable cluster-wide efs

GeorgianaElena:
For:
- [x] 2i2c-aws-us
- [x]  kitware
- [x] opensci
- [x] projectpythia


GeorgianaElena:
cc @sgibson91

sgibson91:
I think you could have just deleted the variable altogether as it is set to true in the variables.tf file?

https://github.com/2i2c-org/infrastructure/blob/96b64772f0510cb6bc61bb4c8f080327fc2b7848/terraform/aws/variables.tf#L202-L203

GeorgianaElena:
That's true, but I wanted to be explicit instead so that it's clear that when we're done with this, we can delete the whole efs-related resources, including the variable.

I plan to open an issue about it and add it to the parent initiative.

PR #5796, opened 2025-03-25
Bump up jupyterhub-fancy-profiles

sunu:
This PR rolls out the the upgrade to NASA VEDA hubs.

cc @batpad

### DIFF
--- a/config/clusters/nasa-veda/common.values.yaml
+++ b/config/clusters/nasa-veda/common.values.yaml
@@ -43,6 +43,8 @@ basehub:
             name: "NASA"
             url: https://www.earthdata.nasa.gov/esds
     hub:
+      image:
+        tag: "0.0.1-0.dev.git.12172.hf6a94ef2"
       allowNamedServers: true
       config:
         JupyterHub:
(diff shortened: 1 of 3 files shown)


batpad:
hmm @sunu this will update fancy profiles for ALL the hubs that are using it?

Is it possible to deploy this only to VEDA staging so we can do a bit of internal testing before further roll-out?

sunu:
@batpad my understanding is that, the hub images won't be updated till we update the tag at https://github.com/2i2c-org/infrastructure/blob/c05d7d8462fc502e742f7e2423104d34631a6966/helm-charts/basehub/values.yaml#L1057

Once we have an published image, we can selectively override the tag for VEDA staging only

batpad:
@sunu that makes sense, thanks for the explanation!

batpad (approved):

sunu:
Looking at https://github.com/2i2c-org/infrastructure/blob/c05d7d8462fc502e742f7e2423104d34631a6966/helm-charts/images/hub/Dockerfile#L4, we might have to manually build and push the image to the quay registry though. Not sure if this will happen automatically.

sunu:
I built and pushed the image to quay locally. In https://github.com/2i2c-org/infrastructure/pull/5797, we are using that image on VEDA staging hub to test jupyterhub-fancy-profiles 0.4.1

Issue #5754, opened 2025-03-19
Create an auth client for MAAP Staging hub in Veda keycloak


yuvipanda:
As the first hub to move over, we need to create an appropriate client in the VEDA Keycloak for the staging maap hub. This requires designing roles, scopes and groups appropriately to replicate the current setup. The current setup gates access via GitHub teams memberships.

## Definition of Done

- [x] Auth client for MAAP staging JupyterHub set up in veda keycloak (via PRs in https://github.com/NASA-IMPACT/veda-keycloak/)
- [x] The NASA MAAP hub is moved to use this client

yuvipanda:
Done now!

PR #5809, opened 2025-03-27
Move maap staging hub to use keycloak

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/5754


yuvipanda:
Currently failing with

image

yuvipanda:
All sorted out! There was an extra trailing slash in the config that was set up in veda-auth's CI system, and removing it got it to work

Issue #5845, opened 2025-04-02
Write smallest possible NetworkPolicy for dask-gateway


yuvipanda:
After https://github.com/2i2c-org/infrastructure/issues/5818, we want to write the smallest possible networkpolicy to get dask-gateway to work with networkpolicy being applied to the various *user controlled* components - dask-scheduler and dask-worker. We don't necessarily need them for the deployer controlled components (traefik, controller, api).

In particular, they should be able to talk freely to the external internet, but be restricted from talking to the internal kubernetes network except for what they need. Here's the [equivalent code for that in z2jh](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/c84e0c27c99493016d16ce67739b4105f9afdf1a/jupyterhub/templates/_helpers-netpol.tpl#L65).

## Definition of done

- [x] The communication lines required for dask-worker and dask-scheduler to function properly are identified.
- [x] NetworkPolicy is written to apply to dask-scheduler, and tested to work without issue on one cluster
- [x] NetworkPolicy is written to apply to dask-worker, and tested to work without issue on one cluster
- [x] Rolled out to all clusters

yuvipanda:
Let's do this as a template in our `basehub` chart for now, and once we have it fully working, we can contribute it upstream to dask-gateway

GeorgianaElena:
Copying here the message in the PR for better visibility:
I've deployed this to two clusters (one on GCP and one on AWS) and will wait for a couple of days to see if we get any reports of things not working there before deploying it everywhere as per Yuvi's suggestion.

Will merge on Monday morning. Moving it it to blocked until then.

PR #5870, opened 2025-04-07
NetworkPolicy for dask-workers

GeorgianaElena:
Ref: https://github.com/2i2c-org/infrastructure/issues/5845

I tested this on the 2i2c-aws-us dask-staging hub with:
` deployer run-hub-health-check 2i2c-aws-us dask-staging --check-dask-scaling`
(validated that the health check didn't pass when all egress traffic to private ips was blocked).

I figured I also need a conditional for this network policy and maybe do also a manual check that everything works?
Putting it in progress until I pick it up tomorrow but figured it's worth opening it now for early feedback from @yuvipanda


helm-charts/basehub/templates/network-policy-dask-worker.yaml
```diff
@@ -0,0 +1,48 @@
...
+          protocol: UDP
+        - port: 53
+          protocol: TCP
+      to:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: kube-system
+          podSelector:
+            matchLabels:
+              k8s-app: kube-dns
```
yuvipanda, line 33: Leave a comment here that while this says `kube-dns`, it also works with core-dns (what is actually used by EKS) because it has that as a label

yuvipanda:
Also, `netstat` while exec'd into a pod, esp in continuous mode (https://serverfault.com/questions/398234/netstat-continuous-refresh-watch-changes-the-output) is often very helpful when dealing with networkpolicy

yuvipanda:
Let's also validate that dask-workers can acquire AWS credentials if need be. Can be verified by running `aws sts get-caller-identity` from within a worker pod.

GeorgianaElena:
I've deployed this to two clusters (one on GCP and one on AWS) and will wait for a couple of days to see if we get any reports of things not working there before deploying it everywhere as per Yuvi's suggestion.

Will merge on Monday morning. Moving it it to blocked until then.

PR #5882, opened 2025-04-09
Add note about deleting EFS backups on decomissioning

yuvipanda:
- Also remove redundant 'pre defined definition of done'. It's never possible to know we never broke any existing functionality :)


GeorgianaElena:
Merging this to have it documented for future decoms

Issue #5410, opened 2025-01-24
[Platform Initiative] VEDA Objective JH-2: Add ability to remember previously used custom images and repositories to the logged in environment selector


jnywong:
### Motivation

When a user logs into the hub, they are presented with (among other things) a way to select the environment they want to use. One of the options is to specify their own Dcker image (‘bring your own image’), and another is to specify a GitHub repository from which an image will be dynamically built (‘bring your own repo’). It would be very helpful if both these remember prior options the user has used so they can immediately go back to a prior environment they have used instead of having to type it in each time.

### Description
As part of this objective, we’ll also be doing a conversion of the code-base from typed JavaScript to TypeScript. We expect this to take little time now because the code base is small, and enable us to move faster both for this objective and in the future.

### Success Criteria
- Bring your own image features remember previously used options so users can quickly select them
- The code-base is converted into TypeScript
- Dynamic image building has been rolled out to all hubs

### Personnel

- Oliver (Development Seed)
- @sunu 
- @yuvipanda

batpad:
Feature to "remember" previously used options / images is being tracked here: https://github.com/2i2c-org/jupyterhub-fancy-profiles/issues/43

Conversion to TypeScript is close to completion in this PR: https://github.com/2i2c-org/jupyterhub-fancy-profiles/pull/101

Dynamic image building roll-out to all hubs: @sunu do we have a ticket for this? If not, could you make one and drop the link here, thanks.

yuvipanda:
I just spoke to @oliverroick and he's going to pick up https://github.com/2i2c-org/jupyterhub-fancy-profiles/issues/43 now

batpad:
This is deployed to the VEDA staging hub. Asked some folks to test and if all looks good, we should be able to roll this out later this week.

PR #5886, opened 2025-04-09
Roll out jupyterhub-fancy-profiles 0.4.1 to all hubs

sunu:
Related to https://github.com/2i2c-org/infrastructure/issues/5410. cc @batpad

### DIFF
--- a/config/clusters/nasa-veda/common.values.yaml
+++ b/config/clusters/nasa-veda/common.values.yaml
@@ -43,8 +43,6 @@ basehub:
             name: "NASA"
             url: https://www.earthdata.nasa.gov/esds
     hub:
-      image:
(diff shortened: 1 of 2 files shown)


sunu:
Not sure why the docs build is failing

sgibson91:
> Not sure why the docs build is failing

Seems to be transient as I triggered a rebuild and it has at least passed the repo cloning stage. Perhaps you opened the PR while GitHub was having an API issue? Either way, docs are irrelevant for this PR.

sgibson91 (approved):
Can we do a quick check of any other hubs that may be currently overriding the default hub image like VEDA is?

sunu:
@sgibson91 I did a quick check. Doesn't look like any other hub is overriding the default hub image.

Issue #5665, opened 2025-03-05
Rollout storage quotas to  openscapes


GeorgianaElena:

yuvipanda:
Re-opened to make sure we don't miss prometheus.

PR #5935, opened 2025-04-22
Migrate openscapes hubs to use jupyterhub-home-nfs

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/5665


GeorgianaElena (commented):
@yuvipanda, great job with this! I noticed prometheus alerting was not setup with this PR. Is the plan to do this later on?

yuvipanda:
@GeorgianaElena I had missed it, I've re-opened it in https://github.com/2i2c-org/infrastructure/issues/5665

PR #5937, opened 2025-04-22
catalystproject-africa: Upgrade to k8s v1.32

### LINKED ISSUE #5702: [catalystproject-africa] Upgrade cluster

sgibson91:
- fixes #5702


sgibson91:
@GeorgianaElena are we deleting the ssh keys from the repo when we do these as well?

GeorgianaElena:
@sgibson91, yes 👍🏼 , I think we can delete those as we're not using them anymore

PR #5947, opened 2025-04-23
fix the echopype image link

eeholmes:


eeholmes:
might try different solution

PR #5984, opened 2025-04-29
Rm tainted node

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/5709


sgibson91:
Thanks @GeorgianaElena!

PR #5990, opened 2025-04-30
Patch tigera-operator deployment to rm NoSchedule toleration

### LINKED ISSUE #5961: [AWS] tigera-operator gets deployed on user nodes blocking scale downs
### Context

While upgrading the aws clusters, I've noticed a few user nodes active but with no active users on them. 
This happened on more clusters.
Upon inspecting them I noticed that `tigera-operator` was running on them.

For context, this also happened https://github.com/2i2c-org/infrastructure/pull/5913

### What we need to do

I guess one option could be to patch the tigera-operator deployement in https://github.com/2i2c-org/infrastructure/blob/339c2b4e8026648dc0e613fc82de8d7350e4f50c/deployer/infra_components/cluster.py#L83-L92 to have a node selector so that we make sure it gets on core nodes?

### Definition of Done

- [ ] The feature/service is technically complete
- [ ] The feature/service been tested with one or more users (if applicable)
- [ ] The feature/service been deployed to a production cluster
- [ ] The feature/service has been shown to be replicably deployable
- [ ] The feature/service is well documented, and the documentation is accessible for the target user base
- [ ] The feature/service has been added to our product menu (if applicable).
- [ ] Lessons learned have been documented
- [ ] The feature's availability has been communicated to Sales
- [ ] The feature has been communicated to our communities

GeorgianaElena:
Fixes #5961

### DIFF
--- a/deployer/infra_components/cluster.py
+++ b/deployer/infra_components/cluster.py
@@ -94,6 +94,35 @@ class Cluster:
             )
             print_colour("Done!")
 
+            # Patch the tigera operator to remove the NoSchedule toleration
+            # otherwise it will schedule on tainted nodes
+            print_colour("Patching tigera operator...")
+            patch_tolerations = {
+                "spec": {
+                    "template": {
+                        "spec": {
+                            "tolerations": [
+                                {"effect": "NoExecute", "operator": "Exists"},
+                            ],
+                        }
+                    }
+                }
+            }
+            patch_tolerations_json = json.dumps(patch_tolerations)
+            subprocess.check_call(
+                [
+                    "kubectl",
+                    "--namespace",
+                    "tigera-operator",
+                    "patch",
+                    "deployment",
+                    "tigera-operator",
+                    "--patch",
+                    patch_tolerations_json,
+                ],
+            )
+            print_colour("Done!")
+
         print_colour("Provisioning support charts...")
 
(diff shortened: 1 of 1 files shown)


deployer/infra_components/cluster.py
```diff
@@ -94,6 +94,35 @@ def deploy_support(self, cert_manager_version, debug):
...
 
+            # Patch the tigera operator to remove the NoSchedule toleration
+            # otherwise it will schedule on tainted nodes
+            print_colour("Patching tigera operator...")
+            patch_tolerations = {
+                "spec": {
+                    "template": {
+                        "spec": {
+                            "tolerations": [
+                                {"effect": "NoExecute", "operator": "Exists"},
```
GeorgianaElena, line 105: This is coming from https://github.com/projectcalico/calico/blob/da6b089ff18f2e4ab104def0f78e68c8741fc221/manifests/tigera-operator.yaml#L535-L539

Wasn't sure if I should keep it, but for now it doens't bother us _I think_

yuvipanda:
I think this is sufficient, but can you change the comment in `--force-apply` to no longer say 'remove this after some task is done' but to document *why* we need it? Given we want the source of truth to be that apply, and our patch is minor afterwards, this feels fine.

PR #5991, opened 2025-04-30
[tigera-operator] Update the manifest directly

### LINKED ISSUE #5961: [AWS] tigera-operator gets deployed on user nodes blocking scale downs

GeorgianaElena:
Fixes #5961 

This updates the previous PR #5990 and updates the manifest directly to not contain the "NoSchedule" toleration before doing a server-side apply.

## Reasoning

This is because `kubectl patch`  cannot be ran server-side and so it will cause ownership conflicts, forcing us to keep the `-force-conflicts` from the initial apply command.

Before doing this I've also considered https://kubernetes.io/docs/reference/using-api/server-side-apply/#transferring-ownership, which worked, but the second apply would still need to take ownership using `-force-conflicts`.

## Open question

We either do this, or keep the `k apply --force-conflicts` followed by the `k patch` which from my understanding just gives ownership back to `kubectl` like it was initially?

### DIFF
--- a/deployer/infra_components/cluster.py
+++ b/deployer/infra_components/cluster.py
@@ -5,6 +5,9 @@ import tempfile
 from contextlib import ExitStack, contextmanager
 from pathlib import Path
 
+import requests
+import yaml
+
 from deployer.infra_components.hub import Hub
 from deployer.utils.env_vars_management import unset_env_vars
 from deployer.utils.file_acquisition import (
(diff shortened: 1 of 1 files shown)


GeorgianaElena:
@yuvipanda can you please take a look at this and help me decide if this is the way to go, or if the approach in https://github.com/2i2c-org/infrastructure/issues/5961 is sensible enough?

deployer/infra_components/cluster.py
```diff
@@ -82,46 +85,43 @@ def deploy_support(self, cert_manager_version, debug):
...
+                    and resource.get("metadata", {}).get("name") == "tigera-operator"
+                ):
+                    spec = resource.get("spec", {}).get("template", {}).get("spec", {})
+                    if spec.get("tolerations", []):
+                        spec["tolerations"] = [
+                            {"effect": "NoExecute", "operator": "Exists"}
+                        ]
+            try:
+                with tempfile.NamedTemporaryFile(
+                    mode="w+", suffix=".yaml", delete=False
```
yuvipanda, line 107: why do delete=False?

Issue #5411, opened 2025-01-24
[P&S Initiative] VEDA Objective JH-3: Deploy per-group shared directories


jnywong:

batpad:
It seems like all the upstream releases we were waiting on to finish this work are done, and we should be able to proceed with making the configuration changes in the `infrastructure` repository to actually enable this for users.

My understanding of what this would look like, roughly:

 - Create a test "team" in Github to test this feature with.
 - Verify that the team is being mapped to "groups" in JupyterHub - this should work correctly.
 - Convert **all** `volumeMounts` across the `infrastructure` repository to be `dicts` instead of `lists`, following up on earlier work by @sunu to allow for this, eg: https://github.com/2i2c-org/infrastructure/blob/main/config/clusters/nasa-veda/common.values.yaml#L105
 - Specify `group_overrides` to add `volumeMounts` like `_shared-group/{group-name}` or so for groups that need group-shared directories, see https://jupyterhub.readthedocs.io/en/stable/reference/api/spawner.html#jupyterhub.spawner.Spawner.group_overrides
 - Test that this works and that the group-shared folders are only visible to members of the group.

@sunu does this make sense and is this something you'd be able to get started on?

@yuvipanda can you confirm that we'd need to change `volumeMounts` to `dicts` for **all** `volumeMounts` for all clusters in the `infrastructure` repository?

yuvipanda:
> @yuvipanda can you confirm that we'd need to change volumeMounts to dicts for all volumeMounts for all clusters in the infrastructure repository?

yes

sunu:
@batpad sounds good! I'll get to this after finishing https://github.com/2i2c-org/infrastructure/issues/5414.

aprilmj:
I'd like us to discuss what it would take to make per-group shared directories more widely available. It's come back up in recent support conversations: https://2i2c.freshdesk.com/a/tickets/3375

Gman0909:
@aprilmj I suggest creating an issue along the lines of "Figure out how to make per-group shared directories more widely available," ideally with a list of candidate hubs, which we can pick up in an upcoming sprint. Alternatively, if it feels like a bigger lift than one or two sprints, we could shape it as an initiative with separate design and implementation phases.

PR #6010, opened 2025-05-02
MAAP: test group shared directories on staging

sunu:
@yuvipanda Would you mind taking a look and letting me know if this seems reasonable?

I’ve added an `extraConfig` block that converts `volumes` and `volume_mounts` from lists to dictionaries so that we can add new entries without replacing the existing values.

It works ok based on my testing, but I’m wondering whether this change should instead live upstream in zero-to-jupyterhub, where these fields are currently defined as lists.

Related to https://github.com/NASA-IMPACT/veda-jupyterhub/issues/66

### DIFF
--- a/config/clusters/maap/staging.values.yaml
+++ b/config/clusters/maap/staging.values.yaml
@@ -20,7 +20,13 @@ jupyterhub:
         command:
           - sh
           - -c
-          - id && chown 1000:1000 /home/jovyan /home/jovyan/shared /home/jovyan/shared-public && ls -lhd /home/jovyan
+          - >
+            id &&
+            chown 1000:1000 /home/jovyan /home/jovyan/shared /home/jovyan/shared-public /home/jovyan/shared-group &&
+            if [ -d "/home/jovyan/shared-group" ] && [ "$(ls -A /home/jovyan/shared-group)" ]; then
+              chown 1000:1000 /home/jovyan/shared-group/* || true;
+            fi &&
+            ls -lhd /home/jovyan
         securityContext:
           runAsUser: 0
         volumeMounts:
@@ -35,6 +41,9 @@ jupyterhub:
           - name: home
             mountPath: /home/jovyan/shared-public
             subPath: _shared-public
+          - name: home
+            mountPath: /home/jovyan/shared-group
+            subPath: _shared-group
 
     profileList:
       - display_name: Choose your environment and resources
--- a/config/clusters/maap/staging.values.yaml
+++ b/config/clusters/maap/staging.values.yaml
@@ -252,6 +252,53 @@ jupyterhub:
           - CPU:XXL
           - CPU:XXXL
           - GPU:T4
+    extraConfig:
+      00-volumes-and-volume-mounts-as-dict: |
+        # The base jupyterhub config in zero-to-jupyterhub defines
+        # volumes and volume_mounts as lists.
+        # But we can't add new volumes or volume_mounts to the list
+        # as that replaces the entire list.
+        # So we convert them to dictionaries, which allows us to
+        # add new volumes and volume_mounts as needed.
+        if isinstance(c.KubeSpawner.volumes, list):
+          existing_volumes = c.KubeSpawner.volumes
+          c.KubeSpawner.volumes = {}
+          for volume in existing_volumes:
+            c.KubeSpawner.volumes[volume["name"]] = volume
+        if isinstance(c.KubeSpawner.volume_mounts, list):
+          existing_volume_mounts = c.KubeSpawner.volume_mounts
+          c.KubeSpawner.volume_mounts = {}
+          for idx, volume_mount in enumerate(existing_volume_mounts):
+            c.KubeSpawner.volume_mounts[f"{idx}-{volume_mount['name']}"] = volume_mount
+      01-group-shared-directories: |
+        c.KubeSpawner.group_overrides = {
+          "00-group-CPU-L-extra-volume-mounts": {
+            "groups": ["CPU:L"],
+            "spawner_override": {
+              "volume_mounts": {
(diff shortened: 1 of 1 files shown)


sunu:
This needs a bit more work to fix permissions on the group shared directories. Marking as draft for now.

sunu:
Hi @GeorgianaElena, I think this is ready for review now! I've deployed it to the MAAP staging hub and done some basic testing with my user. I was able to confirm that the configured group folder is correctly mounted and accessible for read/write, based on group membership for my user.

If you'd like more background, there’s additional context on group shared folders in [this VEDA issue](https://github.com/NASA-IMPACT/veda-jupyterhub/issues/44) and [this 2i2c one](https://github.com/2i2c-org/infrastructure/issues/5411). Let me know if there's anything else I can clarify or help with!

GeorgianaElena (approved):
Thank you @sunu! This PR is awesome and it looks safe to try 🚀 Feel free to merge whenever you are ready.

> I'm still unsure whether the list-to-dict conversion should ideally happen upstream instead. For now though, to get this out the door, I think it's fine to handle it in the hub's extraConfig. It does end up being a bit verbose, but manageable for the short term.

I agree that ideally this would be supported upstream, but that might take a while and you're decision to not block on it makes sense to me.
Also, when we'll roll this feature out to all hubs, we can add this config directly to the basehub chart in https://github.com/2i2c-org/infrastructure/blob/main/helm-charts/basehub/values.yaml#L1129 to make it easier to both maintain and reuse.

> The shared group folders are owned by the user, which means there's nothing preventing them from accidentally deleting the folder. That said, this is consistent with how writable shared folders like shared-public already behave.

I think it's fine as longs as it's documented and the end-goal is to have non-admin users write to this shared group dir as well. Otherwise _I think_ the alternative would be to think about how to integrate admin membership into this, though I feel that's out of scope for now.

> May be we want separate groups configured in Keycloak for resource quotas and file sharing? Naming shared quotas based on groups meant to define resource quota options seems awkward. This is something @wildintellect raised in
https://github.com/NASA-IMPACT/veda-keycloak/issues/93 too.

I think this is something Yuvi can answer best. But do you have an example here that you can share so I can understand the concern better?

sunu:
> But do you have an example here that you can share so I can understand the concern better?

In this PR, we're using groups like `GPU_T4` and `CPU_L` to create shared folders for users. So the shared folder looks like `~/shared-group/CPU_L`. These same groups also control what kinds of servers users can spin up. I’m thinking that in production, we’ll probably want to separate these concerns — have one set of groups in Keycloak for server resource access (like `GPU_T4`), and another set for file access (like `MAAP GeoTrees`).

There’s nothing technically blocking us from doing this — I just wanted to point out that the current group setup in Keycloak is focused only on server access. We should plan to add more descriptive, purpose-specific groups for like file sharing, ideally reflecting real-world team names, and make sure we document that somewhere.

sunu:
There's a failing test. Rerunning the test to see if that fixes the error.

sunu:
Hmm, rerunning didn't fix the error. @GeorgianaElena any suggestion on what to do for the failing test?

GeorgianaElena:
@sunu, the validation test should not run on `common` and it actually shouldn't run at all on this PR as this doesn't touches it. Can you rebase your PR against main? I suspect it won't run the validation after this.

But even if it doesn't fix it, feel free to ignore that validation test and merge it anyway. I'll open an issue to investigate it on our side.

sunu:
@GeorgianaElena thanks! that worked!

config/clusters/maap/staging.values.yaml
```diff
@@ -252,6 +261,53 @@ jupyterhub:
           - CPU:XXL
           - CPU:XXXL
           - GPU:T4
+    extraConfig:
+      00-volumes-and-volume-mounts-as-dict: |
```
yuvipanda, line 265: I think this is unsustainable long term, and should be followed up with a PR to z2jh that allows them to be dicts or lists.

yuvipanda: Alternatively, instead of using `singleuser.profileList`, we could consider using `hub.config.KubeSpawner.profile_list`, or `hub.config.KubeSpawner.volume_mounts` (or similar). What is preventing us from doing that?

yuvipanda: To expand on this, we currently set `singleuser.storage` to set up home directories, here: https://github.com/2i2c-org/infrastructure/blob/main/helm-charts/basehub/values.yaml#L784.

There are two components here:

1. Set up `storage.type` to static, and configure that to mount a PVC. This adds a `home` volume to the singleuser storage.
2. Set up `storage.extraVolumes` to mount some extra Volumes
3. Set up `storage.extraVolumeMounts` to mount the `home` volume in a few other places.

We could instead convert this to:

1. Set `storage.type` to `null` to prevent z2jh special storage config setting
2. Use `hub.config.KubeSpawner.volumes` to add the `home` volume ourself - this is exactly what z2jh itself does (https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/fa6414fc50bd88ff9725cc1228a6c8f177dbd5c3/jupyterhub/files/hub/jupyterhub_config.py#L289). We can simply do this ourselves, and can thus set `volumes` as a dict without needing any changes to z2jh
3. Do (2) and (3) above as `hub.config.KubeSpawner.volumes` and `hub.config.KubeSpawner.volume_mounts`. When I initially designed z2jh, the `extra` in `storage.extraVolumes` was to indicate this was additional volumes on top of `storage.type` injected volumes. We simply can set `storage.type` to `null` and opt out, and use `volumes` and `volume_mounts` directly. 

This would allow us to make implement what we need without having to change z2jh first. We should still change z2jh ofc, but the order doesn't need to be blocking

PR #6016, opened 2025-05-05
Update common.values.yaml

iamzoltan:
Please update our permission group for 2025. can you let us know when it is deployed?

### DIFF
--- a/config/clusters/climatematch/common.values.yaml
+++ b/config/clusters/climatematch/common.values.yaml
@@ -97,7 +97,7 @@ jupyterhub:
         authenticator_class: github
       GitHubOAuthenticator:
         allowed_organizations:
-          - ClimateMatchAcademy:2024students
+          - ClimateMatchAcademy:2025students
         scope:
           - read:org
       Authenticator:


GeorgianaElena:
@iamzoltan this change should now be deployed

iamzoltan:
@GeorgianaElena Any idea why user added to the group cannot access the hub?

GeorgianaElena:
@iamzoltan, I suspect the reason is that the OAuth app wasn't granted the correct permissions.
Can you please temporarily make me an admin of that GitHub org to reset it and then grant the permissions? I will revoke my membership afterwards.

Also, https://infrastructure.2i2c.org/sre-guide/common-problems-solutions/#resetting-a-github-oauth-app-when-users-are-faced-with-a-403-forbidden-error-at-login has more info about this common issue.

iamzoltan:
@GeorgianaElena what permissions do you need? Did the required permissions change from last year?

As an admin, I can access the hub no issues.

GeorgianaElena:
@iamzoltan, I need to temporary be made an admin of the `ClimateMatchAcademy` GitHub org. This would be a one time thing to make sure the correct permissions are granted to the OAuth app at login.

Alternatively if you are an admin of the GitHub organization (not the hub) you could follow step 3 from https://infrastructure.2i2c.org/sre-guide/common-problems-solutions/#resetting-a-github-oauth-app-when-users-are-faced-with-a-403-forbidden-error-at-login yourself as I already did step 1 and 2.

iamzoltan:
Hey @GeorgianaElena Thanks for that link. I think we solved the issue on our end.

That guide your shared to the 2i2c docs could use a little clarity. Do you have a repo for the docs to make a pr?

PR #6017, opened 2025-05-05
Update python image on NASA-Openscapes

ateucher:
Updates the default Python image on NASA Openscapes prod and workshop hubs. Has been tested by users using "bring your own image"


ateucher:
cc @betolink

PR #6020, opened 2025-05-07
Update defaulturl in climaterisk.values.yaml

dhavide:
Please merge soon so I can verify the change in the infrastructure

### DIFF
--- a/config/clusters/opensci/climaterisk.values.yaml
+++ b/config/clusters/opensci/climaterisk.values.yaml
@@ -63,7 +63,7 @@ jupyterhub:
       jupyterHub:
         authenticator_class: dummy
   singleuser:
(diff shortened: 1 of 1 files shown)


dhavide:
@GeorgianaElena this did not work. I'm not sure why but now the infrastructure fails to spin up. I must have made a mistake when entrting the nbgitpuller URL into the `climaterisk.values.yaml` file.

I'm unclear on how to proceed from here to correct this. Do we revert the `main` branch of the repo to its original state?

GeorgianaElena:
@dhavide, I've reverted the PR which should bring the hub back to its previous state. In the new like I see `git-pull` two times, and the first time is refering to the hub url. So I think this is where the issue is.

Issue #5983, opened 2025-04-29
Rollout jupyterhub-groups-exporter for all hubs with `manage_groups` feature


jnywong:
### Context

Following the rollout `manage_groups` feature in #5805, we need to verify that the feature is working as intended before rolling out the `jupyterhub-groups-exporter`. We also need to update the code base to remove legacy code pertaining to upstream PR [upstream PR](https://github.com/jupyterhub/oauthenticator/pull/735). See also related issue https://github.com/2i2c-org/infrastructure/issues/5314.

### Tasklist

- [x] Verify that all GitHub authenticated hubs have `manage_groups` working as expected. This means that users that have logged out and in again are populating the groups table in the JH database as expected.
- [x] Rollout `jupyerhub-groups-exporter` to Earthscope prod hub
- [x] Remove legacy code pertaining to [upstream PR](https://github.com/jupyterhub/oauthenticator/pull/735) from `jupyerhub.hub.extraConfig`in `helm-charts/basehub/values.yaml` in the test hub.
- [x] Validate that the legacy code is indeed no longer needed after removal and that the hub behaves as expected
- [ ] Repeat the above for all eligible hubs
- [ ] Consolidate config change into basehub chart values https://github.com/2i2c-org/infrastructure/pull/6032

### List of GH authenticated hubs

- [x] `2i2c-aws-us` https://github.com/2i2c-org/infrastructure/pull/6078
- [x] `2i2c-jetstream2` https://github.com/2i2c-org/infrastructure/pull/6077
- [x] `2i2c-uk` https://github.com/2i2c-org/infrastructure/pull/6080
- [x] `awi-ciroh` https://github.com/2i2c-org/infrastructure/pull/6104
- [x] `catalyst-project-africa` https://github.com/2i2c-org/infrastructure/pull/6082
- [x] `catalyst-project-latam` https://github.com/2i2c-org/infrastructure/pull/6083
- [x] `climatematch` https://github.com/2i2c-org/infrastructure/pull/6084
- [ ] ~~`cloudbank`~~ these hubs do not have any extra features rolled out
- [ ] `disasters` ? `generic-oauth` class used, managed by keycloak
- [x] `earthscope` https://github.com/2i2c-org/infrastructure/pull/6076
- [x] `jupyterhealth` https://github.com/2i2c-org/infrastructure/pull/6110
- [x] `leap` https://github.com/2i2c-org/infrastructure/pull/6085
[...]

Issue #5805, opened 2025-03-26
Phased rollout of `manage_groups` feature for all GitHub authenticated hub deployments


jnywong:
### Context

Following [upstream work](https://github.com/jupyterhub/oauthenticator/pull/735) to manage groups with JupyterHub's base oauthenticator, we should enable this feature across all of our hubs by setting `hub.config.Authenticator.manage_groups=True` in `common.values.yaml`. This is currently true for Earthscope and JupyterHealth, but not all of our other communities.

This will only affect hubs with GitHub authentication, since groups can be managed at that level. This will not apply to hubs with CILogon.

We need to roll out this feature over time, since group creation at the JupyterHub level will require users to be logged out and back into the hub. In another sprint, we can follow up to verify groups are created and working as expected and then legacy code from `helm-charts/basehub/values.yaml` can be removed.

### Tasklist

- [x] Enable `manage_groups` for a test GitHub authenticated hub
- [x] Verify feature works as expected on the test hub
- [x] Roll out `manage_groups` for all GitHub authenticated hubs
- [x] Lightly monitor and evaluate rollout in preparation for https://github.com/2i2c-org/infrastructure/issues/5983

jnywong:
This is taking longer than expected and likely to run slightly into the next sprint.

The reason is because for the prometheus groups exporter service to work, we need to set `c.Authenticator.manage_groups = True` to populate the hub database with groups.

My main focus is on hubs using GitHubOAuthenticator, which means that I should also set `populate_teams_in_auth_state` to populate the `auth_state` with GH teams data. However, the [documentation](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.github.html#oauthenticator.github.GitHubOAuthenticator.populate_teams_in_auth_state) says that "that authentication state is only be available to a `post_auth_hook`", which is at a stage too late to update the groups table for the hub database.

Therefore my workaround is to override GitHubOAuthenticator with a custom class:

```python
  class CustomGitHubOAuthenticator(GitHubOAuthenticator):
    async def authenticate(self, *args, **kwargs):
        resp = await super().authenticate(*args, **kwargs)
        if resp["auth_state"]["teams"] is None:
          return resp
        resp["auth_state"]["groups"] = [f'{team["organization"]["login"]}:{team["slug"]}' for team in resp["auth_state"]["teams"]]
        return resp

  c.JupyterHub.authenticator_class = CustomGitHubOAuthenticator
```

[...]

Issue #5314, opened 2024-12-20
[Spike] Investigate which special-cased authentication configuration we can remove/update now that some of it has been upstreamed


GeorgianaElena:
After we've migrated the hubs to use z2jh4, we will be using a more recent `oauthenticator` version, 17.1.0 that comes with some [changes](https://oauthenticator.readthedocs.io/en/stable/reference/changelog.html)

We should investigate removing/updating the hubs that still use special-cased configuration that has now been upstreamed in this version of oauthenticator (earthscope, hubs using `managed_groups` and GitHubOauthenticator)

### Definition of Done
- [ ] the config that needs to be updated or removed has been identified (if any)

PR #6032, opened 2025-05-12
Update basehub chart values to enable jupyterhub-groups-exporter service

jnywong:
This PR follows up https://github.com/2i2c-org/infrastructure/issues/5805 and consolidates individual canary deployments to enable `jupyterhub-groups-exporter` service into the `basehub/values.yaml` file.

Key changes:

- `jupyterhub.hub.extraConfig` has a new config called `08-auth-state-groups`
  - If `GitHubOAuthenticator` is enabled, then GH teams, of the form `<gh-org>:<gh-team>` who are members of the `GitHubOAuthenticator.allowed_organizations`, are passed from the  `auth_state` of the GitHub authentication to the JupyterHub state
  - this requires `GitHubOAuthenticator.populate_teams_in_auth_state.allowed_organizations`, `GitHubOAuthenticator.manage_groups` and `GitHubOAuthenticator.enable_auth_state` to be enabled, as well as `scope` set to `read:org`
  - other hubs, such as shared password workshop or unauthenticated binder hubs, **must have these options disabled,** since there is no `auth_state` to pass and so results in error 500 otherwise
  - the above options are set in `basehub/values.yaml` file and conditionally enabled when `JupyterHub.authenticator_class = github`
- `jupyterhub.hub.extraConfig` has a new config called `09-2i2c-add-staff-gh-team`
  - appends 2i2c staff access by GitHub team membership by default for GitHub authenticated hubs.
- the `jupyterhub-groups-exporter` service is added to `jupyterhub.hub.services` and associated role is added to `jupyterhub.hub.loadRoles`
- `jupyterhub-groups-exporter` is enabled by default from the `basehub/values.yaml` file, which is a tradeoff for simpler config for each hub-specific `values.yaml` file

Related https://github.com/2i2c-org/infrastructure/issues/5314

Closes https://github.com/2i2c-org/infrastructure/issues/5983


jnywong:
I'm going to follow up this PR with a separate one to enable [Authenticator.refresh_user](https://jupyterhub.readthedocs.io/en/stable/reference/api/auth.html#jupyterhub.auth.Authenticator.refresh_user) and [Authenticator.auth_refresh_age](https://jupyterhub.readthedocs.io/en/stable/reference/api/auth.html#jupyterhub.auth.Authenticator.auth_refresh_age) so that all affected users are required to log in again to keep `auth_state` information up to date.

helm-charts/basehub/values.yaml
```diff
@@ -1447,17 +1455,32 @@ jupyterhub:
...
+      05-auth-state-groups-key: |
+        # Pass the user's GitHub teams from auth_state as groups to the authenticator
+        async def custom_auth_state_groups_key(auth_state):
+          groups_list = []
+          if auth_state is None:
+            print("Auth state is None")
+            return groups_list
+          else:
+            for team in auth_state["teams"]:
+              if f'{team["organization"]["login"]}:{team["slug"]}' not in c.GitHubOAuthenticator.allowed_organizations:
```
yuvipanda, line 1467: We should unconditionally put all teams into groups, as we may change what's in `c.GitHubOAuthenticator.allowed_organizations` later on. And if we do it conditionally, we'll need to force people to re-log in. This does possibly create extra entries, but I think that's ok.

helm-charts/basehub/values.yaml
```diff
@@ -1447,17 +1455,32 @@ jupyterhub:
...
+                continue
+              else:
+                groups_list.append(f'{team["organization"]["login"]}:{team["slug"]}')
+            return groups_list
+        if c.GitHubOAuthenticator:
+          c.GitHubOAuthenticator.allowed_organizations.append("2i2c-org:hub-access-for-2i2c-staff")
```
yuvipanda, line 1473: This is unrelated, so it should be in its own separate entry rather than bundled in here.

helm-charts/basehub/values.yaml
```diff
@@ -1447,17 +1455,32 @@ jupyterhub:
...
+            print("Auth state is None")
+            return groups_list
+          else:
+            for team in auth_state["teams"]:
+              if f'{team["organization"]["login"]}:{team["slug"]}' not in c.GitHubOAuthenticator.allowed_organizations:
+                continue
+              else:
+                groups_list.append(f'{team["organization"]["login"]}:{team["slug"]}')
+            return groups_list
+        if c.GitHubOAuthenticator:
```
yuvipanda, line 1472: You can use `hasattr` here as an additional guard, so you won't need to put empty `GitHubOAuthenticator` entries in various pieces of config.

yuvipanda (changes requested):
This is excellent work Jenny! I have a few changes for how you can roll this out with a little more simplicity, and reduced ongoing effort. I'll also try to list some of the principles that drive these.

First, I think we should try to minimize the number of decisions that someone deploying a new hub should make. If it is possible for things to be 'automatic', we should bake that design into it from the start. Here are a couple changes I think we should make from that perspective:

1. Unconditionally enable jupyterhub-group-exporter everywhere. For hubs that currently don't use groups, this just means we won't have any data (unless they explicitly create groups in the admin interface). The groups API call should be empty. This is a trade-off - we are choosing to spend a little more resources (RAM and CPU) for simplicity of config (so we don't also need to explicitly enable / disable this each time). 
2. z2jh (and traitlets) uses the following pattern that we can exploit to set 'common config, but only for GitHubOAuthenticator':
	1. You can set `hub.config.GitHubOAuthenticator.populate_teams_in_auth_state`, `hub.config.GitHubOAuthenticator.enable_auth_state` and `hub.config.GitHubOAuthenticator.manage_groups` in `basehub/values.yaml`. This works because `GitHubOAuthenticator` inherits eventually from `Authenticator` - so `hub.config.GitHubOAuthenticator.enable_auth_state` and `hub.config.Authenticator.enable_auth_state` are equivalent!
	2. Whether `GitHubOAuthenticator` is used or not is controlled by `hub.config.JupyterHub.authenticator_class`. If it is not setup to be `GitHubOAuthenticator` or `github`, the config we set via `hub.config.GitHubOAuthenticator` is just ignored. So this lets you set common config in `basehub/values.yaml` without having to set it on every hub, and then creating a potential point for human error - where someone setting up a new hub may forget!

Second is to minimize chances of issues during rollout by changing one thing at a time, and in a form that is easily revertable without causing too many state changes in the hub database. From this perspective, here are some changes:
1. Leave the existing code in `05-profile-groups` alone for now, as it doesn't affect the jupyterhub-groups-exporter functionality. Changing it here increases the chance that trying to roll out one thing (groups-exporter) causes issues in another thing (profile list restriction based on group membership). We can remove this in a separate PR, once this rollout is successfully done and we have confidence in its stability. This also means we don't need to *enforce* logouts right now - profile list functionality is user facing, while prometheus is not. 

I've left some other minor inline comments.

Can you also write out what the logout effects / plan is? I know we had a short conversation in slack, but I'd like to understand that better. Will merging this PR cause any logouts?

Hope this was helpful!

helm-charts/basehub/values.yaml
jnywong: I thought about this and decided against doing this unconditionally.

Users grant `read:org` access to populate GH teams in the `auth_state` that could be completely unrelated to the hub (including private memberships) that will populate the JH groups table in the database and therefore pass to other places like the `jupyterhub_groups_exporter`.

Therefore to keep authorisation up to date with `c.GitHubOAuthenticator.allowed_organizations` we should rely on users to log in again, which I will write up separately.

helm-charts/basehub/values.yaml
jnywong: Now that we can set config for `GitHubOAuthenticator` in `basehub/values.yaml`, I have opted to include this in the list of `allowed_organizations` (see [Separate 2i2c-add-staff-gh-team access](https://github.com/2i2c-org/infrastructure/pull/6032/commits/43a7ae0a82fe5f10cc878ac7db18bb83aebed365))

helm-charts/basehub/values.yaml
jnywong: What object attribute would I consider applying `hasattr` to?

I thought about invoking `custom_auth_state_groups_key(auth_state)` to determine whether `groups_list` is populated, and if so, using that as a condition to set `c.GitHubOAuthenticator.auth_state_groups_key`, but this requires the `auth_state` input outside of runtime so not possible.

I am missing something obvious here 😅

jnywong: I've updated this to say

```python
        if c.JupyterHub.authenticator_class == "github":
          c.GitHubOAuthenticator.auth_state_groups_key = custom_auth_state_groups_key
```

and happy with how that works for now.

jnywong:
> Can you also write out what the logout effects / plan is? I know we had a short conversation in slack, but I'd like to understand that better. Will merging this PR cause any logouts?

I've done more research into this, and here is my overview:

### Rollout

There will be many users who are already logged in to hubs who will not be automatically logged out during rollout. JupyterHub group memberships will only be updated if a user manually logs out and back in again, or by using the *Revoke all user tokens* in the GitHub OAuth App settings.

### Desired behaviour

GitHub team memberships are in sync with JupyterHub user group memberships. 

### Possible solution

The [OAuthenticator docs](https://oauthenticator.readthedocs.io/en/latest/how-to/refresh.html#) has guidance on how to refresh a user authentication. However, the `refresh_user_hook` is only available in `OAuthenticator` version 17.3 but we are using version 4.0.0 of the `z2jh` helm chart that bundles with `OAuthenticator` version 17.1.0.

Sample code for what `refresh_user_hook` could look like, taking inspiration from [oauthenticator/github.py](https://github.com/jupyterhub/oauthenticator/blob/6d08bf21445888193adee225d9749c1098af91ae/oauthenticator/github.py#L167): 

  ```python
  10-refresh-user: |
    # Refresh user data to sync GH teams with JupyterHub groups.

    async def refresh_user_hook(authenticator, user, auth_state):
      access_token = auth_state["token_response"]["access_token"]
      token_type = auth_state["token_response"]["token_type"]
      for org_team in authenticator.allowed_organizations:
          if await authenticator._check_membership_allowed_organizations(
              org_team, user.name, access_token, token_type
          ):
            return True
      message = f"User {user.name} is not part of allowed_organizations"
      print(message)
      return False

    c.GitHubOAuthenticator.refresh_user_hook = refresh_user_hook
  ```

helm-charts/basehub/values.yaml
jnywong: Here's my login/logout strategy writeup: https://github.com/2i2c-org/infrastructure/pull/6032#issuecomment-2944180023.

> And if we do it conditionally, we'll need to force people to re-log in.

I think having it conditional to protect private GH team memberships is okay for now: information from `c.GitHubOAuthenticator.auth_state_groups_key` will only be used to populate JupyterHub groups. JupyterHub groups is currently only used for the `jupyterhub_groups_exporter`, which is not on the critical path for hub functionality such as profile list allowed groups, server launches.

Therefore the extra config in `05-profile-groups` should stay the same and depend on team memberships from the GH auth state and not JH group memberships for now.

helm-charts/basehub/values.yaml
jnywong: Actually, I have reverted that and separated this into its own extraConfig, since staff access needs to be *appended* to `allowed_organizations` specified in hub values files.

[Append staff access to allowed_organizations](https://github.com/2i2c-org/infrastructure/pull/6032/commits/4b6bfd1ec8820e56708441414d6e84d4cf9fd50a)

jnywong:
Had a chat with Yuvi on Slack yesterday and I am happy that merging this will not cause any destructive actions.

PR #6060, opened 2025-05-15
[openscapes] add shared writeable folder to workshop hub

ateucher:
This is based on #6043, to add a shared folder to the Openscapes workshop hub.

Thanks!


ateucher:
Thank you!

Issue #5315, opened 2024-11-06
[Platform Initiative] Associate users with groups for usage and costs monitoring


Gman0909:
# Description

We need to allow admins of hubs to see aggregated usage information based on what groups a user belongs to. This would allow them to have a better understanding of which groups are using how much resources, and advocate for better funding / cost recovery based on their own internal representations and needs.

Please review the insights in the original [Product Board initative](https://2i2c.productboard.com/roadmap/7947557-2i2c-roadmap/features/27195081/insights)

# Tasks

- Allow associating a single user with multiple groups they may belong to. This information must be externally managed, probably by whatever authentication provider is used. 
  - As a starting point for an MVP, let us use GitHub authentication, with groups indicated by membership in GitHub teams. But the implementation should be agnostic to the authentication mechanism used.
- Collect information on what kind of usage reports will be useful to admins. Our existing Grafana Dashboard may be used as a starting point.
- Add ability to filter based on group membership to usage reports grafana dashboards
- Find out how to handle the problem of overlapping groups.
  - For example, a user may be a member of both the statistics and physics departments. When they start a server and consume those resources, who should that be attributed to? What about storage, which is temporally different (it is 'used' as soon as a byte is written, whereas memory / cpu / GPU are only used when the user is actively using it). 

# Definition of Done

- [ ] The feature/service is technically complete
- [ ] It’s been tested with one or more users (if applicable)
- [ ] It’s been deployed to a production cluster
- [ ] It has been shown to be replicably deployable
- [ ] It is well documented
- [ ] It has been added to our product menu.
- [ ] Lessons learned have been documented

PR #6065, opened 2025-05-16
Add "User Group Diagnostics" Grafana dashboard

jnywong:
This PR adds a new "User Group Diagnostics" Grafana dashboard that complements the "User Diagnostics Dashboard" to show resource usage aggregated by user group.

Requires [jupyterhub-groups-exporter](https://github.com/2i2c-org/jupyterhub-groups-exporter) to be set up on the hub. 

For this to universally work across all our infrastructure, I decided to separate the user group dashboard entirely from user name based aggregation. This is because if I combined both user name and user group into one dashboard/PromQL query, then it would break the dashboard for hubs that do not have [jupyterhub-groups-exporter](https://github.com/2i2c-org/jupyterhub-groups-exporter) set up because the `jupyterhub_user_group_info` metric is unavailable to the PromQL. Therefore, if a hub does not have [jupyterhub-groups-exporter](https://github.com/2i2c-org/jupyterhub-groups-exporter) set up, then the "User Diagnostics" dashboard will work as normal but the "User Group Diagnostics" dashboard will show no data.

Ref: https://github.com/2i2c-org/infrastructure/issues/5983

### DIFF
--- a/grafana-dashboards/common.libsonnet
+++ b/grafana-dashboards/common.libsonnet
@@ -9,30 +9,56 @@ local ts = grafonnet.panel.timeSeries;
       var.datasource.new("infinity_datasource", "yesoreyeram-infinity-datasource")
       + var.datasource.generalOptions.showOnDashboard.withNothing()
     ,
+    prometheus:
+      var.datasource.new('PROMETHEUS_DS', 'prometheus')
+      + var.datasource.generalOptions.showOnDashboard.withValueOnly()
+    ,
+    // Limit namespaces to those that run a hub service    
     hub:
-      var.query.new(
-        "hub",
-        {
-          query: "",
-          queryType: "infinity",
-          infinityQuery: {
-            format: "table",
-            parser: "backend",
-            refId: "variable",
-            source: "url",
-            type: "json",
-            url: "http://aws-ce-grafana-backend.support.svc.cluster.local/hub-names?from=${__from:date}&to=${__to:date}",
-            url_options: {
-              data: "",
-              method: "GET"
-            },
-          },
-        },
-      )
-      + var.query.withDatasourceFromVariable(self.infinity_datasource)
-      + var.query.selectionOptions.withIncludeAll(value=true)
-      + var.query.generalOptions.showOnDashboard.withNothing()
-      + var.query.refresh.onTime()
+      var.query.new('hub')
+      + var.query.withDatasourceFromVariable(self.prometheus)
+      + var.query.selectionOptions.withMulti()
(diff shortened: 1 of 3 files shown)


jnywong:
Hey @GeorgianaElena ! Just requested a review from you to just check if you think this is okay to implement across all of our clusters.

I think it should be fine, I value your infra eng experience here to give a quick green/red light :)

jnywong:
Actually, not ready for review yet, just found a bug!

GeorgianaElena (commented):
Great work @jnywong <3 ! I didn't pay too much attention to the queries, so this review is more from an infra perspective.

If you look at https://github.com/2i2c-org/infrastructure/blob/a526d62af27874a2e80adbb491d40261ad0e01a6/deployer/commands/grafana/deploy_dashboards.py#L58-L72

The dashboards here are deployed on aws only and are categorized as cost-related dashboards.

In my opinion, ideally this work should be instead upstream, in `jupyterhub/grafana-dashboards`. I believe 2i2c uses that repo most intensively so we shouldn't be impacting lots of people if we were to merge this ourselves.

If you want to be on the safe side, we could also host them under 2i2c for a month lets say, validate that they are correct, then upstream them. In this time, we would also open an upstream issue about this intention.

If we are to go with option two, then we need to update the deployer command and the general structure of this directory because it looks like it's growing to be more than cost-related dashboards.

What do you think?

jnywong:
Your infra perspective is exactly what I needed :)

Yeah, eventually I would like to upstream this, but like you said, to be on the safe side we can validate and host this on the 2i2c side while we wait.

I'll make changes to this PR to update the deployer command then. Thanks!

jnywong:
Deployer updated, dashboards moved to [2i2c hosted repo](https://github.com/2i2c-org/grafana-dashboards) and [upstream PR](https://github.com/jupyterhub/grafana-dashboards/pull/149) waiting in draft 👍

GeorgianaElena:
this is perfect @jnywong ❤️ I feel comfortable merging your PR upstream as well (I have the rights) if you feel confident too. Just let me know, otherwise feel free to merge this one and maybe also open an internal 2i2c tracking issue to review+merge upstream PR after some testing on our infra to make sure we don't loose track of it.

I believe the biggest challenge is having the 2i2c repo diverge and having to maintain and contribute in two places.

jnywong:
Great, thank you @GeorgianaElena!

I agree, we do not want these two repos to diverge and I will let you know when https://github.com/2i2c-org/infrastructure/issues/5315 is ready to be merged and upstreamed once I am happy.

To not lose track, I have added this task to the DoD in the parent initiative: https://github.com/2i2c-org/infrastructure/issues/5315

Issue #5658, opened 2025-03-05
Rollout storage quotas to  catalystproject-africa


GeorgianaElena:
Hubs with persistent storage to migrate:
- [x] aibst
- [x] bhki
- [x] bon
- [x] kush
- [x] molerhealth
- [x] must
- [x] nm-aist
- [x] staging
- [x] uvri
- [x] wits

This should be 10 points if we were to follow the same rule 1 point per hub, but will make it 8 instead which is the closest size in our Fibonacci series.

yuvipanda:
I deleted old backups and this is done

PR #6073, opened 2025-05-19
Enable jupyterhub-home-nfs for catalystproject-africa

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/5658

### DIFF
--- a/config/clusters/catalystproject-africa/aibst.values.yaml
+++ b/config/clusters/catalystproject-africa/aibst.values.yaml
@@ -25,3 +25,6 @@ jupyterhub:
   singleuser:
     nodeSelector:
       2i2c/hub-name: aibst
+jupyterhub-home-nfs:
+  eks:
+    volumeId: vol-06a16472044aa6ebc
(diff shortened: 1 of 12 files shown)


yuvipanda:
I generated these tfvar values with the following script, which was fed the output of `df -d1` on the existing home directories:


```python

import sys
from textwrap import dedent
from math import ceil, log

df_output_file = sys.argv[1]

hubs = {}
with open(df_output_file) as f:
    for line in f:
        parts = line.split()
        hubs[parts[1][2:]] = int(parts[0])

print(hubs)

for hub, size_in_kb in hubs.items():
    # Size in gb, and round up everything under 1GB to 1GB
    size_in_gb = ceil(size_in_kb / 1024 / 1024)

    # round to nearest power of 2 'up'
    dest_size = pow(2, ceil(log(size_in_gb)/log(2)))

    # Round to nearest 'round'ish number
    print(dedent(f"""
    "{hub}" = {{
        size        = {dest_size}
        type        = "gp3"
        name_suffix = "{hub}"
        tags        = {{ "2i2c:hub-name": "{hub}" }}
    }},"""), end="")
```

PR #6077, opened 2025-05-20
[2i2c-jetstream2] Add jupyterhub-groups-exporter config

jnywong:
Ref: https://github.com/2i2c-org/infrastructure/issues/5983


jnywong:
Hmm, I think because this cluster is driven by openstack, then there is some issue with passing `auth_state` from GitHub to the hub pod.

jnywong:
@GeorgianaElena should I skip deploying the jupyterhub groups exporter to any OpenStack clusters?

jnywong:
I spoke with Georgiana today and actually I just need to enable the configs related to capturing `auth_state` for this hub!

PR #6096, opened 2025-05-21
add writeable shared directory for users

eeholmes:
I don't how to fix the ownership however in our main hub. The workshop hub yaml has this line
https://github.com/2i2c-org/infrastructure/blob/e56e9680767a9785702ac7a95ed738c30067595f/config/clusters/nmfs-openscapes/workshop.values.yaml#L44-L49

But the main hub does not have this line.

### DIFF
--- a/config/clusters/nmfs-openscapes/common.values.yaml
+++ b/config/clusters/nmfs-openscapes/common.values.yaml
@@ -69,6 +69,13 @@ jupyterhub:
     cloudMetadata:
       blockWithIptables: false
     defaultUrl: /lab
+    storage:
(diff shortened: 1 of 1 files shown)


GeorgianaElena (approved):
Hey @eeholmes! I can help with this, but just to make sure I understand what you want to achieve: 
Do you want a `shared-public` directory that is writable for _all hubs_ in `nmfs-openscapes` cluster?

GeorgianaElena:
@eeholmes, I tackled this in https://github.com/2i2c-org/infrastructure/pull/6206

PR #6102, opened 2025-05-23
MAAP staging: Define group shared folders volumes and volume mounts as dict

sunu:
~~Trying out suggestions from https://github.com/2i2c-org/infrastructure/pull/6010#discussion_r2100847326~~

~~Haven't tried this out live yet, because I'm having some trouble logging in to the MAAP staging hub.~~

~~But if I understand correctly, we will run into issues without upstream changes because of how [`extraFiles` is handled](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/c3aa0758a8f436a45cb1ea510bcfa415ffec4c3e/jupyterhub/files/hub/jupyterhub_config.py#L326). This expects the `KubeSpawner.volumes` and `KubeSpawner.volume_mounts` to be lists. Similarly [`extraVolumes` / `extraVolumeMounts` handling](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/fa6414fc50bd88ff9725cc1228a6c8f177dbd5c3/jupyterhub/files/hub/jupyterhub_config.py#L339-L343) also expects `KubeSpawner.volumes` and `KubeSpawner.volume_mounts` to be lists.~~

I have tried this out on MAAP staging and it seems to work as expected.

related to https://github.com/NASA-IMPACT/veda-jupyterhub/issues/66

---
## Update

z2jh version 4.3.0 now supports dict based values for all supported fields and this PR now takes advantage of that. I've tested this on MAAP staging hub manually. But to merge this in, we need to wait for https://github.com/2i2c-org/infrastructure/issues/6915 to roll out z2jh 4.3.0 for all hubs.


config/clusters/maap/staging.values.yaml
```diff
@@ -229,6 +233,93 @@ jupyterhub:
...
+            name: home
+            mountPath: /home/jovyan/shared
+            subPath: _shared
+            readOnly: true
+          02-home-shared-group:
+            # overrides the root of the shared-group folder with an empty dir
+            # so that the user can't see the contents of other groups' folders
+            # that the user is not a member of
+            name: shared-group-placeholder
+            mountPath: /home/jovyan/shared-group
```
sunu, line 280: Not sure if this the best approach to fix visibility of stale group shared folders. Please let me know if there's a better way to do this.

The concern is: If a user previously had a shared group folder mounted (e.g., to `/home/jovyan/shared-group/<group-folder>`), their underlying home PVC (`<home-pvc>/<user>/)` might now contain a persistent directory at `shared-group/<group-folder>` that was created by Kubernetes as a mount point.

Later, if the user is no longer in that group, the specific mount for the actual shared data is removed. However, the `shared-group/<group-folder>` directory within their personal PVC home will still be visible (as part of the main `/home/jovyan` mount). This can lead to confusion, showing outdated or empty folders for groups they no longer have access to.

sunu:
The test is failing because the jupyterhub chart expects `singleuser.storage` [to be defined as an object](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/6cadee8060c55692b3d831d1d10ab06e04e418e2/jupyterhub/values.schema.yaml#L2355) but we are intentionally setting it to null.

yuvipanda:
@sunu hmm, can this be intentionally set to an empty object or something else to test?

sunu:
@yuvipanda yea, my bad, I misread the schema.  `singleuser.storage.type` can be set as `none` as a string.

sunu:
I've updated the PR to take advantage of support for dict values in z2jh 4.3.0. Converting to draft so that we don't merge it before https://github.com/2i2c-org/infrastructure/issues/6915 is fixed and z2jh 4.3.0 is rolled out across all hubs.

GeorgianaElena (commented):
@sunu, I've tried to deployed this to another hub where we need to deploy per groups shared directories.

When using the latest z2jh 4.3.1, I'm getting the following warnings when coalescing the charts.
```
coalesce.go:223: warning: destination for basehub.jupyterhub.singleuser.storage.extraVolumeMounts is a table. Ignoring non-table value ([map[mountPath:/home/jovyan/shared name:home readOnly:true subPath:_shared] map[mountPath:/dev/shm name:dev-shm] map[mountPath:/home/rstudio name:home subPath:{escaped_username}] map[mountPath:/home/rstudio/shared name:home readOnly:true subPath:_shared]])
coalesce.go:223: warning: destination for basehub.jupyterhub.singleuser.storage.extraVolumes is a table. Ignoring non-table value ([map[emptyDir:map[medium:Memory] name:dev-shm]])
coalesce.go:220: warning: cannot overwrite table with non table for basehub.jupyterhub.singleuser.storage.extraVolumeMounts (map[])
coalesce.go:220: warning: cannot overwrite table with non table for basehub.jupyterhub.singleuser.storage.extraVolumes (map[])
coalesce.go:223: warning: destination for basehub.jupyterhub.singleuser.storage.extraVolumeMounts is a table. Ignoring non-table value ([map[mountPath:/home/jovyan/shared name:home readOnly:true subPath:_shared] map[mountPath:/dev/shm name:dev-shm] map[mountPath:/home/rstudio name:home subPath:{escaped_username}] map[mountPath:/home/rstudio/shared name:home readOnly:true subPath:_shared]])
coalesce.go:223: warning: destination for basehub.jupyterhub.singleuser.storage.extraVolumes is a table. Ignoring non-table value ([map[emptyDir:map[medium:Memory] name:dev-shm]])
```

I tracked this down as coming from https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/0a3b1bad94d3e0fd31c3d1e207588b7e81bda4aa/jupyterhub/values.yaml#L412-L413, where we default them to being the empty map.

And we're defining them as arrays in our sub-chart, in [basehub/values.yaml](https://github.com/2i2c-org/infrastructure/blob/417afaab5fcb55ba6c7e98d75735f6b961224a29/helm-charts/basehub/values.yaml#L754-L758). And even though we're re-defining as a map in the hub's own values.yaml file, the coalescence happens before the merge, so it still errors out. 

This means that the change in z2jh is actually a breaking change and it doesn't work for setups that still use arrays. Wondering is using `null` is acceptable to keep it backwards compatible.
Either way, we need to at least mark it as a breaking change in the changelog of z2jh I believe.

Cc-ing @agoose77 because we're reviewed https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/3690 together and haven't catch this.

agoose77:
> I am confused. Isn't it an issue that in z2jh schema we say that extraVolumeMounts can either be an array or an obj, but then when trying to use it as an array in child charts makes the config to be completely ignored?

I missed the z2jh chart values.yaml. Yes!

GeorgianaElena:
I don't think there are any group shared dirs in maap anymore, so closing this

PR #6108, opened 2025-05-23
[nasa-cryo] Add rtc-access role to sub-team

jnywong:
See https://jupyterhub.readthedocs.io/en/stable/tutorial/collaboration-users.html

Ref: https://2i2c.freshdesk.com/a/tickets/3295

### DIFF
--- a/config/clusters/nasa-cryo/common.values.yaml
+++ b/config/clusters/nasa-cryo/common.values.yaml
@@ -78,6 +78,14 @@ basehub:
       services:
         jupyterhub-groups-exporter: {}
       loadRoles:
+        rtc-access:
+          scopes:
+            - access:servers!group=CryoInTheCloud:csdms-2025-workshop
+            - admin:servers!group=CryoInTheCloud:csdms-2025-workshop
+            - admin-ui
+            - list:groups!group=CryoInTheCloud:csdms-2025-workshop
+          groups:
+            - CryoInTheCloud:csdms-2025-workshop
         jupyterhub-groups-exporter:
           services:
             - jupyterhub-groups-exporter


mfisher87:
@jnywong Does this allow access to other users' servers without token? Or is a token still required?

Once @tsnow03 approves, can this fix be applied to the whole hub?

jnywong:
Ah this solution allows access without token, and hence everyone within the sub-team can access each others' servers.

I can see why the token method is preferable if you want to enable for the whole hub though! Will look into that, in theory it should work (https://jupyterhub.readthedocs.io/en/stable/tutorial/sharing.html#issue-a-sharing-code) but will take me a bit longer to double-check all is working as intended.

tsnow03:
Yes go for it!

jnywong:
Okay, so adding the scope `shares!user` to the default user role is all that's needed to [enable sharing with access codes](https://jupyterhub.readthedocs.io/en/stable/reference/sharing.html#enable-sharing) for everyone.

However, you have to generate an API token to talk to the hub API to generate and [share via invitation code](https://jupyterhub.readthedocs.io/en/stable/reference/sharing.html#share-via-invitation-code). See a quick video demo below:


https://github.com/user-attachments/assets/bd783f37-b119-498a-810a-0f5ca9cbecaf

jnywong:
Yes, every user needs to do this, since the "include token" UI in JupyterLab does not work with our JupyterHub infra set up on Kubernetes unfortunately. See https://github.com/jupyterlab/jupyter-collaboration/pull/438 for the current development around the improvement you seek!

mfisher87:
Thank you, Jenny :bow:

mfisher87:
@jnywong one more question:  Does this method enable hub members to share their session with users who are **not** hub members? If not, is this feasible?

jnywong:
I am afraid not: to access any server resource on CryoCloud, you have to authenticate against the hub (through an authorised GitHub account).

The ability to grant authorisation to the hub via an access code in JupyterHub doesn't exist presently.

jnywong:
Not that I am aware of! Please open away :)

mfisher87:
Thanks again, @jnywong :) I think this would be an issue against https://github.com/jupyterhub/jupyterhub, correct? I'm not sure I have the right words for this yet, would you be able to help out with that once I open the issue?

jnywong:
Yes, that would be a good place to start

mfisher87:
Thanks again :) https://github.com/jupyterhub/jupyterhub/issues/5078

mfisher87:
@jnywong I think something about our config isn't allowing the new jupyter-collaboration share UI to display (testing on `ghcr.io/geojupyter/geojupyter:21bdbc2698c4 ` which includes `jupyter-collaboration==4.1.0rc0`, I see the standard share UI).

Based on [this post](https://github.com/jupyterlab/jupyter-collaboration/pull/438#issuecomment-2919010411) do we need to define `extraConfig.customLabScopesOptions`? :pray:

PR #6115, opened 2025-05-24
Change how we autoformat YAML

yuvipanda:
As I was doing some of the jupyterhub-home-directory migrations, one thing that stuck out to me was the constant need to write scripts that manipulated our YAML config files. We have a *lot* of YAML config files, which are similarly structured. And humans err, so should only be doing things that require human judgement (also so we don't get bored out of our minds). Writing scripts to make these config changes is the perfect way to do reduce errors, reduce total toil our engineers make and have everything happen much faster than otherwise.

To be able to write scripts that manipulate our YAML files at scale, we needed a way to use a round-trip YAML parser that can manipulate the config file but preserve comments, as our YAML files are still written by humans.

ruamel.yaml is an excellent round trip YAML parser, and we already use it. However, we were also using using [prettier](https://github.com/pre-commit/mirrors-prettier) to autoformat YAML. So using ruamel.yaml to make refactorings meant that we will *not* produce clean git diffs - as ruamel.yaml has different ideas on how some indents should work than prettier does. Prettier's pre-commit integration has also been broken for a while, and we would be forever stuck in version 3.

Since the goal of autoformatting is to provide *a* consistent style, regardless of what style that is, adopting the ruamel.yaml style brings with it the power of being able to do mass refactorings.

So I timeboxed myself to a couple hours, and wrote a simple autoformatter that uses ruamel.yaml to autoformat YAML files: https://github.com/yuvipanda/ruamelfmt!

This PR switches us to use that, and pre-commit.ci will make a big fixer commit on top. While this adds some noise, it brings with it the utterly mass benefit of being able to write scripts that automate changes in our config files.

This also:

1. removes markdown autoformatting. We can re-add that if we find it necessary via a non-prettier place.
2. Removes an unused .yamllint file - we were not using yamllint
3. Fix a duplicated anchor that this pre-commit hook caught that the old version of prettier did not catch

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,3 +1,4 @@
 # Ignore these revs while doing git blame, as they are autoformats
 # Run `git config blame.ignoreRevsFile .git-blame-ignore-revs`
 b59cf3bb16b61531ec4620737f61a74b0f47d7a5
+6f3886ba818bccbe33d700cab34b51bd65552b6e


yuvipanda:
I don't wanna cause a lot of merge conflicts so I'll wait a little bit

yuvipanda:
We can use https://medium.com/@ramunarasinga/git-blame-ignore-revs-to-ignore-bulk-formatting-changes-f20ac23e6155 to ignore the pre-commit commit for blame :)

yuvipanda:
https://github.com/2i2c-org/infrastructure/pull/6121 is very related in theme

PR #6121, opened 2025-05-24
Cleanup overuse of procedural programming techniques in our deployer

yuvipanda:
As we manage more and more clusters, it becomes important for us to make our work more automatable. We have to build infrastructure to make managing our infrastructure more easier, or we will be forever stuck scaling linearly, which increases toil and is not sustainable.

As part of that, we should be able to write and use automation that help with our day to day tasks faster. Since Clusters and Hubs are what we deal with on a day to day basis, clearer representations of that helps with this task a *lot*.

As our codebase has grown over time, we have added utility methods in various places to do specific things. However, these are hard to find, and hard to keep together. Plus, they are not benefiting from some of the helpful features of Python that help manage complexity. In particular, in this PR, I refactor two patterns:

1. Repeated class construction using utility methods should be using class methods instead. A common pattern here was:

   1. Use a utility method to find a particular path
   2. Load this path into memory
   3. Construct a particular object by calling a class's constructor using what we loaded onto memory.

   This happens so often that we have a python feature to address this -
   class methods! This PR adds a `from_name` classmethod to our `Cluster`
   class, and replaces all repeated usage with that class method. This
   makes life much easier to use Cluster objects everywhere.

2. Utility methods that take as their first parameter information that is then used to construct an object, and then do something with that object. This is what a regular python method is to be used for - Python makes this quite explicit with the `self` being explicit. Makes for much better discovery and containment.

   This PR moves such utility methods for dealing with prometheus and grafana into the Cluster object itself, so it can be reused elsewhere. This was what originally triggered this work - as part of the home directory migration, I wanted to build a deployer command to easily execute promql from the commandline. Even though all the code required to do that was already present, it was scattered and difficult to use. This simplifies that process.

I want us to continue looking through this with an eye towards making automation easier, so we can make our lives easier. Otherwise we will all eventually simply drown in toil and nobody wants that.

A few additional things this PR does:

1. Use `/` to join `Paths` instead of `.joinpath`. Using `/` is a python feature, and not something we need to be afraid of but to learn. Let's use it wherever we are constructing paths.
2. Add some more type hints. We can do this more systematically to get better type checking, but not a priority.
3. Removes some utility methods that were either called only once, or were at most 2 lines long. A utility method adds a layer of indirection and abstraction, and the extra cognitive load of the indirection must be weighed against the benefits the abstraction provides. This is of course subjective, and that's perfectly ok.
4. Added a `dir_path` field to `Cluster` as a lot of code is finding other files from the directory containing the cluster config. Remember we control the whole codebase, and can do things like this to make our lives easier.

You can also check out the following short resources that informed my choices in this PR:

1. https://www.youtube.com/watch?v=o9pEzgHorH0 which goes through places where the utility method pattern is better, but also where classes are better.
2. https://blog.glyph.im/2025/04/stop-writing-init-methods.html a newer post that I'm still slowly digesting, but has definitely convinced me to use classmethods more as the canonical way to construct objects.

Things I have intentionally stayed away from in this PR:

1. Make invalid states unrepresentable (https://hugotunius.se/2020/05/16/making-invalid-state-unrepresentable.html and many other posts describe this). Our validation logic is sprawled around in a lot of functions as well, making it hard to reason about and duplicitous. That should all be handled in the objects themselves. We may also use dataclasses here.
2. Refactor `decision.py`. An important trade-off when writing code is to make choices about what kinds of complexity to introduce, proportionate to the essential complexity of the problem being solved. This code takes as input a list of files changed and produces as output a list of hubs and clusters to be upgraded. However, the very many layers of indirection (spotted by functions that are often called only once) and reliance on mocks of our own code for testing introduce a lot of accidental complexity here. We should refactor this to make it simpler as a pure function, and also convert the tests to use [table driven testing](https://go.dev/wiki/TableDrivenTests)
3. Get rid of the billing module here. I don't believe it's currently used at all.
4. Centralize all path and encryption management into the `Cluster` and `Hub` classes. Currently there's far too much path management going on all over the place, making things hard to reason about and repeated.


agoose77:
I'm a huge fan of 
> Make invalid states unrepresentable 

It's one of the biggest safety principles we can employ. It will be fun to get around to that!

PR #6125, opened 2025-05-26
[awi-ciroh] Add option to ngen 2i2c image

benlee0423:


arpita0911patel (commented):
Possible to change name of the image to NextGen National Water Model(NWM)

PR #6142, opened 2025-05-29
[deployer] Fix quotes in f-string

GeorgianaElena:


yuvipanda:
@GeorgianaElena nah, no problem to have it this way! But I think when you have time upgrading locally to also 3.13 is a good idea.

yuvipanda:
Just as a follow on from the conversation we just had, I think it's fine given we only really have 4-5 people who need to use this code :)

PR #6147, opened 2025-05-30
Increase the Binder Ram allocation

wildintellect:
Some examples require significantly more memory than 2GB to even attempt. For at least 1 demo in June we need to try with more ram. Can discuss with user later if we can turn it back down.

Is this the only place that needs to be changed to allow the Binder launched instances to have more ram?

@batpad @yuvipanda @GeorgianaElena


yuvipanda:
This is ok for testing, but when you do the demo, you should probably increase the limit as well. We also have capped the number of instances, and that may need to be raised too based on what this number ends up being @wildintellect

Issue #6176, opened 2025-06-06
Add an alert for jupyterhub-groups-exporter pod dying


yuvipanda:
We want to monitor our new code for errors, and the simplest way is to have an alert that fires to pagerduty when the pod restarts. We should make this a policy for everything we deploy, so we aren't 'surprised' by bugs in there.

We should *unconditionally roll this out* to all clusters, even if we eventually choose to not deploy this pod on all clusters. The overhead for the alert is low.

## Definition of done

- [x] A prometheus alert is rolled out
- [x] Very minimal documentation is written on what we do when this alert fires (as this is not an emergency). I suggest that the response is to look at the logs, file an issue as appropriate, and flag that for inclusion in the next sprint as something that must be fixed.

## References

https://stackoverflow.com/questions/41452930/how-can-i-alert-for-container-restarted has some promql that may help

PR #6179, opened 2025-06-09
Add pod restart alert for pod names containing groups-exporter substring

jnywong:
Would like some guidance on how i can test this alert without merging into `main`.

Closes https://github.com/2i2c-org/infrastructure/issues/6176

### DIFF
--- a/docs/howto/manage-alerts/index.md
+++ b/docs/howto/manage-alerts/index.md
@@ -1,0 +1,40 @@
+# Managing alerts configured with Jsonnet
+
+In addition to [](#uptime-checks), we also have a set of alerts that are configured in support deployments using [](#topic/jsonnet).
+
+## Configuration
+
+We use the [Prometheus alert manager](https://prometheus.io/docs/alerting/latest/overview/) to set up alerts that are defined in the `helm-charts/support/values.jsonnet` file.
+
+At the time of writing, we have the following classes of alerts:
+
+1. when a persistent volume claim (PVC) is approaching full capacity
+2. when a pod has restarted in the last hour.
+
+When an alert threshold is crossed, an automatic notification is sent to PagerDuty and the `#pagerduty-notifications` channel on the 2i2c Slack.
+
+## When a PVC is approaching full capacity
+
+We monitor the capacity of the following volumes:
+
+- home directories
+- hub database
+- prometheus database
+
+The alert is triggered when the volume is more than 90% full.
+
+To resolve the alert, follow the guidance below
+
+- [](../../howto/filesystem-management/increase-disk-size.md)
+- **TODO: add instructions for hub database**
+- [](../../sre-guide/prometheus-disk-resize.md)
+
+## When a pod has restarted in the last hour
+
+We monitor pod restarts for the following services:
+
+- `jupyterhub-groups-exporter`
+
+If a pod has restarted in the last hour, it may indicate an issue with the service or its configuration. Check the logs of the pod to identify any errors or issues that may have caused the restart. If necessary, redeploy the service or adjust its configuration to resolve the issue.
+
+If you have taken the above actions and the issue persists, then open a GitHub issue capturing the details of the problem for consideration by the wider 2i2c team.
--- a/docs/howto/manage-alerts/index.md
+++ b/docs/howto/manage-alerts/index.md
@@ -9,7 +9,7 @@ We use the [Prometheus alert manager](https://prometheus.io/docs/alerting/latest
 At the time of writing, we have the following classes of alerts:
 
 1. when a persistent volume claim (PVC) is approaching full capacity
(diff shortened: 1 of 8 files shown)


docs/howto/manage-alerts/index.md
```diff
@@ -0,0 +1,40 @@
...
+- home directories
+- hub database
+- prometheus database
+
+The alert is triggered when the volume is more than 90% full.
+
+To resolve the alert, follow the guidance below
+
+- [](../../howto/filesystem-management/increase-disk-size.md)
+- **TODO: add instructions for hub database**
```
jnywong: Help needed: I am unsure if we have docs for increase the hub database disk is written down somewhere, and what the procedure is if not.

docs/howto/manage-alerts/index.md
```diff
@@ -0,0 +1,40 @@
...
+
+When an alert threshold is crossed, an automatic notification is sent to PagerDuty and the `#pagerduty-notifications` channel on the 2i2c Slack.
+
+## When a PVC is approaching full capacity
+
+We monitor the capacity of the following volumes:
+
+- home directories
+- hub database
+- prometheus database
```
agoose77, line 22:
```suggestion
- Prometheus database
```

jnywong: Cheers pal

docs/howto/manage-alerts/index.md
```diff
@@ -0,0 +1,40 @@
+# Managing alerts configured with Jsonnet
```
jnywong:
```suggestion
# Manage alerts configured with Jsonnet
```

helm-charts/support/values.jsonnet
```diff
@@ -36,6 +36,30 @@ local makePVCApproachingFullAlert = function(
...
+  summary,
+  pod_name_substring,
+                            ) {
+  name: name,
+  rules: [
+    {
+      alert: name,
+      expr: |||
+        # Count container restarts with pod name containing 'pod_name_substring' in 1 hour.
+        sum(rate(kube_pod_container_status_restarts_total{pod=~".*%s.*"}[1h])) by (namespace) * 3600 >= 1
```
yuvipanda: This query gets re-evaluated every minute or so (configurable somewhere). I *think*, as written, it looks for '1 restart in the last hour'. I *think* this means this query will evaluate to *false* after an hour has passed, no matter if we did anything about it or not. This will potentially lead to 'flapping' alerts, that come on and go off without us having to do anything, and hence tend to get ignored.

Given that `kube_pod_container_status_restarts_total` is a [Counter](https://prometheus.io/docs/concepts/metric_types/#counter) that will monotonically increase *but* stop existing once the pod goes away (because it is per pod - each 'prometheus metric' is a combination of the metric name and all its labels), I think we can simply alert on `kube_pod_container_status_restarts_total{pod=~".*%s.*"} > 1`. To 'clear' it, whoever investigating must then open an issue, and then manually delete the pod.

The easiest way to trigger a restart to check is to set the memory limit to be lower than what observed usage is :D

docs/howto/manage-alerts/index.md
jnywong: @GeorgianaElena would you be able to point me in the right direction here?

yuvipanda: @jnywong I don't think we have specific instructions, I'd recommend creating an issue here, linking to this and moving forward. Also mostly if the hub db disk runs out of disk space, it's almost always because of log files - so mostly we'll have to trim the log file

docs/howto/manage-alerts/index.md
```diff
@@ -0,0 +1,48 @@
...
+- **TODO: add instructions for hub database**
+- [](../../sre-guide/prometheus-disk-resize.md)
+
+## When a pod has restarted
+
+We monitor pod restarts for the following services:
+
+- `jupyterhub-groups-exporter`
+
+If a pod has restarted, it may indicate an issue with the service or its configuration. Check the logs of the pod to identify any errors or issues that may have caused the restart. If necessary, redeploy the service or adjust its configuration to resolve the issue.
```
yuvipanda:
```suggestion
If a pod has restarted, it may indicate an issue with the service or its configuration. Check the logs of the pod to identify any errors or issues that may have caused the restart. If necessary, add more error handling to the code, redeploy the service or adjust its configuration to resolve the issue.
```

helm-charts/support/values.jsonnet
jnywong: As suggested, I've updated the query to just see whether `kube_pod_container_status_restarts_total{pod=~".*%s.*"} >= 1`.

I tested this by deploying manually on the `2i2c-aws-us` cluster and verified the alert is behaving as expected.

Issue #6177, opened 2025-06-06
Set resource requests and limits appropriately for jupyterhub-groups-exporter


yuvipanda:
All the infrastructure pods we run must have appropriate limits and requests set.

We should observe usage on a couple of clusters, and then use that to set these values everywhere.

## Definition of done

- [ ] Memory requests and limits set for the jupyterhub-groups-exporter pod
- [ ] CPU requests and limits set for the jupyterhub-groups-exporter pod

jnywong:
Ran the following bash script to use `deployer exec promql` (thanks for this new addition!) to get memory and cpu usage across all clusters:

```bash
#!/bin/bash

export FILE="cluster_list.txt"

arr_memory=()
arr_cpu=()

for i in $(cat ${FILE}); do
  echo "Processing cluster: $i"

  mem=$(deployer exec promql $i "max(container_memory_working_set_bytes{name!='', pod=~'.*groups-exporter.*'})" --json  | jq -r ".[0].value")
  arr_memory+=("$mem")

  cpu=$(deployer exec promql $i "max(irate(container_cpu_usage_seconds_total{name!='', pod=~'.*groups-exporter.*'}[5m]))" --json  | jq -r ".[0].value")
  arr_cpu+=("$cpu")
done
```

Getting the max values for memory and CPU usage returned

```bash
Max memory: 40849408 
Max CPU: 0.0002463868161359514
```

PR #6181, opened 2025-06-09
Set jupyterhub-groups-exporter resource requests and limits

jnywong:
Closes https://github.com/2i2c-org/infrastructure/issues/6177


helm-charts/basehub/values.yaml
```diff
@@ -1760,3 +1760,10 @@ jupyterhub-home-nfs:
 
 jupyterhub-groups-exporter:
   enabled: true
+  resources:
+    requests:
+      cpu: 0.001
+      memory: 64Mi
```
yuvipanda: Add a comment with the promql query used to determine this?

jnywong: Added in the comments of the jsonnet file https://github.com/2i2c-org/infrastructure/pull/6181/commits/77bb53f340774990c722bf6e3e1da44a31cb840b

jnywong:
Cool! Manually deployed and used `kubectl describe` to verify that the new resource settings were applied to the pod.

PR #6189, opened 2025-06-10
feat: update node types in Project Pythia BinderHub on Jetstream2

agoose77:
I didn't touch the staging Jetstream2 hub — I am not sure where the infrastructure-as-code is for that. My hunch is that we have imperatively built that configuration, rather than e.g. using Terraform.

@GeorgianaElena can you help me figure out a couple of questions?

1. I'm guessing that the `capi.stackhpc.com/node-group` label is set by the infrastructure. How does it normalise from the actual node-group name, which contains dots? i.e. from `"user-${each.key}"` where I assume that `${each.key}` contains dots, how do we end up with `"user-m3-large"` not `"user-m3.large"`?
2. Am I correct that the staging hub doesn't have infra-as-code in this repo?

Thanks!

### DIFF
--- a/config/clusters/2i2c-jetstream2/staging.values.yaml
+++ b/config/clusters/2i2c-jetstream2/staging.values.yaml
@@ -36,7 +36,7 @@ jupyterhub:
   prePuller:
     hook:
       nodeSelector:
-        capi.stackhpc.com/node-group: user-m3-large
+        capi.stackhpc.com/node-group: user-m3-quad
   ingress:
     hosts: [staging.js.2i2c.cloud]
     tls:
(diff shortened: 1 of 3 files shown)


agoose77:
@jmunroe looking at https://docs.jetstream-cloud.org/general/instance-flavors/#jetstream2-cpu, the cost-per-hour for the `m3.large` is four-times greater than the current `m3.quad`.

Based upon reading the infrastructure configuration, I believe that there will always be at _least_ one node running at all times (the node group scales down to one node). As such, this change will directly affect the baseline cost overhead in addition to per-node-group cost scaling. Is this OK? 

@GeorgianaElena could you perhaps confirm my understanding here, too!

jmunroe:
Yes. That is understood by Project Pythia. Please proceed.

PR #6201, opened 2025-06-11
Add Nix flake for the development environment

agoose77:
This PR adds a Nix flake that I am using to maintain this development environment locally. I spoke to @yuvipanda about the possibility of including this, given that I am using it to set up this environment on multiple machines.

To use this flake:
```shell
nix develop .
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,5 +1,5 @@
 # code-owner: @agoose77
-# This flake sets up an FSH dev-shell that installs all the required 
+# This flake sets up an FSH dev-shell that installs all the required
 # packages for running deployer, and then installs the tool in the virtual environment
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -0,0 +1,61 @@
...
+                ${pkgs.python3.interpreter} -m venv .venv
+                source .venv/bin/activate
+                python -m pip install -e .[dev]
+              else
+                source .venv/bin/activate
+              fi
+              set +e
+            ''
+            + script)}/bin/runScript";
+        })
```
agoose77, line 56: This is lazy. It's better to provision a Nix env with those packages installed from nixpkgs where possible, but perfect is the enemy of good enough.

PR #6207, opened 2025-06-12
[docs] Increase disk size: add check-in step

jnywong:
See [Slack thread](https://2i2c.slack.com/archives/C041E05LVHB/p1749744638158989) for context.


GeorgianaElena (approved):
❤️ Thank you!

PR #6215, opened 2025-06-13
fix: label eksctl autoscaler nodes with GPU accelerator labels

agoose77:
We have observed that the autoscaler needs restarting for clusters with GPUs, in order to fix a bug that is preventing pods requesting GPUs from triggering a scale-up.

This PR applies a possible fix suggested by @yuvipanda, to add proper labels for the autoscaler.

I generated this change with
```shell
sd "^(\s*)('2i2c/has-gpu': 'true',)" $'${1}${2}\n${1}\'k8s.amazonaws.com/accelerator\': \'nvidia-tesla-t4\',' eksctl/*.jsonnet
```

I determined the accelerator name by starting a GPU job on the showcase hub, and running the snippet given in https://github.com/NVIDIA/k8s-device-plugin/issues/294#issue-1154321965, 
which looks correct given the form in https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider/aws#special-note-on-gpu-instances

I am surprised that there's no canonical list of GPU label names.

- [x] disasters
- [x] smithsonian
- [x] nasa-cryo
- [x] maap
- [x] projectpythia
- [x] nasa-veda
- [x] nmfs-openscapes

> [!Warning]
> I couldn't touch `victor`, so I'm merging this as-is for the non-victor clusters such that our configuration matches what's actually in deployment.

### DIFF
--- a/docs/howto/features/gpu.md
+++ b/docs/howto/features/gpu.md
@@ -106,6 +106,7 @@ series nodes.
 We use `eksctl` with `jsonnet` to provision our kubernetes clusters on
 AWS, and we can configure a node group there to provide us GPUs.
 
+
 1. In the `notebookNodes` definition in the appropriate `.jsonnet` file,
    add a node definition for the appropriate GPU node type:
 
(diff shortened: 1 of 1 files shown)


agoose77:
I am pausing for today — I am waiting for the GPU node group to clear for `nmfs-openscapes`.

PR #6236, opened 2025-06-17
profile options for SSIM GHG Workshop 2025

batpad:
Makes a separate Profile Option restricted to the SSIM GHG 2025 Workshop participants' group.

**Desired Outcome**: Existing groups / users see the existing profile options. Members of the `ssim-ghg-2025` group see only the profile option specific for the workshop: with the custom workshop image and resource allocations set at a fixed ~60GB RAM.

**Questions**: Currently, I have disabled dynamic image building and the unlisted image choice for the workshop group - they will only be able to use the defined image.

 - Should users be allowed dynamic image building and / or specifying an arbitrary image with the `unlisted_choice` option?
 - Should we continue to give options for the existing images like Pangeo, Rocker, QGIS?
 - Currently, we are also pulling in the ghgc-docs folder for the Workshop users as part of the init scripts - do we want to skip doing that for workshop participants?

All the above things are just configuration, so happy to implement things either way if you let us know what's best @wildintellect @AgilentGCMS 

cc @sunu

### DIFF
--- a/config/clusters/nasa-ghg/common.values.yaml
+++ b/config/clusters/nasa-ghg/common.values.yaml
@@ -99,6 +99,14 @@ basehub:
       profileList:
       - display_name: Choose environment and resources
         default: true
+        allowed_groups:
+          - US-GHG-Center:ghgc-hub-access
+          - US-GHG-Center:ghg-use-case-1
+          - US-GHG-Center:ghg-use-case-2
+          - US-GHG-Center:ghg-use-case-3
+          - US-GHG-Center:ghg-external-collaborators
+          - US-GHG-Center:ghg-workshop-access
+          - US-GHG-Center:ghg-trial-access        
         profile_options:
           image:
             display_name: Environment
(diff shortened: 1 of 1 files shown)


AgilentGCMS:
> **Questions**: Currently, I have disabled dynamic image building and the unlisted image choice for the workshop group - they will only be able to use the defined image.
> 
> * Should users be allowed dynamic image building and / or specifying an arbitrary image with the `unlisted_choice` option?

I would say no. For the workshop, let's make it so that the participants only see one option. If you want to add a "hidden" mechanism for building custom images, that's fine, but let's not make it an option they can select easily, e.g., from a drop-down list.

> * Should we continue to give options for the existing images like Pangeo, Rocker, QGIS?

Preferably not. Since the custom image works well for all the examples, let's just stick to that.

> * Currently, we are also pulling in the ghgc-docs folder for the Workshop users as part of the init scripts - do we want to skip doing that for workshop participants?

I do not understand this, what is the ghgc-docs folder?

config/clusters/nasa-ghg/common.values.yaml
```diff
@@ -221,7 +229,53 @@ basehub:
                   cpu_limit: 15.695
                   node_selector:
                     node.kubernetes.io/instance-type: r5.4xlarge
-
+      - display_name: SSIM GHG Workshop 2025
+        default: true
```
sunu: This won't matter in practice because of the order but we shouldn't mark this as default or have more than one default profiles in `profileList`

sunu (approved):
Added a minor comment but otherwise looks good to me!

config/clusters/nasa-ghg/common.values.yaml
batpad: oh that's a very good catch and I missed that copying things over. Ty!

I think it's better to remove it - some people will have access to both the workshop group as well as other groups, so in that case would see two profiles. I'll remove the `default` there.

GeorgianaElena (commented):
This looks good to me @batpad! I expected to see a workshop hub and was concerned about this modifying `common.values.yaml`, but this is meant to be used in the prod hub, right?

Do you want to me to first deploy it to staging so you can test it out before going to prod?

config/clusters/nasa-ghg/common.values.yaml
```diff
@@ -99,6 +99,14 @@ basehub:
...
       - display_name: Choose environment and resources
         default: true
+        allowed_groups:
+        - US-GHG-Center:ghgc-hub-access
+        - US-GHG-Center:ghg-use-case-1
+        - US-GHG-Center:ghg-use-case-2
+        - US-GHG-Center:ghg-use-case-3
+        - US-GHG-Center:ghg-external-collaborators
+        - US-GHG-Center:ghg-workshop-access
+        - US-GHG-Center:ghg-trial-access
```
GeorgianaElena, line 109:
```suggestion
        - US-GHG-Center:ghg-trial-access
        - 2i2c-org:hub-access-for-2i2c-staff
```

config/clusters/nasa-ghg/common.values.yaml
```diff
@@ -221,7 +229,52 @@ basehub:
                   cpu_limit: 15.695
                   node_selector:
                     node.kubernetes.io/instance-type: r5.4xlarge
-
+      - display_name: SSIM GHG Workshop 2025
+        allowed_groups:
+        - US-GHG-Center:ssim-ghg-2025
```
GeorgianaElena, line 235:
```suggestion
        - US-GHG-Center:ssim-ghg-2025
        - 2i2c-org:hub-access-for-2i2c-staff
```

PR #6242, opened 2025-06-18
fix: apply GPU pre-labelling to victor config

agoose77:
A successor to #6215 for `victor`. 

## Deployments updated
- [ ] `victor`


agoose77:
I now have credentials! I've applied this configuration, and it can now be merged.

PR #6253, opened 2025-06-19
Change image to d624b28e39c4

tsnow03:
The old image has RTC that might be causing save and home directory bugs.


GeorgianaElena:
Thank you @tsnow03 ✨

Issue #6224, opened 2025-06-13
[New Hub] [Phase 3] Hub Setup - neurohackademy.2i2c.cloud


jmunroe:
### Context

https://github.com/2i2c-org/meta/issues/2317

### How many hubs will be deployed?

2

### Which cluster will the hub(s) be deployed on?

neurohackademy

### Hub Setup Information

For each hub to be deployed, copy the tables below and fill them in.
The tables must be complete for each hub to be considered READY to be
deployed. For a staging/prod pair, a statement such as "Same as staging
but for production" will suffice for the prod hub's specification. Use
the Notes column to provide any contextual information.

Available runbooks:
- https://infrastructure.2i2c.org/hub-deployment-guide/runbooks/phase3/

#### Phase 3.1: Initial setup

| Question | Answer | Notes |
| :--- | :--- | :--- |
| Name of the hub | staging | |
| Dask gateway? | `No` | |
| Splash image URL | https://user-images.githubusercontent.com/118582/178122521-aa5b83e7-f898-415a-89ff-dc288f599b4f.png | |
| Homepage URL | https://neurohackademy.org/ | | 
| Funded by? | The National Institutes of Health grant 2R25MH112480-06 | https://reporter.nih.gov/search/ydTvTwXxk0yd6eGdRznbLQ/project-details/10409452 |
| Authentication Mechanism | github | |
| Admin Users (GitHub handles or emails, depending on Mechanism) | arokem, noahbenson| |

At the end of this phase, both 2i2c engineers and the admin users mentioned can login to the hub.

#### Phase 3.2: Object storage access

| Question | Answer | Notes |
| :--- | :--- | :--- |
| Scratch bucket enabled? | `No` |  |
| Persistent bucket enabled? | `No` |  |
| Requester pays requests to external buckets allowed? | `NA` | |  

At the end of this phase, both 2i2c engineers and the admin users mentioned can access any object storage setup.

#### Phase 3.3: Profile List

These are the standard profile list options to start with. They can be customized later.
Use the Notes column to provide extra information, such as specific image tags to use, or `deployer generate resource-allocation choices` command to use for RAM, etc.

| Question | Answer | Notes |
| :--- | :--- | :--- |
| Pangeo Notebook?  | `No` | |
| RStudio (with Rocker)? | `Yes` |  |
| Allow users to specify any image they want to use? | `Yes` | If Yes, enable `unlisted_choice`. | 
| Max RAM option allowed | 14 GB? | |
[...]

PR #6290, opened 2025-06-27
[new hub] Neurohackademy staging and prod

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/6224


GeorgianaElena:
Assumptions:
- prod volume size is 50GB in total for all users, with a quota of 5GB/per user in prod
- in the profile list, you can combine every image with both machine types (gpu, r5.xlarge)
- it uses the last year's images, only the GitHub org that has access was updated

Issue #6308, opened 2025-07-01
[NASA-GHG][Workshop] Summer School Workshop July 8-18


batpad:
### Context

The GHG Summer School will be holding a workshop with about **35 participants** from July 8 - July 18. They ran a similar workshop last year, and most things seem smooth. Just creating a ticket here for visibility and for making sure we're fully prepared.

The instructors have created a custom image for the workshop and we've already deployed a custom profile for workshop participants that sets the default image and default resource requirements. **Note:** each container is allocated 61GB of RAM, so these are fairly large individual instances.

@yuvipanda has [increased the IOPS](https://github.com/2i2c-org/infrastructure/pull/6306) on the EBS volume used for the home folders to (hopefully) mitigate any issues with the shared disk.

The workshop will begin at 9am Mountain Time on July 8th (realistically, post-registration, etc. people will start using the hub likely around 10am MT). @aprilmj - would you be able to help advising on who they can reach out to just in case of any emergency situations or issues with the hub? 

### What we need to do

We're mostly setup, but there's a few questions -

 - It will be ~35 people each using instances with ~61GB RAM. Are we certain we won't face any issues with any arbitrary AWS limits or max nodepool sizes or anything that might prevent nodes from coming up? They seem __ok__ with the longer startup time when new nodes come online, so I don't think we need to preprovision, I just want to make sure we don't hit any limits.
 - They will be using local `/tmp` folders for some stuff. They are curious what the limit on the size of `/tmp` would be. They don't **need** more than 10GB per user, so I think there's enough space, but they would like to know what the limit is just in case.
 - Home folders / NFS: we should double-check total available size and what they need. (cc @wildintellect )

cc @aprilmj 

TODO:
 - [ ] Double check how containers are getting placed on nodes and ensure the placement is sane for the instance size we are using for the workshop. - @sunu may need your help on this - we essentially want to double-check that we aren't leaving a lot of room on instances, etc.

### Definition of Done

- [ ] Answers are communicated to workshop organizers
- [ ] Workshop runs smoothly

batpad:
Adding here: it seems folks will need more space in home directories. The current EBS volume size seems to be 200GB. I think we should plan for upto ~50GB per user, so we're looking at upping the EBS volume to about 1.8TB.

Sorry, this is a bit unfortunately and hopefully we can do things differently a bit next time - we will need to plan for some cleanup here post-workshop.

batpad:
[...]

PR #6309, opened 2025-07-01
GHG: tune the CPU request for SSIM workshop

sunu:
We tested spinning up multiple pods for the SSIM workshop. It always created a node per pod because the CPU availability was insufficient.

I'm decreasing the CPU request here a bit to test if that helps put 2 user pods in a single node for the workshop.

Changing the option for the non-workshop environment too just to help us test without being part of the workshop group.

refs https://github.com/2i2c-org/infrastructure/issues/6308


yuvipanda:
https://github.com/2i2c-org/infrastructure/pull/6313 is the right fix but we can deploy this as is for now

Issue #6213, opened 2025-06-13
[Spike 4h] Get the documentation and resource allocation script on par with current recommendations


GeorgianaElena:
### Context
This also branched out https://github.com/2i2c-org/infrastructure/issues/6105.

### What we need to do
The infra team wants to be more consistent in our approach to resource allocation. That will eventually lead how we can update the resource allocation script so that it's easily and deterministically usable as well as being straightforward to maintain. The spike represents the time we want to dedicate to conversation and agreement.

### Definition of Done
- [ ] We've agreed on resource allocation approach for different scenarios
- [ ] A new (refined) issue is opened about how v2 of the resource allocation script should look like more 
- [ ] We don't spend more than 4 hours of each person's time total on this.

PR #6313, opened 2025-07-02
Improve resource allocation choice generator

yuvipanda:
- In https://github.com/2i2c-org/infrastructure/pull/6309, we discovered that some of our resource allocations were now outdated, causing some servers to not spawn. We add a 2% wiggle room (clamped at 2GB / 1CPU) to work around this.
- Based on feedback from earthscope's @sarahwilson523 in https://github.com/2i2c-org/infrastructure/pull/6351#issuecomment-3071223299, we've updated the choice descriptions to be clearer to end users. No more fractions, better messaging around CPU guarantees and limits.
- Allow generating allocations for multiple instance types in one go, to reduce copy pasting

### DIFF
--- a/config/clusters/nasa-ghg/staging.values.yaml
+++ b/config/clusters/nasa-ghg/staging.values.yaml
@@ -23,7 +23,7 @@ basehub:
     custom:
       homepage:
         gitRepoBranch: staging
-        gitRepoUrl: "https://github.com/US-GHG-Center/ghgc-hub-homepage"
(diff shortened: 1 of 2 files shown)


GeorgianaElena:
@yuvipanda, I know that this could be improved/documented as part of https://github.com/2i2c-org/infrastructure/issues/6213, BUT since it looks we haven't had time to commit it these past sprints, can we merge this PR as it is so we can use it?

yuvipanda:
Fair enough, @GeorgianaElena - merged.

PR #6316, opened 2025-07-03
[nasa-ghg, prod] Increase EBS volume size

jnywong:
Ref https://github.com/2i2c-org/infrastructure/issues/6308

Also fixes [inline_policy deprecation warning](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_role#example-of-removing-inline-policies) and rules [filter warning](https://github.com/hashicorp/terraform-provider-aws/issues/42714)

### DIFF
--- a/terraform/aws/buckets.tf
+++ b/terraform/aws/buckets.tf
@@ -17,6 +17,10 @@ resource "aws_s3_bucket_lifecycle_configuration" "user_bucket_expiry" {
     expiration {
       days = each.value.delete_after
     }
+
+    filter {
+      prefix = ""
+    }
   }
 
   dynamic "rule" {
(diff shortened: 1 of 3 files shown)


terraform/aws/buckets.tf
```diff
@@ -17,6 +17,10 @@ resource "aws_s3_bucket_lifecycle_configuration" "user_bucket_expiry" {
     expiration {
       days = each.value.delete_after
     }
+
+    filter {
+      prefix = ""
+    }
```
filiperosa, line 23: Hi @jnywong, sorry for the intrusion, arrived here though the [Terraform GH Thread](https://github.com/hashicorp/terraform-provider-aws/issues/42714). 
Did this syntax work on silencing the warnings?

jnywong: Hey @filiperosa ! Yeah this change silenced the warning for us. The `filter {}` syntax did not work. We're currently on AWS Provider 5.89

filiperosa: Humm ok! I'm using the exact same syntax as you but it's still producing warnings on  AWS Provider **5.99.1**
Thank you!

Issue #6185, opened 2025-06-09
Add an alert for when a user pod doesn't trigger a scale up, signalling an autoscaler failure


yuvipanda:
https://2i2c.freshdesk.com/a/tickets/3449 and https://2i2c.freshdesk.com/a/tickets/3454 (and many other prior issues) could be automatically caught if we have an alert for the following conditions:

1. A user pod is created
2. But is unscheduled for a long period of time

This almost always signifies an infrastructure (than user) failure that must be investigated.

## Definition of done

- [ ] Alert described exists
- [ ] We have minimal docs on how to deal with this alert

Issue #2267, opened 2023-02-27
Alerts for consistent server start failures on a hub. If a JupyterHub is somehow failing to start user pods consistently, we should have an alert for that that hopefully reaches us before a user does. Note that this is fairly rare. The use case is to just provide enough evidence that this rare thing isn't happening right now


damianavila:

yuvipanda:
Me and @pnasrat discussed moving this to next quarter, based on capacity of the team this quarter. I think we should! Thoughts, @damianavila?

damianavila:
> Thoughts, @damianavila?

I concur with the move, although we should evaluate it alongside the other potential goals for Q3 and decide if we really want to include it for the next quarter.

PR #6337, opened 2025-07-08
Add an alert for when a user pod is unschedulable

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/6185

### DIFF
--- a/helm-charts/support/values.jsonnet
+++ b/helm-charts/support/values.jsonnet
@@ -95,6 +95,34 @@ local makePodRestartAlert = function(
   ],
 };
 
+local makeUserPodUnschedulableAlert = function(
+  name,
+  summary,
+  labels={},
+                                      ) {
+  name: name,
+  rules: [
+    {
+      alert: name,
+      expr: |||
+        # This alert fires when a user pod is unschedulable for more than 5 minutes.
+        # We use kube_pod_status_unschedulable to detect unschedulable pods.
+        count(
+          kube_pod_status_unschedulable{pod=~"jupyter-.*"} == 1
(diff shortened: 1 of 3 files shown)


helm-charts/support/values.jsonnet
```diff
@@ -95,6 +95,34 @@ local makePodRestartAlert = function(
...
+      alert: name,
+      expr: |||
+        # This alert fires when a user pod is unschedulable for more than 5 minutes.
+        # We use kube_pod_status_unschedulable to detect unschedulable pods.
+        count(
+          kube_pod_status_unschedulable{pod=~'jupyter-.*'} == 1
+          and (time() - kube_pod_created > 300)
+        ) by (namespace, pod) > 0
+      |||,
+      'for': '3m',
```
GeorgianaElena: @yuvipanda, I need your guidance on this one.

The way it is written, the alert self-resolves after the timeout period has passed and the pod is deleted, so I don't think is reliable. So I have looked into a different approach that's based on `jupyterhub_server_spawn_duration_seconds_count` metric, the one below:

```suggestion
        jupyterhub_server_spawn_duration_seconds_count{status='failure'} > 0
      |||,
      'for': '0m',
```

However, every 5 user server restarts, the hub itself gets a restart and the count it's reset to zero. I could potentially trigger the alert also when the hub has had a restart, or create a different alert for this?

What do you think it's the best approach here?

GeorgianaElena: I realize now that the later one, that checks the hub metric matches more https://github.com/2i2c-org/infrastructure/issues/2267

Issue #6342, opened 2025-07-11
VEDA: RStudio home directory does not persist


sunu:
### Context

On the VEDA hubs, when launching a single-user server with the Rocker image, any files saved in the home directory (/home/rstudio) are lost after the server is restarted.

### What we need to do

A quick look suggests the volume mount for home folders in RStudio notebooks is defined in the basehub values here:
https://github.com/2i2c-org/infrastructure/blob/5ee7c8836e0651f410f569b9dfc85bfb51dd323b/helm-charts/basehub/values.yaml#L799-L815

However, this gets overridden and dropped when singleuser.storage.extraVolumeMounts is set in individual hubs, as seen in the veda configuration:
https://github.com/2i2c-org/infrastructure/blob/5ee7c8836e0651f410f569b9dfc85bfb51dd323b/config/clusters/nasa-veda/common.values.yaml#L99-L114

Looks like other hubs like MAAP and nasa-cryo might be affected by this as well.

### Definition of Done

- [ ] Users can persist files in their home directories across restarts when using RStudio-based images on the affected hubs.

PR #6343, opened 2025-07-11
VEDA: include rstudio home directory mounts

sunu:
Lists in Helm values get overwritten in their entirety. We need to ensure that the home directory mount for rstudio is included in the values for individual hubs where we overwrite `singleuser.storage.extraVolumeMounts`.

refs #6342


GeorgianaElena (approved):
Aaaah, great catch @sunu! And I think you are right and other hubs might be affected too :(

PR #6351, opened 2025-07-11
earthscope: Rename the resource request labels

sarahw-earthscope:
Adding clearer display names for resource allocation choices, and removing smallest server option for now.

### DIFF
--- a/config/clusters/earthscope/common.values.yaml
+++ b/config/clusters/earthscope/common.values.yaml
@@ -190,22 +190,22 @@ basehub:
           resource_allocation:
             display_name: Resource Allocation
             choices:
-              mem_3_7:
-                display_name: 3.7 GB RAM, upto 3.7 CPUs
-                default: true
-                allowed_groups:
-                - geolab
-                - geolab:dev
-                - geolab:power
-                kubespawner_override:
-                  mem_guarantee: 3982489550
-                  mem_limit: 3982489550
-                  cpu_guarantee: 0.465625
-                  cpu_limit: 3.725
-                  node_selector:
-                    node.kubernetes.io/instance-type: r5.xlarge
+              # mem_3_7:
+              #   display_name: Try It: 3.7 GB RAM, >= 0.5 CPU
+              #   default: true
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thanks @sarahwilson523! We like to keep the labels consistent, as we may have to make adjustments constantly (one is coming up in https://github.com/2i2c-org/infrastructure/pull/6313 for example). I'm happy to merge this, but would like to revert it post the workshop. Does that seem acceptable?

yuvipanda:
@sarahwilson523 that is extremely valuble feedback! I'm incorporating that into the overall work we're doing in improving these resource options.

I've incorporated your points 2 and 3 into the script (https://github.com/2i2c-org/infrastructure/pull/6313), and the options now look like this:

image

image

The changes are:

1. Much more limited use of fractions - only for CPUs under 2. No fractions over that, everything is rounded to nearest whole number
2. The display name indicates how much CPU is guaranteed
3. The description indicates CPU limits with appropriate wording

I do agree with you on point 1, where descriptive words are better than just the numbers. We had considered using t shirt sizes (XS, S, M, L, XL, XXL, etc) but that implied the 'bigger is better' problem. I love the idea of using suggested names per-workload (light, moderate, etc) - but that can't be systematized as much, as what's 'standard' is going to change community to community. I *think* a pathway to explore is to either allow a community to pick 'standard' and we autogenerate names for instances smaller and bigger than that, or simply allow them to fully customize the labels. I'll spend a little more time thinking about it.

I've deployed these changes (https://github.com/2i2c-org/infrastructure/pull/6366) to the staging geolab (https://staging.geolab.earthscope.cloud/hub/oauth_login). Can you test it out and let me know what you think? In particular, do you think it's worth deploying this to earthscope as is, or should we wait until we have descriptive labels?

Thank you very much for your high quality feedback @sarahwilson523, it helps us make changes not just for geolab but for all our communities! It's well appreciated.

yuvipanda:
We made a lot of progress in different ways here, so closing this now :)

Issue #6370, opened 2025-07-16
[openscapes, workshop] [support] Tune resource allocations for 200 person workshop


jnywong:
### The Freshdesk ticket link

https://2i2c.freshdesk.com/a/tickets/3579

### Ticket request type

Feature Request

### Ticket impact

🟩 Low

### Short ticket description

Follow up to https://github.com/2i2c-org/infrastructure/issues/6270.

We need to tune resources for up to 200 simultaneous users for a workshop on **July 22, 2025**.

### Definition of done

- [ ] We optimize machines provisioned to support 200 users with minimal server startup time
- [ ] Restrict resource allocations to 14.8 GB RAM / up to 3.7 CPU

jnywong:
Update from the ticket: community leader would like to do a test run on Friday but I said that we cannot make such guarantees at short notice @GeorgianaElena

PR #6381, opened 2025-07-17
[openscapes] Offer just one allocation option for the workshop on a bigger machine:

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/6370

### Changes

- This PR uses #6313 to re-generates 5 choices on a r5.4xlarge.
- The option that was closest to the community's `14.8 GB RAM / up to 3.7 CPU` is chosen, i.e. `~15 GB RAM, ~1.9 CPUs, up to 15 cpu`
- It removes all other resource options that are not used and all other image options that are not used in the workshop

It's important to note that this choice is now on a r5.4xlarge instead of a r5.xlarge, which means that it will pack 8 users on a node instead of two, allowing for smaller startup time.
At ~200 users, there will be around 25 scale up events on a r5.4xlarge, whereas on a r5.xlarge there would have been 100 events.

### *This change should be reverted after the workshop

### DIFF
--- a/config/clusters/openscapes/workshop.values.yaml
+++ b/config/clusters/openscapes/workshop.values.yaml
@@ -64,7 +64,7 @@ basehub:
         default: true
         kubespawner_override:
           image: openscapes/python:3ad4036
-        profile_options: &profile_options
+        profile_options:
           requests: &profile_options_resource_allocation
             display_name: Resource Allocation
             choices:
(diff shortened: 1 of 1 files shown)


GeorgianaElena:
This is how it looks like deployed. @jnywong, what do you think?

Screenshot 2025-07-17 at 11 17 03

jnywong (approved):
This is great! The instructor will do a test run tomorrow, so hopefully we know about any potential issues ahead of time.

I've added a follow up issue to revert these changes https://github.com/2i2c-org/infrastructure/issues/6382

Thank you Georgiana :)

Issue #6323, opened 2025-07-03
Validate if higher IOPS were needed for GHG workshop


yuvipanda:
After the workshop #6308 is completed, let's look at our IOPS metrics to see if the higher IOPS we provisioned was needed.

## Definition of Done

- [x] Workshop is over (July 18)
- [x] We check to see how much IOPS was used in peak
- [x] based on that, we reduce IOPS as needed
- [ ] If IOPS increases were needed, we also proactively apply changes to workshops starting July 21 
~- [ ] Update documentation accordingly~ Tracked by https://github.com/2i2c-org/infrastructure/issues/6369

GeorgianaElena:
Based on these graphics fom 8-18 July, I don't think 10k iops were needed. I will open a PR to rm the 10k and let it default to 3k iops because it looks like the default it's enough? 

Not sure how different will the workshop starting today will be in terms of needs, so I'll wait for your feedback on this one Yuvi before changing anything.

PR #6409, opened 2025-07-21
[ghg] Remove temp iops bump

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/6323

Based on the graphics fom 8-18 July, I don't think 10k iops were needed and the default 3k looks like it's more than enough.

Not sure how different will the workshop starting today will be in terms of needs, so I'll wait for your feedback on this one @yuvipanda  before changing anything.

*The terraform changes in this PR were not applied.


GeorgianaElena:
Screenshot 2025-07-21 at 13 59 31
Screenshot 2025-07-21 at 13 57 15
Screenshot 2025-07-21 at 13 56 07

Issue #6265, opened 2025-06-23
Structure 'action timeline' into our pagerduty alerts using Pagerduty priorities


yuvipanda:
We must never have alerts we ignore (and we currently don't!), but this requires that the *timeline* for when some *specific action* must be taken should be clear for each alert.

Let's do this by using Pagerduty priorities. Each priority encodes an expected time period within which some action is *required* to be taken for this alert. The initial mapping of priorities - actions would be:

P1 - *immediate* (community *currently affected*) aka OUTAGES
P2 - *same-business-day* (community *about to be affected* if we don't do something *immediately*)
P3 - *five-business-days* (community *about to be affected* if we don't do something soon, but not immediately)
P4 - *next-sprint* (community not necessarily affected on a specific timeline, but we must take some action into the committed column of next sprint)

## Definition of done

- [ ] These levels and their meanings are documented
- [ ] All existing alerts have their priorities set

PR #6438, opened 2025-07-23
Add one more severity level to pvc full

GeorgianaElena:
Ref https://github.com/2i2c-org/infrastructure/issues/6265


yuvipanda (approved):
love it!

PR #6456, opened 2025-07-25
Updates the NeuroHackademy user images

noahbenson:


GeorgianaElena:
Hi @noahbenson! This got deployed only to staging because I believe there's something wrong with the image that prevents user servers from spawning. Can you take a look?

PR #6459, opened 2025-07-25
[projectpythia, prod] Add admin

jnywong:
Ref https://github.com/2i2c-org/meta/issues/2392


jnywong:
@jukent let me know if you or others would like [admin access](https://jupyterhub.readthedocs.io/en/stable/reference/urls.html#hub-admin) to the 2i2c hub :)

jukent:
Yes please that might be helpful! Thanks!

PR #6479, opened 2025-07-30
insert additional packages before starting singleuser

jmunroe:


jmunroe:
Thanks @yuvipanda and @agoose77 for the feedback.  Please see PR #6490.

PR #6490, opened 2025-07-31
Configure for Project Pythia Cookoff 2025

jmunroe:
This PR reverts some of the changes made in #6479

1.  Removes injection of file content from infrastructure in favour of using git-init instead to pull content
2. Ensure the the security patch ghsa-w3vc-fx9p-wp4v/check-patch-run is still used
3. uses conda instead of pip for injecting jupyterlab-git and gh-scoped-creds

### DIFF
--- a/config/clusters/projectpythia/common.values.yaml
+++ b/config/clusters/projectpythia/common.values.yaml
@@ -87,6 +87,18 @@ jupyterhub:
       securityContext:
         runAsUser: 1000
         runAsGroup: 1000
+    - name: volume-mount-ownership-fix
+      image: busybox:1.36.1
+      command:
+      - sh
+      - -c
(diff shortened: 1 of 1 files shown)


yuvipanda (changes requested):
I agree that's something for us to consider and figure out how to support. However, the current reality is that the way to support that is for these packages to be added to the image. I would like that to be fully reverted, and if that functionality is important, for it to go through our product process. We have fought long and hard to not have one off unique exemptions in our hub config, when we know they will cause issues, and I don't want that to be the case here.

yuvipanda:
Given that I'm off tomorrow, and I think this event starts on Monday, I've consulted with @aprilmj and delegated choice making around this to her.

agoose77:
@jmunroe are we in a position to revert this, now?

PR #6505, opened 2025-08-05
maint: use non-FHS flake

agoose77:
The Nix package manager's most disruptive design choice is the deliberate neglect of the [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html). This means that applications packaged for Nix cannot rely on `/usr/lib` etc, and instead need to use `rpath` and other mechanisms to ensure that they link properly. 

The Python ecosystem has a subtle concept of a "runtime environment" that most users don't see. On Linux, that's `manylinux`. In the existing version of this Flake, Nix builds a FHS environment using namespaces and chroot, so that the assumptions about where to find the `manylinux` dependencies holds true. 

This PR replaces the heavier FHS with a simple `mkShell`, and defines `LD_LIBRARY_PATH` to ensure that Python finds the proper dependencies for `manylinux`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -26,7 +26,7 @@
       python = pkgs.python313;
       packages =
         [
-          (python.withPackages (pypkgs: with pypkgs; [virtualenv]))
+          python
         ]
         ++ (with pkgs; [
           cmake
@@ -45,6 +45,9 @@
           eksctl
         ]);
       shellHook = ''
+        # Unset leaky PYTHONPATH
+        unset PYTHONPATH
+
         # Setup if not defined ####
         if [[ ! ( -d ".venv" && -f ".venv/marker" ) ]]; then
             __setup_env() {
@@ -56,9 +59,6 @@
                 # Stand up new venv
                 ${python.interpreter} -m venv .venv
 
-                # Unset leaky PYTHONPATH
-                unset PYTHONPATH
-
                 ".venv/bin/python" -m pip install -e ".[dev]"
 
                 # Add a marker that marks this venv as "ready"
@@ -66,9 +66,6 @@
             }
 
             __setup_env
-        else
-            # Unset leaky PYTHONPATH
-            unset PYTHONPATH
         fi
         ###########################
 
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -18,67 +18,71 @@
...
 
-              # Setup if not defined ####
-              if [[ ! ( -d ".venv" && -f ".venv/marker" ) ]]; then
-                  __setup_env() {
-                      # Remove existing venv
-                      if [[ -d .venv ]]; then
-                          rm -r .venv
-                      fi
+                # Unset leaky PYTHONPATH
+                unset PYTHONPATH
```
agoose77, line 60: Python packages like `pre-commit` leak `PYTHONPATH`.

flake.nix
```diff
@@ -18,67 +18,68 @@
...
-              azure-cli
-              terraform
-              eksctl
-            ]);
-          # If there is a higher power, why does it allow such horrible autoformatting below
-          runScript = "${pkgs.writeShellScriptBin "runScript" (''
-              set -e
+      packages =
+        [
+          python
```
agoose77, line 29: We don't actually use `virtualenv` (only `venv`), and `pip` is default bundled with `venv`!

flake.nix
```diff
@@ -18,67 +18,68 @@
...
+          sops
+          gdk
+          awscli2
+          azure-cli
+          terraform
+          eksctl
+        ]);
+      shellHook = ''
+        # Unset leaky PYTHONPATH
+        unset PYTHONPATH
```
agoose77, line 49: Certain packages from `nixpkgs` define `PYTHONPATH`, which leaks. It is not required.

agoose77:
@jmunroe if you've been testing out the `nix develop` support for the infra repo, you _might_ need to do the following after pulling this change:
1. Delete `.venv` in `infrastructure/`
2. Close your terminal (to ensure you're not running the old env)
3. Run `nix develop` in `infrastructure/`

PR #6508, opened 2025-08-05
fix: set a bogus username to silence Terraform error

agoose77:
I run into this when running plans on AWS IAAC. @GeorgianaElena I feel like you have added similar fixes for TF errors recently, can you review please? :pray:


agoose77:
I think this is probably one of those "I deleted my local TF lockfile and it's fine again" moments.

Issue #6391, opened 2025-07-18
Cost monitoring backend can be installed as a standalone helm-chart repository (phase 1)


jnywong:
### Context

The current cost reporting implementation is incorporated as part of the 2i2c infrastructure repository through mounted files in the aws-ce-grafana-backend [helm chart](https://github.com/2i2c-org/infrastructure/tree/main/helm-charts/aws-ce-grafana-backend). We should separate this into its own standalone repository so that other users can independently install the cost reporting tool using helm in line with our [right to replicate](https://2i2c.org/right-to-replicate/) policy.

### Definition of Done

- [x] Set up, initialise, rename and transfer [aws-ce-grafana-backend](https://github.com/2i2c-org/infrastructure/tree/main/helm-charts/aws-ce-grafana-backend) into a standalone repository, including git commit history
  - https://github.com/jnywong/jupyterhub-cost-monitoring
  - commit history migrated with https://github.com/newren/git-filter-repo
- [x] Set up CI/CD to publish helm charts with [chartpress](https://github.com/jupyterhub/chartpress)
  - [x] Rename from `aws-ce-grafana-backend` to `jupyterhub-cost-monitoring`
  - https://github.com/2i2c-org/jupyterhub-cost-monitoring/pull/2
- [x] Refactor the code where necessary
- [x] Documentation to support #6396 
  - https://github.com/2i2c-org/jupyterhub-cost-monitoring/pull/4
  - https://github.com/2i2c-org/jupyterhub-cost-monitoring/pull/5
- [x] Manually deploy on a test hub and validate
  - https://github.com/2i2c-org/infrastructure/pull/6481

agoose77:
We split this into two cards, 10 + 3, and marked *this* one as done.

jnywong:
Thanks Angus! Thangus!

PR #6515, opened 2025-08-06
[cost monitoring] Encapsulate repeatable configs as jsonnet

jnywong:
Ref https://github.com/2i2c-org/infrastructure/issues/6391


helm-charts/support/values.jsonnet
```diff
@@ -1,4 +1,6 @@
 local cluster_name = std.extVar('2I2C_VARS.CLUSTER_NAME');
+local provider_name = std.extVar('2I2C_VARS.PROVIDER');
+local cost_monitoring_iam_role = std.extVar('2I2C_VARS.COST_MONITORING_IAM');
```
yuvipanda: so IAM role names are *deterministic*, with the account number the only variable part. so let's make the variable be the account number, and go from there?

yuvipanda (changes requested):

helm-charts/support/values.jsonnet
jnywong: Brilliant! https://github.com/2i2c-org/infrastructure/pull/6515/commits/0f394c6be416c17279cad8f57b7aec43aeb6a961

This means we need AWS account numbers encapsulated in all our `cluster.yaml` files, which i can document in our cluster setup guide

helm-charts/support/values.jsonnet
```diff
@@ -127,6 +129,21 @@ local makeUserPodUnschedulableAlert = function(
...
+  enabled: true,
+  extraEnv: [
+    {
+      name: 'CLUSTER_NAME',
+      value: cluster_name,
+    },
+  ],
+  serviceAccount: {
+    annotations: {
+      'eks.amazonaws.com/role-arn': 'arn:aws:iam::%s:role/jupyterhub_cost_monitoring_iam_role' % aws_account_id,
```
yuvipanda: Add a comment with a link to the terraform line where we create this role? And in terraform, mention that if this name is changed, we must change this in jsonnet as well.

helm-charts/support/values.jsonnet
```diff
@@ -1,4 +1,6 @@
 local cluster_name = std.extVar('2I2C_VARS.CLUSTER_NAME');
+local provider_name = std.extVar('2I2C_VARS.PROVIDER');
+local aws_account_id = std.extVar('2I2C_VARS.AWS_ACCOUNT_ID');
```
yuvipanda: Add a comment on what this variable will resolve to if we aren't on AWS

jnywong: Added comment: "# undefined if provider_name != 'aws'"

helm-charts/support/values.jsonnet
jnywong: https://github.com/2i2c-org/infrastructure/pull/6515/commits/eb19f4b041a014a64dfde506b3c830cf675ec464

helm-charts/support/values.jsonnet
jnywong: https://github.com/2i2c-org/infrastructure/pull/6515/commits/eb19f4b041a014a64dfde506b3c830cf675ec464

jnywong: I think this should be fine, since the variable is currently only needed `if provider_name == 'aws' ` in 

`  'jupyterhub-cost-monitoring': if provider_name == 'aws' then configCostMonitoring else { enabled: false },`

jnywong: Actually! I have changed the approach to include `aws_account_id` as a [top-level argument](https://jsonnet.org/learning/tutorial.html#top-level-arguments) to handle the case where this is undefined for non-AWS cloud providers.

This is passed to the jsonnet via the `deployer deploy-support` command based on the new key `aws.active_cost_tags` under the `cluster.yaml` file.

jnywong:
@yuvipanda i have tested this on openscapeshub and nasa-ghg-hub and all looks good. If you decide to approve this, then I can slowly (and calmly :D) apply the terraform changes across all our AWS clusters over the weekend when user activity is low in case something gets broken.

deployer/commands/validate/cluster.schema.yaml
```diff
@@ -45,6 +45,7 @@ properties:
     description: |
       Account used to access the provider console UI. Most often we use our 2i2c accounts for this,
       but there are a few exceptional clusters and we should mark those.
+      For AWS accounts, provide the account ID (12 digits) and NOT the alias.
```
yuvipanda: Let's mark this as required, and remove comment about 'if no account is specified'

deployer/commands/validate/config.py
```diff
@@ -65,10 +65,6 @@ def _prepare_helm_charts_dependencies_and_schemas():
     _generate_values_schema_json(support_dir)
     subprocess.check_call(["helm", "dep", "up", support_dir])
 
-    aws_ce_grafana_backend = HELM_CHARTS_DIR.joinpath("aws-ce-grafana-backend")
```
yuvipanda, line 68: great to see this removed! I particularly hate this function here for unclear reasons :D

deployer/utils/jsonnet.py
```diff
@@ -28,16 +28,26 @@ def validate_jsonnet_version():
 
 
 @contextmanager
-def render_jsonnet(jsonnet_file: Path, cluster_name: str, hub_name: str | None):
+def render_jsonnet(
+    jsonnet_file: Path,
+    cluster_name: str,
+    hub_name: str = None,
+    provider: str = None,
+    aws_account_id: str = None,
```
yuvipanda, line 36: let's type this as an int

deployer/utils/jsonnet.py
```diff
@@ -47,6 +57,10 @@ def render_jsonnet(jsonnet_file: Path, cluster_name: str, hub_name: str | None):
     ]
     if hub_name is not None:
         command += ["--ext-str", f"2I2C_VARS.HUB_NAME={hub_name}"]
+    if provider is not None:
+        command += ["--ext-str", f"2I2C_VARS.PROVIDER={provider}"]
+    if aws_account_id is not None:
+        command += ["--tla-str", f"aws_account_id={aws_account_id}"]
```
yuvipanda: Is there a reason to use tla-str rather than ext-str here? If so, document it and if not let's use the same one everywhere.

Let's also use `2I2C_VARS.<ALL_CAPS>` as the format for variables we pass in? If that's not something we can do for tla let's find an equivalent one that very clearly marks it as a 2i2c specific variable we are passing in

deployer/infra_components/cluster.py
```diff
@@ -171,8 +171,24 @@ def deploy_support(self, cert_manager_version, debug, dry_run):
...
+                        "provider": self.spec["provider"],
+                        "cluster_name": (
+                            self.spec["aws"]["clusterName"]
+                            if self.spec["provider"] == "aws"
+                            else self.spec["name"]
+                        ),
+                    }
+                    if (
+                        self.spec["provider"] == "aws"
+                        and self.spec["aws"]["billing"]["active_cost_tags"]
```
yuvipanda: we can pass this account id in unconditionally on all AWS things regardless of wether we enable cost monitoring or not. Let's generally try to keep the knowledge about what's in the cluster files and *conditional logic* as minimal as possible

helm-charts/support/values.jsonnet
```diff
@@ -1,210 +1,229 @@
...
+              'The user pod {{ $labels.pod }} is unschedulable on cluster:%s hub:{{ $labels.namespace }}' % [cluster_name],
+            ),
+            diskIOApproachingSaturation,
+          ],
+        },
       },
     },
-  },
-}
+    'jupyterhub-cost-monitoring': if std.type(aws_account_id) != 'null' then configCostMonitoring(aws_account_id) else { enabled: false },
```
yuvipanda: Let's unconditionally enable this on AWS here by checking on provider.

yuvipanda (changes requested):
This mostly looks great! I want to expound a principle for our cluster.yaml and then ask for a specific change here.

The principle is that 'we should keep things in cluster.yaml *as minimal as possible*'. This means that every time we add something in there, we should check to see 'can this be done by another means?'. The primary goal here is to rely on things that are upstream (helm, z2jh) as much as possible. My intuition here makes me want to be particularly careful with things that are bools. `billing.paid_by_us` is not really used anywhere for example.

In this case, it looks like the only point of `active_cost_tags` is to enable or disable cost monitoring. If we were to apply the previous principle here, we can accomplish that goal with helm directly. `support/values.jsonnet` takes precedence over any `support.values.yaml` in a particular cluster. So I suggest we:

1. Remove `billing.active_cost_tags` as a field
2. Document that for cases where we *don't* have cost tags enabled, we should set `jupyterhub-cost-monitoring.enabled` to false in `support.values.yaml`
3. Set this wherever needed.

This also means we don't have to have a field in cluster.yaml that's `true` most of the time!

deployer/commands/validate/cluster.schema.yaml
jnywong: I've placed the `account` key under the `aws` key. This is because if we would like to enforce `int` type on the `account` value, then we can do necessarily do this with AWS account but not non-AWS accounts for other providers.

deployer/utils/jsonnet.py
jnywong: https://github.com/2i2c-org/infrastructure/pull/6515/commits/b5406b4c1b3f840c667762254e7fb3ac12083374

deployer/commands/validate/cluster.schema.yaml
jnywong: https://github.com/2i2c-org/infrastructure/pull/6515/commits/b5406b4c1b3f840c667762254e7fb3ac12083374

deployer/utils/jsonnet.py
jnywong: Yes, tla-str can be referenced in the `support/values.jsonnet` if they are undefined, which is the case for support charts for non-AWS clusters. I've wrapped everything but the global --ext-str variables into a function so that the tla-str can be passed in.

tla-str does not accept the format proposed, so I have tweaked our 2i2c variable names to: `VAR_2I2C_<variable_name>`

jnywong: https://github.com/2i2c-org/infrastructure/pull/6515/commits/32cb4f8cd44c4247e3809921fe2319e509026c3d

deployer/infra_components/cluster.py
jnywong: https://github.com/2i2c-org/infrastructure/pull/6515/commits/f18da6ea6c3d4e52737cf5c05f7080d0d529c80d

helm-charts/support/values.jsonnet
jnywong: Okay! I've checked that the `jupyterhub-cost-monitoring.enabled: false` override in the `support/values.yaml` works as expected and clobbers the jsonnet config.

PR #6529, opened 2025-08-06
Bump prometheus disk to be lower

yuvipanda:
This is expensive cloud cost that doesn't buy us anything - the thing it was trying to mitigate (an explosion in disk space) has been mitigated in other ways since.

Change the berkeley-geojupyter one because it has no real data yet. But others will need to be actively migrated.


agoose77:
Is the plan here to migrate one-by-one through per-cluster support values, and then revert 5f5b4aa0e2 and all of those PRs ?

PR #6536, opened 2025-08-07
[NeuroHackademy] Increase memory for GPU instances

arokem:


config/clusters/neurohackademy/common.values.yaml
```diff
@@ -143,8 +143,8 @@ jupyterhub:
               slug: gpu
               description: Start a container on a dedicated node with a GPU
               kubespawner_override:
-                mem_limit:
-                mem_guarantee: 14G
+                mem_limit: 32G
```
yuvipanda, line 146: The g4dn.xlarge only has 16G of RAM, so this won't work

arokem: Can we change the instance we are using?

yuvipanda: @arokem yep, on it

arokem: Done on my end!

yuvipanda:
Done in https://github.com/2i2c-org/infrastructure/pull/6537

PR #6537, opened 2025-08-07
Bump neurohackademy to g4dn.2xlarge

yuvipanda:
They need same amount of GPU but more RAM

Replaces https://github.com/2i2c-org/infrastructure/pull/6536


arokem:
Oh - OK, let's do 2x with 30G then. Thanks!

yuvipanda:
@arokem great, already deployed! I did provision (but not use) 4xlarge as well, so if you want more RAM (~61G) you can bump up straightforward with a PR (eksctl changes require manual work, which I've just done)

yuvipanda:
Hang on, turns out 30G is not right. Fixing

PR #6555, opened 2025-08-13
feat: add feature to prevent nested kubeconfig contexts

agoose77:
I recently targeted the wrong cluster because (I think) of a nested KUBECONFIG context — I invisibly lost the second subshell, and ended up working in the parent context.

This PR allows _me_ to make that harder for myself.

## Changes

I made some changes to avoid deploying _everything_ to main here:

1. Split out `use-cluster-credentials` into a separate submodule
2. Ignore `deployer/commands/develop/**` in CI 
3. Add new `DEPLOYER_NO_NESTED_KUBECONFIG` flag to deny nested kubeconfig contexts

We should add `__init__.py` entries to our various submodules, but that's not needed here, so I'm leaving it!

## Details
This PR makes it possible for 2i2c engineers to set 
```shell
export DEPLOYER_NO_NESTED_KUBECONFIG=1
```
such that sequential `deployer use-cluster-credentials` invocations fail.

I think this might have avoided the problem I ran into when the kubeconfig context reverted due to a shell error. This would also be ameliorated by setting e.g. a PS1, which I have also done with starship.

I'm biasing to action here -- this is opt-in and I think we are all comfortable with these kind of small additions.

### DIFF
--- a/config/clusters/cloudbank/authoring.values.yaml
+++ b/config/clusters/cloudbank/authoring.values.yaml
@@ -82,3 +82,7 @@ jupyterhub:
 jupyterhub-home-nfs:
   gke:
     volumeId: projects/cb-1003-1696/zones/us-central1-b/disks/hub-nfs-homedirs-authoring
+nfs:
+  pv:
+    serverIP: 10.3.253.47
+    baseShareName: /
--- a/config/clusters/cloudbank/bcc.values.yaml
+++ b/config/clusters/cloudbank/bcc.values.yaml
@@ -55,3 +55,7 @@ jupyterhub:
 jupyterhub-home-nfs:
   gke:
     volumeId: projects/cb-1003-1696/zones/us-central1-b/disks/hub-nfs-homedirs-bcc
+nfs:
+  pv:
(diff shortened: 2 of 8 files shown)


deployer/commands/deployer.py
```diff
@@ -37,6 +37,10 @@ def use_cluster_credentials(
     """
     Pop a new shell or execute a command after authenticating to the given cluster using the deployer's credentials
     """
+    if "DEPLOYER_NO_NESTED_KUBECONFIG" in os.environ and "KUBECONFIG" in os.environ:
```
yuvipanda, line 40: If there is a `~/.kube/config` file, you will still end up in a nested kubeconfig situation. So to fully prevent that, you should check for the existence of that file here as well and fail if that's found.

agoose77:
I've added a test for _non-empty_ `~/.kube/config` files. This lets me define kubeconfig as a `ro` empty file on my system, which helps in other contexts.

agoose77:
I've addressed Yuvi's review — now merging.

PR #6636, opened 2025-08-25
[CloudBank] nfs configs

sean-morris:
- I set the quota to 5 GB a person
- I went through and best guessed the deployments using - Number of Expected Users * 5 * 0.4

### DIFF
--- a/config/clusters/cloudbank/common.values.yaml
+++ b/config/clusters/cloudbank/common.values.yaml
@@ -30,3 +30,5 @@ jupyterhub-home-nfs:
   enabled: true
   gke:
     enabled: true
+  quotaEnforcer:
+    hardQuota: '5' # in GB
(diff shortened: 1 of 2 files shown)


GeorgianaElena:
@sean-morris, I terraform apply the new disk sizes, but had to revert the ones that were decreasing the disk size in https://github.com/2i2c-org/infrastructure/pull/6638/commits/21787484142c27e6cf6f6f6b6baadee0eaede357

That is we cannot actually decrease the size of disk, we ca only increase. If we ever want to decrease it we'll have to create a new, smaller disk, move all the data there and decommission the old one. 

Also, I have only re-deployed evc and haven't deployed _all_ the hubs again. I will let it be done automatically when I'll merge #6638. But let me know if you want that sooner.

sean-morris:
@GeorgianaElena thanks!  yes I knew that on the decrease no problem

Thanks on evc!

We are needing to update the datascience library to avoid a warning message. It is mainly just annoying but I am sure I will start getting emails!  The quicker I could deploy a new user image to all the hubs the better but it can wait. Just tell me when to try it.

s

PR #6638, opened 2025-08-25
Bump jupyterhub-home-nfs chart

### LINKED ISSUE #6329: Bump jupyterhub-home-nfs chart version
### Context

We currently run `jupyterhub-home-nfs` 0.1.0.

### What we need to do

We should upgrade the chart to 0.2.0 which comes with some changes https://github.com/2i2c-org/jupyterhub-home-nfs/blob/main/CHANGELOG.md#v020---2025-04-21 that would allow us to exclude the `shared` directory from the hard quota setting.

This is relevant for this neurohackademy request https://2i2c.freshdesk.com/a/tickets/3502

### Definition of Done

- [ ] Upgrade the chart version in `Chart.yaml`
- [ ] Upgrade the setup to account for the breaking change in this version
- [ ] Deploy this change to a single hub and manually test that everything still works
- [ ] Rollout to all hubs that have `jupyterhub-home-nfs` deployed

GeorgianaElena:
Fixes #6329

### DIFF
--- a/helm-charts/basehub/values.jsonnet
+++ b/helm-charts/basehub/values.jsonnet
@@ -17,6 +17,7 @@ local jupyterhubHomeNFSResources = {
     config: {
       QuotaManager: {
         paths: ['/export/%s' % hub_name],
+        hard_quota: 0,
       },
     },
     resources: {


helm-charts/basehub/values.jsonnet
```diff
@@ -14,15 +14,20 @@ local emitDaskHubCompatibleConfig(basehubConfig) =
 
 local jupyterhubHomeNFSResources = {
   quotaEnforcer: {
-    path: '/export/%s' % hub_name,
+    config: {
+      QuotaManager: {
+        paths: ['/export/%s' % hub_name],
+        hard_quota: 0,
```
GeorgianaElena, line 20: This means no quota. Without this, all hubs got a "10GB" quota by default from the `jupyterhub-home-nfs` chart.

helm-charts/basehub/values.jsonnet
```diff
@@ -14,15 +14,20 @@ local emitDaskHubCompatibleConfig(basehubConfig) =
...
-        cpu: 0.02,
-        memory: '20M',
+        cpu: 0.2,
+        memory: '750M',
       },
       limits: {
-        cpu: 0.04,
-        memory: '30M',
+        cpu: 0.4,
+        memory: '1G',
```
GeorgianaElena, line 30: This container loops through every dir recursively and sets the quota. It the depth is big and there are lots of user home dirs, I've seen the pod getting OOMkilled.

helm-charts/basehub/values.jsonnet
```diff
@@ -68,6 +73,6 @@ local jupyterhubGroupsExporterResources = {
 };
 
 emitDaskHubCompatibleConfig({
-  'jupyterhub-home-nfs': if is_staging then {} else jupyterhubHomeNFSResources,
-  'jupyterhub-groups-exporter': if is_staging then {} else jupyterhubGroupsExporterResources,
+  'jupyterhub-home-nfs': jupyterhubHomeNFSResources,
+  'jupyterhub-groups-exporter': jupyterhubGroupsExporterResources,
```
GeorgianaElena, line 77: If we don't apply this to staging hubs, they will get the default 10G quota set on the entire `/export/staging`

Issue #6592, opened 2025-08-19
Rollout storage quota to the other half of the cloudbank hubs


GeorgianaElena:
### Context

Follow-up to https://github.com/2i2c-org/infrastructure/issues/5656
This issue will be scoped down to 2-3 pts of work before starting

Hubs with persistent storage in this cluster:

- [ ] 1. ccsf
- [ ] 2. csm
- [ ] 3. csum
- [ ] 4. evc
- [ ] 5. fresno
- [ ] 6. humboldt
- [ ] 7. laney
- [ ] 8. moreno
- [ ] 9. norco
- [ ] 10. palomar
- [ ] 11. pasadena
- [ ] 12. redwoods
- [ ] 13. reedley
- [ ] 14. riohondo
- [ ] 15. saddleback
- [ ] 16. sbcc
- [ ] 17. sbcc-dev
- [ ] 18. sierra
- [ ] 19. sjcc
- [ ] 20. sjsu
- [ ] 21. skyline
- [ ] 22. srjc
- [ ] 23. tuskegee
- [ ] 24. ucsc
- [ ] 25. unr
- [ ] 26. wlac

### Definition of Done

- [ ] The feature/service is technically complete
- [ ] The feature/service been tested with one or more users (if applicable)
- [ ] The feature/service been deployed to a production cluster
- [ ] The feature/service has been shown to be replicably deployable
- [ ] The feature/service is well documented, and the documentation is accessible for the target user base
- [ ] The feature/service has been added to our product menu (if applicable).
- [ ] Lessons learned have been documented
- [ ] The feature's availability has been communicated to Sales
- [ ] The feature has been communicated to our communities

yuvipanda:
I'm inclined to either delay this another sprint, or pick this up myself to give @GeorgianaElena more space to work on the demonstrable reliability initiative.

aprilmj:
@yuvipanda and @GeorgianaElena can coordinate offline - we think Georgiana will do most of this half so she doesn't lose context on scripts

yuvipanda:
Thanks to all the excellent work @GeorgianaElena has done, I was able to take this through to completion.

PR #6668, opened 2025-08-29
Migrate all the cloudbank hubs to jupyterhub-home-nfs

yuvipanda:
Ref https://github.com/2i2c-org/infrastructure/issues/6592


yuvipanda:
don't merge yet, will do after all the migration is completed

PR #6673, opened 2025-08-30
Put ucmerced on n2-highmem-8

yuvipanda:
Also bump their user dirs as it was getting lowish.

Ref https://2i2c.freshdesk.com/a/tickets/3825


agoose77:
@yuvi are we anticipating reverting this in the near future, or is this the ongoing stable configuration? :blue_heart:

Issue #5560, opened 2025-02-19
Use `anonuid` and `anongid` on jupyterhub-home-nfs to simplify our infrastructure


yuvipanda:
We currently have two separate 'hacks' that are about making sure permissions on home directories are correct:

1. `nfs-share-creator` run before each deploy, making sure that the *root* directory (like `prod` or `staging`) under which all user home directories are created is owned by user `1000`. 
2. `volume-mount-ownership-fix` initContainer, run before each server start. This makes sure the specific user directory (like `prod/yuvipanda`) is owned by user `1000`.

We needed this because users in user containers run with a non root id of `1000`, and NFS implementations we used (Filestore, EFS) didn't consistently support being mounted by non-root users.

However, now that we run our own NFS server, we can solve this!

The in-kernel NFS server has two options - `anonuid` and `anongid` that allow mounting of the NFS server by a particular non-root uid, and then forcing *all* access to that NFS server to just be those uids. This is perfect for us, as we do only use a single uid - `1000` for everyone.

nfs-ganesha exposes these as `Anonymous_uid` and `Anonymous_gid` for this. 

So if we set those properties to `1000`, we can simply remove both `nfs-share-creator` and `volume-mount-ownership-fix` from our infrastructure.

## Definition of done

- [x] Set `Anonymous_uid` and `Anonymous_gid` in jupyterhub-home-nfs
- [x] Test to make sure that `nfs-share-creator` and `volume-mount-ownership-fix` can be removed with these set

GeorgianaElena:
I actually made progress on this one because I suspected might help with an outage. It didn't, but we should still check if it had solved the issue it should have.

GeorgianaElena:
I have found that if I set the squash option to be root_squash, then these uid and gid take effect.

GeorgianaElena:
The rollout work is now tracked by https://github.com/2i2c-org/infrastructure/issues/6839

PR #6714, opened 2025-09-03
Disable nfs-share-creator globally

yuvipanda:
It's been off on cloudbank for a while without issues, so I think it can be turned off everywhere.

Brings in https://github.com/2i2c-org/jupyterhub-home-nfs/pull/48 which is needed.

Ref https://github.com/2i2c-org/infrastructure/issues/5560 and https://github.com/2i2c-org/infrastructure/issues/6839

### DIFF
--- a/helm-charts/basehub/Chart.yaml
+++ b/helm-charts/basehub/Chart.yaml
@@ -26,7 +26,7 @@ dependencies:
     repository: "https://helm.dask.org/"
     condition: dask-gateway.enabled
   - name: jupyterhub-home-nfs
-    version: 1.1.0
+    version: 1.2.0
     repository: oci://ghcr.io/2i2c-org/jupyterhub-home-nfs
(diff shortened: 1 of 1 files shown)


GeorgianaElena:
@yuvipanda, I deployed the changes from https://github.com/2i2c-org/jupyterhub-home-nfs/pull/55 and this time, it worked.

So we can both remove the nfs share creator and the ownership fix initContainer ❤️

GeorgianaElena:
@yuvipanda, I asked a question in https://github.com/2i2c-org/infrastructure/issues/5560#issuecomment-3338668411 and cross-posting here .

yuvipanda:
@GeorgianaElena great work there! I left two comments in https://github.com/2i2c-org/jupyterhub-home-nfs/pull/55/files, but after those are addressed and deployed this is good to go

GeorgianaElena (approved):

PR #6742, opened 2025-09-09
fix: change server restart alerts to non-resolving delta-fns

agoose77:
Following https://github.com/2i2c-org/infrastructure/pull/6738, this PR changes server restart alerts to delta functions, and turns off the auto-resolution in PD.

This means that we'll trigger and then resolve the alert in Prometheus, but PD will keep the alert open. This reflects the point made in https://github.com/2i2c-org/infrastructure/pull/6738 — restart alerts and failed server starts are events, not system state, so we can't automatically resolve them.

### DIFF
--- a/helm-charts/support/values.jsonnet
+++ b/helm-charts/support/values.jsonnet
@@ -104,7 +104,7 @@ function(VARS_2I2C_AWS_ACCOUNT_ID=null)
         (
               sum by (pod, namespace) (kube_pod_container_status_restarts_total{pod=~".*%s.*"})
             -
-              (sum by (pod, namespace) (kube_pod_container_status_restarts_total{pod=~".*%s.*"} offset 10m))
+              sum by (pod, namespace) (kube_pod_container_status_restarts_total{pod=~".*%s.*"} offset 10m)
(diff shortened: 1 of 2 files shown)


helm-charts/support/values.jsonnet
```diff
@@ -46,10 +46,12 @@ function(VARS_2I2C_AWS_ACCOUNT_ID=null)
...
-      # The 'min' is to reduce the labels being passed to only the necessary ones
-      min(
-        jupyterhub_server_spawn_duration_seconds_count{status="failure"} > 0
-      ) by (namespace)
+      # The 'max' is to reduce the labels being passed to only the necessary ones
+      max by (namespace) (
+          (jupyterhub_server_spawn_duration_seconds_count{status="failure"} > 0)
+        -
+          ((jupyterhub_server_spawn_duration_seconds_count{status="failure"} offset 2m) > 0)
+      ) > 0
```
agoose77, line 54: Trigger an alert if any server has a new spawn failure (since 2m ago). 

Use max because $N(T)-N(T-2m)$ is positive for failures.

PR #6759, opened 2025-09-15
feat: add dashboard shortcut 💻

agoose77:
I find myself often looking at the AWS/GCP console to investigate logs or other resources. Given that most of our providers have dashboard URLs, I thought it would be useful to enshrine this in the structured data about each cluster, and add a shortcut to launch it.

This is something you can already do with
```shell
xdg-open $(rg 'provider.*(https.*)' config/clusters/$CLUSTER_NAME/cluster.yaml -r '$1')
```

But I think we should optimise things we do often.


yuvipanda:
You should add to the new cluster instructions to add this information

PR #6760, opened 2025-09-15
[cloudbank] remove filestore 

agoose77:
Now that we're using persistent disks, and @sean-morris has confirmed that we can remove the disk on Slack: https://2i2c.slack.com/archives/C01J397UX24/p1757710786151509

This PR removes the disk from our Terraform, after separately deleting it from GCP.

### DIFF
--- a/terraform/gcp/projects/cloudbank.tfvars
+++ b/terraform/gcp/projects/cloudbank.tfvars
@@ -6,9 +6,7 @@ region           = "us-central1"
 regional_cluster = false
 
 enable_filestore_backups = true
-filestores = {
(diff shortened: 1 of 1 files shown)


terraform/gcp/projects/cloudbank.tfvars
```diff
@@ -6,9 +6,7 @@ region           = "us-central1"
 regional_cluster = false
 
 enable_filestore_backups = true
```
agoose77:
```suggestion
enable_filestore_backups = false
```

agoose77: This doesn't make any difference, but better reflects our intent.

Issue #6698, opened 2025-09-02
Keep to within 2k USD for the Heliophysics Hub


yuvipanda:
The heliophysics hub we are setting up is supposed to use no more than 2k USD in AWS spending. This requires some effort on our part.

## Definition of Done

- [ ] Setup AWS budget alerts in 500$ increments, so we know where we are. These should come into our support email, and we must know whom to notify when it happens
- [x] Reduce our base cost by using smaller core node instances. This requires a small amount of tuning to make sure we can run on a single r7i.large or r5.xlarge rather than multiple ones

Ultimately if the costs go out because the hub saw too much usage, that's acceptable! But let's do what we can to support them by reducing our base cost.

aprilmj:
@yuvipanda is this FD alert what we're looking for in trying to limit cloud spending? https://2i2c.freshdesk.com/a/tickets/3955

yuvipanda:
@aprilmj nope, but i'm just setting up other ones.

I'm tempted to remove those auto adjusting budget alerts, as we don't really do anything with them.

yuvipanda:
@aprilmj ok, here's what the new alerts will look like: https://2i2c.freshdesk.com/a/tickets/3961

I would say when you receive them next, let's:

1. Let the community know (by cc'ing the appropriate rep)
2. Ask them what they would like to do 

I would imagine we mostly shut down everything after the 2000$ alert.

Since this is still a manual process, we may go over.

Do you accept this new responsibility, @aprilmj?

aprilmj:
Alert is working: https://2i2c.freshdesk.com/a/tickets/3996 

We're at $200 / 10% of their budget.

yuvipanda:
This needs documentation of the new feature we added (% cost alerts) and then can be done.

yuvipanda:
https://github.com/2i2c-org/infrastructure/issues/6836 tracks the documentation piece.

PR #6786, opened 2025-09-17
heliophysics: Add explicit budget alerts, remove auto-adjusting budge…

yuvipanda:
…t alert

- Add budget alerts for $500, 1000, 1500, 2000
- Remove the 'auto adjusting budget alert' as that doesn't seem to be doing much here so far

Ref https://github.com/2i2c-org/infrastructure/issues/6698


yuvipanda:
https://2i2c.freshdesk.com/a/tickets/3961 this is how it'll look like

Issue #6795, opened 2025-09-19
Allow Sean Morris to allocate new disks for CloudBank


agoose77:
Presently, for new Hubs on CloudBank, Sean has to ping us to allocate raw block storage for the home directory volume. It would be nice to prevent these synchronous requests from blocking him. What we do here might serve as a pattern to replicate for future collaborators of the kind that we have in Sean.

We cannot grant the same level of access that 2i2c engineers have, which means that we can't just grant access to the shared terraform backend (even if deployments are scoped to cloudbank, the TF state is global) so we need to find another solution.

@yuvipanda suggested migrating our Terraform backend to the Cloudbank GCP project, so that we can scope IAAC changes to exclusively the CB cluster: https://github.com/2i2c-org/infrastructure/pull/6797#pullrequestreview-3246004031

### Definition of Done
- [x] Cloudbank TF state is held in CB GCP project
- [x] Documentation is updated to reflect this nuance where appropriate
- [x] Sean has written guidance / links to our documentation that describe the process of provisioning a new disk
- [x] It is possible for Sean to modify CB infrastructure without 2i2c involvement

yuvipanda:
I've asked @sean-morris to validate #7301. I'll write up docs for our team, as well as some info on how to do this migration in the future.

yuvipanda:
Added some more docs https://github.com/2i2c-org/infrastructure/pull/7338.

@sean-morris is pretty close to being able to fully validate it https://2i2c.slack.com/archives/C01J397UX24/p1765915286951119

yuvipanda:
Sean was able to work through this and deploy https://github.com/2i2c-org/infrastructure/pull/7339 himself! He's no longer dependent on 2i2c engineers to deploy new hubs!

\o/

PR #6797, opened 2025-09-19
[cloudbank] Add workflow for creating home-directory disks without deployer

agoose77:
This PR addresses https://github.com/2i2c-org/infrastructure/issues/6795 by adding documentation and resources for creating a home-directory disk on GCP using only `gcloud`. This disk can later (non-blocking) be added to Terraform by 2i2c engineers, requiring only limited project-specific `gcloud` credentials for the original PR author.

The idea is simple:
1. Create a disk using gcloud. 
2. Import this disk, later, into Terraform


yuvipanda:
There's the slight additional complexity of having to switch backends when we are working and applying changes across all our clusters, but that's fairly minor and should be fine.

PR #6801, opened 2025-09-19
Explicitly disable admins on climaterisk and run an auth validation step before a deployment

### LINKED ISSUE #6767: Ensure no workshop hubs have admin access
I noticed that at least one workshop hub provided the ability to impersonate 2i2c admin users. This is a security risk.

@jnywong mentioned that @GeorgianaElena had looked at this recently, so I've assigned us both for now!

### Definition of Done
- [ ] All workshop hubs are audited to ensure that we don't expose this any more.
- [ ] A systematic way to disable this (e.g. via basehub config that denies combining shared password with 2i2c user provisioning) as an opt-out is created.

GeorgianaElena:
Fixes #6767

### DIFF
--- a/deployer/commands/validate/config.py
+++ b/deployer/commands/validate/config.py
@@ -219,10 +219,9 @@ def authenticator_config(
     hub_name: str = typer.Argument(None, help="Name of hub to operate on"),
 ):
     """
-    For each hub of a specific cluster:
-     - It asserts that when the JupyterHub GitHubOAuthenticator is used,
-       then `Authenticator.allowed_users` is not set.
-       An error is raised otherwise.
+    For each hub of a specific cluster it asserts that:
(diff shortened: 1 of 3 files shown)


deployer/commands/validate/config.py
```diff
@@ -270,6 +271,13 @@ def authenticator_config(
                     being used for auth so valid members are not refused access.
                 """
             )
+        elif hub.authenticator == "dummy" and admin_users != []:
+            raise ValueError(
+                f"""
+                    For security reasons, please unset `Authenticator.admin_users` for {hub.spec['name']} when the dummy authenticator is
+                    being used for authentication.
+                """
+            )
```
agoose77, line 281: Yes!!! ✨

PR #6804, opened 2025-09-19
Experiment with adding CODEOWNERS for specific instances

jsignell:
We were discussing who needs to be pinged for a each VEDA hub and it occurred to me that it might be possible to automate this a bit. Anyways this is a little experiment.


jsignell:
@yuvipanda pointed out that you need write access for this to work :(

PR #6828, opened 2025-09-29
NASA VEDA: Enable auth for fancy-profile binderhub

sunu:
Make the necessary config changes to test https://github.com/2i2c-org/jupyterhub-fancy-profiles/pull/122 on VEDA staging hub. cc @batpad

@GeorgianaElena, @agoose77 -- I'm not 100% sure that I've made the config changes in the right place and in the right format. I would appreciate it if you could take a quick look before we test this on staging 🙏🏽

### DIFF
--- a/config/clusters/nasa-veda/staging.values.yaml
+++ b/config/clusters/nasa-veda/staging.values.yaml
@@ -32,7 +32,9 @@ basehub:
           subPath: _shared-public
     hub:
       image:
-        tag: 0.0.1-0.dev.git.13134.hb8ed00131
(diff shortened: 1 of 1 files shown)


sunu:
thanks @yuvipanda! I'll pair with @batpad to deploy and test this on the staging hub.

sunu:
@agoose77 Thanks for the flag! nothing blocking really. @batpad and I just need to coordinate among ourselves to deploy and test with a new pre-release version of jupyterhub-fancy-profiles.

PR #6829, opened 2025-09-29
Update Helm chart for NFS server updates

agoose77:
Recently, we made a few updates to `jupyterhub-home-nfs` in the interest of reducing the verbosity of logging and increasing hardness against desync between the file system quota attributes and the quota generator internal state.

This PR updates our helm charts to deploy this new release.

### DIFF
--- a/helm-charts/basehub/values.jsonnet
+++ b/helm-charts/basehub/values.jsonnet
@@ -68,6 +68,9 @@ local jupyterhubHomeNFSConfig = {
       QuotaManager: {
         paths: ['/export/%s' % hub_name],
         hard_quota: 0,
+        projects_file: '/export/%s/.projects' % hub_name,
+        projid_file: '/export/%s/.projid' % hub_name,
+        log_level: 'INFO',
       },
     },
   },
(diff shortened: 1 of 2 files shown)


helm-charts/basehub/values.jsonnet
```diff
@@ -68,6 +68,9 @@ local jupyterhubHomeNFSConfig = {
       QuotaManager: {
         paths: ['/export/%s' % hub_name],
         hard_quota: 0,
+        projects_file: '/export/%s/.projects' % hub_name,
+        projid_file: '/export/%s/.projid' % hub_name,
```
agoose77, line 72: Now we're going to store these quota files on the NFS volume. 

These files do not _need_ to be persisted, but by placing them here we can make restarts faster by avoiding state invalidation.

helm-charts/basehub/values.jsonnet
```diff
@@ -68,6 +68,9 @@ local jupyterhubHomeNFSConfig = {
...
+        log_level: 'INFO',
```
agoose77, line 73: If we don't specify this, the default log level is such that we see nothing.

GeorgianaElena (approved):
Looks good to me, Angus! 🚀 Amazing work ✨

PR #6958, opened 2025-10-16
feat: add command to get current quota usage

agoose77:
This PR adds a command to retrieve the current NFS usage quotas:

```bash
deployer exec get-quota-usage <CLUSTER> <NAME> --json
```

This can be chained with commands to e.g. figure out new quota values (https://gist.github.com/agoose77/733e7ec4e38a9c0d942b1ac67c3ee0ea)


agoose77:
I'm merging this @yuvipanda, because it's useful for me atm, and I think we should not treat dev commands as blockers. I'll cancel the CI.

LMK if that's too bias-for-action!

PR #6974, opened 2025-10-17
fix: bump helm-chart version of jupyterhub-home-nfs (logging)

agoose77:
This PR follows #6973 to pull in a revised logging strategy. This should address the cash associated with OOM.


agoose77:
I've tested this on neurohackademy prod & staging, and it started and reconciled successfully.

Issue #6685, opened 2025-09-02
Develop a process for deploying a sub-chart version upgrade to batches of hubs at a time


GeorgianaElena:
### Context

Currently, if we want to upgrade one of the sub-charts's version and only test it on one hub, we must manually deploy the change and make sure nobody overwrites that via CI/CD.

This makes it hard to do things like:
- deploy and upgrade to a few production hubs and wait a few days before rolling-it out to other hubs if nothing breaks

### What we need to do

- Create a new deployment flow that allows each cluster/hub to use its own, configurable `Chat.yaml` file 

### Definition of Done

- [ ] Such a deployment flow exists or a refined plan to implement it exists
- [ ] Documentation is updated
- [ ] A process for doing sub-chart version upgrades that uses this deployment flow exists

PR #6997, opened 2025-10-21
Support providing custom Chart.yaml files per hub

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/6685

What this PR changes:
- we'll be running schema generation and helm dependencies updates, separate for support and hub. This means that we won't be running `helm dep up basehub` when we deploy the support chart for example
- `deployer validate hub-config` and `deployer validate authenticator-config`don't longer exist. Instead they are called as part of `deployer validate all-hub-config` under a context manager that takes into account a `Chart.yaml` override.
- `deployer validate all` is also gone and replaced by calling `validate all-hub-config`, `validate cluster-config` and `validate support config`
- there is now a context manager called `get_chart_dir` that takes into account a `chart_override` passed in `cluster.yaml` of a cluster. This can be used to run helm commands to make sure that the desired Chart.yaml is used.

### DIFF
--- a/docs/howto/upgrade-cluster/sub-chart-version.md
+++ b/docs/howto/upgrade-cluster/sub-chart-version.md
@@ -16,19 +16,21 @@ This will bump the version of the sub-chart for all hubs that use the `basehub` 
 
 ## Bumping the version of a helm sub-chart for a specific hub
 
-### 1. Create a custom `Chart.yaml` file, e.g. `staging-chart.yaml`
-  a. **Override a basehub**
-    - copy the contents of the original `charts/basehub/Chart.yaml` file into `staging-chart.yaml`
-    - update the version of the sub-chart you want to bump.
+### 1. Create a custom `Chart.yaml` file
+Example `staging-chart.yaml`
 
-  b. **Override a legacy daskhub**
-    - copy the contents of the original `charts/basehub/Chart.yaml` file into `staging-chart.yaml`
-    - override the description and name fields with:
-      ```yaml
-      description: Deployment Chart for a dask-enabled JupyterHub
-      name: daskhub
-      ```
-    - update the version of the sub-chart you want to bump.
+a. **Override a basehub**
+  - copy the contents of the original `charts/basehub/Chart.yaml` file into `staging-chart.yaml`
+  - update the version of the sub-chart you want to bump.
+
+b. **Override a legacy daskhub**
+  - copy the contents of the original `charts/basehub/Chart.yaml` file into `staging-chart.yaml`
+  - override the description and name fields with:
+    ```yaml
+    description: Deployment Chart for a dask-enabled JupyterHub
+    name: daskhub
+    ```
+  - update the version of the sub-chart you want to bump.
 
 ### 2. Configure the hub to use the custom `Chart.yaml` file
 


deployer/commands/deployer.py
```diff
@@ -68,6 +70,21 @@ def deploy_support(
...
+        if chart_override_path:
+            chart_yaml = chart_dir / "Chart.yaml"
+            temp_default_chart_copy = chart_dir / "Chart-copy.yaml"
+
+            shutil.copy(chart_yaml, temp_default_chart_copy)
+            shutil.copy(chart_override_path, chart_yaml)
+        yield chart_dir
+    finally:
+        if chart_override_path:
+            os.replace(temp_default_chart_copy, chart_yaml)
```
GeorgianaElena: Whenever we run a helm command, we can run it under this context manager that:

- makes a copy of the current Chart.yaml
- copies the hub's custom chart in the basehub's Chart.yaml

When it's done, it replaces Chart.yaml with the original file and removes the copy.

yuvipanda: I think this may cause issues with parallel runs of deployment at the same time. I suggest that if an override is set, we copy the entire chart folder to a temp directory and run helm commands there. If there are no overrides, just use the existing directory.

GeorgianaElena: Ah, correct. It's funny, cause this was the first idea I had, then replacing Chart.yaml felt cleaner and more straightforward.

deployer/commands/validate/config.py
```diff
@@ -48,121 +50,159 @@ def _generate_values_schema_json(helm_chart_dir):
...
+        shutil.copytree(default_chart_dir, temp_chart_dir_name, dirs_exist_ok=True)
+        # copy the chart override file into the temporary chart directory
+        shutil.copy(chart_override_path, temp_chart_dir_name)
+        # rename the override file so that it overrides "Chart.yaml"
+        default_chart_yaml = Path(temp_chart_dir_name) / "Chart.yaml"
+        os.rename(Path(temp_chart_dir_name) / chart_override, default_chart_yaml)
+        chart_dir = Path(temp_chart_dir_name)
+    yield chart_dir
+    if chart_override:
+        temp_chart_dir.cleanup()
```
GeorgianaElena: @yuvipanda, can you please take another look at this? It's been updated to use the temporary directory approach.

deployer/commands/validate/config.py
```diff
@@ -48,121 +50,167 @@ def _generate_values_schema_json(helm_chart_dir):
...
+        # copy the chart directory into the temporary location
+        shutil.copytree(default_chart_dir, temp_chart_dir_name, dirs_exist_ok=True)
+        # copy the chart override file into the temporary chart directory
+        shutil.copy(chart_override_path, temp_chart_dir_name)
+        # rename the override file so that it overrides "Chart.yaml"
+        default_chart_yaml = Path(temp_chart_dir_name) / "Chart.yaml"
+        os.rename(Path(temp_chart_dir_name) / chart_override, default_chart_yaml)
+        chart_dir = Path(temp_chart_dir_name)
+    yield chart_dir
+    if chart_override:
```
yuvipanda: This should probably be under a `try...finally` so it catches any exceptions?

GeorgianaElena:
> So maybe we have to copy the entire directory, or specialcase daskhub? I would be ok with specialcasing daskhub, because there's no point in doing a Chart.yaml override for daskhub, only for basehub.

So, what I did was to _always_ copy the contents of `basehub` even if we're overriding a legacy daskhub. We just need to make sure we're calling it `daskhub` in the override Chart.yaml.
Does that make sense?

Issue #6865, opened 2025-10-01
Round 1 of EKS cluster refactor Q4 2025


yuvipanda:
This issue tracks doing migrations of 4 clusters for https://github.com/2i2c-org/infrastructure/issues/6756. As this is the first time someone other than @yuvipanda is going to do this, let's pick 4 varied ones.

## Definition of Done

The following clusters have been migrated

- [x] nasa-veda
- [x] openscapes
- [x] opensci
- [x] nasa-cryo

In addition, we have checked that:

- [x] dask-gateway works fine in nasa-veda
- [x] GPU works fine in nasa-cryo

### To Do Items
- [ ] Tear down openscapeshub legacy node group (with cordoned node)
- [ ] Confirm same for nasa-cryo
- [ ] Confirm same for nasa-veda
- [ ] Document modifications to openscapeshub

Issue #6756, opened 2025-09-12
[Infrastructure Refactor] EKS clusters refactor Q4 2025


yuvipanda:
We want to refactor all our EKS clusters to simplify, standardize and stop doing manual copy-paste work. This will save us future time, make all our infrastructure easier to support, and get us on current EKS and AmazonLinux.  

This is the migration plan for the EKS cluster refactor done in https://github.com/2i2c-org/infrastructure/pull/6721. 

## Changes

- Migrate from unmanaged nodegroups to managed nodegroups
- Migrate from r5.xlarge to r8i-flex.large for core nodes
- Enable EKS control plane logging, retaining logs for 7 days
- Migrate to AmazonLinux2023 from deprecated AmazonLinux2
- EKS version upgrade to 1.33 (or 1.34, if it is available)
- Enable [Node Autorepair](https://docs.aws.amazon.com/eks/latest/userguide/node-health.html)

## Untested

- Does GPU node scalup work correctly?
- Does dask worker scaleup work correctly with spot instances?

## Timeline

We want everything migrated before end of the year of 2025.

## Clusters to Migrate

All AWS clusters must be migrated. 

- [x] 2i2c-aws-us https://github.com/2i2c-org/infrastructure/issues/6866
- [x] berkeley-geojupyter
- [x] catalystproject-africa https://github.com/2i2c-org/infrastructure/issues/6869
- [x] disasters https://github.com/2i2c-org/infrastructure/issues/6868 
- [x] earthscope https://github.com/2i2c-org/infrastructure/issues/6868
- [x] jupyter-health https://github.com/2i2c-org/infrastructure/issues/6866
- [x] maap https://github.com/2i2c-org/infrastructure/issues/6867
- [x] nasa-cryo https://github.com/2i2c-org/infrastructure/issues/6865
- [x] nasa-ghg https://github.com/2i2c-org/infrastructure/issues/6866
- [x] nasa-veda https://github.com/2i2c-org/infrastructure/issues/6865
- [x] nmfs-openscapes https://github.com/2i2c-org/infrastructure/issues/6867
- [x] oceanhackweek
- [x] openscapes https://github.com/2i2c-org/infrastructure/issues/6865
- [x] opensci https://github.com/2i2c-org/infrastructure/issues/6865
- [x] projectpythia https://github.com/2i2c-org/infrastructure/issues/6867
- [x] reflective https://github.com/2i2c-org/infrastructure/issues/6868 
[...]

PR #7003, opened 2025-10-22
refactor: Use new template for EKS round 1

agoose77:
This PR follows https://github.com/2i2c-org/infrastructure/issues/6756 to close https://github.com/2i2c-org/infrastructure/issues/6865.

### Lessons Learned
- For clusters with non-standard/mixed generations, just create the new cluster with a unique generation and manually clear up the old node pools
-

### DIFF
--- a/config/clusters/templates/common/support.values.yaml
+++ b/config/clusters/templates/common/support.values.yaml
@@ -47,4 +47,10 @@ cluster-autoscaler:
 
 calico:
   enabled: true
+
+
+# FIXME: remove this once eksctl is fixed
+nvidiaDevicePlugin:
+  aws:
+    enabled: true
 {% endif %}


eksctl/libsonnet/cluster.jsonnet
```diff
@@ -53,23 +54,25 @@
...
     desiredCapacity: minSize,
-    instanceType: instanceType,
+    // For multiple instance types, we need to set the instanceTypes key
+    [if std.isString(instanceType) then 'instanceType' else 'instanceTypes']: instanceType,
     volumeSize: 80,
     amiFamily: 'AmazonLinux2023',
     nodeRepairConfig: {
       enabled: true,
     },
     labels: {
```
agoose77, line 74: These changes are designed to let us specify multiple instance types. We need this for Dask, because we can't always guarantee spot availability of a particular node. We *can* alternatively specify features that we want, like nCPU, etc., but that's at-risk of provisioning expensive nodes and/or requiring us to to specify every tiny feature we need.

Meanwhile, we use the instance type as a kind of unique-ID component for the nodegroup name. So, if we support multiple instances, now we need to include them all in the name pattern.

agoose77:
I'm not sure if I have access to VEDA for deployment. Let me merge this in, and then we can iterate.

PR #7007, opened 2025-10-22
chore: update maap domains

jjfrench:
## Changes
- updates `jupyterhub.ingress.hosts` and `jupyterhub.ingress.tls` in `<hub-name>.values.yaml`
- updates `GitHubOAuthenticator` in `prod.values.yaml`
- updates `GenericOAuthenticator` in `staging.values.yaml`
- updates `cluster.yaml` domains
- adds redirects in `support.values.yaml` from 2i2c domains to MAAP domains

### DIFF
--- a/config/clusters/maap/support.values.yaml
+++ b/config/clusters/maap/support.values.yaml
@@ -38,7 +38,7 @@ calico:
 
 redirects:
   rules:
-    - from: staging.hub.maap.2i2c.cloud
-      to: staging.hub.maap-project.org
-    - from: hub.maap.2i2c.cloud
-      to: hub.maap-project.org
+  - from: staging.hub.maap.2i2c.cloud
+    to: staging.hub.maap-project.org
+  - from: hub.maap.2i2c.cloud
+    to: hub.maap-project.org


config/clusters/maap/staging.values.yaml
```diff
@@ -195,7 +195,7 @@ jupyterhub:
       Authenticator:
         admin_users: []
       GenericOAuthenticator:
-        oauth_callback_url: https://staging.hub.maap.2i2c.cloud/hub/oauth_callback
+        oauth_callback_url: https://staging.hub.maap-project.org/hub/oauth_callback
```
jjfrench, line 198: GenericOAuthenticator wasn't listed as one of the auth providers that needed changing - should this be changed?

config/clusters/maap/support.values.yaml
```diff
@@ -35,3 +35,10 @@ cluster-autoscaler:
 
 calico:
   enabled: true
+
+redirects:
+  rules:
+  - from: staging.hub.maap.2i2c.cloud
+    to: staging.hub.maap-project.org
+  - from: hub.maap.2i2c.cloud
+    to: hub.maap-project.org
```
jjfrench, line 44: Can someone confirm this is the correct location for the redirects?

config/clusters/maap/staging.values.yaml
yuvipanda: Ya, and there should be an equivalent change in the keycloak config to add this as an allowed callback url as well

PR #7017, opened 2025-10-23
chore: update prod auth

jjfrench:
## Changes
- updates `jupyterhub.hub.config` in `prod.values.yaml` to keycloak config

### DIFF
--- a/config/clusters/maap/prod.values.yaml
+++ b/config/clusters/maap/prod.values.yaml
@@ -12,8 +12,38 @@ jupyterhub:
       2i2c/hub-name: prod
   hub:
     config:
-      GitHubOAuthenticator:
+      JupyterHub:
+        authenticator_class: generic-oauth
+      Authenticator:
+        admin_users: []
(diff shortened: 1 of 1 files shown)


smohiudd:
Before this gets merged there are two additional steps:
1. @yuvipanda can we get your team to add this redirect in the prod CIlogon account used by veda: `https://auth.openveda.cloud/realms/maap/broker/cilogon/endpoint`
2. I will send by secure email the client id and secret to 2i2c for this the prod jupyter hub client in Keycloak

wildintellect:
@smohiudd  looks like the domain switch is working, and Keycloak on [hub.maap-project.org](http://hub.maap-project.org/) is working, a few things to fix ...
- [ ] Should we point prod to EDL prod? unlike staging that point to EDL UAT
- [ ] The redirect after login isn't working correctly (doesn't take user back to the hub)

https://staging.hub.maap-project.org/ works as expected

PR #7019, opened 2025-10-23
[CloudBank] Southern Oregon University

sean-morris:


terraform/gcp/projects/cloudbank.tfvars
```diff
@@ -181,6 +181,10 @@ persistent_disks = {
     size        = 84
     name_suffix = "skyline"
   }
+  "spelman" = {
```
agoose77: Just need to tweak this for sou

PR #7043, opened 2025-10-28
[nasa-veda]: Upgrade EKS k8s

agoose77:
Follows https://github.com/2i2c-org/infrastructure/pull/7003

### DIFF
--- a/eksctl/nasa-veda.jsonnet
+++ b/eksctl/nasa-veda.jsonnet
@@ -1,6 +1,6 @@
 local cluster = import './libsonnet/cluster.jsonnet';
 
-local _c = cluster.makeCluster(
+local c = cluster.makeCluster(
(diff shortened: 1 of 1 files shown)


config/clusters/nasa-veda/binder.values.yaml
```diff
@@ -27,6 +27,8 @@ jupyterhub:
       extraVolumeMounts: []
   singleuser:
     nodeSelector:
+      # Only use small instances
+      node.kubernetes.io/instance-type: r5.xlarge
```
agoose77, line 31: @GeorgianaElena I did this because the previous eksctl deployment for veda only has a single node type. 

Now, there are technically larger node-pools available for the singleuser to run for binder.

Is this overly cautious do you think?

PR #7051, opened 2025-10-29
Fix: correct divergence of branches

agoose77:
Follows https://github.com/2i2c-org/infrastructure/pull/7050


agoose77:
I cancelled the CI in the last PR. As this touches core deployer code, it should trigger the same jobs (all hubs)

PR #7058, opened 2025-10-31
MAAP cluster (staging) - Enhanced  S3FS configuration and custom homepage git repo

bsatoriu:
1. Add homepage configuration with git repository details.
2. Added additional volume mounts and containers for S3FS integration.

### DIFF
--- a/config/clusters/maap/staging.values.yaml
+++ b/config/clusters/maap/staging.values.yaml
@@ -39,32 +39,32 @@ jupyterhub:
         mountPropagation: HostToContainer
         readOnly: true
     extraContainers:
-      - name: s3fs
-        image: mas.dit.maap-project.org/root/che-sidecar-s3fs:2i2c
-        securityContext:
-          privileged: true
-        resources:
-          limits:
+    - name: s3fs
+      image: mas.dit.maap-project.org/root/che-sidecar-s3fs:2i2c
+      securityContext:
+        privileged: true
+      resources:
+        limits:
             # Best effort only. No more than 1 CPU, and if postgres uses more than 512M, restart it
-            memory: 512Mi
-            cpu: 1.0
-        volumeMounts:
-            - name: s3fs-volume
-              mountPath: /my-public-bucket
-              subPath: my-public-bucket
-              mountPropagation: Bidirectional
-            - name: s3fs-volume
-              mountPath: /my-private-bucket
-              subPath: my-private-bucket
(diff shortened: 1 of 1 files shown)


config/clusters/maap/staging.values.yaml
```diff
@@ -8,11 +8,63 @@ jupyterhub:
...
+      gitRepoUrl: https://github.com/MAAP-Project/maap-hub-homepage
   singleuser:
     extraEnv:
       SCRATCH_BUCKET: s3://maap-scratch-staging/$(JUPYTERHUB_USER)
     nodeSelector:
       2i2c/hub-name: staging
+    storage:
+      extraVolumeMounts:
+      - name: s3fs-volume
+        mountPath: /home/jovyan/my-private-bucket
```
riverma, line 22: I see `jovyan` hardcoded here. Should we use `$(JUPYTERHUB_USER)` instead?

bsatoriu: This 'jovyan' path follows a common 2i2c convention as shown here -- https://github.com/2i2c-org/infrastructure/blob/main/config/clusters/maap/staging.values.yaml#L47

$(JUPYTERHUB_USER) is the username of the logged in 2i2c user. You can confirm this by printing out the environment variables in a 2i2c workspace.

riverma: Thanks for the clarification - understood.

config/clusters/maap/staging.values.yaml
```diff
@@ -8,11 +8,63 @@ jupyterhub:
...
+        mountPropagation: HostToContainer
+        readOnly: true
+    extraContainers:
+    - name: s3fs
+      image: mas.dit.maap-project.org/root/che-sidecar-s3fs:2i2c
+      securityContext:
+        privileged: true
+      resources:
+        limits:
+            # Best effort only. No more than 1 CPU, and if postgres uses more than 512M, restart it
```
yuvipanda, line 48: You must still set a request, even if it's very low: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/:

> If you specify a limit for a resource, but do not specify any request, and no admission-time mechanism has applied a default request for that resource, then Kubernetes copies the limit you specified and uses it as the requested value for the resource.

yuvipanda:
Just merging so this can be iterated on faster while we are mostly at jupytercon

PR #7068, opened 2025-11-06
Move auth_state_hook from home page into infrastructure

bsatoriu:
New config settings based on a similar [method used by earthscope](https://github.com/2i2c-org/infrastructure/blob/0046e14a68d7af9e353c494ee6ad39beb0ce970a/config/clusters/earthscope/common.values.yaml#L29).

TODO: move to common.values.yaml once tested in staging environment.

### DIFF
--- a/config/clusters/maap/staging.values.yaml
+++ b/config/clusters/maap/staging.values.yaml
@@ -281,8 +281,6 @@ jupyterhub:
         - GPU:T4
     extraConfig:
       001-username-claim: |
-        from oauthenticator.generic import GenericOAuthenticator
-
         def populate_token(spawner, auth_state):
           # For our deployment-service-check health check user, there is no auth_state.
           # So these env variables need not be set.
@@ -296,8 +294,6 @@ jupyterhub:
             })
 
         c.Spawner.auth_state_hook = populate_token
-
-        c.JupyterHub.authenticator_class = GenericOAuthenticator
       00-volumes-and-volume-mounts-as-dict: |
         # The base jupyterhub config in zero-to-jupyterhub defines
         # volumes and volume_mounts as lists.


config/clusters/maap/staging.values.yaml
```diff
@@ -280,6 +280,24 @@ jupyterhub:
         - CPU:XXXL
         - GPU:T4
     extraConfig:
+      001-username-claim: |
+        from oauthenticator.generic import GenericOAuthenticator
```
yuvipanda: This is not necessary, we already set this elsewhere

config/clusters/maap/staging.values.yaml
```diff
@@ -280,6 +280,24 @@ jupyterhub:
...
+            spawner.environment.update({
+              "MAAP_PGT": f"jwt:{auth_state.get("id_token", "")}",
+              "KC_ACCESS_TOKEN": auth_state.get("access_token", ""),
+              "KC_ID_TOKEN": auth_state.get("id_token", ""),
+              "KC_REFRESH_TOKEN": auth_state.get("refresh_token", "")
+            })
+
+        c.Spawner.auth_state_hook = populate_token
+
+        c.JupyterHub.authenticator_class = GenericOAuthenticator
```
yuvipanda: This is also not required

bsatoriu: @yuvipanda how can we set our environment variables without a custom `auth_state_hook`? This is how earthscope is setting their environment variables: https://github.com/2i2c-org/infrastructure/blob/0046e14a68d7af9e353c494ee6ad39beb0ce970a/config/clusters/earthscope/common.values.yaml#L29

yuvipanda: Just the two lines I commented on need to be removed. The rest shall stay. Auth0 uses a custom authenticator, you don’t need just that bit

bsatoriu: Okay, to be clear, these are the two lines to remove?

[284](https://github.com/bsatoriu/infrastructure/blob/7ae6568bc86ea8679d0fa678d2baf074e8fa8cf8/config/clusters/maap/staging.values.yaml#L284)
[300](https://github.com/bsatoriu/infrastructure/blob/7ae6568bc86ea8679d0fa678d2baf074e8fa8cf8/config/clusters/maap/staging.values.yaml#L300)

Issue #7021, opened 2025-10-24
Trigger the server startup alerts when at least two servers have failed


GeorgianaElena:
### Context

The server startup alert is noisy. For this reason it causes fatigue because we always have to investigate it as it can be the symptom of an outage. But in most of the cases it isn't.

### What we need to do

- We should create a new alerts that triggers when there are at least two servers failing. This should ignore the cases where the servers failed to start because of nodes being slow to come up and other similar reasons. 
- We should keep the current alert as a P2 that we simply close if we can start a new server on the hub it got triggered on

### Definition of Done

- [ ] The new alarm was implemented and the current one changed
- [ ] The alarm been tested
- [ ] The alarm has been documented

Issue #7022, opened 2025-10-24
[Spike: 4h] Change the way we group alerts


GeorgianaElena:
### Context

Currently we have the alerts grouped per category as described in https://infrastructure.2i2c.org/topic/monitoring-alerting/alerting/#configuration.

Each group is mapped to an alertmanager receiver that is assigned a PD service.

When assigning an escalation policy, you assign it to a service. And such service is grouping together all types of priorities. This means, we cannot assign escalation policies based on priorities, only based on service.

### What we need to do

We want to be able to have different escalation policies for different types of priorities (so that we only assign the tech lead to P1 alerts aka outages), as well as having services that make sense to users when presented on the status page.

### Definition of Done

- [ ] Find a new grouping that makes sense
- [ ] Implement it
- [ ] Document it

PR #7120, opened 2025-11-14
Add two server failure alert

GeorgianaElena:
Fixes https://github.com/2i2c-org/infrastructure/issues/7021

### DIFF
--- a/helm-charts/support/values.jsonnet
+++ b/helm-charts/support/values.jsonnet
@@ -71,14 +71,12 @@ function(VARS_2I2C_AWS_ACCOUNT_ID=null)
                                             ) {
     alert: 'Two servers failed to start in the last 30m',
     expr: |||
-      round(
-        increase(
-          (
-            max by (namespace) (
-              jupyterhub_server_spawn_duration_seconds_count{status="failure"}
-            )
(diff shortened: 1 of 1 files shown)


helm-charts/support/values.jsonnet
```diff
@@ -217,9 +242,13 @@ function(VARS_2I2C_AWS_ACCOUNT_ID=null)
               name: 'Server Startup Failure',
               rules: [
                 makeServerStartupFailureAlert(
-                  'Outage alert: Server Startup failed: cluster %s hub:{{ $labels.namespace }}' % [cluster_name],
+                  'A server failed to start: cluster %s hub:{{ $labels.namespace }}' % [cluster_name],
                   'same day action needed'
```
yuvipanda, line 246: Let's reduce this to 'same week'?

GeorgianaElena: Makes sense, ty! I'll tackle this as part of https://github.com/2i2c-org/infrastructure/issues/7022 as I feel like the whole priorities structure needs a bit of a cleanup.

Right now this is a P2 and we should document that the expectation for this alert should be: if I can start a server, mark it as resolved.

PR #7134, opened 2025-11-19
ghg: Add "sdap-dev-zarr"

siddharth0248:
Request for adding the following bucket in the 'nasa-ghg-hub-prod-extra-user-policy'  policy

1."arn:aws:s3:::sdap-dev-zarr",
2."arn:aws:s3:::sdap-dev-zarr/*"


siddharth0248:
@yuvipanda

PR #7156, opened 2025-11-24
Migrate 2i2c/temple to temple/prod

agoose77:
This PR tracks the work to migrate temple from GCP to AWS as a dedicated hub. As described in the docs, it prioritises a cloud-agnostic simple approach that was sufficient for copying within the same region for the `temple` cluster. 

In future, we may wish to tweak this if larger disks or lack of region proximity means that this approach does not scale well.

### Initial phase
- [x] Copy CILogon credentials
- [x] Initial clone of user-storage to AWS

### Reconciliation phase
- [x] Secondary clone of user-storage to AWS before commissioning
- [x] Remove `temple-shared`, add `temple.2i2c.cloud` to `prod`

### DIFF
--- a/docs/howto/filesystem-management/data-transfer-external.md
+++ b/docs/howto/filesystem-management/data-transfer-external.md
@@ -108,7 +108,9 @@ Run this section on the destination cluster.
    ```bash
    kubectl -n <DEST-HUB> get service/openssh-service
    ```
-(migrate-external:setup-src)=
+
+   (migrate-external:setup-src)=
+
 ## Setting up the source server
 
 ```{important}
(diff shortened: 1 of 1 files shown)


config/clusters/temple/common.values.yaml
```diff
@@ -83,10 +83,20 @@ jupyterhub:
         requests:
           display_name: Resource Allocation
           choices:
+            mem_2_gb:
```
yuvipanda: The teaching hub should not have a profileList, because it shouldn't offer users these complex choices. Let's:

1. split the stuff that's common to the advanced and research hubs into their own common.yaml file (perhaps non-teaching-common)
2. Leave the rest of the stuff here in common.yaml

This allows us to still share whatever is common, while also not sharing user facing configuration that *must* differ.

agoose77: I did wonder about that, thanks!

yuvipanda: https://2i2c.freshdesk.com/a/tickets/4432 has a little more info

PR #7176, opened 2025-11-25
[*] Bump `jupyterhub-home-nfs` version

agoose77:
This PR bumps the `jupyterhub-home-nfs` version so that we can define `extraContainers`, as is required by #6550


GeorgianaElena (commented):
You can also just bump it for temple to validate that it works. That is, if you're not confident in that already

agoose77:
> You can also just bump it for temple to validate that it works. That is, if you're not confident in that already

Yes! Thanks @GeorgianaElena, I did already deploy transiently to Temple `prod`. I've since reverted it (via a regular CI deployment in the merged PR). But it's fab to have that mechanism.

PR #7217, opened 2025-12-02
Increase iops for MAAP home directory + increase quota for 1 user

yuvipanda:
We got a few alerts about saturation + a user is copying a large amount of data over. We may want to turn this down later, but it's roughly 10$ a month extra for this 2k iops bump.

Increase quota for one user who is doing the transfer.


yuvipanda:
Don't merge without approval from @wildintellect

wildintellect:
Go ahead and do it for now, we'll see how the rest of this transfer impacts the io before tuning it again.

PR #7219, opened 2025-12-02
Promote MAAP staging hubs to prod

bsatoriu:
Synchronize the latest maap `staging.values.yaml` updates to `prod.values.yaml`.

### DIFF
--- a/config/clusters/maap/staging.values.yaml
+++ b/config/clusters/maap/staging.values.yaml
@@ -44,6 +44,7 @@ jupyterhub:
               description: Pangeo based notebook with a Python environment
               kubespawner_override:
                 image: mas.dit.maap-project.org/root/maap-workspaces/2i2c/pangeo:develop
+                image_pull_policy: Always
                 init_containers:
                 - name: jupyterhub-gitpuller-init
                   image: public.ecr.aws/nasa-veda/jupyterhub-gitpuller-init:97eb45f9d23b128aff810e45911857d5cffd05c2
@@ -64,6 +65,7 @@ jupyterhub:
               description: JupyterHub environment with many R geospatial libraries pre-installed
               kubespawner_override:
                 image: mas.dit.maap-project.org/root/maap-workspaces/2i2c/r:develop
+                image_pull_policy: Always
                 init_containers:
                 - name: jupyterhub-gitpuller-init
                   image: public.ecr.aws/nasa-veda/jupyterhub-gitpuller-init:97eb45f9d23b128aff810e45911857d5cffd05c2
@@ -84,6 +86,7 @@ jupyterhub:
               description: Pangeo based notebook with a Python environment and isce3
               kubespawner_override:
                 image: mas.dit.maap-project.org/root/maap-workspaces/2i2c/isce3:develop
+                image_pull_policy: Always
                 init_containers:
                 - name: jupyterhub-gitpuller-init
                   image: public.ecr.aws/nasa-veda/jupyterhub-gitpuller-init:97eb45f9d23b128aff810e45911857d5cffd05c2


config/clusters/maap/prod.values.yaml
```diff
@@ -5,45 +5,236 @@ userServiceAccount:
...
+    2i2c:
+      add_staff_user_ids_to_admin_users: false
+    homepage:
+      gitRepoBranch: master
+      gitRepoUrl: https://github.com/MAAP-Project/maap-hub-homepage
   singleuser:
     extraEnv:
       SCRATCH_BUCKET: s3://maap-scratch-prod/$(JUPYTERHUB_USER)
+      MAAP_API_HOST: api.maap-project.org
+      DOCKERIMAGE_PATH_DEFAULT: mas.maap-project.org/root/maap-workspaces/custom_images/maap_base:v5.0.0
```
yuvipanda, line 18: This will set this as the environment variable no matter what image is used. Is that what was expected? In the last PR, I saw this was set to be the same as the name of the image, in which case it should use `$(JUPYTER_IMAGE)` as the value.

config/clusters/maap/common.values.yaml
```diff
@@ -85,6 +146,59 @@ jupyterhub:
...
+        readOnly: true
+    extraContainers:
+    - name: s3fs
+      image: mas.dit.maap-project.org/root/che-sidecar-s3fs:2i2c
+      image_pull_policy: Always
+      securityContext:
+        privileged: true
+      resources:
+        limits:
+            # Best effort only. No more than 1 CPU, and if postgres uses more than 512M, restart it
```
yuvipanda: Are you using postgres here? This comment seems wrong

config/clusters/maap/prod.values.yaml
grallewellyn: Okay, extracted out DOCKERIMAGE_PATH_BASE_IMAGE!

config/clusters/maap/prod.values.yaml
```diff
@@ -5,45 +5,236 @@ userServiceAccount:
...
+      WORKSPACE_BUCKET: maap-ops-workspace
     nodeSelector:
       2i2c/hub-name: prod
+    profileList:
```
yuvipanda, line 26: Normally, we would like to keep `profileLists` in `common.yaml`, and use the *same image* in staging and prod. The staging here is primarily for testing infrastructure changes, and we (2i2c) would like to generally keep it the exact same as prod. So that if we have tested something on staging, we're 99% confident it would work in prod.

having different images in staging and prod could cause problems here, in case the images being different causes failure when migrating. It could also cause the other parts of profile Lists (such as resource config) to drift out of sync between these two.

However, we also recognize that you want to probably test out different images as you're onboarding an existing userbase to this hub, and want to be flexible.

So I see two paths forward:

1. Use the same image tags for staging and prod, and put it in common.yaml. Image testing happens purely via unlisted choice. This is the preferred way, and also where we should go long term.
2. If (1) doesn't fit with your existing workflows for building images, leave a block comment above the `profileList` config in staging and prod, documenting that it's duplicated, and that whoever is modifying it should take care to make sure that the *only* differences between these two should be the image tags, and everything else should be kept in sync manually. We can then revisit this in 3-6 months, after the initial migration is completed and the pace of image changes has changed.

I wanna unblock y'all asap, so while I have a preference for (1) happy to do either.

sujen1412: Sorry jumping into this conversation as I come back from leave. 

Let me know if I am phrasing this correctly - 
You are saying that `staging` and `prod` are meant for infrastructure testing and everything else remains the same. In that case, we (MAAP) as tenants of this infrastructure should be deploying 3 versions of your `prod` configuration for our own customers and venues (DIT, UAT and OPS). The tenant should not need to worry about your changes in your staging environment.  
We should be able to deploy multiple `2i2c prod` environments with different MAAP configurations for our testing. 

Does that make sense?

sujen1412: On MAAP, the DIT, UAT and OPS venues come with their associated deployments of the API and data processing clusters which impact the jupyter extensions used in the images. So in terms of testing, we are not just testing the images, but also entire the deployment venue which is isolated in its own cloud env.

grallewellyn: I added a block comment above profileList and we would like to go with option 2

yuvipanda: Thanks @grallewellyn! I've retitled the PR slightly and merged this!

@sujen1412 I opened https://github.com/2i2c-org/infrastructure/issues/7233 to split off the other conversation so we don't lose track of it!

PR #7237, opened 2025-12-04
Remove GitHubOAuthenticator

bsatoriu:
Removed GitHubOAuthenticator configuration since it appears to be overriding our Keycloak login service.

### DIFF
--- a/config/clusters/maap/common.values.yaml
+++ b/config/clusters/maap/common.values.yaml
@@ -47,7 +47,7 @@ jupyterhub:
     allowNamedServers: true
     config:
       JupyterHub:
-        authenticator_class: github
+        authenticator_class: generic-oauth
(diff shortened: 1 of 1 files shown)


config/clusters/maap/common.values.yaml
```diff
@@ -48,12 +48,6 @@ jupyterhub:
     config:
       JupyterHub:
         authenticator_class: github
```
GeorgianaElena:
```suggestion
        authenticator_class: generic-oauth
```

GeorgianaElena: The `authenticator_class` should be set here, under the `JupyterHub` key to be effective.
I believe this is why the allowed orgs config was beeing picked up

PR #7336, opened 2026-01-07
[ucmerced] Create new ucmerced cluster

agoose77:
This PR will (as-yet this configuration has _not_ been applied) define the cluster for ucmerced that they will be migrated to.

### Follow Up
- [ ] Update TTLs on DNS for ucmerced


eksctl/libsonnet/cluster.jsonnet
```diff
@@ -263,7 +264,8 @@ local escapeName(name) = std.strReplace(name, '.', '-');
     notebookCPUInstanceTypes,
     notebookGPUNodeGroups=[],
     daskInstanceTypes=[],
-    nodeGroupGenerations=[]
+    nodeGroupGenerations=[],
+    regionSize=3
```
agoose77, line 268: This enables us to specify vary the number of zones.

Issue #7304, opened 2025-12-17
[Support] Prometheus data & cost service enablement for EarthScope


aprilmj:
### Context

The EarthScope/GeoLab team manages their own usage dashboards that give them detailed insights about who's using what resources. They want to incorporate our cloud cost and Prometheus data into those dashboards.

### What we need to do

Send link and credentials for Prometheus data and cost service to EarthScope team 

### Definition of Done

- [x] The feature/service is technically complete
- [x] The feature/service been tested with one or more users (if applicable)
- [x] The feature/service follows our [definition of minimally documented](https://compass.2i2c.org/product-and-services/documentation/)

Documentation and wider communication will be useful if we do this in a way that's replicable for other communities. TBD whether that's appropriate.

yuvipanda:
The outcomes here are twofold:

1. Have a way to give communities credentials (safely) to access their prometheus data, *along with warnings about PII exposure* if they lose those credentials
2. Have a way to give communities credentials to access their jupyterhub cost monitoring api endpoint

yuvipanda:
Second set of credentials sounds great @jnywong! Let's document this as something we offer too.

Can we do something similar for them to hit our cost monitoring API endpoint too?

jnywong:
I've done this for prometheus access in https://github.com/2i2c-org/docs/pull/290.

I think it's possible to use the same creds and place that in front of an ingress for the cost-monitoring endpoint, to follow in another PR.

jnywong:
In blocked pending response from the community.

- [ ] Remaining action is to securely share access credentials after receiving a public key from the community following instructions in https://infrastructure.2i2c.org/sre-guide/support/prometheus-credentials/

yuvipanda:
@jnywong can you update https://infrastructure.2i2c.org/sre-guide/support/prometheus-credentials/ to refer to bitwarden send docs rather than age?

PR #7354, opened 2026-01-09
Allow additional user credentials to access Prometheus

jnywong:
Ref https://github.com/2i2c-org/infrastructure/issues/7304.

The main change here is the extra nesting of a list under the `users` key in `enc-support.secret.values.yaml`

```
   prometheusIngressAuthSecret:
     users:
       - username: <output of pwgen -s 64 1>
         password: <output of pwgen -s 64 1>
       - username: <output of pwgen -s 64 1>
         password: <output of pwgen -s 64 1>   
```

The documentation outlines the process for sharing these credentials securely to a community, since PII is at risk.

### DIFF
--- a/config/clusters/2i2c-aws-us/enc-support.secret.values.yaml
+++ b/config/clusters/2i2c-aws-us/enc-support.secret.values.yaml
@@ -1,23 +1,24 @@
 prometheusIngressAuthSecret:
+  #ENC[AES256_GCM,data:Y9YBU79Th7DZ0hZW8FqfP2BmmTnrliVrwzawOqAQ73F89VR8kfmuRZwZy/39X8DbhTL2WQQuFszZzjkWG1wdtedNYEjF7Ew0HoNDh+Crx2R/YQ+qywP+AczlDWKFr6gTBIrdGHbuxkYY0ArTAZxCyymMSewkhSXJRfAMKLX6oGHUAsTbLi0DwFcQ+z5TxnTeP5O21Af1GBDiXEuZ,iv:tUaaHs1tpvy23Mot7UDvIAyckiEEYUShurGIMThSGzo=,tag:UuBoELhRrFysh2ckflH3Yw==,type:comment]
   users:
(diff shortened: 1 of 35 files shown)


helm-charts/support/values.schema.yaml
```diff
@@ -106,20 +106,29 @@ properties:
 
   prometheusIngressAuthSecret:
     type: object
-    additionalProperties: false
+    additionalProperties: true
```
jnywong: I couldn't get the validator to stop complaining for properties defined in the array so set this to true. Not sure what the best practise here is, despite staring at the json schema docs.

docs/sre-guide/support/prometheus-credentials.md
```diff
@@ -0,0 +1,51 @@
...
+   prometheusIngressAuthSecret:
+     users:
+       - username: <output of pwgen -s 64 1>
+         password: <output of pwgen -s 64 1>
+       - username: <output of pwgen -s 64 1>
+         password: <output of pwgen -s 64 1>         
+   ```
+
+   ```{tip}
+   Make sure you place the extra user credentials **under** the first entry, since the first entry is reserved for internal 2i2c purposes to [register with our central grafana](/hub-deployment-guide/deploy-support/register-central-grafana.md).
```
GeorgianaElena, line 21: I think it would also be wise to leave a comment about which community a particular secret has been shared with, so we can clean them up or rotate them with more confidence

helm-charts/support/values.schema.yaml
GeorgianaElena: I believe it's because of the remaining outdated config in support/values.yaml
I updated it in https://github.com/2i2c-org/infrastructure/pull/7354/commits/8303f123cf6f8dad756a8441e0e798f96c4e3696

GeorgianaElena (approved):
Thank you @jnywong ✨ 🚀 

I added two commits related to the `validate` command and left one extra comment, that's more of an open question.

docs/sre-guide/support/prometheus-credentials.md
jnywong: Yes great suggestion! 💖

helm-charts/support/values.schema.yaml
jnywong: https://github.com/2i2c-org/infrastructure/pull/7354#discussion_r2681498326 omg thanks for catching that!!! My head was going round in circles cos I knew my schema was right but turns out I just had the wrong input 😄

PR #7405, opened 2026-01-14
Allow all our hubs to delete non-empty dirs

yuvipanda:
Tested successfully just on MAAP. It still has a confirmation dialog.

Ref https://github.com/NASA-IMPACT/veda-jupyterhub/issues/96

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -714,6 +714,10 @@ jupyterhub:
           # https://github.com/jupyterlab/jupyterlab/issues/11304#issuecomment-945466766
           ContentsManager:
             allow_hidden: true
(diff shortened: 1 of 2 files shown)


helm-charts/basehub/values.yaml
```diff
@@ -714,6 +714,10 @@ jupyterhub:
           # https://github.com/jupyterlab/jupyterlab/issues/11304#issuecomment-945466766
           ContentsManager:
             allow_hidden: true
+          # Allow deletion of directories with files in them, after a confirmation dialog
+          # https://github.com/NASA-IMPACT/veda-jupyterhub/issues/96
+          FileContentsManager:
+            alwas_delete_dir: true
```
jnywong:
```suggestion
            always_delete_dir: true
```

PR #7407, opened 2026-01-14
maap: Staging now points to uat api

grallewellyn:
We agreed this makes the most sense in a meeting for users who want to start working with OGC
DIT API is not stable enough for staging to have users


yuvipanda:
With the caveat that we don't expect users to be on staging hubs at all, as their primary purpose is to test infrastructure changes. That means they may be down, have their disks wiped, etc without notice. We can easily add more hubs in this cluster if you desire.

PR #7438, opened 2026-01-19
nasa smce clusters: re-generate deployer credentials

jnywong:
Closes https://github.com/2i2c-org/infrastructure/issues/7328


yuvipanda:
Can you unpin and rebase against latest master? The error about not being able to push to upstream makes me suspect that may fix it.

PR #7444, opened 2026-01-20
Adding pace and ml images to CryoCloud

tsnow03:
Added 3 new images and updated some others. Changed the default to 4GB.

Please check that the slug for ml-notebook environment under the Tesla is supposed to say pytorch. It's a tensorflow environment so I'm not sure if there is another slug word that should be used there.


config/clusters/nasa-cryo/common.values.yaml
```diff
@@ -236,11 +249,18 @@ basehub:
...
+                description: ML environment with PyTorch from Pangeo
                 default: true
                 slug: pytorch
                 kubespawner_override:
-                  image: quay.io/pangeo/pytorch-notebook:2025.05.22
+                  image: quay.io/pangeo/pytorch-notebook:2025.12.30
+              tensorflow:
+                display_name: Machine Learning Image
+                description: ML environment with TensorFlow, Keras, Scikit from Pangeo
+                slug: pytorch
```
agoose77:
```suggestion
                slug: tensorflow
```
Slugs are not required, but should be unique.

PR #7454, opened 2026-01-20
Add homedir archive bucket to NMFS-openscapes

ateucher:
This adds a home directory archive bucket to the NMFS prod environment, based on the model for NASA-Openscapes (https://github.com/2i2c-org/infrastructure/blob/4efe1457efd49a3f2d9196a738c0873c110d4e40/terraform/aws/projects/openscapeshub.tfvars#L53-L56). Please let me know if there is a more formal process I should go through to request this.

Thanks!

### DIFF
--- a/terraform/aws/projects/nmfs-openscapes.tfvars
+++ b/terraform/aws/projects/nmfs-openscapes.tfvars
@@ -7,6 +7,7 @@ default_budget_alert = {
 }
 
 enable_jupyterhub_cost_monitoring = true
+enable_jupyterhub_cost_tags       = true
 
 disable_cluster_wide_filestore = true
 ebs_volumes = {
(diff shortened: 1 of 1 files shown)


agoose77:
I'm AFK at the moment, but I think the context for this is 
- https://2i2c.freshdesk.com/a/tickets/4804
- https://2i2c.freshdesk.com/a/tickets/4635

terraform/aws/projects/nmfs-openscapes.tfvars
```diff
@@ -7,6 +7,7 @@ default_budget_alert = {
 }
 
 enable_jupyterhub_cost_monitoring = true
+enable_jupyterhub_cost_tags       = true
```
agoose77, line 10: This was effectively already on due to an earlier deployment of the cost monitoring system. Adding this line ensures that we do not clobber this.

agoose77:
Thanks for the explanation @ateucher, and for creating the PR alongside it! I've effected these Terraform changes, so you should be able to use the bucket!

ateucher:
Thank you so much @agoose77!

Issue #6219, opened 2025-06-13
EKS control plane logging: enable on all clusters, document and analyse its impact on cost


yuvipanda:
To help debug outages better on AWS, we should have logging on AWS match what we have on GCP. We currently only have logs to immediate logs - as soon as a pod restarts, they are gone.

https://github.com/2i2c-org/infrastructure/issues/6756 turns on control plane logging for EKS clusters sets it up to keep the logs for 7 days. We need to document this is validate that the impact on cost is negligible.

## Definition of done
- [ ] Write a timeboxed amount of docs on how to access these logs
- [ ] Observe the increased cost from these logs (should be negligible, but we must verify)
- [ ] Add any changes to docs.2i2c.org

PR #7465, opened 2026-01-21
[deployer] Get a shell under a SSO account with no context-switching

GeorgianaElena:
Ref: [6219](https://github.com/2i2c-org/infrastructure/issues/6219)
This adds a new `deployer` command that allows getting the right AWS credentials for an account under a SSO account (either the 2i2c one or the SMDC one) with no context-switching or manual copy-pasting.

This is different from the previous `exec aws shell` command because it doesn't require maintaining AWS credentials directly for an account, only an SSO profile entry in `~/.aws/config`

It can be called with:

Screenshot 2026-01-21 at 17 44 03

Or directly:

Screenshot 2026-01-21 at 17 44 48

### DIFF
--- a/docs/topic/access-creds/cloud-auth.md
+++ b/docs/topic/access-creds/cloud-auth.md
@@ -83,9 +83,39 @@ check out [Firefox Multi-Account Containers](https://addons.mozilla.org/en-US/fi
 #### Access AWS from your terminal
 
 To use programs like `eksctl`, or `terraform`, you need to get AWS credentials that
-can be accessed from the terminal on your computer. When set up with AWS SSO, the portal
-easily provides access with _time limited_ credentials. These are valid only for **60 minutes**,
-and will need to be refreshed with new sets whenever that time is up.
+can be accessed from the terminal on your computer.
+
+##### Programmatically
+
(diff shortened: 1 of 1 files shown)


deployer/dev_commands/exec/aws/aws_app.py
```diff
@@ -77,10 +153,116 @@ def shell(
...
+    env = get_role_creds_as_env(profile, account_id, role, access_token)
+
+    print_colour(
+        f"✅ New shell ready for account {account_name}:{account_id}, role {role}"
+    )
+    print_colour(
+        "💡 Run 'eksctl get cluster --region=<cluster_region>' to verify", "yellow"
+    )
+
+    subprocess.check_call([os.environ["SHELL"], "-l"], env=env)
```
agoose77: May I tweak this to support specifying a command to pass to the shell? I have a suggestion to use the same pattern in `deployer/dev_commands/exec/azure.py`

GeorgianaElena: Added https://github.com/2i2c-org/infrastructure/pull/7465/commits/3854ade66d85f53d287e884b923948efbb99178e

PR #7500, opened 2026-01-27
[earthscope, *] Refactor authenticator configuration

agoose77:
The Earthscope hub has some bespoke configuration logic that we've since upstreamed in 
- https://github.com/jupyterhub/oauthenticator/pull/719/changes
- https://github.com/jupyterhub/oauthenticator/pull/735/changes

In order to simplify the administration of this hub, we should remove it.

### DIFF
--- a/config/clusters/earthscope/common.values.yaml
+++ b/config/clusters/earthscope/common.values.yaml
@@ -61,8 +61,9 @@ basehub:
 
           c.Spawner.auth_state_hook = populate_token
 
-          c.JupyterHub.authenticator_class = GenericOAuthenticator
       config:
+        JupyterHub:
+          authenticator_class: generic-oauth
         GenericOAuthenticator:
           allowed_scopes:
             # This allows EarthScope to control who can login to the hub


config/clusters/earthscope/common.values.yaml
```diff
@@ -148,6 +94,7 @@ basehub:
           - google-oauth2|108516754518362587599   # Yuvi
           - google-oauth2|117859169473992122769   # Georgiana
           - google-oauth2|117322480787655244438   # Jenny
+          - google-oauth2|107058254054417943477   # Angus
```
agoose77, line 98: Also, add me as an admin!

config/clusters/earthscope/common.values.yaml
```diff
@@ -129,16 +85,6 @@ basehub:
...
           manage_groups: true
-        CILogonOAuthenticator:
-          allowed_idps:
-            http://github.com/login/oauth/authorize:
-              default: true
-              username_derivation:
-                username_claim: preferred_username
-            http://google.com/accounts/o8/id:
-              username_derivation:
-                username_claim: email
```
agoose77, line 140: My understanding is that we switched away from CILogon, so this is legacy config

Issue #7468, opened 2026-01-21
Support installing helm v4 in https://github.com/jupyterhub/action-k3s-helm


yuvipanda:
To support https://github.com/2i2c-org/initiatives/issues/4, we want to enable zero to jupyterhub to be tested on Helm v3 and Helm v4. To do so, https://github.com/jupyterhub/action-k3s-helm needs to be able to provision helm v4 as well. It currently can't do so.

## Definition of dfone

- [ ] https://github.com/jupyterhub/action-k3s-helm can be optionally set up to support helm v4
- [ ] This support is merged and documented

GeorgianaElena:
Release happened https://github.com/jupyterhub/action-k3s-helm/pull/131

PR #7523, opened 2026-01-29
fix: pin Helm to <=3.17.0

agoose77:
Helm 3.17.1 introduces new null behaviour: https://github.com/helm/helm/issues/30587#issuecomment-3717495744

In Helm 4, it is recognised that this introduces a new regression, which is pertinent to #7468.

_This_ PR pins Helm for the Nix flake dev-shell so that we can avoid this. In CI, we're still using v3.14.4 IIRC.

### DIFF
--- a/config/clusters/projectpythia-binder/binderhub.values.yaml
+++ b/config/clusters/projectpythia-binder/binderhub.values.yaml
@@ -120,14 +120,11 @@ binderhub-service:
       - binder.projectpythia.org
   dockerApi:
     nodeSelector:
+      hub.jupyter.org/node-purpose:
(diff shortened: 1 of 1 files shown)


config/clusters/projectpythia-binder/binderhub.values.yaml
```diff
@@ -120,14 +120,11 @@ binderhub-service:
       - binder.projectpythia.org
   dockerApi:
     nodeSelector:
+      hub.jupyter.org/node-purpose:
```
agoose77, line 123: This was not the bug!

PR #7549, opened 2026-02-02
fix: patch LD_LIBRARY_PATH on Nix

agoose77:
This PR applies some fixes to the Nix environment after I learned more about best-practices.

Most of `flake.nix` is just setting up which packages we want! But, there's a bit of boilerplate to getting the virtualenv up and running. The proper Nix way to handle packages is to package them as Nix derivations. But, I'd prefer to use the flake to bootstrap a development environment where I'm not the sole maintainer keeping a load of Python package derivations up to date. So, instead we just use a venv.

Python wheels make some assumptions about the runtime (manylinux) that don't always hold on esoteric systems, so we have to do a bit of tweaking to expose the manylinux libs to the Python interpreter. Previously I exposed this via the LD_LIBRARY_PATH variable, but this is a sledgehammer that affects every program on the system (even on Nix, some packages listen to this). This PR replaces that approach with a shim that wraps `python3.XX` with the env var, so it's scoped only to the Python binary.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -25,6 +25,7 @@
       };
       inherit (pkgs) lib;
 
+      # Configure packages that need additional deps
       gdk = pkgs.google-cloud-sdk.withExtraComponents (with pkgs.google-cloud-sdk.components; [
         gke-gcloud-auth-plugin
       ]);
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -59,23 +64,45 @@
...
+
+          # Drop bad env vars on activation
+          postShellHook = unwantedEnvPreamble;
+
+          # Setup venv by patching interpreter with LD_LIBRARY_PATH
+          # This is required because ld does not exist on Nix systems
+          postVenvCreation = let
+            # Prepare to wrap venv with LD_LIBRARY_PATH
+            libraryEnvVar =
+              if (!pkgs.stdenv.isDarwin)
```
agoose77, line 90: I don't think this actually works on Mac atm, because manylinux is a Linux thing.

PR #7572, opened 2026-02-03
feat: allow kubelet config customisation

agoose77:
This PR makes it possible to define kubelet configuration for our eksctl-managed clusters.


eksctl/libsonnet/cluster.jsonnet
```diff
@@ -27,7 +28,8 @@ local lowerCaseLetter(i) = std.char(97 + i);
     maxSize,
     extraLabels={},
     extraTaints=[],
-    extraTags={}
+    extraTags={},
+    extraKubeletConfig=null,
```
agoose77, line 32: No string feelings about using `null` vs `{}` here.

PR #7590, opened 2026-02-05
Update to pangeo-notebook-veda-image:2025.12.30-v1

jsignell:
This PR updates all NASA-IMPACT hubs to contain pangeo notebook images based off of 2025.12.30 as requested in https://github.com/NASA-IMPACT/veda-analytics/issues/192

I think it's still @wildintellect, @smk0033, and @freitagb who represent the various hubs and will be able to say whether this is a good time for an update or whether we want to hold off for a workshop or something.

Note that I am pretty sure MAAP doesn't actually use most of what's in common-values. They have overrides for everything in https://github.com/2i2c-org/infrastructure/blob/e5e2c988ecfb52aeb1abcb06880afee633a23a4e/config/clusters/maap/prod.values.yaml#L42-L46

### DIFF
--- a/config/clusters/disasters/common.values.yaml
+++ b/config/clusters/disasters/common.values.yaml
@@ -115,7 +115,7 @@ jupyterhub:
               display_name: Modified Pangeo Notebook
               description: Pangeo based notebook with a Python environment
               kubespawner_override:
-                image: public.ecr.aws/nasa-veda/pangeo-notebook-veda-image:2025.08.14-v2
+                image: public.ecr.aws/nasa-veda/pangeo-notebook-veda-image:2025.12.30-v1
                 init_containers:
                 # this container uses nbgitpuller to mount https://github.com/NASA-IMPACT/veda-docs/ for user pods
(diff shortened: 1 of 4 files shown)


jsignell:
My read is the common-values are meant to be common across staging and prod for a particular VEDA instance. So it seems like MAAP could just move a lot of their config out of the staging and prod specific files and up into their common-values file.

grallewellyn:
With the current approach we don't need to modify the deploy configs for staging because they already point to the develop tag which is what we update during our development cycle 
And we would need to modify the production configs for a release anyway

kyle-lesinger:
> @kyle-lesinger represents Disasters - which if I understand correctly uses a custom image most of the time (separate task to add their custom image to the menu as the default). So should be good there.
> 
> MAAP - adds additional customization, so no longer uses the VEDA built images https://github.com/MAAP-Project/maap-workspaces/pull/189/changes @grallewellyn is lead builder. Which makes me wonder if we should drop the common values from MAAP entirely?

Correct, they are currently using the custom image. So I think it would be good to add to Disasters Hub pangeo image.

PR #7597, opened 2026-02-05
[CloudBank] Staging Submit/Collect Shares

sean-morris:
I'm setting up a "submit/collect" workflow for the East Tennessee hub.

**The Setup:**
- Students: Can drop files into a "submit" folder but have zero visibility into other files in the underlying _shared-private mount.
- Instructors: Have a "collect" folder with full access to everyone's subfolders.

**Issues**
- All instructors see all student folders; ETSU only has one instructor so this is not a problem yet.
- Folder names are the full CILogon email addresses. (I’m not sure yet if I can strip the domain to keep the folder names cleaner).

I'm going to deploy this for ETSU to give it a try - do you see problems with how I handled all this?

### DIFF
--- a/config/clusters/cloudbank/staging.values.yaml
+++ b/config/clusters/cloudbank/staging.values.yaml
@@ -16,22 +16,22 @@ jupyterhub:
       tag: sp26.4
     storage:
       extraVolumeMounts:
-        - name: home
-          mountPath: /home/jovyan/submit
-          subPath: _shared-private/{escaped_username}
-        - name: home
-          mountPath: /home/jovyan/shared
-          subPath: _shared
-          readOnly: true
-        - name: dev-shm
-          mountPath: /dev/shm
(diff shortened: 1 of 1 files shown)


config/clusters/cloudbank/staging.values.yaml
```diff
@@ -13,6 +14,11 @@ jupyterhub:
     image:
       name: quay.io/ccsf/jupyterhub
       tag: sp26.4
+    storage:
+      extraVolumeMounts:
+      - name: home
```
yuvipanda: you need to copy the whole list from https://github.com/2i2c-org/infrastructure/blob/main/helm-charts/basehub/values.yaml#L762

sean-morris: ok good deal -- did not realize lists did not merge!

PR #7599, opened 2026-02-05
[CloudBank] gpu-demo

sean-morris:


yuvipanda (approved):
Let’s try it!

config/clusters/cloudbank/gpu-demo.values.yaml
```diff
@@ -0,0 +1,65 @@
...
+          name: CloudBank Classroom - GPU Demo
+          url: https://www.cloudbank.org/
+  hub:
+    config:
+      JupyterHub:
+        authenticator_class: github
+      GitHubOAuthenticator:
+        oauth_callback_url: https://gpu-demo.cloudbank.2i2c.cloud/hub/oauth_callback
+        allowed_organizations:
+        - ucb-ds
```
yuvipanda, line 67: can you add https://github.com/2i2c-org/infrastructure/pull/7599/changes#diff-7b1f8c8dca49b472e5d34946909a9efb8af7d06d4462b17276cd853b2ceed334R16 here too?

config/clusters/cloudbank/gpu-demo.values.yaml
```diff
@@ -0,0 +1,65 @@
...
+  ingress:
+    hosts: [gpu-demo.cloudbank.2i2c.cloud]
+    tls:
+    - hosts: [gpu-demo.cloudbank.2i2c.cloud]
+      secretName: https-auto-tls
+  singleuser:
+    defaultUrl: /lab
+    memory:
+      guarantee: 512M
+      limit: 1G
```
yuvipanda: these should be much higher, because each node will only have one user.

config/clusters/cloudbank/gpu-demo.values.yaml
```diff
@@ -0,0 +1,65 @@
...
+    profileList:
+    - display_name: NVIDIA Tesla T4, ~16 GB, ~4 CPUs
```
yuvipanda: Look at https://github.com/2i2c-org/infrastructure/blob/d838ebd5d2c36a4263b4b32d73d8613c4604e369/config/clusters/leap/daskhub-common.values.yaml#L368 for an example. I suggest picking one of the images there (but with a newer tag) as an option here, so people can actually use the GPU

config/clusters/cloudbank/gpu-demo.values.yaml
sean-morris: I not seeing what you are seeing here? What do we need to add?

sean-morris:
@yuvipanda I deployed the changed via the deployer -- just waiting on the one piece above.

https://gpu-demo.cloudbank.2i2c.cloud/

PR #7641, opened 2026-02-11
Bump prometheus version 

### LINKED ISSUE #7637: Update prometheus chart version to maintain compatibility with k8s 1.34
### Context

We recently updated half of our GCP hubs to k8s 1.34
While investigating a PD alert on one of the hubs, no pod logs seemed to be available since the upgrade happened.

After some digging, it looks like k8s 1.33 deprecated the use of `endpoints` for `endpointslices` https://kubernetes.io/blog/2025/04/24/endpoints-deprecation/.

The version of `prometheus` chart that we run, has a `kube-state-metrics` version, that doesn't have this change. 

The change is available since the following versions of the helm charts:
- `kube-state-metrics` [7.0.0](https://github.com/prometheus-community/helm-charts/pull/6376/)
- `prometheus` [27.52.0](https://github.com/prometheus-community/helm-charts/pull/6390)

### What we need to do

Update the `prometheus` version that we use in our support chart to >27.52.0

### Definition of Done

- [ ] prometheus chart is upgraded
- [ ] logs are showing up again

GeorgianaElena:
Fixes #7637

### DIFF
--- a/helm-charts/support/Chart.yaml
+++ b/helm-charts/support/Chart.yaml
@@ -13,9 +13,7 @@ dependencies:
   # cost. If we ever have multiple prometheii on the same cluster, we can
   # reconsider using the operator.
   - name: prometheus
-    # NOTE: CHECK INSTRUCTIONS UNDER prometheus.server.command IN support/values.yaml
-    # EACH TIME THIS VERSION IS BUMPED!
-    version: 27.48.0
+    version: 28.9.0
     repository: https://prometheus-community.github.io/helm-charts
 
   # Grafana for dashboarding of metrics.


helm-charts/support/Chart.yaml
```diff
@@ -13,9 +13,7 @@ dependencies:
   # cost. If we ever have multiple prometheii on the same cluster, we can
   # reconsider using the operator.
   - name: prometheus
-    # NOTE: CHECK INSTRUCTIONS UNDER prometheus.server.command IN support/values.yaml
```
yuvipanda, line 16: Do we no longer have to check the command?

GeorgianaElena: I don't think so @yuvipanda! You removed the command a while ago: https://github.com/2i2c-org/infrastructure/commit/84a3d34c644f476c6ca3ec51db2b993ba6d4f847

yuvipanda: wheeee! Thanks for cleaning up, I must've forgotten to remove that!

Issue #5927, opened 2025-04-16
Stop using lists anywhere in YAML for our config


yuvipanda:
Lists in YAML aren't easily overrideable, and that means we have to copy the entire original list and keep that in sync each time we copy / change it. This sucks.

https://github.com/jupyterhub/kubespawner/pull/845 was intended to help with this. Instead of lists, we can use dicts - which *are* overrideable, and 'remove' entries by setting the value to null. We have this change deployed, but haven't changed our config.

We should change our configs to use dicts everywhere we use lists, and makes life much easier. In particular, it also makes it easier to keep profileLists consistent across hubs, when they may change only via the groups that are allowed access.

## Definition of done

- [ ] All the lists in our config that should be dicts are identified
- [ ] All those are changed to dicts
- [ ]

PR #7673, opened 2026-02-16
Use dicts instead of lists wherever possible

### LINKED ISSUE #7561: [Round 2] Rollout latest z2jh version to all remaining hubs
### Context

Follow-up to #6915 

### What we need to do

Upgrade all cluster to latest z2jh chart version.

### Definition of Done

- [x] Update the z2jh chart version
- [ ] Update config to use dicts instead of lists for `singleuser.storage.volumes` and `volume_mounts`
- [x] Rm special cased hub image
- [x] Upgrade the main hub image
- [x] Deploy everywhere
- [ ] Document rolling out such an upgrade in two batches

GeorgianaElena:
Fixes #7561
Fixes https://github.com/2i2c-org/infrastructure/issues/5927

### DIFF
--- a/config/clusters/disasters/common.values.yaml
+++ b/config/clusters/disasters/common.values.yaml
@@ -76,12 +76,12 @@ jupyterhub:
     defaultUrl: /lab
     storage:
       extraVolumeMounts:
-        1-shared-public-volumemount:
+        01-shared-public-volumemount:
           name: home
           mountPath: /home/jovyan/shared-public
           subPath: _shared-public
           readOnly: false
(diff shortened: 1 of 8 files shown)


helm-charts/basehub/values.yaml
```diff
@@ -1289,6 +1289,9 @@ jupyterhub:
                   raise Exception("No container named 'notebook' found in pod definition")
 
               admin_volume_mounts = custom_admin.get('extraVolumeMounts', [])
```
agoose77: @GeorgianaElena do you need to update the default here too, if the later code expects a dict?

PR #7687, opened 2026-02-17
Revert "Revert "Use dicts instead of lists wherever possible""

GeorgianaElena:
Ref https://github.com/2i2c-org/infrastructure/issues/7561
Reverts 2i2c-org/infrastructure#7686


agoose77:
> Upgrade the projectpythia cluster to latest k8s version and check if node labeling works now

We should check this out, but I have a suspicion that it's not fixed -- the code that's being used to justify the statement is 3 years old, whilst the issue was closed in 2025.

PR #7691, opened 2026-02-17
Handle resource exhaustion for hhmi in us-west-2

yuvipanda:
- Spread the load out to all zones, as us-west2 seems much smaller in GCP
- Ask for a smaller VM, so we have less chances of running out of resources. Also helps with cloud cost
- Set max VM count to be smaller

Ref https://github.com/2i2c-org/infrastructure/issues/7657


yuvipanda:
Not actually applied!

PR #7701, opened 2026-02-18
chore: add GetObjectTagging permissions

jjfrench:
## Description
- Adds s3:GetObjectTagging permission to MAAP hubs in regards to [MAAP Community Issue #1251](https://github.com/MAAP-Project/Community/issues/1251#event-22893341379)

### DIFF
--- a/terraform/aws/projects/maap.tfvars
+++ b/terraform/aws/projects/maap.tfvars
@@ -29,18 +29,11 @@ hub_cloud_permissions = {
           {
             "Effect": "Allow",
             "Action": [
-              "s3:PutObject",
-              "s3:PutObjectAcl",
               "s3:GetObject",
               "s3:GetObjectTagging",
-              "s3:ListBucketMultipartUploads",
-              "s3:AbortMultipartUpload",
               "s3:ListBucketVersions",
-              "s3:CreateBucket",
               "s3:ListBucket",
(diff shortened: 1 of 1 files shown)


terraform/aws/projects/maap.tfvars
```diff
@@ -29,17 +29,11 @@ hub_cloud_permissions = {
           {
             "Effect": "Allow",
             "Action": [
-              "s3:PutObject",
```
yuvipanda, line 32: Are all these removals intentional?

jjfrench: Yeah, these get moved down to `maap-ops-workspace` specifically. I don't think they were intended for the majority of these buckets. This essentially mirrors the prod hub's permissions

wildintellect: This is actually an important question, do we want to allow more permissions and rely on the bucket owner policy to determine if we're actually allowed to write?

I think this PR as is will work for the current needs but I wonder if we want to leave open the options to write to some of the other buckets.

jjfrench: I believe it is a shared responsibility - both being least permissive. Not having visibility of other target bucket policies, I wouldn't want someone to accidentally delete objects unknowingly if the bucket owner didn't tighten their permissions initially

PR #7769, opened 2026-02-27
Add a PD incident workflow that runs a health-check for the alerting hub

GeorgianaElena:
Ref https://github.com/2i2c-org/infrastructure/issues/7807


agoose77:
Very excited to see this being merged @GeorgianaElena !!!

PR #7851, opened 2026-03-05
[projectpythia-binder] Redeploy JS2 cluster

### LINKED ISSUE #7843: Upgrade JS2 K8s version for projectpythia-binder
### Context

projectpythia-binder is running 1.31 currently.

### What we need to do

We need to upgrade. However, we never did this and don't have docs. Upgrading might actually mean re-creating the cluster. 

Relevant issues:
- https://github.com/2i2c-org/meta/issues/2917
- https://github.com/2i2c-org/infrastructure/issues/6464

### Definition of Done

- [ ] projectpythia-binder clusters run 1.34
- [ ] We understand if it's possible and update the infra appropriately so that upgrading doesn't mean cluster recreation
- [ ] Docs are created

agoose77:
Closes #7843 by redeploying with K8s 1.33, the latest available version.

### DIFF
--- a/docs/howto/migrate-ingress.md
+++ b/docs/howto/migrate-ingress.md
@@ -37,17 +37,6 @@ We must migrate the DNS records for the domain to point to this external IP. Dur
 
 Once the DNS records have been updated to point to the `cluster-entrypoint` LB, we can safely transition to the official `ingress-nginx` service:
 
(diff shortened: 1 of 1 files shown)


agoose77:
Testing that scale up works...

PR #7881, opened 2026-03-09
fix: set header snippets for websocket upgrade on nginx-ingress

agoose77:
@ktyle reported that kernels and terminals weren't working properly on the Pythia BinderHub. He also saw 502 errors. Both of these, I reproduced.

We recently upgraded that hub to use nginx-ingress, and it turns out that our current configuration does not support websocket upgrades.

This PR sets the location snippet required to support that. There may be other configuration we need to tweak here (like buffering, timeouts, etc).

It also sets a TTL on DNS replies to match the old ingress defaults.

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -431,6 +431,10 @@ jupyterhub:
     ingressClassName: nginx
     annotations:
       nginx.ingress.kubernetes.io/proxy-body-size: 256m
+      # Handle websocket upgrades in nginx-ingress
+      nginx.org/location-snippets: |
+        proxy_set_header Upgrade $http_upgrade;
+        proxy_set_header Connection $connection_upgrade;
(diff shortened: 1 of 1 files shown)


helm-charts/basehub/values.yaml
```diff
@@ -13,6 +13,8 @@ binderhub-service:
     annotations:
       nginx.ingress.kubernetes.io/proxy-body-size: 256m
       cert-manager.io/cluster-issuer: letsencrypt-prod
+      nginx.org/server-snippets: |
+        resolver_timeout 30s;
```
agoose77, line 17: It looks like NGINX caches the DNS responses by default, whereas the existing ingress-nginx uses the /etc/resolve to set a resolver with 30s TTL.

agoose77:
@yuvipanda I pinged you for review. This feels a bit like whack-a-mole — I wonder whether we should do a more systematic comparison of the resolved nginx configurations (I've been using them to guess what's missing).

PR #7883, opened 2026-03-09
fix: revert resolver timeouts

agoose77:
This wasn't needed for the 502 on Pythia, even though the ingress-nginx controller does set resolver timeouts, let's not do this now.


yuvipanda (approved):

helm-charts/basehub/values.yaml
```diff
@@ -437,8 +435,6 @@ jupyterhub:
       nginx.org/location-snippets: |
         proxy_set_header Upgrade $http_upgrade;
```
yuvipanda, line 437: i would think these aren't required either

agoose77: This is required, without it websocket upgrade connections fail: https://2i2c-org.pagerduty.com/postmortems/189b4e6b-7ca9-e7e0-fbff-a07a88814222

Issue #7470, opened 2026-01-21
Roll out helm v4 in our infrastructure


yuvipanda:
Once we know that helm v4 works in zero to jupyterhub, we roll it out in our CI/CD. This would also require that everyone using the deployer locally switch to using helm v4, and the deployer code will refuse to run helm v3.

## Definition of done

- [ ] Modify our deployer code to work with helm v4
- [ ] Modify our deployer code to refuse to work with helm v3
- [ ] Upgrade our CI/CD code to run with helm v4
- [ ] Roll this out to our clusters in a batched fashion

GeorgianaElena:
@Gman0909, I added this to the `UpNext` column as I want to pick up #7886 next sprint and drive it to completion. This means we should also mark https://github.com/2i2c-org/initiatives/issues/4 as in flight once we pick that up. Are you ok with that?

GeorgianaElena:
@yuvipanda, @agoose77, not sure what to do about this!

@yuvipanda, for context the bug was closed upstream, but remains unsolved -> https://github.com/2i2c-org/infrastructure/pull/9054#issuecomment-5715347593

Looking at https://github.com/search?q=repo%3Ahelm%2Fhelm+null&type=issues&s=created&o=desc, there seems to be lots of confusion and inconsistencies around `null` merging behaviour. I am reluctant to deploy this to our hubs until they sort things out consistently.

What do you all think we should do with this one?

Also, @agoose77, I hope it's ok to re-assign this to you. I feel you have the most context and did most of the heavy lifting, finding the bug and filling it upstream together with the reproducer.

yuvipanda:
Is it possible for us to bypass the bug, by either putting the config in jsonnet or something similar? IIRC the primary issue is with Jetstream, and we could potentially special case that in our jsonnet to move this forward, right?

GeorgianaElena:
There are [so many bug reports](https://github.com/search?q=repo%3Ahelm%2Fhelm+null&type=issues&s=created&o=desc) related to unexpected null merging behaviour, that I actually wouldn't feel comfortable deploying this anywhere atm :/ 

Maybe we can do one canary deployment and observe that for a while?

agoose77:
I agree with Georgiana. Whilst we can work-around this specific bug by performing sibling merging ourselves via Jsonnet, a fundamental bug like this in Helm feels like a footgun that we can't easily guard against raising its head in the future. 

[...]

PR #7886, opened 2026-03-10
Bump to helm v4

yuvipanda:
Tested it locally, and seems fine. Also bumps kubectl version.

Remove a lot of superfluous and drifted-to-no-longer-be-accurate comments.

Once this is merged, we would need to make sure everyone upgrades their local install of helm to v4 as well

Ref https://github.com/2i2c-org/infrastructure/issues/7470

### DIFF
--- a/.github/actions/setup-deploy/action.yaml
+++ b/.github/actions/setup-deploy/action.yaml
@@ -69,54 +69,19 @@ runs:
       go install github.com/google/go-jsonnet/cmd/jsonnet@v0.20.0
     shell: bash
 
-    # This action use the github official cache mechanism internally
   - uses: azure/setup-helm@v4
     with:
-        # Manually update a pinning of helm to a minor version based on:
-        #
-        # - it seems to work
-        # - to avoid falling behind
-        #
-        # Related:
-        #
-        # - helm versions: https://github.com/helm/helm/releases
-        #
-      version: v3.14.4
+      version: v4.1.1
 
-    # Manually update a pinning of kubectl to a minor version based on:
-    #
-    # - the current range of k8s version in our k8s clusters, as of 2024-04-17,
-    #   this is k8s 1.27 - 1.29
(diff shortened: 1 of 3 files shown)


yuvipanda:
@agoose77 would upgrading to helm cause this regression to come back? Or is the concern that there will be *other* bugs that we run into?

yuvipanda:
@agoose77 I looked into that, and i agree. I also think we can't fix this with `extraConfig` because we also need to set this for the core pods themselves, which can't be done with extraConfig. Is that right?

agoose77:
https://github.com/helm/helm/pull/31979 merged! Let me check whether it is fit for purpose.

GeorgianaElena:
@yuvipanda, @agoose77, any updates on this?

agoose77:
I performed a dry run of `deployer deploy` for all clusters and `diff`ed the results between our existing Helm binary and Helm 4 with
```bash
#!/usr/bin/env bash
set -eu

cluster=${1:?provide cluster name}
hub=${2:?provide hub name}

nix shell nixpkgs/9b100cf#kubernetes-helm --command deployer deploy --dry-run --skip-refresh ${cluster} ${hub} > /tmp/old.yml
nix shell nixpkgs/350fa6e#kubernetes-helm --command deployer deploy --dry-run --skip-refresh ${cluster} ${hub} > /tmp/new.yml

# Final rg fails if no unique entries exist
# Invert result so we pass if no diff
# First, strip lines containing diffs in secrets / checksums.
# Look for remaining diff lines
! diff /tmp/manifest* | rg -v "description|checksum|secret|token|keys" -i | rg "^<\s*\S+:|^>\s*\S+:" -q
```

I accidentally cleared my history, but I watched the test run, and I saw that most were identical.

agoose77:
Closing this in favour of https://github.com/2i2c-org/infrastructure/pull/9054 (because I rebased it!)

PR #7896, opened 2026-03-11
Refactor populate_token function for sidecar auth_state handling

bsatoriu:
Inject the MAAP_PGT env var into the `s3fs` sidecar via the populate_token hook.

### DIFF
--- a/.github/workflows/pd-triggered-health-check.yaml
+++ b/.github/workflows/pd-triggered-health-check.yaml
@@ -46,6 +46,9 @@ jobs:
     needs: [get_incident_details, health_check]
     runs-on: ubuntu-latest
     steps:
+    - uses: actions/checkout@v6
+      with:
+        submodules: true
     - name: Report Status
       if: always()
       uses: ravsamhq/notify-slack-action@v2
--- a/.github/workflows/run-health-check.yaml
+++ b/.github/workflows/run-health-check.yaml
@@ -89,12 +89,29 @@ jobs:
       fail-fast: false
       matrix:
         jobs: ${{ fromJson(needs.generate-jobs.outputs.staging-jobs) }}
-    uses: ./.github/workflows/reusable-health-check.yaml
-    with:
-      cluster: ${{ matrix.jobs.cluster_name }}
-      hub: ${{ matrix.jobs.hub_name }}
-      provider: ${{ matrix.jobs.provider }}
-    secrets: inherit
+    steps:
+    - name: Install deployer script's Python dependencies
+      run: |
+        pip install --editable .
+        go install github.com/google/go-jsonnet/cmd/jsonnet@v0.20.0
+
+    - name: Setup deploy for ${{ matrix.jobs.cluster_name }}cluster
+      uses: ./.github/actions/setup-deploy
+      with:
+        provider: ${{ matrix.jobs.provider }}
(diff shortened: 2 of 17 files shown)


config/clusters/maap/common.values.yaml
```diff
@@ -77,16 +77,28 @@ jupyterhub:
     extraConfig:
       001-username-claim: |
         def populate_token(spawner, auth_state):
-          # For our deployment-service-check health check user, there is no auth_state.
```
yuvipanda, line 80: Can you leave the comment in place?

config/clusters/maap/common.values.yaml
```diff
@@ -77,16 +77,28 @@ jupyterhub:
...
               "KC_ACCESS_TOKEN": auth_state.get("access_token", ""),
               "KC_ID_TOKEN": auth_state.get("id_token", ""),
               "KC_REFRESH_TOKEN": auth_state.get("refresh_token", "")
             })
+            
+            # Inject MAAP_PGT into the s3fs sidecar container
+            for container in spawner.extra_containers:
+              if container.get("name") == "s3fs":
+                env_list = container.setdefault("env", [])
+                # Remove any existing MAAP_PGT entry to avoid duplicates
```
yuvipanda: Is this an existing problem? If we don't set MAAP_PGT entries, none would be set. If there are duplicates, I'd rather have us look at the source and eliminate that, rather than add this code here to handle it this way

config/clusters/maap/common.values.yaml
```diff
@@ -139,6 +151,12 @@ jupyterhub:
           subPath: triaged-jobs
           mountPropagation: HostToContainer
           readOnly: true
+        07-s3f3-org-buckets-volumemount:
```
yuvipanda: Is this meant to be `s3fs` not `s3f3`?

bsatoriu: @yuvipanda these labels were changed by @GeorgianaElena, not me. I was following the convention from the last commit. See: https://github.com/2i2c-org/infrastructure/commit/07b17187a88a61d05543290d8522f6239e8e7370

config/clusters/maap/common.values.yaml
bsatoriu: done

config/clusters/maap/common.values.yaml
bsatoriu: Duplicate check removed.

config/clusters/maap/common.values.yaml
```diff
@@ -78,15 +78,24 @@ jupyterhub:
...
               "KC_ACCESS_TOKEN": auth_state.get("access_token", ""),
               "KC_ID_TOKEN": auth_state.get("id_token", ""),
               "KC_REFRESH_TOKEN": auth_state.get("refresh_token", "")
             })
+            
+            # Inject MAAP_PGT into the s3fs sidecar container
+            for container in spawner.extra_containers:
+              if container.get("name") == "s3fs":
+                env_list = container.setdefault("env", [])
+                container["env"].append({
```
agoose77:
```suggestion
                env_list.append({
```

config/clusters/maap/common.values.yaml
agoose77: That's a typo, but not a harmful one. Let's change it here for readability. One moment.

PR #8047, opened 2026-04-03
Drastically simplify core-dumps daemonset

yuvipanda:
- Using the critical addon causes problems on GCP, so let's not use those
- Setting any resources at all seems to make it hard to schedule on some full nodes, so let's not. We only run one syscstl command, and it is fine if it fails.
- Use the 'pause' container, as we don't really need to do anything after the daemonset starts. It can just... be.
- Roll this out everywhere, because we forgot to

Follow-up to https://github.com/2i2c-org/infrastructure/pull/7588


yuvipanda:
@sunu fyi!

PR #8068, opened 2026-04-09
Bump fancy profiles

agoose77:
This PR bumps the fancy profile versions across all non-customised hubs.

The motivation for this is that Earthscope need it for prod, and I am unclear whether others do too — if so, we can merge this. Meanwhile, I'll customise EarthScope.


sunu:
@batpad said on Slack that
> We intentionally did not cater to backwards compatibility :disappointed: - so I believe this would likely fail if binder token stuff is not authenticated

~~@agoose77 If I understand correctly, this means we'll need to add the appropriate scopes and enable binderhub authentication on any existing hubs using fancy-profiles while upgrading, otherwise fancy-profiles will break on boot, right? So may be this needs to be done hub by hub rather than all at once?~~

We discussed this on Slack. The latest fancy profiles is backwards compatible with our existing configuration and works fine even when auth is disabled on binderhub

agoose77:
> We discussed this on Slack. The latest fancy profiles is backwards compatible with our existing configuration and works fine even when auth is disabled on binderhub

Yes, to add context — EarthScope staging is not using auth on the BinderHub (I explicitly tested this via CURL). The new parts are (a) the fancy-profiles requesting an API token and (b) passing that as an auth header to the binderhub service. The former should always succeed, whilst the latter is a no-op in non-auth deployments.

Issue #8063, opened 2026-04-08
Fix certificate renewals of nginx-ingress clusters


agoose77:
Clusters running nginx-ingress are failing to renew their certificates. This is because the cert-manager is trying to create temporary ingresses for ACME challenge handling, which breaks under nginx-ingress (it [uses oldest-wins for host collisions](https://docs.nginx.com/nginx-ingress-controller/configuration/host-and-listener-collisions/#winner-selection-algorithm)).

One solution to this is [mergeable ingresses](https://github.com/nginx/kubernetes-ingress/tree/v5.4.1/examples/ingress-resources/mergeable-ingress-types), via master-minion separation. But this is a very invasive change, because the master ingress must contain no `paths`, and the TLS information, whilst the minion must not terminate TLS and may contain paths.

We attempted to avoid this problem by making the ingresses editable by cert-manager. However, there are two problems:

1. Prometheus uses basic-auth, which seems impossible to disable on a per-route level due to how the config is generated by nginx-ingress. This means the ACME challenge resolution encounters HTTP 401 responses.
2. Patching existing ingresses, which was performed by the nginx-ingress migration, does not trigger update of the certificate, which means that nginx-ingress still tries to create a temporary ingress.

I had some luck solving (2) by:
- Manually reconciling the `Certificate` resource with the `Ingress` (adding two annotations, `acme.cert-manager.io/http01-override-ingress-name: <INGRESS NAME>` and `cert-manager.io/issue-temporary-certificate: "true"`)
- Deleting any existing `CertificateRequest` resources associated with the cert.

This triggers a recreation of the request, which ultimately patches the ingress.

But (1) is fairly hard to circumvent because the `basic_auth` config is generated by nginx-ingress, and NGINX doesn't like duplicate directives, etc.

We could handle prometheus on its own by adding an intermediary service, or trying to make prometheus itself enforce basic_auth, but I think it might be time to look again at mergeable ingresses.

yuvipanda:
I also wrote this piece of code to validate that we don't have any certificates becoming invalid in the next 30d:

```
from datetime import datetime
from deployer.infra_components.cluster import Cluster
from dateutil.parser import parse
import OpenSSL
import ssl

for cluster in Cluster.get_all():
    for hub in cluster.hubs:
        domain = hub.spec["domain"]
        if domain == 'staging.js.2i2c.cloud':
            continue

        cert=ssl.get_server_certificate((domain, 443))
[...]

PR #8114, opened 2026-04-16
Move authentication from nginx to prometheus level 

### LINKED ISSUE #8119: Fix certificate renewals of nginx-ingress clusters
### Context

Implement the solution decided in https://github.com/2i2c-org/infrastructure/issues/8063

### What we need to do

https://github.com/2i2c-org/infrastructure/issues/8063

### Definition of Done

- [ ] certificate renewals work on every cluster

GeorgianaElena:
Generated this with https://gist.github.com/GeorgianaElena/9f001026f990f346e55fe8435f3509be
~It excludes earthscope because it has to be treated separately due to having two pairs of credentials.~

Ref #8063 
Fixes https://github.com/2i2c-org/infrastructure/issues/8119


GeorgianaElena:
Merging this, will break individual hub's `grafanas, but they'll be accessible from the central grafana. Looking into fixing it

GeorgianaElena:
It's been fixed and I've verified that both central grafana, as well as individual grafanas are picking up prometheus data, withouth any 401s.

Merging....

PR #8138, opened 2026-04-22
Bump repo2docker version for MAAP

yuvipanda:
They're actively experimenting with dynamic image building & R, and a lot of newer fixes will help


wildintellect (approved):
From a user perspective +1, I can't comment on if the syntax etc is correct.

PR #8146, opened 2026-04-23
Change trade-offs for MAAP server startup

yuvipanda:
- Put default user resource allocation on larger nodes (r5.4xlarge) than smaller nodes (r5.xlarge) so more users fit on a single node. This leads to faster startup times, as fewer node spin-ups need to happen. This is at the cost of higher cloud spend, which will need to be monitored
- Enable continuous image pre-pulling, so all images in our profile list are pulled on all nodes as they come in. This means new images don't need to be pulled, at the cost of it taking up more disk space (that would otherwise be used for tmp or logs etc) - so we may need to increase disk space too.

Ref https://github.com/2i2c-org/initiatives/issues/50
Ref https://github.com/2i2c-org/infrastructure/issues/8148

### DIFF
--- a/config/clusters/maap/common.values.yaml
+++ b/config/clusters/maap/common.values.yaml
@@ -14,6 +14,13 @@ nfs:
 dask-gateway:
   enabled: true
 jupyterhub:
+
+  prePuller:
+    # Pre-pull all explicitly specified images in our profiles on all nodes
+    # when they start up
+    continuous:
+      enabled: true
+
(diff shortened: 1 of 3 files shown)


wildintellect:
I think it makes sense to try these changes on MAAP. They are easy to undo if we change our minds.

wildintellect (commented):
I do have one question, are GPU images only pulled to GPU nodes?

PR #8153, opened 2026-04-24
Reduce baseline cost of hhmi cluster

yuvipanda:
- Switch from n2-highmem-4 nodes to n2-highmem-2 nodes
- Reduce the resources allocated to nfs server, as this sees very light use
- Reduce the prometheus server allocation too

Ref https://github.com/2i2c-org/meta/issues/3097


yuvipanda:
Already deployed!

PR #8155, opened 2026-04-24
Add alert for missing kubelet volume stats

agoose77:
Several clusters, due to k8s bugs, are missing these metrics. Whilst it's not super likely that we'll run into this again, it seems prudent to check for "missing" for things that we alert on. For now, only the volume metrics.

### DIFF
--- a/helm-charts/support/values.jsonnet
+++ b/helm-charts/support/values.jsonnet
@@ -2,9 +2,29 @@ local cluster_name = std.extVar('VARS_2I2C_CLUSTER_NAME');
 local provider_name = std.extVar('VARS_2I2C_PROVIDER');
 
 function(VARS_2I2C_AWS_ACCOUNT_ID=null)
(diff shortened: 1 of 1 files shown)


helm-charts/support/values.jsonnet
```diff
@@ -2,6 +2,30 @@ local cluster_name = std.extVar('VARS_2I2C_CLUSTER_NAME');
 local provider_name = std.extVar('VARS_2I2C_PROVIDER');
 
 function(VARS_2I2C_AWS_ACCOUNT_ID=null)
+  local sameDayAction = 'same day action needed',
+  local sameWeekAction = 'action needed this week',
+  local immediateAction = 'take immediate action',
```
agoose77: Pull these out. I would like to use an enum-like object but they don't exist. Could create an object, but locals is fine for now.

PR #8159, opened 2026-04-24
Revert "Add alert for missing kubelet volume stats"

yuvipanda:
Reverts 2i2c-org/infrastructure#8155

This failed jsonnet validation, due to missing `labels` key (see https://github.com/2i2c-org/infrastructure/actions/runs/24895352862)


agoose77:
Ah damn, good catch. I knew it was a risk pushing this on Friday.

Issue #8192, opened 2026-04-28
Hard to reproduce 403 errors across our cluster


yuvipanda:
We started encountering this in MAAP (https://2i2c.freshdesk.com/a/tickets/5514), but have since encountered it in LEAP (https://2i2c.freshdesk.com/a/tickets/5535) and also in AWI-CIROH workshop hub.

Let's investigate and fix it.

## Definition of done

- [ ] No more unexplained 403 errors reported

yuvipanda:
We believe that https://github.com/2i2c-org/infrastructure/pull/8273 should fix this for now, and https://github.com/jupyterhub/jupyterhub/pull/5386 is the upstream fix.

PR #8193, opened 2026-04-28
Preserve client IPs for JupyterHub

yuvipanda:
JupyterHub uses client IPs as part of its XSRF calculation, and while we aren't sure, preserving client IP is a good idea here regardless.

Ref https://github.com/2i2c-org/infrastructure/issues/8192

### DIFF
--- a/helm-charts/support/templates/traffic-entrypoint.yaml
+++ b/helm-charts/support/templates/traffic-entrypoint.yaml
@@ -7,6 +7,9 @@ metadata:
     app: cluster-entrypoint
 spec:
   type: LoadBalancer
+  # Avoid extra hop and preserve client IP, as JupyterHub
+  # uses that as part of its XSRF check.
+  externalTrafficPolicy: Local
   ports:
     - name: http
       port: 80


minrk (approved):

minrk:
Yeah, this is also required for meaningful ips in logs, etc.

If the request ip is not reliable, the relevant setting is 

```
JUPYTERHUB_XSRF_ANONYMOUS_IP_CIDRS="0.0.0.0/0"
```

to avoid xsrf inconsistency when load-balancers are involved.

I _thought_ only one case was affected by this: initial login, when the client ip is taken into account for the xsrf token to avoid portable credentials across clients. However, it can also [happen](https://github.com/jupyterhub/jupyterhub/blob/6e2521f27351fe67fc3504d40f448e991f44d034/jupyterhub/handlers/base.py#L286) if jupyterhub-session-id cookie has been cleared while the user is still logged in somehow. Failure to ensure jupyterhub-session-id is set should probably be considered a jupyterhub bug.

PR #8195, opened 2026-04-28
Add alert for HTTPS Certificate Renewal failures

### LINKED ISSUE #3278: Alert when certificates are about to expire less than 2 weeks from now
We had our first ever let's encrypt related outage today (https://github.com/2i2c-org/infrastructure/issues/3274). 

We should have alerts for when HTTPS certificates are about to expire. Probably about 2 weeks notice is more than good enough.

yuvipanda:
If a certificate is less than 24 days from expiry, we get an alert. cert-manager should've renewed it before that, so this is a failure. We use a single check for *all* of our uptime checks, so we don't use too many uptime check resources. This also covers our prometheus instances, and in the future, grafana when we set them up

Fixes https://github.com/2i2c-org/infrastructure/issues/3278

### DIFF
--- a/terraform/uptime-checks/checks.tf
+++ b/terraform/uptime-checks/checks.tf
@@ -92,6 +92,42 @@ resource "google_monitoring_alert_policy" "hub_simple_uptime_alert" {
   notification_channels = [google_monitoring_notification_channel.pagerduty_hubs.name]
 }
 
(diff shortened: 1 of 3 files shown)


GeorgianaElena:
Thank you, Yuvi!

Issue #8130, opened 2026-04-22
[Incident Action]: Create cloudwatch alerts for throughput and iops exceeding


agoose77:
### Incident Report

https://2i2c-org.pagerduty.com/postmortems/18a8ac0f-27c6-3176-bacb-1b66d8fe76ef

### What needs to happen?

Degradation of home storage performance is usually a fairly critical failure. We want to detect this as early as possible, in order to ensure that we don't experience a significant outage.

We currently check `%iowait`. Can we use another metric, e.g. `aws_ebs_csi_ec2_exceeded_tp_seconds_total` and `aws_ebs_csi_ec2_exceeded_iops_seconds_total`? These seem flat in EarthScope Prometheus. But, see the dashboard:

<img width="750" height="205" alt="Image" src="https://github.com/user-attachments/assets/efbbfa3f-1b86-4ddd-a95b-9a84b77b8e64" />

Let's look at tweaking this to trigger earlier.

### What is definition of done for this item?

- [ ] Alert configuration for %iowait has been reviewed and modified if possible to reduce latency.
- [ ] This alert is then tested.

yuvipanda:
I would like us to use *cloudwatch* alarms here: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_metric_alarm.

That uses the absolute source of truth that causes throttling, without additional layers of collection.

agoose77:
Closed by #8215

yuvipanda:
let's verify this works, and write some docs for responders

PR #8215, opened 2026-04-30
Add cloudwatch alarms for EBS thresholds

agoose77:
Partly addresses #8130 

We need to get these into PagerDuty somehow.

See also: https://support.pagerduty.com/main/docs/amazon-cloudwatch-integration-guide

### DIFF
--- a/terraform/aws/ebs-volumes.tf
+++ b/terraform/aws/ebs-volumes.tf
@@ -18,6 +18,21 @@ resource "aws_ebs_volume" "nfs_home_dirs" {
   }
 }
 
+data "sops_file" "pagerduty_volume_integration" {
+  # Read sops encrypted file containing integration key for pagerduty
+  source_file = "secret/enc-pagerduty-volume-integration.yaml"
+}
+
+resource "aws_sns_topic" "volume_metric_exceeded" {
+  name = "volume-metric-exceeded-topic"
+}
+
+resource "aws_sns_topic_subscription" "volume_metric_exceeded_https_target" {
+  topic_arn = aws_sns_topic.volume_metric_exceeded.arn
+  protocol  = "https"
+  endpoint  = data.sops_file.pagerduty_volume_integration.data["pagerdutyIntegration.url"]
+}
+
 resource "aws_cloudwatch_metric_alarm" "volume_throughput_alarm" {
   alarm_name                = "warn-volume-throughput-exceeded"
   comparison_operator       = "GreaterThanThreshold"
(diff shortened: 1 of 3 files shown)


terraform/aws/ebs-volumes.tf
```diff
@@ -18,6 +18,31 @@ resource "aws_ebs_volume" "nfs_home_dirs" {
...
 
+resource "aws_cloudwatch_metric_alarm" "volume_throughput_alarm" {
+  alarm_name                = "warn-volume-throughput-exceeded"
+  comparison_operator       = "GreaterThanThreshold"
+  evaluation_periods        = 1
+  metric_name               = "VolumeThroughputExceededCheck"
+  namespace                 = "AWS/EBS"
+  period                    = 300
+  statistic                 = "Average"
+  threshold                 = 0.7
```
yuvipanda, line 45: This means it'll trigger if it's >70% over 5 minutes?

agoose77:
@yuvipanda I realised this might not work — these metrics (e.g. `VolumeThroughputExceededCheck`) are defined only in the context of both a volume ID _and_ and instance ID, which we don't have at Terraform time.

yuvipanda:
@agoose77 i did see the alerts come up. And I think you can only specify one of the dimensions (`VolumeID`) and not specify InstanceID - that's how at least the grafana charts I made work. What error are you running into in specifying just the VolumeID dimension?

agoose77:
> @agoose77 i did see the alerts come up. And I think you can only specify one of the dimensions (`VolumeID`) and not specify InstanceID - that's how at least the grafana charts I made work. What error are you running into in specifying just the VolumeID dimension?

The metric is listed under a category that requires an instance. I assume that's because they only work on nitro instances.

I triggered the cloudwatch alarm manually to test things, which I think is what we saw in PD.

yuvipanda:
@agoose77 I think it's because the metric may be reported only when the instance is actually attached. Where was this categorization you saw? I would assume if it did require both dimensions, it wouldn't have worked on my grafana chart.

agoose77:
Well, I was looking at the alert UI -- it's not possible to create an alert on the metric without providing both dimensions (and in the EBS metric viewer, these metrics are listed under "with Instance" category). But the AWS docs show an `|`: https://docs.aws.amazon.com/ebs/latest/userguide/using_cloudwatch_ebs.html#ebs-volume-metrics

I'll re-apply this and see what happens. If we get an alert, it's working :sweat_smile:

terraform/aws/ebs-volumes.tf
agoose77: Yes. I'm not sure exactly how this metric is computed -- some docs say {0, 1} whilst the UI shows interpolation ([0, 1]). I'm not sure if that's a UI thing or a representation of the actual value having float properties.

agoose77:
I'm applying this, as it's currently in-effect. Let's keen an eye on PD!

Issue #8206, opened 2026-04-29
Angus tackling Pythia PR previews?


agoose77:
Project Pythia's PR previews are painful. They don't use RTD / Netlify due to the registration requirements. The existing workflow is complex, and recently @jukent was banned due to getting caught up in the nightly builds.

I'd like to spend some time on this with a view to enabling Pythia, and wider to support the ecosystem (and ourselves)

agoose77:
Some progress:

I've been tinkering with Yuvi's https://github.com/yuvipanda/jupyterbook.pub/ in my spare time recently. 

I took my local Helm chart and did a bit of work to clean it up: https://github.com/agoose77/jupyterbook-pub-service. It's definitely not pretty, but I'm not optimising for aesthetics at the moment.

I've merged in JupyterHub support and made authentication opt-in. These two factors mean it's possible to run JupyterBook.pub on a Hub.

I've added the ability to pass in builder configuration so that we can disable source builds. This drastically improves the out-of-the-box security, because this removes the option to run arbitrary code. There are still risks from unsandboxed vulnerabilities, but they'll be confined to the service container only. 

I have a WIP here: https://github.com/2i2c-org/infrastructure/pull/8224#issuecomment-4360277792

CSH today: 4hrs

yuvipanda:
My primary concern is about:

1. your time
2. you burning out.
 
@agoose77 with those caveats, please make technical choices as you see fit.

Can you record the time you've spent on this for pythia as well, so we have useful information for invoicing?

aprilmj:
I think this is beneficial to do as long as it's doable within our sprint, and I like @yuvipanda's point to track this as part of Pythia (thus the CSH label and Asana link).

agoose77:
I spent another hour on this today to mitigate copyfail and tie in with Pythia CI/CD. The latter work is tracked under the binderbot initiative, but the former isn't. 

CSH: 1hr.

agoose77:
I spent another ~8 hours on this to wrap it up.

agoose77:
I finished this spike. See https://github.com/agoose77/ptype-ml-cookbook/actions/runs/25468654449 for an example link, and the main deployment at https://hub.projectpythia.org/services/jupyterbook-pub/

There are some outstanding items, but I'll work those through with Yuvi.

agoose77:
An initiative is here: https://github.com/2i2c-org/initiatives/issues/59

PR #8224, opened 2026-05-01
[projectpythia-binder] Add temporary jupyterbook-pub-service

agoose77:
This service uses ephemeral storage for builds and repos. It doesn't use auth, because they likely don't want it authenticated for CI previews. It uses AST previewing only, to reduce the attack surface (for build poisoning).

Part of a spike for #8206

### DIFF
--- a/config/clusters/projectpythia-binder/binderhub.values.yaml
+++ b/config/clusters/projectpythia-binder/binderhub.values.yaml
@@ -83,6 +83,9 @@ jupyterhub:
       nodeSelector:
         capi.stackhpc.com/node-group: user-m3-large
   hub:
+    services:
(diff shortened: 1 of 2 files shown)


agoose77:
Here's an example workflow: https://github.com/agoose77/ptype-ml-cookbook/actions/runs/25222027829

It spits out this build summary:
> Preview this build at https://hub.projectpythia.org/services/jupyterbook-pub/repo/https%3A%2F%2Fgithub.com%2Fagoose77%2Fptype-ml-cookbook%2Factions%2Fruns%2F25222027829%2Fartifacts%2F6752430996/

Whose link opens the preview on the BinderHub

PR #8226, opened 2026-05-03
Populate hubspot_deal_id for some clusters

yuvipanda:
Thinking about experimenting with using the hubspot API to get information about deals (https://developers.hubspot.com/docs/api-reference/latest/crm/objects/deals/get-deal). A deal is how we seem to be encoding 'current contracts', with appropriate end dates.

Nothing concrete, just an exploration.


haroldcampbell:
Sweeet!

yuvipanda:
docs failure is unrelated

PR #8227, opened 2026-05-04
Add copyfail daemonset

GeorgianaElena:
See https://2i2c.slack.com/archives/C07SJJWVCAD/p1777738668937179

### DIFF
--- a/helm-charts/support/values.yaml
+++ b/helm-charts/support/values.yaml
@@ -14,6 +14,12 @@
 cluster-autoscaler:
   enabled: false
 
+copyfail:
+  enabled: true
+  image:
+    repository: ghcr.io/iwanhae/copyfail-ebpf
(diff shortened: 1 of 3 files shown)


helm-charts/support/values.yaml
```diff
@@ -14,6 +14,12 @@
 cluster-autoscaler:
   enabled: false
 
+copyfail:
+  enabled: true
+  image:
+    repository: ghcr.io/iwanhae/copyfail-ebpf
+    tag: latest
```
agoose77, line 21: I reckon we should pin this to a particular version, for security purposes?
Here's `latest` as of now:
```
sha256:7cb17a6e6d734da1b1b0946f4a2157734297dce990f969320163f79671f0ccc0
```

Image

Issue #8123, opened 2026-04-20
[cost-monitoring] Restore authenticated access to Prometheus


jnywong:
### Context

Following great work in https://github.com/2i2c-org/infrastructure/pull/8114, any application on 2i2c infra now needs to authenticate any API call to the Prometheus server.

### Proposal

Following sprint planning on Apr 29, we decided to treat this as an outage and the Definition of Done below has been repurposed to follow the incident process.

### Definition of Done

(may need to be broken up into another sprint)

- [x] Fix this for the cost monitoring system and document
- [x] Double-check that this does not affect any other applications
- [x] Build and rollout to all clusters
- [x] Update communities using "proactive outreach" on FreshDesk
- [ ] Add alerting so that this service outage gets picked up if it happens again
- [ ] Retroactively add incident to PagerDuty
- [ ] Write an incident report
- [ ] Schedule review from Tech Lead and Engineering Manager
- [ ] Publish incident report

jnywong:
Developing and shipping a fix, as well as instrumenting alerting, turns out to be a sizeable lift so bumping the effort on this from a 3 to an 8.

aprilmj:
Consider splitting this into two parts: restore service (done!) and set up monitoring, finish incident process et al.

jnywong:
I've updated the top card with the work remaining

PR #8238, opened 2026-05-06
feat: add alerts for cost monitoring system

jnywong:
Ref https://github.com/2i2c-org/infrastructure/issues/8123

### To Do

- [ ] Update `support/values.jsonnet` with grouped cost-monitoring alerts
- [ ] Ensure PagerDuty system is set up to route these alerts to app engineers

### DIFF
--- a/helm-charts/support/Chart.yaml
+++ b/helm-charts/support/Chart.yaml
@@ -59,6 +59,6 @@ dependencies:
   # jupyterhub-cost-monitoring, cost reporting for Grafana
   # https://github.com/2i2c-org/jupyterhub-cost-monitoring
   - name: jupyterhub-cost-monitoring
-    version: "1.1.2"
+    version: "1.1.4"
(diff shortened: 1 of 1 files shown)


yuvipanda:
@jnywong what is needed to drive this to completion?

jnywong:
The image bumps is so that the cost monitoring services exports Prometheus metrics for every HTTP request made to the system 

- https://github.com/2i2c-org/jupyterhub-cost-monitoring/pull/113

plus updating the helm chart to add pod annotations to tell Prometheus to scrape from the cost-monitoring pod.

The remaining work in this PR is to come up with a sensible alert(s) for the cost monitoring application that I want to be alerted on via PagerDuty. I'll add this to the top card so that's clearer.

I've not been able to push this work to completion because I've been focused on other things, but I hope we can pick this up again as goal-driven work for this quarter.

PR #8247, opened 2026-05-06
Add post-auth webhook to notify MAAP API on user login

bsatoriu:
### Summary

- Adds a post_auth_hook to the MAAP hub that calls GET /api/members/self on the MAAP REST API after a user authenticates via Keycloak
- Authenticates the request using the same jwt:{id_token} format used by the spawner's existing auth_state_hook
- Adds a per-environment maapApiBaseUrl custom config value (staging: api.uat.maap-project.org, prod: api.maap-project.org)
- The webhook is non-blocking — failures are logged but do not prevent login

### DIFF
--- a/config/clusters/maap/common.values.yaml
+++ b/config/clusters/maap/common.values.yaml
@@ -116,18 +116,12 @@ jupyterhub:
 
         c.Spawner.auth_state_hook = populate_token
       002-post-auth-webhook: |
+        import asyncio
         from tornado.httpclient import AsyncHTTPClient
         from z2jh import get_config
 
-        async def maap_post_auth_hook(authenticator, handler, authentication):
-            """Notify MAAP REST API after successful login."""
-            base_url = get_config("custom.maapApiBaseUrl", "")
-            if not base_url:
-                return authentication
-
-            auth_state = authentication.get("auth_state", {})
-            pgt = f"jwt:{auth_state.get('id_token', '')}"
-
+        async def _notify_maap_api(authenticator, base_url, pgt):
+            """Fire-and-forget call to MAAP REST API."""
             http_client = AsyncHTTPClient()
             try:
                 await http_client.fetch(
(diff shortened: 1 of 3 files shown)


config/clusters/maap/common.values.yaml
```diff
@@ -115,6 +115,33 @@ jupyterhub:
...
+            auth_state = authentication.get("auth_state", {})
+            pgt = f"jwt:{auth_state.get('id_token', '')}"
+
+            http_client = AsyncHTTPClient()
+            try:
+                await http_client.fetch(
+                    f"{base_url}/api/members/self",
+                    method="GET",
+                    headers={"Authorization": f"Bearer {pgt}"},
+                    request_timeout=5.0,
```
agoose77, line 124: Does this need to be "blocking" for the login progression? Or can we use a background task?

bsatoriu: @agoose77 I changed this to a non-blocking call.

config/clusters/maap/common.values.yaml
```diff
@@ -115,6 +115,38 @@ jupyterhub:
...
+                )
+            except Exception as e:
+                authenticator.log.warning("MAAP post-auth webhook failed: %s", e)
```
yuvipanda:
```suggestion
                authenticator.log.exception("MAAP post-auth webhook failed: %s", e)
```

Provides better tracebacks I think

config/clusters/maap/prod.values.yaml
```diff
@@ -6,6 +6,7 @@ userServiceAccount:
     eks.amazonaws.com/role-arn: arn:aws:iam::916098889494:role/maap-prod
 jupyterhub:
   custom:
+    maapApiBaseUrl: https://api.maap-project.org
```
yuvipanda: can you nest this under `maap` so you can add further values here in the future? So it would be:

```
maap:
   apiBaseUrl:
```

config/clusters/maap/common.values.yaml
```diff
@@ -115,6 +115,38 @@ jupyterhub:
                 })
 
         c.Spawner.auth_state_hook = populate_token
+      002-post-auth-webhook: |
```
yuvipanda, line 111: Can you add a comment here about who is responsible for this code? 2i2c engineers don't have clear visibility here, and it would be very helpful to know in case we are debugging something.

yuvipanda (changes requested):

yuvipanda (approved):

PR #8257, opened 2026-05-07
Increase resource limits and requests for MAAP sidecar pods

bsatoriu:


agoose77:
See https://2i2c.slack.com/archives/C04R9B69T5W/p1778171836698939 for discussion.

PR #8262, opened 2026-05-07
internaltools: Collect annotations about pods

yuvipanda:
This collects `kube_pod_annotations` metric, which is potentially useful for doing billing related MAU calculations

- closes https://github.com/2i2c-org/meta/issues/3209

### DIFF
--- a/terraform/internaltools/variables.tf
+++ b/terraform/internaltools/variables.tf
@@ -17,6 +17,11 @@ variable "federated_metrics_allowlist" {
       name         = "active-users"
       metric_regex = "jupyterhub_active_users"
       labels_regex = "period|namespace"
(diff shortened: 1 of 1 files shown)


yuvipanda:
> i also want to think about this some more, because this means internaltools will have a lot of PII (usernames as well as when they were online)

don't merge yet

choldgraf:
Noting an issue to keep on top of and try to resolve as we do this work:

- https://github.com/2i2c-org/meta/issues/3153

yuvipanda:
Going to merge this, so we can start keeping session activity info long term. It does mean we get usernames stored long term, which I think is fine. I won't try to backfill this.

Issue #8251, opened 2026-05-07
[Prometheus] Config unable to reload due to 401 errors


jnywong:
### Context

From [Prometheus docs](https://prometheus.io/docs/prometheus/latest/configuration/configuration/):

> Prometheus can reload its configuration at runtime. If the new configuration is not well-formed, the changes will not be applied. A configuration reload is triggered by sending a SIGHUP to the Prometheus process or sending a HTTP POST request to the /-/reload endpoint (when the --web.enable-lifecycle flag is enabled). This will also reload any configured rule files.

We have the `--web.enable-lifecycle flag` enabled, but when inspecting pod logs, there are many errors that say

```bash
level=error ts=2026-05-07T11:28:51.126850253Z caller=runutil.go:117 msg="function failed. Retrying in next tick" err="trigger reload: received non-200 response: 401 Unauthorized; have you set `--web.enable-lifecycle` Prometheus flag?"
```

### Proposal

- Investigate when the reload is triggered
- Figure out how to pass Prometheus creds to reload trigger

### Definition of Done

- [ ] Prometheus auth details are stored in a secret on all clusters
- [ ] Prometheus reload url includes auth credentials on all clusters
- [ ] Prometheus server no longer errors when config is reloaded

sunu:
This is fixed for Earthscope in https://github.com/2i2c-org/infrastructure/pull/8264. I've updated the definition of done to roll out this pattern to all remaining clusters as well.

jnywong:
Can this be included to provide prometheus auth as a secret for all namespaces, not just support? This makes it easier for me to use the same pattern for staging/prod applications that also need these creds.

PR #8249, opened 2026-05-07
feat: update server startup alert

jnywong:
Closes https://github.com/2i2c-org/infrastructure/issues/8203

PR #8264, opened 2026-05-08
earthscope: add auth info to prometheus config reload url

sunu:
related to https://github.com/2i2c-org/infrastructure/pull/8249.

Appends credentials to the Prometheus config reload URL in the form `http://username:password@127.0.0.1:9090/-/reload` to prevent config reload failures caused by 401 authentication errors.

cc @jnywong 

Update: adds a secret to the support chart to store prometheus auth values. the config-map reload URL references auth credentials from that secret.


jnywong:
I think this would close https://github.com/2i2c-org/infrastructure/issues/8251

sunu:
> I think this would close https://github.com/2i2c-org/infrastructure/issues/8251

@jnywong yes, but only once we've updated the config for all the other hubs to match what we did for the earthscope hub in this PR!

config/clusters/earthscope/enc-support.secret.values.yaml
```diff
@@ -1,52 +1,60 @@
...
+    enc: CiUA4OM7eL5BTyD6u6+W6UECOfD5t8/n8WTvxvJaF7DKO1gk8+XIEkgAmy+eknknFRkQtor/h0oMJhVcgoA1nNhqSEBOtJViKorlm5jO57/nvICMoDKWOW7ctJZha6wDMz6bsky/LeMZYYqN7QeiDno=
+  azure_kv: []
+  hc_vault: []
+  age: []
+  lastmodified: '2026-05-11T07:36:45Z'
+  mac: ENC[AES256_GCM,data:747MBFIZzEY/hKj0SSqAEjoOG0RMknBi3S9leT4h7WtyVBkYqmeQFuT2TErnmDNYDkt3U2Vh3WbVNjgZS17OBRUqe43bTVaYClXG+1qEO4SGFQkVwbgNtCtCXkoKz9W92RRbFiKXuwc2QLAgqNReH3bLy+NxYRzDTwzFnwFjLq8=,iv:HfpRUkBJlJwvuV5jaXpyOxGDXDNAOVkdUKWvn22IB2A=,tag:egtu0vyhrj87AO+/M9KQWg==,type:str]
+  pgp: []
   unencrypted_suffix: _unencrypted
-  version: 3.12.2
+  version: 3.9.0
```
GeorgianaElena, line 60: incoming version update 😅 🚀

GeorgianaElena (approved):
This looks good @sunu! thank you!

We just have to do this for all the clusters and use this secret everywhere its needed instead of duplicating it.

Do you want to update the definition of done of.https://github.com/2i2c-org/infrastructure/issues/8251 with these todos?

config/clusters/earthscope/enc-support.secret.values.yaml
sunu: oh good catch, bumped sops to 3.13.0!

sunu:
Thanks @GeorgianaElena! I've updated the definition of done in #8251

Issue #8200, opened 2026-04-29
Do a pass through config and standardize into jsonnet


GeorgianaElena:
### Context

We have repetitive config among the hubs that would benefit from being standardized into jsonnet.
Some examples of Yuvi's recent work standardizing some:
- https://github.com/2i2c-org/infrastructure/pull/8145
- https://github.com/2i2c-org/infrastructure/pull/8191

### What we need to do

We should do a quick, initial, pass though the config and standardize the most straightforward ones.

### Definition of Done

- [ ] Remove all repeated ingress config
- [ ] Remove `nfs.pv.serverIP` on GCP hubs
- [ ] Remove `userServiceAccount.annotations`
- [ ] Remove `nodeSelector`

yuvipanda:
I refined the 'definition of done' to add a few more things

GeorgianaElena:
Updating the effort number on this one from a 3 to a 5 because it was and still is lots of config to centralize 🚀

PR #8283, opened 2026-05-11
Centralize nfs.pv.serverIP config for gcp clusters

GeorgianaElena:
Ref #8200

### DIFF
--- a/helm-charts/basehub/values.jsonnet
+++ b/helm-charts/basehub/values.jsonnet
@@ -109,6 +109,12 @@ local jupyterhubGroupsExporterConfig = {
   },
 };
 
+local pvConfig = {
+  pv: {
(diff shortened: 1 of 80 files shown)


agoose77:
A GIF saying "Oh my god. It's beautiful"

helm-charts/basehub/values.jsonnet
```diff
@@ -109,14 +109,20 @@ local jupyterhubGroupsExporterConfig = {
   },
 };
 
+local pvConfig = {
+  pv: {
+    serverIP: 'storage-quota-home-nfs.%s.svc.cluster.local' % hub_name,
+  },
```
yuvipanda: let's put the conditional here, rather than down below. That way we can do more `pv` config for gcp / aws if needed

Issue #8254, opened 2026-05-07
CIROH DevCon 2026


aprilmj:
### Context

This initiative connects several support tasks so we can find all AWI-CIROH tickets around prep and support for their annual conference in one place.

### What we need to do

See sub-issues

### Definition of Done

- [ ] The CIROH team is prepped for their conference
- [ ] We're aware of their plans so we know what types of issues may pop up
- [ ] As needed (as part of CSH) we respond to tickets during their conference days

### Infrastructure Details
- 150–200 users concurrently with small/medium/large mix
- Maybe GPUs

PR #8284, opened 2026-05-11
[awi-ciroh, workshop] Bump node-sharing setup to n2-standard-64

agoose77:
This workshop needs 32 ~16vCPU ~64GiB instances. To improve startup behaviour (where we are not worried about over provisioning like we are for day-to-day scenarios), this PR adds a new profile option for packing four "Huge" instances into n2-standard-64.

I have not updated the other profile options — I think we should let the community know about that kind of change ahead of time.

Part of #8254

### DIFF
--- a/terraform/gcp/projects/awi-ciroh.tfvars
+++ b/terraform/gcp/projects/awi-ciroh.tfvars
@@ -67,20 +67,6 @@ notebook_nodes = {
     max : 100,
     machine_type : "n2-highmem-4",
   },
-  "n2-highmem-16" : {
-    min : 0,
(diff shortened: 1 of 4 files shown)


terraform/gcp/projects/awi-ciroh.tfvars
```diff
@@ -67,20 +67,6 @@ notebook_nodes = {
     max : 100,
     machine_type : "n2-highmem-4",
   },
-  "n2-highmem-16" : {
```
agoose77, line 70: From GKE upgrade.

PR #8289, opened 2026-05-11
Offboard James from the hub admins

agoose77:
See https://github.com/2i2c-org/meta/issues/3123


yuvipanda:
@agoose77 it may still need an active API call or similar to get the user itself deleted from the hub.

agoose77:
@yuvipanda yes, I have that filed for today!

PR #8304, opened 2026-05-12
Centralize nfs.pv.mountOptions into jsonnet

GeorgianaElena:
Ref #8200


config/clusters/2i2c-aws-us/basehub-common.values.yaml
```diff
@@ -3,14 +3,6 @@ nfs:
   pv:
     enabled: true
     # from https://docs.aws.amazon.com/efs/latest/ug/mounting-fs-nfs-mount-settings.html
```
yuvipanda, line 5: should probably remove this comment too

PR #8325, opened 2026-05-13
[hhmi] Use tmpauthenticator with auto_login:false

### LINKED ISSUE #8303: Investigate trivial roadblock to link crawling on HHMI
c.f. https://github.com/2i2c-org/meta/issues/2911

HHMI hubs might have been seeing demand from link crawler-style bots. We can explore config on `tmpauthenticator` or look at new solutions that introduce a level of human-in-the-loop:
```python
c.TmpAuthenticator.auto_login = False
```

Other ideas include adding a `SharedPasswordAuthenticator`, and writing the password _in_ the home page template (so that it's not actually a secret).

We should try to avoid bespoke work here unless it turns out its needed.

### Definition of Done
- [ ] Link-crawlers cannot trigger resource usage.

agoose77:
I haven't checked how this impacts the Binder flow, which I would like to do before merging

Closes #8303

### DIFF
--- a/config/clusters/hhmi/binder.values.yaml
+++ b/config/clusters/hhmi/binder.values.yaml
@@ -69,7 +69,9 @@ jupyterhub:
       BinderSpawnerMixin:
         auth_enabled: false
       JupyterHub:
-        authenticator_class: 'null'
+        authenticator_class: tmp
+      TmpAuthenticator:
+        auto_login: false
     redirectToServer: false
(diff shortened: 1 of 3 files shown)


agoose77:
Merging this as it's in production. We need to circle back and ensure it behaves properly / solves problems. Particularly w.r.t the BinderHub page itself being unauthenticated.

PR #8330, opened 2026-05-14
fix: EBS alerts via CloudWatch

agoose77:
This PR fixes the EBS CloudWatch threshold alerts. It will trigger if IOPs or throughput saturate for three minutes over a five minute interval.

This can be triggered using the `fio` tool from conda-forge, e.g. for IOPs:
```bash
#!/usr/bin/env bash

set -eu

printf %s\\n {0..12} | xargs -I{} -n 1 -P 0 fio --randrepeat=1 --ioengine=libaio --direct=1 --gtod_reduce=1 --name=test --filename=test{} --bs=4k --iodepth=64 --size=50M --readwrite=randrw --rwmixread=75 --loops 100
```


agoose77:
Applied!

PR #8333, opened 2026-05-14
[earthscope, staging] Update config

jnywong:
Ref https://github.com/2i2c-org/jupyterhub-usage-quotas/issues/38


jnywong:
Thanks for the tweaks @agoose77 🙌

PR #8338, opened 2026-05-15
[spike] Split out custom Hub config

agoose77:
Our JupyterHub config is quite large (1701 lines). Much of this is inline vendored Python. Through time, this has been tested, but if we were to write this from scratch it would be a difficult exercise to avoid introducing string formatting bugs and other problems.

This PR is a spike to pull out the config files into lint-able files. The problem with this spike is that the JupyterHub deployment does not depend upon the checksum of the configmap, so changes to these config files do not lead to a hub restart. It seems quite difficult to plumb that feature in to the base chart, and I suspect that's a good reason for why things are done this way at the moment.

CC @GeorgianaElena because it's relevant to our quarterly objective, but I'm not asking any time from you here!

Some questions:
- Is this worth pursuing?
- Can we use a Helm `postinstall` hook here to trigger a hub reload? Is that acceptable as additional moving parts? https://helm.sh/docs/topics/charts_hooks/
- Could/should we use Jsonnet to handle the error-prone configmap mount expansion? Or perhaps the upstream z2jh should support loading `jupyterhub_config.d/foo/bar.py` automatically.

> [!Note]
> I haven't tested this deployment.


yuvipanda:
What's the smallest unit of work we can do to move this forward?

agoose77:
I did this work when the refactoring efforts we've been making were less mature.

I still think we should aim to split up our hub config _and_ try to lift our Python files into something that conventional tooling can see.

The next step would be to create a new PR that:
1. Adds a `configmap`
2. Lifts _one_ bit of config into it.

It remains true that I am unsure how to solve 
> The problem with this spike is that the JupyterHub deployment does not depend upon the checksum of the configmap

This felt like a blocker requiring e.g. a Helm postinstall hook. But, looking at this again (the best part about spikes), I suspect checksumming might work _or_ we can compute the checksum of the files.

PR #8339, opened 2026-05-15
Update workshop.values.yaml to modify devcon entries

benlee0423:
Removed ewri image and added hillslope and emulator with their respective images.


benlee0423:
@yuvipanda @agoose77 
Please merge it.

Issue #7552, opened 2026-02-02
[Compute usage quotas] General review


jnywong:
### Definition of Done

- [ ] Secure package dependencies
- [ ] Trusted publishing and CI/CD
- [x] Logging, monitoring and alerting
   - https://github.com/2i2c-org/infrastructure/pull/8347
- [x] Testing
- [ ] Resource allocation and stress testing of the hub pod
- [ ] Make sure accounts like `deployment-service-check` don't get resource limited
- [x] #8282
- [x] https://github.com/2i2c-org/jupyterhub-usage-quotas/issues/38
- [ ] Documentation

jnywong:
### Thursday 14 May

Spent a lot of time consolidating config today to ease deployment in a z2jh context. A little stuck on how multiple config files get loaded and merged in https://github.com/2i2c-org/infrastructure/pull/8333 but [pinged for help in Slack](https://2i2c.slack.com/archives/C07SJJWVCAD/p1778778042851379)

jnywong:
I'm keeping a bit of a daily log here since this issue encompasses a lot of things.

---

### Friday 15 May

No longer stuck on the config and https://github.com/2i2c-org/infrastructure/pull/8333 was merged 🚀 

Today I spent time instrumenting metrics to monitor the health of the usage quota system in https://github.com/2i2c-org/jupyterhub-usage-quotas/pull/51, as well as updating the quickstart documentation in https://github.com/2i2c-org/jupyterhub-usage-quotas/pull/50

jnywong:
### Monday 18 May

Spent more time improving the metrics for monitoring and testing these improvements on the staging hub. WIP for alert config https://github.com/2i2c-org/infrastructure/pull/8347

Todo:

- [x] https://infrastructure.2i2c.org/howto/manage-alerts/#how-to-add-a-new-alert 👈 follow this guide for routing alerts to a new receiver and PD service
- [x] update 2i2c infra docs with new alert

jnywong:
### Tuesday 19 May

- Spent time configuring PD and updating docs in #8347
- Review from #8282 
- Bunch of meetings in the afternoon

### Wednesday 20 May

- Slack integration with PD was confusing but I documented it in infrastructure guide now
- Tested alerts with Earthscope staging -- all good and #8347 is now merged

jnywong:
### Thursday 21 May

[...]

PR #8347, opened 2026-05-18
[earthscope, staging] configure alerts

jnywong:
Ref https://github.com/2i2c-org/infrastructure/issues/7552

- [x] Configure usage quota alerts for earthscope staging
- [x] Add pod restart alert for prometheus server, since this is now on critical path for usage quota system

Note that the alert rules are manually updated in the support values file for Earthscope for now, but the idea would be to roll this out network-wide and refactor into jsonnet pending sufficient testing of the Earthscope cluster.

### DIFF
--- a/config/clusters/earthscope/staging.values.yaml
+++ b/config/clusters/earthscope/staging.values.yaml
@@ -10,7 +10,7 @@ basehub:
     hub:
       image:
         name: quay.io/2i2c/jupyterhub-usage-quotas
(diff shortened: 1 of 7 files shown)


jnywong:
@GeorgianaElena this is still WIP but could you take a look at my proposed alerts and PD escalation policies for potential application outages in general?

GeorgianaElena (approved):
LGTM @jnywong! Thank you for going along with the structure 🚀

PR #8372, opened 2026-05-21
awi-ciroh: Deploy a 5Gi 'scratch' space that clears after each session

yuvipanda:
Just a test, to see how this may be useful (or not)

### DIFF
--- a/config/clusters/awi-ciroh/common.values.yaml
+++ b/config/clusters/awi-ciroh/common.values.yaml
@@ -44,6 +44,22 @@ basehub:
       defaultUrl: /lab
       cloudMetadata:
         blockWithIptables: false
+      storage:
+        extraVolumes:
+          002-scratch-session:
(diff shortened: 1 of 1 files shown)


config/clusters/awi-ciroh/common.values.yaml
```diff
@@ -44,6 +44,22 @@ basehub:
...
+            name: scratch-session
+            emptyDir:
+              # Limit total size, to prevent pod being evicted
+              # with over use by any one user. This can be increased,
+              # but may need to increase the size of the node disk
+              sizeLimit: 5Gi
+        extraVolumeMounts:
+          002-scratch-session-jovyan:
+            name: scratch-session
+            mountPath: /home/jovyan/scratch-session
```
yuvipanda, line 59:
```suggestion
            mountPath: /tmp
```

PR #8374, opened 2026-05-21
VICTOR: Image bump and spawn screen details

volcanocyber:


volcanocyber:
@GeorgianaElena The Github Action to update staging failed, so it didnt roll out properly. I tested the new image on staging and it worked, but if i need to fix anything, please let me know.

PR #8406, opened 2026-05-27
feat: add support for commands in various exec routines

agoose77:
I regularly want to be able to do something in a pipeline against a 2i2c cluster, e.g. compute the size of the `/home` disk for disk resizing. This PR adds a standard mechanism (via `typer.Context`) to pass through commands to the shell spawned in a pod (a pattern we use frequently).

This can now be done using
```shell
deployer exec homes 2i2c-aws-us staging -- /bin/df /home
```

> [!Tip]
> For posterity, I often use the deployment labeling to get a shell where I want it, e.g.
> ```bash
> kubectl exec -it -n prod deployment/hub -- /bin/bash -l
> ```

### DIFF
--- a/deployer/dev_commands/exec/infra_components.py
+++ b/deployer/dev_commands/exec/infra_components.py
@@ -18,8 +18,9 @@ yaml = YAML(typ="safe", pure=True)
 UBUNTU_IMAGE = "ubuntu:22.04"
 
 
-@exec_app.command()
+@exec_app.command(context_settings={"allow_extra_args": True})
 def root_homes(
+    ctx: typer.Context,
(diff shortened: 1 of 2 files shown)


deployer/dev_commands/exec/infra_components.py
```diff
@@ -280,44 +284,35 @@ def homes(
...
-
+    ]
     cmd = [
         "kubectl",
         "-n",
         hub_name,
         "run",
         "--rm",  # Deletes the pod when the process completes, successfully or otherwise
         "-it",  # Give us a shell!
+        "--override-type",
```
agoose77, line 308: I switched to JSON Patch here, because the pod merge was not behaving properly w.r.t args / command

PR #8413, opened 2026-05-27
feat: Helm-managed node placeholders

agoose77:
## Motivation
2i2c runs workshops, and we sometimes get requests to pre-warm the infrastructure. It would be nice to make this quicker / easier to manage.

We have an initiative https://github.com/2i2c-org/initiatives/issues/51 that tracks the need to reserve nodes ahead of time.

That initiative aims to implement cloud-specific backends to solve the problem. The proposal is a robust approach, but in the meantime we still would benefit from some system of scaling up hubs in a way that reduces toil.

## Implementation
This PR replaces previous work to use user placeholders (which fails for >1 node) in favour of using node placeholders. These placeholders just ensure that $N$ nodes are available, and do not respond to load i.e. do not scale up as nodes become used. This differs from user placeholders, which attempt to ensure `$N$` replicas are always available.

This can be combined with the `userScheduler` to ensure that user nodes are densely packed, rather than spread out, making scale-down less disruptive.

### DIFF
--- a/.github/workflows/scale-placeholders.yaml
+++ b/.github/workflows/scale-placeholders.yaml
@@ -1,4 +1,4 @@
-name: Scale user placeholders
+name: Scale node placeholders
 on:
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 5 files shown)


agoose77:
I'm skipping deploy on this, because it's turned off by default.

PR #8470, opened 2026-06-02
refactor: set token for BinderHub in base config

agoose77:
This is a tiny change that ensures the API token and OAuth client ID for BinderHub is always set in the service container, simplifying our existing Hub configs.

This token is only needed for BinderHub's "launch" feature, i.e. not fancy profiles!

### DIFF
--- a/helm-charts/basehub/values.yaml
+++ b/helm-charts/basehub/values.yaml
@@ -64,8 +64,6 @@ binderhub-service:
           key: hub.services.binder.apiToken
     - name: JUPYTERHUB_CLIENT_ID
       value: &binderhub_client_id service-binderhub
-    - name: JUPYTERHUB_API_URL
-      value: http://hub:8081/hub/api
   config:
     BinderHub:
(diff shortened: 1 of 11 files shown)


helm-charts/basehub/values.yaml
```diff
@@ -55,6 +55,17 @@ binderhub-service:
...
+    # Z2jh automatically creates the API token
+    - name: JUPYTERHUB_API_TOKEN
+      valueFrom:
+        secretKeyRef:
+          name: hub
+          key: hub.services.binder.apiToken
+    - name: JUPYTERHUB_CLIENT_ID
+      value: &binderhub_client_id service-binderhub
+    - name: JUPYTERHUB_API_URL
+      value: http://hub:8081/hub/api
```
agoose77: I'm sure it's safe to set this in base config, but I'll turn it off for this round until we know https://github.com/2i2c-org/binderhub-service/pull/167 is merged.

```suggestion
```

PR #8475, opened 2026-06-03
Remove unnecessary scopes from the fancy-profiles binderhub role

sunu:
Fancy profiles passes user tokens to binderhub as bearer token and the fancy profiles binderhub only talks to the hub api to validate the token. And since fancy profiles uses binderhub to build images only, and not to spawn user servers, the binderhub service role doesn't really need any scopes at all.


sunu:
@yuvipanda @agoose77 I've manually deployed and tested this on GHG staging hub. But it would be good to get a second pair of eyes before we merge.

yuvipanda (approved):

PR #8494, opened 2026-06-06
Reduce size of home dir disks on showcase / staging hubs

yuvipanda:
This costs us a bunch of money without really being used. Applying this would destroy the disks, but I don't believe there should be any actively useful data in there.


jnywong (approved):
This is fine with me

PR #8521, opened 2026-06-10
VICTOR: updated staging vals + removed jupyter-ai

volcanocyber:


config/clusters/victor/staging.values.yaml
```diff
@@ -18,7 +18,7 @@ basehub:
         #            unlisted_choice to pick a custom image. If you update
         #            either, update the other as well.
         #
-      - display_name: CPU Only
+     - display_name: CPU Only
```
GeorgianaElena:
```suggestion
      - display_name: CPU Only
```

PR #8528, opened 2026-06-12
[earthscope] Add /tmp scratch storage

agoose77:
We already have this implicitly due to the kubelet providing ephemeral storage. This PR now codifies the upper bound via `emptyDir`. This provides 20GiB per user, and sets the performance of the disk to 3000 IOPs, 500 MiB/s. I chose the Throughput with consideration to general performance needs — a regular non-cloud SSD is ~5000MiB/s, and a cloud SSD more like 1200MiB/s — but, I suspect that 500MiB/s is above our home directory performance at the moment, so it's an improvement (and not crazy expensive).

### DIFF
--- a/eksctl/libsonnet/cluster.jsonnet
+++ b/eksctl/libsonnet/cluster.jsonnet
@@ -1,6 +1,7 @@
 local escapeName(name) = std.strReplace(name, '.', '-');
 local lowerCaseLetter(i) = std.char(97 + i);
-local trim63(s) = if std.length(s) > 63 then s[0:62] else s;
+// Build 63-char (max) name with untouched generation.
+local buildName(parts, generation) = std.join('-', parts)[:63 - 1 - std.length(generation)] + '-' + generation;
 
 {
   /**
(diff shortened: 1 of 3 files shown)


eksctl/libsonnet/cluster.jsonnet
```diff
@@ -1,6 +1,7 @@
 local escapeName(name) = std.strReplace(name, '.', '-');
 local lowerCaseLetter(i) = std.char(97 + i);
-local trim63(s) = if std.length(s) > 63 then s[0:62] else s;
+// Build 63-char (max) name with untouched generation.
+local buildName(parts, generation) = std.join('-', parts)[:63 - 1 - std.length(generation)] + '-' + generation;
```
agoose77, line 4: @GeorgianaElena I had to tweak the node names so that we don't trim the generation. Otherwise, it's not possible to create separate generations of dask nodepools (as they map to the same name).

I think this will need handling for all of our eksctl cluster that have dask nodepools, but I don't have time to do all of them today. So, I propose that we fix this on a per-nodepool basis.

GeorgianaElena: ah, haven't thought about this.
Alternativelly, I think we can add the generation as a prefix then?

GeorgianaElena: thanks for fixing it

Issue #8362, opened 2026-05-20
Support - Pythia Cookoff (June 15-18)


aprilmj:
### Context

This card captures during-workshop support for the Cookoff. @choldgraf will be there in person. @agoose77 is traveling to another workshop. This is "tools down in the evening" support - do not plan to spend after hours time. 

### What we need to do

The infra eng team will be supporting remotely. @choldgraf will be there in person. 

We should scale up their nodes to:
```
ceil(
  (45 users * 30 GiB) / memory(r5.4xlarge)
)
```
for the duration of the workshop.

### Definition of Done

- #8532 
- #8537 
- #8539 
- #8556 
- #8557

agoose77:
@GeorgianaElena I think you're going to be the engineer who is in office during the cookoff. As discussed in the notes above, given your timezone I think we will need to be realistic that support will be quite limited. Chris will be at the event, and I may be reachable too.

Here's what I've done to prepare for the workshop:

## Auth and UX
1. Pythia have [some config that propagates the GitHub information](https://github.com/2i2c-org/infrastructure/blob/c8cd52aa85d4add4c9a47afc9f209b3522c06cca/config/clusters/projectpythia/common.values.yaml#L43-L57) into the user environment. 
2. I set up [a temporary credential helper](https://github.com/2i2c-org/infrastructure/blob/c8cd52aa85d4add4c9a47afc9f209b3522c06cca/config/clusters/projectpythia/common.values.yaml#L86-L141) that retrieves the GitHub token from the Hub's auth state. I did this because at one point I was under the impression that the GitHub OAuth login tokens could expire, so I wanted a nice way for the `singleuser` image to avoid needing a restart. I don't actually think that's the case, though — only GitHub Apps have refresh tokens.
3. I [added a link in the hub menu](https://github.com/2i2c-org/infrastructure/blob/c8cd52aa85d4add4c9a47afc9f209b3522c06cca/config/clusters/projectpythia/common.values.yaml#L143-L157) to revoke the GitHub token used for log-in. This can be used if a token is thought to have leaked. It just links to the GitHub app page.

These changes mean that after login, users don't need to touch `gh-cli` or other tools to clone their repos. I view this as a time-limited feature — we don't love long-lived tokens being available like this.

## Nodes
We've committed to spinning up N nodes at the start of the workshop. They have ample cloud credits, so I think we can just leave them running for the 3 days. I'll confirm that, though.

GeorgianaElena:
[...]

PR #8537, opened 2026-06-15
[projectpythia] Bump rate limit for BinderHub service

agoose77:
This is a temporary PR to fix rate limiting for image builds
Ref #8362


agoose77:
Skipping deploy, as I've already done this.

PR #8541, opened 2026-06-15
Document 2i2c cloud cost metadata a bit better

### LINKED ISSUE #8540: Buggy code to create table of hubs
The table of hubs lists the Pythia hub as running on Azure, but this is almost certainly not true, we think it's JetStream2. @agoose77 noted this is probably a bug! He said we should look here:

ref: `docs/helper_programs/utils.py`

choldgraf:
Cleans up our hub table code, by doing a few things:

1. Improves the cloud provider infer code if we use `kubeconfig` (I think this was mistakenly marking things as Azure that were actually jetstream)
2. Clarifies how we mark a cluster as paid by us
3. Adds that as a column in the list of hubs

---

- closes https://github.com/2i2c-org/infrastructure/issues/8540


yuvipanda (dismissed):
I think we should move away from this. The entire `billing` piece is for use by a piece of code that doesn't really work in the deployer. There's no clear process for keeping this info up to date. I also have no idea if the existing information is *correct*.

I think instead, we should get rid of this info in `cluster.yaml`, and it should instead be in hubspot, and we should follow the link from here through the deal id to figure that out.

choldgraf:
Two questions: 

What other pieces of infrastructure information should be in hubspot then? "We pay" might be different depending on the cluster for a given contract 

Can I use this "we pay" variable as a source of truth for doing an initial data transfer to hubspot?

yuvipanda:
> What other pieces of infrastructure information should be in hubspot then?

I think of 'who pays' as *not* infrastructure information, but contract information. So I would say they should live in hubspot for that reason. 

> Can I use this "we pay" variable as a source of truth for doing an initial data transfer to hubspot?

this information has not been actively used by anyone for a while now. I would use whatever the cloud billing invoicing process has been as the source of truth, and not this information.

choldgraf:
Is "we pay" something that is always true or false for a contract? Are there any contracts for which we pay for some infrastructure but not for other infrastructure?

Either way I'm going to descope "who pays" from this PR entirely and just focus it on un breaking the list of hubs

yuvipanda:
+1 to descope.

yuvipanda:
I opened https://github.com/2i2c-org/infrastructure/issues/8548

choldgraf:
I removed the cloud costs stuff and linked to our compass. I could not easily find a source of truth for "do we pay the cloud costs". I suspect that's why @colliand was looking at this infra repo to try and answer this for Pythia. @haroldcampbell may have a better idea. Either way, I think this PR has a tighter scope now and might be ready to go

PR #8545, opened 2026-06-16
Remove the reactive CSH sync workflow

choldgraf:
This removes the custom reactive CSH sync workflow. I'm going to try accomplishing this with a CRON job that runs from the data-sync repo instead...this should reduce the amount of config we have spread across our repos. If it's really too clunky this way, we can add it back in! I just wanted to try it the simpler way for now

---

- closes https://github.com/2i2c-org/data-sync/issues/6


GeorgianaElena:
Aaah, sorry for merging https://github.com/2i2c-org/infrastructure/pull/8555 yesterday.
I wasn't aware of this PR and just got overloaded by a bunch of failed workflows after creating a series of issues and thought I should act 🙄

choldgraf:
no worries at all - I decided to go ahead and just remove this worklfow call in general, so let's merge that in and I can add it back in if we want to re-add it haha

Issue #7411, opened 2026-01-15
[Spike 4h] Integrate resource generation with structural configuration model


agoose77:
### Context
2i2c deploys JupyterHubs to a diverse yet finite set of Kubernetes platforms:
1. GKE
2. EKS
3. AKS
4. Jetstream2
5. misc.

For the 35 clusters we have the following distribution
```bash
$ rg "provider:\s*(\S*)" config/clusters/*/cluster.yaml -o -r '$1' -I | sort |  uniq -c
23 aws
8 gcp
3 kubeconfig
```

For each cloud provider, there are a finite set of node types, and within these there are a reasonable set of cost-efficient nodes for the typical workloads that JupyterHub users impose. Furthermore, there are optimal strategies to apportion individual node resource capacities amongst user workloads (user servers, Dask workers) to efficiently pack pods. 

As such, most hubs should be using profile lists whose resource allocations are derived computationally as a function of the baseline requirements (how many CPUs, and/or RAM) and the platform provider. In #6213 we partially refined some work to improve the consistency of our resource allocation approach. This spike is more opinionated — we should aim to avoid per-hub bespoke resource configurations in favour of something that we can generate at the platform level, and update once for each k8s provider.

#### How 2i2c decides profile options
In order to understand our goals here, it's worth thinking about how we decide upon the resource allocations for a new community. A typical way of deciding on resource allocations is something like:
```mermaid
graph LR;
initial[Pick CPU, RAM for workload] --> packing
packing[Select smallest node for baseline active users] --> run
run[Run hub] --Optimise machine--> packing
```

Communities might not have a strong sense of the requirements of their workflows. In these cases, we typically choose a particular profile(s) for them.

At this point, we have a nominal workload profile, e.g `{ 2GB, 500m }`, and a node type.[^note-degen] For many CPU + RAM combinations, there are several to choose from (e.g. `r5.xlarge`, `r5.2xlarge`) due to the subdivision strategy. By choosing a packing density of user pods, we reduce the number of node choices to one.

### What we need to do
We need to investigate the best way to allow individual hubs to define particular resource requirements. 

Without requiring us to individually modify $N*M$ hub configurations each time we need to refine these choices. It should be possible to e.g. modify the base hub configuration once, and have our deployer pipeline roll out improvements to all $N$ clusters.

[...]

PR #8549, opened 2026-06-17
Draft work for using jsonnet to generate profile lists, taking advantage of the resource allocation script

GeorgianaElena:
Ref https://github.com/2i2c-org/infrastructure/issues/7411

## Feeback needed
Before building more logic around this, I want to get your input first on this. So, at the moment this PR presents the desired state of where a hub's profileList config will come from.

## The general structure

- in the `basehub` helm chart we'll have
  - a base jsonnet function that knows how to build a `profile` 
  - a collection of resource allocation `json` files generated with the resource allocation script, that the hubs will use 
  - *_the idea is to be able to quickly re-generate them, in one place, when we need to (workshops, we need to leave more wiggle room, etc), using the script_
- the `cluster.yaml` will have a new field that we'll tell us where to get the profileList config from
- we could then have a per cluster libsonnet with common stuff, and have each hub dynamically modify what's needed
- *_this should reduce duplicated code_

## Questions
- Is this too complicated?
- Should we instead invest our energy into adding support for dict profileLists? (https://github.com/jupyterhub/kubespawner/pull/919 has stalled)
- Other ideas?


yuvipanda:
This should be independent of https://github.com/jupyterhub/kubespawner/pull/919!

I'm excited this is being worked on.

config/clusters/nmfs-openscapes/cluster.yaml
```diff
@@ -26,6 +26,8 @@ hubs:
   - enc-common.secret.values.yaml
   - staging.values.yaml
   - enc-staging.secret.values.yaml
+  profiles_source_files:
```
yuvipanda, line 29: Can we instead just rely on the fact that you can put regular `jsonnet` files under `helm_chart_values_files` already?

config/clusters/nmfs-openscapes/common.libsonnet
```diff
@@ -0,0 +1,157 @@
+{
+  local profileList = import '../../../helm-charts/basehub/profile_lists_defaults/libsonnet/profileListsFunc.jsonnet',
+  local r5xlarge_5 = import '../../../helm-charts/basehub/profile_lists_defaults/choices/r5.xlarge-5.json',
+  local r54xlarge_2 = import '../../../helm-charts/basehub/profile_lists_defaults/choices/r5.4xlarge-2.json',
+
+  local requestsChoices = profileList.makeChoiceDefault(r5xlarge_5 + r54xlarge_2, 'mem_2_gb'),
+
+  local defaultProfileImageChoices = {
```
yuvipanda, line 8: I would also love for us to keep *image lists* in yaml, and only be using jsonnet for the resource allocation. I suppose that requires https://github.com/jupyterhub/kubespawner/pull/919?

agoose77: @yuvipanda what's your incentive for preferring YAML? I agree that YAML is more readable than JSON, but the whitespace sensitivity is also a bit of a footgun. Jsonnet seems like a nice trade-off between the two, because it's more like JSON5.

My view is that trying to add dict-support to upstream charts is an uphill battle to work around the limitations of the YAML DSL. I'm in favour of acknowledging this by making Jsonnet more first class in our deployment workflow.

That's a bit more work, but fairly easy to move over to systematically. I'll expand in my top-level review comment.

agoose77:
@GeorgianaElena thank you for spiking this!

My gut feeling is that we need the flexibility of Jsonnet — expressing things like "adapt these _kinds_ of profiles to workshop packing vs regular node packing" etc. is much much easier (and more readable) in Jsonnet than YAML. I acknowledge that our third-party contributors are more familiar with YAML, but we don't have to give that up (see more later).

I think I'd be interested in really pushing this further, and _not_ using Helm as the config merging tool — what if instead of 
```yaml
hubs:
- name: staging
  display_name: Reflective Cloud - staging
  domain: staging.reflective.2i2c.cloud
  helm_chart: basehub
  helm_chart_values_files:
  - common.values.yaml
  - staging.values.yaml
  - enc-staging.secret.values.yaml
- name: prod
  display_name: Reflective Cloud - prod
  domain: reflective.2i2c.cloud
  helm_chart: basehub
  helm_chart_values_files:
  - common.values.yaml
  - prod.values.yaml
  - enc-prod.secret.values.yaml
```
we just had 

```yaml
hubs:
- name: staging
  display_name: Reflective Cloud - staging
  domain: staging.reflective.2i2c.cloud
  helm_chart: basehub
  values: staging.jsonnet
  secret-values:
  - enc-staging.secret.values.yaml
- name: prod
  display_name: Reflective Cloud - prod
  domain: reflective.2i2c.cloud
  helm_chart: basehub
  values: prod.jsonnet
  secret-values:
  - enc-prod.secret.values.yaml
```

i.e. we get rid of multiple values files and just use Jsonnet imports. 

I suspect that we'll want to make it easier for users to contribute their own profiles. We could either add documentation for how to do that with Jsonnet, or just write
```json
local images = std.parseYaml(importstr 'images.yaml');
```

config/clusters/nmfs-openscapes/common.libsonnet
GeorgianaElena: @agoose77, I think we should think about this under a framing that @yuvipanda mentioned recently, in the context of this quarter work: _"What's the least amount of work that we can do and could give us max results"_  

I'm not saying that using JSON configs instead of YAML wouldn't make our deployments more potent. If anything, I'm sure they'll allow us to standardize and automate more. But I don't think we're in a place that allows to invest the time and energy in making this move now. 

With upstream `profileLists` support, we've already got a nice starting point to make our life's better.

Issue #8242, opened 2026-05-06
[Support] EarthScope Workshop - MsPASS July 13-17 2026


aprilmj:
### The Freshdesk ticket link

https://2i2c.freshdesk.com/a/tickets/5571

### Ticket request type

Other

### Ticket impact

🟩 Low

### Short ticket description

This is an iteration on the same content that has been taught the last two summers. Anticipated schedule is live lecture M/W/Fri with participants completing work in GeoLab in between those sessions and in the week after. Here's the [workshop website](https://www.earthscope.org/event/2026-technical-course-using-mspass-for-data-processing-on-hpc-and-cloud-systems/).

Largely identical to July 2025 workshop!

### (Optional) Investigation results

Step through updated workshop docs in https://infrastructure.2i2c.org/howto/prepare-for-events/event-prep/

- 110 registered participants
- ~~Group-based access to Dask gateway needed~~
  - https://github.com/2i2c-org/infrastructure/pull/8558
  - `LocalCluster` recommended over gateway
- check resource allocations/packing efficiency and minimise server startup time
- nbgitpuller
- prep work from previous two years:
  - https://github.com/2i2c-org/infrastructure/issues/6278
  - https://2i2c.freshdesk.com/a/tickets/1599
  - https://2i2c.freshdesk.com/a/tickets/3500 

### Definition of done

- [x] temporarily increase IOPs and throughput and drop down after July 20
  - currently set to 10,000 IOPs and 1,000 Mb/s throughput from the last workshop
  - July 14: bumped for node root volumes to max available (throughput ~18%, from 500 to 590, IOPs by 333%, from 3000 to 10000
- [ ] coordinate workshop prep testing with interns, with extra attention to ephemeral storage which cannot be tracked with workshop prep dashboard metrics

jnywong:
I've asked follow up questions in the FD ticket regarding

- course materials + nbgitpuller
- profile option for image
- workshop prep dashboard for resource tuning

jnywong:
Assigning this to @aprilmj @agoose77 to follow up while I am away next week.

There's been no workshop dashboard prep because they are blocked on getting dask gateway working (see my latest message in FD). Last year, it looked like they were just using a dask `LocalCluster`, which could have been related to performance bottlenecks? 

> - course materials + nbgitpuller

instructor is still finalising course material and hung up on dask

> - profile option for image

image linked in FD thread, but should be updated because i think there is a dependency conflict with dask-gateway

[...]

PR #8558, opened 2026-06-19
[earthscope, *] Update scopes

jnywong:
- add scopes for mem. usage quota to staging
- update scopes for vm access
- set `admin_groups` on staging
- add dask-gateway group access
- export all groups for cost-monitoring

Ref https://github.com/2i2c-org/infrastructure/issues/8242
Cc @hamilton-earthscope

### DIFF
--- a/config/clusters/earthscope/common.values.yaml
+++ b/config/clusters/earthscope/common.values.yaml
@@ -389,13 +389,6 @@ basehub:
     scheduling:
       userScheduler:
         enabled: true
-  jupyterhub-groups-exporter:
-    config:
-      groupsExporter:
-        allowed_groups:
-        - geolab
-        - geolab:dev
-        - geolab:power
   binderhub-service:
     enabled: true
     networkPolicy:
--- a/config/clusters/earthscope/common.values.yaml
+++ b/config/clusters/earthscope/common.values.yaml
@@ -127,6 +127,16 @@ basehub:
           - geolab
           - geolab:dev
           - geolab:power
+          - geolab-admin
+          - mem-sm
+          - mem-md
+          - mem-lg
+          - vm-sm
+          - vm-md
+          - vm-lg
+          - vm-xl
+          - gpu
+          - dask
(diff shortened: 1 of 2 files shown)


hamilton-earthscope:
LGTM! Can we go ahead and apply this to the stage cluster?

jnywong:
@hamilton-earthscope i've just done a manual deploy of these configs to the staging hub.

jnywong:
These configs won't "stick" unless this PR is merged, so I'll merge this one in for now and we can open another to mirror the config for prod once you're happy with staging. Let me know if there's anything I can do to help with testing in the meantime @hamilton-earthscope 👍

jnywong:
Hey @hamilton-earthscope how is the testing on the staging hub going? I can see that the new scopes are populating the JH groups yay

Screenshot 2026-06-24 at 15 25 14

hamilton-earthscope:
Hi Jenny. They are working well from my testing. I haven't gotten any feedback from the rest of the team yet. Hopefully soon. Thanks for following up.

config/clusters/earthscope/common.values.yaml
```diff
@@ -127,6 +127,16 @@ basehub:
...
+          - geolab-admin
+          - mem-sm
+          - mem-md
+          - mem-lg
+          - vm-sm
+          - vm-md
+          - vm-lg
+          - vm-xl
+          - gpu
+          - dask
```
hamilton-earthscope, line 139: Hi @jnywong, I think with these here jupyterhub is requesting these scopes during the oauth2 exchange.

If that is their only purpose, we do not need them to be there. Our backend will inject these scopes into the access token out-of-band of the normal oauth2 exchange.

jnywong: I think so: https://jupyterhub.readthedocs.io/en/stable/explanation/oauth.html

I've applied https://github.com/2i2c-org/infrastructure/pull/8606 so that we can test anyhow

PR #8565, opened 2026-06-20
Security: Command Injection via Shell Argument in jsonnet.py

tomaioo:
## Summary

Security: Command Injection via Shell Argument in jsonnet.py

## Problem

**Severity**: `High` | **File**: `deployer/utils/jsonnet.py:L28`

The `render_jsonnet` function in `deployer/utils/jsonnet.py` constructs a shell command by directly interpolating user-controlled variables (`cluster_name`, `provider`, `account_id`, `hub_name`, `hub_domain`) into command-line arguments passed to `jsonnet`. While `shlex.join` is used for display purposes, the actual command execution via `subprocess.check_call(command)` passes a list where external variables are interpolated without sanitization. If any of these variables contain shell metacharacters or malicious content, they could alter the command execution. More critically, the `jsonnet_file` path is passed directly without validation, and if an attacker controls the path or any ext-str values, they could potentially execute arbitrary commands or read sensitive files.

## Solution

Validate and sanitize all external inputs before passing them to subprocess. Use `shlex.quote()` for string arguments, or better yet, avoid shell execution entirely by using the Python jsonnet library directly. Ensure `jsonnet_file` is validated to be within an allowed directory (path traversal check).

## Changes

- `deployer/utils/jsonnet.py` (modified)

### DIFF
--- a/deployer/utils/jsonnet.py
+++ b/deployer/utils/jsonnet.py
@@ -53,14 +53,15 @@ def render_jsonnet(
         "--jpath",
         str(jsonnet_file.parent),
         "--ext-str",
-        f"VARS_2I2C_CLUSTER_NAME={cluster_name}",
+        f"VARS_2I2C_CLUSTER_NAME={shlex.quote(cluster_name)}",
     ]
     if hub_name is not None:
-        command += ["--ext-str", f"VARS_2I2C_HUB_NAME={hub_name}"]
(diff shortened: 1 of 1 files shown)


agoose77:
Anyone with access to sufficient privileges to run `deployer` already has direct and compromising access. This tooling is not user facing, and has a different thread model to distributed code.

Ultimately this is probably the right move, but I'm going to close it on the grounds that it's a contextless PR that seems to be generated as part of a mass agent sweep across many repos.

PR #8596, opened 2026-06-24
Use conda-forge to provision dev environment

### LINKED ISSUE #7525: [Incident Action]: Constrain development environment for CI and local
### Incident Report

https://2i2c-org.pagerduty.com/postmortems/188f3b01-e222-82fd-0551-db4ca5bca436/

### What needs to happen?

Build jobs on Project Pythia were stuck pending. The root cause was a change in their pod spec that followed from a Helm regression. We do not constrain developer's Helm versions, and we should look to do this.

### What is definition of done for this item?

- [ ] A suite of in-use environment files exists for the core dependencies required for developer deployment (Nix, Brew)
- [ ] An equivalent configuration is defined and in use in CI.

agoose77:
We've recently had CI failures due to drifting dependencies (gcloud CLI). We've also historically had incidents due to local environments being unpinned (Helm). Whilst 2i2c engineers should feel free to use their own env managers, we should try to introduce a source of truth for supported tooling.

This PR uses micromamba to provision the dev tools in CI, making it much easier to reproduce. I thought about using a Docker container or Nix env, but I want this to be lightweight for our existing team to use and easy to update.

Closes #7525, and fixes CI!

### DIFF
--- a/.github/actions/setup-deploy/action.yaml
+++ b/.github/actions/setup-deploy/action.yaml
@@ -30,7 +30,7 @@ description: >-
 inputs:
   provider:
     description: Cloud provider a cluster runs on
-    required: true
+    required: false
     default: gcp
 
 # runs (for composite actions) configuration reference:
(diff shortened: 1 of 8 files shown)


agoose77:
I'm skipping deploy because I've already tested this on main. This PR just makes it easy to follow.

PR #8603, opened 2026-06-24
Update CACH image tag

Sheargrub:
Updates the image tag for Community Assets Computation Hub, per https://github.com/CIROH-UA/awi-ciroh-image/issues/114#issuecomment-4793522624. Also switches the ordering a bit so that DevCon images are better lumped together internally (since this allowed one image to slip through unlabeled before).


agoose77:
FYI: I accidentally reverted the branch base rather than HEAD, so I've cherry picked that directly into main.

PR #8604, opened 2026-06-24
[earthscope, *] Update common.values.yaml

sarahw-earthscope:
Since we are only passing the largest of a user's magic strings in each category in the acces token, vm-sizes should enable access to all servers smaller than the users max vm size.


jnywong (approved):
Brilliant, ty 🚢

jnywong:
Ref https://github.com/2i2c-org/infrastructure/issues/8242

PR #8640, opened 2026-06-29
Migrate documentation to Jupyter Book 2

### LINKED ISSUE #8655: Convert Infrastructure docs to Jupyter Book 2
We have been talking about doing this for a while. Let's dogfood the tooling that we recommend, and do so in a non-blocking way.

agoose77:
For various social and technical reasons, there are incentives to move to Jupyter Book 2 (MyST) rather than use our existing Sphinx stack.

There are missing features in the JB2 ecosystem, such as automatic Typer documentation, file-inclusion for CSV tables, etc. However, it's my contention that these are not massive blockers, and we should actually make the move in order to incentivise us to fix them!

This is not a like-for-like port — As these are internal docs, I'm not that concerned if e.g. TOCs change etc. I'd rather just ship this and iterate!

Closes https://github.com/2i2c-org/infrastructure/issues/8655


agoose77:
I'm merging this proactively, because I think it's safe to try.

PR #8642, opened 2026-06-29
[spike] Support standing-up clusters in Terraform (for AWS)

agoose77:
This PR recreates the neurohackademy cluster. 

I generally dislike the need to run `eksctl` — it's slow, and a separate workflow from our other clusters. I noticed #1924 today, and I think it still holds true.

Therefore, I'm taking this opportunity to rebuild neurohackademy using Terraform.

### To Do
- [x] Support kubelet config via launch template
- [x] Support custom labels and tags
- [x] Implement User pool(s)
- [x] Implement Core pool
- [ ] Implement Dask pool(s)
- [ ] Implement GPU pool(s)


yuvipanda:
This was one of the reasons i moved us to using managed nodegroups! If this works, great :) But do timebox yourself, and make sure upgrades are possible

yuvipanda:
Also big +1 for just using the terraform resources directly rather than using a module like https://registry.terraform.io/modules/terraform-aws-modules/eks/aws/latest. Our needs are reasonably simple so this is ok.

terraform/aws/cluster.tf
```diff
@@ -0,0 +1,326 @@
...
+  role       = aws_iam_role.node[0].name
+}
+
+resource "aws_iam_role_policy_attachment" "node-AmazonEC2ContainerRegistryReadOnly" {
+  count      = var.use_eksctl ? 0 : 1
+  policy_arn = "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly"
+  role       = aws_iam_role.node[0].name
+}
+
+resource "aws_launch_template" "core" {
```
yuvipanda, line 135: This is setting up an unmanaged nodegroup, right? We should use https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group instead which should drastically simplify this

yuvipanda (changes requested):

terraform/aws/cluster.tf
agoose77: This should be setting up managed node groups. We need to specify launch templates so that we can patch the kubelet config.

That said, I haven't tested this yet — I mainly tested that the plan looked reasonable — this was a short spike to build out the infra.

PR #8678, opened 2026-07-02
cryocloud: Update admin_users and allowed_groups

tsnow03:
Removed user 'dfelikson' from admin_users and added 'eeholmes'. Added 'CryoInTheCloud:fish-pace' to allowed_groups.

### DIFF
--- a/config/clusters/nasa-cryo/common.values.yaml
+++ b/config/clusters/nasa-cryo/common.values.yaml
@@ -68,12 +68,12 @@ basehub:
           - tsnow03
           - JessicaS11
           - jdmillstein
-          - dfelikson
           - fperez
           - scottyhq
           - jomey
           - rwegener2
           - itcarroll
+          - eeholmes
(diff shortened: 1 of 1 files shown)


config/clusters/nasa-cryo/common.values.yaml
```diff
@@ -68,12 +68,12 @@ basehub:
           - tsnow03
           - JessicaS11
           - jdmillstein
-          - dfelikson
```
yuvipanda, line 71: You probably also need to explicitly remove admin perms from the admin interface

tsnow03: Is that in Grafana? I'm not familiar with it

yuvipanda: @tsnow03 you should be able to access this in https://hub.cryointhecloud.com/hub/admin

PR #8701, opened 2026-07-08
Update cryptnono generator

yuvipanda:
Brings in changes from https://github.com/jupyterhub/mybinder.org-deploy/pull/3810

Ref https://github.com/2i2c-org/meta/issues/3329


yuvipanda:
The test failures are unrelated, and we can fix that by ripping that out

PR #8705, opened 2026-07-09
Improve incident report recurrent issue 

GeorgianaElena:
It's been raised, at least a couple of times during planning, that the recurrence of this issues clutters the Up Next column. This is especially true when the task from previous sprint is not completed.

This should fix it, by adding a new check, that validates if an open issue with the same title isn't already there.

Ref https://github.com/2i2c-org/infrastructure/issues/8720


GeorgianaElena:
Let's see if it works.

PR #8718, opened 2026-07-13
Add "continuous-dems-workshops" team to CryoCloud allowed groups list

mfisher87:
This is to support a workshop in 1 week, July 20th, and another in August. Hope this is enough notice (sorry :pleading_face:)!

@tsnow03 please confirm you approve of this change :relaxed:

### DIFF
--- a/config/clusters/nasa-cryo/common.values.yaml
+++ b/config/clusters/nasa-cryo/common.values.yaml
@@ -61,6 +61,7 @@ basehub:
           - CryoInTheCloud:cryoclouduser
           - CryoInTheCloud:cryocloudadvanced
           - CryoInTheCloud:csdms-2025-workshop
+          - CryoInTheCloud:continuous-dems-workshops
         Authenticator:
           # We are restricting profiles based on GitHub Team membership and
           # so need to persist auth state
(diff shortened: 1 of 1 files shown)


tsnow03:
I approve but @yuvipanda can you verify if we need to add these new groups above as well in two places? I believe adding just here might not be enough

agoose77:
You will also want to add these groups to `large_allowed_groups` if it is required that they can access the 60GiB/120GiB profiles.

agoose77:
And, if this group doesn't already have hub access via one of the other allowed organisations/teams, to the hub `allowed_organizations` too.

mfisher87:
> You will also want to add these groups to `large_allowed_groups` if it is required that they can access the 60GiB/120GiB profiles.

These are not required for this workshop :) 

> And, if this group doesn't already have hub access via one of the other allowed organisations/teams, to the hub `allowed_organizations` too.

Done in 5132a36c9! Thanks @agoose77 

I was following the example of `fish-pace`, so I thought this wasn't necessary. I'll admit I have no idea what I'm doing :face_with_peeking_eye: :grin:

PR #8735, opened 2026-07-14
[earthscope] Bump performance of `/tmp` disks

agoose77:
This triples IOPs, and bumps throughput. 
I think the ceiling for throughput is ~597 for the node, and the performance drops if users consume the max for too long.

We will need to clean up the `-d` generation, which is currently cordoned and the node-pool manually tainted.


jnywong:
@agoose77 i've just closed 

- https://github.com/2i2c-org/infrastructure/issues/8242

but pinging you here in case you need to do something about cleaning up the `-d` generation

PR #8743, opened 2026-07-14
Ignore continuous-image-puller for "pending" checks

yuvipanda:
Our alerts for "pending" checks have become useless with "expected" failures, primarily due to 'continuous image puller' taking 30+min sometimes. This check ignores those, as they don't cause any functional problems.


agoose77:
I'm merging this, as I assume you were waiting for CI. Thanks @yuvipanda!

PR #8769, opened 2026-07-20
[earthscope] Clean up old node generations

agoose77:
Following the workshop, this PR effects the eksctl changes to restore the original configuration.


agoose77:
I've run `kubectl cordon -l "hub.jupyter.org/node-purpose=user,2i2c/node-group-generation=e"`, and manually tainted the nodegroups.

PR #8771, opened 2026-07-20
Remove 'cost-table' transform

yuvipanda:
We no longer use this, as the Cloud Cost process is now entirely handled outside of P&S (by @haroldcampbell)


yuvipanda:
@haroldcampbell uses this!

PR #8779, opened 2026-07-21
Refactor: move `JUPYTERHUB_API_URL` into Jsonnet

agoose77:
Now that we use this in most hub configs, we can lower it to the base config — it should be harmless for non binderhubs, and more complex existing hubs already clobber this list.


agoose77:
Pinging @GeorgianaElena for review — I would like to lean in to having a second reviewer on multi-hub config changes, and you've recently touched a lot of these files!

GeorgianaElena (approved):

PR #8880, opened 2026-08-11
Revert "[2i2c-aws-us] Initial QoS spike to 2i2c-aws-us staging"

agoose77:
Reverts 2i2c-org/infrastructure#8878


agoose77:
OK, it turns out that the current Debian distributions of Ganesha do not enable QOS at compile time. We will need to build a new image.

PR #8887, opened 2026-08-12
Add OceanHackWeek teams and image to CryoCloud

abkfenris:
OceanHackWeek is using CryoCloud for the event this year. Adding OHW teams directly to help track event usage and adds the event image to the menu.

@tsnow03

### DIFF
--- a/config/clusters/nasa-cryo/common.values.yaml
+++ b/config/clusters/nasa-cryo/common.values.yaml
@@ -63,6 +63,8 @@ basehub:
           - CryoInTheCloud:csdms-2025-workshop
           - CryoInTheCloud:fish-pace
           - CryoInTheCloud:continuous-dems-workshops
+          - oceanhackweek:ohw26-organizers-fellows-mentors
+          - oceanhackweek:ohw26-participants
         Authenticator:
           # We are restricting profiles based on GitHub Team membership and
           # so need to persist auth state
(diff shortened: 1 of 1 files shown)


tsnow03 (approved):

config/clusters/nasa-cryo/common.values.yaml
```diff
@@ -63,6 +63,8 @@ basehub:
           - CryoInTheCloud:csdms-2025-workshop
           - CryoInTheCloud:fish-pace
           - CryoInTheCloud:continuous-dems-workshops
+          - oceanhackweek:ohw26-organizers-fellows-mentors
```
choldgraf, line 66: This is just adding some new github teams to the allowlist of users

config/clusters/nasa-cryo/common.values.yaml
```diff
@@ -152,6 +154,11 @@ basehub:
                 kubespawner_override:
                   # From https://2i2c.freshdesk.com/a/tickets/1537
                   image: openscapes/matlab:latest
+              06-oceanhackweek:
```
choldgraf, line 157: This adds a new environment drop-down option that's designed for OHW - I think it's safe to merge because it's just adding a new option and not changing any of the old ones.

config/clusters/nasa-cryo/common.values.yaml
```diff
@@ -165,6 +172,8 @@ basehub:
                 - CryoInTheCloud:fish-pace
                 - CryoInTheCloud:continuous-dems-workshops
                 - 2i2c-org:hub-access-for-2i2c-staff
+                - oceanhackweek:ohw26-organizers-fellows-mentors
```
choldgraf, line 175: This is giving ohw groups access to a basic resource allocation, I think it's just a natural pair to the authentication config addition above

choldgraf (approved):
I took a quick look and this seems safe to me to merge. I left comments for the changes to show my understanding of what they are and why they're safe to merge. I'll merge this!

I think the main risks might be that we're adding a new class of users to use CryoCloud's hub (and thus, hub bill) but @tsnow03 has seen this PR and is happy with it, so I think we're fine there.

abkfenris:
Thanks!

PR #8900, opened 2026-08-13
[openscapes:workshop] Bump the password again

GeorgianaElena:
The initial one was too short.


jnywong:
Ah i closed this with https://github.com/2i2c-org/infrastructure/pull/8902 because we had a PD alert on the workshop hub being unavailable. Marking as closed.

PR #8904, opened 2026-08-14
[uw-escience] New default docker image and remove small RAM profiles

scottyhq:
- Adds new choice for GO-BGC 2026 Workshop and updated existing choices. 
  - Image is for this workshop August 17-21 https://www.go-bgc.org/event/go-bgc-float-data-and-science-workshop
  - Image build repo https://github.com/go-bgc/go-bgc-dataworkshop-2026-docker

- Removed tiny RAM options which are too small for most workshop users. new default: ~7 GB RAM, ~0.9 CPUs

cc @wingho-uw

### DIFF
--- a/config/clusters/uw-escience/common.values.yaml
+++ b/config/clusters/uw-escience/common.values.yaml
@@ -77,7 +77,7 @@ jupyterhub:
             kubespawner_override:
               image: '{value}'
           choices:
-            01-GO-BGC:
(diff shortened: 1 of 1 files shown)


GeorgianaElena:
@scottyhq, we're having some problems with the CI/CD, so I've manually deployed it to both prod and staging.

PR #8905, opened 2026-08-14
[cryocloud] Add opencode and claude configuration to prod.values.yaml

tsnow03:
Added opencode and claude configuration with model details and options. I have no idea how to add the one secret in: OPENAI_API_KEY. I provided that key via email.

I copied these for opencode from @minrk, but I think our `.binder/postbuild` takes care of the `install-opencode.sh`: 
https://github.com/BIDS/hub-deploy/blob/2a0e060f930fe9ff7f2dc0e4a38ed9cb0dd791b4/hubs/demo/config.yaml#L257-L299
https://github.com/BIDS/hub-deploy/blob/2a0e060f930fe9ff7f2dc0e4a38ed9cb0dd791b4/hubs/demo/config.yaml#L229
https://github.com/BIDS/hub-deploy/blob/2a0e060f930fe9ff7f2dc0e4a38ed9cb0dd791b4/images/user/install-opencode.sh#L6

And used claude to wing its own config. Searching the internet, it looks like it should work. I based most of it on what the .json would be to be added manually by users if we don't have this.


config/clusters/nasa-cryo/prod.values.yaml
```diff
@@ -8,6 +8,51 @@ basehub:
...
+            model: nrp/kimi
+            disabled_providers:
+            - opencode
+            provider:
+              nrp:
+                npm: '@ai-sdk/openai-compatible'
+                name: NRP
+                options:
+                  baseURL: https://ellm.nrp-nautilus.io/v1
+                  apiKey: '{env:OPENAI_API_KEY}'
```
choldgraf, line 29: I'm worried that putting the key to an env variable would make the API visible to all users of a hub (not on github, but in an authenticated user session), which feels like a security risk, but it seems like something we already do for other communities so I'm assuming it's fine - eg JupyterHealth has the same pattern:

https://github.com/search?q=repo%3A2i2c-org%2Finfrastructure+openai&type=code

@tsnow03 should probably know about this though! I think any user of the cryocloud hub will be able to view and potentially exfiltrate this key

choldgraf (approved):
This seems safe-ish to merge to me, though I'm hesitant to do so without knowing more about how this hub image works. I've taken the key that @tsnow03 sent me in a DM and encrypted it with `sops` at `singleuser.extraEnv`. I've also added some comments to document where the hub image is configured for this hub.

That said, I'm not going to merge this just yet unless @tsnow03 requests that I do so now before waiting for @agoose77 to wake up. I'm pretty sure this exposes the openai API key to all users - maybe that's OK but I'd want @tsnow03 or @agoose77 to say this was safe to try before merging solo

config/clusters/nasa-cryo/prod.values.yaml
```diff
@@ -8,6 +8,54 @@ basehub:
         SCRATCH_BUCKET: s3://nasa-cryo-scratch/$(JUPYTERHUB_USER)
         PANGEO_SCRATCH: s3://nasa-cryo-scratch/$(JUPYTERHUB_USER)
         PERSISTENT_BUCKET: s3://nasa-cryo-persistent/$(JUPYTERHUB_USER)
+        OPENCODE_CONFIG: /etc/opencode-default/opencode.json
+      extraFiles:
+        # Opencode config for the NRP model provider.
+        # opencode itself is installed by the CryoCloud AI image, I think that's here:
+        # https://github.com/CryoInTheCloud/image-cryo-python-AI/blob/e80c8f17fbfcac53ddf1cd1c51f515915b31441c/appendix#L26-L29
```
choldgraf, line 15: It's not totally clear to me how users are meant to use the image that has opencode etc - are you just going to ask people to use the dynamic image builder UI to past the link to the repo themselves? We should document where that repo is in the config - I took a stab at this but you should confirm that I'm right

tsnow03:
Also, adding @minrk for review

config/clusters/nasa-cryo/prod.values.yaml
agoose77: Yes, that is a risk. I'd like to find better ways to handle API keys, but it would likely require us to support something like using a Hub as an OIDC provider (which isn't yet supported).

config/clusters/nasa-cryo/prod.values.yaml
agoose77: You would use the `Custom image` profile option, and paste the Docker image string.

agoose77 (approved):

agoose77:
This can merge, but there are review comments that @tsnow03 should be aware of :) Thanks for getting to this @choldgraf!

config/clusters/nasa-cryo/prod.values.yaml
minrk: Yes, there is a general purpose application that I think will benefit the JupyterHub community that is an auth proxy that allows JupyterHub tokens to be used to access a REST API via a _proxy_ which uses a credential to access the upstream API. 

But in the absence of that, these shared credential approaches are _okay_ in that you are sharing a single managed credential with all users. I talked to @tsnow03 about this a bit ago, and it's a question of trust and mitigation: to what extent do you trust the users of this hub with that credential (they can absolutely take this and use NRP LLM resources from their home machine), and mitigating that by rotating this key. An NRP LLM token is not the most precious resource because it is not billed and probably most cryocloud users would have access to it themselves if they bothered to go through the process of asking for an account, so there is little benefit to exfiltration _in this particular case_. If the resource protected by this token were either not likely to be available to most users of the hub and/or if there was a cost risk associated with abuse

Key rotation is not easily facilitated by storing the key in SOPS and requiring 2i2c folks to set the secret value, which Tasha has to retrieve from NRP, so this is likely to be infrequent, but can be done on a time scale Tasha feels is appropriate for mitigating that risk. For example, the hub this is copied from runs with a relatively small number of reasonably trusted users, but when we use it to host a workshop, for example, I rotate the keys immediately afterward.

minrk (approved):

choldgraf:
@minrk that makes sense to me! so it sounds like this is safe to merge because @tsnow03 understands the risk that we're taking by making the token accessible, and she's cool w/ it.

choldgraf:
> Yes, there is a general purpose application that I think will benefit the JupyterHub community that is an auth proxy that allows JupyterHub tokens to be used to access a REST API via a proxy which uses a credential to access the upstream API.

I was thinking the same thing - do you know if we're tracking that somewhere? If not then I can write up a little user story for it so we have an issue. I noticed this pattern in a bunch of places in the 2i2c hubs (e.g. accessing earthdata etc) and I suspect it'll be a useful thing to have a layer of security around

tsnow03:
Yes I approve!

tsnow03:
Thank you so much for all of your work on this @agoose77 @choldgraf @minrk!!!

choldgraf:
ok we have like 100 approvals so I'm just going to merge this 😅

PR #8948, opened 2026-08-21
[CloudBank] Retarget GPU-profile hubs' image-puller to GPU nodes only

sean-morris:


agoose77:
@sean-morris have you considered using https://z2jh.jupyter.org/en/stable/resources/reference.html#prepuller-hook-nodeselector rather than `singleuser`?

sean-morris:
Superseded by a shared `config/clusters/cloudbank/common-gpu.values.yaml` approach (one file, referenced from each GPU hub's `cluster.yaml` entry, instead of per-hub image-puller retargeting). Closing in favor of that.

PR #8950, opened 2026-08-21
[Cloudbank] Set prevent_destroy to true

sean-morris:


sean-morris:
@agoose77 I saw this [commit](https://github.com/2i2c-org/infrastructure/pull/8845/changes#diff-d6d8b52a47c105f4ced5bb5e0c0843c6510eed4be9874d0a39b22f75f736b822L14) -- maybe it does not matter and we can leave it in but thought I should bring it back.

agoose77:
@sean-morris No, I accidentally removed it whilst administering another cluster -- thank you for restoring it!

PR #8986, opened 2026-08-26
[CloudBank] Add cloudbank-pilot-hub-users API token to emich, epcc, fgcu, howard, maine, uta

sean-morris:


agoose77:
@sean-morris can you lift this token out into a separate `enc-XXX` file and add that in `cluster.yaml` to save you repeating it N times?

sean-morris:
> Member

@agoose77 I thought about asking how people felt about having one token with list and read access to all hubs and then decided I should not risk it. What do you think? I can pull this back for sure. it would be much easier to have one token.

agoose77:
@sean-morris to clarify, is this API token something that is constant for each hub? Or is it unique to each hub?

sean-morris:
@agoose77 it could be the same on each hub It is whatever we set it to.

PR #9008, opened 2026-08-29
feat: additional bucket policy

jjfrench:
## Description
- Attempting to add an additional bucket policy to one of the scratch buckets generated with the hubs
- Fresh pull request from https://github.com/2i2c-org/infrastructure/pull/8194

Health check action may look larger than expected because it would modify the `user_buckets` object in `terraform/aws/variables.tf` and  combine bucket policies in `terraform/aws/buckets.tf` -- see `src/deployer/commands/plan_upgrade/decision.py`

### DIFF
--- a/terraform/aws/buckets.tf
+++ b/terraform/aws/buckets.tf
@@ -57,6 +57,17 @@ locals {
 # ref: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document
 data "aws_iam_policy_document" "bucket_access" {
   for_each = { for bp in local.bucket_permissions : "${bp.hub_name}.${bp.bucket_name}" => bp }
+  source_policy_documents = compact([
+    try(
+      replace(
+        var.user_buckets[each.value.bucket_name].bucket_policy,
+        "__BUCKET_ARN__",
+        aws_s3_bucket.user_buckets[each.value.bucket_name].arn
+      ),
+      null
+    )
+  ])
+
   statement {
     effect  = "Allow"
     actions = ["s3:*"]
--- a/terraform/aws/buckets.tf
+++ b/terraform/aws/buckets.tf
@@ -57,25 +57,42 @@ locals {
 # ref: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document
 data "aws_iam_policy_document" "bucket_access" {
   for_each = { for bp in local.bucket_permissions : "${bp.hub_name}.${bp.bucket_name}" => bp }
-  source_policy_documents = compact([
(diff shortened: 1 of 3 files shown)


agoose77:
@jjfrench I can see that is a quick way to facilitate cross-account access via resource-based policies. Great idea to use string replacement!

@GeorgianaElena I'm getting up to speed here — have we discussed/considered switching away from bucket policies towards IAM policies instead?

Alternatively, I would possibly prefer to restructure our bucket policy into `Read only` and `Write` statements, and allow multiple principles. I'll start there whilst we discuss.

agoose77:
I've tweaked this to reduce the scope — this PR now only adds the ability to define new AWS principles for read and full privileges respectively.

agoose77:
@jjfrench are you comfortable with the more generalised policy here (taken from [`AmazonS3ReadOnlyAccess`](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonS3ReadOnlyAccess.html))?

I will set the `skip-deploy` label so that we do not trigger hub redeployments (this is infra-level only)

terraform/aws/buckets.tf
```diff
@@ -57,25 +57,42 @@ locals {
...
-        "__BUCKET_ARN__",
-        aws_s3_bucket.user_buckets[each.value.bucket_name].arn
-      ),
-      null
-    )
-  ])
+
+  # Read only
+  dynamic "statement" {
+    for_each = length(var.user_buckets[each.value.bucket_name].extra_read_only_principals) == 0 ? [] : [1]
```
jjfrench, line 63: oh nice, this is definitely much cleaner

terraform/aws/buckets.tf
```diff
@@ -57,25 +57,42 @@ locals {
...
+    content {
+      effect = "Allow"
+      actions = [
+        "s3:Get*",
+        "s3:List*",
+        "s3:Describe*",
+        "s3-object-lambda:Get*",
+        "s3-object-lambda:List*"
+      ]
```
jjfrench, line 72: >@jjfrench are you comfortable with the more generalised policy here (taken from [AmazonS3ReadOnlyAccess](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonS3ReadOnlyAccess.html))?

I think it's fine given that they're just READ permissions - although, it's probably giving a lot more permissions than what should be used

agoose77: Thanks, yes there are more in here. My read is that it's OK to generalise given that it's ultimately RO access, and MAAP are a trusted partner. I'll proceed with merging based on your assent, but we can revisit if needs be.

Issue #8398, opened 2026-05-26
MAAP/VEDA hub start times: gather metrics


aprilmj:
### Context

As part of the Improve Hub Start Experience objective - https://github.com/NASA-IMPACT/veda-jupyterhub/issues/110 - @sunu is working on a track to measure current state (after the changes @yuvipanda made at the start of May). 

Here's the Grafana dashboard Tarashish has built so far: https://grafana.maap.2i2c.cloud/d/ta2zz82/singleuser-server-spawn-time 

And feedback from @wildintellect / Science Support: they like that grafana shows a comfortable median, but would like more fine grained data (specifically outlier - what are min/max boot times & what can we learn about vs specific configs from the extremes)

### What we need to do

- [ ] A 2i2c infrastructure engineer reviews the dashboard, and if needed, meets with Tarashish to refine what logs and measurements to capture 
- [ ] We expand on the metrics @sunu captured in a way that's useful to VEDA Science Support

### Definition of Done

- [ ] Metrics are documented
- [ ] We've created any follow up cards for 2i2c or DevSeed

sunu:
I have gathered some initial metrics for average spawn duration for MAAP prod hub through Prometheus:

For last 30 days, average spawn duration:
```
sum by (namespace, status) (increase(jupyterhub_server_spawn_duration_seconds_sum{status="success"}[30d]))
/
sum by (namespace, status) (increase(jupyterhub_server_spawn_duration_seconds_count{status="success"}[30d]))

{namespace="prod", status="success"} | 79.67548104024878
{namespace="staging", status="success"} | 123.04505353589252

```

average spawn duration for 30 days before that:

```
sum by (namespace, status) (increase(jupyterhub_server_spawn_duration_seconds_sum{status="success"}[30d] offset 30d))
/
sum by (namespace, status) (increase(jupyterhub_server_spawn_duration_seconds_count{status="success"}[30d] offset 30d))


{namespace="prod", status="success"}	108.32812577403575
{namespace="staging", status="success"}	122.50017140444861
```

I think we can add some of these metrics to Grafana. Something like average spawn time over the last x days or weeks and the percentage of spawns exceeding a threshold (e.g. 30 seconds) will be useful I think.

I also need to double check what exactly is tracked as part of `jupyterhub_server_spawn_duration` metric.

aprilmj:
@sunu let me know when you're ready for a 2i2c engineer's help on this one?

sunu:
[...]

PR #9023, opened 2026-09-01
Script to analyze start time metrics from k8s events in CloudWatch

sunu:
Part of https://github.com/2i2c-org/infrastructure/issues/8398 and https://github.com/NASA-IMPACT/veda-jupyterhub/issues/126

We're now shipping Kubernetes Events to CloudWatch through Fluent Bit. This PR adds a script to analyze those events. The Kubernetes events can tell us ,per user server pod, when the autoscaler asked for a node, when the pod started, and how long each image pull took etc. The script turns that stream of individual events back into one row per spawn, so we can ask where the time actually went (node scaling, image pull etc) and how that changes over time.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -10,3 +10,7 @@ chartpress==2.*
 requests==2.*
 aiohttp==3.*
 asyncio==4.*
+# analyze-spawn-timing.py builds its CLI with click, and uses boto3 for its
+# reading events from CloudWatch
+click==8.*
+boto3==1.*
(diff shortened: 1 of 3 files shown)


dev-requirements.txt
```diff
@@ -9,4 +9,7 @@ chartpress==2.*
 # Used by scripts in the /extra-scripts folder
 requests==2.*
 aiohttp==3.*
-asyncio==4.*
```
sunu, line 12: This is unrelated to the PR. But cleaning it up since asyncio is now part of stdlib.

PR #9025, opened 2026-09-01
Run gitpuller as sidecar extra container

### LINKED ISSUE #8965: Switch VEDA Git puller to a sidecar container
This was discussed in our DDM meeting.

### Definition of done
- [ ] Move the nbgitpuller init container into a sidecar on the VEDA hubs
  - [x] MAAP
  - [ ] VEDA
  - [ ] Disasters

sunu:
Move gitpuller from initContainers to extraContainers so its image pull and repo clone/pull won't block the notebook container from starting.

Fixes https://github.com/2i2c-org/infrastructure/issues/8965 for MAAP hubs. ref https://github.com/NASA-IMPACT/veda-jupyterhub/issues/126

### DIFF
--- a/config/clusters/maap/common.values.yaml
+++ b/config/clusters/maap/common.values.yaml
@@ -20,7 +20,7 @@ jupyterhub:
       enabled: true
 
     extraImages:
-      # Keep up to date with the init container used for pulling in docs, under initContainers
+      # Keep up to date with the container used for pulling in docs, under extraContainers
       nbgitpuller:
         name: public.ecr.aws/nasa-veda/jupyterhub-gitpuller-init
         tag: 97eb45f9d23b128aff810e45911857d5cffd05c2
@@ -96,14 +96,14 @@ jupyterhub:
           # For our deployment-service-check health check user, there is no auth_state.
           if auth_state:
             pgt = f"jwt:{auth_state.get('id_token', '')}"
-            
+
             spawner.environment.update({
               "MAAP_PGT": pgt,
               "KC_ACCESS_TOKEN": auth_state.get("access_token", ""),
               "KC_ID_TOKEN": auth_state.get("id_token", ""),
               "KC_REFRESH_TOKEN": auth_state.get("refresh_token", "")
             })
-            
+
             # Inject MAAP_PGT into the s3fs sidecar container
             for container in spawner.extra_containers:
               if container.get("name") == "s3fs":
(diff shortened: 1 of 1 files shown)


config/clusters/maap/common.values.yaml
```diff
@@ -259,6 +243,40 @@ jupyterhub:
...
+      # Keep this up to date with `prePuller.extraImages`
+      image: public.ecr.aws/nasa-veda/jupyterhub-gitpuller-init:97eb45f9d23b128aff810e45911857d5cffd05c2
+      # The image's CMD clones and then exits 0, which is correct for an init
+      # container but not here. We're adding `sleep infinity` to run indefinitely
+      command:
+      - /bin/bash
+      - -c
+      - |
+        bash /opt/k8s-init-gitpuller.sh
+        exec sleep infinity
```
agoose77, line 259: I think we can just set `restartPolicy: Never` instead?

agoose77:
Is this ready for merge (besides comments?)

config/clusters/maap/common.values.yaml
sunu: @agoose77 I don't think that will work but let please me know if I'm thinking about this wrong. 

`restartPolicy` is set at pod level. It's currently set to ` restartPolicy: OnFailure` by Kubespawner and Kubespawner also [expects all containers to be in `ready` state](https://github.com/jupyterhub/kubespawner/blob/721c67d20297ffa452af7168a158fa66296608e5/kubespawner/spawner.py#L2587) to mark the pod as running. So I think we need each container to be running for the pod/user notebook server to be marked as running.

sunu: Looks like it's now possible to set `restartPolicy` at container level too: https://kubernetes.io/blog/2025/08/29/kubernetes-v1-34-per-container-restart-policy/. However I think this is too new and `ready: false` problem will still be there even with container level `restartPolicy` set.

sunu:
> Is this ready for merge (besides comments?)

@agoose77 I'm happy to merge if this looks ok. (MAAP had some workshops and events last week, so I didn't want to disrupt anything)

config/clusters/maap/common.values.yaml
agoose77: Yes, you can set per-container policies!

I had forgotten about the pod ready status including completed containers. That's unfortunate. Let's leave this as is!

PR #9028, opened 2026-09-01
[WIP] Allow sharing EBS snapshots with other AWS accounts

### LINKED ISSUE #8772: Add ability to share EBS snapshots with another AWS account
## Update
The approach we tried isn't viable because of technical complications, listed in https://github.com/2i2c-org/infrastructure/pull/9028#issuecomment-5624968036. See https://2i2c.freshdesk.com/a/tickets/6145 for follow-up
I've gotten back to the community and offered them my initial offer, which was to give them IAM accounts so they can do what they need instead. https://2i2c.freshdesk.com/a/tickets/6145

## Context

Some communities (like UC Merced) would like access to our EBS snapshots in their own AWS accounts (rather than the one the hub is on) for cold archival as well as potential migration to their own managed hub. We would like to facilitate this in the most straightforward way possible, which seems to be via [sharing an EBS volume with another AWS account](https://repost.aws/knowledge-center/share-ebs-volume). It won't be possible when using the default KMS key for encrypting the snapshots (what we do now)

To do so, we would need to:

1. Create a new KMS key that is used for just encrypting EBS volumes
2. Encrypt the EBS *snapshots* (not volumes) with this key
3. Share the KMS key with the target account (provided by the community)
4. Share the EBS snapshots with the target account (provided by the community)

## Definition of Done

- [ ] Our terraform is modified to *optionally* create another KMS key just for EBS snapshots
- [ ] Our terraform is modified to allow sharing this KMS key with another account
- [ ] Our terraform is modified to use this key for the snapshots (ideally in addition to the existing key, but ok if not)
- [ ] Our terraform is modified to allow sharing the snapshots with another AWS account
- [ ] The KMS key and snapshots are shared with the AWS account provided for us by UC Merced
- [ ] We capture documentation for ourselves and future communities on how to request and set up this feature

GeorgianaElena:
Fixes #8772

### DIFF
--- a/terraform/aws/data-lifecycle-manager.tf
+++ b/terraform/aws/data-lifecycle-manager.tf
@@ -80,6 +80,7 @@ resource "aws_dlm_lifecycle_policy" "nfs_backup" {
       }
 
       copy_tags = true
+
     }
 
     target_tags = {
--- a/terraform/aws/irsa.tf
+++ b/terraform/aws/irsa.tf
@@ -96,7 +96,9 @@ output "kubernetes_sa_annotations" {
 }
 
 resource "aws_kms_key" "cross-account-ebs-snapshot-key" {
+  count       = var.external_consumer_account_id ? 1 : 0
   description = "A KMS Key for EBS snapshots and cross-account sharing"
+
   policy = jsonencode({
     Version = "2012-10-17"
     Id      = "cross-account-ebs-snapshot-key-1"
--- a/terraform/aws/irsa.tf
+++ b/terraform/aws/irsa.tf
@@ -96,7 +96,7 @@ output "kubernetes_sa_annotations" {
 }
 
 resource "aws_kms_key" "cross-account-ebs-snapshot-key" {
-  count       = var.external_consumer_account_id ? 1 : 0
+  count       = var.external_consumer_account_id == "" ? 0 : 1
   description = "A KMS Key for EBS snapshots and cross-account sharing"
 
   policy = jsonencode({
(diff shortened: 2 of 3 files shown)


GeorgianaElena:
@yuvipanda, it looks like the snapshots are automatically encrypted with the same key as the volume.
And it's not possible to say "encrypt the snapshots with this particular key" within the DLM setup.

- **One option** would be to copy the snapshots and re-encrypt them with another kms key, then share these copies and the key. However, if we need to do this for every daily snapshot that we take, we'll have to build some form of automation that triggers whenever a new snapshot is created, then copies it and shares it with a target account (possibly with an aws lambda function).

- **Another option** would be to re-encrypt the volume with this custom kms key. But this means moving the data from one place to another I believe, as the kms encryption is an immutable volume field.

### questions
1. WDYT is the best approach?
2. Just to make sure I understood the task: we're supposed to share _all_ daily snapshots that we take, not just one particular snapshot at some moment in time?

_Side note: I also explored making a copy of the snapshot, with a custom kms key, in a different region (that the DLM supports very easily). However, to share this copy, still needs a custom workflow, as the `share_rule` in the DLM applies to the snapshot created by it and not the copy that it makes)_

yuvipanda:
Thanks @GeorgianaElena! So the problem seems to be that:

1. We're using the default KMS key for encryption of the EBS volume
2. To switch over to using a custom key, we'll need to basically take down the hub, create a new EBS volume (restore from a snapshot) with new key, switch us to use new EBS volume. This can not be done without downtime, and is more complicated than we had thought.

I'll let the community know that our initial approach that we and the community agreed on does not work, and we'll have to explore alternatives.

yuvipanda:
See https://2i2c.freshdesk.com/a/tickets/6145 for communication with the community

Issue #8822, opened 2026-07-27
[Round 1] Bump z2jh version


GeorgianaElena:
### Context

Follow-up to #6552 
We should bump a few clusters to use z2jh main and see test how `profileLists` as dicts actually work.

### What we need to do

Deploy a the latest z2jh to a few prod clusters:
- [x] 2i2c-aws-us
- [x] maap
- [x] cloudbank staging
- [x] cloudbank gpu-demo

### Definition of Done

- [x] A new version has been deployed
- [ ] Clusters define profileLists as dictionaries

GeorgianaElena:
@Gman0909, I added this to UpNext because it's important for us to try out this version as it will greatly simplify the hub's config files.
This will reduce duplicate code, make it easier to spot errors but also bespoke hub configs that may or may not be intentional. 

This is also relevant for https://github.com/2i2c-org/infrastructure/issues/8745

Gman0909:
Thanks @GeorgianaElena, noted and prioritized

jnywong:
Okay so 

- https://pypi.org/project/jupyterhub-usage-quotas/0.2.0/

is the release you want to update to `jupyterhub-usage-quotas==0.2.0` in 

https://github.com/2i2c-org/infrastructure/blob/279070045197864555e807d1cbd65185028abbf4/helm-charts/images/hub/requirements.txt#L4

but since this is a breaking change, I think you want to make a copy of the `requirements.txt` and use that in your JH6 Dockerfile so that we don't break the existing deployments (upon reflection, perhaps I should have used version name `1.0.0` 😄 ).

GeorgianaElena:
I plan to deploy this to the following clusters/hubs:
- 2i2c-aws-us (all hubs)
- maap (all hubs)
- cloudbank staging
- cloudbank gpu-demo

Unfortunately there's a problem with having `profileLists` as dicts in `kubespawner` that I'll work on fixing next

## Important info!

Instructions on how to revert if there are issues https://github.com/2i2c-org/infrastructure/pull/9042#issuecomment-5634726300

PR #9042, opened 2026-09-02
Round 1 Z2jh bump version

GeorgianaElena:
Ref #8822


jnywong:
@GeorgianaElena i've updated the `usage-quotas` package to take advantage of the new `SpawnException` feature in JH 6. I'm going to cut a release and we can update the `requirements.txt` for the Dockerfile with it :)

GeorgianaElena:
Thank you @jnywong <3

jnywong:
Posted my comment in the issue and not PR 😅 :

- https://github.com/2i2c-org/infrastructure/issues/8822#issuecomment-5572392866

GeorgianaElena:
There's a small caveat to this upgrade: since it comes with a database schema upgrade, in case something goes wrong, we can't just revert the PR to fix it. Instead we should:

1. take down the hub so that the db is not updated anymore
  `kubectl scale deploy hub --replicas=0 -n showcase`
2. create a new debug pod and attach it to the `hub-db-dir` volume
  ```yaml
  kind: Pod
  apiVersion: v1
  metadata:
    name: db-reconciliation
  spec:
    volumes:
      - name: db
        persistentVolumeClaim:
         claimName: hub-db-dir
    containers:
      - name: debugger
        image: python:3.12-slim
        command: ['sleep', '3600']
        volumeMounts:
          - mountPath: "/data"
            name: db
  ```
  
 `kubectl apply -f debugger-pod.yaml -n showcase `
3. exec into the pod and re-instantiate the db backup that was created before the upgrade
4. delete the debug pod
5. Re-deploy with `JupyterHub5`

PR #9053, opened 2026-09-04
Merge AWS SSO shell handling into regular AWS shell

agoose77:
This morning, I noticed that SSO login to nasa-veda broke. My suspicion is that SMDC / AWS have clamped down on cross-region permissions, breaking our setup.

Here's a typical non-legacy config entry for some SSO:
```ini
[profile dev]
sso_session = my-sso
sso_account_id = 111122223333
sso_role_name = SampleRole

region = us-west-2

[sso-session my-sso]
sso_region = us-east-1
sso_start_url = https://my-sso-portal.awsapps.com/start
```

My understanding is that this configuration is designed to permit account-specific profiles to share/re-use a single sso-session. Such an approach works well for SMDC, where we have N accounts under a single SSO session.

The reason our helper was failing was due to the use of `--profile` — the region from profile was being used, rather than the _SSO_ region.

This PR simplifies this whole arrangement by re-using profiles for SSO sessions as well as regular sessions. It removes the `sso-shell` command for this reason.

> [!Important]
> This PR also changes the `deployer aws shell` command, removing the `exec` prefix.

## How 2i2c needs to update their config
Ensure that each AWS Account has an entry in `~/.aws/config`, that shares a common SSO session e.g.:

```ini
[sso-session 2i2c]
sso_start_url = https://2i2c.awsapps.com/start#/
sso_region = us-east-1
sso_registration_scopes = sso:account:access

[profile reflective]
sso_session = 2i2c
sso_account_id = 916143380841
sso_role_name = AdministratorAccess
region = us-west-2

[profile nasa-cryo]
sso_session = 2i2c
sso_account_id = 574251165169
sso_role_name = AdministratorAccess
region = us-west-2
```


agoose77:
Initially I switched our implementation to creating an SSO session, but I noticed that really we can/should just use regular profiles for this!

src/deployer/dev/commands/aws.py
```diff
@@ -42,106 +43,83 @@ def setup_aws_sts_env(profile, mfa_device_id, auth_token) -> dict[str, str]:
...
+        text=True,
+        capture_output=True,
+        check=False,
+    )
+    # Valid login
+    if not result.returncode:
+        return env
+
+    # Not an SSO log-in problem, throw an error
+    if not re.search(
```
agoose77, line 83: There is no elegant way to check whether the token is valid without digging into the AWS SSO cache. I think it's "equally" hacky to just test the returned CLI strings, except it's much shorter!

src/deployer/dev/commands/aws.py
```diff
@@ -42,106 +43,83 @@ def setup_aws_sts_env(profile, mfa_device_id, auth_token) -> dict[str, str]:
...
+                str(auth_token),
                 "--profile",
                 profile,
-                "--output",
-                "json",
-            ]
+            ],
+            capture_output=True,
+            text=True,
+            check=False,
```
agoose77, line 60: All of this repeated `subprocess` boilerplate is designed to avoid leaking the fact that we're running `awscli` under the hood — it's solely for UX. 

We could just rely on `check_call` and `check_output` otherwise.

PR #9054, opened 2026-09-04
Upgrade to Helm 4

agoose77:
See https://github.com/2i2c-org/infrastructure/issues/7470

### DIFF
--- a/build-tools/environment.yaml
+++ b/build-tools/environment.yaml
@@ -7,7 +7,7 @@ dependencies:
 # Pins by compatibility
 ## Upper bound / hard pins
 - google-cloud-sdk=570.*
-- kubernetes-helm=3.17.0.*
+- kubernetes-helm>=4.0.0,<4.3.0
 ## Lower bound
 - eksctl>=0.221.*,<0.229.0
 #- azure-cli # Out of date / broken
--- a/src/deployer/infra_components/cluster.py
+++ b/src/deployer/infra_components/cluster.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import sys
 import json
 import os
 import subprocess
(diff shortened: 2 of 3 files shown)


build-tools/environment.yaml
```diff
@@ -7,7 +7,7 @@ dependencies:
 # Pins by compatibility
 ## Upper bound / hard pins
 - google-cloud-sdk=570.*
-- kubernetes-helm=3.17.0.*
+- kubernetes-helm>=4.0.0,<4.3.0
```
GeorgianaElena: @agoose77, before looking into it myself, why not `4.3.0`? Does it come with anything weird?

agoose77: My rationale for the build environment is that we should be using similar versions for deployment (i.e. narrow pins). We don't want to be too strict, for cross-platform compatibility (like, if Helm 4.X.Y isn't available on Brew, and you're using Brew).

However, I notice that these limits should be updated to reflect our known-good version of Helm (from https://github.com/helm/helm/commit/c51456c45d4782254ce3ba83bcbbd5e78a0f0070)

We should set `>=4.2.0` as our lower bound, and choose a reasonably narrow upper bound (4.3.0 seems fine to me).

agoose77:
```suggestion
- kubernetes-helm>=4.2.0,<4.3.0
```

agoose77:
This is blocked by the upstream bug remaining un-fixed (see reproducers here: https://github.com/agoose77/reproducer-helm-merge-changes)

We encounter this bug in places where:
1. The base Helm chart (e.g. z2jh) has a map.
2. Our `basehub` chart adds a key.
3. Our specific-hub values nulls that key.

This is `test-3` in the reproducer, which gives (3.17.0 vs `latest`)
```yaml
---
# Source: parent/charts/child/templates/service.yaml
config:
  node_selector: {}
---
# Source: parent/charts/child/templates/service.yaml
config:
  node_selector:
    label: null
    ```

PR #9069, opened 2026-09-09
[deployer] Add commands to retrieve info about versions

GeorgianaElena:
## Motivation

This was motivated by the fact that we need to do a lot of staged deployments in the close feature and we don't have a way to easily know what hub is running what version of a thing we are upgrading and rolling out.

## What the PR does:

This adds new deployer commands under `config` that list what k8s version and what z2jh version each hub is running.

There's also one command that merges all of these into a single matrix that's meant to present a snapshot of various versions that are are running.

Screenshot 2026-09-09 at 15 22 06
Screenshot 2026-09-09 at 15 21 02

## Improvements

There's obviously room for improvements and would also be useful to:
- get a list of hubs that don't run a particular version for example
- get this tables somewhere public (maybe where we have the list of running hubs?)


yuvipanda:
This makes all deployer commands fail for me, with the following error:

```
➜ deployer deploy-support nasa-cryo
  File "/Users/yuvipanda/.local/share/virtualenvs/infrastructure/bin/deployer", line 3, in <module>
    from deployer.__main__ import main
  File "/Users/yuvipanda/code/infrastructure/src/deployer/__main__.py", line 12, in <module>
    from .dev import commands as dev_commands  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yuvipanda/code/infrastructure/src/deployer/dev/commands/__init__.py", line 5, in <module>
    import deployer.dev.commands.config.get.k8s_version  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yuvipanda/code/infrastructure/src/deployer/dev/commands/config/get/k8s_version.py", line 5, in <module>
    import _jsonnet
ModuleNotFoundError: No module named '_jsonnet'

```

Where's the `_jsonnet_ module coming from?

GeorgianaElena:
@yuvipanda, it's under dev requirements https://github.com/2i2c-org/infrastructure/blob/426f03878d1d1c0807fbfed9beda9c0915e500a7/pyproject.toml#L40

yuvipanda:
@GeorgianaElena ah, ty! i was just doing `pip install .`, needed `pip install .[dev]`

PR #9075, opened 2026-09-11
Cleanup how dashboards are deployed

yuvipanda:
- Clone everything into a TemporaryDirectory for automatic cleanup
- Use explict paths everywhere, so we aren't relying on relative paths and current directories
- Don't explicitly clone grafonnet, as that is also now vendored in https://github.com/2i2c-org/jupyterhub-cost-monitoring/pull/130

I'm not touching some of the other structures here, but wanted to do the minimal to support https://github.com/2i2c-org/jupyterhub-cost-monitoring/pull/130

### DIFF
--- a/src/deployer/commands/deploy_dashboards.py
+++ b/src/deployer/commands/deploy_dashboards.py
@@ -1,6 +1,7 @@
 import os
 import shutil
 import subprocess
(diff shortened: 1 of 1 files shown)


jnywong (approved):
Brilliant, ty

Issue #8879, opened 2026-08-10
[Support] LP DAAC and NOAA Openscapes workshop September 14-18


aprilmj:
### The Freshdesk ticket link

https://2i2c.freshdesk.com/a/tickets/6051

### Ticket request type

Feature Request

### Ticket impact

🟩 Low

### Short ticket description

This is a request for the NMFS Openscapes workshop hub - it needs to be ready ahead of the workshop (so I set an end date one week before, September 7)

ateucher:
Hi @aprilmj, I don't think the password has updated for this workshop yet. The previous one still works; can you let me know when it has been set?

Thanks so much!

PR #9078, opened 2026-09-11
[nmfs-openscapes] restore image and resource options in workshop hub

ateucher:
The last workshop replaced all of the image and resource options in the nmfs-openscapes workshop hub. For a workshop next week (See #8879) we need the other options back. So sorry for the late nature of this request!


jules32:
I'm tagging @ronnyhdez here from the Openscapes team as well, for awareness

ateucher:
Thanks so much @yuvipanda for the quick merge! ❤️

PR #9096, opened 2026-09-16
Fix invalid workflow: cancelled()

sean-morris:


sean-morris:
@agoose77 I think these three lines fix the error related to testing for a cancelled job in the deploy-hubs workflow with these changes

agoose77:
Interesting, I suspected this might fail. The GitHub context docs are not great. Reading the docs suggested that the job status would not be set to `cancelled`. However, my new approach to figuring out what's possible is to use GitHub Code searches, and I can see thousands of examples of this syntax. Let's try it!

PR #9106, opened 2026-09-18
[CloudBank] Pin new hubs' db PVC to a WaitForFirstConsumer storage class

sean-morris:


config/clusters/cloudbank/chaminade.values.yaml
```diff
@@ -19,6 +19,13 @@ jupyterhub:
           name: Chaminade University of Honolulu
           url: https://chaminade.edu/
   hub:
+    db:
+      pvc:
+        # Default SC 'standard' is Immediate-binding, so it provisions the
+        # disk before scheduling and can land outside us-central1-b, where
+        # core-pool lives -- leaving the hub pod permanently unschedulable.
+        # standard-rwo is WaitForFirstConsumer, so the disk follows the pod.
+        storageClassName: standard-rwo
```
agoose77, line 28: We should probably do this for all of our hubs on GCP, right @GeorgianaElena? Actually, I should check what the default storage class is.

yuvipanda: +1 I think that is a good idea. It may be the default for new clusters, so maybe this just needs fixing for older clusters

PR #9109, opened 2026-09-18
maap: Make `/tmp` larger for larger nodes

yuvipanda:
This is a temporary way to give some users larger `/tmp`, by having them use a larger resource allocation. Quite wasteful, as there's no way to accomodate users who just need larger /tmp for a short period of time. But this unblocks users temporarily, so let's do this.


yuvipanda:
Closed in favor of https://github.com/2i2c-org/infrastructure/pull/9112

Issue #67, opened 2020-10-12
Document whether / how user data is backed up


choldgraf:
Some folks have asked questions like "should I tell my students to create backups of their work in case their data gets lost?" What should I respond in those cases? Does our NFS server have backups set up? If not, should we recommend people download `.zip`s of their files occasionally? Have we ever had an issue where we lost user storage data?

yuvipanda:
We now do 7day daily 'disaster recovery' backups. However, I think this isn't good enough, since we don't wanna do a lot of work restoring this - especially for recovering a single user's work.

I think we should:

1. Make it easy and obvious for a student to download their entire home directory
2. Make it possible for an admin to get the home directory of all students.

choldgraf:
I'm gonna close this one and we can track it in https://github.com/2i2c-org/pilot-hubs/issues/641

PR #9112, opened 2026-09-19
Enable users to get larger `/tmp` per-session

yuvipanda:
image

Kubernetes now has the concept of an 'ephemeral'
volume (https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/) that lets you tie a PVC to a Pod's lifecycle! This allows us to easily create a temporary volume that can be attached to a specific pod, and then removed when the pod dies automatically, without us having to write a lot of code. Yay!

This is perfect for larger per-user `/tmp`. In this PR, we implement this on the MAAP staging hub.

1. Users get a dropdown that allows them to pick the larger `/tmp`. The default is still the standard, as we don't pay anything extra for that in cloud cost.
2. If users pick the larger `/tmp`, we add a generic extra ephemeral volume of the selected size, with our own custom storageclass, and attach that to `/tmp`.
3. Kubernetes takes care of provisioning, as well as deleting the volume when the user is done

Because this is done via a dropdown in profiles, we are able to restrict this based on user groups in the future as needed!

A VolumeAttributesClass is created because we want to tag this as belonging to this hub for cost purposes, as well as allow customization of the disk (type, iops, etc) easily. A StorageClass is immutable and makes setting these attributes more difficult, especially if we already have a running hub with active users. Ideally we would be able to tag the username too, but at least on AWS that does not work due to limitations in what kinda tags can be set.

This should unlock larger `/tmp` for everyone without needing larger nodes for everyone.

Implements most of https://github.com/2i2c-org/initiatives/issues/61

TODO

- [x] Test on GCP
- [ ] Add how-to doc for enabling
- [ ] Add doc on user facing site


yuvipanda:
Something strange on GCP when I try, as provisioning the PVC fails with `Disk type pd-balanced does not support dynamic provisioning`. Even if I set type to `pd-ssd` in the volume provisioning. It's possible that I need to understand storageclass vs volumeattributesclass better.

yuvipanda:
Figured it out! Have it working on GCP too :)

I couldn't find definitive documentation about storageclass vs volumeattributesclass, but have settled on using storageclass to specify the type (ssd vs standard, etc) and volumeattributesclass for anything else. since we want to specify the aws tags on AWS, there's now a volumeattributesclass set for AWS. But we don't need to configure anything else on GCP right now, so i've just let that be empty.

agoose77:
Something we'll need to explicitly consider here is the minefield of perf and cost/perf: I considered dynamic disks for fast scratch when first implementing #67 for CIROH(?) but found it was better to have node-level large disks. I can't remember whether that was because:
- small fast disks are more $$$ than large disks (especially/or only) if you only need to guarantee 60% of users need peak `/tmp` at any time.
- disk perf is tied heavily to node instance perf limits, and these are complex.

I'm only driving by this PR, and so this is just a placeholder comment — it might be fine. I just remember not choosing this for CIROH, so flagging it.

yuvipanda:
@agoose77 Were you able to observe the actual loads for user needs being slower than what a dynamic disk could handle? At least on AWS, IIRC all our node level disks are also just EBS so this shouldn't be any different there. I thiwould imagine that the need for super-high-perf node-local disks is pretty minimal, and we can always bump up the iops / throughput here if needed. On GCP that can be hyperdisk if needed. 

The big cost saving here is that we don't have to have these nodes be 'always on' just for some folks to use `/tmp`. We can only bring them online when needed.

agoose77:
> The big cost saving here is that we don't have to have these nodes be 'always on' just for some folks to use /tmp. We can only bring them online when needed.

Assume this is a typo, and reading this as "disks be 'always on'"? Agreed this is a major boon of using ephemeral volumes.

OK, I did some Slack digging and I found the thread that I was thinking about w.r.t whether we have concerns here: https://2i2c.slack.com/archives/C07SJJWVCAD/p1779553159176929
I also re-read DMs with @yuvipanda around CIROH's workshop. 

At the time, we were avoiding rolling out _any_ new features that weren't trivial (like `emptyDir`-trivial), so we discarded ephemeral volumes then. That's not the case here, where we're building out new features on purpose. So my gut-feeling caution was about a previous decision to avoid risk ahead of a workshop, rather than a problem with disk performance.

I've gone back through the EBS and GCP docs, and here's the TL;DR:

- The max perf we can grant these volumes is governed by the _instance limit_. This bounds the sum of all attached disks performance.
- GCP has higher "max throughput" with `pd-*` on `N2-*-16`[^gcp] nodes than AWS does with `gp3` on `r5.4xlarge`[^aws]
- `pd-*` on GCP has performance governed by size, so we will have to use large volumes for good performance. This contrasts `hyperdisk` and `gp3` volumes where these numbers are tunable.
- `hyperdisk` requires special node types that we don't normally use on GCP

[^aws]: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-optimized.html#current-memory-optimized:~:text=r5%2E4xlarge,18750
[^gcp]: https://docs.cloud.google.com/compute/docs/disks/performance#pd-balanced_13:~:text=16%2D31,1%2C200

There's probably more I forgot to summarise here. As it stands, my view is that using ephemeral volumes is much nicer than using `/tmp` because it better handles heterogeneous node packing (i.e. different sized user pods on the same nodes). I don't see any advantage for using a common `/tmp` vs smaller ephemeral volumes apart from "peak performance" perspective[^peak], and it's definitely preferable to use a per-user ephemeral volume for workshops and classes, where we likely prefer steady-state to peak perf. Big node `/tmp` is better if we use pd-disks and really care about maximising burst performance vs cost, but not much else.

[^peak]: It's cheaper to offer max throughput on a big shared disk than guaranteeing each user that performance on their own disk. The downside is that only one person can use the peak performance of a big disk before it's amortised over all users. I.e. burst vs steady-state behaviour. Workshops much prefer "give everyone their own disk".

helm-charts/support/templates/storageclass/scratch-disk.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: storage.k8s.io/v1
+kind: VolumeAttributesClass
+metadata:
+  name: scratch-disk
+driverName: {{ .Values.scratchDiskVAC.driverName }}
+{{ with .Values.scratchDiskVAC.parameters }}
+parameters:
+  {{ . | toYaml | nindent 2}}
+{{- end}}
+{{- end}}
```
agoose77, line 11: I'd be tempted to make this template generate several classes so that we can have multiple kinds of scratch. It's easy to do that later on, though.

agoose77 (approved):

yuvipanda:
@agoose77 i'm so sorry, i think i lost some of your commits when i rebased / force pushed.

agoose77:
Ah no worries! I use `--force-with-lease` after being burned by that a number of times.

I have the refs locally, I'll push them.

